PR #1, opened 2026-05-26
fix: portable in-place sed across 5 shell scripts (Linux compat)

AuthenTech-AI:
## Problem

Five shell scripts in this repo use `sed -i ''` (the BSD/macOS form, where `''` is the empty backup-suffix argument). On GNU/Linux, `sed -i` does not take a separate suffix argument — the `''` is parsed as the sed script, so the actual sed expression that follows becomes a filename, and the in-place edit silently no-ops (the `2>/dev/null || true` patterns swallow the resulting error).

This means several scripts that look like they work on Linux actually don't perform their in-place edits:

- `kipi-new-instance.sh` — the subtree path-nest transform (`/q-system/hooks/` → `/q-system/q-system/hooks/`) never runs, so new instances on Linux end up with broken hook paths in `.claude/settings.json`.
- `kipi-update.sh` — same path-nest transform never runs on Linux instances.
- `fix-imports.sh`, `q-system/.q-system/scripts/instance-diet-fix.sh` — broken-import cleanup silently skipped on Linux.
- `q-system/.q-system/agent-pipeline/templates/create-from-template.sh` — `{{DATE}}` and `{{OUTPUT_NAME}}` placeholders never get replaced in template output on Linux.

## Fix

Add a tiny `sed_inplace()` helper near the top of each affected script that branches on `uname` and call it instead of raw `sed -i ''`:

```bash
sed_inplace() {
  if [[ "$(uname)" == "Darwin" ]]; then
    sed -i '' "$@"
  else
    sed -i "$@"
  fi
}
```

Per file:
- `fix-imports.sh` — 1 call site
- `kipi-new-instance.sh` — 3 call sites
- `kipi-update.sh` — 2 call sites
- `q-system/.q-system/scripts/instance-diet-fix.sh` — 1 call site
- `q-system/.q-system/agent-pipeline/templates/create-from-template.sh` — 2 call sites, refactored from `find -exec` to `while read` because `-exec` spawns a new process that can't see shell functions

## Behavior

- **macOS:** no change — still calls `sed -i ''` exactly as before.
- **Linux:** the previously silently-broken in-place edits now actually happen.

## Verification

Each script syntax-checked with `bash -n`. Helper sourced and smoke-tested on Linux Mint 22.1 — `sed_inplace 's/hello/HOWDY/'` correctly rewrites the target file. The refactored find-loop in `create-from-template.sh` verified to substitute both `{{DATE}}` and `{{OUTPUT_NAME}}` placeholders.

## Context

Discovered while running `kipi-update.sh` on a Linux instance — the script reported success but produced a `settings.json` with single-nested hook paths instead of the intended double-nested form. Local fix lived for a day on a private fork; sending upstream so other Linux users don't get bitten.

### DIFF
--- a/fix-imports.sh
+++ b/fix-imports.sh
@@ -1,6 +1,15 @@
 #!/bin/bash
 set -euo pipefail
 
+# Portable in-place sed (BSD/macOS requires `sed -i ''`, GNU/Linux requires `sed -i`)
+sed_inplace() {
+  if [[ "$(uname)" == "Darwin" ]]; then
+    sed -i '' "$@"
+  else
+    sed -i "$@"
+  fi
+}
+
 for f in \
   ~/Desktop/ktlyst-hub/strategy/CLAUDE.md \
   ~/Desktop/ktlyst-hub/product/CLAUDE.md \
@@ -10,7 +19,7 @@ for f in \
   ~/Desktop/ktlyst-hub/lawyer/CLAUDE.md
 do
   if [ -f "$f" ]; then
-    sed -i '' 's|@q-system/q-system/CLAUDE.md|@q-system/CLAUDE.md|g' "$f"
+    sed_inplace 's|@q-system/q-system/CLAUDE.md|@q-system/CLAUDE.md|g' "$f"
     echo "fixed: $f"
   else
     echo "skip: $f"
--- a/kipi-new-instance.sh
+++ b/kipi-new-instance.sh
@@ -10,6 +10,15 @@ SKELETON_REMOTE="https://github.com/assafkip/kipi-system.git"
 SKELETON_BRANCH="main"
 PREFIX="q-system"
 
+# Portable in-place sed (BSD/macOS requires `sed -i ''`, GNU/Linux requires `sed -i`)
+sed_inplace() {
+  if [[ "$(uname)" == "Darwin" ]]; then
+    sed -i '' "$@"
+  else
+    sed -i "$@"
+  fi
+}
+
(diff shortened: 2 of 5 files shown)


AuthenTech-AI:
Validation report against this branch (`fix/portable-sed-inplace`):

- Installed the documented git hooks: `git config core.hooksPath .githooks` (per `CONTRIBUTE.md` step 4)
- Ran `CI=true python3 validate-separation.py 5 --verbose` — matches what GitHub Actions runs

Result:
```
VALIDATION SUMMARY (Phase 5)
==============================
  PASS: 47
  FAIL: 2
  WARN: 0
```

The 2 failures are **pre-existing on `main`** and not introduced by this PR — `git diff upstream/main..HEAD -- <those-files>` is empty. They're both flavors of "KTLYST references in skeleton code":
- `q-system/.q-system/scripts/stat-registry-extract.py` — hardcoded `q-ktlyst/canonical/...` in a stderr message
- `q-system/.q-system/scripts/tests/test-stat-verify.py` — `ktlyst-spec`/`ktlyst-acv` test fixture IDs + docstring flavor

Both are cleared by the companion PR #3. With #3 merged, this branch goes fully green: `PASS 49 / FAIL 0`.

`CI=true` is needed locally because Phases 2-4 check for the maintainer's specific instance directories (`KTLYST_strategy`, etc.) which aren't present on contributor machines — same behavior as the `validate.yml` workflow.

Branch is rebased on `upstream/main` (no divergence). Ready for review when you have a moment.

PR #4, opened 2026-07-01
fix: exempt lessons scrubber from skeleton leak scan

assafkip:
## Summary
- Extends the existing denylist self-reference exemption to `lessons_scrub` and its test.
- Keeps the skeleton leak gate strict for normal scripts while allowing deterministic leak detectors to name the tokens they block.

## Verification
- `python3 validate-separation.py 1 --verbose`
  - PASS: 60
  - FAIL: 0
  - WARN: 1 advisory manifest warning

### DIFF
--- a/validate-separation.py
+++ b/validate-separation.py
@@ -214,11 +214,11 @@ def phase_1():
     build_sched = os.path.join(SCRIPT_DIR, "q-system", "marketing", "templates", "build-schedule.py")
     check("build-schedule.py exists and is non-empty", file_exists(build_sched) and os.path.getsize(build_sched) > 0)
 
-    # No KTLYST in scripts. Exclude the lessons-validator denylist machinery
-    # (the leak-detector and its test): a denylist must name the tokens it
-    # blocks, so it legitimately contains them. Same self-reference exemption
-    # this validator grants itself in the full sweep below.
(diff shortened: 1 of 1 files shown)


assafkip:
Closing as already landed.

This PR adds `lessons_scrub` / `lessons-scrub` to two exclusion sets in `validate-separation.py`. Current `origin/main` already carries both, verbatim:

- `script_exclude = ("lessons-validator", "lessons_scrub", "lessons-scrub")` (validate-separation.py:799)
- `exclude_files = {... "lessons-validator", "lessons_scrub", "lessons-scrub", ...}` (validate-separation.py:967)

So merging this would be a no-op on the exclusion sets. The branch is also 514 commits behind main and currently CONFLICTING.

One correction for the record, since this PR gets cited as the fix for the red `validate` check: it is not the same fault. The failure that had `Skeleton Validation` red on every branch was three founder-specific identifiers in code comments tripping the Phase 1 purity gates (`pr-review-agent.sh`, `linear-worker.sh`, `test/test-worker-project-scope.sh`). That was fixed in ce2073c and main is green. This PR loosens the validator; ce2073c fixed the source files and left the gates alone.

Branch `codex/fix-skeleton-validation-lessons-scrub` is preserved. Reopening costs nothing if I have read this wrong.

PR #5, opened 2026-07-02
fix(prompt-only-enforcement-guard): mention-vs-claim false positive (sp-edf9395d)

assafkip:
## What

The guard flagged 11 lines of its own PRD — including pure frontmatter (`id:`/`title:`/`status:`) — for merely *mentioning* enforcement, even though the doc names its executable blocker throughout. Spillover `sp-edf9395d`.

Root cause: the deterministic-blocker suppressors read the same ±2-line window as the trigger regexes, so a hook/test named 3+ lines from a claim-shaped sentence was invisible to the suppressor. Frontmatter metadata (`title: Prompt Only Guard Stderr`) also matches subject+action patterns despite being a description, not a claim.

## Fix (the two mitigations named in the spillover entry, both minimal)

1. **Leading YAML frontmatter is exempt** — metadata describes the doc, it does not claim enforcement.
2. **Dual windows**: triggers + negation keep ±2; `GUARD_FILENAME` + `DETERMINISTIC_RE` suppressors read ±6. Negation stays narrow on purpose — a distant "must not" cannot mask a real claim.

No changes to the regexes, target extensions, skip marker, or exit-code contract.

## Verification

- New test sections in `test-prompt-only-enforcement-guard-stderr.sh` shown **red first** against the unfixed script (frontmatter fixture reproduced all 5 metadata flags).
- Wider-not-document-wide pinned: a blocker named 8 lines away still blocks.
- Both real FP documents verified passing with their skip markers stripped: the archived PRD `prd-prompt-only-guard-stderr-2026-07-02` and today's plan file (which the guard blocked live on first write — a second reproduction).
- All 21 scripts in `q-system/.q-system/scripts/test/` pass.

## Notes

- Branch includes 4 commits already on local main (token-guard + stderr fixes); the diff collapses to this one commit once main is pushed.
- `sp-edf9395d` stays open in the ledger until this ships through the gated issue flow; this branch is the fix + reproducer.
- Fleet propagation: `kipi update` after merge ships the script to instances.

### DIFF
--- a/.prd-os/issues/prompt-only-guard-stderr.md
+++ b/.prd-os/issues/prompt-only-guard-stderr.md
@@ -1,0 +1,26 @@
+---
+id: prompt-only-guard-stderr
+title: prompt-only-enforcement-guard block message must reach stderr (sp-cd530cc7)
+status: closed
+priority: p2
+parent_prd: prd-prompt-only-guard-stderr-2026-07-02
+allowed_files:
+  - q-system/.q-system/scripts/prompt-only-enforcement-guard.py
+  - q-system/.q-system/scripts/test/test-prompt-only-enforcement-guard-stderr.sh
+disallowed_files: []
+required_checks:
+  - bash q-system/.q-system/scripts/test/test-prompt-only-enforcement-guard-stderr.sh
+required_reviews: []
+bypass_check: "bash q-system/.q-system/scripts/test/test-prompt-only-enforcement-guard-stderr.sh"
+---
+<!-- generated-by: prd_split.py prd=prd-prompt-only-guard-stderr-2026-07-02 finding=finding-1 at=2026-07-02T23:16:54Z -->
+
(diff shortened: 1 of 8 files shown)


assafkip:
Closing as superseded.

This PR (2026-07-02) narrows one false-positive class in the prompt-only guard by widening the window that blocker-suppressors read. Since then:

- `main` has evolved its own false-positive handling in that same file (`NEGATION_TERMS`, plus newly-introduced-claims-only scoping at lines 198 and 309). This branch's variant is an older shape aimed at a file that has since moved underneath it.
- The remaining problem is now tracked more broadly by **ASK-249**, "two false-positive classes (negation vocabulary, diagnostic vs assertive)". That issue covers the class this PR fixed plus a second one.
- The branch is 490 commits behind main, and `validate` is red on it.

Better redriven off current main under ASK-249 than merged from a two-month-old divergence. The head branch is preserved, so reopening costs nothing.

PR #12, opened 2026-07-27
ASK-183: migrate com.kipi.linear-dor to Linear-tracked execution

assafkip:
Closes the migration work for ASK-183. **Do not merge without review** — closeout runs through `/issue-verify` and `/issue-closeout`.

## What the DoR asked for

`com.kipi.linear-dor` runs under launchd on a known schedule, its failures reach Linear rather than dying in a log, its state is recorded, and it has been run once and observed to do correct work.

| Bar | Before | Now |
|-----|--------|-----|
| 1. launchd, never cron | held | held (plist unchanged) |
| 2. failures reach Linear | **failed** | `report_failures()` |
| 3. live or in the pause ledger | held | verified: `launchd-dark: 0`, `launchd-failing: 0` |
| 4. verified by running it | **failed** | ran it, and it found a live defect |

## Running it is what found the bug

The job was loaded, scheduled at 03:00, and reporting `LastExitStatus = 0`. Its declared log files, `~/.config/kipi/linear-dor.out` and `.err`, did not exist — it had never actually run under launchd. The first real `launchctl kickstart`:

```
ASK-147: claude failed (FileNotFoundError)   [x8, ASK-147..ASK-137]
dor-drafter: drafted 0, 122 still lack a DoR
launchctl list com.kipi.linear-dor -> LastExitStatus = 0
```

Eight of eight drafts dead, and launchd recorded a clean exit.

**Root cause:** the plist runs `/bin/bash -lc`. The `-l` sources *bash* login files; the founder's shell is zsh, so the PATH entry for `~/.local/bin` never loads under launchd. Every interactive run inherits PATH from the parent shell and passes, which is why reading this job could never have found it. Only running it under its own scheduler did.

## The two changes

**`report_failures()`** — the drafter exits 0 on every path so launchd does not restart-loop it, which also means `launchd-health-check.py` (it keys on a non-zero `LastExitStatus`) can never see a bad night here. Exit 0 stays; the run now reports itself. `main()` collects each failure instead of only printing it, and files them onto ONE permanent Linear issue, deduped by the same `kipi-key` marker `fleet-health-daily.py` uses, with a comment per failing run rather than a fresh issue every night. The reporter never raises — it runs in the last breath of a launchd job.

**`claude_binary()`** — resolves the CLI (PATH, then known install locations) instead of trusting the scheduler's environment. A missing binary is now a reported failure, not a per-issue traceback.

Neither touches what the job does — which issues it picks, the prompt, or how it writes. That was the DoR's "Not doing" line.

## Verification

RED first, then green, both observed:

```
$ python3 q-system/.q-system/scripts/test/test-linear-dor-failure-reporting.py
AttributeError: module 'dor' has no attribute 'report_failures'     # RED
...
all linear-dor failure-reporting checks passed                       # 22 checks
```

The PATH fix re-run under launchd's actual environment:

```
$ PATH=/usr/bin:/bin:/usr/sbin:/sbin  python3 .../linear-dor-drafter.py --limit 1 --apply
dor-drafter 2026-07-27T06:19:55Z: 122 issue(s) lack a Definition of Ready
  drafted ASK-147  double-scheduled: /Users/assafkipnis/projects/cole-gtm/proje
dor-drafter: drafted 1, 0 failed (none), 121 still lack a DoR
rc: 0, stderr empty
```

Supporting checks:

```
$ python3 q-system/.q-system/scripts/capability-gate.py --repo-root . --check-only
declared: 64 tests ... capability-gate: GREEN

$ python3 q-system/.q-system/scripts/fleet-health-daily.py --quiet
launchd-dark: 0 / launchd-failing: 0
```

The unit test is registered in `capability-manifest.json`, so `capability-gate.py` runs it. A test no runner knows about is the silent-absence class that gate exists to catch.

## Spillover captured

- **`sp-684548f2`** — every job-migration DoR (ASK-151..ASK-186) tells the operator to run `launchd-health-check.py --dry-run`, but the script tests `"--dry" in sys.argv` by exact match. `--dry-run` therefore runs **live**: it Slack-pings and overwrites `~/.config/kipi/launchd-health-state.json` during what reads as a dry run.
- **`sp-14164e33`** — the same launchd PATH bug is latent in every other plist that shells `claude`.

Both leave `prd_runner.py gates run` red by design until they are worked. That is the no-orphan-findings mechanism, not a regression from this branch.

## Blast radius

Machine-local for the plist (LaunchAgents do not propagate via `kipi update`). The two script changes and the test are in the synced `q-system/` tree and reach the fleet on the next update.


assafkip:
# Review: PR #12 — ASK-183 linear-dor migration

Read the diff, read the real producer (`linear-sync.py`), loaded the branch code in-memory and ran five reproducers. Findings below, each with output I actually got.

---

## MAJOR — once the failure issue is closed, every later failing night is silent

`q-system/.q-system/scripts/linear-dor-drafter.py:224`

The comment path dedups on the `kipi-key` marker and never looks at issue state. It cannot: `fetch_remote_state`'s `PROJECT_ISSUES_QUERY` (`linear-sync.py:392`) filters on project only, has no state filter, and does not even *select* state. The ledger keeps the key forever. So the first time the operator closes the failure issue — which the PR's own title (`linear-dor: nightly DoR drafter reported failures`) invites, since you close it after fixing the PATH bug — every subsequent failing night becomes a comment on a `Done` issue. No open board object, no reopen, forever.

Note `fleet-health-daily.py:437` *skips* an existing key (`result["existing"] += 1`, no write). The comment-on-existing behavior is new here, and it is the half without a state check.

```
$ python3 - <<'PY'   # loads FETCH_HEAD drafter, fixture mirrors linear-sync's real return shape
night 1  -> created
  reported 8 failure(s) on ASK-1000
night 40 -> commented
  new issues created : 0
  comments posted    : 1 -> onto Done ASK-1000
  reopen mutations   : 0
  state ever queried : False

CONTRACT: a failing night leaves an OPEN board-visible signal -> False
AssertionError: FAIL: 8 dead drafts reported only as a comment on a Done issue
```

3am action for a human: none, because nothing surfaces. Recoverable by reopening, which is why this is major and not blocker.

---

## MAJOR — when the reporter itself fails, the run reports success and nothing records it

`linear-dor-drafter.py:259` and `:317`

Linear down or rate-limited at 03:00 is the routine unattended case. The failure list is dropped with a stderr line into `~/.config/kipi/linear-dor.err` — the file this PR's own description establishes did not exist and nobody reads — plus `failures_reported: "unreachable"` in `linear-dor-state.json`. `slack-notify.sh`, which `.claude/rules/founder-notifications.md` makes the single channel for exactly this, is not called.

```
$ python3 - <<'PY'   # team lookup succeeds, Linear 503s before the report
  linear unreachable, 8 failure(s) NOT filed: HTTPError 503 Service Unavailable
dor-drafter: drafted 0, 8 failed (unreachable), 8 still lack a DoR

exit code launchd sees : 0
state file content     : { "ran_at": "...", "drafted": 0, "failed": 8,
                           "failures_reported": "unreachable", "remaining": 8 }
```

The state file has no reader anywhere in the repo:

```
$ grep -rl "linear-dor-state" . | grep -v "^./.git/"
q-system/.q-system/scripts/linear-dor-drafter.py
```

The writer is the only mention. Exit 0, green `launchctl`, 8 dead drafts, zero signal — the same shape the PR set out to close, one level up.

---

## MINOR — the test asserts where `claude` is installed, so it goes red on a healthy machine

`test/test-linear-dor-failure-reporting.py:98,107`

`check("the resolved path is executable", ...)` and `check("with PATH blind, the fallbacks still find it", dor.claude_binary(), _binary)` compare `shutil.which("claude")` against the four hardcoded `CLAUDE_FALLBACKS`. Any install outside that list (nvm, volta, asdf — all normal for a node-installed CLI) fails both. The test is registered in `capability-manifest.json`, so `capability-gate.py` goes red with it.

I ran the PR's own suite twice against the PR's own code. Only `shutil.which`'s answer differed:

```
=== A) claude where THIS machine has it (/Users/assafkipnis/.local/bin/claude) ===
all linear-dor failure-reporting checks passed
exit: 0

=== B) same code, claude installed via nvm (~/.nvm/versions/node/v22.11.0/bin/claude) ===
FAIL:
  - the resolved path is executable: got False, want True
  - with PATH blind, the fallbacks still find it: got '/Users/assafkipnis/.local/bin/claude',
    want '/Users/assafkipnis/.nvm/versions/node/v22.11.0/bin/claude'
exit: 1
```

Same root cause both ways: a hardcoded location list. On that machine the PATH fix also would not hold, since launchd's `/usr/bin:/bin:/usr/sbin:/sbin` plus those four paths still misses it. I constructed the nvm path rather than finding such a machine in the fleet, so it is a minor.

---

## MINOR — four distinct root causes reach Linear as one identical line

`linear-dor-drafter.py:299`

`draft_one` writes the specific reason to stderr and returns `None`; `main()` appends a fixed string. The Linear issue tells the operator a count, not a cause.

```
cause=no claude binary anywhere                -> - ASK-147: draft failed (claude call or output rejected)
cause=TimeoutExpired after 300s                -> - ASK-147: draft failed (claude call or output rejected)
cause=rc=1 and 0 chars of output               -> - ASK-147: draft failed (claude call or output rejected)
cause=model returned prose, no Outcome bullet  -> - ASK-147: draft failed (claude call or output rejected)

distinct lines Linear receives for 4 distinct root causes: 1
```

The `FileNotFoundError` that this PR exists because of would not be distinguishable in Linear from a timeout. The actionable detail stays in `linear-dor.err`.

---

## NIT — the `"dry"` verdict is unreachable from `main()`

`linear-dor-drafter.py:217`, tested at `test-...py:152-153`

`main()` returns at the `--apply` guard before `report_failures` is ever called.

```
$ python3 - <<'PY'   # instrument report_failures, run main() with no --apply
dry run. 1 remaining; --apply to write 1.
rc: 0  report_failures called with apply=False: []
the "dry" return in report_failures is reachable from main(): False
```

Two checks guarding a branch production never takes.

---

## What is sound

- **Invented-fixture shape.** This was my primary suspicion and it held. `linear-sync.py` is a module, not a class; `fetch_remote_state` really returns `(team_id, project|None, {key: {linear_id, identifier}})` (`:420-471`), `ISSUE_CREATE`/`COMMENT_CREATE` really are module-level query strings (`:413`, `:474`), `read_ledger`/`append_ledger` really have those signatures (`:85`, `:106`). No key that no producer emits.
- **"Never raises."** Attacked with an unreachable `fetch_remote_state`, an unreachable `graphql`, and an `issueCreate` returning no id. Every path outside the `try` is a return. The claim is true.
- **Wiring is not decoration.** The `main()` test drives the real `main()` with every draft failing and asserts the create landed. Deleting line 312 makes it red. The `fetch_draftable` extraction is what makes that stubbable, and it is behavior-preserving.
- **Test isolation.** `dor.STATE` is redirected to a tempdir and `_linear` is stubbed, so neither the live state file nor `q-system/output/linear-ledger.jsonl` is touched.
- **The fix works on this host.** With `shutil.which` blinded, `claude_binary()` still resolves `/Users/assafkipnis/.local/bin/claude`.
- **Duplicate permanent issues — tried, could not reproduce.** The OR-guard (`remote_keys` else ledger) holds; the ledger is present with 208 records. Dropped.
- **Concurrency — tried, could not reproduce.** launchd will not double-fire the label, and a duplicate needed a hand-constructed interleave. Dropped.

---

## VERDICT: REQUEST CHANGES

Two majors, both recoverable, so this is not a BLOCK — nothing here is permanent if it fires.

**Fix first:** the closed-issue loop at `linear-dor-drafter.py:224`. Fetch state alongside the dedup key and either reopen the issue or file a fresh one when the existing one is `completed`/`canceled`. As written, the detector this PR adds turns itself off the first time the operator triages it — which is the one failure mode the whole change exists to prevent.

FINDINGS:
major|report_failures comments on the deduped issue without checking state, so once the operator closes it every later failing night is a comment on a Done issue with no open board signal|q-system/.q-system/scripts/linear-dor-drafter.py:224
major|When Linear is unreachable at report time the failures are dropped to stderr, the run exits 0, and the only record is linear-dor-state.json which no code in the repo reads|q-system/.q-system/scripts/linear-dor-drafter.py:259
minor|The test asserts the host's claude install location, so capability-gate goes red on any machine whose claude is outside the four hardcoded CLAUDE_FALLBACKS|q-system/.q-system/scripts/test/test-linear-dor-failure-reporting.py:107
minor|All four distinct draft-failure causes collapse into one identical Linear line, leaving the actionable reason only in linear-dor.err|q-system/.q-system/scripts/linear-dor-drafter.py:299
nit|The "dry" verdict in report_failures is unreachable from main(), and two tests guard that dead branch|q-system/.q-system/scripts/linear-dor-drafter.py:217
END FINDINGS

assafkip:
## Review round 2 — every finding answered

All five fixed, none disputed. Commit `b6b8e2b`. RED observed before green on every one.

```
AttributeError: module 'dor' has no attribute 'CLAUDE_BIN_ENV'   # RED, new test vs old code
all linear-dor failure-reporting checks passed                    # 47 checks
```

---

### MAJOR 1 — the detector switched itself off once the operator triaged it

`report_failures` now fetches the permanent issue's LIVE state before it writes.

- `ISSUE_STATE_Q` asks for the state `fetch_remote_state` structurally cannot return.
- `completed` / `canceled` → `_reopen()` moves it to the team's first `unstarted` state (Todo, falling back to Backlog), THEN comments. Verdict `reopened`.
- Reopen refused → verdict `reopen-failed`, the detail still lands on the issue, and the founder gets the one Slack ping that says the board has nothing open. Not silently reported as success.
- The `live is None` case (ledger points at an issue that no longer exists) files a fresh one instead of commenting into a 404. You did not raise this one; it fell out of fetching state.

Tests that go red without the fix: `a completed issue is reopened, not commented into the void`, the same for `canceled`, `...to an open state, not any state`, `a reopen Linear refuses is reported as such, not as success`, `a vanished issue is replaced, not commented on`.

**Fixture-shape check, because you called it your primary suspicion.** Three new operations, all exercised against live Linear, not just the fake:

```
ISSUE_STATE_Q  -> ASK-183 {'name': 'In Progress', 'type': 'started'}
TEAM_STATES_Q  -> 7 states; reopen target would be Todo unstarted
REOPEN_M       -> {'issueUpdate': {'success': True}}
state after    -> {'name': 'In Progress', 'type': 'started'} (unchanged: True )
```

The mutation ran as a no-op (moved ASK-183 to the state it was already in) so the shape is proven without touching the board.

---

### MAJOR 2 — the reporter's own failure was reported nowhere

Two changes, because a ping alone still loses the data.

1. **Slack**, through `slack-notify.sh` — the single channel `founder-notifications.md` names. Fires on `unreachable` and on `reopen-failed`, nothing else.
2. **The failures survive the night.** `pending_failures` in `linear-dor-state.json`, capped at 50, prepended to the next run's list and cleared when they land. `read_pending()` is what gives that file the reader you correctly said it never had. A clean night now drains the backlog: `report_failures` is called with `to_report = pending + failed`, so zero new failures still flushes.

```
dor-drafter: drafted 0, 1 failed (unreachable), 1 STILL UNFILED, 1 still lack a DoR
dor-drafter: drafted 0, 0 failed (created), 1 carried in, 0 still lack a DoR
```

The carried entries are labeled in the Linear body (`(1 carried from an earlier run that could not reach Linear)`) so the report does not claim they happened tonight.

**On the ping test:** `dor.NOTIFY_SCRIPT` is redirected to a script in a tempdir that records its argument. A stub on `dor.notify` would prove the function exists, not that the ping leaves the process.

**Where I deliberately did NOT add a ping**, since silence bought by a fix is the expensive kind:

- **A normal failing night** (`created` / `commented` / `reopened`) stays Slack-silent. The open Linear issue is the signal; a nightly line on top of it trains the founder to ignore the channel. Different from before: before, the board got nothing.
- **Linear unreachable at the TOP of `main()`** (`:274`, team lookup) is still a bare stderr line and exit 0. Deliberate: at that point nothing has been attempted and no data exists to lose, the drip just skips a night. Pinging there would be one line per night for the whole duration of a Linear outage. The report-time path is different because eight real diagnoses are in hand and about to be dropped.

---

### MINOR 3 — the test asserted where `claude` lives on the host

You are right and I could not defend it. The two host-dependent checks are gone. The suite now plants its own executable in a tempdir, blinds `shutil.which`, and asserts against the planted path. Nothing in the file reads the host's install location, so your nvm machine passes.

The underlying point (a hardcoded list cannot be exhaustive) gets `KIPI_CLAUDE_BIN`, checked before PATH, with a non-executable value ignored and logged rather than returned. Tested three ways. The list is no longer the last word, and on a machine where resolution genuinely fails, minor 4 means the operator now reads the cause instead of a count.

---

### MINOR 4 — four causes, one line

`draft_one` returns `(body, reason)`; `main()` files the reason. What Linear receives now:

```
- ASK-1: no claude binary ($KIPI_CLAUDE_BIN unset or bad, not on PATH, not in 4 known install locations)
- ASK-1: claude call failed (TimeoutExpired after 300s)
- ASK-1: claude exited rc=1 with 0 chars of output
- ASK-1: draft had no Energy/Time line (149 chars), not written
```

Asserted as `three distinct causes produce three distinct lines` plus the missing-binary case.

---

### NIT — the dead `dry` branch

The `apply` parameter is gone from `report_failures` and so are the two tests that guarded it. `main()` returns at the `--apply` guard, which is the real gate.

---

### Layer above, walked

- **Who reads what I started writing:** the state file's consumer is the drafter itself. Corrupt or missing → `read_pending()` returns `[]`, never raises. Capped at 50 so a long outage cannot grow it without bound.
- **Does the report still tell the truth:** the summary line now separates this run's failures from carried ones and says `STILL UNFILED` when the queue did not drain. The Linear body labels carried entries.
- **Second code path doing the same job:** yes, and I am not fixing it here. `fleet-health-daily.py:437` has the identical closed-issue blindness. It SKIPS an existing key, so it never writes onto a Done issue, but its detector also goes dark the moment the operator closes one. Out of ASK-183's scope (its DoR is `com.kipi.linear-dor`), captured as **`sp-1c47615e`** so it cannot be forgotten. Fix is the same `_live_issue` / `_reopen` shape.
- **What the operator sees at 3am:** on a normal failing night, an open Linear issue with one line per cause and no phone buzz. Only when the board could not be reached does the phone go off, and that ping says how many failures are held and that they will be retried.

### Verification

```
$ python3 q-system/.q-system/scripts/test/test-linear-dor-failure-reporting.py
all linear-dor failure-reporting checks passed          # 47 checks, exit 0

$ python3 q-system/.q-system/scripts/capability-gate.py --repo-root . --check-only
declared: 64 tests (0 quarantined) ... capability-gate: GREEN

$ python3 q-system/.q-system/scripts/linear-dor-drafter.py --limit 1 --apply
dor-drafter 2026-07-27T06:39:44Z: 121 issue(s) lack a Definition of Ready
  drafted ASK-143  CAP-35 rule verify-before-disagree: When the founder states
dor-drafter: drafted 1, 0 failed (none), 120 still lack a DoR
```

Things you said the code survived and I checked I did not regress: the reporter still never raises (every path out of the new `try` is a return; `_reopen` swallows its own errors by design), the `main()` wiring test still drives the real `main()`, and test isolation still holds — `dor.STATE` goes to a tempdir, `dor.NOTIFY_SCRIPT` goes to a tempdir, `_linear` is stubbed, so neither the live state file nor `linear-ledger.jsonl` nor the real Slack webhook is touched.

assafkip:
Repo left untouched (same 5 untracked files as at session start). Here is my read.

---

# Review: PR #12 — round 2

I read the diff and `linear-sync.py` (the real producer) before opening the earlier rounds. Four attacks, three reproducers that actually failed, one dropped. Everything below was run against the branch code at `653f2bd`.

---

## MAJOR — the `live is None` branch added in round 2 cannot fire; real Linear raises where the fixture returns `None`

`q-system/.q-system/scripts/linear-dor-drafter.py:340` (helper at `:288`)

`_live_issue()` is written as if `issue(id:)` returns `{"issue": None}` for an id Linear does not have. It does not. `linear-sync.py:381` raises on any `errors` array, and Linear puts a missing issue in `errors`. Verified against the live API:

```
$ python3 -c '... ls.graphql("query($id:String!){issue(id:$id){...}}", {"id": "00000000-0000-4000-8000-000000000000"})'
REAL API RAISED: LinearAPIError [{"message": "Entity not found: Issue", "path": ["issue"],
"extensions": {"type": "invalid input", "code": "INPUT_ERROR", "statusCode": 400, "userError": true, ...
```

So `_live_issue` raises inside `report_failures`'s `try`, the `except` at `:384` swallows it, and the run returns `"unreachable"`. I took the PR's own fixture and changed exactly one thing — `issue(id:)` raises like the real `graphql` does — then re-ran the PR's own assertion:

```
$ python3 -c '<PR fixture, issue(id:) raises LinearAPIError like linear-sync.graphql>'
  linear unreachable, 2 failure(s) NOT filed: Entity not found: Issue

verdict          : unreachable
issues created   : 0
comments posted  : 0
founder pings    : ['linear-dor: Linear unreachable, 2 failure(s) could not be filed
                    (Entity not found: Issue). Held for the next run.']

PR test line 297 asserts verdict == created and len(created) == 1
PR assertion holds under the real API contract: False
```

**Falsified citation.** The round-2 answer says *"Fixture-shape check, because you called it your primary suspicion... ISSUE_STATE_Q -> ASK-183 {'name': 'In Progress', 'type': 'started'}"*. That proves the shape for an issue that **exists**. The branch under test is the one where it does not, and that case was never sent to the live API. The check `a vanished issue is replaced, not commented on` is green on a shape no producer in this repo emits.

**3am consequence.** The trigger is the operator deleting the bot-filed issue — a normal reaction to an auto-created "nightly DoR drafter reported failures" object. The ledger (`q-system/output/linear-ledger.jsonl`, gitignored, machine-local, 208 records) keeps the dead id forever. From then on: nothing filed, ever; a nightly Slack ping saying **"Linear unreachable"** while Linear is fine; failures accumulating in `pending_failures`. The detector switches itself off and lies about why. A human recovers it by hand-editing the ledger, which is why this is major and not a block.

Not an escalation: this is code added in round 2, and the round-1 review says explicitly *"You did not raise this one."*

---

## MINOR — round-1 MAJOR 2 is fixed at report time but still live at start time, which is the branch that fires when Linear is down

`q-system/.q-system/scripts/linear-dor-drafter.py:406`

The round-2 fix pings and holds failures when `fetch_remote_state` fails inside `report_failures`. But when Linear is down at 03:00, `main()` never gets there — the team lookup at `:401` raises and returns at `:406`, before `read_pending()`, before `report_failures`, before the state write.

```
$ python3 -c '<preload state with 1 held failure, stub _linear() to raise, run main()>'
state BEFORE the run: {'ran_at': '2026-07-26T03:00:00Z', ..., 'pending_failures':
  ['ASK-147: claude call failed (TimeoutExpired after 300s)']}
linear unreachable: network: [Errno 60] Operation timed out

exit code launchd sees      : 0
founder pings fired         : []
held failures still unfiled : 1
ran_at after the run        : 2026-07-26T03:00:00Z  (unchanged)

CONTRACT: a night Linear is down leaves SOME signal for the founder -> False
```

I am rating this **below** round 1's major on the same underlying issue, deliberately: on a start-time-down night no drafts run, so there are no new failures to lose and the held ones are preserved. What is lost is liveness — the backlog silently stops draining, `launchd-health-check.py` sees exit 0, and `ran_at` does not advance so the state file cannot even be used as a freshness deadman.

---

## NIT — the failure issue the job files becomes a draft target for the job

`q-system/.q-system/scripts/linear-dor-drafter.py:357`

The created payload has no `## Definition of Ready` and no explicit state, so it lands in `backlog` and passes the job's own selector:

```
$ python3 -c '<call report_failures, feed the created description back into needs_dor>'
  filed ASK-1000 with 1 failure(s)
  needs_dor(the failure issue) -> True
```

Tomorrow night it consumes one of the plist's `--limit 8` `claude -p` slots to write LLM prose onto the operator's own failure record. One-time per issue, harmless, but it is a slot.

---

## Dropped

- **`PENDING_CAP` silent truncation** (`:80`). At `--limit 8`, seven fully-failing nights where Linear is reachable at start but not at report would silently discard the oldest lines with no field recording the drop. I could not construct that interleave without hand-forcing both conditions, and the start-time branch above means Linear-down nights produce nothing to accumulate. Dropped rather than reported thin.

## What is sound

- **Fixture shape, everywhere except the one branch above.** `fetch_remote_state` really returns `(team_id, project|None, {key: {"linear_id", "identifier"}})` (`linear-sync.py:463`); `read_ledger` really returns `{key: record}` (`:85`); `ISSUE_CREATE`/`COMMENT_CREATE` really are module-level strings (`:413`, `:474`). The fake's query dispatch is correct — `issueUpdate(id:` does not collide with the `issue(id:` match, and `issues(` does not swallow `ISSUE_STATE_Q`. I looked for a key no producer emits and found exactly one.
- **The Slack channel is real, not a stubbed claim.** `~/.config/kipi/slack-webhook` exists (81 bytes), so `notify()` reaches the founder in production, not just in the test's fake script. The test routing pings through a real `subprocess.run` rather than stubbing `dor.notify` is the right call and it caught nothing because there was nothing to catch.
- **Test isolation.** I ran the suite (47 checks, green). `q-system/output/linear-ledger.jsonl` still stamped `Jul 26 21:14` afterward — the live ledger and live state file were not touched. Round 1's minor 3 is genuinely fixed: the test now plants its own binary and no check reads the host's `claude` location.
- **One cause per line** is real: my run shows `TimeoutExpired`, `rc=1`, and the missing-Energy rejection arriving as three distinct strings. Round 1's minor 4 settled.
- **The closed-issue reopen (round 1's major 1)** holds against the fixture and I could not falsify the live-API proof for `TEAM_STATES_Q`/`REOPEN_M` — those were exercised against real Linear as a no-op state move, which is a genuinely good way to prove a mutation's shape without touching the board.

**On review calibration:** round 1 was well-calibrated. Both of its majors were real, both got fixed, and it correctly dropped the duplicate-issue and concurrency angles rather than padding. My major here is on code round 2 *added*, not on unchanged code — this is not the "new blockers on unchanged code" pattern.

## VERDICT: REQUEST CHANGES

One major, recoverable, so not a BLOCK.

**Fix first:** `linear-dor-drafter.py:288`. Wrap the `_live_issue` call so that a "not found" / `INPUT_ERROR` from Linear returns `None` instead of propagating, and separate that from a genuine transport failure — right now the reporter cannot tell "this issue is gone, file a new one" from "Linear is down, hold everything," and it guesses wrong in the direction that turns the detector off. The test needs a fixture that raises, not one that returns `None`.

FINDINGS:
major|The live-is-None branch cannot fire because real Linear raises Entity not found rather than returning null, so a deleted permanent issue means nothing is ever filed plus a nightly false Linear-unreachable ping|q-system/.q-system/scripts/linear-dor-drafter.py:340
minor|When Linear is down at the start of the run main() returns before report_failures, so held failures are not flushed, no founder ping fires, and ran_at does not advance|q-system/.q-system/scripts/linear-dor-drafter.py:406
nit|The failure issue the job files has no DoR and lands in backlog, so it passes the job's own selector and consumes a nightly claude -p slot|q-system/.q-system/scripts/linear-dor-drafter.py:357
END FINDINGS

assafkip:
## Review round 3 — all three answered

All three fixed, none disputed. Commit `c1a7d6a`. RED observed on each before green.

```
- a vanished issue is replaced, not commented on: got 'unreachable', want 'created'
- the job's own failure record is not a draft target: got True, want False
- ran_at advances, so the state file is a usable freshness deadman: got False, want True
```

---

### MAJOR — the `live is None` branch could not fire

Confirmed against live Linear before touching anything, and confirmed harder than you did: **both** a well-formed uuid with nothing behind it **and** a non-uuid string produce the identical payload.

```
well-formed uuid, no such issue -> RAISED LinearAPIError: [{"message": "Entity not found: Issue",
  "path": ["issue"], "extensions": {"type": "invalid input", "code": "INPUT_ERROR",
  "statusCode": 400, "userError": true, "userPresentableMessage": "Could not find referenced Issue."}}]
garbage, not a uuid             -> RAISED LinearAPIError: [ ...identical... ]
an identifier that exists       -> RETURNED {'issue': {'id': 'bd11c1fa-...', 'identifier': 'ASK-183',
                                              'state': {'name': 'In Progress', 'type': 'started'}}}
```

`_live_issue` now catches and separates the two. **The discrimination is narrow on purpose, and the direction matters:** I match the message `entity not found`, NOT `INPUT_ERROR` alone. A malformed query is also `INPUT_ERROR`, and reading that as "gone" would file a **new permanent issue every single night** — and Linear issues cannot be deleted here (`destructive-op-deny.sh` blocks `*delete*` and archive). So an over-match is unrecoverable while an under-match only holds the failures for the next run. That asymmetry has its own test:

```
ok: Linear down at the state lookup is held, not read as a deletion
ok: ...no second permanent issue is forked
ok: a vanished issue is replaced, not commented on
ok: ...nothing was written at the dead id
ok: ...the ledger is repointed so tomorrow does not chase the dead id again
ok: ...and the founder is NOT told Linear is down, because it is not
```

**On the falsified citation.** You were right and it was the fair hit: round 2 proved the shape for an issue that exists and shipped a branch for the case where it does not. The fixture no longer models an invented shape — it raises the real `ls_real.LinearAPIError` with the payload captured above, so the test now fails if Linear's contract is what round 2 assumed.

I also checked the ledger repoint actually holds one layer out, since a create that is not readable next night would loop: `read_ledger` is last-wins by construction (`linear-sync.py:106` — "Re-appending a known key is harmless"), so the fresh id supersedes the dead one. That is why the ledger-append check is in the list above rather than just the create.

---

### MINOR — down at start, not at report time

You were right that round 2 fixed the wrong half, and right to rate it below round 1's major. `main()` now leaves a heartbeat instead of returning bare: `ran_at` advances, `pending_failures` is written back, `failures_reported` is `unreachable-at-start`, and the printed line says `SKIPPED` rather than implying a run happened.

**On the ping cadence, which is the real design question here.** Three options, none free:

- Round 2's answer (never ping) — your finding: a down night leaves no signal at all.
- The fleet's existing pattern (`launchd-health-check.py:37`, `FAIL_PING_TTL_SECONDS = 6h`) — for a job that runs **once a night**, a 6h TTL degenerates to a ping every single night of the outage. That is the noise round 2 was right to avoid.
- Pure edge-trigger (ping only on the first down night) — bounded noise, but **unbounded silence**: miss the one ping and a three-week outage is three weeks of nothing.

I took night 1, then every 7th. Once a week says "still dark" without training the channel to be swiped away, and the ping text states its own cadence so absence is interpretable:

```
linear-dor: Linear unreachable at start for 7 night(s), the DoR drip is not running.
1 earlier failure(s) still unfiled. Next ping in 7 nights if it stays down.
```

```
ok: ...and the founder is told the drip is not running
ok: a second night of the same outage does not ping again
ok: ...but it still leaves a heartbeat
ok: seven nights down is counted, not just flagged
ok: ...and the weekly re-ping fires exactly once more
ok: ...night 8 is silent again
ok: a new outage after a healthy night pings again
ok: the failure held across the outage is filed once Linear returns
```

**The silence I am buying, stated plainly:** nights 2-6 and 8-13 of an outage are silent. Justification: the ping that fires names the cadence, the state file carries a level-triggered record for anything that polls it, and no data is at risk on those nights (nothing is attempted, so nothing new can be lost). What is lost is only elapsed time on a backlog that is already stalled.

**And the honest limit on my own fix** — the same criticism you levelled in round 1, applied to what I just wrote: `ran_at` advancing makes the state file *capable* of being a freshness deadman, but **nothing in the fleet reads it**. `launchd-health-check.py` keys only on `LastExitStatus` and load state, so a job that exits 0 forever while doing nothing is still invisible. Building that reader is a fleet-wide job with its own design, and ASK-183's DoR says "Not doing: rewriting what the job does." Captured as **`sp-b6516bab`** rather than claimed as wired.

---

### NIT — the failure issue was a draft target for the job

Fixed at the source of the drift: `FAILURE_MARKER` is now ONE constant used by both the writer (the create payload) and the selector (`needs_dor`), so they cannot disagree.

Matched narrowly on **this job's** key rather than on any `kipi-key` marker, deliberately: `fleet-health-daily.py`'s machine-filed issues are real work items and a DoR helps them. Your exact repro is the test, plus a negative self-test so the skip cannot silently become blanket:

```
ok: the job's own failure record is not a draft target
ok: ...but an ordinary backlog issue still is
```

---

### On your dropped finding

`PENDING_CAP` silent truncation: you dropped it rather than report it thin, which was the right call on evidence. It is still a silent cap in a repo whose own rule says a bounded path must log what it discards, and my start-time fix does not touch it. Not bundled into this pass (scope), captured as **`sp-f5bc9f5e`**.

### Layer above, walked

- **Who reads what I changed:** `read_state()` / `read_pending()` are the only readers and never raise on a corrupt or missing file. `remaining` is written as `null` on a skipped night rather than carrying the previous run's count forward, because a stale number reads as a fresh measurement.
- **Second code path:** `linear-sync.py:502` (`cmd_progress`) has the same dead `if not issue` branch for the same reason — but its behaviour is already correct, because it catches `LinearAPIError` and returns `BLOCK`. Dead defensive code, not a defect, so no spillover item. `fleet-health-daily.py` never calls `issue(id:)` at all, so this class does not reach it; its separate closed-issue blindness stays at `sp-1c47615e`.
- **Timing change:** `read_pending()` moved from after the drafting loop to the top of the run. Same value (nothing writes the file in between) and strictly safer under a hypothetical concurrent run, since reading later would widen a clobber window.
- **What the operator sees at 3am:** unchanged on a normal failing night (open Linear issue, no buzz). A deleted permanent issue now produces a fresh Linear issue and **no** ping, where before it produced nothing and a false "Linear unreachable" buzz every night.

### Things you said the code survived, re-checked as not regressed

`report_failures` still never raises — the new `_live_issue` re-raise lands inside its existing `try`, and every path out is a return. The `main()` wiring test still drives the real `main()`. Test isolation still holds: `dor.STATE` and `dor.NOTIFY_SCRIPT` both point into a tempdir, `_linear` is stubbed, and `ls_real` is loaded only for its exception class (no module-level side effects, no network) — the live state file, `linear-ledger.jsonl`, and the real webhook are all untouched.

### Verification

```
$ python3 q-system/.q-system/scripts/test/test-linear-dor-failure-reporting.py
all linear-dor failure-reporting checks passed        # 73 checks, exit 0

$ python3 q-system/.q-system/scripts/capability-gate.py --repo-root . --check-only
capability-gate: GREEN

$ python3 q-system/.q-system/scripts/linear-dor-drafter.py --limit 1 --apply
dor-drafter 2026-07-27T07:03:16Z: 120 issue(s) lack a Definition of Ready
  drafted ASK-142  CAP-31 rule stop-hook-routing: When the Codex stop-hook (or
dor-drafter: drafted 1, 0 failed (none), 119 still lack a DoR

$ python3 q-system/.q-system/scripts/launchd-health-check.py --dry-run | grep -i linear-dor
(no problem reported for com.kipi.linear-dor)
```

assafkip:
# Review round 3 — PR #12, ASK-183 linear-dor migration

Read the diff and `linear-sync.py` first, staged the branch at `c1a7d6a` into a tmp sandbox, ran the suite, mutation-tested it, then read rounds 1 and 2. Three findings survived reproduction. All minor.

---

## MINOR 1 — an error one step after the team lookup escapes every guard the PR added

`q-system/.q-system/scripts/linear-dor-drafter.py:508`

Round 2 raised this class at the team lookup and rated it minor; the fix covered the team lookup only. `fetch_draftable(ls, tid)` on the very next line has no guard. `write_state`'s own docstring (`:262`) says "Every exit path from main() goes through it, including the ones that did no work, so `ran_at` is a real heartbeat." Two exit paths do not.

```
$ python3 $TMPDIR/rev12/repro.py
=== A) Linear errors on the ISSUE FETCH, one step after the team lookup ===
  main() RAISED RuntimeError: network: [Errno 60] Operation timed out
  -> traceback out of sys.exit(main()); launchd LastExitStatus = 1, not the designed 0
  founder pings fired     : 0
  ran_at frozen at        : 2026-07-26T03:00:00Z
  held failures unflushed : ['ASK-147: claude call failed (TimeoutExpired after 300s)']
```

The second falsification of the same docstring, benign but it is the same claim:

```
dor-drafter 2026-07-27T07:14:29Z: 0 issue(s) lack a Definition of Ready
dry run. 0 remaining; --apply to write 0.
dry (no --apply) rc: 0 | ran_at after: 2026-07-26T03:00:00Z
```

**Why this stays minor, not major.** Same underlying issue round 2 already rated minor, so no escalation. And it fails safe in the alerting direction: I confirmed both jobs are loaded, and `launchd-health-check.py` pings on a non-zero `LastExitStatus`, so a crash here surfaces on the founder's phone within hours.

```
$ launchctl list | grep -E "launchd-health|linear-dor"
-	0	com.kipi.launchd-health
-	0	com.kipi.linear-dor
```

3am action: none required, the watchdog covers it. The cost is `ran_at` freezing so the state file stops being the freshness deadman the code says it is.

---

## MINOR 2 — the comment path never checks `success`, and main() clears the queue on its word

`linear-dor-drafter.py:430`

The create path checks the returned id and pings when Linear refuses (`:445`). The comment path checks nothing. `main():552` clears `pending_failures` on any verdict except `unreachable`.

```
=== B) commentCreate returns success:false with no errors array ===
  reported 2 failure(s) on ASK-999 (commented)
  verdict report_failures returned : commented
  main() would hold the failures   : False
  -> unfiled = [] so pending_failures is cleared; the 2 failures are gone
```

**Dropped a level on purpose.** I constructed that payload. I could not show live Linear ever returns `commentCreate.success = false` without an `errors` array, and `graphql()` (`linear-sync.py:381`) raises on any errors array. So the asymmetry is real; the trigger is unproven.

---

## MINOR 3 — the held-failure cap drops the oldest silently

`linear-dor-drafter.py:258` and `:550`

```
=== C) PENDING_CAP silent truncation ===
  60 held failures written, read_pending() returned: 50
  oldest kept: ASK-10: claude call failed
  -> 10 dropped, zero log lines mentioning the drop
```

`.claude/rules/no-orphan-findings.md` and the repo's no-silent-caps norm both want the drop logged. Needs ~7 straight nights of report-time outage at a full batch of 8 to reach, so it is narrow, and the run summary already prints `STILL UNFILED` with a count. Bounded.

---

## What is sound

Attacks I ran that this code survived:

- **The invented-fixture attack, which is what I came for.** `FakeLinear` matches the real producer on every surface it stands in for: `fetch_remote_state` really returns `(team_id, project|None, {key: {linear_id, identifier}})` (`linear-sync.py:420-471`), `ISSUE_CREATE`/`COMMENT_CREATE` really are module-level strings (`:413`, `:474`), `read_ledger` really returns `{key: record}` (`:85`), and `slack-notify.sh` really takes the message as `$1` (`:14`), which is what the fake ping script reads. The `NOT_FOUND_PAYLOAD` fixture is a captured live payload, not an invention. No key that no producer emits.
- **Tests that could not fail.** I mutated the sandbox copy three ways and re-ran the suite:
  ```
  MUTANT: reopen disabled                    -> exit 1
  MUTANT: deleted-issue detection disabled   -> exit 1 | 4 checks red
       - a vanished issue is replaced, not commented on: got 'unreachable', want 'created'
  MUTANT: held backlog never flushed         -> exit 1
  RESTORED -> exit 0 | all linear-dor failure-reporting checks passed
  ```
  Not decoration.
- **The gate wiring.** The test is registered in `expected_tests` (64 entries), the list `capability-gate.py` actually reads, with `runner: python3`. Not parked in `declared_inert` or `skeleton_only`.
- **Dedup durability.** The `kipi-system` project exists on the ASK board and the failure issue is created with its `projectId`, so `fetch_remote_state` is the authority and the gitignored machine-local ledger is only the fallback. A wiped ledger does not fork a permanent issue.
- **Archived-issue path — tried, could not reproduce, dropping it.** My best remaining theory was that Linear auto-archiving a closed failure issue would make `_reopen` return success while the comment lands somewhere invisible, with verdict `reopened` and no ping. Confirming it needs archiving a real Linear issue, which is a board mutation I will not make in a read-only review, and `ISSUE_STATE_Q` does not select `archivedAt` so I cannot detect it from the outside either. No repro, so no finding. Flagging it only as the one stone I left unturned.
- **Concurrency — tried, could not reproduce.** launchd will not double-fire the label; a lost `pending_failures` write needs a hand-built interleave. Round 1 dropped it too; not re-raising.

Rounds 1 and 2 were calibrated, not inflated. Both majors were genuine unattended-silence bugs, both correctly rated recoverable rather than blocking, and my repros show both are actually dead now, not papered over. Round 3 producing only minors is the convergence signal.

The one thing worth naming: every round's major has lived inside the previous round's fix. R1's reopen fix introduced the `live is None` branch R2 killed; my minor 1 lives in the guard R2 asked for. The file is now 560 lines with six verdict states for a job that drafts eight issues a night, and all three of my findings sit in that guard-on-guard layer rather than in what the job does. The remaining signal points at stopping, not at another round.

## VERDICT: APPROVE WITH NITS

Nothing here is permanent or unrecoverable, nothing files duplicate Linear objects, nothing cries wolf on a normal night. Ship it.

**Fix first:** wrap `fetch_draftable` at `:508` in the same guard the team lookup got, so an error there leaves a heartbeat instead of a traceback, or correct the `write_state` docstring to say what the code does. One or the other, not both.

FINDINGS:
minor|An exception out of fetch_draftable escapes every guard so main() raises, ran_at freezes, no founder ping fires, and the module's stated exit-0 contract breaks|q-system/.q-system/scripts/linear-dor-drafter.py:508
minor|The comment path does not check commentCreate success while the create path checks the returned id, so a refused comment reports verdict commented and main() clears the held-failure queue|q-system/.q-system/scripts/linear-dor-drafter.py:430
minor|PENDING_CAP drops the oldest held failures with no log line, against the repo's own no-silent-caps norm|q-system/.q-system/scripts/linear-dor-drafter.py:258
END FINDINGS

PR #13, opened 2026-07-27
fix(heartbeat): a failed sweep must exit non-zero so it reaches Linear (ASK-184)

assafkip:
## ASK-184 — Migrate `com.kipi.openloops-heartbeat` to Linear-tracked execution

Linear: https://linear.app/ask-consulting/issue/ASK-184/migrate-comkipiopenloops-heartbeat-to-linear-tracked-execution

A migration of **how the job is scheduled, reported and verified** — not a rewrite of what
it does. Per the DoR's `Not doing` line, the sweep logic is untouched.

### The four bars

| Bar | Status | Evidence |
|-----|--------|----------|
| 1. launchd, never cron | **already passed** | `launchctl list com.kipi.openloops-heartbeat` → loaded. `crontab -l \| grep open-loops` → no match. launchd is the only scheduler, so nothing shells `claude` without a keychain (ASK-150). |
| 2. Failures reach Linear, not only a log | **was broken — fixed here** | see below |
| 3. Live or in the pause ledger, never dark | **already passed** | `job_status()` → `('ok', 0)`; not in the pause ledger; `fleet-health-daily.py` → `launchd-dark: 0, launchd-failing: 0` |
| 4. Verified by running it | **live run** | posted as a comment on this PR and the issue |

### Bar 2 was the real defect

`open-loops-heartbeat.sh` ended in an unconditional `exit 0` with no `set -e`. launchd
recorded `LastExitStatus=0` even when an instance's headless agent died or the post-sweep
step-audit failed.

That made `fleet-health-daily.py`'s `launchd-failing` detector **structurally blind** to
this job: its failures reached Slack and a freeform log, and never Linear. The exit code IS
the wire to Linear — non-zero → `LastExitStatus` non-zero → a deduped Linear issue on the
next fleet-health run.

Fix: count sweep failures and `exit 1` if the count is non-zero. Counted rather than
`set -e`, so one bad instance still lets the rest of the fleet sweep finish.

A `skipped` step (missing instance path, pre-propagation instance) is registry drift, not a
job failure, and is deliberately **not** counted — paging weekly on known drift is the
alert-fatigue that teaches the founder to ignore the channel.

### Reproducer, observed RED then GREEN

`q-system/.q-system/scripts/test/test-open-loops-heartbeat-exit.sh` runs the real script
against a temp `SKEL`, a fake instance with one `[needs you]` loop, a `claude` shim, and a
stubbed `slack-notify`. Hermetic — it never touches the live fleet.

```
BEFORE: FAIL failed agent run exits non-zero (exit 0, expected nonzero) / run-log status=failed
        PASS clean sweep exits zero
        passed=1 failed=1 | SUITE_EXIT=1

AFTER:  PASS failed agent run exits non-zero (exit 1, expected nonzero)
        PASS clean sweep exits zero
        passed=2 failed=0 | SUITE_EXIT=0
```

The second case pins the other direction: a clean sweep still exits 0, so the fix cannot
page on every run.

Declared in `capability-manifest.json` — the capability gate only runs tests the manifest
declares, so an undeclared reproducer is a silent absence:

```
python3 q-system/.q-system/scripts/capability-gate.py
declared: 64 tests, tests: ran=64 quarantined=0, capability-gate: GREEN
```

### Bar 3, proven positively

The DoR's check is `launchd-health-check.py --dry-run | grep com.kipi.openloops-heartbeat`.
That script's `discover_problems()` only emits `failing` / `paused` / `not_loaded` lines — a
healthy job emits nothing, so an empty grep is the **passing** case and a match would mean
failure. Rather than rely on that absence, visibility is proven directly:

```
plist on disk          : True /Users/assafkipnis/Library/LaunchAgents/com.kipi.openloops-heartbeat.plist
matched watch prefix   : ['com.kipi.']
in watchdog glob set   : True
in pause ledger        : False
job_status()           : ('ok', 0)
appears as a problem   : no -- classified ok (loaded, last exit clean)

python3 q-system/.q-system/scripts/fleet-health-daily.py   # what `kipi health` dispatches to
  launchd-dark: 0
  launchd-failing: 0
  schedule-duplicate: 0
```

### Blast radius

Machine-local for the scheduling half: LaunchAgents and crontab do not propagate via
`kipi update`. The script itself **does** sync — it lives at
`q-system/.q-system/scripts/open-loops-heartbeat.sh`, so the exit-contract change reaches
every instance on the next `kipi update`. The only behavioral delta there is the exit code;
the sweep does the same work it did before.

### Files

- `q-system/.q-system/scripts/open-loops-heartbeat.sh` — failure counter + non-zero exit
- `q-system/.q-system/scripts/test/test-open-loops-heartbeat-exit.sh` — new reproducer
- `q-system/.q-system/capability-manifest.json` — declares the test
- root `.ask184_fetch.py` / `.ask184_recon.sh` — removed; session scratch an unattended
  auto-commit hook swept into the repo root

Do not merge — closeout runs through `/issue-verify` and `/issue-closeout`.


assafkip:
## Bar 4 — live run, verified

Ran the fixed script by hand exactly as launchd invokes it (`KIPI_REPO=/Users/assafkipnis/projects/kipi-system`, real registry, real log).

```
bash q-system/output/ask184-live-run.sh
HEARTBEAT_EXIT=0
```

**It did real work, not a no-op sweep:**

```
heartbeat-run-last.json  ts 2026-07-27 00:35:44
steps logged : 25  {'completed': 24, 'skipped': 1}

agent-waking instances (all completed):
  kipi-system, travel-agent, fractional-cxo, interview-coach, negotiator,
  Alice, all_points_setup, Prodigy_Gold, thaena, norri ai   -- 'agent ran (1 loops)'

failed  : none
skipped : ['reddit-build-radar']   (pre-propagation -- registry drift, deliberately not counted)
```

```
2026-07-27 00:35:44 heartbeat: fleet sweep complete
2026-07-27 00:35:44 heartbeat AUDIT: [open-loops-heartbeat] OK: all 25 expected steps accounted for
```

This proves the second half of the exit contract **live**, which the hermetic reproducer
proves in isolation: a *failed* sweep exits non-zero so `fleet-health-daily.py`'s
`launchd-failing` detector can file it to Linear, and a *clean* sweep still exits 0 so the
fix cannot page on every run.

All four bars pass. Not merging — closeout runs through `/issue-verify` and `/issue-closeout`.

assafkip:
## Setup used for every repro

```bash
T=$(mktemp -d); mkdir -p "$T/scripts/test"
REF=52f609e63e2dcab568d000d2ea4ed799305bee6e
B=repos/assafkip/kipi-system/contents/q-system/.q-system/scripts
for f in open-loops-heartbeat.sh run-step-audit.py; do
  gh api "$B/$f?ref=$REF" -q .content | base64 -d > "$T/scripts/$f"; done
gh api "$B/test/test-open-loops-heartbeat-exit.sh?ref=$REF" -q .content | base64 -d \
  > "$T/scripts/test/test-open-loops-heartbeat-exit.sh"
echo "$T" > /tmp/pr13_root
```

Baseline, PR head, unmodified:

```
test-open-loops-heartbeat-exit
PASS  failed agent run exits non-zero (exit 1, expected nonzero)
PASS  clean sweep exits zero (exit 0, expected zero)
passed=2 failed=0   SUITE_EXIT=0
```

Matches the PR body. Repo never modified; everything below runs against copies in `$TMPDIR`.

---

## Findings

### 1. minor — the new test stays green when you delete the line the PR presents as the fix

`q-system/.q-system/scripts/open-loops-heartbeat.sh:94`

`run-step-audit.py:47-50` already treats any `failed` step as a problem and exits 1, which trips the *second* counter at line 136. So the per-instance counter at line 94 is never the thing that decides the test's outcome.

```bash
T=$(cat /tmp/pr13_root); M=$(mktemp -d); mkdir -p "$M/scripts/test"
cp "$T/scripts/run-step-audit.py" "$M/scripts/"
cp "$T/scripts/test/test-open-loops-heartbeat-exit.sh" "$M/scripts/test/"
awk 'NR!=94' "$T/scripts/open-loops-heartbeat.sh" > "$M/scripts/open-loops-heartbeat.sh"
bash "$M/scripts/test/test-open-loops-heartbeat-exit.sh"
```

```
MUTANT: deleted line 94 (the per-instance failure counter)
32:SWEEP_FAILURES=0
135:  SWEEP_FAILURES=$((SWEEP_FAILURES + 1))
145:if [ "$SWEEP_FAILURES" -gt 0 ]; then
=== REPRO 2: PR test against the mutant ===
test-open-loops-heartbeat-exit
PASS  failed agent run exits non-zero (exit 1, expected nonzero)
PASS  clean sweep exits zero (exit 0, expected zero)
---
passed=2 failed=0
SUITE_EXIT=0
```

The RED-then-GREEN in the PR body is honest (pre-fix, both counters were absent). The gap is narrower: a future regression that drops line 94 ships green. Fix: a third case whose failure is an agent failure with a clean audit, or assert on the `heartbeat: N failure(s)` log line.

---

### 2. minor — an unreadable `instance-registry.json` skips the whole fleet, and the sweep still exits 0 with the audit reporting OK

`q-system/.q-system/scripts/open-loops-heartbeat.sh:106-118`

The comment on line 106-107 says the audit "Catches the case launchd-health cannot: the sweep exits 0 but never reached an instance." It does not catch it when the registry read itself is what failed. Both the sweep generator (line 97-102) and the expected-list builder (line 113-118) wrap `json.load` in `try/except: pass`, so they collapse together to `['kipi-system']` and match.

```bash
# CASE B in the harness: registry truncated to invalid JSON, 1 instance with open loops
bash "$(cat /tmp/pr13_root)/repro.sh"
```

```
== CASE B: instance-registry.json is truncated / invalid JSON ==
  expect 1 swept
    exit code            : 0
    agents actually woken: 0
    audit verdict        : heartbeat AUDIT: [open-loops-heartbeat] OK: all 1 expected steps accounted for
    last log line        : heartbeat AUDIT: [open-loops-heartbeat] OK: all 1 expected steps accounted for
```

24 registered instances, 0 swept, exit 0, launchd clean, nothing in Linear, log says OK. This is the exact shape of bar 2. It is pre-existing structure the diff does not touch, and I had to hand-corrupt the registry (the live one is well-formed, all 24 entries carry `name` + `path`), so per the calibration rule I dropped it from major to minor and am saying so. One-line fix: make the expected-list builder fail loudly instead of `except: pass`.

---

### 3. minor — `claude` missing from PATH exits 0 having done nothing, with no run-log and no audit at all

`q-system/.q-system/scripts/open-loops-heartbeat.sh:34`

Same "success for work that did not happen" class the PR fixes, 60 lines above the fix, untouched by the diff.

```
== CASE C: claude CLI not on PATH ==
    exit code            : 0
    agents actually woken: 0
    audit verdict        :            <- audit never ran
    last log line        : heartbeat: no claude CLI -> skip
```

Not currently reachable: the plist pins `PATH=/Users/assafkipnis/.local/bin:/opt/homebrew/bin:/usr/bin:/bin:/usr/sbin:/sbin`, and the live log shows real agent runs. No live producer, so minor.

---

### 4. minor — the Linear issue this PR newly enables points the operator at a file that does not exist

`q-system/.q-system/scripts/open-loops-heartbeat.sh:146` (alert text at `fleet-health-daily.py:154`)

`detect_failing_jobs` files an issue whose Action reads "Read its StandardErrorPath, fix the cause." Every diagnostic this script writes goes to `q-system/output/open-loops-heartbeat.log`; nothing goes to stderr, including the new `N failure(s) this sweep` line.

```bash
ls -l /tmp/kipi-openloops-heartbeat.err
```

```
ls: /tmp/kipi-openloops-heartbeat.err: No such file or directory
```

37 sweeps in the log, `StandardErrorPath` never created. At 3am the operator opens the issue, follows the Action, finds nothing. Fix at line 146: also write the failure summary to stderr (`>&2`), or point `StandardErrorPath` at the job log.

---

## What is sound

**The subshell trap.** I expected `SWEEP_FAILURES` in `work_instance` to be lost, since it is called from a `while read` loop. It is `done < <(...)`, process substitution, not a pipe, so the loop body runs in the current shell. Proved directly by deleting the *audit* counter (line 136) and re-running the suite: `PASS failed agent run exits non-zero (exit 1)`. The headline mechanism is real, not a no-op.

**The fixture-shape trap.** The fixture emits `{"instances":[{"name","path"}]}`. The live `instance-registry.json` has 24 entries, key union `has_git, instance_q_dir, name, note, path, skip_agent_check, subtree_prefix, type`, zero missing `name` or `path`, and no `status` key anywhere. The script reads only `name`, `path`, and the `status` filter. The fixture is faithful to the producer.

**"skipped is not counted" is true in code, not just in the comment.** Registry drift (path deleted from disk) sweeps the rest of the fleet and exits 0: `exit code: 0, agents woken: 1, audit: OK: all 3 expected steps accounted for`. No weekly wolf-crying on known drift.

**A registry entry missing `path` is NOT a blind spot.** I expected the KeyError to truncate both generators identically and hide the tail. It does not: the expected-list builder reads only `name`, so it keeps going. Result was `SILENTLY SKIPPED (2)` and `exit code: 1`. The audit caught it.

**No launchd respawn storm.** `plistlib` read of the plist: no `KeepAlive`, `RunAtLoad = False`, `StartCalendarInterval = {Hour 9, Minute 0, Weekday 1}`. A non-zero exit cannot turn into launchd relaunching a job that spawns headless agents in 24 instances.

**No Linear issue spam.** `finding_key` is `fleet-health/launchd-failing/<slug(label)>`, subject is the job label only, and `file_findings` refetches remote state as the dedup guard. One permanent issue per job, not one per failing run. (The flip side, that the key stays in `known` after the issue is closed so a later failure never files again, is fleet-health dedup semantics and predates this PR. I have no repro for it that does not write to Linear, and Linear objects are permanent, so I did not run it.)

**Concurrency.** Two sweeps would collide on the fixed `$RUNLOG_TMP`. Weekly schedule, no `KeepAlive`, no `RunAtLoad`, and the longest observed sweep is minutes. I could not construct a realistic overlap. Dropped.

**Wolf-crying against real history.** 37 sweeps in the log, 3 `agent run failed/timeout` lines, and the last 4 weekly runs all `OK: all N expected steps accounted for`. The 8 consecutive `SILENTLY SKIPPED` runs in early July were the real tail-skip bug, fixed 2026-07-05. Under this PR those would have exited 1, correctly.

**Test hermeticity.** Ran the suite 4 times. Temp `SKEL` via `KIPI_REPO`, `claude` shimmed at the front of `PATH`, `slack-notify.sh` stubbed. Live log and Slack untouched.

---

## VERDICT: APPROVE WITH NITS

Four minors, no major, no blocker. The fix does what it says, the mechanism survives the subshell it looked like it would die in, and the plist rules out the failure mode that would have made a non-zero exit dangerous.

**Fix first: finding 4.** The alert this PR turns on is one line from being debuggable. Add `>&2` alongside the log write at line 146 (or repoint `StandardErrorPath`) so the filed Linear issue has evidence attached instead of a path to a file that has never existed.

Second: finding 1, a third test case so the per-instance counter is actually pinned.

FINDINGS:
minor|The new test passes with the per-instance failure counter deleted, so it does not pin the line the PR presents as the fix|q-system/.q-system/scripts/open-loops-heartbeat.sh:94
minor|An unreadable instance-registry.json makes the sweep skip all 24 instances and still exit 0 with the audit reporting OK, contradicting the audit's own comment|q-system/.q-system/scripts/open-loops-heartbeat.sh:106
minor|A missing claude CLI exits 0 having swept nothing, writing no run-log and running no audit|q-system/.q-system/scripts/open-loops-heartbeat.sh:34
minor|The failure detail is written only to the job log, so the Linear issue this PR enables points at a StandardErrorPath that does not exist|q-system/.q-system/scripts/open-loops-heartbeat.sh:146
END FINDINGS

PR #14, opened 2026-07-27
fix(worker): take the claim inside the worktree so the board stops serializing (ASK-188)

assafkip:
Closes the serialization cap on the autonomous board. Linear: [ASK-188](https://linear.app/ask-consulting/issue/ASK-188)

## The defect

`linear-worker.sh` took its claim at step 1, **before** it created or entered `$TREE`, so the claim ran with the skeleton as its cwd. `linear-claim.py::claims_path()` resolves the lock from `git rev-parse --show-toplevel` **of the caller's cwd**, so every issue on the board contended for one file at the skeleton root. One lock, whole repo, total serialization.

Measured 2026-07-27: ASK-150 converged in 25 min, ASK-183 in 67 min. With 50+ ready issues that is 20+ hours serial.

## The fix

Inside a linked worktree, `git rev-parse --show-toplevel` returns **that worktree's** path (verified empirically before the edit). Moving the claim to after the worktree exists, and running it from inside `$TREE`, gives each issue its own lock in the tree it actually protects.

Two workers on the SAME issue still share one worktree and still collide with exit 3. The mutex behind commit 53f2eeb is unweakened, and case 4 of the new suite pins that.

## Three things found while verifying the blast radius

The DoR named the release path and the mutex explicitly, so these are in scope rather than adjacent:

1. **Release ran in the wrong cwd too.** A claim taken in one cwd and released in another does not error. Release reads a different file, finds nothing, prints `not held` and exits 0, while the real lock sits in the worktree forever and wedges that issue permanently. Now released from inside `$TREE`, asserted by reading the lock files back after the run.

2. **`rc=$?` under `if ! cmd` reads the negation, not the command.** Bash sets `$?` from the `!`, so it was 0 on every failure and the collision branch was unreachable: a real collision reported as `INFRA: claim failed rc=0`. That is exactly what the reproducer printed on its first RED run. Status is now captured directly.

3. **Worktree-creation race reported a phantom outage.** Two workers on the same issue could both see `! -d $TREE`; the loser logged INFRA. It now falls through and lets the claim adjudicate, which is what the claim is for.

## How this is tested

`test-linear-worker-parallel.sh` boots the **real** `linear-worker.sh` twice, concurrently, and asserts on where the locks landed and which runs reached the work phase.

The bug is a cwd. A grep for line ordering cannot see a cwd, and a re-implementation of the sequence inside the test would only agree with whatever I believed while writing it. So `python3`/`gh`/`claude` are PATH stubs (Linear, GitHub and the model stay out of it) while `git` stays real, because real linked worktrees are the mechanism under test. Worker A is held inside its work phase while B runs start to finish, so the overlap is real rather than two runs that happened not to collide.

Isolation: `KIPI_SKEL` and `KIPI_STATE_DIR` are added as test seams, same discipline as the existing `KIPI_LINEAR_CLAIMS`. Without them the suite would create real worktrees and real `sana/*` branches in the founder's checkout. `KIPI_LINEAR_CLAIMS` is deliberately left UNSET here, because the per-tree default path is the thing under test.

### Observed RED, before the change

```
FAIL: SERIALIZED: worker B on ASK-BBB never reached the work phase while A held
      the lock on ASK-AAA. B said: INFRA: claim failed rc=0 on ASK-BBB
```

### Observed GREEN, after

```
  ok: two workers on DIFFERENT issues both reached the work phase concurrently
  ok: the claim was held in the issue's OWN worktree while it was being worked
  ok: no claim was taken at the skeleton root
  ok: both claims were released in the tree they were taken in (no leaked lock)
  ok: same issue, same tree: exactly one worker reached the work phase
  ok: the refused run reports the collision (exit 3), not a generic infra failure
  ok: linear-worker.sh parses (bash -n)
PASS: linear-worker parallel (7 checks)
```

### Negative self-test

A test that cannot fail is not a test. Three mutations applied by exact string replacement, each caught by its intended assertion:

| Mutation | Caught by |
|---|---|
| claim taken outside the worktree | `SERIALIZED: worker B ... never reached the work phase` |
| release taken outside the worktree | `a claim was left behind ... leaked a permanent lock` |
| old `if ! cmd; then rc=$?` shape | `the refused run did not report a collision` |

### Neighbouring suites, unchanged and green

| Suite | Result |
|---|---|
| `test-linear-claim.sh` | PASS (30 checks) |
| `test-severity-floor.sh` | PASS (24/24) |
| `test-converge.sh` | PASS (13/13) |
| `test-linear-worker-parallel.sh` | PASS (7 checks) |

`test-severity-floor.sh` line 147 asserts the severity gate fires BEFORE the claim. The claim moved later, so that ordering still holds.

## Not doing (DoR binding)

No launchd scheduling, no change to the four refusals, no change to the remote-collision half of the lock.

## Registered

`q-system/.q-system/capability-manifest.json` gains the new suite, so it is not an orphan.

Review runs next. Do not merge without it.

### DIFF
--- a/q-system/.q-system/capability-manifest.json
+++ b/q-system/.q-system/capability-manifest.json
@@ -94,6 +94,10 @@
       "runner": "bash"
     },
     {
+      "path": "q-system/.q-system/scripts/test/test-linear-worker-parallel.sh",
+      "runner": "bash"
+    },
+    {
       "path": "q-system/.q-system/scripts/test/test-linear-queue.sh",
       "runner": "bash"
     },
--- a/q-system/.q-system/scripts/linear-worker.sh
+++ b/q-system/.q-system/scripts/linear-worker.sh
@@ -40,11 +40,18 @@
 set -uo pipefail
 
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
-SKEL="$(cd "$SCRIPT_DIR/../../.." && pwd)"
+# KIPI_SKEL / KIPI_STATE_DIR are TEST-ISOLATION SEAMS, same discipline as
+# KIPI_LINEAR_CLAIMS / KIPI_LINEAR_LEDGER / KIPI_LINEAR_QUEUE. Without them a
+# suite that drives this script end-to-end would create real worktrees and real
+# sana/* branches in the founder's checkout and stomp the live attempts ledger --
+# so the ordering this script's whole correctness rests on could only ever be
+# asserted by grepping its source. Unset in production; the defaults are the
+# real skeleton and the real state dir.
+SKEL="${KIPI_SKEL:-$(cd "$SCRIPT_DIR/../../.." && pwd)}"
 CLAIM="$SCRIPT_DIR/linear-claim.py"
 SYNC="$SCRIPT_DIR/linear-sync.py"
 NOTIFY="$SCRIPT_DIR/slack-notify.sh"
-STATE_DIR="$HOME/.config/kipi"
+STATE_DIR="${KIPI_STATE_DIR:-$HOME/.config/kipi}"
 ATTEMPTS="$STATE_DIR/linear-worker-attempts.json"
 LOG="$STATE_DIR/linear-worker.log"
 REVIEWS_DIR="$STATE_DIR/pr-reviews"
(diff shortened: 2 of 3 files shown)


assafkip:
Killed the background jobs. Writing up.

## Review: PR #14 — `fix(worker): take the claim inside the worktree` (ASK-188)

### What I ran

Mirrored the PR's scripts into an isolated tree and drove the real suite both ways.

**Suite green on the PR branch:**
```
  ok: two workers on DIFFERENT issues both reached the work phase concurrently
  ok: the claim was held in the issue's OWN worktree while it was being worked
  ok: no claim was taken at the skeleton root
  ok: both claims were released in the tree they were taken in (no leaked lock)
  ok: same issue, same tree: exactly one worker reached the work phase
  ok: the refused run reports the collision (exit 3), not a generic infra failure
  ok: linear-worker.sh parses (bash -n)
PASS: linear-worker parallel (7 checks)   [1.7s]
```

**Same suite, same command, pre-PR worker swapped in:**
```
FAIL: SERIALIZED: worker B on ASK-BBB never reached the work phase while A held
      the lock on ASK-AAA. B said: 2026-07-27T08:05:18Z INFRA: claim failed rc=0 on ASK-BBB
```

That is a real reproducer, not a re-implementation. It boots the actual script and reads where the lock files landed. It earns its place.

---

### Finding 1 — minor

**Claim:** when `$TREE` exists but is no longer a valid worktree, `linear-claim.py` returns 3 (`TreeUnknown`, not a collision), and the worker at `linear-worker.sh:265` relabels that as `"claimed by another session"` — a false cause, on every run, with no attempt counted and no page.

**File:** `q-system/.q-system/scripts/linear-worker.sh:265` (PR branch)

The claim script conflates two exits:

```python
except TreeUnknown as exc:
    ...
    return EXIT_COLLISION   # == 3
```

The worker reads only the number:

```bash
265:    if [ "$rc" = "3" ]; then say "skip $ISSUE: working tree is claimed by another session"; continue; fi
```

Pre-PR the claim ran from `$SKEL`, which is always a git repo, so `TreeUnknown` was unreachable. Routing the claim through `$TREE` is what puts a non-git cwd on that path.

**Reproducer:** `.review-scratch/probe-claim.sh` — builds a real worktree, then moves the skeleton (worktree `.git` files hold an absolute path to `$SKEL/.git/worktrees/<name>`, and the worktrees live outside the repo under `$STATE_DIR`, so they survive it).

**Real output:**
```
=== healthy worktree ===
/private/var/.../state/worktrees/ask-aaa
claimed ASK-AAA for sana (session s1)
  claim exit=0

=== now the skeleton is moved/re-cloned; the worktree dir survives ===
  dir still present: yes
  .git file says:    gitdir: /private/var/.../tmp.iagRRK2Uuy/skel/.git/worktrees/ask-aaa
  rev-parse output:
    fatal: not a git repository: (null)
  claim from inside it:
    REFUSED: cannot identify the working tree to lock (`git rev-parse --show-toplevel` failed...)
  >>> claim EXIT CODE = 3   (3 == EXIT_COLLISION, which linear-worker.sh maps to 'claimed by another session')
```

**Severity rationale, stated plainly:** this is major-shaped (a permanently dark issue reporting a benign cause, unattended, no page since `continue` skips `bump_attempt`). I dropped it one level because **I constructed the input with `mv` and no script in this repo moves or re-clones the skeleton.** `git worktree prune` and a re-clone both produce it, but neither is emitted by code here.

Fix is one line: give `TreeUnknown` its own exit code, or have the worker check `[ -d "$TREE/.git" ] || [ -f "$TREE/.git" ]` before trusting a 3.

---

### What is sound — attacks that failed

- **Lock leaking into the public repo.** The claim file now lands inside a tree that gets committed and pushed. `git check-ignore -v` confirms `.gitignore:32` and `:33` cover `.linear-claims.json` and `.linear-claims.json.guard`. No leak.
- **Lock relocating to an ancestor git repo.** If any ancestor of `$STATE_DIR/worktrees/` were a repo, `rev-parse` would walk up and re-serialize the board there. Checked: neither `$HOME` nor `~/.config` is a git repo on this machine. Dropped.
- **Case 4 as decoration.** I traced whether worker D could be skipped before reaching the claim (attempts counter, severity-floor gate). It cannot — `gh` returns no PR, attempts are 0. The `"claimed by another session"` assertion pins the mutex genuinely.
- **The `rc=$?` fix.** `( cd … ) >/dev/null 2>&1; rc=$?` reads the subshell, not a negation. Correct, and the pre-PR `rc=0` is visible in the RED output above.
- **Release-without-claim wedging the tree.** The worktree-add failure path no longer calls `release`, which is right — there is no claim to release at that point.

**What I could not finish:** an end-to-end worker run against the broken worktree. My harness hung twice at the first worker invocation for a reason I did not isolate before running out of budget, so the finding above rests on the claim-script probe plus the two-line mapping in the diff, not on a full worker run. Saying that rather than dressing it up.

**Untested, not a finding:** the suite gives workers A and B *different* `KIPI_STATE_DIR` values (`test-linear-worker-parallel.sh:96-102`), while production shares one. So the topology this PR unlocks — concurrent workers doing read-modify-write on one `linear-worker-attempts.json` with no lock — is never exercised. I did not reproduce a lost update, so it stays out of the findings block.

---

Only minor findings survived reproduction. The core change is correct, the reproducer is real, and the mutex is measurably unweakened.

**Fix first:** stop returning `3` for `TreeUnknown`, so a broken worktree cannot masquerade as a healthy skip.

**Left behind:** `.review-scratch/` in the repo root (untracked). The `rm -r` guard blocked cleanup. Want me to remove it another way?

FINDINGS:
minor|TreeUnknown returns exit 3 so a broken worktree is reported as "claimed by another session", silently darkening the issue with no attempt counted and no page|q-system/.q-system/scripts/linear-worker.sh:265
END FINDINGS

PR #16, opened 2026-07-27
feat(launchd): watchdog findings reach Linear, and --dry-run means dry (ASK-181)

assafkip:
Migrates `com.kipi.launchd-health` onto Linear-tracked execution. ASK-181.

## The four bars

| Bar | Before | After |
|--|--|--|
| 1. launchd, never cron | met (09:30 + 21:30, `StartCalendarInterval`) | unchanged; `schedule-duplicate: 0` confirms no crontab twin |
| 2. failures and findings reach Linear | **not met** — Slack line only | every finding files a deduped Linear issue |
| 3. live or in the pause ledger | met | unchanged; `launchctl list` rc=0, `LastExitStatus = 0` |
| 4. verified by running it | — | ran dry, ran live, ran a live Linear probe |

## What changed

**Findings reach Linear.** A failing or dark job used to produce a Slack line and nothing else. It now also files a Linear issue through `fleet-health-daily.py`'s existing filer, against its existing detector keys (`launchd-failing` / `launchd-dark`).

Sharing those keys is the design, not a shortcut: fleet-health runs the same two checks at 08:15. A private key namespace here would file a **second permanent issue** for every job both of them see, and Linear issues are not deleted in this fleet. `file_findings` gained a `filer` param so the issue names the script that actually filed it. A paused job is never filed, matching "never pinged".

**`--dry-run` now means dry.** Found while running this migration's own checklist. Dry mode was `"--dry" in sys.argv`, an exact string match. The DoR tells the verifier to type `--dry-run`, which did not match — so a read-only check silently took the LIVE path, wrote state, and pinged the founder's phone. `is_dry_run()` accepts `--dry` / `--dry-run` / `-n`.

**Linear down cannot kill the watchdog.** A filing error is caught, counted and printed. A watchdog that dies on a network error stops watching launchd, which is the exact silent death it exists to catch.

## Verification (reproducer-first, RED observed)

RED first: `AttributeError: module 'wd' has no attribute 'is_dry_run'`.

```
$ python3 q-system/.q-system/scripts/test_launchd_health_check.py
PASS: all launchd-health-check logic checks green

PASS: fleet-health-daily contract holds        (12 checks, filer param is additive)

$ python3 q-system/.q-system/scripts/launchd-health-check.py --dry-run | tail -2
[dry] would ping 0 job(s)
[dry] would file 0 linear finding(s)

$ python3 q-system/.q-system/scripts/launchd-health-check.py | tail -1
linear: filed=0 already-tracked=0

$ python3 q-system/.q-system/scripts/fleet-health-daily.py
  launchd-dark: 0        <- this job is loaded, not dark
  launchd-failing: 0
```

16 new test cases. `test_launchd_health_check.py` is already in `capability-manifest.json`, so no manifest line needed.

**A zero result must prove it is empty, not broken.** The fleet is clean, so a live run files 0 — indistinguishable from a broken filer. A probe pushed one synthetic failing job through the real path with `apply=False`:

```
built 1 finding(s) from 2 problem(s) (paused excluded)
  key=fleet-health/launchd-failing/com-kipi-probe-not-a-real-job
  title=launchd job failing: com.kipi.PROBE-not-a-real-job (exit 127)
DRY outcome: {'created': 1, 'existing': 0, 'skipped_no_key': 0}
RESULT: PASS - Linear round trip succeeded; 1 would be filed, 0 already tracked
```

Real `fetch_remote_state` round trip against Linear, real ledger dedup, nothing created. The probe lives in `.review-scratch/probe-linear-path.py` (untracked scratch — it hits the live network, so it does not belong in the capability-gate runner).

## One DoR check that cannot pass as written

The DoR says:

```bash
launchd-health-check.py --dry-run | grep com.kipi.launchd-health
```

That returns nothing, and correctly so — `SELF_LABEL` is excluded from the watchdog's own scan. A job cannot watch itself; an unloaded job cannot report that it is unloaded. Self-coverage comes from `fleet-health-daily.py`, which scans all `com.kipi.*` with no self-exclusion. `launchd-dark: 0` above is that proof.

## Blast radius

Machine-local scheduling, unchanged. The two script edits ship via `kipi update` to the fleet; both are additive (a new flag alias, a new filing call, a defaulted param). Live effect on the founder's machine lands when `~/projects/kipi-system` pulls main — the plist points at that checkout.

## Not doing

What the job checks is unchanged. No prefix-list, classifier, or pause-ledger edits.

### DIFF
--- a/q-system/.q-system/scripts/fleet-health-daily.py
+++ b/q-system/.q-system/scripts/fleet-health-daily.py
@@ -465,6 +465,23 @@ def file_findings(findings: list, apply: bool, filer: str = "fleet-health-daily.
     return result
 
 
+def outcome_line(outcome: dict) -> str:
+    """The filing result as one line that cannot lie by omission.
+
+    Scar (ASK-181 review): this printed `created` and `existing` only. Because
+    `file_findings` catches its own network errors and returns
+    {created: 0, existing: 0, skipped_no_key: N}, 'Linear was unreachable and N
+    findings went nowhere' printed byte-identically to 'the fleet is clean,
+    nothing to file'. `unfiled` is what separates empty from broken.
+
+    launchd-health-check.py keeps its OWN copy of this formatter on purpose: it
+    has to be able to report that THIS file is missing, which is the rsync
+    --delete scar it was built for. Do not consolidate them into a dependency
+    that disappears exactly when it is needed."""
+    return (f"  filed={outcome['created']} already-tracked={outcome['existing']} "
+            f"unfiled={outcome['skipped_no_key']}")
+
+
 def main() -> int:
     ap = argparse.ArgumentParser(description=__doc__.split("\n")[0])
     ap.add_argument("--apply", action="store_true", help="actually file Linear issues")
@@ -497,7 +514,7 @@ def main() -> int:
         print(f"  {did}: {n}")
 
     outcome = file_findings(all_findings, args.apply)
-    print(f"  filed={outcome['created']} already-tracked={outcome['existing']}")
+    print(outcome_line(outcome))
 
     STATE.parent.mkdir(parents=True, exist_ok=True)
     STATE.write_text(json.dumps(
--- a/q-system/.q-system/scripts/fleet-health-daily.py
+++ b/q-system/.q-system/scripts/fleet-health-daily.py
@@ -407,12 +407,17 @@ def finding_key(detector_id: str, subject: str) -> str:
     return f"fleet-health/{detector_id}/{slug(subject)}"
 
 
-def file_findings(findings: list, apply: bool) -> dict:
+def file_findings(findings: list, apply: bool, filer: str = "fleet-health-daily.py") -> dict:
     """Create a Linear issue per finding, deduped by kipi-key.
 
     Reuses linear-sync's graphql + remote guard so 'already exists' has exactly one
     definition fleet-wide. Linear objects are permanent, so the guard is refetched
     here rather than trusted from any cache.
+
+    `filer` names the script that produced the finding. It is a parameter rather
+    than a constant because this is the fleet's ONE filer: launchd-health-check.py
+    files its 09:30/21:30 findings through here too (ASK-181), against these same
+    detector keys, so a finding both jobs see stays one issue instead of two.
     """
     result = {"created": 0, "existing": 0, "skipped_no_key": 0}
     if not findings:
@@ -443,7 +448,7 @@ def file_findings(findings: list, apply: bool) -> dict:
             continue
         payload = {
             "title": f["title"][:250],
-            "description": f"<!-- kipi-key: {key} -->\n\n{f['body']}\n\nFiled by `fleet-health-daily.py`.",
+            "description": f"<!-- kipi-key: {key} -->\n\n{f['body']}\n\nFiled by `{filer}`.",
             "teamId": team_id,
         }
         if project:
--- a/q-system/.q-system/scripts/launchd-health-check.py
+++ b/q-system/.q-system/scripts/launchd-health-check.py
@@ -17,14 +17,19 @@ wiped their scripts out from under the plists. Nothing surfaced it -- the jobs j
 silently stopped hunting income. A prompt cannot watch launchd; this job can. It is
 the deterministic backstop that turns a silent 127 into a phone ping within hours.
 
+Every finding also becomes a Linear issue (ASK-181), deduped forever by the SAME
+kipi-key fleet-health-daily.py uses for its own launchd checks. A Slack line is
+read once and scrolls away; the issue holds state. Slack stays the pointer.
+
 Single notification channel: slack-notify.sh (founder-notifications rule). Silent
 no-op if no webhook is configured, so this watchdog never breaks anything.
 
 The watchdog always exits 0 -- it must never become the failing job it reports.
 
 Usage:
-  launchd-health-check.py            # check; ping on newly/again-failing jobs
-  launchd-health-check.py --dry      # print findings only; no ping, no state write
+  launchd-health-check.py             # check; ping + file Linear issues
+  launchd-health-check.py --dry-run   # print findings only; no ping, no issue,
+  launchd-health-check.py --dry       #   no state write (--dry / -n are aliases)
 """
 import json
 import re
@@ -37,6 +42,7 @@ SELF_LABEL = "com.kipi.launchd-health"
 FAIL_PING_TTL_SECONDS = 6 * 3600  # re-ping a still-failing job at most this often
 HERE = Path(__file__).resolve().parent
 NOTIFY_SCRIPT = HERE / "slack-notify.sh"
+DRY_FLAGS = ("--dry", "--dry-run", "-n")
 STATE_FILE = Path.home() / ".config" / "kipi" / "launchd-health-state.json"
 LAUNCH_AGENTS = Path.home() / "Library" / "LaunchAgents"
 
@@ -209,6 +215,105 @@ def send_ping(message):
         pass
 
 
+def is_dry_run(args):
+    """True when the operator asked for a read-only pass.
+
+    Scar (ASK-181): this used to be `"--dry" in sys.argv`, an exact string match.
+    `--dry-run` -- the flag this job's own migration checklist tells the verifier
+    to type -- did not match, so a read-only check silently took the LIVE path and
+    pinged the founder's phone. An unrecognized flag must not arm anything."""
+    return any(arg in DRY_FLAGS for arg in args)
+
+
+# --- findings reach Linear, not only Slack -----------------------------------
+# Bar 2 of the job-migration contract (ASK-181). The detector ids below are
+# fleet-health-daily.py's, not new ones: that job runs the same two checks at
+# 08:15 and files against those keys. Sharing the key means a job both of them see
+# is ONE permanent issue; a private namespace here would file a second one every
+# time, and Linear issues are not deleted in this fleet.
+LINEAR_DETECTOR_BY_KIND = {
+    "failing": "launchd-failing",
+    "not_loaded": "launchd-dark",
+    # "paused" is deliberate, recorded in the pause ledger, and never filed.
+}
+
+_FLEET_HEALTH = None
+
+
+def _fleet_health():
+    """Lazily load fleet-health-daily.py -- the one place that defines a finding's
+    Linear key and how it gets filed. Loaded on demand so a healthy run (the
+    common case) pays nothing, and so an import error cannot stop the watchdog."""
+    global _FLEET_HEALTH
+    if _FLEET_HEALTH is None:
+        import importlib.util
+
+        spec = importlib.util.spec_from_file_location("fh", HERE / "fleet-health-daily.py")
+        module = importlib.util.module_from_spec(spec)
+        spec.loader.exec_module(module)
+        _FLEET_HEALTH = module
+    return _FLEET_HEALTH
+
+
+def linear_findings(problems):
+    """Turn watchdog problems into fleet-health finding dicts, ready to file."""
+    fleet_health = _fleet_health()
+    findings = []
+    for label, kind, detail in problems:
+        detector = LINEAR_DETECTOR_BY_KIND.get(kind)
+        if detector is None:
+            continue
+        if kind == "failing":
+            title = f"launchd job failing: {label} ({detail})"
+            body = (
+                f"`{label}` is loaded but its last run exited non-zero (**{detail}**).\n\n"
+                "## Action\nRead its `StandardErrorPath`, fix the cause, and confirm a clean "
+                "run. If the failure is environmental (auth expiry, server down), say so on "
+                "this issue rather than retrying -- `self-healing-retry.md` rule 5 stops "
+                "environmental failures on attempt 1."
+            )
+        else:
+            title = f"launchd job is dark: {label}"
+            body = (
+                f"`{label}` has a plist in `~/Library/LaunchAgents/` but is not loaded, and "
+                "it is NOT in the paused ledger -- so nothing recorded a decision to stop "
+                "it. An unloaded job cannot report its own death.\n\n"
+                "## Action\n"
+                f"- Resume: `launchctl bootstrap gui/$(id -u) ~/Library/LaunchAgents/{label}.plist`\n"
+                f"- Or record the pause: add `{label}` to `~/.config/kipi/launchd-paused.txt`"
+            )
+        findings.append({
+            "key": fleet_health.finding_key(detector, label),
+            "detector": detector,
+            "subject": label,
+            "title": title,
+            "body": body,
+        })
+    return findings
+
+
+def file_linear_findings(problems):
+    """File the findings as deduped Linear issues. Never raises.
+
+    A watchdog that dies because Linear is unreachable stops watching launchd --
+    which is precisely the silent death it exists to catch. A filing failure is
+    counted and printed, never propagated."""
+    empty = {"created": 0, "existing": 0, "skipped_no_key": 0}
+    try:
+        findings = linear_findings(problems)
+    except Exception as exc:  # noqa: BLE001
+        print(f"linear findings could not be built: {exc}", file=sys.stderr)
+        return empty
+    if not findings:
+        return empty
+    try:
+        return _fleet_health().file_findings(
+            findings, apply=True, filer="launchd-health-check.py")
+    except Exception as exc:  # noqa: BLE001
+        print(f"linear filing skipped: {exc}", file=sys.stderr)
+        return {"created": 0, "existing": 0, "skipped_no_key": len(findings)}
+
+
 def problems_to_ping(problems, state, now):
     """Problems whose kind changed since the last ping, or whose last ping is
     older than the TTL (dedupe spam, but re-ping when failing -> not_loaded)."""
@@ -247,7 +352,11 @@ def run(dry_run):
 
     if dry_run:
         print(f"[dry] would ping {len(due)} job(s)")
+        print(f"[dry] would file {len(linear_findings(problems))} linear finding(s)")
         return
+
+    filed = file_linear_findings(problems)
+    print(f"linear: filed={filed['created']} already-tracked={filed['existing']}")
 
     if due:
         parts = [
@@ -268,6 +377,6 @@ def run(dry_run):
 
 if __name__ == "__main__":
     try:
-        run("--dry" in sys.argv)
+        run(is_dry_run(sys.argv[1:]))
     finally:
         sys.exit(0)  # a watchdog must never report itself as the failing job
--- a/q-system/.q-system/scripts/launchd-health-check.py
+++ b/q-system/.q-system/scripts/launchd-health-check.py
@@ -30,12 +30,17 @@ Usage:
   launchd-health-check.py             # check; ping + file Linear issues
   launchd-health-check.py --dry-run   # print findings only; no ping, no issue,
   launchd-health-check.py --dry       #   no state write (--dry / -n are aliases)
+
+Dry mode makes ONE read-only Linear query (the dedup fetch) so its preview of a
+permanent, undeletable action is the real number. It writes nothing: no issue, no
+state file, no ping. Any flag that is not a dry alias is REFUSED, never run live.
 """
 import json
 import re
 import subprocess
 import sys
 import time
+import traceback
 from pathlib import Path
 
 SELF_LABEL = "com.kipi.launchd-health"
@@ -216,13 +221,32 @@ def send_ping(message):
 
 
 def is_dry_run(args):
-    """True when the operator asked for a read-only pass.
-
-    Scar (ASK-181): this used to be `"--dry" in sys.argv`, an exact string match.
-    `--dry-run` -- the flag this job's own migration checklist tells the verifier
-    to type -- did not match, so a read-only check silently took the LIVE path and
-    pinged the founder's phone. An unrecognized flag must not arm anything."""
+    """True when a dry alias was typed. Answers only that; see parse_mode."""
     return any(arg in DRY_FLAGS for arg in args)
+
+
+def parse_mode(args):
+    """('live'|'dry'|'refuse', unrecognized_flags).
+
+    Scar (ASK-181): dry mode used to be `"--dry" in sys.argv`, an exact string
+    match. `--dry-run` -- the flag this job's own migration checklist tells the
+    verifier to type -- did not match, so a read-only check silently took the LIVE
+    path and pinged the founder's phone.
+
+    Widening the accept-list fixed that one typo and left the shape intact: every
+    OTHER near-miss (`--dry-run=1`, `--dryrun`, `--dry_run`, `-dry`) still armed
+    the live path. So an unrecognized flag now REFUSES the run instead of guessing.
+    An unknown flag alongside a valid dry one also refuses -- if part of the
+    command line is unparsed, the operator's intent is unknown, and the safe
+    reading of an unknown intent is 'do nothing'.
+
+    launchd invokes this script with no arguments, so refusing cannot stop the
+    scheduled run; it only stops a hand-typed command from doing more than the
+    operator asked."""
+    unrecognized = [arg for arg in args if arg not in DRY_FLAGS]
+    if unrecognized:
+        return ("refuse", unrecognized)
+    return ("dry" if is_dry_run(args) else "live", [])
 
 
 # --- findings reach Linear, not only Slack -----------------------------------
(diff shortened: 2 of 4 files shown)


assafkip:
## Review: PR #16 — `feat(launchd): watchdog findings reach Linear, and --dry-run means dry (ASK-181)`

Repro scripts live in `.review-scratch/pr16-*.py` (same untracked scratch convention the PR body uses). Every output below is pasted verbatim from a run.

---

### FINDING 1 — major

**A dead Linear filer prints the exact same stdout line as a clean run. `skipped_no_key` is computed, then thrown away.**

`q-system/.q-system/scripts/launchd-health-check.py:359`

`file_findings` does not raise when Linear is unreachable — it catches internally and returns `{created: 0, existing: 0, skipped_no_key: N}`. The new print reads only `created` and `existing`. Both are 0 on total failure and 0 on a clean run.

Reproducer, real `file_findings`, real `urllib` → Linear, bogus key (the branch every fleet instance without `~/.config/kipi/linear-api-key` takes — and this ships fleet-wide via `kipi update`):

```
$ python3 .review-scratch/pr16-repro-real.py
--- what launchd writes to StandardOutPath (launchd-health.out) ---
FAILING: com.claudedaddy.pinterest -- exit 127
NOT_LOADED: com.kipi.opp-scan -- installed but not running
linear: filed=0 already-tracked=0
--- what launchd writes to StandardErrorPath (launchd-health.err) ---
  linear unreachable, findings NOT filed: HTTP 401: {"errors":[{"message":"Authentication required, not authenticated",...

2 real findings existed. Issues actually filed: 0
stdout line is byte-identical to a clean run: True
ledger file written? False
```

And the three-way comparison (`pr16-repro.py`, REPRO A):

```
A1 all-already-tracked stdout tail : 'linear: filed=0 already-tracked=2'
A2 FILER TOTALLY DEAD  stdout tail : 'linear: filed=0 already-tracked=0'
A3 nothing to file     stdout tail : 'linear: filed=0 already-tracked=0'
A2 == A3 last line?                : True
skipped_no_key appears anywhere in stdout?        : False
```

At 3am this is a false all-clear, which is the same class of harm as crying wolf. The PR body's own PASS evidence is `linear: filed=0 already-tracked=0` — the author's verification cannot distinguish working from dead either. This is the fleet's own lesson *"a zero result must prove it is empty, not broken"*, which the PR quotes.

Honest mitigation: `fleet-health-daily.py` runs `--apply` at 08:15 against the same two detectors, so `com.kipi.*` / `com.cole.*` / `com.ask.*` / `com.assaf.*` findings still land via that path. Uncovered if the watchdog's filer is dead: `com.claudedaddy.*` failing jobs and every `EXTRA_PREFIXES_FILE` family. Fix is one string — `filed['skipped_no_key']` is already in the dict.

---

### FINDING 2 — minor

**`[dry] would file N` counts before dedup; the live run files 0.**

`launchd-health-check.py:355`

`linear_findings()` builds the list; dedup happens later inside `file_findings` against ledger + remote. The dry preview of a permanent, undeletable action over-reports.

```
$ python3 .review-scratch/pr16-repro.py   # REPRO B
dry  : ['[dry] would file 2 linear finding(s)']
live : ['linear: filed=0 already-tracked=2']
```

Errs conservative (over-states creation), so bounded.

---

### FINDING 3 — minor

**The docstring says "An unrecognized flag must not arm anything." No code path refuses; every unrecognized flag arms the live path.**

`launchd-health-check.py:225` (claim at `:224`)

```
$ python3 .review-scratch/pr16-repro.py   # REPRO C
  --dry-run            is_dry_run=True  -> DRY
  --dry-run=1          is_dry_run=False -> LIVE (files permanent Linear issues + pings)
  --dryrun             is_dry_run=False -> LIVE (files permanent Linear issues + pings)
  --dry_run            is_dry_run=False -> LIVE (files permanent Linear issues + pings)
  -dry                 is_dry_run=False -> LIVE (files permanent Linear issues + pings)
  --check              is_dry_run=False -> LIVE (files permanent Linear issues + pings)
```

The new test `check("an unrelated flag is live", wd.is_dry_run(["--verbose"]), False)` codifies the opposite of the docstring. The scar was a typo'd flag taking the live path; the fix widened the accept-list instead of making unknown args fail closed. Live blast radius is genuinely low (it files real findings for real problems, deduped) — hence minor, not major. Either add an `argparse`/unknown-arg refusal or delete the sentence.

---

### FINDING 4 — minor

**`_fleet_health`'s docstring says "an import error cannot stop the watchdog." On the dry path the call is unguarded, and the crash is completely silent — no traceback, exit 0.**

`launchd-health-check.py:246` (unguarded call at `:355`)

`file_linear_findings` wraps it, so the live path holds. `run()`'s dry branch calls `linear_findings()` bare. `sys.exit(0)` in the `finally` supersedes the propagating exception, so the traceback never prints.

Removed `fleet-health-daily.py` from beside the script (literally the `kipi update` rsync `--delete` scar this watchdog exists for) and ran `--dry-run`:

```
exit code : 0   <- launchd sees SUCCESS
stdout    : '...PAUSED: com.assaf.competitive-analysis.morning -- paused on purpose\n[dry] would ping 0 job(s)\n'
stderr    : ''
```

The `[dry] would file` line just vanishes. Nothing anywhere says why.

---

### FINDING 5 — minor

**The new key-collision tests pin the watchdog to `finding_key()`'s formula, not to what fleet-health's producers emit. Producer drift keeps the suite green while the fleet files duplicate permanent issues.**

`q-system/.q-system/scripts/test_launchd_health_check.py:157`

`check(..., _found[0]["key"], _fh.finding_key("launchd-failing", "com.kipi.alpha"))` restates what `linear_findings` already does. It never calls `detect_dark_jobs` / `detect_failing_jobs`. I changed fleet-health's dark-job `subject` from `label` to `f"{label}.plist"` — a plausible "improve the issue" edit — and re-ran the suite:

```
$ python3 .review-scratch/pr16-fixture-shape.py
baseline suite: ... | exit 0
after producer drift, suite: ... | exit 0
  fleet-health 08:15 files : fleet-health/launchd-dark/com-kipi-opp-scan-plist
  watchdog 09:30 files     : fleet-health/launchd-dark/com-kipi-opp-scan
  -> keys collide?         : False   (False = TWO permanent Linear issues for one job, forever)
```

Green both times. No producer in the repo emits this today (I confirmed the keys currently match — see below), so per the calibration rule this stays minor. One assert driving `fh.detect_dark_jobs`/`detect_failing_jobs` and comparing to the watchdog's key would close it.

---

## What is sound

- **The central claim — shared keys really do collide.** I drove fleet-health's *real* producers (`detect_dark_jobs` with stubbed `_launchd_labels`/`_is_loaded`, `detect_failing_jobs` through its real `launchctl list` parser) and compared against the watchdog's filed key. Identical for both detectors: `fleet-health/launchd-dark/com-claudedaddy-pinterest-poster`, `fleet-health/launchd-failing/com-kipi-opp-scan`. Registry ids `launchd-dark` / `launchd-failing` match the hardcoded map; both producers emit `subject: label`. No duplicate permanent issues today. This was my main line of attack and it held.
- **`--dry-run` is genuinely dry.** No network, no state file, no ping: `side effects during dry run: NONE`, `state file created?: False`.
- **The new capability does not gate the old one.** With Linear returning 401, the Slack ping still fires: `ping fired despite Linear 401: True` — `launchd watchdog: 2 job issue(s) -- com.claudedaddy.pinterest-poster (installed but NOT running), com.kipi.opp-scan (exit 127)`. Filing runs first but its failure is caught, so the primary path is intact.
- **Paused is never filed and never pinged.** `wd.linear_findings([...paused...])` returns `[]`; a paused-only live run prints the job and pings nothing.
- **No unbounded hang.** `linear-sync.graphql` uses `timeout=30` per call, so the pre-ping network work is bounded, not a wedge.
- **No import recursion** despite `fleet-health-daily._paused_labels()` loading the watchdog and `launchd-health-check._fleet_health()` loading fleet-health — both are lazy and neither runs at module top level.
- **Concurrency:** I tried to build a double-file scenario. 08:15 vs 09:30/21:30 don't overlap, and the remote guard is refetched at create time. I could only make it duplicate with a hand-constructed simultaneous run no scheduler produces. **Dropped.**

---

## VERDICT: REQUEST CHANGES

One major, four minor. Not a BLOCK — nothing here is unrecoverable, and the shared-key design (the part with permanent blast radius) is correct and I verified it against the real producers.

**Fix first:** `launchd-health-check.py:359` — print `skipped_no_key`, so a dead filer cannot render as a clean run. The value is already in the dict; it is a one-line change and it is the difference between this watchdog watching and this watchdog claiming to watch.

---

FINDINGS:
major|A dead Linear filer prints `linear: filed=0 already-tracked=0`, byte-identical to a clean run, because skipped_no_key is computed and discarded|q-system/.q-system/scripts/launchd-health-check.py:359
minor|`[dry] would file N` counts findings before dedup, so dry previews 2 where the live run files 0|q-system/.q-system/scripts/launchd-health-check.py:355
minor|is_dry_run's docstring claims an unrecognized flag must not arm anything, but --dry-run=1 / --dryrun / --dry_run all take the live path|q-system/.q-system/scripts/launchd-health-check.py:225
minor|_fleet_health claims an import error cannot stop the watchdog, but the dry path's call is unguarded and the crash is silent with exit 0|q-system/.q-system/scripts/launchd-health-check.py:246
minor|The key-collision tests restate finding_key's formula instead of driving fleet-health's producers, so producer drift keeps the suite green while duplicate permanent issues get filed|q-system/.q-system/scripts/test_launchd_health_check.py:157
END FINDINGS

assafkip:
## Round 2: all five findings fixed, each with a test that fails without the fix

Commit `ec9301d`. Repro scripts in `.review-scratch/pr16r2-*` (untracked, same convention). Every output below is pasted from a run.

---

### FINDING 1 (major) — FIXED

Your reproducer, re-run against the fixed code, real `file_findings` / real `urllib` / bogus key:

```
$ python3 .review-scratch/pr16r2-dead-filer.py
--- StandardOutPath (launchd-health.out) with the filer DEAD ---
FAILING: com.claudedaddy.pinterest -- exit 127
NOT_LOADED: com.kipi.opp-scan -- installed but not running
linear: filed=0 already-tracked=0 unfiled=2
--- StandardErrorPath ---
linear unreachable, findings NOT filed: HTTP 401: {"errors":[{"message":"Authentication required, ...
--- the Slack ping the founder actually reads ---
launchd watchdog: 2 job issue(s) -- com.claudedaddy.pinterest (exit 127), com.kipi.opp-scan (installed but NOT running) [NOT filed to Linear: 2]

dead-filer last line : linear: filed=0 already-tracked=0 unfiled=2
clean-run last line  : linear: filed=0 already-tracked=0 unfiled=0
stdout byte-identical to a clean run: False
the ping says the findings never reached Linear: True
```

Three things, not the one string:

**a. stdout.** `unfiled` is in the line. Your A2/A3 case is now distinguishable.

**b. The layer above the counter — the report.** stdout is a file nobody opens at 3am. The Slack ping is the channel actually read, and it was saying "here are 2 problems" while the issues meant to hold them never reached the board. It now carries `[NOT filed to Linear: N]`. **No new ping is created and no new cadence exists** — the suffix rides the ping that was already firing, so ping volume is unchanged. The honest cost: if a job is inside its 6h TTL, no ping fires and the dead filer is stdout-only until the TTL expires. I chose that over a second ping path, because a keyless fleet instance would otherwise page every 6h forever about a config state.

**c. One layer deeper than your finding.** If the findings cannot even be *built* — `fleet-health-daily.py` gone, literally the rsync `--delete` scar — `file_linear_findings` returned all-zeros and printed the clean line again. So `skipped_no_key` now counts problems **owed** an issue, not findings successfully constructed:

```
a findings-build crash reports the problems it could not file  -> unfiled=2
a findings-build crash does not count the paused job           -> unfiled=0
```

**Second code path, which your finding did not cover and I am flagging rather than burying:** `fleet-health-daily.py:500` discarded `skipped_no_key` identically, and that is the 08:15 job filing the *same* findings. Fixed the same way (`outcome_line()`), with its own tests in `test-fleet-health-daily.py` including `main() reports through outcome_line` so the function cannot exist uncalled. The two formatters are deliberately NOT shared — the watchdog has to be able to report that `fleet-health-daily.py` is missing, so depending on it for that message would break exactly when needed. Comment says so at both ends.

---

### FINDING 2 (minor) — FIXED

Root cause was the shape you named elsewhere: two readers of one input. Dry counted findings built, live counted what survived dedup. Now there is one reader — the same `file_findings`, with `apply=False`:

```
the dry preview does not claim it would file already-tracked findings -> would-file=0
the dry preview reports what is already tracked                       -> already-tracked=2
dry runs the SAME filer, with apply off  : [False]
dry still writes no state                : []
dry still sends no ping                  : []
```

**Tradeoff I am stating rather than hiding, because it touches something you verified as sound:** dry now makes **one read-only GraphQL query** (the dedup fetch). Your `side effects during dry run: NONE` and `state file created?: False` both still hold — nothing is created, written or pinged, and the tests above pin all three. But your "no network" observation is deliberately no longer true. A preview of a permanent, undeletable action is worth one read; the alternative was a second dedup estimator that disagrees with the real one, which is the defect itself.

---

### FINDING 3 (minor) — FIXED (refusal, not the deleted sentence)

You offered two exits; I took the argparse-style one, because the scar was a typo'd flag going live and widening the accept-list only moved the typo.

```
  --dry-run=1 -> exit 0 | unrecognized flag(s): --dry-run=1 -- refusing to run. Dry: --dry / --dry-run / -n. Live: no flags at all.
  --dryrun    -> exit 0 | unrecognized flag(s): --dryrun -- refusing to run. ...
  --dry_run   -> exit 0 | ...
  -dry        -> exit 0 | ...
  --check     -> exit 0 | ...
```

An unknown flag *next to* a valid dry flag also refuses — if part of the command line is unparsed, intent is unknown, and the safe reading of unknown intent is "do nothing". launchd invokes with no arguments, so refusal cannot touch the scheduled run; a test pins that a refused flag never reaches `run()` and still exits 0.

Your `check("an unrelated flag is live", ...)` line — you were right that it codified the opposite. `is_dry_run` is now only the predicate "was a dry alias typed"; `parse_mode` owns the behavior, and the old assertions are renamed to say only what they test.

---

### FINDING 4 (minor) — FIXED, both halves

The dry-path hole closes because dry now goes through the guarded `file_linear_findings`. But the silence was the worse half and it was general — **any** unhandled crash printed nothing and launchd recorded SUCCESS, because `sys.exit(0)` in the `finally` supersedes a propagating exception. Extracted `main(argv)`, which catches, prints the traceback, and still returns 0:

```
a crash still exits 0            : 0
a crash prints its traceback     : True
the traceback names the real error: True
```

Exiting 0 is the contract. Exiting 0 *silently* was the watchdog dying the death it exists to catch.

---

### FINDING 5 (minor) — FIXED, and I ran your drift edit to prove it

The tests now drive fleet-health's real `detect_dark_jobs` / `detect_failing_jobs` (stubbed `_launchd_labels` / `_paused_labels` / `_is_loaded` / `subprocess`, real parsers) and compare the key each side would actually file. I applied your exact edit (`subject: label` -> `f"{label}.plist"`) under try/finally:

```
$ python3 .review-scratch/pr16r2-producer-drift.py
baseline suite            : exit 0 | PASS: all launchd-health-check logic checks green
after producer drift      : exit 1
  FAIL:
    - fleet-health's dark producer emitted the job under test: got ['com.kipi.beta.plist'], want ['com.kipi.beta']
    - the watchdog's dark key equals the key fleet-health's PRODUCER files: got 'fleet-health/launchd-dark/com-kipi-beta', want 'fleet-health/launchd-dark/com-kipi-beta-plist'
after revert              : exit 0 | PASS: all launchd-health-check logic checks green
RESULT: PASS - producer drift turns the suite RED
```

Also added: every id in `LINEAR_DETECTOR_BY_KIND` must exist in fleet-health's `DETECTORS` registry, so a detector rename cannot leave the shared namespace shared with nothing.

---

## What I did NOT regress

Re-ran the attacks you said held:

```
$ bash .review-scratch/pr16r2-verify.sh
=== 1. --dry-run ===
PAUSED: com.assaf.competitive-analysis.morning -- paused on purpose
[dry] linear: would-file=0 already-tracked=0 unfiled=0
[dry] would ping 0 job(s)
=== 2. live ===
linear: filed=0 already-tracked=0 unfiled=0
=== 4. fleet-health-daily.py, the 08:15 sibling ===
  filed=0 already-tracked=0 unfiled=0
```

- **Shared keys still collide** — now pinned by the producer-driven test above, not by restating the formula.
- **Dry writes nothing** — no issue, no state, no ping, all three asserted. Network read is the one deliberate change, called out above.
- **Linear down does not gate the Slack ping** — the ping still fires on a 401 (see finding 1's output), now with the truth appended.
- **Paused is never filed and never pinged** — unchanged, tests unchanged.
- **The watchdog still always exits 0** — including on a refused flag and on an unhandled crash.

Suites: `test_launchd_health_check.py` PASS (now 44 cases), `test/test-fleet-health-daily.py` PASS (17 cases). Both already in `capability-manifest.json`.

assafkip:
Read the diff and both scripts cold, ran my own repros, then read rounds 1 and 2. Findings below.

## FINDING 1 — minor

**The comment that justifies sharing the permanent dedup namespace says fleet-health "runs the same two checks." It does not. The two checks disagree on real machine state.**

`q-system/.q-system/scripts/launchd-health-check.py:254`

```python
# fleet-health-daily.py's, not new ones: that job runs the same two checks at
# 08:15 and files against those keys.
```

The watchdog's failing-check is `classify_status` (`normalize_exit` + the SIGTERM/live-PID carve-out, scar 2026-07-18). `fleet-health-daily.py:129` `detect_failing_jobs` has neither: it does `int(status)` on the raw `launchctl list` column. Same key, different verdict.

Reproducer, one launchctl fixture per case fed to both real code paths:

```
$ python3 - <<'PY'   # (full script pasted below the review)
com.cole.linkedin-session
  shared key    = fleet-health/launchd-failing/com-cole-linkedin-session
  watchdog      -> (files nothing)
  fleet-health  -> launchd job failing: com.cole.linkedin-session (exit 15)
  DIVERGE on identical machine state

com.kipi.opp-scan
  shared key    = fleet-health/launchd-failing/com-kipi-opp-scan
  watchdog      -> launchd job failing: com.kipi.opp-scan (exit 3)
  fleet-health  -> launchd job failing: com.kipi.opp-scan (exit 768)
  DIVERGE on identical machine state

RESULT: 2/2 diverge -> FAIL (claim falsified)
```

`com.cole.linkedin-session` is a live label on this machine and is the exact false positive `classify_status` was written to suppress. Both fixtures are real launchctl shapes (raw wait-status 15 with a live PID; `exit(3)` arriving as `3<<8 = 768`).

Consequence, bounded: 08:15 fires before 09:30, so fleet-health owns the permanent title. A job that really exited 3 gets an undeletable issue titled `exit 768`, and the linkedin-session false positive lands under a key the watchdog can now never correct. Both are `fleet-health-daily.py`'s unchanged behavior, which is why this is minor and not major. What this PR adds is the false equivalence as load-bearing documentation for the next person editing either detector.

Why not major: nothing new is filed by this diff, no human action is created at 3am, and the misstatement is recoverable by editing a comment or unifying the classifier.

## FINDING 2 — minor

**The PR description's Verification block does not reproduce against `ec9301d`. Those receipts are from the pre-round-2 revision.**

`q-system/.q-system/scripts/launchd-health-check.py:350` (`linear_report_line`)

PR body claims:

```
$ launchd-health-check.py --dry-run | tail -2
[dry] would ping 0 job(s)
[dry] would file 0 linear finding(s)

$ launchd-health-check.py | tail -1
linear: filed=0 already-tracked=0
```

Actual, PR head, live machine:

```
$ python3 .../pr16rev2/work/q-system/.q-system/scripts/launchd-health-check.py --dry-run | tail -3
PAUSED: com.assaf.competitive-analysis.morning -- paused on purpose
[dry] linear: would-file=0 already-tracked=0 unfiled=0
[dry] would ping 0 job(s)
```

Different text, different order, and `unfiled=` is absent from both pasted lines. `[dry] would file 0 linear finding(s)` is not a string this code can emit. So the "ran dry, ran live" evidence in the permanent merge record was never re-run after the round-1 fixes. In a repo whose gate is receipts rather than prompts, a stale receipt is the defect; the code itself is fine here.

## What is sound

Attacks I ran that held:

- **The round-1 major is really fixed, against the real scar and not a stub.** I physically moved `fleet-health-daily.py` out from beside the script (the `kipi update` rsync `--delete` case) and drove both modes:
  ```
  DRY  exit=0  stderr: linear findings could not be built: [Errno 2] No such file or directory: ...
  LIVE exit=0  stdout: linear: filed=0 already-tracked=0 unfiled=2
               ping  : launchd watchdog: 2 job issue(s) -- ... [NOT filed to Linear: 2]
  ```
  `owed`-based counting is the right call: counting built findings would have printed `unfiled=0` here.
- **Round-1 finding 5's fixture-shape fix is genuine.** The new asserts drive `fh.detect_dark_jobs` / `fh.detect_failing_jobs` through a tab-separated `launchctl list` fixture that matches the real producer, and `LINEAR_DETECTOR_BY_KIND.values() - registry_ids` catches a detector rename. Both can go red.
- **Both suites are actually wired.** `test_launchd_health_check.py:198` and `test/test-fleet-health-daily.py:17` are in `capability-manifest.json`, so the new checks run in the gate rather than being declared and skipped. Both green on PR head.
- **The test suite makes zero network calls.** Verified by run; every Linear edge is stubbed or `_Boom`.
- **`--dry-run` writes nothing.** `file_findings(apply=False)` reaches `result["created"] += 1; continue` before any mutation; `fetch_remote_state` issues only queries. No ledger append, no state file, no ping.
- **`unfiled` inside the 6h TTL is stdout-only, exactly as the author disclosed.** I reproduced it (2 problems pinged 1h ago, dead filer, live run: `unfiled=2`, `slack pings: []`) and then dropped it. The author named this tradeoff in round 2, and the schedule makes it narrower than they claimed: 09:30 and 21:30 are 12h apart against a 6h TTL, so a scheduled run always re-pings. Only a manual re-run inside 6h hits it. **Dropped, not re-litigated.**
- **Recurrence after an issue is closed.** `known = set(ledger) | set(remote_keys)`, the ledger is append-only and never pruned on close, and `PROJECT_ISSUES_QUERY` has no state filter, so a re-failure after a fix would report `already-tracked` and file nothing. I could not build a network-free reproducer of the shipped `file_findings` (it `importlib`-loads `linear-sync` internally, so there is no injection point), and it is unchanged pre-existing design. **Dropped for lack of an executed repro.**
- **Concurrency.** 08:15 vs 09:30/21:30 do not overlap and the remote guard refetches at create time. Same conclusion as round 1, same reason. Dropped.

**Review-process note:** round 1 was calibrated. Its one major was real, all five fixes ship with asserts that fail without them, and round 2 (this pass) produced only minors on the same code. That is convergence, not the round-3 pathology. The one gap in round 1's "what is sound": it verified that the shared *key* collides, and treated that as verifying the shared-key design. Finding 1 is the half it did not check, and it is a minor, not an escalation.

## VERDICT: APPROVE WITH NITS

Nothing here is unrecoverable and nothing pages a human wrongly. Merge is fine.

**Fix first:** `launchd-health-check.py:254`. Either drop the "same two checks" claim or route fleet-health's `detect_failing_jobs` through `classify_status`, because that comment is now the reason the next person will trust the shared namespace.

Repro used for finding 1:

```python
import importlib.util, subprocess
from pathlib import Path
S = Path("<repo>/q-system/.q-system/scripts")     # PR head ec9301d
def load(n,p):
    s=importlib.util.spec_from_file_location(n,p); m=importlib.util.module_from_spec(s)
    s.loader.exec_module(m); return m
wd=load("wd",S/"launchd-health-check.py"); fh=load("fh",S/"fleet-health-daily.py")
class C:
    def __init__(s,rc,out): s.returncode,s.stdout=rc,out
CASES=[("com.cole.linkedin-session",'{\n\t"PID" = 4242;\n\t"LastExitStatus" = 15;\n}',"-\t15\tcom.cole.linkedin-session"),
       ("com.kipi.opp-scan",'{\n\t"LastExitStatus" = 768;\n}',"-\t768\tcom.kipi.opp-scan")]
for label, per_label, row in CASES:
    kind,code = wd.classify_status(0, per_label)
    wd_t = f"launchd job failing: {label} (exit {code})" if kind=="failing" else "(files nothing)"
    class Fake:
        TimeoutExpired=subprocess.TimeoutExpired
        @staticmethod
        def run(cmd,**kw):
            return C(0,"PID\tStatus\tLabel\n"+row+"\n") if list(cmd)==["launchctl","list"] else C(1,"")
    sv=fh.subprocess; fh.subprocess=Fake
    try: found=fh.detect_failing_jobs(None)
    finally: fh.subprocess=sv
    print(label, "|", wd_t, "|", found[0]["title"] if found else "(files nothing)")
```

FINDINGS:
minor|The comment justifying the shared permanent dedup namespace says fleet-health runs the same two launchd checks, but its failing-detector lacks normalize_exit and the SIGTERM carve-out, so the two disagree on real machine state|q-system/.q-system/scripts/launchd-health-check.py:254
minor|The PR description's Verification block pastes output the merged code cannot emit, so the dry-run and live-run receipts in the merge record are from the pre-round-2 revision|q-system/.q-system/scripts/launchd-health-check.py:350
END FINDINGS

PR #15, opened 2026-07-27
fix(lessons-daily): a failed run must exit non-zero so it reaches Linear (ASK-182)

assafkip:
Migrates `com.kipi.lessons-daily` onto Linear-tracked execution (ASK-182).

## The gap, observed live

The 06:00 run on 2026-07-27 published 10 lessons, then **failed to propagate to the fleet**. It logged it and Slacked it. And:

```
launchctl list com.kipi.lessons-daily -> rc=0
    "LastExitStatus" = 0;
```

`fleet-health-daily.py`'s `launchd-failing` detector keys on that status, so this job's failures lived in Slack and a 2591-line log and never reached the board. The exit code is the wire to Linear. Same contract as ASK-184.

## The four bars

| # | Bar | State |
|---|-----|-------|
| 1 | launchd, never cron | Already met. Plist loaded; `crontab -l` has no `lessons` line. Plist pins `PATH` to include `~/.local/bin`, so `claude` resolves. |
| 2 | Failures reach Linear, not only a log | **This PR.** Four silent-success paths now exit non-zero. |
| 3 | Live or in the pause ledger, never dark | Already met. Loaded, absent from both pause ledgers, and the watchdog reports no problem for it. |
| 4 | Verified by running it, not reading it | The job ran for real at 06:00 today; its output was read from the log (10 lessons published, propagation failed). Exit contract pinned by a hermetic suite. |

## What changed

`lessons-daily.sh`, reporting only. No change to what the job does.

- `propagate FAILED` -> exit 1. Previously fell off the end of the script on an `echo`, so exit 0.
- `lessons-distill.py` exits non-zero -> exit 1.
- `lessons-distill.py` emits no parseable JSON -> exit 1. A dead distiller printed nothing, every count parsed as 0, and the run logged `nothing new`. A broken night was indistinguishable from a quiet one.
- No `claude` on PATH -> exit 1, was a silent skip. The plist pins PATH, so a missing binary is a broken machine, not a normal night, and a silent skip is exactly the dark-with-a-clean-status state the bars forbid.

Held lessons stay a non-failure: the scrub gate holding client data is the gate working, and paging on it is how a channel becomes one nobody reads.

## Check

`q-system/.q-system/scripts/test/test-lessons-daily-exit.sh`, registered in `capability-manifest.json`. Fully hermetic: temp SKEL git repo, stubbed distill / kipi-update / claude / slack-notify. It cannot touch the live fleet, the real log, Slack, or git.

Before the fix:
```
[...]

PR #17, opened 2026-07-27
fix(capability-gate): the deadline is on child exit, not pipe EOF (ASK-190)

assafkip:
Closes the `validate` RED that has five reviewed PRs (#11 #12 #13 #14 #15) stuck on
*"the base branch policy prohibits the merge"*.

## What was actually wrong

Run `30274434715`, failing step **Capability gate**, one error:

```
RED: test-timeout (60s): q-system/.q-system/scripts/test/test-converge.sh
```

The suite **passes**. Every assertion prints `ok`, including the last, and it still
times out.

`capability-gate.py` ran every declared test through
`subprocess.run(capture_output=True, timeout=60)`, which waits on **pipe EOF, not
child exit**. `test-converge.sh`'s kill case backgrounds a worker whose `sleep 120`
inherits stdout, so the write end of that pipe stayed open for 120s after the suite
had already exited 0. Unpiped it passes in milliseconds, which is why it shipped
(introduced by `4b5a7af`, the ASK-113 kill-leak reproducer).

## The two fixes

**1. The gate** — `run_contained()` replaces `subprocess.run` for test artifacts:

- `start_new_session=True`, so each test owns a process group.
- The deadline is on `proc.wait(timeout=...)` (child exit), not on pipe EOF.
- stdout/stderr drain on concurrent threads, so a test writing past the 64K pipe
  buffer is not misreported as a hang.
- The group reap is **unconditional**, not timeout-only: a test that exits clean
  while leaking a child is the common case, and that orphan both holds the pipe and
  outlives the run.
- Partial output still survives a real timeout (the tail is what names the hang).

**Safety, since this touches how every declared test is run:** cleanup checks
`pgid != os.getpgid(0)` before signalling, and the pgid is read at spawn (not during
cleanup, where the reaped pid would raise and silently skip the orphan). The DoR
flags this exact shape — a cleanup trap re-raised a signal at its own pid and killed
its caller earlier that night. Cleanup can never reach the parent here.

**2. The test** — stop leaking the orphan:

- `set -m` puts the kill case in its own process group.
- The TERM goes to the **group**. `converge.sh` runs the worker as a foreground
  command and bash defers a trap until that returns, so a TERM aimed only at the
  driver pid sat unhandled for the worker's full `sleep 120`. That is what made the
  suite 122s. A real terminal or launchd stop signals the group too, so this is also
  the more faithful case.
- Reap by group, not by name: `pkill -f bin/slowworker` never matched the child,
  whose own command line is just `sleep 120`.
- New assertion: the suite fails if anything survives it.

## Evidence

**RED, gate side** — `subprocess.run(..., timeout=60)` against the HEAD suite:

```
RED: test-timeout (60s): .../.red-converge-HEAD.sh  [elapsed 60.0s]
  the suite's own last lines: ok: a run with no commits still exits 7 ... | ok: claim is held while the run is in flight ...
```

**RED, suite side** — HEAD suite piped:

```
PASS: 16/16 converge checks
PIPED elapsed: 123s
```

**RED, new reap suite against the pre-fix gate:**

```
FAIL: a test that exits 0 was reported as timed out; the runner waited on pipe EOF, not child exit
rc=1
```

**GREEN** — `bash q-system/.q-system/scripts/test/test-capability-gate-reap.sh`:

```
  ok: a test that exits 0 while leaking a child is reported as PASSED, not test-timeout
  ok: stdout is still captured when the child is reaped (a green test keeps its output)
  ok: the backgrounded child is reaped, not left running
  ok: a real hang still times out, and the deadline is not doubled by the cleanup read
  ok: partial output survives the timeout (the tail names what hung)
  ok: timeout kills the whole process group, grandchildren included
  ok: wiring: run_tests uses run_contained, own session, group kill, and parses
PASS: 7/7 capability-gate reap checks
```

**GREEN** — `test-converge.sh` piped, post-fix: `PASS: 17/17 converge checks`,
`PIPED elapsed: 2s` (was 123s).

**GREEN** — the real gate, `python3 q-system/.q-system/scripts/capability-gate.py --repo-root .`:

```
  tests: ran=66 quarantined=0 skipped-skeleton-only=0
capability-gate: GREEN
gate rc=0
```

## Wiring

`test-capability-gate-reap.sh` is declared in `capability-manifest.json` (`runner:
bash`), so it runs in CI with the other 65 — the silent-absence gate would flag it
otherwise. Its last block greps the gate for `run_contained(`, `start_new_session`,
`os.killpg` and `proc.wait(timeout=timeout)`, and **fails if `subprocess.run(cmd`
comes back**, so the unit cannot drift away from its caller or silently regress.

## Not doing (per the DoR)

The 46 Gate 1.3b containment findings (ASK-58/ASK-59), any ratchet, any merge. Those
were an unverified downstream blocker — GitHub fail-fast stopped at the capability
gate, so `validate-separation.py:589` was never reached.

**If `validate` is still red after this**, it fails at **Phase 1** with a findings
count, and the answer is the ratchet this repo already ships
(`instruction-budget-audit.py --ratchet` + `instruction-budget-baseline.json`).
Gate 1.3b is binary today (`semantic_violations == []`) with no baseline; giving it
one is a code change in `validate-separation.py`, not a merge-time judgment call.

### DIFF
--- a/q-system/.q-system/capability-manifest.json
+++ b/q-system/.q-system/capability-manifest.json
@@ -261,6 +261,10 @@
     {
       "path": "q-system/.q-system/scripts/test/test-converge.sh",
       "runner": "bash"
+    },
+    {
+      "path": "q-system/.q-system/scripts/test/test-capability-gate-reap.sh",
+      "runner": "bash"
     }
   ],
   "required_data": [],
--- a/q-system/.q-system/scripts/capability-gate.py
+++ b/q-system/.q-system/scripts/capability-gate.py
@@ -20,8 +20,10 @@ import argparse
 import datetime
 import json
 import os
+import signal
 import subprocess
 import sys
+import threading
 from pathlib import Path
 
 SCHEMA_VERSION = 1
@@ -285,6 +287,120 @@ def check_required_data(root, manifest, mode, errors):
             errors.append(f"required-data-missing: {entry.get('path')} (scope={scope})")
 
 
+class ContainedResult:
+    """What run_contained observed. `timed_out` is the deadline verdict; stdout /
+    stderr are always populated, partial on a timeout."""
+
+    def __init__(self, returncode, stdout, stderr, timed_out):
+        self.returncode = returncode
+        self.stdout = stdout
+        self.stderr = stderr
+        self.timed_out = timed_out
+
+
+def run_contained(cmd, cwd, env, timeout):
+    """Run a test artifact so that NOTHING it spawns can outlive it or hold the
+    gate past `timeout`.
+
+    Two failures this closes (ASK-190):
+
+    1. `subprocess.run(capture_output=True, timeout=...)` waits on pipe EOF, not
+       on child exit. A test that backgrounds a child which inherits stdout keeps
+       that pipe's write end open, so the gate blocks after the test is already
+       gone -- reporting a PASSING test as RED with no way to tell the two apart.
+    2. On timeout `run()` kills only the direct child. Grandchildren survive as
+       orphans, and the next thing that reads the same pipe inherits the hang.
+
+    `start_new_session=True` puts the child in its own process group, so the
+    group signal below reaches the whole subtree. The group id is the child's
+    own pid -- never this process's group -- so cleanup can never reach the gate
+    itself. (A cleanup that re-raised at its own pid killed its caller once; that
+    is the shape being avoided here.)
+    """
+    proc = subprocess.Popen(
+        cmd, cwd=cwd, env=env, text=True,
+        stdin=subprocess.DEVNULL,
+        stdout=subprocess.PIPE, stderr=subprocess.PIPE,
+        start_new_session=True,
+    )
+    # Read the pgid NOW. `start_new_session=True` makes it equal to proc.pid, but
+    # once the child is reaped its pid is gone and getpgid() would raise -- so
+    # looking it up during cleanup would silently skip the very orphan we are
+    # here to kill.
+    pgid = proc.pid
+
+    # Drain concurrently rather than reading after the wait: a test that writes
+    # more than one pipe buffer (64K) would block on write, never exit, and be
+    # reported as a timeout it did not earn.
+    chunks = {"out": [], "err": []}
+
+    def drain(stream, key):
+        try:
+            for line in iter(stream.readline, ""):
+                chunks[key].append(line)
+        except (ValueError, OSError):
+            pass  # closed under us by the reap; whatever we already read stands
+        finally:
+            try:
+                stream.close()
+            except (ValueError, OSError):
+                pass
+
+    readers = [
+        threading.Thread(target=drain, args=(proc.stdout, "out"), daemon=True),
+        threading.Thread(target=drain, args=(proc.stderr, "err"), daemon=True),
+    ]
+    for t in readers:
+        t.start()
+
+    # The deadline is on the CHILD'S EXIT, not on pipe EOF. That is the entire
+    # fix: a leaked grandchild holding the write end can no longer make a test
+    # that already exited 0 look like a hang.
+    timed_out = False
+    try:
+        proc.wait(timeout=timeout)
+    except subprocess.TimeoutExpired:
+        timed_out = True
+
+    # Unconditional, not timeout-only: a test that exits cleanly while leaking a
+    # child is the common case here, and that orphan both holds this pipe and
+    # outlives the run into whatever reads next.
+    reap_group(proc, pgid)
+    for t in readers:
+        t.join(timeout=5)
+
+    return ContainedResult(
+        None if timed_out else proc.returncode,
+        "".join(chunks["out"]), "".join(chunks["err"]), timed_out,
+    )
+
+
+def reap_group(proc, pgid):
+    """SIGKILL the child's whole process group, then the child, then reap it.
+
+    TERM-then-KILL is not worth the extra deadline: by the time this runs the
+    process either finished or is over budget, and a test that ignores TERM
+    would spend the grace period twice.
+    """
+    # The one line that keeps cleanup off the parent. `pgid` is a child pid, so
+    # this should never match -- but a cleanup that signalled its own group once
+    # killed the harness that called it, and a cheap identity check is worth
+    # more than the assumption that it cannot happen.
+    if pgid and pgid != os.getpgid(0):
+        try:
+            os.killpg(pgid, signal.SIGKILL)
+        except (ProcessLookupError, PermissionError, OSError):
+            pass  # already gone, or the child never got its own session
+    try:
+        proc.kill()
+    except (ProcessLookupError, OSError):
+        pass
+    try:
+        proc.wait(timeout=5)
+    except subprocess.TimeoutExpired:
+        pass
+
+
 def run_tests(root, manifest, mode, errors, notes):
     skeleton_only = set(manifest.get("skeleton_only", []))
     env = dict(os.environ, QROOT=str(root / "q-system"))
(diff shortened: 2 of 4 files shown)


assafkip:
CI result, posting before this turn ends rather than after.

**The capability gate step is GREEN.** `gh api .../jobs/90032015027`:

```
success  Capability gate (declared-vs-actual + full test suite)
failure  Run skeleton validation (Phase 1)
```

Gate log line: `capability-gate: GREEN`. The `test-timeout` this PR fixes is gone, and fail-fast now reaches Phase 1 for the first time.

**`validate` is still red, at a different and real error** — which is the DoR's stated second acceptable outcome and Sana's own falsifier firing as written:

```
PASS: 42 / FAIL: 2 / WARN: 12389

FAILURES:
  - Repository-derived generic targets contain no semantic instance facts (12388 findings)
  - Full skeleton sweep: zero KTLYST/hardcoded refs (2 files)
GATE FAILED. Do not proceed to Phase 2.
```

The containment count is **12388**, not the 46 assumed in ASK-58/ASK-59. That assumption was never verified because nothing had ever reached Gate 1.3b.

Captured as spillover `sp-7d840426`, out of ASK-190's scope per its **Not doing** line. Gate 1.3b is binary today (`semantic_violations == []`) with no baseline; giving it one is a code change in `validate-separation.py` modeled on `instruction-budget-audit.py --ratchet` + `instruction-budget-baseline.json`. The spillover ledger keeps `prd_runner.py gates run` red until that lands, so it cannot be quietly dropped.

Not merging, and not reaching for `--admin`.

assafkip:
Full gate run finished. Writing up.

---

# Review: PR #17 — `fix(capability-gate): the deadline is on child exit, not pipe EOF (ASK-190)`

The central diagnosis is correct and I reproduced it end to end. Base suite under the old runner:

```
BASE via subprocess.run(timeout=60): TimeoutExpired at 60.0s
  suite's own last line:   ok: claim is held while the run is in flight (kill case is live, not vacuous)
```

Head, same suite through `run_contained`:

```
test-converge.sh timed_out= False rc= 0 elapsed= 2.5
PASS: 17/17 converge checks
```

123s → 2.5s. The fix is real. Three findings survived reproduction, all minor.

---

## Findings

### 1. minor — `run_contained` has no `try/finally`, and `start_new_session=True` now puts the leak out of reach of the shell's group signal
`q-system/.q-system/scripts/capability-gate.py:361`

The docstring at line 302 says: *"Run a test artifact so that NOTHING it spawns can outlive it."* The code does not enforce that. `reap_group` is reachable only on the normal path — any exception out of `proc.wait()` skips it. And because the child is now in its own session, the group SIGINT that used to reach it no longer does. The change makes this case strictly worse than base.

Repro (group SIGINT to the gate, i.e. Ctrl-C on `kipi check`), real output:

```
[base] running subtree: 2
[base] gate rc=-2  survivors AFTER group SIGINT = 1
          62602 sleep 9991
[head] running subtree: 2
[head] gate rc=-2  survivors AFTER group SIGINT = 2
          62609 sleep 9991
          62610 sleep 9992
```

`sleep 9992` is the test's own foreground process. Base killed it (shared pgroup + `subprocess.run`'s `except: process.kill()`); head does not. Fix is `try: ... finally: reap_group(...)`.

Scoped honestly: SIGTERM is *not* a regression — I checked, base and head both leak 2 (`gate rc=-15 -> SURVIVORS: 2` on both). That hole predates this PR.

### 2. minor — the new reap suite uses machine-wide sentinels, so two concurrent gate runs both go RED with a message that misdiagnoses the cause
`q-system/.q-system/scripts/test/test-capability-gate-reap.sh:32`

`BG_SENTINEL=987` / `HANG_SENTINEL=986` are constants, and the assertions are `pgrep -f "sleep 987"` (line 92) with `pkill -f "sleep 987"` in the EXIT trap (line 35). Both are process-table-wide. Two gate runs on one machine see each other's fixtures and kill each other's fixtures.

Repro — two concurrent runs of the suite, real output:

```
===== concurrent run #0  rc=1 =====
  ok: a test that exits 0 while leaking a child is reported as PASSED, not test-timeout
  ok: stdout is still captured when the child is reaped (a green test keeps its output)
FAIL: the backgrounded child outlived the run; orphans are what re-create this bug
===== concurrent run #1  rc=1 =====
FAIL: the backgrounded child outlived the run; orphans are what re-create this bug
...test-capability-gate-reap.sh: line 34: 62942 Terminated: 15          exit 1
```

The failure text blames the code under test; the actual cause is the neighbouring run. CI is safe (one gate per isolated runner) and `fleet-capability-verify.py:60` iterates instances sequentially, so nothing in the repo triggers this automatically — it needs two humans/agents running `kipi check` at once, which this fleet does do. Bounded and self-clearing on re-run. `$$`-derived sentinels close it.

### 3. minor — the new orphan assertion in `test-converge.sh` cannot go red for the regression it guards
`q-system/.q-system/scripts/test/test-converge.sh:260`

`LEAKED="$( { pgrep -g "$CONV_PID" ...` runs on the line immediately after `kill -KILL -"$CONV_PID"` (line 251). It asserts that a group is empty right after SIGKILLing that group. If the containment is removed, `CONV_PID` is no longer a pgid, `pgrep -g` matches nothing, and the assertion passes vacuously.

Mutation repro — I reverted the whole containment change (dropped `set -m`/`set +m`, restored `pkill -f 'bin/slowworker'`, dropped the EXIT-trap group kill) and ran it:

```
MUTANT suite rc = 0
['ok: kill case leaves no orphan behind (nothing outlives the suite)', 'PASS: 17/17 converge checks']
real orphans alive AFTER the suite exited: []
```

Green, 17/17, on fully-reverted code. (The mutant also blew past my 120s harness timeout, which is the *real* regression signal — and it is the capability gate's own 60s deadline, not this assertion, that catches it. Coverage isn't zero; this particular assertion just contributes nothing.) A duration bound on the kill case would actually gate it.

---

## What is sound

- **The >64K pipe-buffer claim** in the drain-thread comment (line 344). Not covered by any test in the PR, so I checked it: `timed_out=False rc=0 bytes=1388894 elapsed=0.1s`, last line intact. The claim holds.
- **Cleanup reaching the parent.** `pgid` is captured at spawn as `proc.pid` and checked against `os.getpgid(0)` before `killpg`. I could not construct a path where the gate signals its own group. The stated safety story is real, not decorative.
- **Recycled-pid kill via `proc.kill()`.** CPython's `send_signal` polls first and no-ops once `returncode` is set, so the post-reap `proc.kill()` cannot hit a reused pid. Attacked, dropped.
- **pgid recycling on `os.killpg` / `kill -KILL -$CONV_PID`.** Theoretically live (nothing guards the pgid the way `proc.kill()` is guarded), but the window is microseconds and I could not build a reproducer. Dropped per the standing rule.
- **The timeout path still works.** Real hang still reports `timed_out=True`, still keeps partial output, still kills grandchildren — `PASS: 7/7 capability-gate reap checks` in 3.3s.
- **The unmodified paired test.** `test_capability_gate.py` was not touched by this PR; it still passes, including its four runner-contract cases (`runner: timeout RED with partial output tail`, `runner: cwd=repo root and QROOT set`). `ALL PASS`.
- **`stdin=subprocess.DEVNULL`** is a new behaviour, and it is an improvement — a test that reads stdin used to inherit the operator's terminal.
- **Rollback:** one commit, four files, no state written anywhere. Trivially revertable.

One caveat on evidence: my full-gate run returned `capability-gate: RED (2)`, but both REDs are `test-updater-issue-sequence.py` failing with `fatal: not a git repository` — my workspace is a `git archive` extraction with no `.git`, and that test probes real commit ancestry. That is my environment, not the PR. I could not confirm a full end-to-end green from here; CI can.

---

## VERDICT: APPROVE WITH NITS

Three minor findings, no major, no blocker. The fix is correct, the diagnosis is right, and the five wedged PRs should unwedge. None of these gate the merge.

**Fix first:** the missing `try/finally` in `run_contained` (finding 1). It is two lines, it is the one place where the code contradicts its own docstring, and it is the only finding where this PR makes something worse than base.

FINDINGS:
minor|run_contained has no try/finally, so an interrupt skips reap_group and the now-detached test subtree survives — worse than base|q-system/.q-system/scripts/capability-gate.py:361
minor|Hardcoded machine-wide sleep sentinels plus pgrep/pkill make two concurrent gate runs mutually fail with a message that misdiagnoses the cause|q-system/.q-system/scripts/test/test-capability-gate-reap.sh:32
minor|The new orphan assertion runs immediately after SIGKILLing the group, so it stays green with the entire containment change reverted|q-system/.q-system/scripts/test/test-converge.sh:260
END FINDINGS

PR #18, opened 2026-07-27
fix(validate): Gate 1.3b gates on real findings; two plists stop hardcoding a home dir (ASK-191)

assafkip:
Closes the `validate` RED on `main` that is holding PRs #11 #12 #13 #14 #15 #16
behind *"the base branch policy prohibits the merge"*.

Carries the ASK-190 fix (Step 0: `git merge origin/sana/ask-190`, clean
fast-forward) so CI reaches step 7 where this issue's work lives. **PR #17
should close as included.**

## Result

```
bash q-system/output/run-validate.sh
  PASS: 45   FAIL: 0   WARN: 1
  ALL CHECKS PASSED. Phase 1 gate is GREEN.
  VALIDATE_EXIT=0
```

Gate 1.3b went from `FAIL ... (12388 findings)` to:

```
advisory (non-gating): cited_source=6, client_identity=1, pricing_mention=6,
                       source_identity=2, sourced_interaction=2,
                       unclassified_populated_record=6049
PASS Repository-derived generic targets contain no semantic instance facts
     (0 gating, 6066 advisory)
```

No baseline, no ratchet, no blanket suppression. Every one of the 47 classified
findings is dispositioned.

## Part A1 — the gate stopped counting its own noise

`unclassified_populated_record` was 12,341 of 12,388 findings, including 418
lines of one MCP server source file. It is now advisory: counted every run,
listed under `--verbose`, non-gating. Not deleted, because the day that number
collapses is the day the classifier learned the schema.

Pinned by `test-gate-13b-scope.py`: a fixture of 50 unclassified findings on a
**propagated** path must produce 0 gating, **and** one `client_identity` added
to that same set must still gate. Without the second half a partition that
discarded everything would pass.

## Part A2 — scan scope now matches propagation

`kipi update` never copies `q-system/{my-project,canonical,memory,output,
research,.q-system/data,.q-system/agent-pipeline/bus}`. Gate 1.3b was scanning
directories the Full skeleton sweep *in the same file* already excludes on
purpose.

The filter lives in `validate-separation.py`, **not** in `containment-targets.py`
— `propagation-leak-gate.py` loads that same enumerator, and narrowing it there
would have silently shrunk a second, unrelated gate.

Drift protection is a test, not a comment: `test-gate-13b-scope.py` parses
`INSTANCE_OWNED_SUBTREES` out of `kipi-update.sh` and fails if the two lists
disagree. It fired for real during this work.

## Part A3 — the remaining 34, each dispositioned

**Classifier tightenings** (each measured over the propagated set first):

- **A declaration is not an assertion.** The bare `label: value` grammar no
  longer applies in code (`.py/.js/.ts/...`) or tabular data (`.csv`). It was
  reading Python type annotations (`source: dict[str, Any],`), docstring
  parameters and CSV prose cells as canonical records — 15 of the 34, not one a
  fact. Markdown grammars still apply everywhere, so `**Client:** Acme` inside
  a docstring is still caught. `.yaml` is deliberately **not** exempt (YAML is a
  real roster format); `prospect: string` is handled by an exact allowlist of
  schema primitive type names, which a company name cannot pass through.
- **A cited document is not an identity.** A `Source:` line carrying a locator
  (URL, markdown link, arXiv, DOI) is `cited_source`, advisory. A bare
  `Source: Acme Corp` has no locator and still gates.
- **Currency dominance, not currency presence.** Strip the figures and see
  whether a sentence is left. `$5,000` → 0 words, `Maren quoted $5,000 on July
  24, 2026` → 4, all prices. The Apify and Gamma operating-cost lines → 13 and
  14. Bound at 6, in the middle of a measured valley.
- `[Bracketed Title Case]` joins `{{HANDLEBARS}}` as a placeholder, and a value
  that is entirely a parenthetical is a template prompt.

**File-level dispositions:**

- `q-system/research/` was **not** non-propagated, contrary to the issue text —
  the q-system rsync copies the whole tree minus the owned subtrees, so four
  kipi-system-specific notes were shipping to every instance. Fixed at the
  source (`research` added to `INSTANCE_OWNED_SUBTREES`), not by excluding the
  path from the gate that noticed.
- `lessons-distill.py:204` wrote `source: <absolute path>` into every held
  lesson. Fixed in the **writer** (`held_source_link`) so it cannot regenerate.
- Three markdown tables had a label cell colliding with a field name
  (`| source |`, `| Pricing |`, `| Proof gaps |`). Backticked or reworded in
  place — the detector keeps its strength.
- `research-receipts.md` gained the subreddit locators it was already citing.

## Part B — two plists, and the substituter that never existed

The DoR asked me to *"confirm whatever materializes the placeholders at install
time also covers these two."* **Nothing did.** `__KIPI_REPO__` / `__HOME__` in
`com.kipi.openloops-heartbeat.plist` were never substituted by anything in the
repo — copying that plist into `~/Library/LaunchAgents` produced a job that
tried to exec literal `__KIPI_REPO__`. Applying the convention alone would have
traded a hardcoded path for a broken job.

So this ships `install-plist.sh` as the single renderer for all three
templates, with a fail-loud guard on any surviving placeholder, and
`test-install-plist.sh` pinning it — including a negative self-test that re-runs
the guard against a deliberately neutered renderer and fails if a no-op render
still exits 0.

## Corrections to the issue text

Two claims in the DoR turned out to be wrong, both verified before disagreeing:

1. `plugins/kipi-core/kipi-mcp/sources/linkedin-prospects.yaml:14` is
   `prospect: string` inside `output.schema` — a type declaration, not "a
   prospect list shipping to every instance".
2. `q-system/research/` is listed as non-propagated. It propagates. Excluding it
   without the `kipi-update.sh` change would have hidden a real fan-out instead
   of stopping it.

## Two detectors I broke and put back

The A3 tightening turned 6 existing pinning tests red. Both causes were real
losses of strength, so the fixtures stood and the code changed:

1. `pricing_mention` as "currency without a pricing label" read
   `**Package:** $5,000` as advisory — precisely the leak this gate exists for.
2. The bracket-placeholder form ate the **label** of a markdown link, so
   `**Client:** [Oriole Systems](https://oriole.example)` collapsed to a bare
   parenthetical and vanished entirely. Fixed with a `(?!\()` lookahead.

Found by the repo's own reach probe and boundary fixtures.

## Verification

```
bash q-system/output/run-validate.sh
  -> PASS: 45  FAIL: 0  VALIDATE_EXIT=0

python3 q-system/.q-system/scripts/capability-gate.py
  -> capability-gate: GREEN (ran=68, up from 66)

python3 -m pytest plugins/prd-os/tests/ -q
  -> 318 passed, 1 skipped

python3 -m pytest q-system/.q-system/tests/separation/ -q
  -> 163 passed, 4 failed   (see below)

bash q-system/.q-system/scripts/test/test-install-plist.sh    -> PASS (12)
python3 q-system/.q-system/scripts/test/test-gate-13b-scope.py -> PASS (9)
```

**The 4 separation failures are pre-existing and unrelated.** Verified red at
the base commit `1a5d0c3` in a detached worktree before touching anything. They
are not in the capability manifest, so CI is unaffected. Captured as spillover
`sp-3b7bae7d` rather than fixed here — ASK-191's scope is Gate 1.3 and the two
plists. Root cause for two of them: they grep for a literal `--exclude=/canonical/`
on the rsync command line, but `kipi-update.sh` now builds excludes via the
`$(rsync_owned_excludes)` helper.

## Note on sequencing

PR #12 (ASK-183) touches `com.kipi.linear-dor`, one of the two Part B files.
This PR rewrites that file wholesale, so expect a conflict and take this
version plus whatever schedule/args change #12 makes.

### DIFF
--- a/kipi-update.sh
+++ b/kipi-update.sh
@@ -51,11 +51,22 @@ fi
 # the staging list) and drifted the moment a sixth entry was added -- the sync
 # wrote files it then refused to stage, leaving the instance dirty with no
 # commit. One list, four consumers.
+# The single answer to "what does kipi update NEVER copy into an instance?".
+# validate-separation.py's NON_PROPAGATED_PREFIXES mirrors this list and
+# test-gate-13b-scope.py parses this array to block drift between them.
+#
+# `research` added 2026-07-27 (ASK-191): q-system/research/ held four notes
+# written BY and ABOUT this instance (Claude Code workflow learnings, an RCA
+# resource list, a SkillOpt paper + PDF) and shipped all four to every
+# instance in the fleet, which has no use for them. That is a placement bug,
+# and the honest fix is to call the directory instance-owned rather than to
+# exclude the path from the gate that noticed.
 INSTANCE_OWNED_SUBTREES=(
   my-project
   canonical
   memory
   output
+  research
   .q-system/data
   .q-system/agent-pipeline/bus
 )
--- a/lesson-candidates/held-753e1ad7227a25b0.md
+++ b/lesson-candidates/held-753e1ad7227a25b0.md
@@ -1,7 +1,7 @@
 # HELD lesson (not published)
 
 reason: LLM semantic check flagged a residual real entity
-source: /Users/assafkipnis/projects/cole-gtm/q-system/output/rca/rca-reddit-emit-drafter-abort-2026-07-22.md
+source: [rca-reddit-emit-drafter-abort-2026-07-22.md](../cole-gtm/q-system/output/rca/rca-reddit-emit-drafter-abort-2026-07-22.md)
 
 proposed title: Isolate per-item failures in every batch stage
 
--- a/plugins/kipi-core/.claude-plugin/plugin.json
+++ b/plugins/kipi-core/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "kipi-core",
-  "version": "1.5.13",
-  "description": "Core founder OS — voice, research, AUDHD, LinkedIn/brand, RCA skills + kipi-mcp server",
+  "version": "1.5.14",
+  "description": "Core founder OS \u2014 voice, research, AUDHD, LinkedIn/brand, RCA skills + kipi-mcp server",
   "author": {
     "name": "Assaf Kipnis",
     "email": "[REDACTED]"
--- a/plugins/kipi-core/skills/headline-engineering/references/research-receipts.md
+++ b/plugins/kipi-core/skills/headline-engineering/references/research-receipts.md
@@ -20,7 +20,12 @@ Caveat: these stats come from social-media-marketing aggregators with skin in th
 
 ## Reddit top text headlines (last 30 days, AI subs)
 
-Source: direct Reddit MCP pulls on 2026-05-20, top-of-month sorted across r/ChatGPT, r/ClaudeAI, r/OpenAI, r/singularity, r/LocalLLaMA, r/ArtificialIntelligence.
+Source: direct Reddit MCP pulls on 2026-05-20, top-of-month sorted across [r/ChatGPT](https://www.reddit.com/r/ChatGPT/),
+[r/ClaudeAI](https://www.reddit.com/r/ClaudeAI/),
+[r/OpenAI](https://www.reddit.com/r/OpenAI/),
+[r/singularity](https://www.reddit.com/r/singularity/),
+[r/LocalLLaMA](https://www.reddit.com/r/LocalLLaMA/),
+[r/ArtificialIntelligence](https://www.reddit.com/r/ArtificialIntelligence/).
 
 Top text-driven headlines and their lengths:
 
--- a/plugins/kipi-design/.claude-plugin/plugin.json
+++ b/plugins/kipi-design/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "kipi-design",
-  "version": "1.2.4",
+  "version": "1.2.5",
   "description": "Brand, UI/UX, and design skills",
   "author": {
     "name": "Assaf Kipnis",
--- a/plugins/kipi-design/skills/brand/references/asset-organization.md
+++ b/plugins/kipi-design/skills/brand/references/asset-organization.md
@@ -110,11 +110,11 @@ infographic_evergreen_pricing-comparison_20251209.png
 ### Standard Tags
 | Category | Values |
 |----------|--------|
-| status | draft, review, approved, archived |
-| platform | instagram, twitter, linkedin, facebook, youtube, email, web |
-| content-type | promotional, educational, brand, product, testimonial |
-| format | 1x1, 4x5, 9x16, 16x9, story, reel, banner |
-| source | imagen-4, veo-3, user-upload, canva, figma |
+| `status` | draft, review, approved, archived |
+| `platform` | instagram, twitter, linkedin, facebook, youtube, email, web |
+| `content-type` | promotional, educational, brand, product, testimonial |
+| `format` | 1x1, 4x5, 9x16, 16x9, story, reel, banner |
+| `source` | imagen-4, veo-3, user-upload, canva, figma |
 
 ### Tag Usage
 - Each asset should have: status + platform + content-type
--- a/plugins/kipi-design/skills/design/references/slides-copywriting-formulas.md
+++ b/plugins/kipi-design/skills/design/references/slides-copywriting-formulas.md
@@ -40,7 +40,7 @@
 | Features | FAB | confidence |
 | Traction | Proof Stack | trust |
 | Social Proof | Testimonial | trust |
-| Pricing | Value Stack | confidence |
+| Pricing slide | Value Stack | confidence |
 | CTA | AIDA, Urgency | urgency |
 
 ## Headline Patterns
--- a/q-system/.q-system/capability-manifest.json
+++ b/q-system/.q-system/capability-manifest.json
@@ -265,6 +265,14 @@
     {
       "path": "q-system/.q-system/scripts/test/test-capability-gate-reap.sh",
       "runner": "bash"
+    },
+    {
+      "path": "q-system/.q-system/scripts/test/test-install-plist.sh",
+      "runner": "bash"
+    },
+    {
+      "path": "q-system/.q-system/scripts/test/test-gate-13b-scope.py",
+      "runner": "python3"
     }
   ],
   "required_data": [],
--- a/q-system/.q-system/capability-manifest.json
+++ b/q-system/.q-system/capability-manifest.json
@@ -261,6 +261,10 @@
     {
       "path": "q-system/.q-system/scripts/test/test-converge.sh",
       "runner": "bash"
+    },
+    {
+      "path": "q-system/.q-system/scripts/test/test-capability-gate-reap.sh",
+      "runner": "bash"
     }
   ],
   "required_data": [],
--- a/q-system/.q-system/scripts/capability-gate.py
+++ b/q-system/.q-system/scripts/capability-gate.py
@@ -20,8 +20,10 @@ import argparse
 import datetime
 import json
 import os
+import signal
 import subprocess
 import sys
+import threading
 from pathlib import Path
 
 SCHEMA_VERSION = 1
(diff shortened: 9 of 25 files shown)


assafkip:
# Review: PR #18 `fix(validate): restore the two detectors the A3 tightening broke (ASK-191)`

Read cold from the PR head worktree (`93bdcc4`). Ran every new test, ran the old and new detector over the identical 1073-file target tree, and mutation-tested the new guards. Five findings survived reproduction. None gate.

---

## Findings

### minor 1 — `held_source_link` drops the source instance on the fleet's nested layout, and its docstring claims the opposite
`q-system/.q-system/scripts/lessons-distill.py:163`

`fleet_root = Path(__file__).resolve().parents[4]` assumes the instance repo sits directly under the fleet root. The repo's own cluster tombstone says five instances are nested (`cole-gtm/projects/strategy`, `consulting/projects/lawyer`, `intel/projects/deliverables`, …). For those, `parents[4]` is the persona's `projects/` dir, so a source doc from a sibling top-level instance misses `relative_to`, falls to `Path(source.name)`, and the link loses which instance it came from. The docstring says "the link resolves for any checkout layout rather than only the author's."

```
$ python3 - <<'PY'   # reimplements held_source_link verbatim, varies only __file__
FLAT   instance -> ('[rca-x-2026-07-22.md](../intel/q-system/output/rca/rca-x-2026-07-22.md)', '/Users/assafkipnis/projects')
NESTED instance -> ('[rca-x-2026-07-22.md](../rca-x-2026-07-22.md)', '/Users/assafkipnis/projects/cole-gtm/projects')
```

Runs unattended in auto-learn, and the ledger marks the hash processed, so the held file is written once and never regenerated. Kept at minor: nobody acts on it at 3am, and the filename still identifies the doc.

### minor 2 — the new "no orphan" assertion in `test-converge.sh` cannot fail
`q-system/.q-system/scripts/test/test-converge.sh:260`

`pgrep -g "$CONV_PID"` is anchored on a group id that only exists if `set -m` took effect. If it does not, `pgrep -g` finds no such group, `LEAKED` is `0`, and the check passes while orphans are alive. It cannot tell "group empty because everything died" from "group never existed" — which is the exact failure mode it was added to catch.

```
--- CASE A: with set -m (as the PR writes it) ---
pgid of child == its pid?  child pid=89819  actual pgid=89819
LEAKED=0

--- CASE B: identical assertion, but WITHOUT set -m (the mutation) ---
child pid=89833  actual pgid=89816  (shell pgid=89816)
  (group kill FAILED -- pgid is not the child pid)
LEAKED2=0   <-- assertion still passes
orphans actually still alive:
89835 sleep 92
89836 sleep 92
```

### minor 3 — a deal record written as one sentence is silent on a propagated path
`validate-separation.py:499`

`_states_a_price` gates only when the residue is ≤ 6 words. The commit message's own example (`Oriole Systems signed for $45,000 today.`, 5 words) gates; one clause longer and the same client-plus-price line drops to `pricing_mention`, which is advisory.

```
GATES  | ['pricing']         | - **Package:** $5,000
SILENT | ['pricing_mention'] | - **Deal:** Oriole Systems signed a $45,000 annual contract on July 24
SILENT | ['pricing_mention'] | - **Notes:** Oriole Systems agreed to $45,000 for the pilot this quarter
```

Constructed input. No producer in this repo emits that shape today (I checked all 6 real `pricing_mention` hits: Apify and Gamma operating costs). Severity dropped a level for that.

### minor 4 — two leak shapes go fully invisible, not even advisory
`validate-separation.py:275` and `validate-separation.py:~380` (`_record_patterns_for`)

`[Title Case]` now counts as a placeholder, and the bare `label: value` grammar is dropped in `.py/.js/.ts/.csv`. Both are disclosed in comments; both produce zero findings, so they are not counted and do not appear under `--verbose`.

```
SILENT | no-finding | - **Client:** [Oriole Systems]                 (plugins/.../notes.md)
SILENT | no-finding | # Client: Oriole Systems                       (plugins/prd-os/scripts/runner.py)
SILENT | no-finding | # Source: Maren Wu 2026-07-12                  (plugins/prd-os/scripts/runner.py)
```

Constructed input; the 27 real findings the exemption removed are all type annotations and template blanks (see below). Severity dropped a level.

### minor 5 — the test's docstring contradicts the test body about `research/`
`q-system/.q-system/scripts/test/test-gate-13b-scope.py:11`

```
--- docstring claim ---
   q-system/research/ is asserted GATING on purpose: the q-system rsync copies
   the whole tree minus INSTANCE_OWNED_SUBTREES, so research/ propagates.
--- what the code actually asserts ---
    non_propagated = [
        ...
        finding("q-system/research/cc-workflow-learnings-2026-06-02.md", "source_identity", 4),
    ]
    expect("classified findings on non-propagated paths do not gate", gating == [])

$ python3 .../test-gate-13b-scope.py
  PASS classified findings on non-propagated paths do not gate
```

Stale from commit `1adab8b`, reversed by `93bdcc4`. Same block's label says "all 6" for a 7-item list.

### nit — five agent scratch files committed to a public repo, three hardcoding the author's worktree
`q-system/output/ask191-triage.py:4`, `ask191-lines.py:5`, `read-ask191.py:3`

```
93bdcc4:q-system/output/ask191-lines.py:5:ROOT  = "/Users/assafkipnis/.config/kipi/worktrees/ask-191"
93bdcc4:q-system/output/ask191-triage.py:4:ROOT = "/Users/assafkipnis/.config/kipi/worktrees/ask-191"
```

No secrets in them. Worth naming because this PR's other half exists to remove `/Users/assafkipnis` from committed files, and `q-system/output/` is exempt from both gates that would have caught it.

---

## What is sound

Eight attacks that the code survived:

- **"The path exclusion is dead code and the fixtures are invented."** It is live. `semantic_separation_violations` emits repo-relative paths; 33 of 1073 real targets are non-propagated and 5 real classified findings land there. The fixtures use exactly the producer's dict shape and real repo-relative paths.
- **"The tightening silenced a real leak."** Ran old and new over the identical tree and diffed. 27 classified findings dropped. I read all 27: Python type annotations (`source: dict[str, Any],`), docstring params, `httpx.AsyncClient`, one CSV prose cell repeated 4x, and `- **Source:** [Person] - [Date]` template blanks. Not one was a fact.
- **"The new tests could not fail."** `test-gate-13b-scope.py`'s mixed-set negative genuinely catches a partition that drops everything. `test-install-plist.sh`'s neutered-renderer negative really exits 1, with a placeholder-free control proving the rejection came from the placeholder, not the sed edit.
- **"The reap change doubles deadlines or eats output."** `test-capability-gate-reap.sh` 7/7: the 15s-deadline fixture returns in under 10s with stdout intact, the 3s hang is not doubled, and both sentinel orphans are gone.
- **"The '163 passed, 4 failed, pre-existing' claim is author spin."** True as written. Same 4 tests fail at the ASK-190 base worktree.
- **"`install-plist.sh` tests touch real launchd."** They do not. `--render-only` throughout; all three templates render placeholder-free and pass `plutil -lint`. `~/Library/LaunchAgents` untouched.
- **"The drift test doesn't really parse `kipi-update.sh`."** It does, reports 7 subtrees, and raises `SystemExit` if the array is unfindable.
- **`validate-separation.py 1`** exits 0 at the PR head.

**Dropped for lack of a reproducer, as the process requires:** the PID-reuse race in `reap_group` (unconditional `os.killpg` on a pid `proc.wait()` already reaped) — the mechanism is real but I could not force pid reuse deterministically. And `install-plist.sh` overwriting a working `~/Library/LaunchAgents` entry before `assert_rendered`/`plutil` validate it — I would not run that path against the founder's real launchd, and `--render-only` does not exercise it.

---

## VERDICT: APPROVE WITH NITS

No blocker, no major. The detector work is measured against the real corpus rather than asserted, the negative tests are genuine, and the reap fix is proven against fixture processes.

**Fix first:** `lessons-distill.py:163`. Resolve the fleet root by walking up to the directory containing `instance-registry.json` instead of `parents[4]`, so the nested instances get a real link instead of a bare filename.

FINDINGS:
minor|held_source_link's parents[4] fleet-root assumption drops the source instance name for nested-layout instances, contradicting its own docstring|q-system/.q-system/scripts/lessons-distill.py:163
minor|The new no-orphan assertion passes vacuously when set -m is not in effect, since pgrep -g finds nothing whether the group is empty or never existed|q-system/.q-system/scripts/test/test-converge.sh:260
minor|A client name plus deal size written as one sentence classifies as advisory pricing_mention and stops gating once the residue exceeds 6 words|validate-separation.py:499
minor|Bracketed Title Case values and bare label lines in .py/.js/.csv produce no finding at all, so those leak shapes are invisible even under --verbose|validate-separation.py:275
minor|test-gate-13b-scope.py's docstring says research/ is asserted GATING while the test body asserts it is non-propagated|q-system/.q-system/scripts/test/test-gate-13b-scope.py:11
nit|Five agent scratch files are committed to a public repo, three hardcoding the author's worktree path, in a PR whose other half removes hardcoded homes|q-system/output/ask191-triage.py:4
END FINDINGS

PR #19, opened 2026-07-27
fix(fleet-health): the cron detector publishes a line reference, not the line (ASK-204)

assafkip:
Replaces #11 (ASK-150), closed unmerged after nine review rounds. Resolves spillover `sp-638006cc`.

Branched from `origin/main` at `72c782d` (fetched first, per the issue). The ask-150 branch was **read, not merged** — its detection logic and test corpora are salvaged; its conflicting `fleet-health-daily.py` is not.

## The defect

The detector's behaviour was correct. Its output channel was not.

It took arbitrary shell text from `crontab -l`, ran best-effort regex redaction over it, and published the result into a **permanent** Linear issue. That is a denylist against unbounded syntax, and it leaked in every round anyone looked: a Slack webhook path, a Bearer token, a `bash -lc 'KEY=... claude'` credential (one of the detector's own headline supported shapes), a backtick word start, `lin_api_` / `ntn_` token shapes. Nine rounds, nine bypasses. Each fix was correct. The architecture was the defect.

## The fix

**A finding carries a reference, never the text.** `- line 2` instead of the line's contents. The operator opens their own crontab.

Deleted, not tightened: `_redact_secrets`, `_ASSIGNMENT_RE`, `_SECRET_PATTERNS`. There is nothing left to redact and no bypass left to find. `file_findings`' unreachable-Linear path gets the same treatment — the exception **type** reaches the outcome and the Slack line; the message stays on stderr in the run log (a GraphQL error array echoes the request, and this fleet's requests carry an `Authorization` header).

**Detection is not weakened.** The parser still reads each line in full — quoting, comments, `-c` strings, command substitution, wrapper/flag/operand tracking. What changed is what the finding *carries*.

One behaviour change, and it makes detection **stronger**: `_split_substitutions` masks a substitution span before the segment walk lexes. Previously a substitution containing a space split an assignment prefix across two whitespace tokens, so `` VAR=`echo secret` claude -p x `` — a real invocation — scored `` secret` `` as the command and answered False. Silent false negative, surfaced by the reproducer this issue's acceptance criterion demands.

## Also fixed, carried from #11's review rounds

- **major** — `_refresh_one` no longer replaces the whole `description`. It renders only the region above `<!-- kipi-managed-end -->` and splices the operator's tail back, with a graded fallback (sentinel → v1 trailer → preserve the whole body) for the pre-sentinel issues already on the board. Pinned by a test asserting a note survives **two consecutive** rewrites without duplicating.
- **minor** — a ledger key whose issue Linear cannot locate is now **re-filed** instead of counted and pinged every morning forever. `read_ledger` is last-wins over an append-only file, so the new record replaces the dangling `linear_id` and the next run resolves the key normally. One `relisted` line, once.

## ASK-181's contract is preserved

`file_findings` keeps its `filer` parameter and the `skipped_no_key` bucket that `launchd-health-check.py` reads by name; `outcome_line` tolerates the 3-key outcome that script builds by hand. Both are pinned by tests here, and `test_launchd_health_check.py` still passes.

## Evidence

The reproducer takes the target file as an argument, so the same script proves both sides.

```
$ python3 q-system/output/ask204-repro.py <ask-150 head>
what the finding publishes:
    - `0 3 * * * bash -lc '`LINEAR_API_KEY=lin_api_backtickleak claude -p sweep`'`
    - `0 4 * * * claude -p 'sweep' --header lin_api_bareleak`
    - `0 5 * * * claude -p 'sweep' --header ntn_barenotionleak`

RED: 3 secret(s) reach the Linear issue body verbatim
EXIT=1

$ python3 q-system/output/ask204-repro.py <this head>
GREEN: zero characters of any source line reach the body
EXIT=0
```

```
PASS: fleet-health-daily contract holds          # exit 0

$ python3 q-system/.q-system/scripts/test_launchd_health_check.py
PASS: all launchd-health-check logic checks green # exit 0

$ python3 q-system/.q-system/scripts/capability-gate.py --repo-root .
  tests: ran=72 quarantined=0 skipped-skeleton-only=0
capability-gate: GREEN                            # exit 0
```

Live smoke against the real crontab, no writes:

```
fleet-health 2026-07-27T18:42:37Z
  unwired-untracked: 0
  launchd-dark: 0
  launchd-failing: 0
  schedule-duplicate: 0
  cron-shells-claude: 0
  open-spillover: 0
  filed=0 already-tracked=0 unfiled=0 rewritten=0 reopened=0 relisted=0 errors=0
```

## One place I did not follow the issue

ASK-204 lists "the quote-state tracking in `_substitutions`" among what the rewrite deletes. It stays. That tracking is **detection** work, not redaction work: single quotes suppress substitution and double quotes do not, so `` echo 'run `claude -p x` now' `` is prose and ``echo "`claude -p x`"`` is an invocation. Dropping it would file a permanent false-positive issue on the first — which the same issue forbids two paragraphs later ("Do not weaken detection to simplify output"). Both cases are pinned by tests. Flagging rather than silently skipping.

## Blast radius

`q-system/.q-system/scripts/` syncs to every instance via `kipi update`, so this detector runs fleet-wide. `linear-sync.py` gains `fetch_issue`, `reopen_state_id`, `ISSUE_UPDATE`, and richer `fetch_remote_state` records — all additive; no existing caller's shape changed.

### DIFF
--- a/q-system/.q-system/scripts/fleet-health-daily.py
+++ b/q-system/.q-system/scripts/fleet-health-daily.py
@@ -722,7 +722,12 @@ def _shells_claude(command: str, _depth: int = 0) -> bool:
     if _depth > 5:
         return False
     command = _strip_comment(command)
-    for tokens in _shell_segments(command):
+    # Substitutions come OUT before the segment walk lexes, and go back in as one
+    # argument-shaped token each. Their contents are walked separately below, at
+    # the same depth cap. Leaving them inline let a substitution's internal space
+    # split an assignment prefix across two tokens (see _split_substitutions).
+    masked, inner_commands = _split_substitutions(command)
+    for tokens in _shell_segments(masked):
         command_index = _command_index(tokens)
         command_token = tokens[command_index] if command_index >= 0 else ""
         if _is_claude_token(command_token):
@@ -732,9 +737,9 @@ def _shells_claude(command: str, _depth: int = 0) -> bool:
             if inner and _shells_claude(inner, _depth + 1):
                 return True
     # A substitution is a command position of its own, and the segment walk above
-    # cannot reach one that a quote has kept whole. Same recursion, same depth cap
-    # as the `-c` string: what runs inside `$( )` or backticks is a command, not text.
-    for inner in _substitutions(command):
+    # cannot reach one: it saw a placeholder. Same recursion, same depth cap as
+    # the `-c` string: what runs inside `$( )` or backticks is a command, not text.
+    for inner in inner_commands:
         if _shells_claude(inner, _depth + 1):
             return True
     return False
--- a/q-system/.q-system/scripts/fleet-health-daily.py
+++ b/q-system/.q-system/scripts/fleet-health-daily.py
@@ -119,17 +119,62 @@ def _is_loaded(label: str) -> bool:
         return True  # cannot tell -> do not cry wolf
 
 
-def detect_dark_jobs(_ctx) -> list:
-    """A plist on disk, not loaded, and NOT in the paused ledger = silent death."""
-    paused = _paused_labels()
-    out = []
-    for label in _launchd_labels():
-        if not label.startswith(("com.kipi.", "com.cole.", "com.ask.", "com.assaf.",
-                                 "com.claudedaddy.", "com.purespectrum.", "com.personal.")):
-            continue
-        if label in paused or _is_loaded(label):
-            continue
-        out.append({
+# ---------------------------------------------------------------------------
+# ONE rendering per launchd kipi-key, shared by BOTH filers
+# ---------------------------------------------------------------------------
+# Two scripts file against these keys: this one at 08:15 and
+# launchd-health-check.py at 09:30 and 21:30 (ASK-181 made them share the key on
+# purpose -- a private namespace on either side files a SECOND permanent issue for
+# the same dead job). `finding_hash` covers title + body, so sharing a key while
+# rendering different text means every alternating run sees a stale hash and
+# rewrites the issue: 13 Linear mutations a week and a "1 updated ... nothing to
+# do now" Slack line every morning on a fleet where nothing moved. That is the
+# checker training the operator to skip the line the real alert has to compete
+# with (PR #19 round-3 review, major). One key, one renderer -- both callers come
+# through here so the next edit cannot land on one side only.
+LAUNCHD_FINDING_IDS = ("launchd-failing", "launchd-dark")
+
+
+def launchd_exit_detail(list_status) -> str:
+    """`exit N` for a `launchctl list` STATUS column value.
+
+    The two readers see one machine state in TWO encodings: `launchctl list`
+    prints a SIGNED status while `launchctl list <label>` prints the RAW wait
+    status that launchd-health-check.py's `normalize_exit` decodes. Measured on
+    this machine 2026-07-27 across 210 non-zero rows: a SIGKILLed job reads `-9`
+    in the column and `"LastExitStatus" = 9`; `com.apple.BiomeAgent` reads `5`
+    and `1280`. `abs()` lands both encodings in the same space, so the detail
+    this renderer interpolates does not depend on WHICH script looked.
+
+    A column value that is not a number keeps its text rather than becoming a
+    fabricated `exit 0` -- an unparsable status is a thing to show, not to round.
+    """
+    try:
+        return f"exit {abs(int(list_status))}"
+    except (TypeError, ValueError):
+        return f"exit {list_status}"
+
+
+def launchd_finding(detector_id: str, label: str, detail: str = "") -> dict:
+    """The finding dict for one launchd problem: {subject, title, body}.
+
+    The ONLY place either filer renders these. `detail` is a `launchd_exit_detail`
+    string and is ignored for `launchd-dark`, which has no exit status.
+    """
+    if detector_id == "launchd-failing":
+        return {
+            "subject": label,
+            "title": f"launchd job failing: {label} ({detail})",
+            "body": (
+                f"`{label}` is loaded but its last run exited non-zero (**{detail}**).\n\n"
+                "## Action\nRead its `StandardErrorPath`, fix the cause, and confirm a clean "
+                "run. If the failure is environmental (auth expiry, server down), say so on "
+                "this issue rather than retrying -- `self-healing-retry.md` rule 5 stops "
+                "environmental failures on attempt 1."
+            ),
+        }
+    if detector_id == "launchd-dark":
+        return {
             "subject": label,
             "title": f"launchd job is dark: {label}",
             "body": (
@@ -140,7 +185,23 @@ def detect_dark_jobs(_ctx) -> list:
                 f"- Resume: `launchctl bootstrap gui/$(id -u) ~/Library/LaunchAgents/{label}.plist`\n"
                 f"- Or record the pause: add `{label}` to `~/.config/kipi/launchd-paused.txt`"
             ),
-        })
+        }
+    # Loud, not silent: a caller inventing a launchd detector id would otherwise
+    # file an issue with an empty body under a key the other filer also writes.
+    raise ValueError(f"no launchd rendering for detector id {detector_id!r}")
+
+
+def detect_dark_jobs(_ctx) -> list:
+    """A plist on disk, not loaded, and NOT in the paused ledger = silent death."""
+    paused = _paused_labels()
+    out = []
+    for label in _launchd_labels():
+        if not label.startswith(("com.kipi.", "com.cole.", "com.ask.", "com.assaf.",
+                                 "com.claudedaddy.", "com.purespectrum.", "com.personal.")):
+            continue
+        if label in paused or _is_loaded(label):
+            continue
+        out.append(launchd_finding("launchd-dark", label))
     return out
 
 
@@ -162,17 +223,8 @@ def detect_failing_jobs(_ctx) -> list:
             continue
         if status in ("0", "-"):
             continue
-        out.append({
-            "subject": label,
-            "title": f"launchd job failing: {label} (exit {status})",
-            "body": (
-                f"`{label}` is loaded but its last run exited **{status}**.\n\n"
-                "## Action\nRead its `StandardErrorPath`, fix the cause, and confirm a clean "
-                "run. If the failure is environmental (auth expiry, server down), say so on "
-                "this issue rather than retrying -- `self-healing-retry.md` rule 5 stops "
-                "environmental failures on attempt 1."
-            ),
-        })
+        out.append(launchd_finding("launchd-failing", label,
+                                   launchd_exit_detail(status)))
     return out
 
 
@@ -381,6 +433,19 @@ _WRAPPER_VALUE_FLAGS = {
 # a lock file or a host named `claude`.
 _WRAPPER_OPERANDS = {"flock": 1, "ssh": 1}
 
+# ssh does NOT exec its remote operands the way flock and sudo do. It joins
+# everything after the destination with single spaces and hands ONE string to a
+# shell on the far side, which lexes it itself. So the remote command has to be
+# re-parsed as a command string -- exactly like a `-c` argument -- and reading it
+# as tokens in the local segment made the QUOTED form (`ssh mini 'claude -p
+# sweep'`, the form you write so the LOCAL shell does not expand it) lex to a
+# single token whose basename is `claude -p sweep`, not `claude`. The unquoted
+# form was pinned by the suite and the quoted one was a silent false negative
+# (PR #19 round-3 review, minor 2). flock stays OUT of this set on purpose: it
+# execs its operands directly, so `flock /tmp/x.lock echo 'a; claude -p x'` runs
+# no claude and re-lexing it would invent one.
+_REMOTE_SHELL_WRAPPERS = {"ssh"}
+
 # A shell in command position does not RUN what follows; it runs the string
 # handed to its `-c` flag. `bash -lc 'claude -p ...'` is the shape a cron line
 # uses to get a login environment, so the command inside the quotes has to be
@@ -463,18 +528,32 @@ def _takes_next_token(token: str, wrapper: str) -> bool:
 
 
 def _command_index(tokens: list) -> int:
-    """Index of the real command in a shell segment, past env prefixes and
-    wrappers. -1 when the segment holds no command.
+    """Index of the real command in a shell segment. See `_command_scan`."""
+    return _command_scan(tokens)[0]
+
+
+def _command_scan(tokens: list) -> tuple:
+    """(command index, remote-command index) for a shell segment.
+
+    The command index is the real command past env prefixes and wrappers, or -1
+    when the segment holds no command.
 
     Tracks WHICH wrapper is in effect, because a bare `startswith("-")` skip
     reads a flag and leaves its value exposed: `sudo -u claude run.sh` scored
     `claude` -- the value of `-u` -- as the command (fourth review, finding 4).
     An option's argument belongs to the option, and a wrapper's leading operand
     (flock's lock file, ssh's host) belongs to the wrapper.
+
+    The remote index is where a `_REMOTE_SHELL_WRAPPERS` wrapper's remote command
+    STARTS (-1 when there is none). Recorded here rather than derived from the
+    command index because a second wrapper resets the tracked one: in `ssh mini
+    timeout 30 'claude -p x'` the wrapper in effect at the command is `timeout`,
+    yet everything from `timeout` onward is still what ssh hands the far shell.
     """
     skip_value = False
     operands_left = 0
     wrapper = ""
+    remote_index = -1
     for index, token in enumerate(tokens):
         if skip_value:
             skip_value = False
@@ -483,7 +562,10 @@ def _command_index(tokens: list) -> int:
             continue  # VAR=value prefix
         if token.startswith("-"):
             if _is_lookup_flag(token, wrapper):
-                return -1  # `command -v claude` resolves a path, it runs nothing
+                # `command -v claude` resolves a path, it runs nothing. The
+                # remote index rides along: over ssh the far shell re-parses the
+                # string and reaches the same answer for the same reason.
+                return -1, remote_index
             skip_value = _takes_next_token(token, wrapper)
             continue
         if token in SHELL_KEYWORDS:
@@ -497,9 +579,11 @@ def _command_index(tokens: list) -> int:
             continue
         if operands_left:
             operands_left -= 1
+            if not operands_left and wrapper in _REMOTE_SHELL_WRAPPERS:
+                remote_index = index + 1  # past ssh's destination: the remote command
             continue  # flock's lock file / ssh's destination, not the command
-        return index
-    return -1
+        return index, remote_index
+    return -1, remote_index
 
 
 def _command_token(tokens: list) -> str:
@@ -772,7 +856,7 @@ def _shells_claude(command: str, _depth: int = 0) -> bool:
     # split an assignment prefix across two tokens (see _split_substitutions).
     masked, inner_commands = _split_substitutions(command)
     for tokens in _shell_segments(masked):
-        command_index = _command_index(tokens)
+        command_index, remote_index = _command_scan(tokens)
         command_token = tokens[command_index] if command_index >= 0 else ""
         if _is_claude_token(command_token):
             return True
@@ -780,6 +864,25 @@ def _shells_claude(command: str, _depth: int = 0) -> bool:
             inner = _shell_c_argument(tokens, command_index)
             if inner and _shells_claude(inner, _depth + 1):
                 return True
+        # ssh's remote operands are JOINED with spaces and re-lexed by the far
+        # shell, so they are a command STRING, not tokens in this segment. Joining
+        # the already-lexed tokens is the faithful model: ssh concatenates its
+        # argv after the local shell removed the quotes, which is why `ssh mini
+        # echo "a; claude -p x"` really does run claude on the far side.
+        if remote_index >= 0:
+            remote = " ".join(tokens[remote_index:])
+            if remote and _shells_claude(remote, _depth + 1):
+                return True
+    # A line sh REFUSES TO RUN runs nothing -- including whatever sits inside its
+    # substitutions. `_shell_segments` already obeys that (an unbalanced quote
+    # collapses to one coarse segment and never splits on an operator), and its
+    # rule is "one segment can only be coarser than sh, never finer". The walk
+    # below is FINER: `_split_substitutions` happily extracts `$(claude -p x)`
+    # out of an unterminated double-quoted span and scored it as an invocation,
+    # filing a permanent Linear issue for a line /bin/sh answers with a syntax
+    # error (PR #19 round-3 review, minor 3).
+    if _lex(_strip_comment(masked)) is None:  # the exact test `_shell_segments` makes
+        return False
     # A substitution is a command position of its own, and the segment walk above
     # cannot reach one: it saw a placeholder. Same recursion, same depth cap as
     # the `-c` string: what runs inside `$( )` or backticks is a command, not text.
--- a/q-system/.q-system/scripts/fleet-health-daily.py
+++ b/q-system/.q-system/scripts/fleet-health-daily.py
@@ -222,7 +222,44 @@ def _crontab_text() -> str:
     return _read_crontab_result(res.returncode, res.stdout, res.stderr)
 
 
-def detect_duplicate_schedules(_ctx, cron_text=None) -> list:
+class RunContext:
+    """Per-run shared state for the detectors. Today: ONE `crontab -l` per run.
+
+    Two detectors read the crontab. `run_detectors` passed None as the context, so
+    each one shelled out for itself and a test asserting only that a `cron_text`
+    PARAMETER existed was labelled proof that "neither shells out twice" (PR #19
+    review, minor 4). The parameter existed; the wiring did not.
+
+    The FAILURE is cached alongside the value and re-raised per caller, so one read
+    still produces two independent blind spots. Collapsing them would buy the dedup
+    with a silent all-clear for whichever detector ran second -- the exact trade
+    `run_detectors` exists to refuse.
+    """
+
+    def __init__(self, read_crontab=None):
+        self._read_crontab = read_crontab or _crontab_text
+        self._crontab = None  # (text, exception), whichever the read produced
+
+    def crontab(self) -> str:
+        if self._crontab is None:
+            try:
+                self._crontab = (self._read_crontab(), None)
+            except Exception as exc:  # noqa: BLE001 - replayed to every caller
+                self._crontab = (None, exc)
+        text, exc = self._crontab
+        if exc is not None:
+            raise exc
+        return text
+
+
+def _crontab_for(ctx, cron_text):
+    """The crontab a detector should read: injected > shared > its own read."""
+    if cron_text is not None:
+        return cron_text
+    return ctx.crontab() if isinstance(ctx, RunContext) else _crontab_text()
+
+
+def detect_duplicate_schedules(ctx, cron_text=None) -> list:
     """The same script scheduled by BOTH launchd and crontab.
 
     Found live 2026-07-26: reddit-build-radar runs at 08:00 from
@@ -234,7 +271,7 @@ def detect_duplicate_schedules(_ctx, cron_text=None) -> list:
     that reaches the body is one that already appears in `~/Library/LaunchAgents/`.
     A path that matched nothing is never published.
     """
-    cron = _crontab_text() if cron_text is None else cron_text
+    cron = _crontab_for(ctx, cron_text)
     # Resolve to ABSOLUTE paths, never basenames. Measured 2026-07-26: matching on
     # the basename `run_daily.sh` hit 3 plists (reddit-radar-daily, daily-podcast,
     # story-podcast) when only ONE is a real duplicate -- the other two are
@@ -592,7 +629,14 @@ def _split_substitutions(command: str) -> tuple:
             masked.append(command[index:index + 2])
             index += 2  # escaped next char, in double quotes or unquoted
             continue
-        if char == "'":
+        if char == "'" and not quote:
+            # `and not quote` is load-bearing: reaching here with quote == '"'
+            # means an APOSTROPHE inside a double-quoted span, which sh treats as
+            # a literal character. Opening a single-quote span on it inverted the
+            # state for the rest of the line, so `echo "don't $(claude -p x)"` --
+            # a real invocation -- was missed, and `echo "don't" 'run `claude -p
+            # x` now'` -- prose -- was scored as one and would have filed a
+            # PERMANENT false-positive issue (PR #19 review, minor 2).
             quote = "'"
         elif char == '"':
             quote = "" if quote == '"' else '"'
@@ -760,7 +804,7 @@ def offending_cron_lines(cron: str) -> list:
             if _shells_claude(_cron_command(line))]
 
 
-def detect_cron_shells_claude(_ctx, cron_text=None) -> list:
+def detect_cron_shells_claude(ctx, cron_text=None) -> list:
     """A crontab line that invokes `claude`. It cannot work: cron has no keychain.
 
     Probed 2026-07-23 (`reddit-build-radar/logs/cron-probe/result.txt`):
@@ -782,7 +826,7 @@ def detect_cron_shells_claude(_ctx, cron_text=None) -> list:
     Raises CrontabUnavailable when the crontab could not be read at all, so a
     blind run is reported as an error rather than as a clean bill of health.
     """
-    cron = _crontab_text() if cron_text is None else cron_text
+    cron = _crontab_for(ctx, cron_text)
     numbers = offending_cron_lines(cron)
     if not numbers:
         return []
@@ -1349,9 +1393,11 @@ def run_detectors(detectors=None) -> tuple:
     """
     all_findings = []
     per_detector = {}
+    # ONE context for the whole run, so the two crontab readers share a read.
+    context = RunContext()
     for d in detectors if detectors is not None else DETECTORS:
         try:
-            found = d["detect"](None) or []
+            found = d["detect"](context) or []
         except Exception as exc:  # noqa: BLE001 - a health check must not crash its job
             print(f"  detector {d['id']} errored: {exc}", file=sys.stderr)
             per_detector[d["id"]] = DETECTOR_ERROR
--- a/q-system/.q-system/scripts/fleet-health-daily.py
+++ b/q-system/.q-system/scripts/fleet-health-daily.py
@@ -28,6 +28,24 @@ So a detector here is not shippable without all three legs:
 `validate_detectors()` refuses to run if any detector breaks that contract, and
 test-fleet-health-daily.py asserts it. Prose cannot hold this line; the registry can.
 
+A FINDING CARRIES A REFERENCE, NEVER UNTRUSTED TEXT (ASK-204):
+
+    A Linear issue filed here is PERMANENT. Any detector that reads arbitrary
+    operator input -- a crontab line, an exception message -- publishes a
+    reference to it (which line, which exception type) and never the input
+    itself. The previous design copied the crontab line in with best-effort
+    regex redaction over it; that is a denylist against unbounded shell syntax,
+    and nine review rounds found nine distinct bypasses (a Slack webhook path, a
+    Bearer token, a `bash -lc 'KEY=... claude'` credential, a backtick word
+    start, `lin_api_`/`ntn_` shapes). Each fix was correct; the architecture was
+    not. The operator opens their own crontab. Nothing untrusted crosses into a
+    permanent external record, so there is nothing to redact and no bypass left
+    to find.
+
+    This constrains what a finding CARRIES, not what a detector READS. Deciding
+    whether a line shells `claude` still parses the line in full, including
+    quoting and command substitution.
+
 ALERTING: ONE Slack summary line, never one ping per finding. The findings live in
 Linear where they hold state. Slack is a pointer, not a ledger.
 
@@ -38,13 +56,15 @@ goes to stdout and the ledger.
 from __future__ import annotations
 
 import argparse
+import hashlib
 import json
 import os
 import re
+import shlex
 import subprocess
 import sys
 from datetime import datetime, timezone
-from pathlib import Path
+from pathlib import Path, PurePosixPath
 
 HERE = Path(__file__).resolve().parent
 QROOT = HERE.parent.parent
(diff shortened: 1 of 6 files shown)


assafkip:
## Findings

---

### 1. `major` — a Linear write that fails now reports byte-identically to a clean run on the launchd watchdog

`q-system/.q-system/scripts/fleet-health-daily.py:1210`

The new per-finding `except` converts a Linear failure into `result["errors"]` instead of letting it propagate. `launchd-health-check.py:361` reads exactly three keys (`created`, `existing`, `skipped_no_key`) and never `errors`, so its report and its Slack annotation both go silent. On `origin/main` the same failure raised out of `file_findings`, hit the caller's handler at `launchd-health-check.py:345`, and was counted as `unfiled=N` plus a `[NOT filed to Linear: N]` on the ping. This is a regression, in the job whose stated purpose is catching silent death, running unattended at 09:30 and 21:30.

**Repro** (real output, both halves):

```
$ python3 - <<'PY'   # inject a linear-sync stub whose graphql raises
...
=== a Linear write that FAILS, seen through launchd-health-check.py ===
BASE origin/main -> file_findings RAISED RuntimeError
                    watchdog stdout : linear: filed=0 already-tracked=0 unfiled=1
                    Slack NOT-filed : True
PR #19           -> file_findings returned {'created': 0, 'existing': 0, 'skipped_no_key': 0,
                                            'updated': 0, 'reopened': 0, 'relisted': 0, 'errors': 1}
                    watchdog stdout : linear: filed=0 already-tracked=0 unfiled=0
                    Slack NOT-filed : False
```

And against a refused mutation (`issueUpdate.success == false`) rather than a raise:

```
  filing fleet-health/launchd-dark/com-kipi-x FAILED: issueUpdate refused for ...
fleet-health-daily report : filed=0 already-tracked=0 unfiled=0 rewritten=0 reopened=0 relisted=0 errors=1
launchd-health-check report: linear: filed=0 already-tracked=0 unfiled=0
clean run reads            : linear: filed=0 already-tracked=0 unfiled=0
IDENTICAL to failed run?   : True
```

`fleet-health-daily.py`'s own `outcome_line` was updated and does surface `errors`. Its duplicate at `launchd-health-check.py:361` was not, and the docstring above that duplicate still calls it "the filing result as one line that cannot lie by omission."

Mitigation that keeps this off `blocker`: the 08:15 job's own `should_notify` does read `errors`, so a *persistent* Linear failure still pings the next morning. A transient one at 09:30 is silent.

**3am cost:** the watchdog says the board has the findings; it does not.

---

### 2. `minor` — a single quote inside a double-quoted span inverts the substitution walker's quote state

`q-system/.q-system/scripts/fleet-health-daily.py:595`

```python
if char == "'":
    quote = "'"
```

That branch is reached while `quote == '"'`. In sh, an apostrophe inside double quotes is literal. Here it opens a single-quote span, and the `quote == "'"` fast path at the top of the loop then skips backtick and `$(` detection for the rest of the line. Both directions of the error reproduce. The docstring two lines up claims "Single quotes SUPPRESS substitution and double quotes do not" — the code does not implement that.

**Repro** (stub `claude` on PATH touches a marker; marker present == sh really invoked it):

```
REAL   DETECT  COMMAND
True   False   echo "don't $(claude -p x)"        <-- MISMATCH
True   False   echo "don't" `claude -p z`         <-- MISMATCH
True   False   echo "isn't `claude -p q` done"    <-- MISMATCH
True   True    echo "it is $(claude -p x)"        (same line without the apostrophe)
```

The false-positive direction files a permanent issue:

```
REAL   DETECT  COMMAND
False  True    echo "don't" 'run `claude -p x` now'          <-- FALSE POSITIVE
False  True    echo "won't" 'note: $(claude -p z)'           <-- FALSE POSITIVE
False  True    /usr/bin/say "it's 3am" ; echo 'tip: `claude -p sweep`'  <-- FALSE POSITIVE
```

End-to-end, as a crontab:

```
crontab under test (neither line runs claude):
   line 1: 0 2 * * * /usr/bin/rsync -a ~/a ~/b
   line 2: 0 3 * * * echo "don't forget" ; echo 'tip: `claude -p sweep`'

offending_cron_lines -> [2]
findings filed       -> 1
issue title          -> 1 crontab line(s) shell `claude` — cron has no keychain access
```

`echo 'run \`claude -p x\` now'` is in the PR's own `_MUST_NOT_DETECT` list. One apostrophe earlier in the line defeats it.

**Severity dropped a level, deliberately:** the operator's live crontab is empty (`crontab -l` rc=0, 0 lines, 0 lines with an apostrophe inside double quotes), so nothing emits this shape today. It is a real defect in operator-authored input, not a live one.

---

### 3. `minor` — the committed reproducer can only ever print GREEN

`q-system/output/ask204-repro.py:16`

The default target is `q-system/output/ask150-fleet-health-daily.py`. That file is not on the branch (`git ls-tree -r pr19 | grep ask150` returns nothing). The RED half of the PR's headline evidence is unreproducible by anyone reading it, and the missing-file crash exits 1 too, which an exit-code wrapper reads as "the defect reproduces."

**Repro:**

```
$ python3 q-system/output/ask204-repro.py           # its documented default
stderr: FileNotFoundError: ... 'q-system/output/ask150-fleet-health-daily.py'
EXIT = 1

$ python3 q-system/output/ask204-repro.py q-system/.q-system/scripts/fleet-health-daily.py
what the finding publishes:

GREEN: zero characters of any source line reach the body
EXIT = 0
```

The actual regression guard is in `test-fleet-health-daily.py:169-194`, which does assert zero leakage and would go red if a pattern table came back. That net holds; this artifact just does not do what its own docstring says.

---

### 4. `minor` — the test's name claims a property the production path does not have

`q-system/.q-system/scripts/test/test-fleet-health-daily.py:412`

The assertion is `"cron_text" in inspect.signature(fn).parameters` for both cron detectors, labelled "so neither shells out twice." `run_detectors` at `fleet-health-daily.py:1354` calls `d["detect"](None)` — one positional argument — so `cron_text` stays `None` and each detector runs its own `crontab -l`.

**Repro** (spy on `subprocess.run` inside the module, then the exact production call shape):

```
`crontab -l` invocations during ONE run of the two cron detectors: 2
    ['crontab', '-l']
    ['crontab', '-l']
```

Harmless in itself (an idempotent read), but the assertion reads as coverage of a deduplication that was never wired.

---

## What is sound

- **520-case differential fuzz of `_shells_claude` against real `/bin/sh`** — prefix × core × suffix over wrappers, operators, brace groups, keywords, substitutions, comments, redirects, and a decoy `./claude/` directory, with a stub `claude` on PATH proving whether sh actually invoked it. **Zero genuine mismatches.** All 38 apparent false positives were harness artifacts: `flock` and `timeout` are not on macOS PATH (`shutil.which` → `None`), so those lines never ran anything, and the `# claude -p later` suffix commented out the `; fi` of the `if` cases. The detector's answers were right in all 520.
- **The ASK-204 core claim holds.** Fed a fixture carrying a backtick-word-start `LINEAR_API_KEY=lin_api_...` assignment and a `VAR=\`echo secret\`` substitution: the finding was detected, and zero characters of either command survived into the body. I could not construct a crontab shape that puts source text in the issue.
- **`unfiled_reason` really is type-only.** A `RuntimeError("HTTP 401: Authorization: lin_api_...")` reaches neither the outcome dict nor `notify_text`.
- **Operator-note preservation survives two consecutive rewrites** without duplicating, including the migration path from a v1 body with no sentinel, and for both filer names.
- **Hash-based staleness does not churn.** An unchanged finding against a body this renderer wrote issues zero mutations (`sent == []`). I tried to make Linear's markdown re-serialization matter and could not — the hash rides in an HTML comment.
- **`fetch_issue` refuses to confuse "deleted" with "flaky."** A 429 or dropped connection re-raises instead of returning `{}`; only an `Entity not found` message yields the empty record that triggers a re-file. I tried to turn a transient error into a duplicate permanent issue and could not.
- **The two filers do not ping-pong each other's bodies.** `filer` sits inside the managed region but is excluded from `finding_hash`, so the 08:15 job and the 09:30 watchdog writing the same key do not rewrite each other daily. I tried to build that loop and could not.
- **`launchd-health-check.py` does not KeyError** — all three keys it reads with `[]` still exist in the widened result dict. The ASK-181 shape claim is true as stated.
- The full suite passes (`PASS: fleet-health-daily contract holds`), and `_MUST_DETECT` / `_MUST_NOT_DETECT` are real tests — I verified their cases against live sh, and flipping any of them goes red.

---

## VERDICT: REQUEST CHANGES

One `major`, so it does not merge as-is. **Not BLOCK** — nothing here is permanent or unrecoverable, and the reference-only architecture this PR is actually about is correct and well-defended.

**Fix first:** `launchd-health-check.py:361` has to read the buckets `file_findings` now returns. Either fold `errors` + `relisted` into that line (and into the `[NOT filed to Linear: N]` Slack annotation at line 418), or have `file_findings` roll write failures into `skipped_no_key` so the existing reader keeps working. The second is smaller and preserves the pre-PR contract exactly. Whichever you pick, the assertion that pins it belongs in `test_launchd_health_check.py`, not here — the current suite proves `fleet-health-daily`'s own line is honest while the watchdog's copy is not.

FINDINGS:
major|The new `errors` bucket in file_findings is unread by launchd-health-check.py, so a failed or refused Linear write reports byte-identically to a clean run and sends no Slack warning — a regression from origin/main, which raised and let the caller count it as unfiled|q-system/.q-system/scripts/fleet-health-daily.py:1210
minor|_split_substitutions treats an apostrophe inside a double-quoted span as opening a single-quote span, so a substitution after it is missed (false negative) and a genuinely single-quoted backtick after it is scored as an invocation (false positive, a permanent Linear issue)|q-system/.q-system/scripts/fleet-health-daily.py:595
minor|ask204-repro.py's default target q-system/output/ask150-fleet-health-daily.py is not on the branch, so the reproducer can never print RED and its FileNotFoundError also exits 1|q-system/output/ask204-repro.py:16
minor|The test labelled "so neither shells out twice" only asserts a parameter exists; run_detectors calls detect(None), so both cron detectors each run `crontab -l`|q-system/.q-system/scripts/test/test-fleet-health-daily.py:412
END FINDINGS

assafkip:
## Review round 3 — all four findings fixed, each with a test that went red first

Commit `9cab3ac`. Nothing disagreed with; nothing deferred.

---

### 1. `major` — a failed Linear write reported byte-identically to a clean run

**Fixed, and not the way either option in the review suggested.** Both of those
fix the four buckets that exist today. The finding is that the reporter reads
buckets **by name**, so it goes dark again the next time `file_findings` grows
one — which is exactly how this regression happened (ASK-181 fixed the line,
ASK-204 added `errors`, the line went silent). Rolling `errors` into
`skipped_no_key` would also make `notify_text`'s "Linear unreachable" sentence
false for a refused write, and zero out `errors=` on fleet-health-daily's own
line: one lie traded for two.

So `unfiled` is now **owed minus landed**, against an allowlist:

```python
LANDED_BUCKETS = ("created", "existing", "updated", "reopened", "relisted")

def unfiled_count(outcome):
    fallback = outcome.get("skipped_no_key", 0)
    if "owed" not in outcome:
        return fallback
    landed = sum(outcome.get(bucket, 0) for bucket in LANDED_BUCKETS)
    return max(outcome["owed"] - landed, fallback)
```

`file_linear_findings` puts `owed` on all four return paths (it is the only side
that knows the denominator). A bucket nobody has taught the reporter about now
counts as **unfiled until someone declares it landed** — fail-closed instead of
fail-open. An outcome without `owed` (hand-built, or an older copy of the pair
mid-`kipi update`) degrades to the pre-fix number, not to a wrong one.

Both readers fixed, not just the counter: `linear_report_line` **and** the
`[NOT filed to Linear: N]` Slack annotation at what is now line 465.

The two formatters stay separate, deliberately, and the docstring now says why:
this script must be able to report that `fleet-health-daily.py` is **missing**,
so it cannot import its reporter from the file whose absence it reports.

**Test (`test_launchd_health_check.py`), red before the fix:**

```
FAIL:
  - a refused write is counted as unfiled: got False, want True
  - an all-refused run does not print the same line as a run with nothing to file: got True, want False
  - the ping says the board does NOT have the refused finding: got False, want True
  - a bucket the reporter has never heard of still counts as unfiled: got False, want True
```

The second one is your exact shape (every write refused vs nothing to file, both
printing `filed=0 already-tracked=0`). The fourth is the layer-above guard: a
stub returning `{"created": 1, "quarantined": 1}` — a bucket that does not exist
anywhere in the codebase — must still report `unfiled=1`. That assertion is what
makes the *next* bucket safe, and it is the one that could not have been written
against either suggested fix.

Also pinned: a `relisted` finding counts as **filed**, so the fix does not buy
this silence back as a nightly false alarm.

Green after: `PASS: all launchd-health-check logic checks green` (exit 0).

---

### 2. `minor` — apostrophe inside a double-quoted span inverted the quote state

**Fixed.** `if char == "'" and not quote:`. Reaching that branch with
`quote == '"'` is an apostrophe inside double quotes, which sh treats as literal.

Your six cases plus three controls, differentialled against real `/bin/sh` with a
stub `claude` on PATH (marker file present == sh actually invoked it):

```
REAL   DETECT  COMMAND
True   True    echo "don't $(claude -p x)"
True   True    echo "don't" `claude -p z`
True   True    echo "isn't `claude -p q` done"
True   True    echo "it is $(claude -p x)"
False  False   echo "don't" 'run `claude -p x` now'
False  False   echo "won't" 'note: $(claude -p z)'
False  False   /usr/bin/say "it's 3am" ; echo 'tip: `claude -p sweep`'
False  False   echo 'run `claude -p x` now'
True   True    claude -p "x"

mismatches: 0
```

Three cases added to `_MUST_DETECT`, two to `_MUST_NOT_DETECT` (both error
directions). Red before the fix:

```
  - still detects: echo "don't $(claude -p x)": got False, want True
  - still detects: echo "don't" `claude -p z`: got False, want True
  - still detects: echo "isn't `claude -p q` done": got False, want True
  - still refuses: echo "don't" 'run `claude -p x` now': got True, want False
  - still refuses: echo "won't" 'note: $(claude -p z)': got True, want False
```

---

### 3. `minor` — the committed reproducer could only ever print GREEN

**Fixed.** Three changes, because the artifact had three separate ways of being
useless:

- default target is `q-system/.q-system/scripts/fleet-health-daily.py`, resolved
  from `__file__` so it runs from any cwd;
- a target that will not load exits **2** (INCONCLUSIVE), never 1;
- `--legacy` reconstructs the ask-150 **output channel** — same detector, publish
  the line instead of its number — so the RED half runs from this branch without
  a copy of the deleted file.

`--self-test` asserts all three exit codes, so it cannot go one-sided again:

```
$ python3 q-system/output/ask204-repro.py --self-test
  ok : the old output channel reproduces the leak (exit 1, want 1)
  ok : the shipped detector publishes no source text (exit 0, want 0)
  ok : a missing target is INCONCLUSIVE, not RED (exit 2, want 2)

SELF-TEST PASS: RED, GREEN and INCONCLUSIVE all reachable from this branch
```

```
$ python3 q-system/output/ask204-repro.py --legacy
what the ask-150 output channel publishes:
    - `0 3 * * * bash -lc '`LINEAR_API_KEY=lin_api_backtickleak claude -p sweep`'`
    - `0 4 * * * claude -p 'sweep' --header lin_api_bareleak`
    - `0 5 * * * claude -p 'sweep' --header ntn_barenotionleak`

RED: 3 secret(s) reach the Linear issue body verbatim          # exit 1

$ python3 q-system/output/ask204-repro.py
what the finding publishes:
    - line 1
    - line 2
    - line 3

GREEN: zero characters of any source line reach the body       # exit 0
```

The self-test is not wired into the fleet suite on purpose: `q-system/output/`
is gitignored and does not sync via `kipi update`, so a synced test that required
this file would fail in every instance. The regression net you named
(`test-fleet-health-daily.py:169-194`) remains the one that guards the behaviour.

---

### 4. `minor` — the test claimed a dedup that was never wired

**Wired it, then measured it.** `RunContext` reads the crontab once per run and
`run_detectors` passes it to every detector.

The layer above: the context caches the **failure** alongside the value and
re-raises it per caller, so one read still produces two independent blind spots.
Collapsing them would buy the dedup with a silent all-clear for whichever cron
detector ran second — the trade `run_detectors` exists to refuse.

The old assertion (parameter-exists) is kept, honestly relabelled "both cron
detectors accept injected text". The claim it used to make is now measured by
spying on the module's `subprocess.run`. Red before the fix:

```
  - run_detectors reads the crontab ONCE for both cron detectors: got 2, want 1
  - an unreadable crontab is still read only once: got 2, want 1
```

and the blind-spot half, which passes on both sides and must keep doing so:

```
  ok: an unreadable crontab is still read only once
  ok: ...and BOTH cron detectors are reported blind, not clean
```

---

## What I did not regress

Re-ran everything the review said it tried and could not break. Full suites, and
a live dry run of the watchdog against the real fleet:

```
PASS: fleet-health-daily contract holds                    # exit 0

$ python3 q-system/.q-system/scripts/test_launchd_health_check.py
PASS: all launchd-health-check logic checks green          # exit 0

$ python3 q-system/.q-system/scripts/launchd-health-check.py --dry
...
[dry] linear: would-file=0 already-tracked=0 unfiled=0
[dry] would ping 0 job(s)                                  # exit 0
```

One existing assertion changed, and it is mine: `a paused-only run files nothing`
now expects `{"created": 0, "existing": 0, "skipped_no_key": 0, "owed": 0}`.
`owed` rides on every return path — a path that omitted it would silently fall
back to reading one bucket, which is the defect.

Nothing here makes a downstream signal quieter. Finding 1 makes the 09:30/21:30
ping **louder** in exactly one case (a refused write) and unchanged everywhere
else; the `relisted`-is-landed assertion is what stops that from becoming noise.

Fleet gate and a live no-write run of the detector, with the shared-crontab
wiring in place:

```
$ python3 q-system/.q-system/scripts/capability-gate.py --repo-root .
  tests: ran=72 quarantined=0 skipped-skeleton-only=0
capability-gate: GREEN                                     # exit 0

$ python3 q-system/.q-system/scripts/fleet-health-daily.py --quiet
fleet-health 2026-07-27T19:09:47Z
  unwired-untracked: 0
  launchd-dark: 0
  launchd-failing: 0
  schedule-duplicate: 0
  cron-shells-claude: 0
  open-spillover: 1
  filed=0 already-tracked=0 unfiled=0 rewritten=1 reopened=0 relisted=0 errors=0
```

assafkip:
Read the code first, then the prior rounds. Three findings survived reproduction.

## Findings

### 1. `major` — two filers render different bodies for the same kipi-key, so the new update path rewrites the issue and pings Slack twice a day forever

`q-system/.q-system/scripts/fleet-health-daily.py:1323`

`file_findings` is the fleet's ONE filer (ASK-181). Both `fleet-health-daily.py` (08:15) and `launchd-health-check.py` (09:30, 21:30) file against the same key, `fleet-health/launchd-failing/<label>`. Their `dark` bodies are byte-identical. Their `failing` bodies are not:

- `fleet-health-daily.py:169` → ``  `{label}` is loaded but its last run exited **{status}**.``
- `launchd-health-check.py:293` → ``  `{label}` is loaded but its last run exited non-zero (**{detail}**).``

`finding_hash` covers title + body, so the hash flips on every alternating run and `_refresh_one` issues a mutation. This is a regression from `origin/main`, where `file_findings` did `result["existing"] += 1; continue` on a known key (base `fleet-health-daily.py:443`) and never wrote.

Reproducer (calls the real `wd.linear_findings` producer with the tuple `run()` builds at `launchd-health-check.py:192`):

```
python3 .review-ask204/repro.py
```

Real output:

```
same key?        True
same title?      True
same body?       False
fleet hash       0ad24a272a9112bc
launchd hash     bbdbc114c8c6cf7c

  day 1 08:15 fleet-health-daily  updated=0 existing=1 linear-writes=0  slack=False
  day 1 09:30 launchd-watchdog    updated=1 existing=0 linear-writes=1  slack=True
  day 2 08:15 fleet-health-daily  updated=1 existing=0 linear-writes=1  slack=True
  day 2 09:30 launchd-watchdog    updated=1 existing=0 linear-writes=1  slack=True

  7 days, 14 runs -> 13 Linear mutations, 6 fleet-health Slack lines, ZERO change on the fleet
  the Slack line the operator gets every morning:
    fleet health: 0 new issue(s) filed, 0 reopened, 1 updated, 3 already tracked. Board has them; nothing to do now.
```

What a human does at 3am: nothing, and that is the cost. The line says "1 updated" and closes with "nothing to do now" on a fleet where nothing moved. That is the checker training the operator to skip it, which is what the real alert will be competing with later. Recoverable (make the two bodies agree, or hash only the fields both filers own), so major, not blocker. Fires the first time any `com.*` job exits non-zero, which is the detector's core case, not an exotic one.

### 2. `minor` — a quoted remote command over ssh is a false negative

`q-system/.q-system/scripts/fleet-health-daily.py:382`

`_WRAPPER_OPERANDS = {"ssh": 1}` skips the destination, and the suite pins `ssh mini claude -p 'x'`. The quoted form (the one you write so the local shell does not expand it) lexes to a single token and `_is_claude_token` sees `claude -p sweep`, not `claude`. Nothing re-parses an ssh remote command the way `_shell_c_argument` re-parses a `-c` string.

Same reproducer, R2 section. Real output:

```
  _shells_claude("ssh mini claude -p 'x'") -> True
  _shells_claude("ssh mini 'claude -p sweep'") -> False
  _shells_claude('ssh mini "claude -p sweep"') -> False
  _shells_claude("ssh mini -- 'claude -p sweep'") -> False
```

### 3. `minor` — an unbalanced double quote around a substitution scores as an invocation

`q-system/.q-system/scripts/fleet-health-daily.py:786`

`_shell_segments` deliberately refuses to invent command positions on an unparsable line (`_MUST_NOT_DETECT` pins `echo 'reminder: && claude -p x`). The `for inner in inner_commands` loop bypasses that guard: `_split_substitutions` extracts and recurses into a substitution that sits inside an unterminated double-quoted span, which `/bin/sh` never runs at all. A permanent Linear issue for a line that cannot execute.

Same reproducer, R3 section. Real output:

```
  _shells_claude("echo 'reminder: && claude -p x") -> False
  _shells_claude('echo "reminder: $(claude -p x)') -> True
  _shells_claude('echo "note: `claude -p x`') -> True
```

Severity note per the calibration rule: this needs an operator typo in a crontab, not a shape any producer in this repo emits, so it stays minor.

## What is sound

- **The leak, which is the whole point of the PR.** I could not get one character of crontab text into a Linear body. `offending_cron_lines` returns ints, the body interpolates `f"- line {n}"`, and the redaction machinery is gone rather than tightened. `detect_duplicate_schedules` does publish a path, but the string has to be both a crontab token and a verbatim substring of a fleet-owned plist, so I could not push arbitrary text through it either.
- **`unfiled_reason`.** The unreachable-Linear branch carries `type(exc).__name__` only; the message stays on stderr. Their `_DeadLinear` fixture proves it and I re-ran it green.
- **`read_ledger` corruption crashing the 08:15 job.** I chased this and it died: `linear-sync.py:85` skips malformed lines, so a corrupt cache degrades rather than raising out of `main()`. Dropped.
- **Double-counting in `unfiled_count`.** Every finding lands in exactly one bucket in `file_findings`, including the `relisted` path, so `owed - landed` cannot go negative or overcount. The round-1 major is genuinely fixed, and the allowlist shape is the right fix (an unknown bucket counts as unfiled).
- **Round-1's apostrophe minor and crontab-read-twice minor.** Re-ran the suite; the substitution battery is green and the `subprocess.run` spy measures 1 read. Not re-litigating either.
- **Operator-note splice.** Two consecutive rewrites preserve the note once with no duplication.

Test suites, run against the PR head:

```
test-fleet-health-daily.py       PASS   exit 0
test_launchd_health_check.py     PASS   exit 0
```

On review calibration: round 1's four findings and my R1 all land on code this PR introduces (the base file has no cron detector and no update path at all). That is not the round-3 pathology of new blockers on unchanged code. The signal worth naming is narrower: every round so far has found a defect in the newly added `_refresh_one` / shared-key path specifically. That path, not the parser, is where the remaining risk sits.

## VERDICT: REQUEST CHANGES

Fix first: make the two filers render the same body for `launchd-failing`, or narrow `finding_hash` to fields both filers own. Until then the fleet's own health check writes to a permanent Linear object twice a day and Slacks the founder every morning about a fleet that did not change.

Housekeeping: `.review-ask204/` is my untracked scratch tree (the repro plus an unpacked copy of the PR head). No tracked file was modified. The destructive-op hook blocked my cleanup, so removal is `rm -rf .review-ask204` from your shell when you want it gone.

FINDINGS:
major|fleet-health-daily.py and launchd-health-check.py render different bodies for the same kipi-key on launchd-failing, so the new hash-staleness path rewrites the Linear issue and sends a "1 updated" Slack line twice a day forever on an unchanged fleet|q-system/.q-system/scripts/fleet-health-daily.py:1323
minor|A quoted remote command over ssh (`ssh mini 'claude -p sweep'`) is never re-parsed, so it is a silent false negative while the unquoted form is detected|q-system/.q-system/scripts/fleet-health-daily.py:382
minor|_shells_claude recurses into a substitution extracted from inside an unterminated double-quoted span, bypassing the unbalanced-quote guard and filing a permanent issue for a line /bin/sh refuses to run|q-system/.q-system/scripts/fleet-health-daily.py:786
END FINDINGS

assafkip:
## Review round 4 — three findings fixed, each with a test observed red first

Commit `eb4e5eb`. Nothing disagreed with, nothing deferred. Two adjacent defects
found while fixing R1 are captured in the spillover ledger rather than mentioned.

---

### 1. `major` — two filers, one kipi-key, two renderings

**Fixed at the layer you named, and the divergence is wider than the repro
showed.** Your fixture fed both sides the same pre-formatted detail and measured
`same title? True`. On real data the title diverges too:

```
  SIGKILL (`launchctl list` says '-9', LastExitStatus=9):
    same key?   True   fleet-health/launchd-failing/com-kipi-demo-failing
    same title? False
        fleet-health : launchd job failing: com.kipi.demo-failing (exit -9)
        watchdog     : launchd job failing: com.kipi.demo-failing (exit 9)
    same body?  False
    same hash?  False  (ff873ddb4ecd6a3b vs 9a4400248cbd9b6c)
```

The two readers see ONE machine state in TWO encodings. `launchctl list` prints a
signed status; `launchctl list <label>` prints the raw wait status that
`normalize_exit` decodes. Measured live on this machine, 210 non-zero rows:

```
  com.apple.progressd  : list-column='-9'  LastExitStatus=9      (SIGKILL)
  com.apple.Finder     : list-column='-15' LastExitStatus=15     (SIGTERM)
  com.apple.BiomeAgent : list-column='5'   LastExitStatus=1280   (exit 5)
```

209 of the 210 are the signal shape, so the title half is the common case, not the
exotic one. `abs()` lands both encodings in one space; that correspondence is what
`launchd_exit_detail` encodes, with the measurement in the docstring.

So the fix is not "make the two body strings match" — that leaves two renderers
one edit away from diverging again, which is how this got here. Both filers now
call **one** renderer, `launchd_finding(detector_id, label, detail)`, keyed by the
same detector id the kipi-key is built from. `launchd-health-check.py` renders
nothing of its own. Delegating rendering is safe for the same reason delegating
the key already was: that path only runs when `fleet-health-daily.py` loaded, and
`file_linear_findings` already reports its ABSENCE as owed-but-unfiled rather than
letting it raise. An unknown detector id raises instead of filing an empty body
under a key the other script also writes.

**Test (`test_launchd_health_check.py`), driven from ONE fake launchctl answering
both readers** — not two hand-built fixtures, because the encoding split is half
the defect. A test that fed both sides the same detail would have passed while
production titled one issue `(exit -9)` and the other `(exit 9)`.

Four layers pinned:

- both renderings byte-identical (key, title, body, hash) for `exit 1` and for `SIGKILL`;
- **the consumer, not just the strings**: `_refresh_one` handed an issue carrying the
  *other* filer's rendering must answer `existing` — in both directions;
- **table-driven over `LINEAR_DETECTOR_BY_KIND`**, so adding a kind without covering
  it fails here instead of shipping a second rendering of a shared key;
- the stubs (`_Boom`, `_fh_stub`, `_fh_stub_shape`) **delegate** `launchd_finding`
  to the real module rather than re-implementing it — a stub with its own
  rendering goes on passing after the two filers drift apart again.

Red before the fix (production reverted, tests kept):

```
FAIL:
  - exit 1: both render the same body: got '`com.kipi.pairing-demo` is loaded but its last run exited **1**...', want '...exited non-zero (**exit 1**)...'
  - exit 1: so the staleness hash agrees: got 'b4efd6121fdd987c', want 'ccdd88a30dc11f53'
  - SIGKILL: both render the same title: got 'launchd job failing: com.kipi.pairing-demo (exit -9)', want '... (exit 9)'
  - SIGKILL: both render the same body / so the staleness hash agrees
  - the watchdog leaves fleet-health-daily's issue alone: got 'updated', want 'existing'
  - and fleet-health-daily leaves the watchdog's issue alone: got 'updated', want 'existing'
```

**This fix makes a downstream signal quieter, deliberately, and here is the
accounting.** It removes exactly one thing: the daily `1 updated ... Board has
them; nothing to do now` line for a fleet where nothing moved, plus the ~13
Linear mutations a week behind it. It removes nothing else — `updated` still
fires when the rendering genuinely changes (a deploy that edits the body), and
the `_refresh_one` assertions above are two-directional so the silence cannot be
bought by making either side permanently stale. The line that was removed was
false, which is the only kind worth buying silence with.

---

### 2. `minor` — a quoted remote command over ssh

**Fixed.** ssh does not exec its remote operands the way flock and sudo do: it
joins everything after the destination with single spaces and hands ONE string to
a shell on the far side, which lexes it there. So the remote command is re-parsed
as a command string, exactly like a `-c` argument. `flock` stays out of that set
on purpose — it execs its operands directly, so re-lexing them would invent an
invocation.

The remote start index is recorded during the scan rather than derived from the
command index, because a second wrapper resets the tracked one: in `ssh mini
timeout 30 'claude -p x'` the wrapper in effect at the command is `timeout`, yet
everything from `timeout` onward is still what ssh hands the far shell.

Differentialled against real `/bin/sh`, with the far side modelled the way ssh
actually builds it (`sh -c "<joined operands>"`) and a stub `claude` on PATH —
marker present means sh really invoked it. `timeout` is stubbed too, so the
round-1 harness artifact (not on macOS PATH) cannot recur:

```
REAL   DETECT  COMMAND
True   True    ssh mini claude -p 'x'
True   True    ssh mini 'claude -p sweep'
True   True    ssh mini "claude -p sweep"
True   True    ssh mini -- 'claude -p sweep'
True   True    ssh mini timeout 30 'claude -p x'
True   True    ssh mini echo "a; claude -p x"
False  False   ssh mini tar -czf /tmp/b.tgz /tmp/projects/claude
False  False   ssh mini echo 'run claude -p tomorrow'
False  False   ssh claude@mini ./run.sh

mismatches: 0
```

`ssh mini echo "a; claude -p x"` is the one worth naming: it really does run
claude remotely, because ssh's join drops the quotes the local shell already
removed. Joining lexed tokens models that faithfully rather than by accident.
`ssh claude@mini ./run.sh` (your `_MUST_NOT_DETECT` entry, a remote user named
claude) still refuses.

Five cases added to `_MUST_DETECT`, two to `_MUST_NOT_DETECT`. Red before:

```
  - still detects: ssh mini 'claude -p sweep': got False, want True
  - still detects: ssh mini "claude -p sweep": got False, want True
  - still detects: ssh mini -- 'claude -p sweep': got False, want True
  - still detects: ssh mini timeout 30 'claude -p x': got False, want True
  - still detects: ssh mini echo "a; claude -p x": got False, want True
```

---

### 3. `minor` — recursing into a substitution inside an unterminated quote

**Fixed, by making the walk obey the rule `_shell_segments` already wrote down.**
That docstring's rule is *"one segment can only be coarser than sh, never finer."*
The substitution walk was FINER: it extracted `$(claude -p x)` out of an
unterminated double-quoted span and scored it, inventing a command position sh
never creates. `_shells_claude` now returns False when the masked line does not
lex — the exact test the segment walk makes.

The asymmetry with `claude -p 'sweep the repo` (still detected, still in
`_MUST_DETECT`) is deliberate and is that same rule: the coarse fallback only
reports what is ALREADY at line-lead command position, it never invents a new one.

Three cases added to `_MUST_NOT_DETECT`, red before:

```
  - still refuses: echo "reminder: $(claude -p x): got True, want False
  - still refuses: echo "note: `claude -p x`: got True, want False
  - still refuses: VAR="x $(claude -p sweep): got True, want False
```

Same differential, real `/bin/sh`, zero mismatches — including the controls that
must stay True (`echo "it is $(claude -p x)"`, `echo "don't $(claude -p x)"`).

---

## Captured, not mentioned

Two real defects found while fixing R1, both **pre-existing on `origin/main`** and
outside this issue's scope. In `.prd-os/spillover.jsonl`, so `gates run` stays red
until they are resolved:

- **`sp-b0758bc1`** — `normalize_exit` reports a SIGNAL death as an exit code. Raw
  wait-status 9 is `< 256` so it passes through as "exit 9"; `launchctl list`
  calls the same job `-9`. Both filers now agree on that label — they agree on a
  wrong one, and an operator reading "exit 9" hunts an exit code that was never
  returned.
- **`sp-410309f2`** — the two readers disagree on whether to **file at all**, not
  just on wording. `classify_status` carves out a benign KeepAlive restart
  (SIGTERM + live PID → ok, scar 2026-07-18, `com.cole.linkedin-session`);
  `detect_failing_jobs` has no such carve-out and files `-15` as failing. So the
  08:15 job opens a permanent Linear issue for exactly the case the 09:30 watchdog
  was taught to suppress. Fixing it means changing what `fleet-health-daily.py`
  FILES, which is a behaviour change this issue does not scope.

---

## What I did not regress

Everything you said you tried and could not break, re-run against this head:

```
PASS: fleet-health-daily contract holds                       # exit 0

$ python3 q-system/.q-system/scripts/test_launchd_health_check.py
PASS: all launchd-health-check logic checks green             # exit 0

$ python3 q-system/output/ask204-repro.py --self-test
SELF-TEST PASS: RED, GREEN and INCONCLUSIVE all reachable     # exit 0

$ python3 q-system/output/ask204-r3-differential.py
mismatches: 0                                                 # exit 0

$ python3 q-system/.q-system/scripts/capability-gate.py --repo-root .
  tests: ran=72 quarantined=0 skipped-skeleton-only=0
capability-gate: GREEN                                        # exit 0

$ python3 q-system/.q-system/scripts/launchd-health-check.py --dry
[dry] linear: would-file=0 already-tracked=0 unfiled=0
[dry] would ping 0 job(s)                                     # exit 0

$ python3 q-system/.q-system/scripts/fleet-health-daily.py --quiet
fleet-health 2026-07-27T19:35:22Z
  unwired-untracked: 0   launchd-dark: 0   launchd-failing: 0
  schedule-duplicate: 0  cron-shells-claude: 0  open-spillover: 1
  filed=0 already-tracked=0 unfiled=0 rewritten=1 reopened=0 relisted=0 errors=0
```

That live `rewritten=1` is the load-bearing half of the silence accounting: the
open-spillover rollup really did change (the ledger went from 1 item to 3 when I
captured the two above), so a genuine content change still writes and still
reports. What stopped is the rewrite for content that did not change.

The ASK-204 core claim is untouched: `offending_cron_lines` still returns ints and
the leak battery is green. `unfiled_count`'s owed-minus-landed allowlist is
untouched. Round-1's apostrophe fix and the shared-crontab read are untouched and
still green.

One assertion changed shape and it is mine: the watchdog suite's fleet-health
stubs now delegate `launchd_finding` instead of only `finding_key`, and
`_BuildBoom` raises on it too — rendering is delegated now, so the missing-module
scar has to be modelled on the call that actually happens first.

Your calibration note landed: every round has found the defect in the shared-key /
`_refresh_one` path, not the parser. This round's fix moved that path from "two
producers that happen to agree" to "one producer", and the table-driven assertion
over `LINEAR_DETECTOR_BY_KIND` is aimed at the next kind rather than at this one.

assafkip:
Read the code first, then the prior rounds. Repros live in `security-remediation/ask204-review/` (gitignored; `git status` is unchanged — no repo modification).

## Findings

### 1. `minor` — a wrapper whose argument is a command substitution loses command-position tracking, so the new masking silently misses a real invocation

`fleet-health-daily.py:674` (`_SUBST_PLACEHOLDER`, introduced by this PR)

`_split_substitutions` collapses each substitution to `__kipi_subst__`. That token matches no rule in `_command_scan` — not `VAR=`, not `-flag`, not a keyword, not `_WRAPPER_ARG_RE`, not a wrapper name — so line 585 returns it as **the command**. `claude` never reaches command position. Ground truth is `/bin/sh` with a stub `claude` on PATH:

```
$ python3 security-remediation/ask204-review/repro_head_vs_sh.py
cron command                                 sh runs claude  detector 
------------------------------------------------------------------------
timeout $(echo 30) claude -p sweep           False           False    
timeout `echo 30` claude -p sweep            False           False    
nice -n $(echo 5) claude -p sweep            True            True     
env $(echo FOO=1) claude -p sweep            True            False      <-- FALSE NEGATIVE
VAR=`echo secret` claude -p x                True            True     
claude -p sweep                              True            True     
cd /tmp && claude -p sweep                   True            True     

disagreements with /bin/sh: 1
EXIT=1
```

`nice -n $(...)` survives only because `-n` eats the placeholder as its value. The `timeout` rows are inconclusive on macOS (no GNU `timeout` on PATH), not clean passes.

Calibration, stated plainly: no line in this machine's crontab emits this shape, so this is a narrow miss on an input nobody currently hits. It stays at minor rather than dropping to nit because the code is new in this PR and the failure is silent in a detector whose entire value is a negative result.

---

### 2. `minor` — `cd ~/...` is never expanded, so the duplicate-schedule detector cannot fire on the ordinary cron idiom

`fleet-health-daily.py:341`

`Path("~/projects/x")` keeps a literal tilde. A launchd plist stores absolute paths, so `name in text` can never hit. Repro built from a **real plist off this machine**, not an invented fixture:

```
$ python3 security-remediation/ask204-review/repro_dupe_tilde.py
real plist:   com.assaf.competitive-analysis.morning
real script:  /Users/assafkipnis/projects/cole-gtm/projects/competitive-analysis/scripts/run_morning.sh

cd ~/...  (the ordinary idiom)   -> 0 finding(s)
    cron line: 0 8 * * * cd ~/projects/cole-gtm/projects/competitive-analysis/scripts && ./run_morning.sh >/dev/null 2>&1
cd /Users/... (absolute)         -> 1 finding(s)
    cron line: 0 8 * * * cd /Users/assafkipnis/projects/cole-gtm/projects/competitive-analysis/scripts && ./run_morning.sh >/dev/null 2>&1

RED: the same double-schedule is detected with an absolute cd and MISSED with `cd ~/...`
EXIT=1
```

This is **unchanged code** — `git show 72c782d` has the identical `base_dir` line. This PR carries it forward and adds a docstring above it, it did not introduce it. Not raised in any prior round: round 2 probed `detect_duplicate_schedules` for *leaks* (comment line 482) and correctly cleared it, but never checked whether it *detects*. One-line fix (`.expanduser()`); at 3am the cost is a "paused" job that keeps running with nothing on the board.

---

- **The leak, which is the whole point of the PR.** I could not push one character of crontab text into a Linear body. `offending_cron_lines` returns ints; the body interpolates `f"- line {n}"`. `detect_duplicate_schedules` does publish a path, but the string must be *both* a crontab token *and* a verbatim substring of a fleet-owned plist — I tried and could not get arbitrary text through that gate.
- **The renderer consolidation (commit `eb4e5eb`).** I re-ran the encoding claim against **all 210 non-zero `launchctl list` rows on this machine**, comparing `launchd_exit_detail(column)` to `normalize_exit(LastExitStatus)`: **209/210 agree**. The single divergence is `com.apple.Finder` (`-15`, live PID) — a *classification* split, not an encoding one, and the author already captured it as `sp-410309f2` with an accurate description. `sp-b0758bc1` covers the signal-vs-exit-code labelling. Both confirmed by my repro; I am dropping both rather than re-filing them.
- **`unfiled_reason` carries `type(exc).__name__` only** (`file_findings:1326`); the message stays on stderr.
- **Operator-tail splice** survives two consecutive rewrites without duplicating, and the graded fallback preserves an unrecognised body whole.
- **Unterminated-quote guard** (`_shells_claude:884`) correctly refuses to invent a command position inside a span `/bin/sh` would reject.
- **Test fixtures.** I specifically checked `_FakeLinear` against the real producer: `fetch_remote_state` returns `(team_id, project, {key: {linear_id, identifier, description, state_type, state_name, team_id}})` and `fetch_issue` returns the same record shape from `ISSUE_BY_ID_QUERY`, which does select `description` and `team { id }`. The fixture matches. No mutex-fixture-mismatch class here.

## A finding about the review process, not the code

Round 3 produced **zero blockers and zero majors**. Every prior-round finding I could reach is either fixed with an executed differential or captured in the spillover ledger with an honest scope note. The two minors above are both in the *cron parser / duplicate detector*, not in the shared-key `_refresh_one` path where rounds 1–2 concentrated — and one of them is on code this PR never touched.

That pattern says the thing worth saying: **this PR is being reviewed past the point of return.** Nine rounds on #11 plus three here, and the architectural change (a reference, never the text) has been sound since round 1 of #19. Continuing to hunt an unchanged 500-line parser for a detector that files at most one rollup issue is not risk reduction, it is a wedge. Ship it and let the spillover ledger do its job.

## VERDICT: APPROVE WITH NITS

Only minor findings survived reproduction. Both are tracked as follow-ups by the pipeline.

**Fix first:** the tilde in `fleet-health-daily.py:341`. It silently zeroes a detector against the exact idiom its own headline catch (`reddit-build-radar`) was written from, and it is a one-line `.expanduser()`.

FINDINGS:
minor|`cd ~/...` is never expanded, so detect_duplicate_schedules cannot match any plist and silently reports zero for a real double-schedule|q-system/.q-system/scripts/fleet-health-daily.py:341
minor|The substitution placeholder is scored as command position, so `env $(...) claude -p x` — which /bin/sh really does run — answers False|q-system/.q-system/scripts/fleet-health-daily.py:674
END FINDINGS

PR #21, opened 2026-07-27
fix(prd-os): spillover resolve verifies a Linear issue, so finished work can leave the ledger (ASK-209)

assafkip:
Closes ASK-209. Resolves spillover `sp-4c59c1f5`.

## The defect

`spillover resolve --resolution-ref` verified closure by looking in `.prd-os/issues/`.
This repo's autonomous work ships through **Linear**, so a Linear identifier was never
found and the resolve was refused:

```
$ prd_runner.py spillover resolve sp-638006cc --resolution-ref ASK-204
cannot resolve sp-638006cc: issue 'ASK-204' is not closed.
```

ASK-204 **was** closed (PR #19, merge `990d7c1`). `gates run` fails while any item is
open, so every item fixed the normal way accumulated as permanently open and the gate
went RED on finished work. That is the failure mode where a gate stops carrying
information: the reader learns the red is meaningless and stops reading it.

## Which option, and why

The DoR offered two. **Picked option 1: verify the identifier against Linear.**

Option 2 (record an operator-typed external reference with evidence) cannot prove
closure. The operator's assertion would be the only thing standing between a finding
and a clean ledger, which is exactly the hand-clear this command exists to prevent. The
DoR's own line settles it: an external ref that nobody verifies is worse than the bug.

Option 2's *auditability* goal is kept without its cost: every verified resolution now
records which tracker answered, what state it reported, and when, plus an optional
operator note. The record is readable offline; the *decision* was never the operator's.

## How it works

- **Local specs answer first.** A repo tracking its own issues under `.prd-os/issues/`
  keeps resolving offline even when its ids look like Linear keys. Existing behaviour is
  unchanged, including the exact refusal string.
- A `TEAMKEY-123`-shaped ref with **no local spec** is looked up via the Linear API,
  using the same auth path `linear-sync.py` uses: `KIPI_LINEAR_API_KEY`, then the 0600
  `~/.config/kipi/linear-api-key`. Read in `prd_runner.py` rather than imported because
  prd-os ships standalone into repos with no `q-system` tree, so what is shared is the
  **convention** (env name, file path), not code that could drift apart in behaviour.
- Only workflow-state type `completed` counts. **`canceled` is refused** and pointed at
  `--void`: a canceled issue shipped no fix, so clearing an item with it would green the
  gate on work that never happened.
- Linear answers HTTP 200 with an `errors` array for application-level failures, so the
  fetch checks `errors` too. A status-code-only check would read a failed lookup as a
  verified one, which is the exact shape of bug this command exists to prevent.
- Optional `--evidence` records a note (PR, merge commit) alongside a **verified**
  resolution. It is never consulted when deciding.

## Offline / no-auth behaviour: REFUSE

No key, no network, unreachable API, unknown identifier — all refuse. The item stays
open and the gate stays red until someone can actually prove it.

Recording an unverifiable ref as `pending` was the alternative and is rejected: an item
that leaves `open` on an operator's word is a hand-clear with extra steps, and if
`pending` still counted as open it would only be a noisier refusal. Pinned by
`test_resolve_refuses_when_linear_auth_is_missing` (runs with `HOME` pointed at an empty
dir and `KIPI_LINEAR_API_KEY` stripped — no network reached, item still open, gate red).

## How hand-clearing stays impossible

Nothing on the resolve path reads an operator-supplied claim **about closure**. Every
accepted resolution is gated on what a tracker said: a local spec's `status: closed`, or
Linear's workflow-state `type`. `--evidence` is free text but is written only *after*
that gate passes and is never read by it, so it can annotate a proof and cannot become
one. There is still no third way out of the ledger: verified-closed, or `--void` with a
recorded reason.

The Linear fetch is stubbed **at the function boundary** in tests rather than through an
env-var API-URL override, deliberately. An override would be a live bypass surface on the
very path that must not be bypassable; a test seam production also honours is a hand-clear
with extra steps.

## Verification

**Reproducer first.** 7 new tests in `plugins/prd-os/tests/test_spillover.py`; 6 failed
RED before the fix existed:

```
FAILED test_resolve_verifies_closed_linear_issue
FAILED test_resolve_refuses_open_linear_issue
FAILED test_resolve_refuses_canceled_linear_issue
FAILED test_resolve_refuses_unknown_linear_identifier
FAILED test_resolve_refuses_malformed_ref_without_calling_linear
FAILED test_local_issue_spec_still_wins_over_linear
6 failed, 9 passed in 1.43s
E   AttributeError: module 'prd_runner' has no attribute '_linear_issue_state'
```

Green after:

```
$ python3 -m pytest plugins/prd-os/tests/ -q
325 passed, 1 skipped in 11.58s
```

**Acceptance demo, against the live Linear API** (not a stub):

```
$ prd_runner.py --repo-root ~/projects/kipi-system \
    spillover resolve sp-638006cc --resolution-ref ASK-204 \
    --evidence "PR #19 / merge 990d7c1"
{"id": "sp-638006cc", "status": "resolved"}
EXIT=0
```

Ledger record — `resolution_verified_state` came from Linear, not from the caller:

```json
{
  "id": "sp-638006cc",
  "status": "resolved",
  "resolution_ref": "ASK-204",
  "resolution_tracker": "linear",
  "resolution_verified_state": "Done",
  "resolution_verified_at": "2026-07-27T20:01:16Z",
  "resolution_evidence": "PR #19 / merge 990d7c1"
}
```

**Negative self-tests, same live API:**

```
$ ... spillover resolve sp-4c59c1f5 --resolution-ref ASK-209
cannot resolve sp-4c59c1f5: Linear issue 'ASK-209' is not completed (state: Backlog).
Close it first, or record a non-item with --void <reason>.
EXIT=2

$ ... spillover resolve sp-4c59c1f5 --resolution-ref ASK-99999
cannot resolve sp-4c59c1f5: Linear rejected the lookup for ASK-99999:
[{"message": "Entity not found: Issue", ...}]
EXIT=2
```

`plugins/prd-os/.claude-plugin/plugin.json` bumped 0.7.1 -> 0.7.2 (the
`plugin-version-bump` pre-commit gate blocked the first commit; without it the
version-keyed runtime cache would keep serving the stale copy).

## Out of scope, untouched

What creates spillover items, the `deferred`-triage auto-capture, and `gates run`
failure semantics are unchanged. This is only about how an item legitimately leaves.

### DIFF
--- a/plugins/prd-os/.claude-plugin/plugin.json
+++ b/plugins/prd-os/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "prd-os",
-  "version": "0.7.1",
+  "version": "0.7.2",
   "description": "PRD operating system: capture rough ideas, draft reviewable PRDs, run standard and adversarial review (Codex or a Claude senior-staff-engineer subagent, recorded truthfully), triage findings, decompose approved PRDs into atomic issue specs, and execute those issues with scope enforcement, stop-gate receipts, and reviewer-attributed findings.",
   "author": {
     "name": "Assaf Kipnis"
--- a/plugins/prd-os/scripts/prd_runner.py
+++ b/plugins/prd-os/scripts/prd_runner.py
@@ -1039,6 +1039,115 @@ def _issue_is_closed(cfg: Config, issue_id: str) -> bool:
     return False
 
 
+class LinearRefError(Exception):
+    """A resolution reference could not be PROVEN closed.
+
+    Covers every unverifiable case alike — no API key, no network, no such
+    issue, issue still open. They collapse to one class on purpose: the caller's
+    only correct response to any of them is to refuse, so a code path that could
+    tell them apart would only invite one of them being downgraded to a warning.
+    """
+
+
+LINEAR_API_URL = "https://api.linear.app/graphql"
+# Linear identifiers are TEAMKEY-number. Anything else is a local spec id (or a
+# typo), and must never become a live lookup.
+LINEAR_ID_RE = re.compile(r"^[A-Z][A-Z0-9]{0,9}-\d+$")
+LINEAR_STATE_QUERY = "query($id: String!) { issue(id: $id) { state { name type } } }"
+
+
+def _linear_api_key() -> str:
+    """The same auth path linear-sync.py uses: env first, then the 0600 secret.
+
+    Read here rather than imported from linear-sync.py because prd-os ships as a
+    standalone plugin into repos with no q-system tree. What is shared is the
+    CONVENTION (this env name, this file path), not code that could drift.
+    """
+    env = os.environ.get("KIPI_LINEAR_API_KEY", "").strip()
+    if env:
+        return env
+    path = Path(os.path.expanduser("~/.config/kipi/linear-api-key"))
+    if not path.is_file():
+        raise LinearRefError(
+            "closure cannot be verified: no Linear API key. Create one at "
+            "https://linear.app/settings/api then:\n"
+            f"  umask 077 && printf '%s' '<key>' > {path}\n"
+            "or export KIPI_LINEAR_API_KEY. An unverified reference is never recorded")
+    key = path.read_text().strip()
+    if not key:
+        raise LinearRefError(f"closure cannot be verified: {path} is empty")
+    return key
+
+
+def _linear_issue_state(identifier: str) -> dict:
+    """The `{name, type}` of a Linear issue's workflow state.
+
+    Raises LinearRefError for anything short of a definite answer, including an
+    unreachable API — offline is a refusal, not an assumption.
+    """
+    import urllib.error
+    import urllib.request
+
+    body = json.dumps({"query": LINEAR_STATE_QUERY, "variables": {"id": identifier}}).encode("utf-8")
+    request = urllib.request.Request(
+        LINEAR_API_URL, data=body, method="POST",
+        headers={"Content-Type": "application/json", "Authorization": _linear_api_key()},
+    )
+    try:
+        with urllib.request.urlopen(request, timeout=20) as response:
+            payload = json.loads(response.read().decode("utf-8"))
+    except urllib.error.HTTPError as exc:
+        raise LinearRefError(f"Linear returned HTTP {exc.code} for {identifier}") from exc
+    except urllib.error.URLError as exc:
+        raise LinearRefError(f"cannot reach Linear to verify {identifier}: {exc.reason}") from exc
+    except json.JSONDecodeError as exc:
+        raise LinearRefError(f"Linear sent a non-JSON answer for {identifier}: {exc}") from exc
+    # Linear answers HTTP 200 with an `errors` array for application-level
+    # failures, so a status-code-only check would read a failed lookup as a
+    # verified one -- the exact shape of bug this command exists to prevent.
+    if payload.get("errors"):
+        raise LinearRefError(
+            f"Linear rejected the lookup for {identifier}: {json.dumps(payload['errors'])[:200]}")
+    issue = (payload.get("data") or {}).get("issue")
+    if not isinstance(issue, dict) or not isinstance(issue.get("state"), dict):
+        raise LinearRefError(f"Linear has no issue {identifier}")
+    return issue["state"]
+
+
+def _verify_resolution_ref(cfg: Config, ref: str) -> dict:
+    """Prove `ref` names a closed issue; return the evidence that proved it.
+
+    Raises LinearRefError with an operator-readable reason when closure cannot
+    be proven. Nothing here accepts the operator's word: the returned evidence
+    describes what the TRACKER said, which is why a resolution still cannot be
+    hand-flipped through this command.
+
+    Local specs answer first. A repo that tracks its own issues under
+    `.prd-os/issues/` keeps resolving offline even when its ids look like Linear
+    keys; the Linear path only opens for a ref this repo has no spec for.
+    """
+    local_spec = cfg.issues_dir / f"{ref}.md"
+    if local_spec.is_file() or not LINEAR_ID_RE.match(ref):
+        if _issue_is_closed(cfg, ref):
+            return {"resolution_tracker": "prd-os"}
+        raise LinearRefError(
+            f"issue '{ref}' is not closed. Build it through the normal "
+            "reproducer-first issue flow and close it first.")
+    state = _linear_issue_state(ref)
+    if state.get("type") != "completed":
+        # `canceled` lands here too, and should: a canceled issue shipped no fix,
+        # so clearing the item with it would green the gate on work that never
+        # happened. The honest exit for a non-item is --void, which records why.
+        raise LinearRefError(
+            f"Linear issue '{ref}' is not completed (state: {state.get('name')}). "
+            "Close it first, or record a non-item with --void <reason>.")
+    return {
+        "resolution_tracker": "linear",
+        "resolution_verified_state": state.get("name"),
+        "resolution_verified_at": _now_iso(),
+    }
+
+
 def cmd_spillover(cfg: Config, args) -> int:
     """add | list | check | resolve — the out-of-scope finding ledger."""
     import hashlib as _hashlib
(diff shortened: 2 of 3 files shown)


assafkip:
## Review: PR #21 — `spillover resolve` verifies a Linear issue (ASK-209)

Read cold. Ran the suite, mutation-tested the new tests, and reproduced every claim below. Repro scratch: `$TMPDIR/rev21` (repo never touched).

---

### Findings

**1. minor — a hung Linear API escapes as a raw `TimeoutError`, not the designed refusal**

`plugins/prd-os/scripts/prd_runner.py:1101` — the fetch catches `HTTPError`, `URLError`, and `JSONDecodeError`. A read/response timeout is a bare `TimeoutError` from `h.getresponse()`, which `urllib` does **not** wrap in `URLError`. The docstring two lines up says *"Raises LinearRefError for anything short of a definite answer, including an unreachable API."* A hung API is exactly that, and it does not.

Reproducer (local socket that accepts the TCP connection and never answers, standing in for a hung `api.linear.app`; `LINEAR_API_URL` repointed in-process only):

```
python3 - <<'PY'   # full script in $TMPDIR/rev21, 20s wait for the real timeout
... m.LINEAR_API_URL = "http://127.0.0.1:<port>/graphql"
    m.main([... "spillover","resolve","sp1","--resolution-ref","ASK-204"])
PY
```

Real output:

```
--- resolving against a Linear API that accepts the connection and never answers ---
UNCAUGHT EXCEPTION out of main():
  File ".../prd_runner.py", line 1097, in _linear_issue_state
    with urllib.request.urlopen(request, timeout=20) as response:
  ...
TimeoutError: timed out
--- ledger after ---
{"id": "sp1", ..., "status": "open"}
```

Fail direction is safe (item stays open, exit non-zero), which is why this is minor and not major. What the operator loses at 3am is the actionable line `cannot resolve sp1: cannot reach Linear...` — they get a stack trace instead. One-word fix: add `OSError` to the `URLError` clause (`URLError` is already an `OSError` subclass, so it collapses to one handler).

---

**2. minor — a Linear-verified resolution leaks its `verified_state`/`verified_at` onto a later local re-resolve**

`plugins/prd-os/scripts/prd_runner.py:1202` — `new = dict(rec)` carries the prior record forward, and the local branch's evidence dict is only `{"resolution_tracker": "prd-os"}`, so it overwrites the tracker but not the two Linear-only fields.

Real output:

```
after Linear resolve: {... "resolution_ref": "ASK-204", "resolution_tracker": "linear",
                       "resolution_verified_state": "Done", "resolution_verified_at": "2026-07-27T20:14:01Z"}
after LOCAL re-resolve: {... "resolution_ref": "local-fix", "resolution_tracker": "prd-os",
                         "resolution_verified_state": "Done", "resolution_verified_at": "2026-07-27T20:14:01Z"}
>>> tracker='prd-os' but it still carries verified_state='Done' verified_at='2026-07-27T20:14:01Z'
```

The record now asserts a Linear verification timestamp for a resolution Linear never answered. This is the ledger whose entire purpose is being auditable offline, so a self-contradicting row costs exactly what the feature buys. Bounded and recoverable (append-only; the history is still readable). Fix: strip the `resolution_*` keys from `new` before applying evidence.

---

**3. minor — `--evidence` is silently discarded when paired with `--void`**

`plugins/prd-os/scripts/prd_runner.py:1194` — the void branch returns before `args.evidence` is ever read. Exit 0, no warning.

Real output:

```
exit: 0 record: {"id": "sp2", ..., "void_reason": "not a real item", "resolved_at": "..."}
>>> resolution_evidence present? False
```

`--void` is where an operator most wants to attach a pointer ("dupe of PR #19"), and the help text scopes evidence to a *verified* resolution without saying it is dropped otherwise. Either record it or refuse the combination; silently accepting and dropping is the one option that misleads.

---

**4. minor — the unknown-identifier stub emits an error string the real API never produces, and the comment claims it does**

`plugins/prd-os/tests/test_spillover.py:164` — `_stub_linear`'s comment: *"Unknown identifier -> the same error the real fetch raises."* It is not. The stub raises from the `data.issue is null` branch; the live API (per this PR's own pasted run for `ASK-99999`) returns HTTP 200 with an `errors` array, which exits through the `errors` branch instead.

Reproducer drives the real `_linear_issue_state` with the payload shape from the PR's own live output:

```
REAL fetch raises  : Linear rejected the lookup for ASK-99999: [{"message": "Entity not found: Issue"}]
TEST stub raises   : Linear has no issue ASK-99999
```

Behaviour is identical (both refuse), so this is a false comment, not a bug. It matters because it points at the coverage hole underneath it: `_linear_issue_state` is stubbed in every test, so its body has **zero** coverage. Mutation-testing the suite:

```
A accept-any-linear-state          exit=1   2 failed, 13 passed     <- caught
D kill local-spec-first routing    exit=1   3 failed, 12 passed     <- caught
B no-key-means-empty-key           exit=0   15 passed               <- SURVIVED
C ignore-graphql-errors-array      exit=0   15 passed               <- SURVIVED
```

Mutation C deletes the `errors`-array check — the single behaviour the PR body argues hardest for ("a status-code-only check would read a failed lookup as a verified one") — and the suite stays green. I traced it: with the check gone, an unknown id still refuses via `data.issue is null`, so it is defence-in-depth rather than an exploitable hole. But the claim is unpinned. One test feeding a canned payload through `_linear_issue_state` with `urlopen` stubbed would kill B and C both.

---

### What is sound

- **Hand-clear the gate.** No path writes `status: resolved` from an operator assertion. `--evidence` is written strictly after the tracker gate and is never read by it. Mutation A (drop the `completed` check) and mutation D (kill local-spec-first routing) both go red — those two tests bite.
- **`canceled` clearing the gate.** Correctly refused and pointed at `--void`. The state-type vocabulary (`completed`/`canceled`/`started`) matches Linear's real workflow-state types, and the PR's live run confirms `Done`/`Backlog` round-trip.
- **Offline / no-key = silent pass.** Refuses; item stays open; `gates run` stays red. Verified the auth test is genuinely hermetic — `_linear_api_key()` raises before the request is built, and the whole file runs in 2.98s, so nothing reaches the network.
- **Key exfiltration via a redirected endpoint.** `LINEAR_API_URL` is a module constant with no env override — deliberately *unlike* `linear-sync.py`, which honours `KIPI_LINEAR_API_URL`. Right call on a credential-bearing path.
- **Auth-convention drift.** Checked against `q-system/.q-system/scripts/linear-sync.py:336-371`: same env name, same `~/.config/kipi/linear-api-key` fallback, same raw (non-`Bearer`) `Authorization` header, same 200-with-`errors` handling. The "shared convention, not shared code" claim holds.
- **GraphQL injection.** The ref goes through a bound `$id` variable, never string-interpolated.
- **Blast radius.** Read-only against Linear. Nothing is created, commented, or mutated on a Linear object. Straight code revert rolls it back; the append-only ledger keeps prior rows readable either way.
- **Routing collision between local and Linear ids.** Real `.prd-os/issues/` ids in this repo are lowercase slugs (`ath-durable-recovery.md`), which cannot match `^[A-Z][A-Z0-9]{0,9}-\d+$`. The two namespaces do not overlap in practice. (The `test_local_issue_spec_still_wins_over_linear` fixture uses `ASK-204` as a *local* spec id, a shape no producer here emits — but it is guarding a defensive branch, so it costs nothing.)
- **Path traversal via `--resolution-ref ../../x`.** Reachable, but byte-identical to pre-PR `_issue_is_closed` behaviour and operator-supplied. Not this PR's.
- **Full plugin suite.** 319 passed / 6 skipped. The single failure I saw (`test_execution_discipline_wiring.py::test_kipi_core_no_longer_wires_the_lint`) was an artifact of my extracting only `plugins/prd-os`; it passes 4/4 in the full tree.

---

### VERDICT: **APPROVE WITH NITS**

Four minors, no majors, no blockers. Every failure mode I could construct fails toward *refuse and stay red*, which is the correct direction for this command, and nothing here can write to a Linear object or destroy founder work.

**Fix first:** finding 1. Add `OSError` to the `URLError` handler at `prd_runner.py:1101`. It is the one that changes what a human sees at 3am, and it is the one case where the docstring currently promises something the code does not deliver.

FINDINGS:
minor|A read timeout on the Linear API escapes as an uncaught TimeoutError instead of the documented LinearRefError refusal|plugins/prd-os/scripts/prd_runner.py:1101
minor|A Linear-verified resolution leaks resolution_verified_state/at onto a later local re-resolve, producing a self-contradicting ledger row|plugins/prd-os/scripts/prd_runner.py:1202
minor|--evidence is silently discarded when paired with --void, exiting 0 with no warning|plugins/prd-os/scripts/prd_runner.py:1194
minor|The unknown-identifier stub raises an error string the live API never produces, and _linear_issue_state has zero test coverage (errors-array mutation survives)|plugins/prd-os/tests/test_spillover.py:164
END FINDINGS

PR #22, opened 2026-07-27
fix(loop): the worker fetches, sees mergeability, keeps operator directives, and cannot commit its own scratch (ASK-208)

assafkip:
Closes the four autonomous-loop defects in ASK-208, all surfaced live on 2026-07-27 while shipping PRs #11-#20. Resolves `sp-28ced3d6`, `sp-71b63e62`, `sp-fd76af2f`, `sp-1aae7516`.

Scoped to `linear-worker.sh`, `converge.sh` and `pr-verdict-lib.sh` as the DoR requires, plus three new test files and their capability-manifest lines.

## 1. The worker never fetched (sp-28ced3d6)

`linear-worker.sh` contained no `git fetch` anywhere. It cut every worktree from whatever local `origin/main` ref happened to exist.

**Observed:** ASK-150 was dispatched to resolve a conflict against `main`, merged `3b60af0`, and the conflict survived because `main` was already `72c782d`.

One fetch per run, placed above the picker so no code path can create a worktree first. A fetch failure is environmental (self-healing-retry rule 5): it stops on attempt 1 and burns no issue's retry budget.

## 2. The rework gate ignored mergeability (sp-71b63e62)

`rework_gate` decided "is there work to do" from the stored verdict alone, so a PR approved earlier that **later stopped being mergeable** was invisible.

**Observed:** PR #11 approved 06:08Z; #16 landed 17:30Z and broke it; both `converge` and a direct worker run skipped it in under 2 seconds as "waiting on founder merge only".

`rework_gate <verdict> [mergeable]` now treats approved + `CONFLICTING` as rework. Only a **stated** `CONFLICTING` counts: `UNKNOWN` (GitHub still computing) and empty (gh failed) keep the old behaviour, because a wrong refusal here stalls every instance's worker at once. Both callers pass it, read once through `pr_mergeable` so there is one reader of that state. When the conflict is the reason, the rework prompt says so first instead of claiming the review is the spec.

## 3. An operator directive died after one round (sp-fd76af2f)

The rework spec was whatever the newest review `.md` said, and the reviewer regenerates findings every round.

**Observed:** a verdict scoped to *"resolve the merge conflict, nothing else"* was overwritten; rounds 1 and 2 both did code polish while the conflict went untouched.

Directives now live at `$KIPI_STATE_DIR/directives/<issue>.md`, render **verbatim** in every prompt, and sit **alongside** the review rather than replacing it. Cleared only by deleting the file: no review round, successful run, or verdict retires one. Deliberately not a priority system.

## 4. The worker committed its own scratch (sp-1aae7516)

Seven working files reached `main` through PRs #18 and #19 before the #20 sweep, in three consecutive PRs.

The worker now installs a **per-worktree** `pre-commit` guard via `git config --worktree core.hooksPath`, so the founder's checkout is untouched (linked worktrees share `.git/hooks`, which is why a plain hook install was not an option). It refuses script/draft files dropped at the **root** of `q-system/output/` and points at a scratch dir outside the repo. It **chains** to the repo's existing `pre-commit` instead of replacing it, so gitleaks, blocked-paths and the commit-msg linear gate keep running.

The shape is derived from the real payload, not invented: all seven were `.py`/`.sh`/`.txt` at that root, and no such file has ever been legitimately tracked there (the 24 tracked top-level files are `.md`/`.patch`/`.out`/`.err`). Files inside a named subdir (`rca/`, `plans/`, `architecture-report/`) are deliverables and are not matched.

## Reproducers: RED observed before each fix, GREEN after

| Test | RED (before the fix) | GREEN (after) |
|---|---|---|
| `test-linear-worker-fetch.sh` | `FAIL: STALE BASE: the worktree was cut from the stale origin/main (f16dc03), true remote head 53cd72e` | `PASS: linear-worker fetch (4 checks)` |
| `test-rework-gate-mergeability.sh` | `FAIL: rework_gate 'APPROVE WITH NITS' 'CONFLICTING' -> 10, want 0` | `PASS: rework gate mergeability (15 checks)` |
| `test-linear-worker-directive-scratch.sh` | `FAIL: directive lost between rounds` | `PASS: worker directive + scratch guard (11 checks)` |

Each drives the real script rather than asserting on its source: the fetch bug **is** the SHA the worktree landed on, and the directive bug **is** what the agent was actually told, so the tests read back the real base SHA and the real captured prompts.

**No regressions:** `test-linear-worker-parallel.sh` 7/7, `test-converge.sh` 17/17.

**Wrong-refusal risk, asserted at scale:** case 6 sweeps all **1442 paths tracked on main** through the real installed hook and requires zero refusals; case 5 pins that an RCA, a top-level `.md` and a real harness still commit; case 7 pins the chaining (a planted `exit 9` pre-commit must still run); case 8 pins that the founder's checkout keeps `core.hooksPath` unset.

## One thing my own reproducer caught

The first version of the guard baked the delegate hook path in at install time. Case 7 failed: a hook added after the worktree was created would have been silently skipped. The guard now resolves the delegate at runtime, so the day gitleaks is added is not the day it silently does not run.

## Blast radius

`q-system/.q-system/scripts/` syncs to every instance via `kipi update`, so this is the fleet's autonomous loop. Fixes 1 and 2 change **when** work is dispatched; fix 4 changes **when** a commit is refused.

Do not merge without review.


assafkip:
Check results, posted after opening the PR.

**New reproducers (each RED before its fix, green after)**
- `test-linear-worker-fetch.sh` -> `PASS: linear-worker fetch (4 checks)`
- `test-rework-gate-mergeability.sh` -> `PASS: rework gate mergeability (15 checks)`
- `test-linear-worker-directive-scratch.sh` -> `PASS: worker directive + scratch guard (11 checks)`

**No regressions on the scripts this touches**
- `test-linear-worker-parallel.sh` -> `PASS: linear-worker parallel (7 checks)`
- `test-converge.sh` -> `PASS: 17/17 converge checks`

**validate-separation.py 1 --verbose**: PASS 68, FAIL 1, WARN 1. The single FAIL is the capability gate, whose only RED is `test-updater-issue-sequence.py` hitting the gate's 60s per-test timeout. That test passes when run standalone: `PASS: updater issues landed in order and every prior check still holds`. It drives 7+ kipi-update sub-suites serially, so it outruns the budget rather than failing. Pre-existing and unrelated to this diff; captured as `sp-e0740c76` rather than fixed here, since the DoR scopes this issue to three files.

Inside the capability gate itself the run was `ran=74 quarantined=0` with 1 RED, and it was that one. All three new tests are declared in `capability-manifest.json` and ran green there.

The DoR's stated form of the command (`CI=true CAPABILITY_GATE_SKIP=1`) skips exactly that gate, so it exits 0; I could not set env-prefixed vars from this sandbox, so the numbers above are from the unskipped run, which is the stricter one.

assafkip:
Review complete. Repro scratch lives in the gitignored `template-repo/pr22-review/` (working tree unmodified, `git status` clean); remove with `git worktree remove template-repo/pr22-review/tree && rm -rf template-repo/`.

---

# PR #22 — adversarial review

## Findings

### 1. major — a `git fetch` failure stops the entire run, does zero work, exits 0, and pages nobody

`q-system/.q-system/scripts/linear-worker.sh:246`

The new fetch guard is the first thing in the script and it gates every issue. Its failure path is `say` + `exit 0`. `$NOTIFY` is in scope (used 340 lines later for "stuck after 3 attempts") and is not called. `MAX_ATTEMPTS` is never bumped, so the issue never becomes stuck and never pages that way either. The header comment cites self-healing-retry rule 5, which says *surface it immediately*; a line in `~/.config/kipi/linear-worker.log` is not surfacing.

An expired git credential or an unreachable origin at 3am produces a worker that is byte-for-byte indistinguishable from a healthy no-work run.

```
bash template-repo/pr22-review/repro-C-fetch-silent.sh
```

```
=== worker exit code with an unreachable origin ===
  rc = 0   <-- SUCCESS. Indistinguishable from a healthy no-op run.

=== work actually performed ===
  rework/work rounds dispatched: 0

=== did anything page the founder? (slack-notify.sh recorder) ===
  (nothing. zero Slack notifications.)

=== the only trace, stdout+log ===
2026-07-27T20:24:29Z INFRA: git fetch failed in /var/.../skel. Stopping before any worktree is cut from a stale base.

=== for contrast: the SAME script's healthy no-work run ===
  rc = 0  (healthy, nothing ready)
```

Honest mitigation: `converge.sh` catches it downstream (no PR → exit 7 + Slack). The batch path (`kipi work --apply --limit N`) does not. This script syncs to every instance via `kipi update`, so the silent-stop ships fleet-wide.

Fix is one line next to the existing `say`: `bash "$NOTIFY" "worker: git fetch failed in $SKEL, run did no work" 2>/dev/null || true`, and a non-zero exit so a caller can tell the difference.

---

### 2. minor — test section C passes for the wrong reason and cannot fail

`q-system/.q-system/scripts/test/test-rework-gate-mergeability.sh:124-142`

Section C claims to prove "an approved AND mergeable PR is still left alone for the founder" — the guard against this fix turning every approved PR into a rework loop. It writes the verdict record to `$WORK/state/pr-reviews/pr-778.verdict.json` and then runs the worker with `KIPI_STATE_DIR="$WORK/state-ok"`. The worker reads a different directory, finds no verdict, and skips at gate **20 (unreviewed)** — never reaching the `APPROVE|APPROVE WITH NITS` branch at all. `[ ! -s worked.txt ]` would hold no matter what that branch did.

```
bash template-repo/pr22-review/repro-A-vacuous-case-C.sh
```

```
=== the shipped test's own assertion ===
  ASSERTION PASSES (worked.txt empty)

=== but WHY did it skip? the worker's own log line ===
2026-07-27T20:21:07Z skip ASK-AAA: PR #778 has no recorded review verdict -- run: kipi review 778 --issue ASK-AAA --post

=== did the verdict record the test wrote even exist where the worker looks? ===
  test wrote   : $WORK/state/pr-reviews/pr-778.verdict.json  -> EXISTS
  worker reads : $WORK/state-ok/pr-reviews/pr-778.verdict.json -> MISSING
```

Not major: the unit-level `gate_is 10 "APPROVE" "MERGEABLE"` does pin the semantics. Only the end-to-end half is decoration. One-char fix: write the record into `$WORK/state-ok/pr-reviews/`.

---

### 3. minor — the operator-directive feature ships a reader and no writer

`q-system/.q-system/scripts/linear-worker.sh:62`

`$KIPI_STATE_DIR/directives/<issue>.md` is a new state file. Nothing in the repo produces it, the `kipi` CLI has zero occurrences of "directive", and `DIRECTIVES_DIR` is the one path constant never passed to `mkdir -p` (lines 85, 149, 380, 537 cover the others). This repo's own `wiring-check.md` says a new state file needs both a producer and a consumer.

```
bash template-repo/pr22-review/repro-D-directive-no-producer.sh
```

```
=== every reference to the directives path in the whole PR head ===
  .../test/test-linear-worker-directive-scratch.sh:49,86,124   (the test)
  .../scripts/linear-worker.sh:58,62                            (the reader)

=== does the kipi CLI expose any way to set one? ===
  75:  review)
  81:  converge)
  94:  work)
  occurrences of "directive" in the kipi CLI: 0

=== is the directory even created for the operator to drop a file in? ===
  85:mkdir -p "$STATE_DIR"
  149:  mkdir -p "$hooks" "$scratch"
  380:    mkdir -p "$(dirname "$TREE")"
  537:  mkdir -p "$SCRATCH"
  -> DIRECTIVES_DIR is defined at line 62 and never passed to mkdir.
```

The fix for sp-fd76af2f works, but only for someone who has read `linear-worker.sh:515` and infers the path convention by hand. `kipi directive <issue> [--clear]` plus a `mkdir -p "$DIRECTIVES_DIR"` closes it.

---

### 4. minor — `APPROVE` stops being terminal, and nothing caps the rework rounds on the direct worker path

`q-system/.q-system/scripts/pr-verdict-lib.sh:108-109`

Before this PR, `APPROVE` was terminal for the worker. Now an approved PR that GitHub reports `CONFLICTING` re-enters rework. `MAX_ATTEMPTS` only counts runs where `claude` exits **non-zero** — and the PR's own cited failure mode is an agent that exits 0 having done the wrong thing ("rounds 1 and 2 both did code polish while the conflict went untouched"). The `rounds` counter is written and never read as a gate.

```
bash template-repo/pr22-review/repro-B-unbounded-rework.sh
```

```
after scheduled run 1: rework rounds dispatched = 1
...
after scheduled run 6: rework rounds dispatched = 6

=== totals over 6 scheduled runs on ONE approved+CONFLICTING PR ===
  claude -p rework rounds dispatched : 6   (each bounded at TIMEOUT_SECONDS=1800)
  progress notes on the Linear issue : 18   (permanent, cannot be deleted)

=== is there a cap? the attempts ledger the worker keeps ===
{ "ASK-AAA": { "rounds": 6 } }

=== did the worker ever refuse? ===
  refusals: 0
```

Calibration, stated plainly: `converge.sh` **does** bound this (MAX_ROUNDS=4 plus the exit-5 same-verdict-same-sha stall check), so the uncapped path is repeated `kipi work --apply --issue X`, which is attended. That is why this is minor and not major. The residue that is real: 3 permanent Linear comments per round on an object that cannot be deleted.

---

### 5. nit — the passthrough mirror copies a file git will never invoke

`q-system/.q-system/scripts/linear-worker.sh:155`

The exclusion list is `*.sample|*.old|*.bak|pre-commit`. Against this repo's real `.git/hooks`, that lets `pre-commit.before-gitleaks` through into every agent worktree's hooks dir as a dead file. Harmless; an allowlist of git's actual hook names would be tighter than a denylist of suffixes.

```
bash template-repo/pr22-review/repro-E-mirror-list.sh
```

```
  MIRRORED (git will invoke it) : commit-msg
  MIRRORED (git will invoke it) : post-commit
  MIRRORED (git NEVER invokes it, dead file) : pre-commit.before-gitleaks
```

---

- **`git rev-parse --git-common-dir` returning a relative path.** Both `install_scratch_guard` and the generated hook string-concatenate its output; a relative result would have silently skipped the mirror and the gitleaks delegation. Ran it from both the worker's `-C $tree` position and a hook's cwd on git 2.54.0: absolute in both. Dropped.
- **The scratch-guard payload claim.** `git show cf14292 --stat` — all seven swept files are `.py`/`.sh`/`.txt` at the root of `q-system/output/`. The regex is derived from the real payload, exactly as claimed, not invented.
- **The scale negative self-test is genuine.** Case 6 pipes all 1445 paths tracked at PR head through the real installed hook; zero false positives. Ran it, saw the count.
- **The chaining fix is real, and the author's own case 7 found it.** Planting `exit 9` at the common `pre-commit` does block the commit — the delegate is resolved at hook-run time, not baked in at install.
- **Test isolation holds.** A stray `~/.config/kipi/worktrees/ask-aaa` + `sana/ask-aaa` branch exists in the founder's checkout, but its reflog dates it to `2026-07-27 01:05` — before this PR's commits — and running all three new tests today added nothing (`git status` clean, reflog unchanged). Pre-existing leak from an earlier session, not this PR.
- **Both no-regression claims verified, not taken on the author's word.** `test-linear-worker-parallel.sh` → `PASS (7 checks)`, `test-converge.sh` → `PASS: 17/17`.
- **`UNKNOWN` / empty mergeability is correctly excluded.** GitHub returns `UNKNOWN` while it computes; treating that as a conflict would manufacture rework rounds fleet-wide. The gate requires a stated `CONFLICTING` and the unit cases pin all ten (verdict × mergeability) pairs. This is the right side to fail on.
- **Prompt injection via the directive file.** `$(cat "$DIRECTIVE_FILE")` result is not re-evaluated, and `$PROMPT` reaches `claude` as a positional `$1`. No shell escape.

---

## VERDICT: REQUEST CHANGES

One major finding, so by the rule this is REQUEST CHANGES — not BLOCK. Nothing here is permanent or unrecoverable, and the four fixes themselves are the right shape: each one is a code refusal rather than an instruction, each drives the real script rather than grepping its source, and the guard's negative case is asserted harder than its positive case.

**Fix first: `linear-worker.sh:246`.** Route the fetch failure through `$NOTIFY` and exit non-zero. Everything else on this list can ride a follow-up; a fleet-wide worker that goes permanently dark while reporting success cannot.

FINDINGS:
major|A git fetch failure stops the whole worker run with exit 0, does zero work, and sends no Slack notification, so a dead credential is indistinguishable from a healthy no-op run|q-system/.q-system/scripts/linear-worker.sh:246
minor|Test section C runs the worker against a different KIPI_STATE_DIR than the one it wrote the verdict record to, so it skips as unreviewed and can never fail|q-system/.q-system/scripts/test/test-rework-gate-mergeability.sh:140
minor|The operator-directive state file has a reader but no producer, no kipi CLI surface, and its directory is never created|q-system/.q-system/scripts/linear-worker.sh:62
minor|APPROVE stops being a terminal state and no round cap bounds rework on the direct worker path, so an unresolvable conflict yields unbounded rounds and permanent Linear comments|q-system/.q-system/scripts/pr-verdict-lib.sh:108
nit|The hook passthrough mirror uses a suffix denylist and copies pre-commit.before-gitleaks, a file git never invokes, into every agent worktree|q-system/.q-system/scripts/linear-worker.sh:155
END FINDINGS

assafkip:
Round-3 rework pushed as `5600eba` + `abeabb8`. Every finding is answered below; all five are fixed, none disputed. Each fix has a reproducer observed RED first, and the real command output is quoted.

## 1. major -- a fetch failure goes dark. FIXED.

`linear-worker.sh` now pages through `$NOTIFY` and exits **9**, and `$NOTIFY` becomes `${KIPI_NOTIFY:-...}` -- the same seam `converge.sh` already had, so the test reads back *what* was paged instead of grepping the source.

RED (before), driving the real worker against a genuinely unreachable origin:

```
FAIL: SILENT SUCCESS: the worker exited 0 after a fetch failure that stopped
      the entire run. A caller -- launchd, a wrapper, converge -- cannot tell
      this apart from a healthy run with nothing ready.
rc=0
```

GREEN: `PASS: worker infra exit (10 checks)`.

**Layer above, and one thing I made quieter on purpose.** `converge.sh` already caught this downstream as "no PR -> exit 7 + Slack". Left alone, the founder would now get **two** Slack messages for one cause, and converge's message would be false -- it blames Sana for not opening a PR when she was never dispatched. So converge maps `rc=9` to its own exit 7, names the infra cause, and **does not page again**. That is a deliberate reduction: converge is now silent on this one event, because the worker owns the page for its own environment. Net across both paths goes from 1 page (converge only, wrong cause) to 1 page (worker, right cause) plus coverage of the batch path that had zero. Case 6 of the reproducer asserts the healthy no-work run still exits 0 and pages nobody, so the fix cannot pass by always failing.

**The reproducer found a second defect while proving the first.** `converge.sh` never created `$STATE_DIR`, and its `>>"$LOG"` redirect is evaluated by converge's shell *before* the worker starts. On a state dir that does not exist yet, bash returns 1 and flattens **every** worker exit code -- including the new 9. The first GREEN attempt failed with `worker rc=1`, not 9. One line: `mkdir -p "$STATE_DIR"`.

## 2. minor -- test section C passes for the wrong reason. FIXED, and I verified your claim rather than assuming it.

Mutated the worker to ignore gate 10 entirely (the exact bug section C claims to guard) and ran the **shipped** test:

```
=== MUTANT worker (approved+mergeable IS reworked) vs the CURRENT test ===
test rc = 0
['  ok: an approved AND mergeable PR is still left alone for the founder',
 'PASS: rework gate mergeability (15 checks)']
```

Vacuous, exactly as stated.

I took the one-char path plus one more thing, because the path fix alone leaves the assertion still unable to distinguish *why* it declined. An absence-of-work assertion passes for any reason the worker skips. So section C now also pins **which gate fired**:

```bash
grep -q "nothing to rework, waiting on founder merge" "$WORK/run-ok.out"
```

Same mutant, fixed test: `rc = 1`. Real code: `PASS: rework gate mergeability (16 checks)`. (The assertion that fires on the mutant is the new reason-pin, not the `worked.txt` one -- so the reason-pin is doing the catching, which is the point.)

## 3. minor -- a reader with no writer. FIXED, with a scope note.

Added `kipi directive <ASK-nnn> "<text>" | --clear | (no args = show)`, writing the exact path the worker reads from the same `KIPI_STATE_DIR` env var with the same default. Worker now `mkdir -p`s `DIRECTIVES_DIR` (it was the one path constant never created), and the log line says `retire it with: kipi directive <issue> --clear` instead of telling the operator to delete a file they would have to find by reading source.

**Both ends are pinned in one test**, because this is a cross-process handshake and a test that pins one end passes happily while the two disagree about the path:

```
ok: kipi directive writes the file the worker reads (producer exists)
ok: a directive set through the CLI reaches the agent's prompt verbatim
ok: kipi directive --clear retires it, and the next prompt is clean
ok: the worker creates the directives directory it reads from
```

Scope note, stated rather than buried: this touches `kipi`, a fourth file, and the DoR says the issue is scoped to three. I judged `wiring-check.md` ("a new state file needs both a producer and a consumer") to outrank that line for a feature this issue itself introduced -- shipping the reader without the writer is shipping a dead feature. Flagging it so the call is yours to reverse, not mine to hide.

## 4. minor -- APPROVE stops being terminal with no cap. FIXED.

Reproduced your measurement first:

```
FAIL: UNBOUNDED: 6 scheduled runs on ONE approved-but-CONFLICTING PR dispatched
      6 rework rounds against a cap of 3. Each is a bounded claude run
      and 3 permanent Linear comments; nothing ever refuses.
```

`MAX_ROUNDS=6` (`KIPI_MAX_ROUNDS`-overridable), read from the `rounds` counter that was previously write-only. Three decisions worth stating:

- **A cap, not a stall check.** I started on a stall check mirroring converge's exit-5, then dropped it: exit-5 needs an unchanged head sha, and your own cited mode is an agent that *pushes* every round (code polish while the conflict survives). The sha moves, so a stall check never fires on that shape. Only a ceiling bounds it.
- **Checked above the dry branch, not down at the verdict gate.** It needs no network, so `kipi work` (dry) reports it -- it used to print `would work ASK-x` for an issue it would in fact refuse, which is the fix-landed-on-the-detector-not-the-report failure. And it fires before the claim and before the "Picked up" note, so **a refusal costs zero permanent Linear comments**. Asserted: `9 [notes] for 3 dispatched rounds`, no additional notes for the 3 refused runs.
- **6, above converge's default 4**, so the attended driver still hits its own cap first and this one only binds the unattended `kipi work` path.

**The cap does not buy silence.** It pages **once, on the transition** (`capped_notified` in the ledger), not once per refused run -- a capped issue pinging every scheduled run is a channel that stops being read. And it is resettable: `kipi work --reset-rounds --issue X`, which runs before the fetch so an operator unwedging an issue does not need the network to be healthy. That is deliberate: a counter that only grows would make the worker go permanently dark on an issue right after the operator did the right thing, which is the round-2 regression on this very PR. Cases 5 and 6 pin the reset and pin that the cap is per-issue.

GREEN: `PASS: worker round cap (8 checks)`.

## 5. nit -- the mirror copies a file git never invokes. FIXED.

Suffix denylist replaced with an allowlist of git's actual hook names. Asserted both directions, since an allowlist can fail the other way:

```
ok: a backup file next to a real hook is not mirrored
ok: a real git hook (post-checkout) is still mirrored
```

## Regression set -- nothing you said you could not break got quieter

```
test-linear-worker-fetch                     rc=0  PASS: linear-worker fetch (4 checks)
test-rework-gate-mergeability                rc=0  PASS: rework gate mergeability (16 checks)
test-linear-worker-directive-scratch         rc=0  PASS: worker directive + scratch guard (17 checks)
test-linear-worker-infra-exit                rc=0  PASS: worker infra exit (10 checks)
test-linear-worker-round-cap                 rc=0  PASS: worker round cap (8 checks)
test-linear-worker-parallel                  rc=0  PASS: linear-worker parallel (7 checks)
test-converge                                rc=0  PASS: 17/17 converge checks
```

The scale negative self-test you verified by hand still sweeps the full tracked list: `ok: the guard refuses none of the 1446 paths tracked on main`. The chaining case, the per-worktree `core.hooksPath` isolation, and all ten (verdict x mergeability) unit pairs are untouched and green.

`python3 validate-separation.py 1 --verbose` -> **PASS: 69, FAIL: 0, WARN: 1**. The capability-gate FAIL from round 2 (`test-updater-issue-sequence.py` outrunning the 60s per-test budget) is gone from this run; it remains captured as `sp-e0740c76` rather than fixed here.

One unrelated fix rode along because the reproducer tripped it: `test-linear-worker-fetch.sh` used `grep -c ... || echo 0`, and `grep -c` exits 1 on a zero count, so the branch that reports "no fetch at all" died on a bad integer comparison instead of failing cleanly. Swapped for `wc -l`.

assafkip:
# Review round 2, PR #22 (`sana/ask-208`)

Read the diff and the code before the prior rounds, as instructed. Both findings below are in code that **did not exist** in round 1. Neither re-raises a settled finding, neither escalates one.

---

## 1. major: `--reset-rounds` reports success unconditionally, so the only way out of the new round cap silently no-ops

`q-system/.q-system/scripts/linear-worker.sh:140`

The reset writes the ledger with `>/dev/null 2>&1 || true`, then unconditionally says "reset to 0; the next run will dispatch again" and exits 0. It never validates that the issue key exists, never normalizes case, and never reads back what it wrote. `kipi directive`, added in this same PR, **does** normalize case (`tr 'a-z' 'A-Z'`), so lowercase issue ids are an input shape this PR itself accepts.

```
$ cd sites/rev22
$ echo '{"ASK-208":{"count":0,"rounds":6}}' > state/linear-worker-attempts.json
$ KIPI_STATE_DIR=.../state bash lw.sh --reset-rounds --issue ask-208
2026-07-27T20:52:06Z ask-208: review round count reset to 0; the next run will dispatch again
exit=0
--- AFTER ---
{
  "ASK-208": { "count": 0, "rounds": 6 },
  "ask-208": { "rounds": 0 }
}
```

Second repro, ledger not writable (`chmod 444`), correct-case id:

```
$ KIPI_STATE_DIR=.../state bash lw.sh --reset-rounds --issue ASK-208
2026-07-27T20:49:27Z ASK-208: review round count reset to 0; the next run will dispatch again
=== EXIT 0 ===
--- attempts file AFTER the claimed reset ---
{"ASK-208":{"count":0,"rounds":6}}
```

**Blast radius at 3am.** The worker pages once on the cap transition, then `mark_capped` sets `capped_notified: true` and it never pages again (`linear-worker.sh:96` cap, `mark_capped`). The operator runs the reset, is told it worked, and walks away. `ASK-208` stays at `rounds: 6`, keeps `capped_notified: true`, and goes permanently dark with no further signal. Cleanup is hand-editing `~/.config/kipi/linear-worker-attempts.json` to drop the phantom key and zero the real one, after first diagnosing why an issue that reported "reset" never resumed.

`test-linear-worker-round-cap.sh:161` covers the happy path only (writable ledger, correct case), so the suite is green and cannot see this.

**Fix shape:** drop `2>&1 || true` on the reset, `tr 'a-z' 'A-Z'` the id the way `kipi directive` does, and print the value read back (`was` is already computed and thrown away).

---

## 2. minor: the new `kipi directive` help line is unquoted, so `kipi help` errors, exits 1, and drops both its own line and the pre-existing `kipi dor` line

`kipi:37`

```
  echo "  kipi directive <ASK-n> "<text>"  Standing instruction ...
  kipi dor [--apply]   Draft a Definition of Ready ..."
```

Bash closes the string at `"<`, so `<text>` parses as `< text > "  Standing instruction...(bounded)"`. `bash -n` passes; the damage is semantic.

```
$ bash ../kipi_head.sh help
...
  kipi converge --issue <ASK-n> [--max-rounds N]  Drive one issue to an APPROVED PR (work+review+rework, capped)
../kipi_head.sh: line 37: text: No such file or directory
=== EXIT: 1 ===
```

Both `kipi directive` and the pre-existing `kipi dor` are gone from the output. `usage` is the `*` fallthrough for every unrecognized subcommand, so any typo now prints an error and exits 1.

Amplification, when a file named `text` happens to exist in cwd the redirection succeeds and creates a junk file:

```
$ echo dummy > text && bash ../kipi_head.sh help; ls -1b
=== EXIT 0 ===
  Standing instruction the worker renders into EVERY rework prompt (--clear to retire)\n  kipi dor [--apply]   Draft a Definition of Ready onto issues that lack one (bounded)
text
```

`kipi directive` itself works (set / show / clear all verified, case-normalizing). It is just undiscoverable: broken in `usage`, and `grep -rn "kipi directive" CLAUDE.md q-system/CLAUDE.md .claude/rules/` returns nothing. `wiring-check.md` asks that a new command be registered.

Kept at minor: no unattended consequence, no automated caller (`grep -rn "kipi help"` finds only a settings allowlist entry).

---

## 3. minor: `git commit --no-verify` bypasses the scratch guard, and the prompt tells the agent the mechanism

`q-system/.q-system/scripts/linear-worker.sh:193` (guard), `:666` (the prompt line that names it)

From my end-to-end repro, immediately after the guard correctly refused:

```
=== CASE B: the SAME staged scratch file with --no-verify ===
  -> commit count: 2
  -> tracked now:  q-system/output/ask999-helper.py q-system/output/rca/rca-x.md
```

Every git hook has this property, so this is not an argument against the design. The specific new thing is that the prompt at `:666` tells the agent "a commit hook in this worktree refuses them", which is the standard cue for a model to reach for `-n`. Naming the constraint without naming the mechanism costs nothing.

---

## What is sound: attacks I ran that the code survived

- **Per-worktree `hooksPath` isolation holds.** Built a synthetic repo with a linked worktree, ran the real `install_scratch_guard`. Agent tree gets `core.hooksPath`; founder checkout stays `[]`. Confirmed live too: `git config --get core.hooksPath` in kipi-system returns rc=1 (unset) with a real `ask-208` worktree present. Case F: the founder committed `q-system/output/founder-thing.py` with no refusal.
- **Chaining is real, not claimed.** Case C/D: both the delegate `pre-commit` and `commit-msg` fired on legitimate commits. The guard adds a gate without subtracting one. Runtime delegate resolution works.
- **The regex is derived from the real payload, not invented.** Checked PR #20's actual sweep: all seven files are `.py`/`.txt`/`.sh` at the root of `q-system/output/`. The claim survives.
- **`.md` scratch at that root is not caught** (Case D), but `*.json`/`*.html`/`*.log` there are already gitignored, so `.md` is the only hole and it is a stated scope call. Not raising it.
- **`extensions.worktreeConfig`** is written to the shared `.git/config`. Git's docs warn this requires moving `core.bare`/`core.worktree`. Checked live: `core.bare=false`, `core.worktree` unset. No hazard. Dropped.
- **Test isolation is correct.** `test-linear-worker-round-cap.sh` sets `HOME`, `KIPI_SKEL`, `KIPI_STATE_DIR` to its own `mktemp -d` and stubs `python3`/`gh`/`claude`. I found a stray `sana/ask-aaa` worktree in the live checkout, tried to attribute it to this test, and could not. Dropped.
- **Fail-open on `UNKNOWN` mergeability.** converge reads `gh pr view` seconds after a push, when GitHub often has not computed mergeability yet, so an approved-but-conflicting PR could be reported "ready to merge" for one cycle. I could not build a repro without pushing to the real remote, and the next scheduled `kipi work` re-reads it. Dropped.
- The new tests are not decoration. `test-linear-worker-round-cap.sh:163` asserts a real dispatch count that would go red if the reset stopped working on the happy path.

## On the review process

Round 1 raised 5 findings; all 5 were fixed in `abeabb8`. Both of my findings sit in the code those fixes added (the directive producer, the round cap and its reset). This is **not** the "new blockers on unchanged code" pathology, so the earlier rounds read as correctly calibrated. The failure mode here is the one this PR's own rework prompt names: a fix that is correct in isolation shipping a new defect one layer out.

## VERDICT: REQUEST CHANGES

Not BLOCK. Nothing here is permanent or unrecoverable, and the central mechanism (the scratch guard) works and does not touch the founder's checkout.

**Fix first:** `linear-worker.sh:140`. The round cap this PR adds has exactly one escape hatch, and right now it tells the operator it worked whether or not it did.

**Cleanup on my side:** my repro scratch is at `sites/rev22/` (gitignored, `git status` clean, real repo untouched). Delete it whenever.

FINDINGS:
major|kipi work --reset-rounds reports "reset to 0; the next run will dispatch again" and exits 0 without validating the issue id or checking the write, so a lowercase or typo'd id creates a phantom ledger key and leaves the real issue capped and permanently silent|q-system/.q-system/scripts/linear-worker.sh:140
minor|The new kipi directive help line is unquoted, so kipi help emits a redirection error, exits 1, and drops both its own line and the pre-existing kipi dor line, leaving the new command undiscoverable|kipi:37
minor|git commit --no-verify bypasses the scratch guard entirely while the agent prompt explicitly names the hook as the mechanism|q-system/.q-system/scripts/linear-worker.sh:666
END FINDINGS

assafkip:
Round-4 rework pushed as `fa74b1d` + `ddd42be`. All three findings fixed, none disputed. Each has a reproducer observed RED first, and the real output is quoted.

## 1. major -- `--reset-rounds` reports success unconditionally. FIXED.

RED first, driving the real script exactly as you did:

```
FAIL: PHANTOM KEY: --reset-rounds --issue ask-aaa wrote a new lowercase entry
      {"rounds": 0} while the real ASK-AAA stayed {"count": 0, "rounds": 9, "capped_notified": true}.
      The operator is told it worked; the issue stays capped and never pages again.
```

The write is now checked, read back off disk, and the previous value is **reported rather than claimed**: `ASK-AAA: review rounds 9 -> 0, verified by read-back`. A failed write says `reset FAILED, round count UNCHANGED` and exits 1. An id with no recorded rounds says `nothing to reset -- no round count recorded for it. Check the id` and exits 1, instead of inventing an entry.

**The layer above, which is where I put the case normalization.** Your fix shape was `tr 'a-z' 'A-Z'` inside the reset branch. I did it at the arg-parse instead, because `--issue` has a **second reader**: the picker matches `i["identifier"] == only` exactly, so `--issue ask-208` was *also* silently returning zero ready issues on the normal work path. Fixing only the reset would have left the operator with a reset that reports correctly and a `kipi work --issue ask-208` that still does nothing and says `0 ready issue(s)`. One normalization, both readers. The test asserts both halves, and the picker stub reproduces the real exact-match semantics rather than returning the issue regardless of the filter -- a lenient stub would have hidden this half entirely:

```
  ok: a lowercase issue id resets the REAL entry, with no phantom key
  ok: the same lowercase id also reaches the picker (one normalization, both readers)
```

**The reproducer found a third failure mode while proving yours.** The old body was `except Exception: d={}` followed by a full `json.dump`. On a corrupt ledger that does not fail -- it rebuilds the file from scratch and **every other issue's attempt and round counters go with it**, reported as success. Silent data loss, not just a silent no-op. The new reader refuses to write what it could not parse, and case 5 asserts the file is left byte-identical.

**Nothing got quieter.** Case 6 pins that after a *failed* reset the issue is still visibly capped in the dry run, so the failure path cannot buy silence by half-writing the ledger. `PASS: worker reset-rounds (9 checks)`.

Case 2 (unwritable ledger) is only meaningful unprivileged, since root can write a 0444 file. It announces itself as not-counted under root rather than silently passing.

## 2. minor -- the `kipi directive` help line is unquoted. FIXED.

RED:

```
FAIL: kipi help exited 1. stderr: .../kipi: line 37: text: No such file or directory
```

Inner quotes escaped, and the two lines split into two `echo`s.

**The assertions are shaped to catch the class, not this instance.** Grepping for the two missing lines would pin this one bug and nothing else. A stray redirection anywhere in `usage` shows up as a non-zero exit, a stderr line, or a file touched in the cwd -- so the test runs `kipi help` from an empty temp dir and asserts all three, plus your amplification case (a pre-existing file named `text` survives untouched, which is the variant that produces *no* error at all).

**On discoverability, with a correction to the framing.** `kipi help` **is** the registration surface for kipi subcommands -- `review`, `work`, `converge`, `jobs` are not in `CLAUDE.md` either, so adding `kipi directive` there and not its peers would be inconsistent. Fixing `usage` is the registration. Case 5 makes that deterministic: every subcommand the case statement handles must appear in `kipi help`.

**One line rode along, flagged rather than buried.** That coverage assertion went red on a pre-existing gap -- `kipi sync-skills` is handled and was never documented. I added its one-line entry (marked DEPRECATED, which is what it prints) because the guard cannot ship while an existing subcommand fails it. Reversible in one line if you would rather the guard carry an allowlist.

`PASS: kipi usage (8 checks)`.

## 3. minor -- `--no-verify` bypasses the guard, and the prompt names the mechanism. FIXED, both halves in code.

I took your point that every git hook has this property, and I did **not** treat the reword as the fix. A reworded prompt is prompt-only enforcement, which this repo does not accept -- so the mechanism-naming half got the wording change and the bypass half got a second gate.

RED:

```
FAIL: SCRATCH PUSHED: q-system/output/ask208-helper.py was committed with
      --no-verify and then pushed with nothing refusing it.
```

The generated guard is now installed as **both** `pre-commit` and `pre-push`, from one heredoc, byte-identical under the two names. Three decisions worth stating:

- **One file, two names, `cmp`-asserted.** Two copies of the scratch regex is the defect, not the fix; they would drift on the next edit to either.
- **The delegate gets its stdin replayed.** `pre-push`'s spec arrives on stdin and the guard consumes it. Handing the repo's own `pre-push` an empty stdin would make it see zero refs and pass everything -- a gate subtracted while still appearing to run. Case 5 plants an `exit 9` pre-push and asserts the push is refused by it.
- **`pre-push` left the mirror allowlist**, for the same reason `pre-commit` is not in it: the guard *is* that hook now and chains to the original at runtime.

**Stated plainly: this is not unbypassable.** `git push --no-verify` exists. It makes one lazy `-n` insufficient, which is the shape you actually measured. The prompt now says a commit or push "is refused" without naming what does the refusing, and case 6 asserts the prompt contains neither `no-verify` nor `commit hook` **and** still states the rule -- dropping the mechanism must not drop the constraint with it.

**The negative side is asserted harder than the positive side**, because a pre-push hook that over-refuses stops every instance's agent from opening a PR at all, which is worse than the leak:

```
  ok: an RCA, a top-level .md and a real harness still push
  ok: a follow-up push onto an existing remote branch is not refused
  ok: the push guard refuses none of the 1449 paths tracked on main
  ok: the repo's existing pre-push still runs after the guard (chained, not replaced)
```

The second-push case exists because the two ref shapes take different code paths (a real remote sha vs all-zeros for a new branch); asserting only the first push would leave the common path untested. `PASS: scratch guard at push (10 checks)`.

## Regression set -- nothing you said you could not break got quieter

```
test-linear-worker-fetch                       rc=0  PASS: linear-worker fetch (4 checks)
test-rework-gate-mergeability                  rc=0  PASS: rework gate mergeability (16 checks)
test-linear-worker-directive-scratch           rc=0  PASS: worker directive + scratch guard (17 checks)
test-linear-worker-infra-exit                  rc=0  PASS: worker infra exit (10 checks)
test-linear-worker-round-cap                   rc=0  PASS: worker round cap (8 checks)
test-linear-worker-reset-rounds                rc=0  PASS: worker reset-rounds (9 checks)      [new]
test-scratch-guard-push                        rc=0  PASS: scratch guard at push (10 checks)   [new]
test-kipi-usage                                rc=0  PASS: kipi usage (8 checks)                [new]
test-linear-worker-parallel                    rc=0  PASS: linear-worker parallel (7 checks)
test-converge                                  rc=0  PASS: 17/17 converge checks
```

The per-worktree `core.hooksPath` isolation, the pre-commit chaining, the ten (verdict x mergeability) unit pairs and the round cap's page-once behaviour are all untouched and green. Your round-3 verification that the scale sweep is genuine still holds and now covers the push path too (1449 tracked paths, zero refusals).

All three new tests are declared in `capability-manifest.json`. `capability-gate.py` standalone: `tests: ran=80 quarantined=0` -> **capability-gate: GREEN**.

`python3 validate-separation.py 1 --verbose` -> **PASS: 68, FAIL: 1, WARN: 1** on one run and the capability gate PASS on the next. The FAIL is the known `sp-e0740c76` flake (`test-updater-issue-sequence.py` outrunning the gate's 60s per-test budget, not failing), which is why the gate is green when run directly. Still captured, still not fixed here.

## Ledger

`prd_runner.py gates run` is RED on two open spillover items, both carried in from earlier rounds of this issue and both out of scope for it:

- `sp-e0740c76` -- the capability-gate 60s per-test budget flake above.
- `sp-762c7d6b` -- `grep -c PATTERN FILE || echo 0` used as a zero-safe count across test scripts (`grep -c` exits 1 on a zero count).

Nothing new was deferred in this pass.

assafkip:
# Review: PR #22 — round 5 (the PR's own numbering; the pipeline called it round 3)

I read the diff first, ran everything, then read the prior comments. Three findings survived reproduction. One is new-code, not re-litigation: I grepped every prior round for `atomic|concurren|race|ledger corrupt` and got zero hits.

---

## 1. major — `mark_capped` wipes the whole attempts ledger on any parse failure, disarming both caps

`q-system/.q-system/scripts/linear-worker.sh:495`

The round cap this PR adds is the first thing that ever *reads* `rounds`. That makes `~/.config/kipi/linear-worker-attempts.json` load-bearing for dispatch gating. The new `mark_capped` still does `except Exception: d={}` and then writes `d` back — the exact shape the author's round-4 response declared fixed ("the old `except Exception: d={}` was also silent data loss ... the reader below refuses to write what it could not parse"). That fix landed only in the `--reset-rounds` branch.

**Repro, ledger level** (`.pr22rev/repro.sh`, section A — the literal body of `mark_capped` lines 495-502):

```
--- ledger BEFORE:
{ "ASK-100": {"count": 3, "rounds": 6, ...}, "ASK-200": {"count": 0, "rounds": 5} }
--- ledger AFTER a torn write:
{ "ASK-100": {"count": 3, "rounds": 6, "why": "claude ru
mark_capped ASK-300 -> 1
--- ledger AFTER mark_capped:
{ "ASK-300": { "capped_notified": true } }
rounds_for ASK-100 = 0   (was 6, cap is 6)
rounds_for ASK-200 = 0   (was 5)
```

**Repro, end-to-end through the real worker** (`repro2.sh`, sections B1/B2):

```
== B1: healthy ledger, ASK-100 at 6/6 rounds
skip ASK-100: 6/6 review round cap reached; parked for a human.

== B2: SAME ledger, torn
{"ASK-100": {"count": 0, "rounds
[dry] would work ASK-100 (attempt 1/3, round 1/6)
```

**Repro, why a torn ledger happens** (`repro2.sh`, section C — 4 processes doing the worker's own read-modify-write, no lock, `open(p,'w')` is not atomic):

```
reads that found an UNPARSEABLE ledger and fell back to d={}: 593
```

Two `kipi work` runs (launchd + `converge.sh`, or the parallel mode `test-linear-worker-parallel.sh` exists for) write this one file with no lock. Each worker run writes it 2-3 times.

**3am:** both caps reset to zero, nothing pages. A parked issue resumes dispatching and posts more permanent Linear comments. Recoverable — hand-edit the ledger — so major, not blocker.

---

## 2. minor — one accented character in a filename walks past both gates

`q-system/.q-system/scripts/linear-worker.sh:286`

`git diff --cached --name-only` emits the C-quoted form for non-ASCII paths, so the `^q-system/output/...` anchor never matches.

**Repro** (`repro3.sh`, real worker, real installed guard):

```
--- control: the plain ASCII name
   refused: BLOCK: agent scratch cannot be committed.
--- same file, one accented character in the name
   git stages it as: "q-system/output/ask208-h\303\251lper.py"
   COMMITTED. The guard did not refuse it.
--- and at push?
   PUSHED. Neither gate saw it.
   on the remote: "q-system/output/ask208-h\303\251lper.py"
```

Dropped a level on purpose: I had to construct this input. All seven files swept in PR #20 were ASCII, and no producer in this repo emits a non-ASCII scratch name today.

---

## 3. minor — the first worker run permanently mutates the founder's shared `.git/config`

`q-system/.q-system/scripts/linear-worker.sh:382`

`git -C "$tree" config extensions.worktreeConfig true` has no `--worktree`, so it lands in the common config, not the worktree's. The paired test asserts the main checkout's `core.hooksPath` stays unset — it does — but nothing covers this one, and it is the change that outlives the worktree.

**Repro** (`repro3.sh`):

```
=== founder checkout .git/config BEFORE the worker runs:
  (no extensions.*, no core.hooksPath)
=== founder checkout .git/config AFTER one worker run:
extensions.worktreeconfig=true
=== which config file holds it:
18:	worktreeConfig = true
```

The flag is genuinely required for `config --worktree`. What is missing is that it is declared, asserted, and reversible. Git's own docs warn that enabling it changes how `core.bare` / `core.worktree` in the common config are honored, and older git refuses a repo carrying it. `kipi update` puts this script in every instance.

---

## What is sound

- **Directive → prompt injection.** `PROMPT` reaches the inner shell as positional `$1` and is referenced as `"$1"`, so directive text cannot break out. Tried; could not.
- **`rework_gate` false refusal.** `UNKNOWN` and empty both keep the old behavior. I wanted this to be a cry-wolf source and it is not — the conservative direction is the right one.
- **Guard false positives at scale.** 1450 tracked paths swept through the real installed hook, zero refusals. An RCA, a top-level `.md`, and a real harness all commit and push.
- **Chaining.** Planted `exit 9` pre-push and pre-commit both still fire. The runtime delegate resolution (rather than install-time baking) holds.
- **Main-checkout isolation.** `core.hooksPath` stays unset on the founder's tree. Verified independently in my own fixture.
- **Fetch ordering.** The test reads back the real base SHA, not the source. That is the right shape and it passes.
- **`kipi help`.** Parses, exits 0, and the coverage assertion over the case statement can actually fail.
- **Regressions.** `test-linear-worker-parallel.sh` 7/7, `test-converge.sh` 17/17. Every one of the 8 new tests green.

The test fixtures here are not the invented kind. `test-scratch-guard-push.sh` drives real `git push` against a real bare remote and reads the refusal back; `test-linear-worker-directive-scratch.sh` captures the actual prompt string the model receives. Those could fail.

## On the review process

This is round 5. Rounds 3, 4 and 5 each found their major inside code the *previous* round asked for:

- round 3's major → the fetch guard (added in round 2)
- round 4's major → `--reset-rounds` (added in round 3 to fix round 3's finding 4)
- this round's major → `mark_capped` (added in round 3 to fix round 3's finding 4, carrying the defect round 4's response declared fixed)

So the earlier rounds were not miscalibrated — they were right, and they are still right. The problem is the other one: 2074 added lines for a 4-defect PR, and each fix arrives faster than it gets verified. Worth considering landing fixes 1 and 2 (fetch + mergeability, ~80 lines, clean) separately from the ledger-and-cap machinery, which is where every regression has come from.

## VERDICT: REQUEST CHANGES

Not BLOCK — nothing here writes an unrecoverable object or destroys founder work.

**Fix first:** give the ledger one writer with an atomic write and a refuse-on-unparseable read, the same rule already applied in `--reset-rounds`, applied to `mark_capped`, `bump_attempt`, and the rounds bump. Everything else on this list is bounded.

---

Housekeeping: my scratch is at `/Users/assafkipnis/projects/kipi-system/.pr22rev/` (untracked, three repro scripts). The `rm -rf` deny hook blocked my cleanup, correctly. Remove it with `ALLOW_DESTRUCTIVE=1 rm -rf .pr22rev` when convenient. The review worktree is already gone. Separately: `git worktree list` shows a leaked `~/.config/kipi/worktrees/ask-aaa` on branch `sana/ask-aaa` and a `template-repo/pr22-review/tree` from an earlier round — both predate this diff.

FINDINGS:
major|mark_capped rebuilds the attempts ledger from {} on any parse failure and writes it back, wiping every issue's count/rounds/capped_notified and silently disarming both the new round cap and the attempt cap|q-system/.q-system/scripts/linear-worker.sh:495
minor|A non-ASCII scratch filename is emitted C-quoted by git diff --cached --name-only, so the anchored SCRATCH_RE never matches and the file commits and pushes past both gates|q-system/.q-system/scripts/linear-worker.sh:286
minor|install_scratch_guard writes extensions.worktreeConfig into the founder's shared .git/config permanently, with no cleanup path and no test asserting it|q-system/.q-system/scripts/linear-worker.sh:382
END FINDINGS

assafkip:
Closed unmerged at the 3-round cap. Not a reviewer/Sana disagreement — the scope was mine and it was too wide.

Four fixes bundled into one file meant every round re-swept all four surfaces. Findings went 5 -> 3 -> 3: flat, and each round's major was newly introduced by the previous round's fix. Round 3's major (`mark_capped` rebuilding the attempts ledger from {} on any parse failure) would silently disarm the very caps this PR adds, which disqualifies it regardless of the rest.

Branch `sana/ask-208` is kept. The four fixes will be re-filed one at a time, smallest first. Evidence for the scope theory: ASK-209 shipped one small change and converged in a single round.

Spillover unchanged and still open: sp-28ced3d6, sp-71b63e62, sp-fd76af2f, sp-1aae7516.

PR #24, opened 2026-07-27
fix(worker): fetch origin before cutting a worktree, and page when the fetch fails (ASK-211)

assafkip:
Fix 1 of 4 from ASK-208 (PR #22, closed unmerged at the round cap), re-filed on its own per the round-3 reviewer: *"worth landing fixes 1 and 2 (fetch + mergeability, ~80 lines, clean) separately from the ledger-and-cap machinery, which is where every regression has come from."*

Resolves spillover `sp-28ced3d6`.

## The defect

`linear-worker.sh` contained no `git fetch` anywhere. Every worktree was cut from whatever local `origin/main` ref happened to be lying around.

**Observed 2026-07-27:** ASK-150 was dispatched to resolve a merge conflict. It merged `3b60af0` and the conflict survived, because `main` was already `72c782d`. The agent did the right thing to the wrong target and two rounds were burned.

## The change (3 files, +52 lines)

- Fetch runs **once per run**, above the picker, so no code path can create a worktree before it.
- A fetch failure is environmental (`self-healing-retry.md` rule 5): stops on attempt 1, is **not** counted against any issue, pages through `$NOTIFY`, exits **9**.
- `NOTIFY` is overridable via `KIPI_NOTIFY` (the seam `converge.sh` already uses) so the suite reads back what was paged instead of grepping the source.
- Exit-code contract documented in the header: `0` healthy, `1` usage, `9` infra.

Exit 9 + the page are the half that **failed review the first time**. A guard whose failure path is `say` + `exit 0` makes an expired credential at 3am byte-for-byte indistinguishable from a healthy no-work run, and `MAX_ATTEMPTS` only counts DISPATCHED runs, so the issue never becomes stuck and never pages that way either.

## Reproducer: observed RED twice, then green

`q-system/.q-system/scripts/test/test-linear-worker-fetch.sh` (new, in the capability manifest).

It never greps the source. It stales a **real** remote-tracking ref and asserts the SHA the worktree actually landed on (a grep for `git fetch` would pass on a fetch placed *after* the worktree is created), and it runs against a genuinely unreachable origin with `$NOTIFY` redirected to a recorder file.

**RED 1** — against the unfetched script:
```
FAIL: STALE BASE: the worktree was cut from the stale origin/main (4cd5b85...).
      The true remote head is f6e5584...
```

**RED 2** — against a staged fetch guard that exited 0 (the exact shape round 3 rejected):
```
  ok: the worktree was cut from the TRUE remote head
  ok: the skeleton's refs/remotes/origin/main was updated to the remote head
  ok: a 2-issue run fetched exactly once
FAIL: SILENT SUCCESS: the worker exited 0 after a fetch failure that stopped
      the entire run. A caller cannot tell this apart from a healthy run.
```

**GREEN:**
```
  ok: the worktree was cut from the TRUE remote head (fetch happened before worktree add)
  ok: the skeleton's refs/remotes/origin/main was updated to the remote head
  ok: a 2-issue run fetched exactly once
  ok: a fetch failure exits non-zero (rc=9), so a caller can tell
  ok: the fetch failure pages the founder through $NOTIFY
  ok: the page names the cause (git fetch), not just 'something failed'
  ok: no worktree was cut and no agent was dispatched
  ok: a healthy run with nothing ready still exits 0
  ok: a healthy no-work run pages nobody
  ok: linear-worker.sh parses (bash -n)
PASS: linear-worker fetch (10 checks)
```

Case 7 is load-bearing: without it the fix could pass by always failing and always paging, which is the same defect wearing the other hat.

**Capability gate:**
```
  tests: ran=73 quarantined=0 skipped-skeleton-only=0
capability-gate: GREEN
```

## Not doing (binding, per the DoR)

mergeability in the rework gate; the scratch guard; operator directives; `--reset-rounds`; the round-cap ledger; `converge.sh`, `pr-verdict-lib.sh`, `pr-review-agent.sh`. All still live on `sana/ask-208`.

## Known consequence, captured not fixed

`converge.sh` on `main` stops with its own exit 7 + Slack ping when no PR exists, so a fetch failure under `converge` now pages **twice** for one event. Mapping worker exit 9 to a single converge page is part of the `converge.sh` work this issue explicitly does not touch. Captured as spillover rather than mentioned.

### DIFF
--- a/q-system/.q-system/capability-manifest.json
+++ b/q-system/.q-system/capability-manifest.json
@@ -102,6 +102,10 @@
       "runner": "bash"
     },
     {
+      "path": "q-system/.q-system/scripts/test/test-linear-worker-fetch.sh",
+      "runner": "bash"
+    },
+    {
       "path": "q-system/.q-system/scripts/test/test-linear-worker-parallel.sh",
       "runner": "bash"
     },
--- a/q-system/.q-system/scripts/linear-worker.sh
+++ b/q-system/.q-system/scripts/linear-worker.sh
@@ -28,6 +28,13 @@
 #   human interrupt                  -> destructive-op-deny.sh; and it cannot merge
 #   goal met                         -> the PR + the closeout gates, not this script
 #
+# EXIT CODES
+#   0  ran (or had nothing to run). A caller may treat this as healthy.
+#   1  usage error
+#   9  INFRA: the environment is down (git fetch failed) and the run did NO
+#      work. Paged on the way out. Distinct from 1 so a caller can tell a dead
+#      environment from a bad invocation.
+#
 # WHY INFRA FAILURE IS COUNTED SEPARATELY
 # ---------------------------------------
 # An expired auth token or a Linear outage is not the issue's fault. Counting it
@@ -50,7 +57,11 @@ SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
 SKEL="${KIPI_SKEL:-$(cd "$SCRIPT_DIR/../../.." && pwd)}"
 CLAIM="$SCRIPT_DIR/linear-claim.py"
 SYNC="$SCRIPT_DIR/linear-sync.py"
-NOTIFY="$SCRIPT_DIR/slack-notify.sh"
+# Overridable so the suite can read back WHAT was paged without paging the
+# founder, and so "did anyone get told?" is answered by a file instead of by a
+# grep of this source. Same seam and same name converge.sh already uses -- one
+# convention, not two. Default is always the real Slack sink.
+NOTIFY="${KIPI_NOTIFY:-$SCRIPT_DIR/slack-notify.sh}"
 STATE_DIR="${KIPI_STATE_DIR:-$HOME/.config/kipi}"
 ATTEMPTS="$STATE_DIR/linear-worker-attempts.json"
 LOG="$STATE_DIR/linear-worker.log"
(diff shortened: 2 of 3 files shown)


assafkip:
Token guard hit at 50 calls, so I'm stopping here and reporting. Repro C (converge's handling of the new exit 9) was written but never executed — per the standing rule it is **not** a finding and I've dropped it.

# Review: PR #24 — fetch origin before cutting a worktree (ASK-211)

Scratch lives at `.pr24rev/` (untracked, same pattern as the existing `.pr23rev/`). Repo was not modified.

---

## Findings

### 1. minor — the fetch is the worker's only unbounded call, so a blackholing remote hangs it forever with zero pages

`q-system/.q-system/scripts/linear-worker.sh:150`

`run_bounded()` is defined 47 lines above at :103 and is not applied to the fetch. A remote that accepts the TCP connection and never answers (captive portal, hung proxy, middlebox dropping an established session) blocks `git fetch` indefinitely. The worker then never picks an issue, never pages, never exits — the go-dark failure this PR exists to remove, relocated into the fix. `converge.sh:22` also documents "wall clock 4 -> inherited: each round is bounded inside the worker", which is no longer true.

```
bash .pr24rev/repro-b-hang.sh
```

```
starting the worker against a blackholing origin on port 60747 ...
  t=5s: worker still alive; pages so far=0; log=
  t=10s: worker still alive; pages so far=0; log=
  t=15s: worker still alive; pages so far=0; log=
  t=20s: worker still alive; pages so far=0; log=
  t=25s: worker still alive; pages so far=0; log=
  t=30s: worker still alive; pages so far=0; log=
RESULT: after 30s the worker is STILL BLOCKED in git fetch.
        pages sent: 0   (nobody has been told)
        worktrees cut: 0
        run_bounded() is defined at line 103 and is NOT applied to the fetch at line 150
```

**Severity is dropped one level on purpose.** I had to construct the blackhole listener; nothing in this repo emits that shape, and `origin` is GitHub over https, which fails rather than hangs. Base severity would have been major (unattended, dark, no alert). Also worth stating plainly: no launchd plist currently schedules `linear-worker.sh` — the three plists in `q-system/.q-system/scripts/` are fleet-health, linear-dor and openloops-heartbeat — so the "3am unattended" case is the stated intent, not yet the wiring.

### 2. minor — a dry run now performs a network write and pages Slack

`q-system/.q-system/scripts/linear-worker.sh:150`

The fetch sits above the `$APPLY` check, so `kipi work` with no `--apply` — documented in the header as "Dry by default: prints what it would pick and stops" — sends a Slack page and exits 9 when the network is down, and prints nothing about what it would have picked. A founder previewing the board from a plane pages their own Slack.

```
bash .pr24rev/repro-a-dry.sh
```

```
dry-run rc = 9   (documented dry contract: print what it would pick, stop)
pages sent from a DRY run: 1
page text: worker: git fetch failed in /var/folders/.../skel -- the run did NO work. Check credentials/network.
stdout: 2026-07-27T22:20:20Z INFRA: git fetch failed in /var/folders/.../skel. Stopping before any worktree is cut from a stale base.
```

---

## What is sound

**Mutation-tested the new suite. Five mutants, five killed.** This is the check I expected to break it and it did not.

| mutant | killed by |
|---|---|
| fetch removed entirely | case 1 — `FAIL: STALE BASE ... efb6e4a` vs true head `cf48632` |
| fetch moved down / made per-issue | case 3 — `FAIL: a 2-issue run fetched 2 times` |
| `exit 9` → `exit 0` (the shape round 3 rejected) | case 4 — `FAIL: SILENT SUCCESS` |
| page removed, exit 9 kept | case 5 — `FAIL: NOBODY WAS PAGED` |
| guard inverted (always fail, always page) | case 1 — `FAIL: no worktree was created (rc=9)` |

**The fixture matches the real producer.** Real bare repo, real second clone advancing origin, real remote-tracking refs, assertion on the SHA the worktree actually landed on. Not a source grep. I could not find a shape here that no producer emits.

**The new fetch does not silently neuter the existing mutex suite.** This was my main hypothesis — a fetch that fails in an old fixture would make `test-linear-worker-parallel.sh` pass vacuously. Ran it on this branch: 7/7 green, and all seven checks still report reaching the work phase, so the fetch is not short-circuiting those runs.

**Concurrency.** The fetch is before the claim, before the worktree, before any Linear write. Two workers fetching the same repo concurrently is handled by git's ref locking, and exit 9 cannot strand a claim or leak a lock because no claim exists yet.

**Blast radius on permanent objects is zero.** Exit 9 fires before the Linear "Picked up" progress note and before any `gh pr create`, so a fetch outage cannot write to an undeletable Linear issue or open a PR. Change is additive across 3 files with no state migration; `git revert` restores prior behaviour exactly. Rollback is clean.

---

## VERDICT: APPROVE WITH NITS

**Fix first:** wrap the fetch in `run_bounded` (finding 1) — the guard currently cannot survive the failure mode it was built for, and the helper is already sitting 47 lines above it. Finding 2 is a one-liner in the same block.

FINDINGS:
minor|The new git fetch is the worker's only unbounded call and has no run_bounded wall clock, so a blackholing remote hangs the run forever with zero pages|q-system/.q-system/scripts/linear-worker.sh:150
minor|The fetch runs above the $APPLY check, so a documented dry-run preview performs a network write, pages Slack and exits 9|q-system/.q-system/scripts/linear-worker.sh:150
END FINDINGS

PR #27, opened 2026-07-27
fix(token-guard): a gate-refused commit gets a bounded budget, not a deadlock (ASK-215)

assafkip:
Resolves ASK-215 (spillover `sp-97303649`).

## The deadlock

The volume ceiling exempts the commit **command** but not the commit's
**preconditions**. A commit refused by a pre-commit gate is not a checkpoint, and
satisfying that gate needs an `Edit` the ceiling blocks.

Observed 2026-07-27 on ASK-214: lefthook's `plugin-version-bump` refused the
checkpoint, the agent had 18 of 20 tests done, and the run stranded with
everything staged and nothing committed. The block message said `50 tool calls`,
the wrong reason, which sent it back into the deadlock instead of at the gate.

## The fix

A refused commit grants `GATE_GRACE = 8` calls. Bounded three ways:

- spent **only** at the ceiling, on calls the ceiling would otherwise block, so a
  run that never reaches the ceiling never touches the budget
- at most **one grant per user-message window**, so there is no refuse / spend 8 /
  refuse ratchet. A second refusal updates the gate *name* without topping the
  budget up: the budget is per deadlock episode, not per gate
- a landed commit or a new user message clears it, so no stale budget silently
  widens a later ceiling

When it is spent the run hard-stops with a message naming the gate:
`plugin-version-bump refused the checkpoint and the 8-call grace budget is spent.`

Blind to which gate fired, by design. This repo has seven blocking pre-commit
gates and each can demand a different file, so exempting `plugin.json` is
whack-a-mole the next gate defeats.

## Reproducers (RED first, cases 7-10, extended in place)

| Case | Pins |
|---|---|
| 7 | at the ceiling with a recorded gate-refused commit, an `Edit` is allowed |
| 8 | `nothing to commit` mints no budget, so grace cannot be minted with an empty commit |
| 9 | the 9th call after the grant blocks, and names `plugin-version-bump` |
| 10 | a landed commit clears the budget |

Observed RED before the fix:

```
FAIL: Edit after a gate-refused commit exited 2, want 0 (deadlock): 50 tool
calls without user input or a commit. Commit finished work now ...
```

## Checks run

```
$ bash q-system/.q-system/scripts/test/test-token-guard-hook-behavior.sh
ok: warn() emits nested PreToolUse hookSpecificOutput
ok: PostToolUse successful Edit clears edit_targets
ok: PostToolUse failed Edit keeps the spiral counter
ok: guard-blocked attempts stay volume-blocked and are un-counted
ok: volume block names the commit escape hatch
ok: exact-retry still fires for executed identical calls
ok: a gate-refused commit grants a bounded budget so the ceiling is escapable
ok: a no-op commit mints no budget; the ceiling still blocks
ok: budget is bounded at 8 and the hard stop names the refusing gate
ok: a landed commit clears the budget
PASS: token-guard hook behavior (... + gate-refusal grace)

$ python3 -m pytest q-system/.q-system/tests/test_token_guard.py -q
14 passed in 0.52s

tests: ran=75 quarantined=0 skipped-skeleton-only=0
capability-gate: GREEN
```

## Grounding, not remembering

The lefthook fixture is real captured output (lefthook v2.1.6, 2026-07-27), not
recalled text. Two facts came out of that capture and shaped the design:

1. A hook-refused commit prints **only the hook's own output**, exit 1. git adds
   no marker of its own, so the gate name can only come from the gate's text, and
   a positive "a hook refused this" signal does not exist to key off.
2. lefthook marks the failed command in its summary with a boxing glove and
   passing ones with a check mark, so the failing command is extractable. The
   `lefthook v2.1.6` banner also carries the glove; the extractor anchors on the
   stripped line *starting* with it, which the banner does not.

Detection is therefore a negative list: a `git commit` that failed and is not one
of the known non-gate failures (no-op, empty message, not a repo, unmerged, no
identity) is treated as refused. A mis-read costs at most 8 calls; missing a real
refusal costs the whole run.

## Also wired

- `PostToolUse` matcher `Edit|Write|MultiEdit|Bash` in `.claude/settings.json`
  already covers the Bash leg, so the recorder has a live path (verified, not
  assumed).
- Wiring B (`reset_volume_if_committed`) clears the budget too, so a commit that
  lands without a `PostToolUse` event still ends the episode.

## Not in scope (per the DoR)

No threshold changes, no lefthook-gate changes, no other detectors, no
`converge.sh` / `linear-worker.sh` / `pr-verdict-lib.sh`.

One adjacent defect found and captured rather than fixed here: `sp-1078fbe2` -
`_is_successful_commit` reads a commit as successful when the response carries no
`error` and no exit code, even with no "1 file changed" marker, so a refused
commit in that response shape silently resets the ceiling. It does not cause a
deadlock (a reset ceiling blocks nothing), which is why it is separate.

### DIFF
--- a/q-system/.q-system/scripts/test/test-token-guard-hook-behavior.sh
+++ b/q-system/.q-system/scripts/test/test-token-guard-hook-behavior.sh
@@ -465,4 +465,36 @@ assert cache.get("gate_grace_gate") == "plugin-version-bump", \
 EOF
 echo "ok: a refusal carried by exit_code alone is read the same as one carried by error"
 
+# --- 16) Holding the volume warning must not lose it --------------------------
+# The finding-3 fix defers the volume warning so the checks below it can run. A
+# later warning must therefore CARRY it, not replace it -- otherwise the fix
+# buys reachability by making the ceiling warning silent, which is the trade the
+# operator would never agree to.
+python3 - "$CACHE" "$SESSION" "$GUARD" <<'EOF'
+import json, re, sys, time
+src = open(sys.argv[3]).read()
+warning = int(re.search(r"^VOLUME_WARNING = (\d+)", src, re.M).group(1))
+spiral = int(re.search(r"^READ_SPIRAL_LIMIT = (\d+)", src, re.M).group(1))
+json.dump({
+    "actor_key": sys.argv[2], "tool_calls_since_user": warning,
+    "agent_calls_since_user": 0, "mcp_timestamps": [], "repeat_map": {},
+    "consecutive_reads": spiral - 1, "warnings_issued": 0, "file_read_counts": {},
+    "greps_since_write": 0, "edit_targets": {}, "agents_without_write": 0,
+    "last_write_time": time.time(), "calls_since_write": 1,
+    "last_volume_reset": time.time(),
+}, open(sys.argv[1], "w"))
+EOF
+OUT="$(run_guard "{\"session_id\": \"$SESSION\", \"hook_event_name\": \"PreToolUse\", \"tool_name\": \"Read\", \"tool_input\": {\"file_path\": \"/tmp/both-warnings-probe\"}}")" \
+  || { echo "FAIL: combined-warning path exited non-zero"; exit 1; }
+python3 - <<EOF
+import json
+context = json.loads(r'''$OUT''')["hookSpecificOutput"]["additionalContext"]
+volume, _, spiral = context.partition("\n\n")
+assert "before hard stop" in volume, \
+    f"the held volume warning was dropped by a later warning: {context!r}"
+assert spiral.strip(), \
+    f"the later read-spiral warning was dropped: {context!r}"
+EOF
+echo "ok: a held volume warning is carried by the next warning, not replaced"
+
 echo "PASS: token-guard hook behavior (warn shape + PostToolUse edit reset + blocked-attempt un-count + gate-refusal grace + non-gate classifier)"
--- a/q-system/.q-system/scripts/test/test-token-guard-hook-behavior.sh
+++ b/q-system/.q-system/scripts/test/test-token-guard-hook-behavior.sh
@@ -21,6 +21,10 @@ run_guard() { # $1 = event JSON on stdin; guard runs with fixture as project dir
   printf '%s' "$1" | CLAUDECODE=1 CLAUDE_PROJECT_DIR="$FIXTURE" python3 "$GUARD"
 }
 
+run_guard_file() { # $1 = path to a payload JSON file (for payloads with unicode/newlines)
+  CLAUDECODE=1 CLAUDE_PROJECT_DIR="$FIXTURE" python3 "$GUARD" < "$1"
+}
+
 # --- 1) warn() output must use the nested PreToolUse hookSpecificOutput form ---
 # Seed a cache one read short of the read-spiral threshold; the next Read warns.
 python3 - "$CACHE" "$SESSION" <<'EOF'
@@ -176,4 +180,132 @@ case "$ERR" in
   *) echo "FAIL: 3rd executed identical call blocked with wrong reason: $ERR"; exit 1;;
 esac
 
-echo "PASS: token-guard hook behavior (warn shape + PostToolUse edit reset + blocked-attempt un-count)"
+# --- 7-10) The ceiling's escape hatch must survive a pre-commit gate (ASK-215) --
+# Scar 4 (2026-07-27, ASK-214): at the ceiling the ONLY exempt call is `git
+# commit`. lefthook's plugin-version-bump REFUSED that commit, and bumping
+# plugin.json needs an Edit the ceiling blocks. The run stranded with 18/20 tests
+# done and everything staged. A refused commit is not a checkpoint, so the
+# commit-command exemption alone does not reach the deadlock: the gate's
+# PRECONDITION needs a bounded budget of its own.
+#
+# Fixture is a real lefthook v2.1.6 refusal (captured 2026-07-27). git passes a
+# hook's output through verbatim and adds no marker of its own, so the gate name
+# can only come from the gate's own text.
+mk_payloads() {
+python3 - "$FIXTURE" "$SESSION" <<'EOF'
+import json, os, sys
+fixture, session = sys.argv[1], sys.argv[2]
+
+LEFTHOOK_REFUSAL = (
+    "╭──────╮\n"
+    "│ \U0001f94a lefthook v2.1.6  hook: pre-commit │\n"
+    "╰──────╯\n"
+    "┃  gitleaks ❯ \n\nclean\n\n"
+    "┃  plugin-version-bump ❯ \n\nBLOCK: bump plugin.json\n\n"
+    "exit status 1\n"
+    "summary: (done in 0.10 seconds)\n"
+    "✔️ gitleaks (0.10 seconds)\n"
+    "\U0001f94a plugin-version-bump: A changed plugin must bump its "
+    ".claude-plugin/plugin.json version. (0.10 seconds)\n"
+)
+
+def write(name, payload):
+    with open(os.path.join(fixture, name), "w") as fh:
+        json.dump(payload, fh)
+
+def post_commit(response):
+    return {"session_id": session, "hook_event_name": "PostToolUse",
+            "tool_name": "Bash",
+            "tool_input": {"command": "git commit -m 'wip (ASK-214)'"},
+            "tool_response": response}
+
+write("refused.json", post_commit(
+    {"stdout": "", "stderr": LEFTHOOK_REFUSAL, "error": "Exit code 1"}))
+write("noop.json", post_commit(
+    {"stdout": "On branch main\nnothing to commit, working tree clean",
+     "stderr": "", "error": "Exit code 1"}))
+write("landed.json", post_commit(
+    {"stdout": "[sana/ask-215 abc1234] fix (ASK-215)\n 1 file changed", "stderr": ""}))
+EOF
+}
+mk_payloads
+
+seed_ceiling() { # reset to a clean at-the-ceiling cache, no grace recorded
+python3 - "$CACHE" "$SESSION" <<'EOF'
+import json, sys, time
+json.dump({
+    "actor_key": sys.argv[2], "tool_calls_since_user": 50,
+    "agent_calls_since_user": 0, "mcp_timestamps": [], "repeat_map": {},
+    "consecutive_reads": 0, "warnings_issued": 1, "file_read_counts": {},
+    "greps_since_write": 0, "edit_targets": {}, "agents_without_write": 0,
+    "last_write_time": time.time(), "calls_since_write": 1,
+    "last_volume_reset": time.time(),
+}, open(sys.argv[1], "w"))
+EOF
+}
+
+# Each probe Edit targets a DISTINCT path so neither the exact-retry hash nor the
+# edit-spiral counter fires -- this test is about the volume ceiling only.
+ceiling_edit() { # $1 = probe index -> exit code in CODE, stderr in ERR
+  local payload
+  payload="{\"session_id\": \"$SESSION\", \"hook_event_name\": \"PreToolUse\", \"tool_name\": \"Edit\", \"tool_input\": {\"file_path\": \"/tmp/gate-probe-$1.json\", \"old_string\": \"version-$1\", \"new_string\": \"bumped-$1\"}}"
+  set +e
+  ERR="$(printf '%s' "$payload" | CLAUDECODE=1 CLAUDE_PROJECT_DIR="$FIXTURE" python3 "$GUARD" 2>&1 >/dev/null)"
+  CODE=$?
+  set -e
+}
+
+# --- 7) A gate-refused commit grants the remediation budget ------------------
+seed_ceiling
+run_guard_file "$FIXTURE/refused.json" >/dev/null \
+  || { echo "FAIL: PostToolUse gate-refusal leg exited non-zero"; exit 1; }
+ceiling_edit 1
+[ "$CODE" -eq 0 ] || { echo "FAIL: Edit after a gate-refused commit exited $CODE, want 0 (deadlock): $ERR"; exit 1; }
+echo "ok: a gate-refused commit grants a bounded budget so the ceiling is escapable"
+
+# --- 8) A no-op commit must NOT mint budget ---------------------------------
+# Otherwise an agent clears the ceiling forever with an empty commit.
+seed_ceiling
+run_guard_file "$FIXTURE/noop.json" >/dev/null \
+  || { echo "FAIL: PostToolUse no-op commit leg exited non-zero"; exit 1; }
+ceiling_edit 2
+[ "$CODE" -eq 2 ] || { echo "FAIL: 'nothing to commit' minted grace (Edit exited $CODE, want 2)"; exit 1; }
+case "$ERR" in
+  *"plugin-version-bump"*) echo "FAIL: no-op commit recorded a refusing gate: $ERR"; exit 1;;
+  *"tool calls"*) echo "ok: a no-op commit mints no budget; the ceiling still blocks";;
+  *) echo "FAIL: no-op path blocked with the wrong reason: $ERR"; exit 1;;
+esac
+
+# --- 9) Bounded: the call after the budget is spent blocks, naming the gate --
+seed_ceiling
+run_guard_file "$FIXTURE/refused.json" >/dev/null
+GRACE="$(python3 -c "import re,sys; print(re.search(r'^GATE_GRACE = (\d+)', open(sys.argv[1]).read(), re.M).group(1))" "$GUARD")"
+probe=0
+while [ "$probe" -lt "$GRACE" ]; do
+  probe=$(( probe + 1 ))
+  ceiling_edit "9$probe"
+  [ "$CODE" -eq 0 ] || { echo "FAIL: grace call $probe/$GRACE blocked early (exit $CODE): $ERR"; exit 1; }
+done
+ceiling_edit "9over"
+[ "$CODE" -eq 2 ] || { echo "FAIL: call $(( GRACE + 1 )) after the grant exited $CODE, want 2 (budget is unbounded)"; exit 1; }
+case "$ERR" in
+  *plugin-version-bump*) echo "ok: budget is bounded at $GRACE and the hard stop names the refusing gate";;
+  *) echo "FAIL: spent-budget block does not name the refusing gate: $ERR"; exit 1;;
+esac
+
+# --- 10) A landed commit clears the budget (no stale grace at a later ceiling) --
+seed_ceiling
+run_guard_file "$FIXTURE/refused.json" >/dev/null
+run_guard_file "$FIXTURE/landed.json" >/dev/null \
+  || { echo "FAIL: PostToolUse landed-commit leg exited non-zero"; exit 1; }
+python3 - "$CACHE" <<'EOF'
+import json, sys
+cache = json.load(open(sys.argv[1]))
+assert cache.get("gate_grace_remaining", 0) == 0, \
+    f"a landed commit must clear the grace budget: {cache.get('gate_grace_remaining')}"
+assert not cache.get("gate_grace_gate"), \
+    f"a landed commit must clear the recorded gate: {cache.get('gate_grace_gate')}"
+EOF
+echo "ok: a landed commit clears the budget"
+
+echo "PASS: token-guard hook behavior (warn shape + PostToolUse edit reset + blocked-attempt un-count + gate-refusal grace)"
--- a/q-system/.q-system/scripts/test/test-token-guard-hook-behavior.sh
+++ b/q-system/.q-system/scripts/test/test-token-guard-hook-behavior.sh
@@ -438,8 +438,19 @@ import importlib.util, os, subprocess, sys, tempfile
 spec = importlib.util.spec_from_file_location("tg", sys.argv[1])
 tg = importlib.util.module_from_spec(spec)
 spec.loader.exec_module(tg)
-repo = os.path.dirname(os.path.dirname(os.path.dirname(os.path.dirname(sys.argv[1]))))
+# THREE dirnames: $GUARD is <repo>/q-system/.q-system/token-guard.py, so the
+# repo root is three levels up, not four. Four resolved to the repo's PARENT and
+# the gate path did not exist, so this check never ran its producer at all -- it
+# graded python3's own "can't open file" message. It passed on macOS (that text
+# starts with the interpreter's absolute path, which falls through to the generic
+# label) and failed on Linux CI (which prints "python3: can't open file", so the
+# reader took 'python3' as the gate name). A green test that never reached the
+# code under test is the exact defect class this suite exists to catch.
+repo = os.path.dirname(os.path.dirname(os.path.dirname(sys.argv[1])))
 gate = os.path.join(repo, "q-system/.q-system/scripts/linear-issue-ref-check.py")
+# Fail LOUDLY if the producer is missing, instead of silently grading the
+# interpreter's error text the way the 4-dirname version did.
+assert os.path.exists(gate), f"fixture cannot reach the real commit-msg gate: {gate}"
 msg = os.path.join(tempfile.mkdtemp(), "msg.txt")
 open(msg, "w").write("chore: no issue id here\n")
 run = subprocess.run(["python3", gate, msg], capture_output=True, text=True)
--- a/q-system/.q-system/scripts/test/test-token-guard-hook-behavior.sh
+++ b/q-system/.q-system/scripts/test/test-token-guard-hook-behavior.sh
@@ -226,6 +226,48 @@ write("noop.json", post_commit(
      "stderr": "", "error": "Exit code 1"}))
 write("landed.json", post_commit(
     {"stdout": "[sana/ask-215 abc1234] fix (ASK-215)\n 1 file changed", "stderr": ""}))
+
+# --- Real transient git failures (captured 2026-07-28 by
+# q-system/output/capture-git-transient-failures.sh). Every one of these is git
+# itself failing at exit 128; a hook refusal is exit 1. They are NOT a gate.
+INDEX_LOCK = (
+    "fatal: Unable to create '/tmp/repo/.git/index.lock': File exists.\n\n"
+    "Another git process seems to be running in this repository, or the lock "
+    "file may be stale\n")
+REF_LOCK = (
+    "fatal: cannot lock ref 'HEAD': Unable to create "
+    "'/tmp/repo/.git/refs/heads/main.lock': File exists.\n\n"
+    "Another git process seems to be running in this repository, or the lock "
+    "file may be stale\n")
+GPG_FAIL = (
+    "fatal: cannot exec '/usr/bin/gpg': No such file or directory\n"
+    "error: gpg failed to sign the data:\n(no gpg output)\n"
+    "fatal: failed to write commit object\n")
+
+write("indexlock.json", post_commit(
+    {"stdout": "", "stderr": INDEX_LOCK, "error": "Exit code 128"}))
+write("reflock.json", post_commit(
+    {"stdout": "", "stderr": REF_LOCK, "error": "Exit code 128"}))
+write("gpgfail.json", post_commit(
+    {"stdout": "", "stderr": GPG_FAIL, "error": "Exit code 128"}))
+# A git-self failure whose text is NOT in the phrase list. Only the exit code
+# separates it from a gate refusal, so this pins the exit-128 rule on its own.
+write("unknown128.json", post_commit(
+    {"stdout": "", "stderr": "fatal: a failure mode nobody enumerated\n",
+     "error": "Exit code 128"}))
+
+# A failing command that merely MENTIONS the string "git commit" and never
+# invokes it. Real producer: a grep over the canonical corpus with no match.
+write("mentions.json", {
+    "session_id": session, "hook_event_name": "PostToolUse", "tool_name": "Bash",
+    "tool_input": {"command": 'grep -rn "git commit" q-system/canonical/'},
+    "tool_response": {"stdout": "", "stderr": "", "error": "Exit code 1"}})
+
+# The same lefthook refusal delivered with an integer exit_code and NO error
+# key. The Bash response shape is not pinned by contract, so both readers must
+# work; without this fixture the exit_code branch is untested.
+write("refused_exitcode.json", post_commit(
+    {"stdout": "", "stderr": LEFTHOOK_REFUSAL, "exit_code": 1}))
 EOF
 }
 mk_payloads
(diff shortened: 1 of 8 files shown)


assafkip:
Cleanup blocked by the destructive-op hook (expected). Leaving the scratch dir; removal command at the bottom.

---

# Review: PR #27 — token-guard gate-refusal grace (ASK-215)

Blast radius note first: `q-system/.q-system/token-guard.py` ships to every registered instance via `kipi-update.sh:1277` (`rsync -a --delete "$ARCHIVE_TMP/q-system/"`). This is a fleet-wide circuit breaker, not one repo's.

---

## FINDING 1 — major

**A transient git failure is classified as a pre-commit gate refusal, and the run hard-stops with a fabricated diagnosis.**

`q-system/.q-system/token-guard.py:487` (`NON_GATE_COMMIT_FAILURES`) and `:550` (`_commit_gate_refusal`)

`index.lock` is not in the negative list. This repo runs a 15-minute auto-committer (`AUTO_COMMIT_SUBJECT = "chore: update project files"`, live at `fbc9404`) and the fleet runs parallel sessions against one checkout. A session commit racing either one fails with `index.lock`, which a retry clears.

```
$ bash .pr27rev/repro4-indexlock.sh
gate_grace_remaining = 8
gate_grace_gate      = 'a pre-commit gate'
9th call exit=2
operator sees: a pre-commit gate refused the checkpoint and the 8-call grace budget is spent. Stop. Report what a pre-commit gate is asking for, what you tried, and that the work is staged and uncommitted.
```

At 3am the run burns 8 calls chasing a gate that does not exist, then stops and tells the human to go find what "a pre-commit gate" wants. The real cause was a lock. Two of the three claims in that message are false: nothing refused the checkpoint, and the state is not "staged and uncommitted" in the way the message implies.

Fix shape: add `index.lock`, `another git process`, `cannot lock ref`, `gpg failed to sign` to `NON_GATE_COMMIT_FAILURES`, or require a positive gate signal instead of a negative list.

---

## FINDING 2 — minor

**A failing Bash command that never touched git mints the budget and hijacks the ceiling message for the rest of the user-message window.**

`q-system/.q-system/token-guard.py:557` — the test is `"git commit" not in command`, a substring match on the command text, not "this call is a commit".

Producer: `grep -rn "git commit" q-system/canonical/` with zero matches exits 1. I ran one live in this session (`echo "...git commit -m x"; false`).

```
$ bash .pr27rev/repro1-bogus-gate.sh
=============== BASE (main) ===============
exit=2
msg: 50 tool calls without user input or a commit. Commit finished work now (git commit is exempt from this ceiling and resets it), or stop and summarize what you've accomplished and what's remaining.

=============== PR #27 ===============
gate_grace_remaining = 8
gate_grace_gate      = 'a pre-commit gate'
-- 8 grace calls consumed (each one an Edit the ceiling would have blocked on main) --
exit=2
msg: a pre-commit gate refused the checkpoint and the 8-call grace budget is spent. Stop. Report what a pre-commit gate is asking for, what you tried, and that the work is staged and uncommitted.
```

`gate_grace_gate` is set unconditionally at `:713`, including when the grant is refused, and `check_volume:286` then serves the gate message for every later ceiling in that window. No commit moved HEAD, so wiring B never rescues it. Same root as Finding 1, separate fix.

---

## FINDING 3 — minor

**During the 8 grace calls, the subagent ceiling, MCP rate limit, and edit-spiral blocks are unreachable.**

`q-system/.q-system/token-guard.py:283` returns `("warn", ...)`; `main()` calls `warn()`, which prints and `sys.exit(0)` before checks 4 through 11 run.

```
$ bash .pr27rev/repro2-detectors-off.sh
AGENT_CEILING=30, cache has agent_calls_since_user=40 (10 over)
EDIT_FAIL_LIMIT=3, cache has 9 failed edits on /tmp/stuck.py

--- gate_grace_remaining = 0 ---
  Agent spawn  -> exit=2  :: 50 tool calls without user input or a commit. Commit finished work now (git commit is exempt fr
  spiral Edit  -> exit=2  :: 50 tool calls without user input or a commit. Commit finished work now (git commit is exempt fr
--- gate_grace_remaining = 8 ---
  Agent spawn  -> exit=0  ::
  spiral Edit  -> exit=0  ::
```

The warn text says "Do nothing else with these calls." That is prompt-only enforcement, in a repo whose CLAUDE.md bans it. An agent spiraling on the gate file gets 8 unchecked edit attempts and never hears "your edit approach is wrong", then gets a block that blames the gate. Bounded at 8, recoverable.

---

## FINDING 4 — minor

**The fallback gate-name reader names this repo's own commit-msg gate `subject`.**

`q-system/.q-system/token-guard.py:541-546`

The PR's own docstring states git prints only the hook's output with no marker. Without lefthook's glove summary in the captured text, the second reader runs. Fed the real output of `q-system/.q-system/scripts/linear-issue-ref-check.py`:

```
$ python3 q-system/.q-system/scripts/linear-issue-ref-check.py .pr27rev/msg.txt
exit=1
BLOCK: this commit names no Linear issue.
  subject: chore: no issue id here
  ...

$ python3 -c "...; print(tg._refusing_gate_name(open('gate-stderr.txt').read()))"
REAL repo commit-msg gate output -> 'subject'
```

`BLOCK` is correctly excluded by `_NOT_A_GATE_NAME`; the next line's `subject:` is not. Block message becomes "subject refused the checkpoint". Cosmetic, no control-flow change.

---

## FINDING 5 — nit

`_response_text_and_code` (`:515`) reads `exit_code`/`returncode`, but no test exercises that branch, and no fixture in the suite sets those keys. Dead-but-harmless (see below).

---

## What is sound

Attacks I ran that the code survived. All output real.

**The fixture shape. I tried hardest here and was wrong.** Session transcripts show `toolUseResult` for Bash has keys `('interrupted','isImage','noOutputExpected','stderr','stdout')`, and `toolUseResult dicts with an 'error' key: 0 / 216`. My hypothesis was that the `"error": "Exit code 1"` fixture was invented and the whole feature was inert. The live control experiment killed it:

```
tool_calls_since_user before                                        = 35
after a FAILING command containing "git commit"                     = 37   (no reset)
after a SUCCEEDING command containing "git commit"                  = 1    (reset to 0)
```

Only a truthy `error` in the hook payload explains that pair, since `_is_successful_commit` gates on exactly that key. The hook payload is not the transcript record. Fixture shape is real. Finding dropped.

**Everything the PR body claims, verified independently** (`.pr27rev/repro3-attacks-that-failed.sh`):

```
A) headline case: real lefthook refusal at the ceiling -> Edit allowed?
   exit=0  grace_left=7   (want exit 0)
B) ratchet: refuse, spend all 8, refuse AGAIN -> topped back up to 8?
   grace after 2nd refusal = 0  (want 0)
   next call exit=2 :: gitleaks refused the checkpoint and the 8-call grace budget is spent.
C) --dry-run commit failure mints budget?      grace=None  (want 0/None)
D) no-op commit mints budget?                  grace=None  (want 0/None)
E) landed commit clears budget?                grace=0     (want 0)
F) grace never granted below the ceiling ->    exit=0  grace_left=8 (unspent)
```

The no-ratchet bound holds. The gate name updates on a second refusal without topping up, exactly as documented. Budget is spent only at the ceiling.

**Gate-name extraction on garbage input.** I expected a URL line to yield `https`. It does not, the space check catches it:

```
git lock failure         -> 'a pre-commit gate'
gpg sign failure         -> 'a pre-commit gate'
gate output w/ a URL     -> 'a pre-commit gate'
empty output             -> 'a pre-commit gate'
```

Dropped.

**Tests 7 through 10 can fail.** Each one has a plausible mutation that turns it red (remove the grant, drop `nothing to commit` from the list, remove the grants cap, remove `clear_gate_grace`). Not decoration.

---

## VERDICT: REQUEST CHANGES

One major. Not BLOCK: nothing here is permanent or unrecoverable, no Linear object is written, and a revert restores main's behavior exactly.

**Fix first:** Finding 1. Add the transient-git failures to `NON_GATE_COMMIT_FAILURES` so a lock collision with your own auto-committer stops being reported as a gate refusal. That single change also narrows Finding 2's surface.

The design is right and the bounds hold. The classifier is the weak part, and it is weak on the failure this fleet actually produces on a timer.

---

## Cleanup

I left `.pr27rev/` (scratch) and ref `refs/pr27-review` behind. The destructive-op hook refused my `rm -f`, correctly. To clear:

```bash
ALLOW_DESTRUCTIVE=1 rm -rf .pr27rev && git update-ref -d refs/pr27-review
```

No repo files were modified.

```
FINDINGS:
major|A transient git failure (index.lock from the repo's own auto-committer or a parallel session) is classified as a pre-commit gate refusal, burning 8 calls and hard-stopping the run with a fabricated diagnosis|q-system/.q-system/token-guard.py:487
minor|Any failing Bash command whose text merely contains the substring "git commit" mints the grace budget and hijacks the volume-ceiling block message for the rest of the user-message window|q-system/.q-system/token-guard.py:557
minor|The grace warn path exits before checks 4-11, so the subagent ceiling, MCP rate limit, and edit-spiral blocks are unreachable for all 8 grace calls|q-system/.q-system/token-guard.py:283
minor|The fallback gate-name reader extracts "subject" from this repo's own linear-issue-ref-check output when lefthook's summary is absent|q-system/.q-system/token-guard.py:541
nit|No test exercises the exit_code/returncode branch of _response_text_and_code; every fixture sets only the error key|q-system/.q-system/token-guard.py:515
END FINDINGS
```

assafkip:
# Round 2 reply — all five findings fixed, each with a case that fails without the fix

Thanks for the index.lock finding. It is the one this fleet actually produces on
a timer, and you are right that the block message was lying about two of its
three claims.

Head: `bf641ad`. Suite goes 10 → 16 cases, extended in place (no new test file,
no `capability-manifest.json` change).

## Ground truth first

I did not want to fix a classifier by adding phrases I remembered. New capture
script, committed as the provenance the code comments cite:
`q-system/output/capture-git-transient-failures.sh`.

```
===== 1. index.lock held =====
exit_code=128
fatal: Unable to create '.../.git/index.lock': File exists.
Another git process seems to be running in this repository, or the lock file may be stale

===== 2. gpg signing failure =====
exit_code=128
error: gpg failed to sign the data:
fatal: failed to write commit object

===== 3. pre-commit hook REFUSAL =====
exit_code=1
BLOCK: bump plugin.json

===== 5. cannot lock ref =====
exit_code=128
fatal: cannot lock ref 'HEAD': Unable to create '.../refs/heads/master.lock': File exists.

===== 6. hook exit-code propagation =====
hook exit 1   -> git exit 1
hook exit 2   -> git exit 1
hook exit 3   -> git exit 1
hook exit 42  -> git exit 1
hook exit 128 -> git exit 1
```

Two facts fall out. **git's own failures exit 128**, and **git normalises every
hook refusal to exit 1** whatever the hook returned. So 128 is the near-positive
signal you asked for: it is git failing, never a gate refusing, and it covers
transient failures whose text nobody has enumerated yet. A fully positive signal
still does not exist (exit 1 is also `nothing to commit`), so the negative list
stays — it is just no longer the only thing holding the line.

## FINDING 1 (major) — fixed

`NON_GATE_COMMIT_FAILURES` gains `index.lock`, `another git process`,
`cannot lock ref`, `gpg failed to sign`, `failed to write commit object`, and
`_commit_gate_refusal` now returns `None` on exit 128 regardless of text.

**Layer above.** A lock collision now mints no budget, so the ceiling serves its
generic message instead. That is not a hole: the recovery from a lock is to retry
`git commit`, which the ceiling already exempts, so the retry needs no budget.
Deadlock-free either way.

**Case 11** (four probes: index.lock, ref-lock, gpg, and an unenumerated failure
that only exit 128 separates from a gate). Observed RED:

```
AssertionError: indexlock: a transient git failure minted budget: 8
```

## FINDING 2 (minor) — fixed

New `_invokes_git_commit()` tokenises with `shlex` instead of substring-matching,
so `grep -rn "git commit" canonical/` no longer reads as a commit while
`cd x && git commit -m y` and `git -c user.name=x commit` still do.

**Two readers, deliberately.** I did **not** point `_is_commit_command` at it.
There a false negative blocks the checkpoint the ceiling is asking for and
deadlocks the run; a false positive only exempts one harmless call. For the grant
and for `_is_successful_commit` the costs invert. That asymmetry is now written
into the docstring so the next reader does not "unify" them.

**Case 12.** Observed RED (probe against the unfixed guard):

```
F2 failing grep      -> 'a pre-commit gate'
F2 grep resets vol   -> True
```

That second line is the same defect in `_is_successful_commit`, one function up,
which you did not flag: a *succeeding* grep that mentions `git commit` reset the
volume ceiling outright. Fixed in the same change — the tightening is safe
because wiring B (`reset_volume_if_committed`, HEAD epoch) still catches a real
commit whose command form the tokeniser misses.

## FINDING 3 (minor) — fixed

The volume warning is now **held**, not emitted inline, so checks 4-11 run. A
later block outranks it; a later warning carries it.

**What I did not do:** buy reachability by making the ceiling warning silent.
**Case 16** pins that — it asserts the emitted context still opens with the
volume warning and still carries the later one. Without the carry it passes as
"reachable" while the operator loses the 35-call warning.

**Case 13** seeds `edit_targets` over `EDIT_FAIL_LIMIT` with a live grace budget
and requires exit 2 with the edit-spiral message. On the old code that Edit
exited 0, exactly as your `repro2-detectors-off.sh` showed.

## FINDING 4 (minor) — fixed

The fallback reader now only considers **unindented** lines. A gate prints its
name at the left margin and indents the detail beneath it. Structural rule, not
another word added to `_NOT_A_GATE_NAME`.

**Case 14** runs the real `linear-issue-ref-check.py` (not a remembered fixture)
and asserts the generic label. Observed RED / GREEN:

```
F4 real gate name    -> 'subject'            # before
F4 real gate name    -> 'a pre-commit gate'  # after
```

## FINDING 5 (nit) — fixed, and it was hiding something

**Case 15** delivers the lefthook refusal as `exit_code: 1` with no `error` key.
It went red for a reason I did not expect:

```
AssertionError: a refusal delivered as exit_code (no error key) minted no budget
```

`_is_successful_commit` never read the exit code at all — only `error`. So that
one response was a gate refusal to `_commit_gate_refusal` and a **landed commit**
to `_is_successful_commit`: it reset the ceiling and cleared the budget it had
just minted, one branch earlier in the same `if/elif` chain. Two readers of the
same input with different semantics, which is what your nit was pointing at
without either of us seeing the consequence. Fixed in `80b4eec`.

Narrower than `sp-1078fbe2` (response with **no** failure signal at all). That
one stays captured, not fixed here.

## What got quieter (stated, per your bar)

One thing: a transient git failure no longer produces
`<gate> refused the checkpoint`. It produces the generic ceiling message instead.
That is the point of finding 1 — the old line was false in two of three clauses
— and nothing else lost a signal. Findings 3 and 5 both make the guard *louder*
(three detectors come back online during grace; a refused commit no longer
resets the ceiling).

## Not regressed

Your `repro3-attacks-that-failed.sh` cases A-F are cases 7-10 in the suite and
all still pass, including the no-ratchet bound and "budget spent only at the
ceiling".

## Checks

```
$ bash q-system/.q-system/scripts/test/test-token-guard-hook-behavior.sh
... 16 ok lines ...
PASS: token-guard hook behavior (warn shape + PostToolUse edit reset +
blocked-attempt un-count + gate-refusal grace + non-gate classifier)

$ python3 -m pytest q-system/.q-system/tests/test_token_guard.py -q
14 passed in 0.50s

tests: ran=75 quarantined=0 skipped-skeleton-only=0
capability-gate: GREEN
```

Nothing in the DoR's Not-doing list was touched: no threshold changed, no
lefthook gate changed, `converge.sh` / `linear-worker.sh` / `pr-verdict-lib.sh`
untouched. The finding-3 fix moves *when* the volume warning is emitted; it does
not change the edit-spiral / read-spiral / grep-drift / stall detectors
themselves, it stops the grace path from hiding them.

## Housekeeping

The scratch you left is not in this worktree (`.pr27rev/` and `refs/pr27-review`
live in whichever checkout you reviewed from), so I have not touched it.

assafkip:
Unstaged — the repo's auto-committer had staged my 18MB scratch tar; it is untracked again. No tracked file touched.

---

# Review: PR #27 — token-guard gate-refusal grace (ASK-215), round 2

Head `bf641ad`. I read the code first, ran my own repros, then read rounds 1–2.

Baseline: the 16-case suite is green on my extraction of head, and all five round-1 findings are genuinely fixed — I falsified none of the author's citations. Both findings below are on **code this round introduced**, not re-litigation.

---

## FINDING 1 — minor

**The held volume warning is destroyed, not deferred, when a later check blocks — and `warnings_issued` was already set, so it never fires again in that window.**

`q-system/.q-system/token-guard.py:862` (`pending_warning = msg`), consumed at `:293` (`cache["warnings_issued"] = 1`), emitted only at `:917`.

The finding-3 fix defers the 35-call warning so checks 4–11 stay reachable. Checks 4, 5 and 9 `block()`, which exits before line 917. `check_volume` already flipped `warnings_issued` to 1 and that write is saved, so the warning is gone for the rest of the user-message window.

Reproducer — one cache at 34 calls with `edit_targets` at `EDIT_FAIL_LIMIT`, same guard invocation, main vs head:

```
$ bash .pr27rev/r2/reproA-held-warning-lost.sh
=== main ===
exit=0
stdout: {"hookSpecificOutput": {"hookEventName": "PreToolUse", "additionalContext": "You've made 35 tool calls since the last user message. You have 15 remaining before hard stop. Committing finished work res
cache: tool_calls_since_user=35 warnings_issued=1
next call in same window: exit=0 stdout=<none>

=== prhead ===
exit=2
stdout:
stderr: 4 edit attempts on stuck.py. The approach isn't working. Read the file again, find the exact string, or tell the founder what's wrong.
cache: tool_calls_since_user=35 warnings_issued=1
next call in same window: exit=0 stdout=<none>
```

The spiral block firing is the intended fix — that part is right. The defect is the empty `stdout` on head: the run goes 35 → 50 with zero heads-up and then hard-stops.

**Falsifying a citation.** The round-2 reply says *"Case 16 pins that — it asserts the emitted context still opens with the volume warning."* Case 16 only exercises warn → warn (read-spiral). Nothing exercises warn → block. My M1 mutation confirms case 16's actual scope: it catches the carry being dropped, and it stays green through this path.

Fix shape: emit the held warning on the block path too (prepend to the block message), or don't set `warnings_issued` until the warning is actually delivered.

---

## FINDING 2 — minor

**At the ceiling, a call that spends a grace unit and is then blocked by check 4/5/9 never executes, but the grace unit is not refunded — 8 blocked edits consume the entire remediation budget with zero remediation done.**

`q-system/.q-system/token-guard.py:286` (spend) vs `:398` (`uncount_blocked_attempt`, which refunds `repeat_map` and `edit_targets` but not `gate_grace_remaining`).

Reachable only because the finding-3 fix made those blocks reachable during grace. The agent is by definition editing the file the gate demands, so an edit spiral on that file is the likely collision.

```
$ python3 .pr27rev/r2/reproBC.py
=== B) blocked calls burn the gate-remediation budget ===
GATE_GRACE=8  EDIT_FAIL_LIMIT=3
  call 1: exit=2 grace_left=7 :: 4 edit attempts on plugin.json. The approach isn't working. Read the file again, find the
  ...
  call 8: exit=2 grace_left=0 :: 4 edit attempts on plugin.json. The approach isn't working. Read the file again, find the
  call 9: exit=2 grace_left=0 :: plugin-version-bump refused the checkpoint and the 8-call grace budget is spent. Stop. Rep
  remediation calls that actually EXECUTED: 0 of 8 granted
```

At 3am the operator is told the 8-call budget is spent. Zero calls ran.

**Why I am not calling this a major.** The outcome is main's outcome — stop with work staged and uncommitted, human finishes the commit by hand. It is an incomplete fix, not new damage, and it is bounded at 8. Round 1's "fabricated diagnosis" was major because the message named a gate that did not exist; here the gate is real and the message is only misleading about *why* the budget went.

Fix shape: refund the grace unit inside `uncount_blocked_attempt`, same principle already applied to `repeat_map`.

---

## What is sound

Attacks I ran that the code survived. Every line below is real output.

**Are the round-2 tests decoration?** No. Four mutations, four caught:

```
$ bash .pr27rev/r2/mutate.sh
M1 -> caught: AssertionError: the held volume warning was dropped by a later warning
M2 -> caught: AssertionError: unknown128: a transient git failure minted budget: 8
M3 -> caught: AssertionError: a failing grep minted grace: 8
M4 -> caught: FAIL: edit-spiral is unreachable during the grace budget (exit 0, want 2)
```

**The gate-name reader against this repo's real gate set** (round-1 finding 4). I expected a bogus name somewhere. There isn't one — gitleaks was my best shot because it prints unindented colon-delimited lines, and the real binary against a real staged canary still lands on the fallback:

```
$ python3 .pr27rev/r2/reproE-gitleaks-name.py
gitleaks exit=1, 3 lines of real output
  | 5:20PM INF 0 commits scanned.
  | 5:20PM WRN leaks found: 1
--- _refusing_gate_name(real gitleaks output) ---   -> 'a pre-commit gate'
--- with the lefthook summary appended ---          -> 'gitleaks'

$ python3 .pr27rev/r2/reproD-real-gate-names.py
  OK  linear-issue-ref  exit=1 -> 'a pre-commit gate'
  OK  gitleaks / large-files / settings-template-sync / instruction-budget (glove reader) -> exact names
```
Caveat stated honestly: only `linear-issue-ref` actually refused; the other python gates exited 0 with nothing staged, so their refusal *text* was not exercised bare — only their lefthook summary shape was.

**The `shlex` tokeniser losing a real commit.** This was my main regression hypothesis and it failed:

```
$ python3 .pr27rev/r2/reproBC.py
  git -C                     new=True  main=False   <-- CHANGED vs main   (improvement)
  heredoc (harness default)  new=True  main=True
  msg with a #               new=True  main=True
  subshell bash -c           new=False main=True    <-- CHANGED vs main
=== C2) real commit of the harness's default heredoc shape ===
  real `git commit` exit=0 stdout='[master (root-commit) 5db90c8] fix(token-guard): a thing (ASK-215)'
  _is_successful_commit -> True
```
`bash -c 'git commit …'` is the one real behavior loss, and wiring B (`reset_volume_if_committed`, HEAD epoch) catches it — the author's cited mitigation holds. No producer in this repo emits that form. Dropped.

**A `git commit` that hits the Bash 2-minute timeout** (no exit code, no matching phrase, so it would classify as a gate refusal — the documented gitleaks-on-a-big-staged-set hang). I could not obtain the real timeout `tool_response` shape, and constructing one would be exactly the invented-fixture defect I am here to catch. Dropped, not softened.

**Round-2 process calibration.** This is the healthy shape, not the bad one: round 2 found nothing new on unchanged code, and both surviving findings are consequences of the round-2 fixes themselves. Round 1 was calibrated correctly.

---

## VERDICT: APPROVE WITH NITS

No blocker, no major. Nothing here is permanent; the file ships fleet-wide via `kipi-update.sh` rsync, and a revert restores main exactly.

**Fix first:** Finding 1. It is the smaller change and it closes the gap the author's own case 16 was written to close — the ceiling warning going silent is the one trade the reply explicitly promised not to make.

Scratch left at `.pr27rev/r2/` (untracked; the repo's auto-committer staged it once and I unstaged it). To clear: `ALLOW_DESTRUCTIVE=1 rm -rf .pr27rev/r2`.

```
FINDINGS:
minor|The held volume warning is destroyed rather than deferred when a later check blocks, and warnings_issued is already set so it never fires again in that window|q-system/.q-system/token-guard.py:862
minor|A grace call spent at the ceiling is not refunded when a later check blocks the same call, so 8 blocked edits consume the whole remediation budget with zero calls executed|q-system/.q-system/token-guard.py:286
END FINDINGS
```

assafkip:
Reviewed at head `c063c3d`. Here is my read.

## Findings

---

**minor — the grace budget is spent by calls the guard then BLOCKS, so the escape hatch drains on calls that never executed**
`q-system/.q-system/token-guard.py:286`

`check_volume` decrements `gate_grace_remaining` before checks 4/5/9 run. Those checks call `uncount_blocked_attempt` (which refunds `repeat_map` and `edit_targets`) but nothing refunds grace. The realistic path is the worst one: the file the agent is editing to satisfy the gate is the file already in an edit spiral.

Repro: `python3 .pr27rev/r3/repro.py` (section B)

```
B) grace spent by calls the guard BLOCKS (they never execute)
   grace after the grant           : 8
   blocked edit #1 -> exit 2, grace now 7 | 4 edit attempts on gate-file.py. The approach isn't working. Rea
   blocked edit #2 -> exit 2, grace now 6 | 4 edit attempts on gate-file.py. The approach isn't working. Rea
   blocked edit #3 -> exit 2, grace now 5 | 4 edit attempts on gate-file.py. The approach isn't working. Rea
   blocked edit #4 -> exit 2, grace now 4 | 4 edit attempts on gate-file.py. The approach isn't working. Rea
   -> 4 Edits that NEVER RAN consumed 4 of the 8 call escape hatch
```

Round 2 raised this (`token-guard.py:286`). Still live at head, repro above. Severity held at round 2's minor.

---

**minor — the held volume warning is consumed but never delivered when a later check blocks, and it can never fire again in that window**
`q-system/.q-system/token-guard.py:293`

`check_volume` sets `warnings_issued = 1` at the moment it *builds* the warning, not when it is delivered. The finding-3 refactor moved delivery to the end of `main()`, so a block between the two destroys the message while the flag stays set. `check_volume` only emits the 35-call warning while `warnings_issued == 0`, so it is gone for the rest of the window.

Repro: `python3 .pr27rev/r3/repro2.py`

```
call at the 35-call warning threshold, edit-spiral also tripped
  exit          : 2
  stderr(block) : 4 edit attempts on stuck.py. The approach isn't working. Read the file again, find the exact string, or tell the founder
  stdout(warn)  : ''
  warnings_issued after: 1

next clean call at 38 calls:
  exit   : 0
  stdout : ''
  -> no '15 remaining before hard stop' warning; it is gone for good.
```

Round 2 raised this (`token-guard.py:862`). Still live at head, repro above. Severity held.

---

**minor — a gate refusal never expires, so one recorded early in the window re-labels an unrelated 50-call spin as a gate deadlock and widens the ceiling to 58**
`q-system/.q-system/token-guard.py:652`

`grant_gate_grace` records the gate at any call count and nothing ages it out. The docstring at `check_volume:277` claims "granting it eagerly (on any refused commit, at any call count) costs nothing: a run that never reaches the ceiling never touches it." That is true for the run that never reaches the ceiling. It is false for the run that reaches it 42 calls later for reasons unrelated to the gate.

Repro: `python3 .pr27rev/r3/repro.py` (section C)

```
C) a gate refusal early in the window hijacks a LATER, unrelated ceiling
   at call 8 a commit is refused; agent abandons committing
   ...42 unrelated calls later the run hits the ceiling:
      call 9 -> exit 2: large-files refused the checkpoint and the 8-call grace budget is spent. Stop. Report what large-files is aski
   -> ceiling widened 50 -> 58 for a run with no gate deadlock
```

This is the same underlying issue as round 1's finding 2 ("mints the grace budget and hijacks the volume-ceiling block message for the rest of the user-message window"). That fix narrowed *what counts as a refusal*; it did not bound *how long a recorded refusal stays authoritative*. Per the no-escalation rule I hold it at round 1's minor. One cheap shape: stamp `gate_grace_call` at grant time and drop the record if `tool_calls_since_user - gate_grace_call` exceeds some window.

---

**nit — `_invokes_git_commit`'s ValueError fallback reverts to the exact substring semantics finding 2 removed**
`q-system/.q-system/token-guard.py:472`

When `shlex.split` raises, the fallback is a bare `command.split()`, which re-admits a bare `git`/`commit` token pair from prose.

Repro: `python3 .pr27rev/r3/repro.py` (section D)

```
   fallback detail (shlex raises on the unbalanced quote):
      shlex ValueError: No closing quotation
      _invokes_git_commit -> True
      _commit_gate_refusal -> 'a pre-commit gate'
```

Severity dropped: I had to construct that input and found no producer in this repo that emits an unbalanced-quote command carrying `git commit`. Stated per the calibration rule.

---

**The fixture-shape attack, which is the one I expected to land.** The PR's whole classifier keys off `tool_response` being a dict with `error` and text. I read this project's real session transcripts and found the recorded result for a gate-refused commit is a *string*: `'Error: Exit code 1\n🥊 plugin-version-bump: ...'` — no `error` key, no `exit_code`, and across 375 Bash records **zero** carry `error`/`exit_code`/`returncode`. If that were what the hook sees, `_commit_gate_refusal` would return `None` forever and this feature would be dead code.

It is not what the hook sees. Two live oracles against the running main-branch guard:

1. A genuinely-failed `git commit` (gate refused, exit 1) did **not** reset `tool_calls_since_user` (36 → 41, `last_volume_reset` unchanged). The only branch in `_is_successful_commit` that returns False there is `isinstance(tool_response, dict) and tool_response.get("error")`. So the hook's payload **is** a dict with a truthy `error`.
2. `git commit -m "..." || echo "nothing to commit"` (exit 0) also did not reset, which requires `text` to have been read and matched. So the dict **does** carry `stdout`.

Both halves of the fixture shape match the real producer. Dropped.

**The gate-name reader on real output, not the author's fixture.** I ran `git commit` in kipi-system for real; lefthook v2.1.6 refused via `large-files`. Fed that captured text (ANSI codes, banner, `BLOCK:` line, `exit status 1┃` run-on and all) to the PR's reader:

```
A) REAL lefthook refusal captured from kipi-system (large-files)
   _refusing_gate_name -> 'large-files'
   _commit_gate_refusal -> 'large-files'
```

The banner-glove claim in the PR body is true against real output, and the `BLOCK:` detail line is correctly rejected.

**Command tokenising.** `git -C /repo commit`, `cd /repo && git commit`, the `$(cat <<'EOF' ... )` heredoc commit form, `grep -rn "git commit"`, `--dry-run` — all classify correctly.

**Also survived:** exit-128 transient exclusions, `nothing to commit` minting nothing, a landed commit clearing the budget, the no-ratchet bound at `GATE_GRACE_GRANTS = 1`.

Severity is converging, not diverging: round 1 was 1 major + 3 minor + 1 nit, round 2 was 2 minor, this round is those same 2 minors plus 1 same-class minor and a nit. No new blocker on unchanged code. That is a well-calibrated chain — I have nothing to fault in the earlier rounds' framing.

The one thing worth naming: round 2 approved with two minors and **both are still live at head**, and neither appears in `.prd-os/spillover.jsonl` (250 entries, no match on grace/warning/refund). This repo's own `no-orphan-findings.md` says a real out-of-scope finding is captured, not mentioned. Worth confirming that path fires on APPROVE-WITH-NITS.

## VERDICT: APPROVE WITH NITS

Only minor/nit findings survived reproduction. Fix first: **`token-guard.py:293`** — move `warnings_issued = 1` to the point of delivery rather than construction. It is a two-line change and it is the one that silently removes an operator signal rather than merely bounding a budget.

Two operational notes, not findings:
- This checkout's index currently has `.pr27rev/r2/head.tar` (17MB) staged, so `large-files` refuses **every** commit here right now. That is the ASK-215 shape live in the working tree.
- My scratch is `.pr27rev/r3/` (untracked, same pattern as the r2 review). Nothing tracked was touched, nothing committed, nothing pushed.

```
FINDINGS:
minor|A grace call spent at the ceiling is not refunded when a later check blocks the same call, so blocked Edits that never executed drain the remediation budget|q-system/.q-system/token-guard.py:286
minor|check_volume sets warnings_issued=1 when it builds the volume warning rather than when it is delivered, so a later block destroys the 35-call warning permanently for that window|q-system/.q-system/token-guard.py:293
minor|A recorded gate refusal never expires, so one from early in the window widens a later unrelated ceiling from 50 to 58 and re-labels a plain spin as a gate deadlock|q-system/.q-system/token-guard.py:652
nit|_invokes_git_commit falls back to bare command.split() when shlex raises, restoring the substring semantics the tokeniser was added to remove|q-system/.q-system/token-guard.py:472
END FINDINGS
```

PR #23, opened 2026-07-27
fix(seam): a Linear-flow PR cannot merge without a prd-os receipt (ASK-210)

assafkip:
Closes step 2 of 3 on the prd-os / Linear seam. Step 1 (ASK-209, `531118e`) made the ledger drainable. This makes proof-of-done a merge requirement.

## The gap

`linear-worker.sh:342` correctly stops the agent at *"open a PR, do not merge — closeout runs through /issue-verify and /issue-closeout."* Nothing invoked that closeout. On 2026-07-27 seven agent PRs merged with **zero** entries in `.prd-os/receipts.jsonl` and no gate noticed. prd-os was capture-only for Linear work: findings flowed in, completion was never proved.

## What this adds

`q-system/.q-system/scripts/pr-receipt-gate.py` — resolves the PR's head branch (`--branch`, or `--pr N` via `gh`), maps `sana/ask-<n>` → `ASK-<n>`, and refuses the merge when no receipt in `.prd-os/receipts.jsonl` carries that id. It **writes nothing**; receipt production stays owned by kipi-dsse's `issue_runner`.

Wired as a PR-only step in `.github/workflows/validate.yml`, next to the plugin version-bump guard. CI passes `--branch "${{ github.head_ref }}"` so the gate needs neither `gh` nor a token.

## Bootstrap decision — and what it does NOT cover

**Only `sana/ask-<n>` branches are gated.** That is the shape `linear-worker.sh` creates and the exact population that merged unreceipted.

**Not gated, deliberately:** human PRs, chore branches, other agent prefixes, `main`. They pass with an explicit `not gated` line on stdout rather than silently — a gate whose blind spot is undocumented reads as full coverage. Pinned by three test cases so the blind spot cannot be widened or narrowed without the test noticing.

The cutoff-commit alternative was rejected: the cutoff is invisible from a branch name, and it would fail every currently-open PR including this one.

**This PR's own receipt:** it has none. Its branch *is* `sana/ask-210`, so the new step will fail on this PR by design until `/issue-verify` + `/issue-closeout` run at closeout. That is the gate working, not a defect.

## Matching rules

Word-bounded, case-insensitive, across any string field of a **well-formed** receipt record.

- *Any field* rather than `issue_id` only: the DSSE spec id is a slug (`ask-210-receipt-gate`) and no convention yet pins where the Linear id lands. Pinning one key would be brittle against a naming choice nobody has made.
- *Word-bounded* so `ask-9990` cannot satisfy `ask-999`.
[...]

PR #29, opened 2026-07-28
feat(review): the reviewer emits its verdict as a commit status on the sha it read, so the platform can gate on it (ASK-217)

assafkip:
Closes the last piece of the review loop that was stuck on disk. **Build item 5a** — producer only, nothing is made required here.

## The defect

The verdict is a **local file** (`~/.config/kipi/pr-reviews/pr-<N>.verdict.json`). GitHub cannot see it, so no platform mechanism can gate on it, so every approved PR ends its life waiting on a human. `pr-receipt-gate.py` is already a CI step, which is the right shape; the reviewer's verdict being a local file was the one piece that was wrong.

## The change

`pr-review-agent.sh` posts a `kipi/reviewer-approved` commit status where it already writes the verdict record.

| | |
|---|---|
| **sha** | the `HEAD_SHA` captured *before* the reviewer was dispatched (ASK-216). Never re-read — a fresh lookup would pin a commit the reviewer never saw, which is worse than no status because it looks authoritative. |
| **state** | `success` for `APPROVE` / `APPROVE WITH NITS`; `failure` for everything else, **including an unstated verdict**. Reuses `$VERDICT` (the derived-over-stated value), not a second grep of the review prose. |
| **when** | only under `--post`. A human running `kipi review 23` for a dry read moves no gate. |
| **description** | the verdict text, truncated to GitHub's 140-char limit. |
| **target_url** | the PR comment URL `--post` already created; **omitted rather than invented** when the comment failed. |

A commit **status**, not a PR **review**: this agent runs as the account that authors these PRs and GitHub forbids self-approval, so a review would deadlock. Proven live against PR #23 on 2026-07-27 — no bot login, no second identity, the existing `repo` scope is enough.

## Failure modes (this repo's defect class is "fails while reporting success")

- A failed POST prints a WARN naming **the sha and the context**, keeps the verdict record, and exits 0. A lost status must not lose the review.
- Empty `HEAD_SHA` ⇒ no status attempted at all, and it says so on stdout.
- The reviewer failing or timing out already exits above the verdict block, so nothing is posted. **Absent = not approved.** That is the safe direction, and it is what will hold the PR once 5b makes this context required.

## Verification

`test-severity-floor.sh` extended in place, section N — no new test file, so no `capability-manifest.json` edit and no conflict on the magnet file (`sp-f3a2ad81`). The **real** `pr-review-agent.sh` runs end to end with `gh` and `claude` stubbed and `HOME` redirected; **every assertion is on the `gh` call log, never on stdout prose**. No live GitHub call from the suite.

**RED first, observed:**

```
FAIL: THE DEFECT: the reviewer approved PR #901 and posted NOTHING to GitHub. The
      verdict stayed a local file, so no required check can ever read it and the PR waits on a
      human forever. gh was called with:
        pr view 901 --json headRefOid,title -q .headRefOid + "\t" + .title
        pr comment 901 --body-file .../pr-901-20260727-171014.md
```

**GREEN after (77 existing checks + 10 new):**

```
  ok: an approving review posts a commit status to GitHub
  ok: the status is posted on the exact sha the reviewer read (the stub's headRefOid)
  ok: APPROVE maps to state=success on context kipi/reviewer-approved
  ok: target_url is the PR comment URL --post actually created
  ok: REQUEST CHANGES maps to state=failure (the gate can refuse)
  ok: without --post no status call is made (a dry read moves nothing)
  ok: a failed status POST is loud on stderr, keeps the record, and exits 0
  ok: an empty head sha posts no status at all, and says so
  ok: no comment URL means no target_url (omitted, not invented)
  ok: the reviewer parses (bash -n)
PASS: 87/87 severity-floor checks
```

**Negative self-test** — a test that cannot fail is decoration. Mutating the state mapping on a throwaway copy so `REQUEST CHANGES` also maps to `success` takes the suite red:

```
FAIL: REQUEST CHANGES did not map to state=failure. A reviewer that only ever posts success
      is a gate that cannot refuse. Call was: api -X POST repos/{owner}/{repo}/statuses/a1b2... 
      -f state=success -f context=kipi/reviewer-approved -f description=REQUEST CHANGES ...
```

## Blast radius

Writes a commit status to a real PR on a public repo. **Statuses cannot be deleted**, only superseded by a later status on the same sha and context. Today the context is **not required**, so a wrong status is inert — that is deliberate, and it is why the producer ships before the gate.

## Not doing (binding, from the DoR)

- Adding `kipi/reviewer-approved` to `required_status_checks.contexts` — that is **5b**, and only after this producer is seen emitting on a live PR. A required context that nothing produces blocks every PR forever; that exact inversion is what PR #23 was.
- `gh pr merge --auto` / auto-merge opt-in (5c)
- Removing or changing the local verdict JSON — `converge.sh` and `linear-worker.sh` still read it; this is additive
- `rework_gate`, `pr-verdict-lib.sh`, `linear-worker.sh`, `converge.sh`, `pr-receipt-gate.py`
- Posting a GitHub PR *review* — impossible for this identity

### DIFF
--- a/q-system/.q-system/scripts/pr-review-agent.sh
+++ b/q-system/.q-system/scripts/pr-review-agent.sh
@@ -299,9 +299,60 @@ EOF
   echo "  minors captured as spillover: $CAPTURED of $MINOR_COUNT"
 fi
 
+# The verdict as a COMMIT STATUS on the sha the reviewer read (ASK-217).
+#
+# WHY THIS EXISTS: the verdict record above is a LOCAL file. GitHub cannot see
+# it, so no platform mechanism can gate on it, so every approved PR ends its
+# life waiting on a human. Every prior-art integrator (merge queue, Bors,
+# Mergify, Kodiak) has one shape: every precondition is a required status check
+# and the platform does the merging. pr-receipt-gate.py is already a CI step;
+# this was the one piece still stuck on disk.
+#
+# WHY A STATUS AND NOT A PR REVIEW: a commit status needs no second identity. A
+# PR *review* would deadlock -- this agent runs as the account that authors
+# these PRs, and GitHub forbids self-approval. Proven live on PR #23, 2026-07-27.
+#
+# ABSENT IS NOT APPROVED, and that is the point. A reviewer that fails or times
+# out exits well above this, before the verdict is even computed, so no status is
+# posted at all. Once this context becomes a REQUIRED check, "absent" is what
+# holds the PR -- the safe direction. Nothing on an error path here invents one.
+post_reviewer_status() {
+  local sha="$1" verdict="$2" target="$3"
+  local context="kipi/reviewer-approved" state="failure" desc
+  # ONE reader of the verdict. $VERDICT is the derived-over-stated value already
+  # written to the record; re-grepping the review prose here would be a second
+  # reader with its own semantics, which is the defect class this repo keeps
+  # finding. Anything that is not an approval -- including an unstated verdict --
+  # is a failure, so an unparseable review cannot pass a gate by accident.
+  case "$verdict" in
+    "APPROVE"|"APPROVE WITH NITS") state="success" ;;
+  esac
+  desc="$(printf '%.140s' "${verdict:-unstated: no verdict parsed from the review}")"
+  local args=(api -X POST "repos/{owner}/{repo}/statuses/$sha"
+              -f "state=$state" -f "context=$context" -f "description=$desc")
+  # Link only a real URL. The PR comment just above is what --post creates; when
+  # that failed there is nothing to link, and a local file path is not a URL.
+  case "$target" in https://*) args+=(-f "target_url=$target") ;; esac
+  if gh "${args[@]}" >/dev/null 2>&1; then
+    echo "  commit status posted: $context=$state on $sha"
+  else
+    echo "  WARN: could not post commit status '$context' (state=$state) on sha $sha; the review is recorded but NO gate moved" >&2
+  fi
+}
+
 if [ "$POST" = "1" ]; then
-  gh pr comment "$PR" --body-file "$REVIEW" >/dev/null 2>&1 \
-    && echo "  posted to PR #$PR" || echo "  WARN: could not comment on PR" >&2
+  COMMENT_URL=""
+  COMMENT_URL="$(gh pr comment "$PR" --body-file "$REVIEW" 2>/dev/null)" \
+    && echo "  posted to PR #$PR" \
+    || { COMMENT_URL=""; echo "  WARN: could not comment on PR" >&2; }
+  # No sha, no status. A status on a guessed commit is worse than none because
+  # it looks authoritative -- the same reason ASK-216 captured the sha before
+  # dispatch instead of looking it up afterwards.
+  if [ -n "$HEAD_SHA" ]; then
+    post_reviewer_status "$HEAD_SHA" "$VERDICT" "$COMMENT_URL"
+  else
+    echo "  no head sha for PR #$PR: posting NO commit status (a status on a guessed sha looks authoritative)"
+  fi
   if [ -n "$ISSUE" ]; then
     python3 "$SYNC" progress "$ISSUE" \
       "Adversarial review of PR #$PR complete. Verdict: ${VERDICT:-unstated}. Reviewer: Netflix-staff persona, fresh eyes, every finding required to ship an executed reproducer." \
--- a/q-system/.q-system/scripts/test/test-severity-floor.sh
+++ b/q-system/.q-system/scripts/test/test-severity-floor.sh
@@ -863,6 +863,187 @@ RUN_LINE="$(grep -n 'run_bounded "\$TIMEOUT_SECONDS"' "$REVIEWER" | head -1 | cu
       landing mid-review would make the record claim a commit the reviewer never read."
 ok "the head sha is captured before the review is taken, not looked up afterwards"
 
+# --- N. the verdict leaves the machine as a COMMIT STATUS (ASK-217) ----------
+# THE DEFECT: the verdict is a LOCAL file (~/.config/kipi/pr-reviews/...json).
+# GitHub cannot see it, so no required check can gate on it, so every approved
+# PR ends its life waiting on a human. Same harness as section M -- the REAL
+# pr-review-agent.sh runs end to end with `gh` and `claude` stubbed and HOME
+# redirected -- and every assertion below is on the gh CALL LOG, never on stdout
+# prose. "posted" printed while nothing left the machine is this repo's whole
+# defect class (something fails while reporting success), so the prose is not
+# admissible evidence here.
+#
+# A commit STATUS, not a PR review: this agent runs as the account that authors
+# these PRs and GitHub forbids self-approval, so a review would deadlock.
+STATUS_CONTEXT="kipi/reviewer-approved"
+COMMENT_URL_FIXTURE="https://github.com/o/r/pull/901#issuecomment-4242"
+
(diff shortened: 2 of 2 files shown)


assafkip:
Three repros run, all with real output. Here is the review.

---

## PR #29 — reviewer emits its verdict as a commit status (ASK-217)

### Findings

---

**minor** — When the status POST fails, the loop still reports a clean converge; the only signal is a stderr line, and the POST is never retried.

`q-system/.q-system/scripts/pr-review-agent.sh:339`

The reviewer exits 0 and the verdict record still says `APPROVE`. `linear-worker.sh:713` reacts only to a non-zero exit, so its `|| say "WARN: reviewer failed"` never fires; line 718 reads the record and line 721 announces converged. Stdout carries no trace. The POST is attempted exactly once — no retry — against the fleet's own 3-attempts-per-step contract in `.claude/rules/self-healing-retry.md`.

```
bash /Users/assafkipnis/projects/kipi-system/.pr29rev/repro-3-silent-nogate.sh
```

```
=== reviewer exit code ===
0
=== stdout the operator sees ===
  2026-07-28T00:24:08Z reviewing PR #901: reviewer status emission
    head sha under review: a1b2c3d4e5f60718293a4b5c6d7e8f9012345678
    round: 1
    verdict: APPROVE
    posted to PR #901
  2026-07-28T00:24:09Z done
=== stderr ===
    WARN: could not post commit status 'kipi/reviewer-approved' (state=success) on sha a1b2c... ; the review is recorded but NO gate moved
=== gh calls (note: the statuses/ POST was attempted ONCE, no retry) ===
  api -X POST repos/{owner}/{repo}/statuses/a1b2... -f state=success -f context=kipi/reviewer-approved ...
=== replay of linear-worker.sh:713-721 against that record ===
  worker line 714 does NOT fire (reviewer exited 0)
  worker line 721 says: "ASK-901 converged: APPROVE after N round(s); PR #901 waits on founder merge"
```

Minor and not major on purpose: today the context is not required, so "converged, waits on founder merge" is a **true** statement — nothing is lost. The moment 5b makes `kipi/reviewer-approved` required, that same line becomes a false success on a silently blocked PR. Fix it before 5b, not as a gate on this PR.

---

**minor** — Nothing in the suite pins the status to the pre-dispatch sha; the status path can be changed to re-read the head after the review and stay 87/87 green.

`q-system/.q-system/scripts/test/test-severity-floor.sh:951`

This is the PR's headline safety claim ("never re-read — a fresh lookup would pin a commit the reviewer never saw"). Section M has a line-order guard for the *record*; the status path has no equivalent, and the fixture answers every `pr view` with the same oid, so pinned and re-looked-up are indistinguishable to it.

```
bash /Users/assafkipnis/projects/kipi-system/.pr29rev/repro-2-sha-relookup.sh
```

```
=== mutated post-time sha lookup ===
352:    LATE_SHA="$(gh pr view "$PR" --json headRefOid -q .headRefOid 2>/dev/null)"
353:    post_reviewer_status "${LATE_SHA:-$HEAD_SHA}" "$VERDICT" "$COMMENT_URL"

=== re-running the FULL suite against the MUTATED reviewer ===
  ok: without --post no status call is made (a dry read moves nothing)
  ok: an empty head sha posts no status at all, and says so
  ok: the reviewer parses (bash -n)
PASS: 87/87 severity-floor checks
```

Shipped code is correct. The guard is what is missing. Cheapest fix is the section-M shape: assert `post_reviewer_status "$HEAD_SHA"` textually, or have the stub return a different oid on its second `pr view`.

---

**minor** — The "and says so" half of the empty-head-sha check cannot fail; it is satisfied by a pre-existing line printed on every run.

`q-system/.q-system/scripts/test/test-severity-floor.sh:1027`

`grep -qi 'head sha' "$N5/out.txt"` matches `head sha under review:` from `pr-review-agent.sh:92`, not the `else`-branch message at `:354` the check exists to prove.

```
bash /Users/assafkipnis/projects/kipi-system/.pr29rev/repro-1-nosha-decoration.sh
```

```
=== after mutation (should be empty) ===
(gone)
=== re-running the suite against the MUTATED reviewer ===
  ok: an empty head sha posts no status at all, and says so
PASS: 87/87 severity-floor checks
```

And the line that actually satisfies it, from repro 3:

```
=== FINDING 1 cross-check: which stdout line satisfies N5's grep -qi 'head sha'? ===
  2:  head sha under review: a1b2c3d4e5f60718293a4b5c6d7e8f9012345678
```

Anchor on `posting NO commit status` instead. The sibling half (`[ -z "$(status_call ...)" ]`) is real and does fail.

---

### What is sound — attacks that did not survive

- **State mapping escape.** I traced the closed vocabulary `extract_verdict` / `verdict_from_findings` can emit: `APPROVE`, `APPROVE WITH NITS`, `REQUEST CHANGES`, `BLOCK`, empty. Only the two approvals hit the `case`; everything else, including empty, falls to the `state="failure"` initializer. No value reaches a wrong `success`. Lowercase or qualified prose yields empty, which is also failure. Fail-closed holds.
- **Second reader of the verdict.** I looked for a re-grep of the review prose inside the status path. There is none; `$VERDICT` is passed as a parameter. The claim in the comment matches the code.
- **Invented `target_url`.** `case "$target" in https://*)` rejects both an empty string and a local file path. Removing it is caught by N6.
- **`--post` guard.** Moving the call outside the `if` is caught by N3 on the call log, not on prose.
- **Comment failure taking the status down.** Confirmed independent in repro 3: `pr comment` and the status POST are separate `gh` invocations; a failed comment still posts the status, minus the URL.
- **Verdict record survival on POST failure.** Confirmed in repro 3 — the record was written intact with `head_sha` populated, and the exit code was 0.
- **Assertions on prose instead of effects.** Every section-N check reads `gh-calls.log`, not stdout, except the two operator-message checks. That is the right split.
- **Test isolation reaching live state.** `HOME` is redirected (repro 3 wrote to `.pr29rev/r3/home/.config/kipi/...`), the stub PR title carries no `ASK-nnn` so `linear-sync.py` is unreachable, and the fixture derives `APPROVE` from an empty findings block so the spillover ledger path is never entered. No live write escaped.
- **Wiring.** The suite is registered at `q-system/.q-system/capability-manifest.json:193`, so section N actually runs rather than sitting inert.
- **bash 3.2.** `local args=(...)` plus `args+=(...)` parse and run on this machine's default bash; the full suite executed green.

I also tried to build a repro for cwd-dependent `{owner}/{repo}` resolution writing a permanent status to the wrong repo (`kipi` never `cd`s, and `$SKEL` is script-relative while `gh` resolves from the caller's directory). I could not produce an executed failure without a live cross-repo POST, and the same exposure already exists for `gh pr view` / `gh pr comment`. Dropping it.

---

Three minors, no blocker, no major. The producer-before-gate sequencing is right, absent-is-not-approved holds on every error path I could exercise, and the tests assert on effects rather than prose.

**Fix first:** make the failed-status outcome visible to the caller — record it in `pr-<N>.verdict.json` or surface it on stdout — before 5b makes the context required. That is the one that turns from harmless into a false "converged" the day the gate arms.

Scratch dir left at `/Users/assafkipnis/projects/kipi-system/.pr29rev/` (untracked, holds the three repro scripts). Nothing in the repo was modified, committed, or pushed.

FINDINGS:
minor|A failed status POST exits 0 with no retry, so the worker reports "converged" and the only signal is a stderr line|q-system/.q-system/scripts/pr-review-agent.sh:339
minor|No check pins the status to the pre-dispatch sha, so a post-review re-lookup passes the suite 87/87|q-system/.q-system/scripts/test/test-severity-floor.sh:951
minor|The "and says so" assertion is satisfied by the unrelated "head sha under review" line and cannot fail|q-system/.q-system/scripts/test/test-severity-floor.sh:1027
END FINDINGS

PR #30, opened 2026-07-28
fix(seam): both callers pass the reviewed sha and the head, so ASK-216's drift exit can fire (ASK-219)

assafkip:
Resolves spillover `sp-a27722e7`. Blocks 5c (auto-merge). Linear: ASK-219.

## The defect: a shipped reader with no caller

ASK-216 (`1b7e75c`) gave `rework_gate` optional 3rd/4th arguments and **exit 40**: an approving verdict at a sha that is no longer the head is a re-review, not terminal. Nothing passed those arguments.

```
converge.sh:161      rework_gate "$VERDICT"; GATE=$?
linear-worker.sh:394 rework_gate "$PR_VERDICT" "$MERGE_STATE"; GATE=$?
```

One argument and two. Exit 40 could never fire; the `head_sha` the reviewer writes faithfully was read by nobody.

**Observed live 2026-07-28:** `pr-27.verdict.json` approved `bf641ad`, `sana/ask-215` was pushed to `c063c3d`, and `kipi converge --issue ASK-215` exited in three seconds with `verdict 'APPROVE WITH NITS' ... Waiting on founder merge only.` Once 5c enables `gh pr merge --auto`, that merges code no reviewer ever read into a `main` that fans out fleet-wide through `kipi update`.

## The change

- **`pr-verdict-lib.sh`** — new `pr_head_sha`, the one reader of `headRefOid`, mirroring `pr_merge_state`. Both drivers now compare the head against a pinned sha; two private `gh pr view` copies is how those comparisons drift apart. `rework_gate`'s body is untouched (out of scope, and the reader is correct).
- **`converge.sh`** — passes `""` for the merge state (still the worker's job, ASK-212), plus the reviewed sha from the record and the head it already read one line up. No second read, no second source of truth. Gate 40 names both shas, then falls **through** to the no-progress guard so a stuck PR stops at exit 5 instead of re-reviewing to the cap.
- **`linear-worker.sh`** — sha arguments **appended**, so `$MERGE_STATE` keeps argument 2 and ASK-212's rebase rounds still fire. Gate 40 falls through to the dispatch, whose step-5 review writes a record pinned to the current head — the only thing that clears the drift. It spends **no conflict round**: a drift round is a review round, and the rebase prompt would tell the agent to force-push a diff nobody reviewed.

## Check: RED observed, then green

[...]

PR #31, opened 2026-07-28
fix(claim): a killed holder's claim is reclaimable on read, so a SIGKILL cannot wedge the board (ASK-189)

assafkip:
Closes the ASK-189 DoR. **Do not merge without review.**

## The defect

Measured twice on 2026-07-27: a run was killed, the claim stayed held by a session with **zero processes alive**, and every issue on the board was blocked until a human ran `release --holder`. SIGKILL cannot be trapped, so no in-process cleanup can ever close this. The lock has to tell, on READ, that its holder is gone.

## Approach: option 1 (holder pid), and why

The gap was never that pids are useless here. It is that **nothing long-lived was ever written down**. The claiming `python3` exits within milliseconds, which is exactly why the claim record deliberately carried no pid. The worker shell lives for the entire run, and now names itself: `--holder-pid "$$"`.

Smaller than option 2 (lease + TTL) and it reuses `_pid_alive`, which already exists for the critical-section guard. No refresh path to get wrong.

## Pid recycling: addressed, not accepted

The DoR calls this the trap in option 1. The record stores the holder's absolute start time (`ps -o lstart=`), so a process that later inherits the dead holder's pid reads as a *different* holder rather than as the original still running. Residual: `lstart` has 1-second resolution, so a recycled pid would have to land in the same wall-clock second on a pid space that wraps at ~99999.

## Zombies earn the second half of the check

`os.kill(pid, 0)` **succeeds** on a SIGKILLed process until its parent reaps it. A pid-only check would have read the measured scar as "holder alive" and changed nothing. `ps -o state=` catches it. Verified on this machine before the design was written.

## Blast radius: the direction of doubt is one-way

The DoR names the failure to avoid above all: a liveness check that reads "dead" for a HEALTHY long-running worker, because that silently disables the mutex while the suite stays green. So a reclaim happens **only on positive proof of death**. Missing field, unreadable `ps`, permission error, ps timeout, unparseable output: all read as ALIVE and the claim stands.

Leaking a lock wedges a board and a human notices. Reclaiming a live holder's tree hands two workers the same files, which is scar `53f2eeb`, and nobody notices.

**Five of the seven new checks assert the mutex is NOT weakened**, not that it reclaims.

## Reproducer first, observed RED

```
FAIL: a claim carrying --holder-pid was refused (exit 2): linear-claim.py: error: unrecognized arguments: --holder-pid 19937
EXIT=1
```

Then green, 30 -> 37 checks:

```
ok: a claim whose holder process is alive is still refused
ok: a claim whose SIGKILLed holder is an unreaped zombie is reclaimable
ok: a claim whose holder is fully gone is reclaimed without --break-stale, and says so
ok: a claim record with no liveness field is still respected
ok: a live pid whose start time does not match the record is a recycled pid, not the holder
ok: a live holder with no recorded start time is respected (unknown is not proof of death)
ok: a --holder-pid that is not a usable pid is a usage error, never a silent drop
PASS: linear-claim (37 checks)
```

## Regression sweep (real output)

```
=== test-linear-claim.sh rc=0 :: PASS: linear-claim (37 checks)
=== test-linear-worker-parallel.sh rc=0 :: PASS: linear-worker parallel (7 checks)
=== test-linear-worker-fetch.sh rc=0 :: PASS: linear-worker fetch (10 checks)
=== test-converge.sh rc=0 :: PASS: 17/17 converge checks
=== test-severity-floor.sh rc=0 :: PASS: 87/87 severity-floor checks
=== test-linear-wiring.sh rc=0 :: pass=13 fail=0
```

## Two residuals, stated rather than papered over

1. **Orphaned child.** If the worker shell alone is killed and its backgrounded `claude` child is orphaned, the holder reads dead while work continues. That needs a targeted kill of that one pid; a timeout, a killed process group, a slept laptop, or a reboot (every case actually observed) takes the whole tree down together. Documented at the call site.
2. **`$$` not `$BASHPID`.** The claim runs inside a subshell inside the `while` loop's pipeline subshell. `$$` stays the script's pid through both (verified); `$BASHPID` would be the innermost subshell, dead on the next line, and does not exist in the bash 3.2 macOS ships.

## Captured, not fixed here

`sp-763f21b5`: `converge.sh:111` compares `rec.get("issue")` against a record whose key is `issue_id`, so that TERM/INT/HUP trap has been a no-op since it was written (`4b5a7af`). Out of the DoR's file scope.

## Not doing (DoR-binding)

No change to the remote-collision half, no change to the worker's four refusals, no scheduling.

### DIFF
--- a/q-system/.q-system/scripts/linear-claim.py
+++ b/q-system/.q-system/scripts/linear-claim.py
@@ -96,6 +96,29 @@ GUARD_TIMEOUT_SECONDS = 5.0
 # which reads as alive.
 PS_TIMEOUT_SECONDS = 5.0
 
+# `ps -o lstart=` renders the start time in the READING process's timezone and
+# locale. The write and the read are different processes, hours apart, so an
+# unpinned render turned a LIVE holder into "the pid was recycled" the moment the
+# system timezone moved -- macOS sets the zone automatically and a laptop that
+# travels flips it. That is the mutex handing a live worker's tree to a second
+# worker, which is scar 53f2eeb and the one direction this feature must never go.
+# Measured 2026-07-27, one pid, same instant:
+#   TZ=UTC        -> 'Tue Jul 28 01:52:53 2026'
+#   TZ=Asia/Tokyo -> 'Tue Jul 28 10:52:53 2026'
+# Both the write and the read go through _proc_snapshot, so pinning it in one
+# place is what keeps the two sides of the comparison rendering identically.
+_PS_ENV = {**os.environ, "TZ": "UTC", "LC_ALL": "C"}
+
+# THE KEY CARRIES ITS FORMAT, because the format changed once. Records written
+# before the pin above hold a LOCAL-time render under the old name
+# `holder_pid_start`; comparing one of those against a UTC render says "recycled
+# pid" about a holder that is alive and working -- this fix causing the exact
+# defect it fixes, for the length of the upgrade window (a `kipi update` can
+# replace this script while a claim is held). Under a new name an old record
+# simply has no recorded start time, which already reads as unknown, which reads
+# as alive. Downgrade is safe for the same reason, in the other direction.
+HOLDER_START_KEY = "holder_pid_start_utc"
+
 
 class CorruptLock(Exception):
     """The lock exists but cannot be understood. Callers refuse, never grant."""
@@ -199,6 +222,9 @@ def _proc_snapshot(pid: int) -> tuple[str, str] | None:
     reaps it, and `os.kill(pid, 0)` succeeds on a zombie -- so the pid check
     alone reads a killed holder as alive for as long as nobody waits on it.
 
+    The environment is PINNED (`_PS_ENV`): `lstart` is rendered in the reading
+    process's timezone, and the writer and the reader are different processes.
+
     None on any doubt (ps missing, non-zero, unparseable, slow). Callers must
     treat None as "cannot prove death", never as death.
     """
@@ -209,6 +235,7 @@ def _proc_snapshot(pid: int) -> tuple[str, str] | None:
             text=True,
             check=False,
             timeout=PS_TIMEOUT_SECONDS,
+            env=_PS_ENV,
         )
     except (OSError, ValueError, TypeError, subprocess.SubprocessError):
         return None
@@ -237,7 +264,7 @@ def holder_death_reason(record: dict) -> str | None:
     started, state = snapshot
     if state.upper().startswith("Z"):
         return f"holder process {pid} was killed and is a zombie awaiting reap"
-    recorded_start = record.get("holder_pid_start")
+    recorded_start = record.get(HOLDER_START_KEY)
     if isinstance(recorded_start, str) and recorded_start and recorded_start != started:
         return (
             f"pid {pid} now belongs to a process started {started!r}, not the "
@@ -264,7 +291,7 @@ def holder_fields(pid: int | None) -> dict:
     fields: dict = {"holder_pid": pid}
     snapshot = _proc_snapshot(pid)
     if snapshot is not None:
-        fields["holder_pid_start"] = snapshot[0]
+        fields[HOLDER_START_KEY] = snapshot[0]
     return fields
 
 
@@ -648,7 +675,22 @@ def cmd_status(_args: argparse.Namespace) -> int:
     if held is None:
         print("no claim held in this working tree")
         return EXIT_OK
-    print(f"{held.get('issue_id')} claimed by {_describe(held)}")
+    line = f"{held.get('issue_id')} claimed by {_describe(held)}"
+    # `status` is what a human runs on a board that looks wedged. Reporting a
+    # provably-dead holder as a live claim is the REPORT contradicting the
+    # detector: `claim` reclaims that same record one line later. The fix landing
+    # on the detector alone would leave the operator reading the wrong answer
+    # from the one command they actually type.
+    #
+    # APPENDED, never substituted, and the exit code stays 0: the existing text
+    # has a reader (test-linear-claim.sh case 20 pulls the session token out of
+    # this line with sed), and a rewritten line or a new non-zero exit would
+    # break callers while every assertion about staleness still passed. This
+    # tells the truth; it does not become a new gate.
+    gone = holder_death_reason(held)
+    if gone is not None:
+        line += f" [STALE: {gone} -- the next claim will reclaim this tree]"
+    print(line)
     return EXIT_OK
 
 
--- a/q-system/.q-system/scripts/linear-claim.py
+++ b/q-system/.q-system/scripts/linear-claim.py
@@ -35,6 +35,26 @@ A pid cannot serve as that token: this is a CLI that exits immediately after
 writing the lock, so the recorded pid is a dead process within milliseconds.
 The token comes from the session's own environment.
 
+THE HOLDER'S LIVENESS IS A SEPARATE, OPTIONAL FIELD (ASK-189)
+-------------------------------------------------------------
+Because the session token says WHO holds the tree but nothing says whether that
+holder still exists, a killed session leaked its claim forever. Measured twice
+2026-07-27: a run was SIGKILLed, the claim stayed held by a session with zero
+processes alive, and a human had to run `release --holder`. SIGKILL cannot be
+trapped, so no in-process cleanup can ever close this -- the lock must be able
+to tell ON READ that its holder is gone.
+
+The caller therefore names a process that lives for the WHOLE run via
+`--holder-pid` (linear-worker.sh passes its own `$$`). That is the piece that was
+missing: not that pids are useless here, but that nothing long-lived was ever
+written down.
+
+The direction of doubt is one-way and load-bearing: a reclaim happens only on
+POSITIVE PROOF of death. A missing field, an unreadable `ps`, a permission error
+-- anything unknown -- reads as ALIVE and the claim stands. Leaking a lock wedges
+a board and a human notices; reclaiming a LIVE holder's tree hands two workers
+the same files, which is scar 53f2eeb and nobody notices.
+
 FAIL CLOSED, EVERYWHERE
 -----------------------
 A mutex that grants under doubt is worse than none, because callers trust it.
@@ -71,6 +91,10 @@ STARTED_STATUS_TYPES = ("started",)
 STARTED_STATUS_NAMES = ("in progress", "in review")
 
 GUARD_TIMEOUT_SECONDS = 5.0
+# `ps` is a local read that normally returns in milliseconds. The bound exists so
+# a wedged process table can never hang the mutex; a timeout reads as "unknown",
+# which reads as alive.
+PS_TIMEOUT_SECONDS = 5.0
 
 
 class CorruptLock(Exception):
@@ -164,6 +188,84 @@ def _pid_alive(pid: int) -> bool:
     except (OverflowError, ValueError, TypeError):
         return False
     return True
+
+
+def _proc_snapshot(pid: int) -> tuple[str, str] | None:
+    """(start_time, state) for `pid` as `ps` sees it, or None when it cannot say.
+
+    `lstart` is the process's ABSOLUTE start time, which is what distinguishes
+    the recorded holder from an unrelated process that later inherited its pid.
+    `state` is here because a SIGKILLed process stays a ZOMBIE until its parent
+    reaps it, and `os.kill(pid, 0)` succeeds on a zombie -- so the pid check
+    alone reads a killed holder as alive for as long as nobody waits on it.
+
+    None on any doubt (ps missing, non-zero, unparseable, slow). Callers must
+    treat None as "cannot prove death", never as death.
+    """
+    try:
+        result = subprocess.run(
+            ["ps", "-p", str(int(pid)), "-o", "lstart=,state="],
+            capture_output=True,
+            text=True,
+            check=False,
+            timeout=PS_TIMEOUT_SECONDS,
+        )
+    except (OSError, ValueError, TypeError, subprocess.SubprocessError):
+        return None
+    if result.returncode != 0:
+        return None
+    parts = result.stdout.strip().rsplit(None, 1)
+    if len(parts) != 2 or not parts[0].strip():
+        return None
+    return parts[0].strip(), parts[1].strip()
+
+
+def holder_death_reason(record: dict) -> str | None:
+    """Why this claim's holder is PROVABLY gone, or None when it must be respected.
+
+    The one-way doubt rule from the module docstring is implemented here and
+    nowhere else: every path that cannot prove death returns None.
+    """
+    pid = record.get("holder_pid")
+    if not isinstance(pid, int) or isinstance(pid, bool) or pid <= 0:
+        return None  # no long-lived holder was ever recorded; respect the claim
+    if not _pid_alive(pid):
+        return f"holder process {pid} no longer exists"
+    snapshot = _proc_snapshot(pid)
+    if snapshot is None:
+        return None  # ps could not answer, and silence is not proof
+    started, state = snapshot
+    if state.upper().startswith("Z"):
+        return f"holder process {pid} was killed and is a zombie awaiting reap"
+    recorded_start = record.get("holder_pid_start")
+    if isinstance(recorded_start, str) and recorded_start and recorded_start != started:
+        return (
+            f"pid {pid} now belongs to a process started {started!r}, not the "
+            f"holder recorded at {recorded_start!r} (the pid was recycled)"
+        )
+    return None
+
+
+def holder_fields(pid: int | None) -> dict:
+    """The liveness fields to store for a claim, empty when no holder was named.
+
+    Empty is the correct answer for a caller that has nothing long-lived to
+    point at, and it is why THIS process never records its own pid: the CLI
+    exits milliseconds after writing the lock, so a self-recorded pid would make
+    every claim instantly reclaimable -- the mutex disabled outright.
+
+    A pid whose start time cannot be read is still recorded, with no start key.
+    That is strictly better than recording nothing: `_pid_alive` alone still
+    catches the plain "holder is gone" case, and the absent key only costs the
+    recycling refinement.
+    """
+    if pid is None:
+        return {}
+    fields: dict = {"holder_pid": pid}
+    snapshot = _proc_snapshot(pid)
+    if snapshot is not None:
+        fields["holder_pid_start"] = snapshot[0]
+    return fields
 
 
 # ---------------------------------------------------------------------------
@@ -429,10 +531,28 @@ def cmd_claim(args: argparse.Namespace) -> int:
                 held.update(
                     issue_id=args.issue_id, refreshed_at=_now_iso(), agent=args.agent
                 )
+                # Refresh liveness too, so a long run's record keeps describing a
+                # process that is really there. A re-claim that names no holder
+                # leaves the recorded one alone rather than dropping to "no
+                # liveness" -- same direction of doubt as everywhere else.
+                held.update(holder_fields(args.holder_pid))
                 write_claim(held)
                 print(f"already held by this session: {args.issue_id}")
                 return EXIT_OK
 
+            gone = holder_death_reason(held)
+            if gone is not None:
+                # Said out loud, on stderr, every time. A mutex that silently
+                # hands a resource from one holder to another is indistinguishable
+                # from a mutex that never worked, and the operator debugging a
+                # two-workers-one-tree collision has to be able to see this line.
+                sys.stderr.write(
+                    f"RECLAIMED: the claim held by {_describe(held)} is stale -- "
+                    f"{gone}. Taking the tree; no --break-stale needed.\n"
+                )
+                held = None
+
+        if held is not None:
             if args.break_stale:
                 # Compare-and-swap, not a blind steal. Scar, adversarial review
                 # 2026-07-26: --break-stale took a demonstrably LIVE claim on
@@ -486,6 +606,7 @@ def cmd_claim(args: argparse.Namespace) -> int:
                 "agent": args.agent,
                 "session": session,
                 "acquired_at": _now_iso(),
+                **holder_fields(args.holder_pid),
             }
         )
         print(f"claimed {args.issue_id} for {args.agent} (session {session})")
@@ -545,6 +666,13 @@ def main() -> int:
     p.add_argument("--remote-state", help="verbatim mcp__linear__get_issue response")
     p.add_argument("--break-stale", action="store_true", help="requires --holder")
     p.add_argument("--holder", help="the session token you are deliberately breaking")
+    # Deliberately NOT type=int: argparse exits 2 on a bad value, and 2 is not
+    # this tool's vocabulary -- same reason --agent is validated by hand below.
+    p.add_argument(
+        "--holder-pid",
+        help="pid of a process that lives for the WHOLE run (the worker shell's "
+        "$$), so a killed holder's claim can be reclaimed on read",
+    )
     p.set_defaults(func=cmd_claim)
 
     p = sub.add_parser("release", help="drop this working tree's claim")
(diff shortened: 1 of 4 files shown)


assafkip:
Review of PR #31, read-only, from a checkout of head `7de67d7`.

---

## Findings

### 1. MAJOR — `holder_pid_start` is a locale/timezone-rendered string, so a live holder reads as a recycled pid and its tree gets taken

`q-system/.q-system/scripts/linear-claim.py:207` (and the equality compare at `:241`)

`ps -o lstart=` renders in the **reading process's** local time. The write and the read are different processes, hours apart. If the timezone changes between them, the string differs and `holder_death_reason()` returns "the pid was recycled" for a process that is alive and working. That is the one direction the PR body says must never happen.

```
$ bash .pr31rev/repro-tz.sh
holder pid=45039
  TZ=UTC              lstart -> [Tue Jul 28 01:46:54 2026    ]
  TZ=America/New_York lstart -> [Mon Jul 27 21:46:54 2026    ]
recorded: {"issue_id": "ASK-TZ", ..., "holder_pid": 45039, "holder_pid_start": "Tue Jul 28 01:46:54 2026"}
second session exit=0  (3=refused/correct, 0=RECLAIMED a LIVE holder)
stderr: RECLAIMED: ... pid 45039 now belongs to a process started 'Mon Jul 27 21:46:54 2026',
        not the holder recorded at 'Tue Jul 28 01:46:54 2026' (the pid was recycled).
RESULT: LIVE HOLDER RECLAIMED
```

Honest scoping: my repro sets `TZ` by hand, and I grepped `q-system/.q-system/scripts` for `TZ=` and found **zero** producers that set it. The reachable production trigger is a *system* timezone change between claim and read (macOS "Set time zone automatically" is on by default; a laptop that travels flips it), which I did not observe firing. I am holding this at major rather than dropping it because the consequence is exactly scar `53f2eeb` — two workers on one tree, unattended, and per finding 2 with no trace anywhere.

Fix is one line: pin the environment on both `ps` calls so write and read render identically, e.g. `env={**os.environ, "TZ": "UTC", "LC_ALL": "C"}` in `_proc_snapshot`. Both sides go through that one function, so it stays symmetric.

---

### 2. MINOR — the `RECLAIMED:` audit line goes to `/dev/null` at the only production caller

`q-system/.q-system/scripts/linear-worker.sh:516`

The code comment at `linear-claim.py:545-548` says the operator "has to be able to see this line." The single production call site discards it.

```
$ bash .pr31rev/repro-silent.sh
--- grep of the production call site ---
515:  ( cd "$TREE" && python3 "$CLAIM" claim "$ISSUE" --agent "$AGENT" --session "$SESSION" \
516-      --holder-pid "$$" ) >/dev/null 2>&1
--- worker-shaped invocation (the exact redirect from linear-worker.sh) ---
worker sees: rc=0
worker LOG contents: []
operator-visible mention of the steal: 0 lines in the run log

--- same call WITHOUT the redirect, for contrast ---
RECLAIMED: the claim held by sana (session worker-1) on ASK-DEAD since 2026-07-28T01:47:38Z
is stale -- holder process 999999 no longer exists. Taking the tree; no --break-stale needed.
```

At 3am a tree changes hands and `$LOG` shows a normal `start ASK-xxx` line and nothing else. Minor by the anchors — nothing pages, no human action — but it is the evidence trail finding 1 would need.

---

### 3. MINOR — the zombie case is decoration: the suite passes with zombie detection deleted

`q-system/.q-system/scripts/test/test-linear-claim.sh:332`

The PR body sells this as the load-bearing half ("A pid-only check would have read the measured scar as 'holder alive' and changed nothing"). The fixture never produces a zombie — bash reaps the background child in its SIGCHLD handler before the claim runs, so the assertion travels the `_pid_alive == False` path and duplicates case 22.

Mutation, `if state.upper().startswith("Z"):` → `if False:`:

```
$ bash .pr31rev/mut/.../test-linear-claim.sh | tail
  ok: a claim whose SIGKILLed holder is an unreaped zombie is reclaimable
  ...
PASS: linear-claim (37 checks)
```

Mechanism proof:

```
$ bash .pr31rev/repro-zombie.sh
spawned holder pid=44307  ps-state=[RN  ]
after kill -9, NO wait called:
  ps -p 44307 -o lstart=,state= -> [] rc=1
  os.kill(44307, 0) -> ProcessLookupError (pid GONE: already reaped)
```

To make it real the holder needs a parent that is alive and not reaping — e.g. spawn a grandchild via an intermediate `sh` that ignores SIGCHLD and stays blocked, then kill the grandchild.

---

### 4. MINOR — `status` reports a provably-dead holder as a live claim

`q-system/.q-system/scripts/linear-claim.py:646`

`cmd_status` never calls `holder_death_reason()`. It is the command a human runs on a wedged board, and it says the opposite of what `claim` says one line later.

```
$ bash .pr31rev/repro-status.sh
holder pid 48901 is now gone; claim record still on disk.
status says: ASK-DEAD claimed by sana (session worker-1) on ASK-DEAD since 2026-07-28T01:48:47Z
claim  says: RECLAIMED: ... holder process 48901 no longer exists. Taking the tree; no --break-stale needed.
```

---

## What is sound

- **Reclaim + concurrency.** The reclaim sits inside `_CriticalSection()`, so two claimants cannot both decide the holder is dead. The 12-way race check still reports exactly 1 winner, 11 refused, 0 crashed.
- **Reclaim does not open a hole in the remote half.** After `held = None` control falls through to the `--remote-state` gate at `:581`, not around it. I traced every path out of the new block.
- **Stale-start pairing on re-claim.** I chased the case where `held.update(holder_fields(...))` writes a new `holder_pid` while keeping the *old* `holder_pid_start` (ps fails on the new pid) — which would read as a recycled pid and reclaim a live tree. It is unreachable: `linear-worker.sh:114` derives `SESSION="worker-$(date +%s)-$$"`, unique per run, so `same_session` is never true across two different holder pids. Dropped.
- **Self-pid trap avoided.** `holder_fields` is only ever fed a caller-supplied pid; the CLI never records its own. A self-recorded pid would make every claim instantly reclaimable, and it does not happen.
- **Type guards on the record.** `isinstance(pid, bool)` is checked explicitly, so a JSON `true` is not read as pid 1. Junk `--holder-pid` (`not-a-pid`, `0`) exits 1, not a silent drop — verified in the suite run.
- **Pre-liveness records.** A record with no `holder_pid` still refuses, so nothing already on disk becomes reclaimable.

---

## VERDICT: REQUEST CHANGES

Not a BLOCK — nothing here is unrecoverable, and the fail-closed direction is genuinely well built everywhere I pushed on it.

**Fix first:** pin `TZ` (and `LC_ALL`) on the `ps` call in `_proc_snapshot` at `linear-claim.py:207`. One line, both sides of the comparison go through it, and it closes the only path I found that hands a live worker's tree to a second worker.

FINDINGS:
major|`holder_pid_start` compares a timezone-rendered `ps -o lstart=` string, so a TZ change between claim and read reclaims a LIVE holder's tree|q-system/.q-system/scripts/linear-claim.py:207
minor|The `RECLAIMED:` audit line is discarded by `>/dev/null 2>&1` at the only production caller, so a tree changing hands leaves no trace in the run log|q-system/.q-system/scripts/linear-worker.sh:516
minor|The "unreaped zombie" check passes with zombie detection deleted; bash reaps the fixture before the claim runs, so it duplicates the pid-gone case|q-system/.q-system/scripts/test/test-linear-claim.sh:332
minor|`cmd_status` never consults `holder_death_reason`, so it reports a provably-dead holder as a live claim|q-system/.q-system/scripts/linear-claim.py:646
END FINDINGS

Scratch files are in the untracked `.pr31rev/` (same convention as `.pr27rev/`, `.pr30rev/`); nothing in the repo was modified.

assafkip:
## Round 3 rework, pushed as `27c3880`

All four findings fixed. Each one has a test that fails without its fix, and three of them are mutation-proved below. Nothing you listed under "what is sound" was touched.

---

### 1. MAJOR, `holder_pid_start` is a locale/timezone render — FIXED

Confirmed on this machine, one pid, same instant:

```
$ bash .pr31rev/probe-tz.sh
pid=58864
  TZ inherited : [Mon Jul 27 18:52:53 2026    ]
  TZ=UTC       : [Tue Jul 28 01:52:53 2026    ]
  TZ=NewYork   : [Mon Jul 27 21:52:53 2026    ]
  TZ=Tokyo     : [Tue Jul 28 10:52:53 2026    ]
  TZ=UTC again : [Tue Jul 28 01:52:53 2026    ]
```

`_proc_snapshot` now runs `ps` under a pinned `env` (`TZ=UTC`, `LC_ALL=C`). Both the write and the read go through that one function, so the two sides of the comparison render identically.

Reproducer added as case 26, observed RED before the fix:

```
FAIL: a LIVE holder was reclaimed because the timezone changed between claim and read (exit 0):
RECLAIMED: the claim held by agent-a (session sess-A) on ASK-TZ since 2026-07-28T01:58:13Z is stale
-- pid 9755 now belongs to a process started 'Tue Jul 28 10:58:12 2026', not the holder recorded at
'Tue Jul 28 01:58:12 2026' (the pid was recycled). Taking the tree; no --break-stale needed.
```

**The layer above that fix, which the one-line version would have broken.** A claim record written by the pre-pin build holds a LOCAL-time render. A UTC-pinned reader compares those two strings, sees a difference, and reclaims a holder that is alive and working — the fix causing the exact defect it fixes, for the length of the upgrade window. That window is real: `kipi update` can replace this script while a claim is held.

So the render format now lives in the KEY NAME. `HOLDER_START_KEY = "holder_pid_start_utc"`. An old record simply has no start time under the new key, which is `unknown`, which already reads as alive. Downgrade is safe the same way: an old build reading a new record finds no `holder_pid_start` and falls back to the pid check alone.

Case 27 pins it (writes the record with `TZ=Asia/Tokyo`, exactly what the pre-pin build wrote, against a live holder):

```
ok: a claim record in the pre-pin start-time format is respected, not read as recycled
```

Mutation, `env=_PS_ENV` deleted:

```
FAIL: a LIVE holder was reclaimed because the timezone changed between claim and read (exit 0)
```

---

### 2. MINOR, the `RECLAIMED:` line goes to `/dev/null` — FIXED

`linear-worker.sh` keeps stderr and echoes the `RECLAIMED:` line through `say`, so it lands in `$LOG`.

The test drives the **real worker** rather than grepping its source, because the defect was a shell redirect and a grep cannot see one. It creates a real worktree, seeds its lock with a holder pid that was spawned, killed and reaped, then runs `linear-worker.sh` and asserts on the run log.

Observed RED before the fix:

```
FAIL: the worker took a tree from a dead holder and said NOTHING in its log. It logged:
2026-07-28T01:58:31Z start ASK-EEE on sana/ask-eee in .../state-e/worktrees/ask-eee (attempt 1/3)
```

Green after:

```
ok: a claim left by a killed holder is reclaimed by the real worker, no human needed
ok: the reclaim is reported in the worker's run log, not swallowed by the redirect
```

**What I deliberately did NOT make louder, since a fix that buys silence is the expensive kind.** Only the `RECLAIMED:` line is echoed, not all of the claim's stderr. An ordinary refusal already produces `skip $ISSUE: working tree is claimed by another session` two lines below; echoing its `REFUSED:` text too would turn every normal collision into two log lines saying the same thing. And no Slack ping: a reclaim is this feature's SUCCESS path (the board self-healed instead of wedging), so paging on it at 3am would be noise. The run log is the right altitude for evidence nobody needs to act on immediately.

---

### 3. MINOR, the zombie case is decoration — FIXED

You were right, and the mechanism was exactly what you said. Confirmed the fixture that does work:

```
$ bash .pr31rev/probe-zombie.sh
parent=58949 grandchild=58951
before kill: state=[SN  ] ppid=[58949]
after kill -9 (nobody waits): state=[Z]
full ps line: [Mon Jul 27 18:52:53 2026     Z   ]
os.kill(pid,0) -> SUCCEEDS (a pid-only check reads this as ALIVE)
```

The holder is now a grandchild: `sh` forks it, then `exec`s into `sleep`, which has no SIGCHLD handling at all, so nothing ever reaps it.

The case now **asserts its preconditions** instead of assuming them, which is what stops it decaying back into a duplicate of case 22:

```
ok: the fixture really makes a zombie: ps says Z while os.kill(pid,0) still succeeds
ok: a claim whose SIGKILLed holder is an unreaped zombie is reclaimable
```

Your mutation, re-run against the new fixture:

```
if state.upper().startswith("Z"):  ->  if False:

  ok: a claim whose holder process is alive is still refused
  ok: the fixture really makes a zombie: ps says Z while os.kill(pid,0) still succeeds
FAIL: a SIGKILLed holder that is still an unreaped zombie blocked the tree (exit 3):
REFUSED: this working tree is claimed by agent-a (session sess-DEAD) on ASK-ZOMB ...
```

Case 21 lost its zombie half and is now just "a live holder is respected", which is what it actually tested.

---

### 4. MINOR, `status` reports a dead holder as live — FIXED

`cmd_status` consults `holder_death_reason` and appends a note:

```
ASK-STAT claimed by agent-a (session sess-STAT) on ASK-STAT since ...
  [STALE: holder process 48901 no longer exists -- the next claim will reclaim this tree]
```

**Appended, not substituted, and the exit code stays 0** — deliberately. Case 20 reads the session token out of this exact line with `sed -n 's/.*session \([^)]*\)).*/\1/p'`, and `converge.sh` reads status output too. A rewritten line or a new non-zero exit would break those readers while every new assertion about staleness still passed. Case 28 pins the parse alongside the behaviour:

```
ok: status does not call a live holder's claim stale
ok: status says so when the holder it reports is provably gone
ok: the status line still parses for its existing reader
```

Mutation, the STALE line deleted:

```
FAIL: status reported a holder with zero processes alive as a live claim:
ASK-STAT claimed by agent-a (session sess-STAT) on ASK-STAT since 2026-07-28T01:59:52Z
```

---

### Regression sweep, real output

```
=== test-linear-claim.sh rc=0 :: PASS: linear-claim (43 checks)
=== test-linear-worker-parallel.sh rc=0 :: PASS: linear-worker parallel (9 checks)
=== test-linear-worker-fetch.sh rc=0 :: PASS: linear-worker fetch (10 checks)
=== test-converge.sh rc=0 :: PASS: 17/17 converge checks
=== test-severity-floor.sh rc=0 :: PASS: 87/87 severity-floor checks
=== test-linear-wiring.sh rc=0 ::   pass=13 fail=0
```

37 -> 43 checks on the claim suite, 7 -> 9 on the worker-parallel suite.

### Not regressed

The reclaim still sits inside `_CriticalSection()` (the 12-way race still reports 1 winner / 11 refused / 0 crashed), still falls through to the `--remote-state` gate rather than around it, still never records the CLI's own pid, still type-guards the record, and every unknown still reads as alive.

Scratch files are in the untracked `.pr31rev/` (probes + the mutation harness); nothing outside the four files in the diff was modified.

assafkip:
## Independent Codex review (run by hand, 2026-07-28)

Second opinion from a **different lab and model** — Claude reviewed this PR too, and this is the decorrelated read. Run manually because ASK-221 (the automation) has not landed yet. Head sha reviewed: `7de67d76`.

**VERDICT: REQUEST CHANGES**

```
major|The mutex can reclaim a live holder when two invocations use different timezone or locale settings.|q-system/.q-system/scripts/linear-claim.py:241
minor|The production worker suppresses the reclaim audit message that the implementation claims operators can see.|q-system/.q-system/scripts/linear-worker.sh:515
minor|The zombie test does not create an unreaped zombie and therefore passes if zombie detection is removed.|q-system/.q-system/scripts/test/test-linear-claim.sh:317
minor|status reports a provably dead holder as an active claim.|q-system/.q-system/scripts/linear-claim.py:647
END FINDINGS
FINDINGS:
major|The mutex can reclaim a live holder when two invocations use different timezone or locale settings.|q-system/.q-system/scripts/linear-claim.py:241
minor|The production worker suppresses the reclaim audit message that the implementation claims operators can see.|q-system/.q-system/scripts/linear-worker.sh:515
```

The third one is the same class Claude caught on PR #30 earlier tonight: a test that passes whether or not the behaviour it names exists. Worth fixing on its own merits.

_Posted by hand. Not a required status — `kipi/codex-approved` becomes a gate only after ASK-221 ships and is seen emitting on a live PR._

assafkip:
## Review of PR #31 — round 2 (head `27c3880`), read-only from an isolated checkout

I read the code before the comments. Baseline first: both suites are green in a tree I built from the PR head, not from the author's claim.

```
$ bash q-system/.q-system/scripts/test/test-linear-claim.sh | tail -1
PASS: linear-claim (43 checks)
$ bash q-system/.q-system/scripts/test/test-linear-worker-parallel.sh | tail -1
PASS: linear-worker parallel (9 checks)
```

---

## Findings

### 1. minor — `holder_death_reason` reads "cannot evaluate this pid" as POSITIVE PROOF OF DEATH, which is the one thing the module docstring says it never does

`q-system/.q-system/scripts/linear-claim.py:259` (via `_pid_alive`'s `except (OverflowError, ValueError, TypeError): return False` at `:211`)

The docstring at `:52-56` states the invariant this whole feature rests on: "A missing field, an unreadable `ps`, a permission error -- anything unknown -- reads as ALIVE and the claim stands." `_pid_alive` breaks it in one branch: a pid `os.kill` cannot represent raises `OverflowError`, which is mapped to `False`, and `holder_death_reason` reads `False` as `"holder process N no longer exists"` and takes the tree.

```
$ bash .pr31rev/r2/repro2.sh
=== N8: an out-of-range holder_pid is read as PROOF OF DEATH, not as unknown ===
{"issue_id": "ASK-BIG", "agent": "sana", "session": "sess-LIVE", "acquired_at": "2026-07-28T00:00:00Z", "holder_pid": 9223372036854775808}
--- direct probe of the primitive:
_pid_alive(2**63)      = False
holder_death_reason ->  holder process 9223372036854775808 no longer exists
--- and the CLI takes the tree (0 = reclaimed, 3 = respected):
RECLAIMED: the claim held by sana (session sess-LIVE) on ASK-BIG since 2026-07-28T00:00:00Z is stale -- holder process 9223372036854775808 no longer exists. Taking the tree; no --break-stale needed.
claimed ASK-NEW for b (session S2)
EXIT=0
```

**Severity dropped a level, and here is why.** No producer in this repo emits that shape: the only writer is `--holder-pid "$$"`, and `$$` is always a small int. `_pid_alive` predates this PR — what is new is the consumer that turns its `False` into a decision to take a working tree. Before this change that branch only swept temp files. So: real, reproducible, bounded, nothing a human does about it at 3am. Minor.

The fix is one line in `holder_death_reason`, not in `_pid_alive` (the sweeper wants the current semantics): range-check `pid` against `2**31` alongside the existing `pid <= 0` guard at `:257`, so an unrepresentable pid joins "unknown" instead of "dead."

---

### 2. minor — this PR makes `converge.sh`'s stated rationale false, and that comment is what an operator reads while debugging a wedged board

`q-system/.q-system/scripts/converge.sh:75`

```
$ grep -n "does NOT pid-check" q-system/.q-system/scripts/converge.sh
75:# linear-claim.py deliberately does NOT pid-check the claim itself (only the
```

Full text at `:75-79`: *"linear-claim.py deliberately does NOT pid-check the claim itself (only the critical-section guard) ... Correct design, real operational hole: a SIGKILL, a harness timeout, a laptop sleeping, or a ctrl-c leaves the lock held with nothing to reclaim it."*

After this PR the claim **is** pid-checked and there **is** something to reclaim it. My repro above is the executed proof — `claim` reclaimed a stale record with no `--break-stale`. The comment is now the opposite of the truth, and it is the justification block for `release_stale_claim_for_issue`, i.e. the code path an operator reads when a board looks stuck.

converge.sh is outside the DoR's file scope, and the sibling defect in that same function (`rec.get("issue")` vs `issue_id`) is already captured as `sp-763f21b5`. This is a different, newly-created falsity and it is not captured. Follow-up, not a gate.

---

### 3. nit — case 27's `holder_pid_start` fixture is a record shape no shipped producer ever wrote

`q-system/.q-system/scripts/test/test-linear-claim.sh:507`

```
$ git log --oneline origin/main -S'holder_pid_start' -- q-system/.q-system/scripts/linear-claim.py
(no output = no shipped producer ever wrote that key)
```

The "pre-pin upgrade window" the `HOLDER_START_KEY` rename defends against never existed on `main` — `holder_pid_start` lived only in `7de67d7`, the intermediate commit on this same branch. The behavior case 27 asserts (a record with no start key under the new name is respected) is already asserted by case 23. Cheap, harmless defense-in-depth; I am naming it only because this repo's own fixture rule is that a fixture with no real producer proves nothing.

Same file, `:450` and `:467` are both labelled `# --- 26.`

---

## What is sound

Attacks I ran that the code survived. Two of these are findings I built, reproduced against, and then **dropped** — reporting that is the point.

- **Stale start-key pairing on re-claim — chased, could not reproduce, DROPPED.** `cmd_claim:565` does `held.update(holder_fields(...))`, which can write a new `holder_pid` while keeping the *old* start string when `ps` fails on the new pid — a live holder then reads as recycled. I forced it with a `ps` stub that exits 1 (`.pr31rev/r2/repro.sh`): the second session was still **REFUSED, exit 3**. The prior round dropped this citing `linear-worker.sh:114`; I tried to falsify that citation and could not — `SESSION="worker-$(date +%s)-$$"` embeds `$$`, so `same_session` and a changed `holder_pid` are mutually exclusive by construction. The citation holds.
- **The two new round-3 tests are load-bearing, verified by my own mutations, not the author's.** Reverting the worker redirect to `>/dev/null 2>&1` → `FAIL: the worker took a tree from a dead holder and said NOTHING in its log.` Deleting the `cmd_status` STALE append → `FAIL: status reported a holder with zero processes alive as a live claim`. Neither is decoration.
- **The "appended, never substituted" claim about `status` checks out against real callers.** The only production consumer is `kipi:179` (`claims) python3 "$CLAIM" status "$@"`), a human-facing passthrough. Nothing parses that line in production; the sed reader is the suite's own case 20, and it still parses.
- **The fix reaches the caller that produced the scar.** I traced the whole unattended path — launchd `com.kipi.dispatch.plist` → `kipi-dispatch.sh:98` (`nohup ./kipi converge`) → `converge.sh` → `linear-worker.sh`. `linear-worker.sh` is the only process that ever writes a claim (converge only *releases*), and it is the one that now names `$$`. This is not a fix wired to the wrong half.
- **The worker shell genuinely outlives the run.** `run_bounded` at `:121-136` does `wait "$job"`, so the recorded `$$` is alive for the whole `claude -p` dispatch and through the release at `:770`. The pid means what the record says it means.
- **Doubt is one-way everywhere else I pushed.** JSON `true` as a pid (explicit `isinstance(pid, bool)` guard), a float pid, a string pid, `PermissionError` on another user's process, `ps` unreadable, `ps` timeout — every one reads as alive. Only the `OverflowError` branch in finding 1 leaks.
- **Reclaim is inside the critical section and does not bypass the remote half.** After `held = None` at `:580`, control falls *through* to the `--remote-state` gate at `:608`, not around it.

---

## Calibration note on the review process itself

Worth stating because the standing rule asks for it. Round 1 raised 1 major + 3 minors; all four were fixed, each with a reproducer observed red, and I independently mutation-verified two of those fixes rather than taking the commit message's word. This round found zero majors on the new code. That is a **converging** PR, not the "keeps producing new blockers on unchanged code" pattern — the earlier rounds were calibrated correctly, and the right move is to ship it rather than mine for a fourth round.

---

## VERDICT: APPROVE WITH NITS

Nothing here is unrecoverable, nothing cries wolf, and the direction-of-doubt discipline is the strongest part of the change — I attacked it from six angles and it held on all but one branch that no producer can reach.

**Fix first:** the `OverflowError` leak (finding 1) — one range check next to the existing `pid <= 0` guard at `linear-claim.py:257`, so the code's behavior matches the invariant its own docstring advertises.

**One operational heads-up, not a finding:** auto-merge is already armed on this PR (`squash`, `mergeStateStatus: BLOCKED`), which per commit `8a359a3` is the intended flow. An approval lands it. That is what you want here; saying it out loud so it is not a surprise.

Scratch files are in the untracked `.pr31rev/r2/` (same convention as `.pr27rev/`, `.pr30rev/`). Nothing tracked in the repo was modified.

FINDINGS:
minor|`holder_death_reason` treats an unrepresentable `holder_pid` (OverflowError in `_pid_alive`) as positive proof of death and reclaims the tree, contradicting the module's one-way-doubt invariant|q-system/.q-system/scripts/linear-claim.py:259
minor|This PR makes converge.sh's comment "linear-claim.py deliberately does NOT pid-check the claim itself" false, in the block an operator reads when a board looks wedged|q-system/.q-system/scripts/converge.sh:75
nit|Case 27's `holder_pid_start` fixture guards an upgrade window that never shipped; `git log -S` on origin/main shows no producer ever wrote that key, and case 23 already asserts the behavior|q-system/.q-system/scripts/test/test-linear-claim.sh:507
END FINDINGS

assafkip:
ASK-224 followed up on the Codex major from [this review](https://github.com/assafkip/kipi-system/pull/31#issuecomment-5099123046). Verdict: **Codex was right about the head it read (`7de67d76`) and the finding was already fixed in the merged head (`2936aef`).**

Reproducer, four writer→reader TZ/locale pairings with the holder alive throughout, run against the real historical file via `git cat-file -p 7de67d76:...` rather than a mutant:

| build | result |
| -- | -- |
| `7de67d76` | REPRODUCED 4/4 — live holder reclaimed |
| `2936aef` (merged) | not reproduced 4/4 — exit 3, live holder respected |

The `_PS_ENV` pin at `linear-claim.py:110` (`TZ=UTC`, `LC_ALL=C`) landed in this PR after the reviewed head. All three minors are addressed in the merged head too: the worker echoes `RECLAIMED:` at `linear-worker.sh:817`, the zombie fixture asserts its own preconditions at case 21b, and `cmd_status` appends `[STALE: ...]` with case 28 covering it.

One real gap did come out of it, just not the one Codex named: case 26 varied `TZ` only, so the **locale** half of the pin was unasserted. Deleting `"LC_ALL": "C"` from `_PS_ENV` left all 43 checks green. Fixed in #38 (test-only, 43 → 45 checks); `linear-claim.py` untouched.

PR #32, opened 2026-07-28
fix(fleet): a registered instance that receives no skeleton propagation declares it or fails the run (ASK-117)

assafkip:
Closes ASK-117 (do not merge until /issue-verify and /issue-closeout run).

## The finding, and which of the two cases it is

`reddit-build-radar` was in `instance-registry.json`'s `instances[]` with **0 of its 28 capabilities originating from the skeleton**, while every other instance carried 268-281.

It is **case 1: deliberate.** Evidence, not assumption:

```
$ python3 q-system/.q-system/scripts/capability-map-gen.py \
    --root ~/projects/cole-gtm/projects/reddit-build-radar \
    --repo reddit-build-radar --out /tmp/rbr.json
reddit-build-radar: 28 capabilities (LIVE=17, UNWIRED=11) [local=28] -> 11 actionable+local
origin_counts = {'local': 28}          # no 'skeleton' key at all
```

The entry already had `subtree_prefix: null`, `type: "standalone"`, and a note saying local git lacks git-subtree. `kipi-update.sh` already refused to sync it. So propagation was never broken; the record was.

## What was actually wrong

The skip was **indistinguishable from an oversight**. One line, `SKIP: standalone (not skeleton-managed)`, exit 0. The fleet counted 24 governed instances while one of them ran with no `token-guard.py`, no `.claude/rules/`, no `auto-commit.py`, no capability gate. Every fleet-wide claim about gates holding had an exception nobody could see.

## The change

**`instance-registry.json`** — `"skeleton_managed": false` on `reddit-build-radar`, note extended to state what it therefore does not carry and that fleet governance counts must exclude it. Machine-readable, so a counter can honor it; prose in `note` could not.

**`kipi-update.sh`** — the standalone skip splits in two:

| State | Output | Run |
|---|---|---|
| `skeleton_managed: false` | `SKIP: declared not skeleton-managed (skeleton_managed=false)` | green |
| no declaration | `UNDECLARED NON-PROPAGATING` inline **and in the summary**, with the fix in the message | **fails** |

Only an explicit `false` counts. A missing key is the reddit-build-radar state and must not read like a deliberate opt-out. A missing instance path keeps its own separate message, so a moved directory is not misreported as ungoverned.

The summary line matters: a real run prints ~40 lines per instance, and a flag that only appears on line 300 of 900 is still, in practice, silent.

## Verification

Reproducer first. RED, before the fix:

```
$ bash q-system/.q-system/scripts/test/test-kipi-update-unmanaged-instance.sh
FAIL: a declared non-propagating instance was not named as declared:
  ...
  --- declared (standalone) ---
    SKIP: standalone (not skeleton-managed)
  === Summary ===
    Updated: 0 / Failed: 0 / Skipped: 1
```

GREEN, after:

```
PASS: a declared non-propagating instance is named as declared and does not fail the run
PASS: an undeclared non-propagating instance is flagged in the summary and fails the run
PASS: a missing instance path keeps its own message and is not the non-propagation flag
PASS: the live instance-registry.json declares every entry that receives no propagation
PASS: reddit-build-radar is on the record as deliberately not skeleton-managed
PASS: a registered instance that receives nothing either declares it or fails the run
```

Two of the five assertions run against the **live** `instance-registry.json`, so a future entry that receives nothing and says nothing fails the suite, not just the updater. Registered in `capability-manifest.json` (77 expected tests).

The real registry entry, through the real updater, in an isolated fixture (no instance touched):

```
--- reddit-build-radar (standalone) ---
  SKIP: declared not skeleton-managed (skeleton_managed=false)
=== Summary ===
  Updated: 0 / Failed: 0 / Skipped: 1        EXIT=0
```

No regression in the adjacent updater suites:

```
$ bash q-system/.q-system/scripts/test/test-kipi-update-leak-preflight.sh
PASS: kipi update leak preflight is fail-closed, version-locked, and explicit when unarmed
$ bash q-system/.q-system/scripts/test/test-kipi-update-safety.sh
PASS: differing config content is protected; identical residue and dot-named dirs are not
```

## Not doing

`kipi update` was not run against any real instance, per the DoR. All updater runs here are `--dry-run` against synthetic skeletons in `mktemp -d`.

## Found, not fixed (captured, `sp-d31681f2`)

The DoR's literal check `./kipi update --dry 2>&1 | grep -i reddit-build-radar` **cannot run on `main` right now**, for a reason that predates this branch:

```
propagation leak gate: ABORT -- baseline was built by classifier cc1bbe24488d, running cb5c31f32e12
  Findings MOVED: added 0, removed 28, recounted 0.
```

The committed baseline is stamped against a classifier that has since changed, so every `kipi update` run, `--dry` included, aborts at preflight before the instance loop. Re-baselining requires per-entry human justification by design, so an agent cannot clear it. Hence the isolated-fixture evidence above. Captured as spillover `sp-d31681f2`, not silently worked around.

### DIFF
--- a/instance-registry.json
+++ b/instance-registry.json
@@ -161,7 +161,8 @@
       "type": "standalone",
       "has_git": true,
       "skip_agent_check": true,
-      "note": "Standalone Reddit consensus signal engine for weekly build recommendations. No q-system subtree because local git lacks git-subtree."
+      "skeleton_managed": false,
+      "note": "Standalone Reddit consensus signal engine for weekly build recommendations. No q-system subtree because local git lacks git-subtree. skeleton_managed=false is the deliberate record (ASK-117): it receives zero skeleton propagation, so it carries none of the fleet enforcement layer (no token-guard.py, no .claude/rules/, no capability gate) and must be excluded from fleet-wide governance counts rather than counted as a governed instance."
     },
     {
       "name": "Alice",
--- a/kipi-update.sh
+++ b/kipi-update.sh
@@ -290,6 +290,7 @@ PASS=0
 FAIL=0
 SKIP=0
 GATE_FAIL=""
+UNDECLARED=""
 MODEL_RUN=0
 DRY_MODEL_ROOT=""
 ARCHIVE_TMP=""
@@ -840,7 +841,7 @@ reject_untracked_config_collisions() {
 
 trap cleanup_updater_temps EXIT
 
-while IFS='|' read -r name path prefix itype; do
+while IFS='|' read -r name path prefix itype declared; do
   # Filter INSIDE the loop, not in the feed, so an --only name that matches
   # nothing is caught by the post-loop check rather than reading as an empty
   # registry.
@@ -859,9 +860,26 @@ while IFS='|' read -r name path prefix itype; do
   # Standalone repos have no skeleton subtree; nothing to sync and the updater
   # must not auto-commit or rsync into them. (A null subtree_prefix used to
   # crash the registry parser below -- keep this guard before any mutation.)
+  #
+  # Receiving nothing is legitimate. Receiving nothing SILENTLY is not: this
+  # branch printed one identical line either way, so `reddit-build-radar` sat in
+  # the fleet's 24 governed instances with 0 skeleton capabilities -- no
+  # token-guard, no .claude/rules/, no capability gate -- and every fleet-wide
+  # claim about gates holding had an exception nobody could see (ASK-117). The
+  # declaration is what separates a decision from an oversight, so an entry
+  # without one is a run failure, not a skip.
   if [ "$itype" = "standalone" ] || [ -z "$prefix" ]; then
-    echo "  SKIP: standalone (not skeleton-managed)"
-    SKIP=$((SKIP + 1))
+    if [ "$declared" = "declared" ]; then
+      echo "  SKIP: declared not skeleton-managed (skeleton_managed=false)"
+      SKIP=$((SKIP + 1))
+    else
+      echo "  UNDECLARED NON-PROPAGATING: registered as an instance but receives"
+      echo "    no skeleton propagation, and nothing on record says that is intended."
+      echo "    Add \"skeleton_managed\": false to its instance-registry.json entry"
+      echo "    with a note, or give it a subtree_prefix so it actually syncs."
+      UNDECLARED="$UNDECLARED $name"
+      FAIL=$((FAIL + 1))
+    fi
     echo ""
     continue
   fi
@@ -1491,7 +1509,11 @@ for i in d['instances']:
         continue
     t = i.get('type', 'subtree')
     prefix = i.get('subtree_prefix') or ''
-    print(i['name'] + '|' + i['path'] + '|' + prefix + '|' + t)
+    # Only an explicit false counts as a declaration. A missing key is the
+    # reddit-build-radar state -- registered, receiving nothing, nobody on the
+    # record for it -- and must not read the same as a deliberate opt-out.
+    declared = 'declared' if i.get('skeleton_managed') is False else 'undeclared'
+    print(i['name'] + '|' + i['path'] + '|' + prefix + '|' + t + '|' + declared)
 ")
 
 if [ -n "$ONLY" ] && [ "$((PASS+FAIL+SKIP))" -eq 0 ]; then
@@ -1505,6 +1527,12 @@ echo "  Failed:  $FAIL"
 echo "  Skipped: $SKIP"
 if [ -n "${GATE_FAIL:-}" ]; then
   echo "  CAPABILITY GATE RED in:$GATE_FAIL"
+fi
+# In the summary, not only inline: a real run prints ~40 lines per instance and
+# the summary is what gets read. An ungoverned instance that only appears on
+# line 300 of 900 is still, in practice, silent.
+if [ -n "${UNDECLARED:-}" ]; then
+  echo "  UNDECLARED NON-PROPAGATING:$UNDECLARED"
 fi
 
 [ "$FAIL" -eq 0 ] && [ -z "${GATE_FAIL:-}" ] && exit 0 || exit 1
--- a/q-system/.q-system/capability-manifest.json
+++ b/q-system/.q-system/capability-manifest.json
@@ -82,6 +82,10 @@
       "runner": "bash"
     },
     {
+      "path": "q-system/.q-system/scripts/test/test-kipi-update-unmanaged-instance.sh",
+      "runner": "bash"
+    },
+    {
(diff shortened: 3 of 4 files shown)


assafkip:
## Review: PR #32 — `fix(fleet): a registered instance that receives no skeleton propagation declares it or fails the run (ASK-117)`

Read the diff, extracted the head tree to `.pr32rev/head` (read-only, no tracked file touched), and ran every claim. Repro scripts are in `.pr32rev/repro.sh`, `.pr32rev/repro2.sh`, `.pr32rev/repro3.py`.

---

### minor — `skeleton_managed: false` is honored only when the entry already receives nothing; on a subtree entry it is silently ignored and the instance is fully synced

`kipi-update.sh:861` — the declaration is read *inside* the `standalone || -z prefix` branch, so it has no meaning anywhere else. An entry that says "not skeleton-managed" and has a `subtree_prefix` gets rsync + auto-commit, "Updated: 1", exit 0. The PR's own live-registry assertion (`test-kipi-update-unmanaged-instance.sh:135`) stays green on it, because it only inspects entries that receive nothing. That is the mirror image of the lie ASK-117 set out to kill: the registry states one thing, the updater does another, and nothing refuses the combination.

```
$ bash .pr32rev/repro.sh
############ REPRO B: skeleton_managed=false on a SUBTREE entry
--- optout (subtree) ---
  Syncing q-system/ from skeleton...
[master 38cb405] chore: sync q-system from skeleton 2026-07-27
 4 files changed, 1482 insertions(+), 1 deletion(-)
  OK (4 files updated)
=== Summary ===
  Updated: 1
  Failed:  0
EXIT=0
B: skeleton_managed=false IGNORED -- entry went down the sync path

############ REPRO B2: the PR's live-registry assertion stays green on that same entry
property-3 verdict: PASS (no undeclared non-propagating entry)
```

Dropped a level per the calibration rule: nothing in this repo emits that shape today (`kipi-new-instance.sh:179` always writes a real `subtree_prefix`), so I had to construct it. Recoverable via git in the instance. One line closes it: refuse `skeleton_managed:false` alongside a non-empty `subtree_prefix`, in the feed parser or in property 3.

---

### minor — `kipi check`'s registry loop still prints the exact line the PR removed from the updater

`validate-separation.py:1123-1126` carries a byte-identical copy of the old branch. The founder's health-check command reports an undeclared non-propagating instance as a benign yellow SKIP, indistinguishable from a deliberate one — the precise defect the PR describes, surviving in the sibling loop.

```
$ python3 .pr32rev/repro3.py
=== Phase 4: All instances ===

  --- undeclared-standalone (standalone) ---
  SKIP undeclared-standalone: standalone (not skeleton-managed)
```

(The entry handed to `phase_4` is `type: standalone`, `subtree_prefix: null`, no `skeleton_managed` key.) Bounded, because the updater and the manifest-registered test still hold the invariant — but the validator's output now contradicts the updater's.

---

### minor — an undeclared non-propagating instance whose path has moved bypasses the guard entirely

`kipi-update.sh:849` (missing-path check) runs before `:861` (declaration check), so a stale path short-circuits first: quiet skip, exit 0, no `UNDECLARED NON-PROPAGATING`. Stale registry paths are a state this fleet actually produces (the cole-gtm and ktlyst-hub reorgs both moved instance directories), so this is not a constructed input.

```
$ bash .pr32rev/repro.sh
############ REPRO A: undeclared non-propagating instance whose path is missing
--- ghost-undeclared (standalone) ---
  SKIP: path /var/folders/.../does-not-exist does not exist
=== Summary ===
  Updated: 0
  Failed:  0
  Skipped: 1
EXIT=0
A: NOT FLAGGED (guard did not fire)
```

The operator does get *a* message, and property 3 catches such an entry in the test suite regardless of path, which is why this is minor and not higher.

---

### nit — the new manifest-registered test leaks three temp dirs per run

`test-kipi-update-unmanaged-instance.sh` calls `mktemp -d` three times with no cleanup trap; it is wired into `capability-manifest.json`, so every capability run adds ~1.4 MB.

```
$ bash .pr32rev/repro2.sh
tmp.* dirs before=11880 after=11883 delta=3
480K	/var/folders/.../tmp.GNhzvaj5m9
480K	/var/folders/.../tmp.UfkSvDCtFi
480K	/var/folders/.../tmp.pp8gjAJ6le
grep for a cleanup trap in the test: 0
```

Half the suite already does this (36 tests use `mktemp -d`, 18 have a cleanup trap), so this is convention, not a regression this PR introduces. The 11,880 standing dirs are the aggregate.

---

## What is sound

- **"The test is decoration."** Swapped `origin/main`'s `kipi-update.sh` under the new test. It goes genuinely RED, on the real assertion, with real output: `FAIL: a declared non-propagating instance was not named as declared: ... SKIP: standalone (not skeleton-managed) ... Updated: 0 / Failed: 0 / Skipped: 1`, exit 1. Restored head → all six PASS lines, exit 0.
- **"The fixture is invented."** It is not a mock of the thing under test: it copies the real `kipi-update.sh` and runs it end to end, and two of the five assertions read the *live* `instance-registry.json`. Its one synthetic artifact (the leak-gate baseline) is honest about disarming the gate — the run prints `propagation leak gate: NOT ENFORCING`, so nothing is hidden behind a fake pass.
- **"This turns the fleet red tomorrow."** Ran the real 24-entry registry through the new guard: one non-propagating entry, declared. No currently-registered instance newly fails. `kipi-new-instance.sh` writes a real `subtree_prefix`, so the normal creation path cannot emit an undeclared entry either.
- **"It cries wolf on a moved directory."** It does not. A missing path with a valid prefix keeps its own message and is never labeled ungoverned — verified, and the test pins it.
- **"A missing key reads as an opt-out."** It does not. `i.get('skeleton_managed') is False` rejects absent, `null`, `0`, and `"false"`.
- **Rollback:** a bash branch plus a JSON key, no state written anywhere. `git revert` is complete.

## VERDICT: APPROVE WITH NITS

No blocker, no major. Nothing here is permanent, nothing runs unattended into an unrecoverable state, and the guard does what the title says for every shape the fleet actually produces today.

**Fix first:** make `skeleton_managed: false` mean one thing. Right now it is honored on a standalone entry and ignored on a subtree entry, which lets the registry assert "ungoverned" about an instance the updater is actively syncing — and the PR's own live-registry test cannot see it.

FINDINGS:
minor|skeleton_managed:false is ignored on a subtree entry, which is synced anyway while the registry declares it unmanaged|kipi-update.sh:861
minor|validate-separation.py phase_4 still prints the identical undifferentiated "standalone (not skeleton-managed)" skip that the updater no longer prints|validate-separation.py:1123
minor|the missing-path check precedes the declaration check, so an undeclared non-propagating instance with a stale path is skipped quietly at exit 0|kipi-update.sh:849
nit|the new manifest-registered test leaks three ~480K temp dirs per run with no cleanup trap|q-system/.q-system/scripts/test/test-kipi-update-unmanaged-instance.sh:70
END FINDINGS

PR #34, opened 2026-07-28
feat(review): codex reviews every PR as an independent second opinion, with an Opus fallback that pages when it degrades (ASK-221)

assafkip:
Closes ASK-221.

Codex (`gpt-5.6-sol`, senior-staff-at-Meta persona) becomes THE reviewer of Sana's work, unattended. Founder directive 2026-07-29.

The gap this closes: `linear-worker.sh` already routed review through `pr-review-agent.sh`. This makes that script's primary engine codex, so the required `kipi/reviewer-approved` status is a Codex verdict rather than a Claude one. The reviewer was `claude -p` and the PR author is also Claude: a different process with no shared memory, genuinely useful, but the same lab and model family, so the blind spots stay correlated. Fresh context is not an independent mind.

## Why an engine flag and not a separate script

Sha capture (ASK-216), verdict derivation from labelled severities, the commit-status post (ASK-217) and spillover capture all stay shared and identical. A second script would be a second writer with its own semantics. What the engine changes is exactly three things: which binary runs, which status context it posts, which directory its artifacts land in.

Codex is PRIMARY: it owns `kipi/reviewer-approved` (the required context) and writes the one verdict record `converge.sh` and `linear-worker.sh` gate on. Claude keeps the same script but posts an advisory `kipi/claude-approved` and writes its record out of the gate's way. When codex is down, Opus fills the primary slot and the status says DEGRADED out loud, so an outage degrades independence instead of wedging every open PR.

## The three Codex findings on this branch, each verified with a reproducer

The fixes were already on the branch. None of them had been RUN. Reading a fix is not running it.

| # | Sev | Claim | Fixed at | Reproducer, before -> after |
|---|-----|-------|----------|----------------------------|
| A | minor | `--agent` matched the attribution marker anywhere in the body | `de2a9c3` | `KIPI_TEST_LINEAR_SYNC_REF=63f81de` -> 13 passed, 2 failed; working tree -> 15 passed, 0 failed |
| B | minor | first-line anchor still accepted `**sana** prose` | `403bc0b` | `ref de2a9c3` -> 14 passed, 1 failed (anchor passes, delimiter fails) |
| C | **major** | no tree-vs-PR-head guard: codex ran and a verdict was derived on a head not in the tree | `403bc0b` | `KIPI_TEST_REVIEWER_REF=de2a9c3` -> FAIL; working tree -> PASS 10/10 |

Each `--agent` case now maps to exactly one fix commit, so neither is a test that cannot fail.

**C is the one that mattered.** Its refusal branch had never once executed. Every reviewer case in `test-severity-floor.sh` reports `SHA_A=a1b2c3d4...`, a fabricated sha that `git cat-file -e` misses, so all of them take the guard's tier-1 WARN branch. Reaching the refusal needs a sha that is a real object AND not an ancestor, which needs its own sandbox repo: `git commit-tree` on HEAD's tree with no parent. Deterministic, and it does not depend on which remote branches happen to be fetched, which is what makes a "use another branch's sha" version of the test pass or fail by accident on a fresh clone.

`test-review-tree-guard.sh` against `de2a9c3` reproduces the live symptom verbatim: `verdict: APPROVE` on a head not in the tree, codex dispatched, verdict record written. Case 2 is the negative self-test: a guard that refuses everything would pass case 1 while breaking every real review, so the same harness drives the sandbox's own HEAD and asserts codex DID run and APPROVE WAS derived. The codex stub touches a marker file rather than printing, because the live symptom was `codex_ran=yes` printed next to a bogus verdict, and prose is not evidence.

## One findings-block reader (sp-c0a9dac3)

Three call sites each ran their own `sed -n '/^FINDINGS:/,/^END FINDINGS/p'`, two of them inside the file that exists specifically to stop two readers of one input. sed restarts that range after every closing line, so the expression is wrong in two directions at once. Both directions are now held by `findings_block()` in `pr-verdict-lib.sh` and covered by `test-findings-block-reader.sh`, which fails on ref `403bc0b` and passes on this branch:

- **Blocks concatenate.** `ROUND_RULE` carries the previous round's findings into the round-2 review text, so two blocks in one stream is a routine shape, not an edge case. A review that quoted round 1's blocker and then refuted it derived `BLOCK` from the quoted one, and an approved PR then wedges forever on a finding that does not exist. Reproduced: `BLOCK` on ref `403bc0b`, `APPROVE` after.
- **An unclosed block runs to EOF.** No severity lines derives APPROVE: a green gate for a review nobody read. Now empty, which reads as unstated.

Then the reproducer found the deeper case: "last COMPLETE block" quietly falls back to a QUOTED prior-round block when the real trailing block is cut off mid-write, so a truncated review derived a verdict from findings it had already withdrawn. A block still open at EOF now voids the whole review, earlier complete blocks included.

The completeness predicate was a second definition of "complete" (both markers, anywhere, in any order). It gates `REVIEW_UNUSABLE`, the one thing stopping a truncated review from posting `state=success` on the required context, and it accepted exactly that shape. Now `has_complete_findings_block` in the lib, defined as `findings_block is non-empty`, so the predicate and the extractor cannot answer differently.

The third site was the reviewer's Linear comment. Left alone it could carry a different set of findings than the verdict came from: Sana answering findings that never set the gate, on a gate set by findings she never saw.

`extract_minor_findings` feeds `spillover add`, which writes permanent ledger items. The multi-block bug filed follow-ups for findings the same review WITHDREW, and the standing gate then stays red on them. The reproducer asserts 1 captured, not 3.

Proof of the property, comments stripped so it asserts on code rather than on prose about code (`findings_block`'s own comment quotes the expression it replaced):

```
$ for f in pr-review-agent.sh pr-verdict-lib.sh; do
    echo "$(grep -v '^[[:space:]]*#' $f | grep -c 'FINDINGS:/,/')  $f"; done
0  pr-review-agent.sh
0  pr-verdict-lib.sh
```

One `findings_block()` definition, three call sites. `test-findings-block-reader.sh` asserts that count so it cannot drift back.

## verify-codex-review-live.sh, salvaged from the closed PR #45

PR #45 (ASK-253, Linear-agent delegation) is closed unmerged: its status was advisory so it could not gate, five live attempts failed, and it added another verdict reader. Its load-path discipline was the best thing on it and is kept, retargeted from the delegation path to the engine path: resolve the launchd plist, read `ProgramArguments.1`, derive the live checkout, assert against the files the scheduler will actually execute, and RUN the live tree's own test files instead of greping them for strings.

First run against `main` at `99f213e`: `RESULT: NOT WIRED (7 failed)`. Correctly, because this work is on a branch. That is the check doing its job, not a bug.

## Evidence

```
test-review-tree-guard.sh        PASS 10/10
test-findings-block-reader.sh    PASS 9/9
test_linear_sync_comments.py     15 passed, 0 failed
test-severity-floor.sh           PASS 170/170   (no regression)
capability-gate.py               exit 0
validate (CI)                    pass
```

## Ledger

- `sp-c0a9dac3` fixed here.
- `sp-c5a17b4a` voided with reason: the shipping reviewer builds its own prompt inline and invokes `codex exec -C SKEL "$PROMPT"`, so the codex-engine path never loads the third-party `adversarial-review.md` and has no vendored prompt to scope-filter. The uncommitted marketplace edit that prompted it is reverted (diff preserved), because it tuned the reviewer of Sana's work toward easier approval and lived in a clone that plugin updates silently overwrite.
- `sp-b249b1b1` and `sp-d1b71b27` carried into **ASK-254**: the verifier has no automated caller, which is the same failure mode it exists to catch.

## Reverted from this branch

Nothing. The four commits an interactive session wrote (`099c324`, `63f81de`, `de2a9c3`, `403bc0b`) were each put against a reproducer and each held. What they lacked was proof, not correctness. The two acceptance comments that reported them fixed without running them are withdrawn on ASK-221, append-only, originals untouched.


assafkip:
**PARKED** 2026-07-29, founder decision.

This PR makes codex the PR reviewer through `pr-review-agent.sh` (shell glue: an `--engine codex` flag, a commit status, and hand-built Linear comment plumbing). It is green (`validate=SUCCESS` @ 099c324) and it works.

It is parked because the NATIVE path is better and is already installed. The founder installed the Codex Linear agent on 2026-07-29 20:08 UTC specifically to review Sana's work in the issues, plus GitHub on the Codex side.

The decisive difference is not line count. This PR posts a commit status called `kipi/reviewer-approved` instead of a real GitHub review, because the reviewer runs as the same account that authors these PRs and GitHub forbids self-approval (see the WHY A STATUS AND NOT A PR REVIEW comment in the script). A GitHub App is a separate identity and can actually approve. The native path removes the workaround rather than maintaining it.

Kept from this branch's work regardless of what happens here:
- the base refresh that cleared the inherited capability-gate REDs (same code, fresh base, CI failure -> success)
- `test-severity-floor.sh` `timeout_s 420`, measured: 174s here, 158s at the pre-change baseline, so this suite was always past the 60s default and had been passing the gate on luck

Two defects found on this branch apply to `pr-review-agent.sh` ON MAIN as well, so they stay open and are NOT parked with this PR:
- `sp-c0a9dac3` `verdict_from_findings` parses the whole review file, so a reviewed diff containing FINDINGS: text feeds fixture findings into the real verdict
- `sp-a72a9567` nothing asserts the working tree matches the PR's head sha, so a reviewer run from the wrong worktree reports findings absent from the diff under a sha that makes them look authoritative

Not closed, not merged. Reopen with `gh pr ready 34` if the native reviewer turns out not to cover issue-side review.

assafkip:
## Triage of codex review round 1 (by the implementing agent, not a human)

Round 1 ran for real against `1d5347d`: engine `codex`/`gpt-5.6-sol`, `degraded=0`,
verdict `REQUEST CHANGES` (stated and derived agreed, source `findings`), and it
posted `kipi/reviewer-approved=failure` on that sha. Both findings are **accepted**.
Neither was rejected or deferred.

### major -- the tree guard rejected every autonomous review. ACCEPTED, fixed.

Codex was right and the consequence is worse than it stated. `$SKEL` follows
`BASH_SOURCE`, and the autonomous caller is `linear-worker.sh:1133`, which runs
`bash $SCRIPT_DIR/pr-review-agent.sh` out of the **main checkout** while the PR's
commits live in a worktree cut under `$STATE_DIR/worktrees/<issue>`. cwd never
enters into it. So the PR head could never be an ancestor, the guard refused
always, and the call site swallows a non-zero exit as
`|| say WARN ... (the PR stands, unreviewed)`.

Merging `1d5347d` would have made the autonomous loop review nothing and say almost
nothing about it. Cases 1-3 of the guard's own test all ran the reviewer from the
same checkout they asked about, which is why they were green while the live shape
was broken.

Fixed in `b090ba7`: the guard now asks *which tree holds this commit* (worktrees
share one object database) and reviews that tree, with `codex -C` and the prompt
following it. Refusal is kept for the case where no tree holds the commit, which is
the original `sp-a72a9567` shape.

Reproducer is case 4, built to the worker's exact shape:

```
KIPI_TEST_REVIEWER_REF=1d5347d bash test/test-review-tree-guard.sh
  -> FAIL: THE DEFECT: the reviewer REFUSED the autonomous call shape
     REFUSING: PR #901 is at 72a1ea88..., which is NOT in the history of .../repo
bash test/test-review-tree-guard.sh
  -> PASS: 15/15 tree-guard checks
```

Cases 1-3 still pass, so case 1 is case 4's negative self-test: a resolver that
devolved into "always proceed" turns it red.

### minor -- the severity-floor fixture wrote to the real object database. ACCEPTED, fixed.

Also correct. `git -C "$ROOT" commit-tree` wrote a loose object into the founder's
live repository on every run, and cannot work at all in a read-only checkout. That
is the live-data-path rule this fleet lints for.

The block is removed rather than sandboxed in place, because the assertion already
lives in `test-review-tree-guard.sh` case 1 in a `mktemp` sandbox repo, and that
version is a strict superset (non-zero exit, the `REFUSING` reason, no codex
dispatch, no Opus dispatch, no verdict record, no commit status). A pointer comment
replaces it so the move is not silent.

### Adjacent, captured not fixed

`test-severity-floor.sh` blocks on inherited stdin and runs for minutes; without
`</dev/null` it does not terminate. Root cause folded onto the existing ledger item
`sp-d3a22cb6` (same suite, same timeout) rather than filed as a duplicate.

Round 2 follows on the new head.

assafkip:
## Round 2 triage and merge record (implementing agent)

Round 2 ran for real against `fa4608a`: engine `codex`, `degraded=0`, one **minor**
finding, derived verdict `APPROVE WITH NITS`, and it posted
`kipi/reviewer-approved=success` on that sha. `validate` was already green. Merged as
`f277389`.

Note the reviewer's own PR comment failed on that run (`WARN: could not comment on
PR`), which is why this is posted by hand. That is the same class as the minor it
found, and it is on the ledger.

### The minor: ACCEPTED, captured, not fixed here

`pr-review-agent.sh:660` -- a rejected Linear review comment is silently discarded, so
an unattended run exits 0 and prints `done` while the issue never receives the review.
Captured as `sp-583dc1a0`. Not fixed in this PR: it is a genuine defect but outside the
two findings this round was answering, and the ledger gate keeps it from being
forgotten.

### One more, found by watching this run rather than reading the diff

`extract_verdict` returned `REQUEST CHANGES` for a review whose own headings said
`## VERDICT: APPROVE WITH NITS`. Cause: from round 2 the prompt hands the reviewer the
earlier rounds, so codex discussed round 1's verdict in prose and the prose read picked
it up. Derived-over-stated caught it here because the findings block won. A review with
no findings block would fall back to prose and inherit the wrong round's verdict.
Captured as `sp-337b40ef`; the fix is to anchor the stated read to the LAST
`## VERDICT:` heading, which is the same last-complete-block rule the findings reader
already uses.

### What is now live, and what is still unproven

The merge is live in the copy launchd actually executes. The loaded plist runs
`/Users/assafkipnis/projects/kipi-system/kipi-dispatch.sh`, that checkout was at
`1597eaf` and did not contain any of this, and `kipi-dispatch.sh` has no `git pull` of
its own -- so the merge alone would have left the loop running the old Claude-only
reviewer. Fast-forwarded to `f277389`; `pr-review-agent.sh` now carries `REVIEW_ROOT`
at 222/240 and `linear-worker.sh:1133` calls `--engine codex`.

Still unproven, and it cannot be proven tonight: a real dispatcher run. The daily cap
is spent (`dispatch-count-2026-07-29` = 3, loaded `KIPI_DISPATCH_DAILY_MAX` = 3) and a
budget day runs 07:00 -> 06:59, so the allowance refills at 07:00 local. The dispatcher
keeps beating every 15 minutes until then and each tick is a no-op skip.

PR #3, opened 2026-05-28
chore: replace KTLYST flavor in skeleton stat-verify code

AuthenTech-AI:
Small companion to #2 (the hardcoded-patterns discussion) — surgical mechanical change while that broader design conversation plays out.

## Problem

Two files in the skeleton contain references to `KTLYST` / `q-ktlyst` / `ktlyst-*` that aren't real instance content leaks — they're leftover example flavor baked into the skeleton itself:

- `q-system/.q-system/scripts/stat-registry-extract.py:147` — stderr message hardcodes `q-ktlyst/canonical/stat-registry.json` as the path the user should edit
- `q-system/.q-system/scripts/tests/test-stat-verify.py` — docstring (`q-ktlyst-shaped tree`) plus two test fixture IDs (`ktlyst-spec`, `ktlyst-acv`)

These trip the leak-detection gate (`validate-separation.py` + `kipi-mcp` validator) in any fork that runs the validators against this skeleton's own content. Forks have to either patch the validator's pattern list locally or live with red CI.

## Fix

Replace with generic placeholders:

- `q-ktlyst/canonical/...` → `the instance's canonical/...` (stderr message)
- `q-ktlyst-shaped tree` → `instance-shaped tree` (docstring)
- `ktlyst-spec` → `test-spec`, `ktlyst-acv` → `test-acv` (test IDs)

## Safety

- `git grep "ktlyst-spec\|ktlyst-acv"` confirms the test IDs are referenced **only** within `test-stat-verify.py` — no cross-references, safe to rename.
- No logic change. No assertion change. Tests still describe and exercise the same shapes.
- The stderr message preserves the same information (what to update); just doesn't tie it to one specific instance name.

## Why now

Forks running `validate-separation.py 1` on this skeleton hit 3 KTLYST-related failures today (1 from `init-bus-day.sh` already independently fixed upstream; 2 from these files). This PR clears the 2 remaining and brings the leak-detection gate to green for any fork. The broader pattern-config discussion in #2 stands on its own.

Issue #2, opened 2026-05-26
Eliminate hardcoded paths, GitHub URLs, and instance-specific patterns (so forks work without local patches)


AuthenTech-AI:
Following up on #1 (the portable-sed fix), there's a broader portability theme worth discussing before sending more PRs. Anyone forking `kipi-system` today has to maintain a local patchset of 20+ edits before the skeleton works as their own. Most of it is mechanically replaceable; some of it needs a small design decision first.

Filing this as an issue rather than a PR because the design choices warrant your input before code lands.

### 1. Hardcoded GitHub URL (`https://github.com/assafkip/kipi-system.git`) — 16 sites

`git grep "github.com/assafkip" upstream/main` finds:

**Code (5 files):**
- `kipi-new-instance.sh:9`, `kipi-push-upstream.sh:8`, `kipi-update.sh:14`, `kipi-migrate.py:33`, `q-system/hooks/auto-update.sh:34`
- `instance-registry.json:4`

**Plugin manifests (5 files):**
- `plugins/{kipi-core,kipi-design,kipi-dsse,kipi-notebooklm,kipi-ops}/.claude-plugin/plugin.json` (each `\"repository\"` field)

**Docs (5 files):**
- `README.md:68`, `SETUP.md:25`, `INSTANCE-SETUP.md:88`, `ARCHITECTURE.md:19`, `UPDATE.md:119`

**Proposed fix:** env-var override pattern — `SKELETON_REMOTE=\"${SKELETON_REMOTE:-https://github.com/assafkip/kipi-system.git}\"`. The default stays your URL; forks override via env. For plugin manifests and docs, the URL is informational; keep it as the canonical reference but mention the override in `INSTANCE-SETUP.md`.

Effort: ~30 lines of code change, no behavior change for users who don't set the env var.

### 2. Hardcoded Mac filesystem paths (`/Users/assafkip/...`) — ~3 live-code sites

Most matches are in historical PRD docs (`q-system/output/prd-*-2026-04-*.md`) and aren't worth touching. The live-code sites:

- `plugins/kipi-core/kipi-mcp/src/kipi_mcp/validator.py:12` — `HARDCODED_PATH_PATTERNS = [r\"/Users/assafkip\", r\"q-ktlyst/\"]`
- `plugins/kipi-core/kipi-mcp/tests/test_validator.py:230` — test fixture string
[...]

PR #35, opened 2026-07-28
feat(triage): a duplicate-template family collapses to one issue whose DoR names every member's target (ASK-226)

assafkip:
Closes the arithmetic in ASK-226: the ~200-issue ASK board is roughly 20 real changes wearing 200 hats.

## The measurement

From `q-system/output/triage-applied-2026-07-28.log` (106 issues classified): 56 triaged `batch`, and 49 of those 56 are the same machine-filed template repeated — 23 near-identical job-migration issues, the CAP-0n rule family, plus scanner output. The triage reasons say it outright ("First of five identical CAP-0n rule issues", "Runs the identical script as ASK-156").

## Why collapse and not "just close them"

Closing 49 issues deletes the record of what each one named. Some carry a distinct path or job label that the family fix has to cover, and a fix that skips one member's target is how a job stays dark.

So: ONE survivor per family whose description enumerates **every** member's specific target, and every absorbed issue closed with a comment pointing at the survivor. Nothing is deleted; reopen is always available, and the pointer is what makes reopening meaningful.

## What ships

- `q-system/.q-system/scripts/linear-collapse.py` — deterministic family detector, dry by default, `--apply` is the only thing that writes.
- `q-system/.q-system/scripts/test/test-linear-collapse.sh` — 21 cases, no network.
- `q-system/.q-system/capability-manifest.json` — the test in `expected_tests`, the engine in `declared_inert` (both in sorted position, no re-sort).

## The three decisions worth reviewing

**1. The detector is strict, not fuzzy.** Two issues collapse only when their titles normalise to the byte-identical template *and* the same producer filed them. No model call, no similarity threshold. This under-collapses in the tail (a hand-edited title drops out of its family) and that is the safe direction: under-collapse leaves an issue on the board for a human, over-collapse closes real distinct work behind a fix that never covers it. `is_collapsible_template()` refuses a template with no literal content, so two identifier-only titles cannot collide into a family.

**2. Write order is the contract.** Survivor roster first, then per member: comment, *then* close. A close with no pointer is an orphan — the issue is off the board and nothing links it to the survivor that inherited its target. Asserted by ordering, not by presence.

**3. The record is appended per issue, never batched.** sp-b5dcf944: `linear-triage.py --apply` died mid-run after closing 32 issues and left no audit file at all, because the verdict file was written after the loop. Here every step appends and fsyncs before the next step runs. The suite drives a fake writer that fails if a close happens before its own record is on disk. A member commented by a dead run is *resumed*, not skipped, or it sits open forever behind a pointer saying it was collapsed.

## Checks (DoR)

```
$ bash q-system/.q-system/scripts/test/test-linear-collapse.sh
== detect_families: one family per template, never per issue ==
  ok   the 5 CAP issues group into exactly one family
  ok   the survivor is the lowest-numbered member
  ok   every other member is absorbed, none dropped
  ok   the 3 job-migration issues group into one family
== over-collapse guards: destroying real work is the worse failure ==
  ok   a near-miss sharing a word is not absorbed
  ok   the near-miss stays out of the family roster
  ok   identical template from a different producer is a separate family
  ok   a template with no literal content is refused
  ok   a lone issue is not a family
== survivor DoR: the union of targets, not a sample ==
  ok   every CAP member's rule file is named in the survivor DoR
  ok   every job-migration label is named in the survivor DoR
  ok   every member identifier is named in the survivor DoR
  ok   the absorb comment names the survivor
== apply_family: comment before close, record before both ==
  ok   each member is commented before it is closed
  ok   every absorbed member is both commented and closed
  ok   the record is on disk before the close that follows it
  ok   a crash after the first close still leaves the completed steps
== idempotency: a second --apply must not double-write ==
  ok   an already-absorbed, already-closed member is not touched again
  ok   a commented-but-open member is still closed on resume
  ok   re-splicing an unchanged survivor block is a no-op
== dry by default: --apply is the only thing that writes ==
  ok   no --apply writes nothing and says so

passed 21, failed 0
```

RED was observed first: `passed 0, failed 21` against the missing script.

`capability-gate.py --check-only` → GREEN. The full run (78 tests) is in flight and its result is posted as a comment below.

## Unwired by design

`linear-collapse.py` ships with a `declared_inert` entry (sp-ffc69720), same posture as `linear-triage.py`: `--apply` closes dozens of permanent Linear objects, so no automatic trigger exists until the founder reads a dry-run roster of the real board. The wiring destination is the same gate — collapse must run BEFORE the DoR drafter manufactures readiness on duplicate-template issues.

## Not doing (binding, per the DoR)

- Retiring or un-pausing any `com.cole.*` job. They are paused deliberately pending the Linear migration; this only collapses the ISSUES that track them.
- The `needs-scope` (10) or `founder-decision` (2) buckets.
- Re-running the crashed triage tail (9 unclassified issues).
- Parallel dispatch (ASK-225).

Do not merge — closeout runs through `/issue-verify` and `/issue-closeout`.


assafkip:
Full capability gate finished (it was in flight when the PR opened).

```
capability-gate mode=skeleton
  declared: 78 tests (0 quarantined), 17 skeleton-only, 14 declared-inert
  ...
  DECLARED-INERT (sp-ffc69720): q-system/.q-system/scripts/linear-collapse.py — built and tested (21 cases) but deliberately unwired: --apply closes dozens of permanent Linear objects in one pass, so no trigger exists until the founder reads a full dry-run roster of the real board
  tests: ran=78 quarantined=0 skipped-skeleton-only=0
capability-gate: GREEN
```

All 9 DoR checks are covered:

| # | Check | Where |
|---|---|---|
| 1 | dry by default, `--apply` is the only writer | `no --apply writes nothing and says so` (drives the real entry point via `KIPI_COLLAPSE_FIXTURE`) |
| 2 | the 5 CAP issues → one family, one survivor | `the 5 CAP issues group into exactly one family` + `the survivor is the lowest-numbered member` |
| 3 | survivor DoR carries the **union**, not a sample | `every CAP member's rule file is named in the survivor DoR` (asserts `len(want)==5 and all(...)`) |
| 4 | a near-miss is not collapsed | `a near-miss sharing a word is not absorbed` + `identical template from a different producer is a separate family` + `a template with no literal content is refused` |
| 5 | comment before close, ordering asserted | `each member is commented before it is closed` (index comparison, not presence) |
| 6 | record appended per issue, never batched | the fake writer fails the suite if a close lands before its own record is on disk; plus `a crash after the first close still leaves the completed steps` |
| 7 | re-running `--apply` is idempotent and resumes | `an already-absorbed, already-closed member is not touched again` + `a commented-but-open member is still closed on resume` |
| 8 | `test-linear-collapse.sh` green | passed 21, failed 0 |
| 9 | `capability-gate.py` green | above |

Two notes for the reviewer, neither hidden:

**The manifest got two entries, not one.** The DoR said exactly one new line. `expected_tests` needs the test, and `capability-gate.py` goes RED with `inert-engine: linear-collapse.py has no reference on any wiring surface and no declared_inert entry` — observed, then fixed. Check 9 and a one-entry manifest are mutually exclusive here, so the engine got a `declared_inert` entry backed by a real spillover item (sp-ffc69720) rather than a bypass. Both entries went into sorted position; the file was not re-sorted (sp-f3a2ad81).

**Under-collapse is the deliberate residual.** Strict template equality means a hand-edited title drops out of its family and stays on the board. That is chosen, not missed: over-collapse closes real distinct work behind a fix that never covers it, and check 4 is the guard in that direction. If the dry run on the real board shows a family splitting on a title variant, the fix is to look at that variant, not to loosen the matcher.

assafkip:
## Review: PR #35 — `linear-collapse.py` (ASK-226)

Ran the shipped suite first: **21/21 green**. Then attacked it. Three of my attacks reproduced; the two that matter both hit the survivor roster, which is the one artifact this whole PR exists to produce.

I could **not** run the shipped script against the live ASK board — the session sandbox blocks `~/.config/kipi/` and the network, so the Linear API key is unreachable. The biggest open question ("does the detector over-collapse on the real 200-issue board?") is therefore **unverified by me**. Saying that loudly because it is the one check that would settle the risk this design is built around.

---

### 1. major — the survivor's target list is built from its own description, so on the second write it ingests the block the script just wrote and truncates the survivor's real target out

`q-system/.q-system/scripts/linear-collapse.py:203`

`member_targets()` scans `title + description`. After pass 1, the survivor's description **contains the roster block**, which lists every other member's target. Pass 2 re-reads it, and the survivor's own row now claims to own the whole family's targets. `MAX_TARGETS_PER_ISSUE = 20` (`:152`) then truncates — and what gets cut is the survivor's own real target.

```
python3 .pr35rev/repro_b_c.py
```

Real output (22 job-migration issues built from the real producer's real format, `linear-job-migration.py:166`):

```
family size: 22
survivor targets, pass 1: ['com.cole.job01', 'q-system/.q-system/scripts/job01.py']

survivor targets, AFTER the pass-1 write (what pass 2 will read):
    com.cole.job01
    com.cole.job02
    ...
    com.cole.job20
count: 20  (MAX_TARGETS_PER_ISSUE = 20 )
survivor's OWN target is com.cole.job01 -> True
truncated at the cap? -> True

pass-2 survivor row (truncated to 300 chars):
    | ASK-301 (this issue) | Migrate com.cole.job01 to Linear-tracked execution | `com.cole.job01`, `com.cole.job02`, `com.cole.job03`, ...
job labels NOT named in the survivor's own row: ['com.cole.job21', 'com.cole.job22'] ...
```

`q-system/.q-system/scripts/job01.py` — the survivor's actual script path — is gone from the row. In the 5-issue CAP run the ingested string included `` `linear-collapse.py` ``, scraped out of the block's own footer sentence.

The docstring at `:247` says "a fix that covers the shape but skips one of these rows leaves that target dark." That is what this does to its own row. A human has to hand-repair the survivor description.

---

### 2. major — a resumed pass DELETES the already-closed members' rows from the survivor roster, and the header count lies about it

`q-system/.q-system/scripts/linear-collapse.py:432` (`fetch_open` drops every closed issue) + `:355` (`write_survivor_block` → `splice_block` replaces the whole block)

Production only ever feeds `detect_families()` from `fetch_open()`. So on pass 2 the members that **pass 1 already closed** are not in the input. The family shrinks, and the roster is rewritten without them — while their pointer comments still say "enumerated in ASK-130, so the family fix has to cover it."

```
python3 .pr35rev/repro_roster_shrink.py     # crash after 2 closes, then resume
python3 .pr35rev/dbg_dump.py                # same run, dumps the description
```

Real output (5 CAP issues, run 1 dies after closing ASK-131 and ASK-132, run 2 is the documented resume):

```
run 1 died: simulated mid-run death (the sp-b5dcf944 scenario)
run 1 roster rows present: ['ASK-130', 'ASK-131', 'ASK-132', 'ASK-133', 'ASK-134']
closed by run 1: ['ASK-131', 'ASK-132']

run 2 sees 3 open issue(s), 1 family/families
run 2 roster rows present: ['ASK-130', 'ASK-133', 'ASK-134']
```

The block header goes from `## Collapsed family — 5 issues, one change` to `## Collapsed family — 3 issues, one change`. ASK-131 and ASK-132 are closed, off the board, and no longer named in the record that closed them.

No crash is needed either: after a clean pass, one new sibling filed later gives a family of `{survivor, new}` and rewrites the roster down to 2 rows.

Why the suite missed it: the test `an already-absorbed, already-closed member is not touched again` (`test-linear-collapse.sh`, the a5 case) hand-sets `state.type = "canceled"` on members and passes them to `detect_families`. `fetch_open()` can never emit that. The one idempotency case covering closed members drives a state the real producer does not produce, and the real resume path — roster shrinkage — has no case at all.

---

### 3. minor — with no `canceled`-type state on the team, every absorbed issue gets its permanent "collapsed" comment, nothing closes, and the run exits 0 saying "applied"

`q-system/.q-system/scripts/linear-collapse.py:373`, `:524`, `:536`

`close_issue()` returns the string `"NOT-CLOSED (...)"` instead of raising. `apply_family` records it as the step outcome, `main()` does `done += 1` unconditionally, prints `1/1 family/families applied.` and returns 0.

```
python3 .pr35rev/repro_b_c.py    # section C
```

```
  ASK-400  survivor-dor  dor-written
  ASK-401  commented     comment-added
  ASK-401  closed        NOT-CLOSED (no canceled-type state on this team)

apply_family raised? no. main() would now do done+=1 and print '1/1 family/families applied.' and return 0
graphql mutations actually sent: [['description'], ['body', 'issueId']]
ASK-401 state after the run: backlog <- still OPEN, carrying a permanent comment that says it was collapsed
```

Dropped to minor per the calibration rule: I could not query the ASK team's workflow states to confirm a team without a `canceled` state exists, so the trigger is constructed. The success-reporting half (`done += 1` regardless of per-step outcome) is unconditional code, not input-dependent.

---

### 4. minor — `producer()` is documented as "which scanner filed this", but on the real board it returns the project name for most issues, so the different-producer guard is inert there

`q-system/.q-system/scripts/linear-collapse.py:182-183`

`KIPI_KEY_RE` takes the first path segment of the kipi-key. The suite's fixtures invent `cap-scan/`, `job-migration/`, `other-scan/` — scanner namespaces — and the test `identical template from a different producer is a separate family` rests on that invention. The real ledger:

```
python3 -c "import json,collections; rows=[json.loads(l) for l in open('q-system/output/linear-ledger.jsonl') if l.strip()]; iss=[r for r in rows if r.get('kind')=='issue']; print('issue rows:',len(iss)); [print('%5d  %s'%(v,k)) for k,v in collections.Counter(r['key'].split('/')[0] for r in iss).most_common(6)]"
```

```
issue rows: 188
   74  kipi-system
   48  cole-gtm
   36  job-migration
    6  fleet-health
    4  ktlyst
    2  4-points-consulting
```

122 of 188 resolve to a *project*, not a scanner. For those the producer half of the family key is one bucket and the title template is the only thing standing between two unrelated issues and a collapse. I could not obtain real `kipi-system/*` titles to show an actual bad collapse, so this stays minor — but the docstring and the code disagree, and the guard is weaker on the real board than the test implies.

---

## What is sound

- **Dry-by-default.** Read every path in `main()` and ran the shipped subprocess case: without `--apply` nothing constructs a `LinearWriter` and nothing reaches `graphql`. Exit 0, plan printed. Holds.
- **Comment-before-close ordering.** Drove `apply_family` with an ordering-recording fake across a 5-member family; comment always precedes close for every member. Holds.
- **Record-before-write durability.** `append_record` flushes + `fsync`s before the caller's next mutation. Confirmed by the crash case — after death mid-close, `[["ASK-130","survivor-dor"],["ASK-131","commented"]]` is on disk. This one is genuinely well built and directly answers sp-b5dcf944.
- **The headline family is real, not invented.** Checked the actual producer: `linear-job-migration.py:166` emits `f"Migrate {label} to Linear-tracked execution"` with key `job-migration/{label}`. That normalises to `migrate <x> to <x> execution` and the ledger has 36 rows under `job-migration/`. The 36-issue family the PR is built to collapse exists.
- **Over-collapse near-miss guard.** `migrate com.kipi/nightly-sweep off launchd entirely` does not join the job family; `is_collapsible_template` refuses identifier-only titles. Holds on every shape I threw at it.
- **Resume-closes-a-commented-but-open member.** Works. Correct call, and the right direction.
- **Nothing here is unrecoverable.** Absorbed issues are canceled (never `completed`), keep their pointer comment, and Linear retains description history. Rollback is manual but possible. No credential exposure, no founder-work destruction.

---

## VERDICT: REQUEST CHANGES

Two major findings, neither permanent — so changes, not block.

**Fix first: #1.** `member_targets(survivor)` must read the survivor's description *with the collapse block stripped* (or read only the pre-block region). Right now the second write already drops the survivor's own target, which is precisely the failure mode the PR's own docstring names.

#2 is the same surface and probably the same patch session: the roster has to be the union of the new family and whatever rows the existing block already carries, not a replacement built from open issues only.

---

FINDINGS:
major|member_targets reads the survivor's own description, so the second write ingests the roster block the script wrote and the 20-target cap truncates the survivor's real target out|q-system/.q-system/scripts/linear-collapse.py:203
major|a resumed pass rebuilds the survivor roster from open issues only, deleting the rows and targets of members the previous pass already closed, and the header count understates the family|q-system/.q-system/scripts/linear-collapse.py:432
minor|with no canceled-type state on the team every absorbed issue gets a permanent collapsed comment, nothing is closed, and main() still prints N/N applied and exits 0|q-system/.q-system/scripts/linear-collapse.py:373
minor|producer() is documented as the filing scanner but returns the project name for 122 of 188 real ledger issues, making the different-producer family guard inert on most of the board|q-system/.q-system/scripts/linear-collapse.py:182
END FINDINGS

Repro scripts are in `.pr35rev/` (untracked scratch, same pattern as the existing `.pr3Xrev` dirs). Nothing in the repo was modified.

assafkip:
## Rework pass: all four findings answered (commit `8c1ac24`)

Every finding reproduced first. **10 cases RED before the fix**, including your exact
`["ASK-130", "ASK-133", "ASK-134"]`. Suite 21 -> 33, all green, and `capability-gate.py`
GREEN (`ran=78 quarantined=0`).

Both majors turned out to be one root cause: **this script reads its own output back,
and it was reading it with the wrong semantics.** So the fix names the boundary once —
`strip_block()` reads everything OUTSIDE the markers (what the issue itself says),
`parse_block_rows()` reads everything INSIDE them (what a previous pass recorded) — and
neither function is allowed into the other's territory. That was your "two readers of the
same input with different semantics" hazard sitting in the code already, unnamed.

---

### 1 major — survivor ingests its own roster block. FIXED

`strip_block()` (`linear-collapse.py:196`) removes the block before `member_targets()`
and `producer()` look at a description. `all_targets()` is the uncapped read;
`member_targets()` is that capped.

Reproducers, on your 22-issue real-format family (`job_family_big(22)`, built from
`linear-job-migration.py:166`):

```
FAIL the survivor's own targets survive a read of its own roster block: expected [true] got [false]
FAIL no other member's target lands in the survivor's own row on pass 2: expected [true] got [false]
```

Both green now. The second case asserts the row itself, not just the target list:
`job01.py` present, `com.cole.job02` absent.

**Layer above:** the other consumer of `member_targets` is `print_plan`, which now also
prints uncontaminated targets in the dry run. `producer()` got the same strip — the block
carries no `kipi-key:` today, so that one is prophylactic, not a bug fix.

I did **not** raise `MAX_TARGETS_PER_ISSUE` or make the truncation loud. With the block
stripped, only an issue that genuinely names 20+ labels and paths reaches the cap, and the
roster's own text still claims it enumerates everything. That gap is real but it is not
this finding — captured as **sp-d3fd8070**, named in the `member_targets` docstring.

---

### 2 major — a resumed pass deletes the closed members' rows. FIXED

`family_rows()` (`:389`) merges the prior block's rows with this pass's members. **The
roster only ever grows**: a row is the record of an issue that was closed against this
one, and closing it is precisely why the row has to outlive it. A member seen in this pass
wins over its prior row (title/targets may have been edited); ordered by issue number so a
rewrite is a diff, not a reshuffle.

Reproducer drives your exact scenario — run 1 dies after two closes, run 2 gets only what
`fetch_open()` would return:

```
FAIL a resumed pass keeps the rows of the members the first pass closed:
     expected [["ASK-130","ASK-131","ASK-132","ASK-133","ASK-134"]] got [["ASK-130","ASK-133","ASK-134"]]
FAIL the roster header counts the whole family, not just what is still open:
     expected [## Collapsed family — 5 issues, one change] got [## Collapsed family — 3 issues, one change]
```

**Your note that the fix has to land on the report too — four places, not one:**

| reader | before | now |
|---|---|---|
| block header | counted open members | counts merged roster |
| absorb comment "one of N" | would have said "one of 3" on a permanent object | takes `roster_size` |
| audit JSONL | `family_size` only | `family_size` (this pass) **and** `roster_size` (the whole family) |
| dry-run plan | showed 3 for a family of 5 | prints `kept ASK-131 ... (closed by an earlier pass)` + `roster after this pass: N` |

The merge only works if the block round-trips, so there is a case for that too, with a
pipe in a title (the one character that breaks a markdown table parse).

**Honest note on your "why the suite missed it":** you were right about the a5 case, and
one of my new cases passed even *before* the fix —
`every member's target is still enumerated after the resumed pass`. Finding 1 was masking
finding 2: the survivor had ingested all five rule paths, so they re-appeared in its own
row. It only passes for the right reason now. Kept as a regression guard, flagging it so
it is not read as a clean RED.

**One thing this fix makes quieter, on purpose:** `block_is_current()` (`:341`) skips the
description write when the only difference is the footer timestamp. Without it every
resumed pass rewrites a permanent object to say the same thing with a newer clock. What is
suppressed is the mutation, not the signal — the pass still appends its `survivor-dor`
record with outcome `dor-unchanged`, so the audit file still says the pass ran and what it
decided. The decision lives in `apply_family`, not in `LinearWriter`, so the test double
is not reproducing production logic in order to observe it.

---

### 3 minor — NOT-CLOSED counted as applied. FIXED, and one layer up

You called the trigger constructed and the success-reporting half unconditional. Both
halves are fixed, but the important one is that the **order** was wrong: every absorbed
issue gets its permanent "collapsed into X" comment *immediately before* its close, so by
the time `close_issue` discovers it cannot close, the lie is already posted and this script
cannot retract it.

So `main()` now refuses **before the first write** (`:606`):

```
REFUSED: team ASK has no canceled-type workflow state, so absorbed issues cannot be
closed. Nothing was written. Commenting 'collapsed' on 48 permanent issues that then
stay open is worse than not running.
```

exit 9. `close_issue()` raises as the backstop for any other caller (RED: `expected [true]
got [false]`). And the counting is now driven by what the records SAY happened —
`incomplete_steps()` — not by "the loop did not raise":

```
FAIL a step whose outcome is not a real write makes the family incomplete:
     expected [[["ASK-131","closed","NOT-CLOSED (stubbed)"]]] got []
```

A family counts as applied only when every recorded step reports a real write; otherwise
`PARTIAL <survivor>`, the dead steps on stderr, exit 9.

While proving this I hit a related landmine of my own: `--apply` under
`KIPI_COLLAPSE_FIXTURE` died in `closed_state_id()` with
`'NoneType' object has no attribute 'graphql'`. A traceback is not a refusal — the seam now
says what it is and exits 1, with a case asserting no `Traceback` in stderr.

---

### 4 minor — `producer()` is not the scanner. FIXED (as documentation + fixture), residual captured

You are right and the ledger says so. Rather than argue it, the fixture now uses the
**real** key shapes: `kipi-system/rule-anti-misclassification` for the CAP family (was an
invented `cap-scan/`), `cole-gtm/cap-45-plugin-fleet` for the different-namespace case. The
docstring says what the function actually returns and names the residual.

The residual is now an executable case rather than a paragraph:

```
ok  residual: one kipi-key namespace is one bucket, scanners inside it are not split
```

Two issues, `kipi-system/rule-alpha-one` and `kipi-system/unwired-engine-audit`, same title
template -> one family. If anyone strengthens the family key, that case fails and they
update it deliberately. Captured as **sp-ab2d1067** with the three candidate strengthenings
(filer marker, project, key second segment).

I did not strengthen it in this pass: it would change which issues collapse on the real
board, and that decision belongs with the dry run you and I both still owe this thing.

---

### Still unverified, same as your read

I could not run against the live ASK board either — this worktree has no network and no
Linear key. **The over-collapse question on the real 200-issue board is still open**, and
nothing in this pass closes it. The ledger has no titles in it, so I could not even check
offline whether two unrelated `kipi-system/*` issues share a title template. That check
happens on the first real dry run, which is exactly why the engine is still
`declared_inert` (sp-ffc69720) rather than wired.

### What I did not regress

Re-ran everything you listed as sound: dry-by-default through the shipped subprocess path,
comment-before-close ordering across the 5-member family, record-before-write durability
(`[["ASK-130","survivor-dor"],["ASK-131","commented"]]` still on disk after the simulated
death), the near-miss guard, and resume-closes-a-commented-but-open-member. All still
green, none of them touched.

```
$ bash q-system/.q-system/scripts/test/test-linear-collapse.sh
passed 33, failed 0

tests: ran=78 quarantined=0 skipped-skeleton-only=0
capability-gate: GREEN
```

assafkip:
Review complete. I ran the shipped suite (33/33 green), then attacked it — and I could do the one thing both earlier rounds explicitly said they could not: **run the detector against the live ASK board.**

---

## Review: PR #35 — `linear-collapse.py` (ASK-226), round 2

Both prior rounds closed with the same caveat: *"The over-collapse question on the real 200-issue board is still open."* This session has network + the Linear key, so I closed it. The answer is not over-collapse. It is **zero collapse**.

---

### 1. major — on the live board the detector finds ZERO families, including the 12 CAP-0n issues the PR names as the win

`q-system/.q-system/scripts/linear-collapse.py:174` (`title_template`), fixture at `q-system/.q-system/scripts/test/test-linear-collapse.sh:59`

The shipped script, dry, against the real ASK board:

```
$ python3 q-system/.q-system/scripts/linear-collapse.py --project all
129 open issue(s) scanned; no duplicate-template family found.

$ python3 q-system/.q-system/scripts/linear-collapse.py          # default --project kipi-system
55 open issue(s) scanned; no duplicate-template family found.
```

The CAP-0n family (ASK-130..141) is **still open right now** — it is the family the PR body and the triage log both name ("First of five identical CAP-0n rule issues"). Real titles vs the fixture's:

```
$ python3 .pr35rev/r2/repro_cap_family.py
REAL CAP-0n issues fetched from the board: 12

ASK-130  ns=kipi-system  tmpl=<x> rule <x> <x> guardrails for content generation
ASK-131  ns=kipi-system  tmpl=<x> rule <x> audhd executive function and <x> interaction rules
ASK-132  ns=kipi-system  tmpl=<x> rule <x> <x> coding rules - structure communication emotional scaffolding
ASK-133  ns=kipi-system  tmpl=<x> rule <x> code style and naming conventions for python js shell and json
ASK-134  ns=kipi-system  tmpl=<x> rule <x> <x> design skills only for <x> pages and assets
ASK-135  ns=kipi-system  tmpl=<x> rule <x> <x> development skills when building skills plugins mcp servers hooks
ASK-136  ns=kipi-system  tmpl=<x> rule <x> a lightweight <x> planning <x> the moment the founder has an idea a bug a pasted
ASK-137  ns=kipi-system  tmpl=<x> rule <x> <x> analysis is handled by the rca skill in the <x> plugin
ASK-138  ns=kipi-system  tmpl=<x> rule <x> fires when the founder shares someone else's content and asks to comment reply qu
ASK-139  ns=kipi-system  tmpl=<x> rule <x> token consumption guardrails and <x> rules
ASK-140  ns=kipi-system  tmpl=<x> rule <x> founder voice enforcement for <x> written content
ASK-141  ns=ktlyst       tmpl=<x> rule <x> every implementation diff goes through codex review before commit this

families detected on the REAL 11: 0
distinct (namespace, template) keys: 12

--- what test-linear-collapse.sh's cap_family() fixture emits ---
ASK-130  tmpl=<x> rule <x> <x> guardrails
ASK-131  tmpl=<x> rule <x> <x> guardrails
ASK-132  tmpl=<x> rule <x> <x> guardrails
ASK-133  tmpl=<x> rule <x> <x> guardrails
ASK-134  tmpl=<x> rule <x> <x> guardrails
```

The real producer puts the rule's **description** after the colon — free-form prose, different per rule. The fixture at `test-linear-collapse.sh:59` builds `"CAP-%02d rule %s: %s guardrails"` — a tail that is the same literal word for every member. 12 distinct templates on the board, 1 identical template in the fixture. Byte-exact equality never matches, so the suite is green on a shape no producer emits.

This is **not** round 1's finding 4. That one was `producer()` (`:223`) resolving to a repo instead of a scanner, and it was answered with `sp-ab2d1067` + a pinned residual case. I am not falsifying that citation — `producer()` correctly returns `kipi-system` for all 11. The template half is what fails, in a different function, with evidence neither round could obtain.

The other named family is already gone: 36 `job-migration/` keys in the ledger, **1 still open** (ASK-151) — the other 31 were closed by an earlier one-off run (`q-system/output/collapse-jobmigration-2026-07-28.jsonl`, survivor ASK-151, family_size 32). So round 1's "the 36-issue family the PR is built to collapse exists" is true of the ledger and false of the current board.

**3am answer:** nothing pages, because the engine is `declared_inert` and dry-by-default — this is why it is major, not blocker. Nothing writes, nothing is permanent, rollback is `git revert`. What merges is a green 33-case suite, a `capability-manifest.json` entry asserting a built-and-tested engine, and a founder who runs the dry run the PR asks for and gets a clean `no duplicate-template family found` with no hint the matcher is mis-tuned against its own target data.

---

### 2. minor — `--min-family N>2` plus a mid-run death strands a member with a permanent "collapsed" comment, still open, never revisited

`q-system/.q-system/scripts/linear-collapse.py:694` (the flag) + `:273` (`detect_families` counts only OPEN members)

`--min-family` is a documented flag in the usage block. Resume re-derives the family from `fetch_open()`, so every close shrinks it — below the threshold, the family vanishes and the resume path the docstring promises ("*A member commented by a dead run is resumed, not skipped, or it sits open forever behind a pointer saying it was collapsed*") never runs.

```
$ python3 .pr35rev/r2/repro_minfamily.py
pass 1: families at --min-family 5 = 1
pass 1 died: simulated mid-run death right after the comment

state after the crash:
  ASK-130  backlog   absorbed-comment=False
  ASK-131  canceled  absorbed-comment=True
  ASK-132  canceled  absorbed-comment=True
  ASK-133  backlog   absorbed-comment=True
  ASK-134  backlog   absorbed-comment=False

pass 2 sees 3 open issue(s): ['ASK-130', 'ASK-133', 'ASK-134']
pass 2: families at --min-family 5 = 0

STRANDED (permanent 'Collapsed into ASK-130' comment, still open,
and pass 2 will never touch them again): ['ASK-133']
audit rows: [('ASK-130', 'survivor-dor', 'dor-written'), ('ASK-131', 'commented', 'comment-added'), ('ASK-131', 'closed', 'closed'), ('ASK-132', 'commented', 'comment-added'), ('ASK-132', 'closed', 'closed')]
```

Note the last line: ASK-133's comment landed on a permanent object and **no audit row exists for it** — `append_record` runs after the mutation returns, so the in-flight write is unreceipted. Minor, not major: it needs a non-default flag plus a crash, the stranded issue stays visibly open, and reopening/closing by hand is the whole cleanup.

---

- **Round 1's two majors are genuinely dead.** `strip_block()`/`family_rows()` hold: the survivor's own targets survive reading its own block, and a resumed pass keeps the closed members' rows and header count. RED-then-green cases exist for both, and I re-ran them.
- **The foreign-marker resurrection of round 1 finding 1 — I tried and could not fire it.** ASK-151 is live and carries a roster under a *different* marker (`<!-- kipi-collapse: job-migration -->`); `strip_block()` does not strip it, and `member_targets(ASK-151)` returns 20 entries of which 19 belong to other jobs, 13 silently truncated. But that earlier run also wiped ASK-151's `kipi-key`, so `producer(ASK-151) == ""` and it cannot group with any `job-migration/` issue the live producer files. **Not reachable. Dropped.** (`.pr35rev/r2/repro_foreign_block.py`)
- **Comment pagination → duplicate permanent comments.** `ISSUES_Q` requests `comments{nodes}` unpaginated (Linear defaults to 50). Max on any open issue today: **29**. No producer gets near it. Dropped per the constructed-input rule.
- **Over-collapse on the real board.** Zero groups of size ≥2 across all 129 open issues. The risk this design is built around does not materialise — the matcher errs entirely in the safe direction, just past the point of usefulness.
- **Dry-by-default, comment-before-close ordering, record-before-next-write, the `NOT-CLOSED`→raise + preflight refusal, `incomplete_steps` driving the count.** All hold. The preflight-before-first-write fix for finding 3 is the right layer — refusing before the lie is posted, not after.
- **Nothing is unrecoverable.** No writes on merge, `canceled` never `completed`, pointer comments preserved, `git revert` clean.

---

## VERDICT: REQUEST CHANGES

One major, one minor. Nothing permanent, so changes rather than block.

**Fix first: #1.** Before touching `title_template`, run the dry pass on the real board and read the 12 CAP templates above — the tail after the colon is a rule *description*, and byte-exact template equality can never see through it. That is a design decision (truncate the title at a delimiter? key on the `rule-` prefix of the kipi-key second segment, which sp-ab2d1067 already lists as a candidate?), not a regex tweak, and the fixture has to be rebuilt from real titles either way. A suite whose fixtures come from the real board would have gone RED on day one.

**Process note, offered not as a nit:** rounds 1 and 2 both ended with "the real-board check is still open," and the work shipped anyway. Every finding in both rounds was about what the code does to invented inputs. The single check that would have caught the biggest problem was the one deferred each time. Worth making a live dry-run receipt a DoR item for anything that ships a detector.

---

Repro scripts are in `.pr35rev/r2/` (untracked scratch, same pattern as the existing `.pr3Xrev` dirs). Nothing in the repo was modified; no writes reached Linear.

FINDINGS:
major|title_template requires byte-identical templates but the real CAP-0n producer emits a free-form prose tail per issue, so the detector finds 0 families on the live 129-issue board including the 12 CAP issues the PR names as its win|q-system/.q-system/scripts/linear-collapse.py:174
minor|--min-family N greater than 2 plus a mid-run death strands a commented-but-open member: pass 2's family drops below the threshold, the member is never revisited, and its commented row never reached the audit file|q-system/.q-system/scripts/linear-collapse.py:694
END FINDINGS

Want me to post this as a PR comment on #35?

assafkip:
## Rework pass 2: both findings answered, and this time with the live-board receipt (`87a801b`, `14255a2`)

You were right about the thing that mattered most, and you were right about *why* it
survived two rounds. I ran the live dry run before touching any code, and reproduced your
result exactly:

```
$ python3 q-system/.q-system/scripts/linear-collapse.py --project all
129 open issue(s) scanned; no duplicate-template family found.
```

Suite **33 -> 47**, all green. The same command now:

```
$ python3 q-system/.q-system/scripts/linear-collapse.py --project all
129 open issue(s) scanned; 2 family/families found covering 14 issue(s).
FAMILY  ASK-130    survivor, 10 absorbed   template: <x> rule <x>
        kipi-key namespace: kipi-system
FAMILY  ASK-141    survivor, 2 absorbed   template: <x> rule <x>
        kipi-key namespace: ktlyst
```

---

### 1 major — zero families on the live board. FIXED

`title_head()` (`linear-collapse.py:174`) splits the title at the first `": "` and
`title_template()` keys on the head only. Your read of the producer was the correct one:
the tail is the rule's own summary line, different prose in every member.

The head is thin though — `<x> rule <x>`, one literal word — so it falls under
`MIN_LITERAL_WORDS`. I did not lower the floor, because that is the guard you confirmed
holds. `is_collapsible_template()` (`:246`) gained a second, narrower path instead:

> a head below the floor is admissible only when the **kipi-key's kind segment repeats one
> of the head's literal words**. `<x> rule <x>` + `kipi-system/rule-anti-misclassification`
> qualifies. Two independent fields of the issue, one of them machine-written, naming the
> same kind of thing.

That substitution is structurally unavailable to anything a person filed: no kipi-key, no
`kind`, floor only. So the over-collapse guard is byte-for-byte unchanged for human issues.

**Your "the fixture has to be rebuilt from real titles either way" — done.** `CAP_REAL`
(`test-linear-collapse.sh:46`) is ASK-130..134 verbatim, prose tails and all, and `CAP_DESC`
is the real table body. Running the **new** suite against the **pre-fix** script:

```
FAIL the real CAP-0n titles group into one family: expected [1] got [0]
FAIL the 5 CAP issues group into exactly one family: expected [1] got [0]
FAIL the survivor is the lowest-numbered member: expected ["ASK-130"] got []
FAIL every other member is absorbed, none dropped: expected [["ASK-131", ...]] got []
```

Three of those are cases that were green yesterday. The fixture was the only thing holding
them up, which is your point stated as a test result.

**The over-collapse canary.** Head-truncation is exactly the change that could merge 45
distinct capabilities, so the 45-issue `cole-gtm` CAP registry (`CAP-01 Persona and voice
routing`, `CAP-02 Rule registry`, ...) is now a fixture with two cases: it must produce 0
families alone, and 1 family with zero leakage when mixed with the rule family. On the real
129 the only groups of size >= 2 are the two above — I enumerated every group, not just the
ones I hoped for.

---

### The layer above this fix, which is where it nearly went wrong

Making the family collapse is only half the job, and the other half would have shipped
broken. `PATH_RE` requires backticks; the CAP scanner writes its entry point into an
un-backticked table cell:

```
| Entry point | .claude/rules/anti-misclassification.md |
```

Measured on the 14 real family members: **11 of 14 resolved to zero targets.** So the fix
that made the family collapsible would have written a roster of 11 rows saying
`_(no explicit target)_` — the exact silent drop DoR check 3 and this script's whole
docstring exist to prevent, and worse than not collapsing at all.

`BARE_PATH_RE` (`:186`) matches an un-backticked path that contains a directory separator
(so a bare word in prose cannot match), plus a kipi-key-item fallback so a machine-filed
issue can never produce a nameless row. Members with no target: **11/14 -> 0/14**. The live
plan reads:

```
        SURVIVOR ASK-130    .claude/rules/anti-misclassification.md
        absorb   ASK-131    .claude/rules/audhd-interaction.md
        ...
        absorb   ASK-136    plan.md, .claude/rules/quick-plan.md
        absorb   ASK-140    .claude/rules/voice-enforcement.md
```

The fallback is more than the literal finding asked for. Flagging it rather than burying it:
without it the collapse can still emit a nameless row, and that row is the failure mode the
blast-radius line calls out by name.

---

### 2 minor — `--min-family` + a mid-run death strands a member. FIXED

Same root cause as round 1's finding 2, one function further out: the threshold was measured
against what `fetch_open()` can see, so every close shrank the family. `detect_families()`
(`:296`) now splits the question in two:

- `len(members) < 2` -> nothing to do this pass. Skip.
- `len(family_rows(...)) < max(2, min_family)` -> family too small to bother with. Skip.

The roster is the family; the open members are this pass's work. Your scenario, as a case:

```
FAIL a resumed pass below --min-family still sees the family: expected [1] got [0]
FAIL the resumed pass closes the member stranded behind its own comment: expected [true] got [false]
```

Both green. Plus a case that a fully-absorbed family does **not** reappear as a no-op family
in every subsequent dry run, because roster-aware counting could easily have caused that.

**Your note that the in-flight write is unreceipted** — fixed too, and you were right that it
undercut a claim the code already makes. `append_record`'s docstring says the audit file can
say what happened to every permanent object the run touched; that was true of all of them
except the one being written when the process died. `apply_family` now writes an
`outcome: "attempting"` row before each mutation:

```
FAIL a write killed in flight still left an intent row on disk: expected [true] got [false]
```

The intent row goes to disk only, never into the returned records, so `incomplete_steps()`
cannot mistake a completed step for a dead one — asserted directly. And your durability case
keeps its original expected value (`[[ASK-130,survivor-dor],[ASK-131,commented]]`) by
filtering intents, rather than being relaxed to accommodate the new rows.

---

### One thing I changed that you did not ask for, deliberately

The `sp-ab2d1067` residual case flipped from `1` to `0` and I rewrote it rather than let it
fail quietly. It pinned that two scanners in one namespace share a bucket:
`kipi-system/rule-alpha-one` and `kipi-system/unwired-engine-audit` with the same title
template merged into one family. Under corroboration only `rule-` matches, so they no longer
merge.

**This is the fix making something quieter, so: that pair now under-collapses where it used
to collapse.** Safe direction, and it partially closes the residual — but only partially, and
the case now says so in its own text. Two scanners whose kinds *do* match, or any family
whose head clears the floor on its own, still merge on namespace + template alone.
`sp-ab2d1067` stays open.

---

### Your process note — captured, not just agreed with

> Worth making a live dry-run receipt a DoR item for anything that ships a detector.

`sp-6bfdf349`. Two rounds closed on "the real-board check is still open" and the work shipped
anyway; the deferred check is the one that found the matcher mis-tuned against its own target
data. The ledger item covers both halves: a live dry-run receipt as a required DoR check for
any detector, and board-captured fixtures instead of invented ones.

### What I did not regress

Re-ran everything you listed as sound: dry-by-default through the shipped subprocess path,
comment-before-close ordering, record-before-next-write durability, the near-miss guard,
`strip_block`/`family_rows` (round 1's two majors), the `NOT-CLOSED` raise + preflight
refusal, and `incomplete_steps` driving the count. All green, none weakened.

```
$ bash q-system/.q-system/scripts/test/test-linear-collapse.sh
passed 47, failed 0

  tests: ran=78 quarantined=0 skipped-skeleton-only=0
capability-gate: GREEN
```

### Still true

No writes reached Linear in this pass. Every run above was dry. The engine remains
`declared_inert` (`sp-ffc69720`): `--apply` would now close 12 permanent objects, and that
first real write is the founder's call after reading the roster above, not mine.

assafkip:
Read the code and ran my own attacks before opening the prior rounds. Three attacks reproduced; two I dropped.

## Review: PR #35 — `linear-collapse.py` (ASK-226), round 3

Baseline: shipped suite **47/47 green** on the branch head (`14255a2`), run from a scratch copy so the repo was never touched.

---

### 1. major — the tool's own escape hatch is the one thing it undoes: a human's reopen is silently re-closed on the next `--apply`

`q-system/.q-system/scripts/linear-collapse.py:693` + `:704`

Every absorbed issue gets this permanent comment (`:524`):

> *"If the family fix does not cover this target, reopen this issue — that is what the pointer is for."*

The resume branch is `already_absorbed(m) and not is_closed(m)` → close. That predicate is byte-identical for (a) a member a crashed run never closed and (b) a member a human deliberately reopened on the tool's own instruction. Case (b) gets closed again, with **no new comment**, an audit row indistinguishable from a clean resume (`commented: already-present`, `closed: closed`), and `main()` printing `applied ASK-130 (+N absorbed)`.

```
$ python3 .pr35rev/r3/attack_reopen.py .pr35rev/r3/scripts/linear-collapse.py .pr35rev/r3/run
```

Real output:

```
pass 1 incomplete_steps: []
pass 1 -> ASK-131 state=canceled absorbed=True
human reopened ASK-131 -> state=unstarted
fetch_open() now returns: ['ASK-130', 'ASK-131']
pass 2 writer log: [['close', 'ASK-131']]
pass 2 incomplete_steps: []
pass 2 -> ASK-131 state=canceled
new comment posted explaining the re-close? False

VERDICT reopened-issue-silently-reclosed: true
```

**3am answer:** nothing pages — the engine is `declared_inert`, so this only fires when the founder re-runs `--apply`, which the docstring calls "safe and resumes". Cleanup is: notice the re-close among N absorbed, reopen again, and never re-run `--apply` after a reopen. Recoverable, so major not blocker. The PR body's claim *"reopen is always available, and the pointer is what makes reopening meaningful"* is what the code contradicts.

Not raised in round 1 or 2. Round 1 listed this exact path under **What is sound** ("Resume-closes-a-commented-but-open member. Works."). I am not falsifying that — resume *is* correct for the crash case. The gap is that nothing distinguishes the two states.

---

### 2. minor — the collapse that already ran on this board is invisible to the new engine, so the same family gets a second survivor

`q-system/.q-system/scripts/linear-collapse.py:102` (`BLOCK_OPEN`) and `:108` (`ABSORB_MARKER`)

`linear-collapse-jobmigration.py` (merged, `16c5fc3`) ran on 2026-07-28 and, per its own audit, absorbed 31 issues into ASK-151 and **overwrote** ASK-151's whole description (`{"input": {"description": survivor_body}}` — no splice). That roster carries neither of the new engine's markers, and the overwrite deleted ASK-151's `kipi-key` line.

Survivor body below is not invented — it is the merged predecessor's own `body_lines` code run over the exact `jobs` list its audit file recorded.

```
$ python3 .pr35rev/r3/attack_predecessor.py .pr35rev/r3/scripts/linear-collapse.py
```

```
predecessor roster enumerates 32 job labels
strip_block() removed the predecessor roster? False
parse_block_rows() recovered rows?            []
producer(ASK-151) (kipi-key namespace)       -> ''
member_targets(ASK-151) -> 20 entries, capped at MAX_TARGETS_PER_ISSUE=20
  labels belonging to OTHER (already-closed) members: 19/20

=== detect_families() on that board ===
  family: survivor=ASK-401 producer='job-migration' members=['ASK-401', 'ASK-402']
ASK-151 (the live survivor holding all 32 targets) is in a family? False

VERDICT live-survivor-split-off-family: true
```

Round 2 attacked the adjacent surface (roster ingestion via a foreign marker) and correctly **dropped** it as unreachable. I am not re-raising that. The consequence I reproduced is different and round 2 did not test it: once `linear-job-migration.py` files any new launchd job, the engine builds a **second** survivor for a family that already has one, and permanently comments the new members pointing at the wrong issue.

Minor, honestly: ASK-401/402 are constructed. The producer can emit them (it dedupes on ledger key, so a new job files a new issue) but none exist today. Harm is an extra board object plus a pointer comment, both cleanable.

---

### 3. minor — the default `--project` drops most of the board, the Usage block never mentions it, and the empty-result message does not name the scope it applied

`q-system/.q-system/scripts/linear-collapse.py:811`, Usage block at `:96-100`

The author's own live receipt shows the size of this: `--project all` → **129 scanned**; default → **55 scanned**. 74 issues silently out of scope.

```
$ python3 .pr35rev/r3/attack_scope.py .pr35rev/r3/scripts/linear-collapse.py \
      q-system/.q-system/scripts/linear-triage.py
```

```
board has 5 open CAP issues, all in Linear project 'cole-gtm'
fetch_open(project=None)          -> 5 issue(s)
fetch_open(project='kipi-system') -> 0 issue(s)   <- argparse DEFAULT

what main() prints on that empty result (linear-collapse.py:646):
   print("no open issues to scan")
what the sibling linear-triage.py prints for the same situation:
   print(f"no open issues in project {args.project!r}")

the module Usage block:
  linear-collapse.py                          # dry, prints the plan
  linear-collapse.py --apply                  # writes: survivor DoR, comments, closes
  linear-collapse.py --min-family 5           # only collapse families this big
  linear-collapse.py --out path.jsonl         # override the audit file location
mentions --project? False
```

A filtered-to-zero scan reads identically to an empty board. The sibling script already got this right. This repo's own lesson file says it: *"A zero result must prove it is empty, not broken."*

---

### 4. nit — the shipped manifest entry misstates the case count

`q-system/.q-system/capability-manifest.json:361` — `"built and tested (33 cases)"`. The suite is **47**. The PR body says 21. Three numbers, one suite.

---

- **Fixture fidelity against the real CAP producer.** This was round 2's kill shot, so I checked it directly rather than trusting the rework note. `linear-sync.py:150-176` emits `f"{cap_id} {cap['name']}: {summary}"` and writes `| Entry point | .claude/rules/x.md |` **un-backticked** — exactly what `CAP_DESC` and `CAP_REAL` claim. I also checked the no-summary path (title has no `": "` at all): `title_head` returns the whole title and the template is still `<x> rule <x>`. Fixture matches producer. Holds.
- **Roster round-trip under adversarial titles.** A title already containing `\|` double-escapes on write and un-escapes correctly on read; a title ending in a backslash still splits on the right pipe. `_cells` rejects malformed rows instead of mis-parsing them. Could not break the merge.
- **`attempt()` closure binding.** Both lambdas bind `m=member`/`b=body` as defaults — no late-binding bug in the loop.
- **The intent row cannot poison the count.** `attempt` writes `outcome:"attempting"` to disk only and never into `records`, so `incomplete_steps` cannot see it. Asserted in the suite and I re-derived it from the code.
- **Round 1's and round 2's majors are all genuinely dead.** `strip_block`, `family_rows`, the preflight refusal, `incomplete_steps` driving `done`, and the head-truncation matcher. I re-ran every one; none re-fires.
- **Nothing permanent lands on merge.** No writes, `canceled` never `completed`, pointer comments preserved, `git revert` clean, no credential surface.

Dropped after failing to reproduce: the roster-block ingestion resurrection via ASK-151's foreign marker (ASK-151's own label sorts first, so the cap truncates 12 *other* labels, not its own — my "eviction" claim was wrong and I am dropping it), and comment-pagination-driven duplicate comments (`ISSUES_Q` is unpaginated at Linear's default 50; no issue is near it).

---

## Process note — this is the calibration finding, and it is worth more than #2 and #3

Three rounds, and each one found a fresh major on code the previous round had explicitly cleared. Round 1 cleared the matcher; round 2 killed it. Round 1 listed the resume path under *What is sound*; #1 above reopens it. That is not severity inflation — the majors are getting smaller each round (round 2's meant the tool found zero families at all; mine means one reverted click) — it is **surface sampling**. Each round attacked whichever surface the previous round's findings pointed at, and the untouched surface produced the next major.

Round 2 already named half of this and the author captured it as `sp-6bfdf349` (live dry-run receipt as a DoR item for any detector). The other half is missing from that item: **the operator-action surface**. Every round attacked what the script does to issues. None attacked what happens when a human acts on the board between two runs, which is where #1 lives. Worth adding to `sp-6bfdf349` rather than filing new.

One thing I could not check: the author's post-fix receipt only shows `--project all`. The **default** invocation's family count on the live board is still unreceipted, and finding #3 is why that matters. No board access here, so no repro — flagging it as unverified, not as a finding.

---

## VERDICT: REQUEST CHANGES

One major, two minors, one nit. Nothing permanent, nothing unrecoverable, so changes — not block.

**Fix first: #1.** The cheapest correct fix is to make the two states distinguishable rather than to stop resuming: record the close in the absorb comment (or check whether the issue's state changed *after* the marker comment's `createdAt`, which `ISSUES_Q` already fetches on comments) and skip any member whose reopen post-dates its own pointer. A member reopened after being collapsed is a human decision; treat it as terminal and print it in the plan instead of closing it.

---

FINDINGS:
major|a human who reopens an absorbed issue on the tool's own instruction has it silently re-closed by the next --apply, with no new comment and an audit row identical to a clean resume|q-system/.q-system/scripts/linear-collapse.py:693
minor|the merged predecessor linear-collapse-jobmigration.py writes different markers and wiped ASK-151's kipi-key, so the live survivor is invisible to this engine and a second survivor is built for the same family once any new job-migration issue is filed|q-system/.q-system/scripts/linear-collapse.py:102
minor|the default --project kipi-system silently narrows the scan from 129 to 55 issues, is absent from the module Usage block, and the empty-result message does not name the scope so a filtered-to-zero run reads as an empty board|q-system/.q-system/scripts/linear-collapse.py:811
nit|the shipped declared_inert manifest entry says the engine is tested with 33 cases; the suite has 47 and the PR body says 21|q-system/.q-system/capability-manifest.json:361
END FINDINGS

Repro scripts are in `.pr35rev/r3/` (untracked scratch, same pattern as the existing `.pr3Xrev` dirs). Nothing in the repo was modified; no writes reached Linear.

assafkip:
Closing as a redrive, not as a rejection. ASK-226 stays In Progress and the branch is preserved.

State of this PR: CONFLICTING with main, 71 commits behind, opened 2026-07-28.

The work itself is still wanted. `linear-collapse.py` is not on main, and the problem it solves is live: the board is dominated by auto-filed scanner issues, and collapsing a duplicate-template family into one issue whose DoR names every member is the right shape for that.

But this branch cannot merge and cannot be brought up to date without a conflict resolution, and I will not resolve conflicts blind on someone else's branch. Sitting open, it is a permanently red required check that mails a CI failure on every push.

Redrive off current main under ASK-226. Nothing is lost: branch `sana/ask-226` holds the full implementation of `linear-collapse.py` and its test, and ASK-226 remains the tracker.

PR #36, opened 2026-07-28
feat(ops): dispatch picks file-disjoint issues so it can run in parallel safely, plus an on-demand burst that ignores the daily cap (ASK-225)

assafkip:
Resolves [ASK-225](https://linear.app/ask-consulting/issue/ASK-225), spillover `sp-e45251f7`.

## Read this first: the branch carries 5 commits that are not mine

`kipi-dispatch.sh` — the file the DoR names — **does not exist on `origin/main`**. It lives only on the founder's local `main`, which is 20 commits ahead of origin and unpushed. There was nothing on my base to build on.

So the first five commits here are cherry-picked from local `main` (`6f1a99f 308d62b 6a9f73e 61c0101 a0bba9c`, the ASK-113 dispatch heartbeat). They touch exactly two files, `kipi-dispatch.sh` and `com.kipi.dispatch.plist`, and are byte-identical to local main:

```
$ git diff main a9f9788~1 -- kipi-dispatch.sh q-system/.q-system/scripts/com.kipi.dispatch.plist
(empty)
```

Nothing on the DoR's "Not doing" list is touched by them. **Only the last commit is ASK-225 work.** Review `a9f9788`.

## What changed

Dispatch picked by READINESS ALONE, so it could not see which files a candidate touches. Two concurrent runs could land in the same file — observed 2026-07-28, ASK-223 edits the same `linear-worker.sh` region as the then-live ASK-222. The cap sat at 1 as a stopgap.

Every ready issue carries a `**Files:**` list in its DoR, so dispatch is now a set-intersection problem:

- the union of every LIVE run's file set is built from **the issue each run is working**, read from the process table's `--issue`, never from the worktree (a worktree can be stale or left behind by a killed run)
- a candidate goes only if its set is disjoint from that union
- **no `**Files:**` line, or an unreadable set, means never dispatched in parallel.** Fail closed — an unknown set intersects everything by assumption
- `capability-manifest.json` is exempt (magnet file, `sp-f3a2ad81`), stated in a comment because a silent exemption is how the conflict comes back

`--burst N [--parallel P]` dispatches on demand. It **neither reads nor spends the daily counter**: that cap exists to stop the unattended heartbeat spending the subscription overnight, not to limit an explicit founder request made while they are present. It prints the picks, every skip with its reason, and the session estimate *before* launching.

## Three defects the build turned up, all fixed here

[...]

PR #37, opened 2026-07-28
fix(ci): main was RED, blocking every open PR on two defects of mine (ASK-113)

assafkip:
`validate` is required, so a red main blocks 100% of PRs. #23, #34, #35 and #36 all failed the same two checks, none of which they caused.

**1. `com.kipi.dispatch.plist` hardcoded `/Users/assafkipnis`** in five places, while the other three templates use `__KIPI_REPO__`/`__HOME__` placeholders that `install-plist.sh` substitutes. On any other machine the job installs and then execs a path that does not exist — launchd reports a failed job, not a bad template.

**2. `linear-collapse-jobmigration.py` was neither wired nor declared.** It is a genuine one-shot (it collapsed the 32-member job-migration family into ASK-151 and has no second family). Declared inert with `sp-fb332466`, which is the plan to delete it when ASK-226's general collapse tool lands.

### Checks (real output, same commands CI runs)

| check | before | after |
|---|---|---|
| `capability-gate.py` | RED (2) | **GREEN**, ran=77, exit 0 |
| `test-install-plist.sh` | FAIL hardcoded home | **PASS** incl. both negative self-tests |
| `plutil -lint` | — | OK |

Appended to `capability-manifest.json`, never re-sorted — it is the magnet file (`sp-f3a2ad81`).

The installed job at `~/Library/LaunchAgents` is untouched and still running; this is the template it renders from.

### DIFF
--- a/q-system/.q-system/capability-manifest.json
+++ b/q-system/.q-system/capability-manifest.json
@@ -396,6 +396,11 @@
       "path": "q-system/.q-system/scripts/verify-containment-export.py",
       "reason": "built for prd-skeleton-data-containment's export/quarantine flow and never wired",
       "spillover_id": "sp-0f773063"
+    },
+    {
+      "path": "q-system/.q-system/scripts/linear-collapse-jobmigration.py",
+      "reason": "one-shot family collapse: it absorbed the 32-member job-migration family into ASK-151 on 2026-07-28 and has no second family to run on. ASK-226 generalises it; retire this one when that lands",
+      "spillover_id": "sp-fb332466"
     }
   ],
   "uncovered_known": [
--- a/q-system/.q-system/scripts/com.kipi.dispatch.plist
+++ b/q-system/.q-system/scripts/com.kipi.dispatch.plist
@@ -8,12 +8,12 @@
 	<key>EnvironmentVariables</key>
 	<dict>
 		<key>KIPI_REPO</key>
-		<string>/Users/assafkipnis/projects/kipi-system</string>
+		<string>__KIPI_REPO__</string>
 		<!-- Same PATH as com.kipi.openloops-heartbeat, which is the proven
 		     working `claude -p` under launchd. gh and claude both live in
 		     /opt/homebrew/bin; without it the dispatcher pages and exits. -->
 		<key>PATH</key>
-		<string>/Users/assafkipnis/.local/bin:/opt/homebrew/bin:/usr/bin:/bin:/usr/sbin:/sbin</string>
+		<string>__HOME__/.local/bin:/opt/homebrew/bin:/usr/bin:/bin:/usr/sbin:/sbin</string>
 		<!-- ONE at a time, deliberately. Each run spawns a `claude -p` agent
 		     plus a reviewer, so this is also the spend ceiling -- but the real
 		     reason is conflicts. The dispatcher picks by readiness and has NO
@@ -53,7 +53,7 @@
 	<key>ProgramArguments</key>
 	<array>
 		<string>/bin/bash</string>
-		<string>/Users/assafkipnis/projects/kipi-system/kipi-dispatch.sh</string>
+		<string>__KIPI_REPO__/kipi-dispatch.sh</string>
 	</array>
 
 	<!-- Every 15 minutes. The heartbeat only TOPS UP to the concurrency cap;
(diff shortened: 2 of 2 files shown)


assafkip:
Review complete. I ran everything below myself; no prior-round comments exist on this PR to inherit.

## Findings

**MINOR — two operator pages now name a file that no longer contains the value they tell you to fix**
`kipi-dispatch.sh:50` and `kipi-dispatch.sh:153`

Both pages send the operator to `com.kipi.dispatch.plist`. Before this PR that file held the literal repo path and the literal PATH, so the instruction was self-consistent. After this PR the repo copy holds `__KIPI_REPO__` and `__HOME__`, and editing it does nothing until `install-plist.sh` re-renders and re-bootstraps. The file that actually drives the running job is `~/Library/LaunchAgents/com.kipi.dispatch.plist`.

Reproducer and real output:

```
$ sed -n '48,51p;150,154p' kipi-dispatch.sh
cd "$REPO" 2>/dev/null || {
  say "FATAL: repo not found at $REPO"
  page "kipi dispatch: repo not found at $REPO -- the Linear loop is DEAD. Do: check the path in com.kipi.dispatch.plist."
  exit 1
if ! command -v gh >/dev/null 2>&1; then
  say "FATAL: gh not on PATH ($PATH)"
  page "kipi dispatch: gh CLI not on PATH under launchd, so no PR can be opened. The Linear loop is stalled. Do: fix PATH in com.kipi.dispatch.plist."
  exit 1

$ gh api "repos/assafkip/kipi-system/contents/q-system/.q-system/scripts/com.kipi.dispatch.plist?ref=ff26b29" --jq .content | base64 -d | grep -n "__KIPI_REPO__\|__HOME__"
11:		<string>__KIPI_REPO__</string>
16:		<string>__HOME__/.local/bin:/opt/homebrew/bin:/usr/bin:/bin:/usr/sbin:/sbin</string>
56:		<string>__KIPI_REPO__/kipi-dispatch.sh</string>
71:		<string>__HOME__/.config/kipi/dispatch.out</string>
73:		<string>__HOME__/.config/kipi/dispatch.err</string>
```

At 3am the loop is dead, the page says open a file, the operator opens it and sees a placeholder. Bounded and recoverable, so minor. The `KIPI_DISPATCH_DAILY_MAX` page at line 143 is unaffected: the template is the right file to edit for that, and it already prints the reload commands.

## What is sound

- **Verbatim-copy path.** The blocker shape here would be some code path that copies the template straight into `~/Library/LaunchAgents`, which after this PR would install a job exec'ing `__KIPI_REPO__/kipi-dispatch.sh`. Grepped every `.sh`/`.py` in the repo for `LaunchAgents`: the only writers are `install-plist.sh:80-83` (renders, then asserts) and `install-lessons-daily.sh` for its own unrelated label. No verbatim path exists.
- **The declared-inert claim.** `linear-collapse-jobmigration.py` has exactly one reference in the entire repo, inside its own usage string at line 36. It is genuinely unwired, so the manifest entry is not suppressing a live script.
- **The spillover receipt is real, not a dangling id.** `sp-fb332466` exists in `.prd-os/spillover.jsonl:274`, `status: open`, with a concrete delete-on-ASK-226-merge plan. The gate cannot be hand-cleared while it stands.
- **The test is not decoration.** Ran `test-install-plist.sh` against the PR-state templates in a scratch tree: 15 PASS, 0 FAIL, including the negative self-test (neutered renderer rejected, exit 1) and its control (clean template still passes). The control is what makes the negative test mean something.
- **sed injection.** A `$HOME` or `$KIPI_REPO` containing `&` corrupts the replacement into `/Users/a__KIPI_REPO__b/...`, and `assert_rendered` (`install-plist.sh:64-70`) greps for exactly that and exits 1. Fails loud, not silent.
- **CI.** `validate` passes on head `ff26b29`.

Could not verify, stated honestly: whether the currently-installed `~/Library/LaunchAgents/com.kipi.dispatch.plist` byte-matches a render of the new template. The sandbox blocks reads outside the repo. I have no evidence of drift, so I am not raising one.

**Review-process note:** the round-2 rules assume earlier findings are on the PR. `gh pr view 37 --comments` returns nothing. If a round 1 happened, it was never recorded here, so no finding could be checked against a prior citation.

## VERDICT: APPROVE WITH NITS

Fix first: point the two fatal pages at `~/Library/LaunchAgents/com.kipi.dispatch.plist` and add `bash q-system/.q-system/scripts/install-plist.sh com.kipi.dispatch` as the repair command, so the 3am instruction matches what the code reads.

FINDINGS:
minor|Two dispatcher fatal pages tell the operator to fix the path and PATH "in com.kipi.dispatch.plist", but that file now holds placeholders and editing it does not affect the running job|kipi-dispatch.sh:50
END FINDINGS

PR #39, opened 2026-07-28
fix(ops): launchd was killing every converge the instant it was dispatched (ASK-113)

assafkip:
**The autonomous loop has been dispatching and doing no work. Not slowly — zero.**

`kipi-dispatch.sh` launched with `nohup ./kipi converge ... & disown`. Correct in an interactive shell, wrong under launchd, which reaps the job's whole process group when the main process exits. `nohup` only blocks SIGHUP.

### Why nothing caught it — every visible signal said healthy
- the converge log **exists** (the redirect creates it before the child dies) and is **0 bytes**, which reads as "just started"
- `say "dispatched $NEXT"` ran unconditionally, so the log claimed success
- the budget counter was already spent, because it is spent *before* launching
- the liveness beacon kept reporting a healthy heartbeat

Observed 2026-07-28: ASK-224 dispatched at 15:01, 15:16, 15:31 — three budget slots spent, three 0-byte logs, no converge process alive at any point.

### Proven, not reasoned about
A launchd job whose only act was `nohup bash -c "sleep 25; touch F" & disown`:

| context | result |
|---|---|
| under launchd | F never written (**reaped**) |
| identical script, interactive shell | F written (survived) |
| under launchd, with `setsid(2)` | F written (**fix confirmed**) |

### The fix
Give the child its own session, outside the group launchd tears down. macOS ships no `setsid(1)`, so `python3`'s `start_new_session=True` calls `setsid(2)`. Log opened append-only so a re-dispatch cannot erase the previous run's evidence.

**And the dispatch is now proved alive before it is claimed** — the process shows up in the table within 10s, or the founder is paged and the run exits non-zero. Asserts the outcome, not one cause.

The liveness match deliberately avoids `pgrep -f "...$NEXT\b"`: `\b` is a GNU extension BSD pgrep ignores, so on the platform this runs on it would never match and a *healthy* run would report as died.

### Checks (real output)
```
bash -n kipi-dispatch.sh          OK
test-dispatch-liveness.sh         10 passed, 0 failed
capability-gate.py                GREEN, ran=78, exit 0
```

Section 3 of the test is structural on purpose: CI is Linux, no launchd, so the reap cannot be reproduced there. It stops a rewrite quietly restoring `nohup ... & disown` — the same silent-revert shape that dropped `RESET_HOUR` from this file earlier today.

Includes #37 (merged in) so this stands alone.

### DIFF
--- a/kipi-dispatch.sh
+++ b/kipi-dispatch.sh
@@ -184,9 +184,68 @@ fi
 printf '%s' "$((DISPATCHED_TODAY + 1))" > "$COUNT_FILE"
 
 say "dispatching $NEXT (live=$LIVE cap=$MAX_CONCURRENT rounds=$MAX_ROUNDS budget=$((DISPATCHED_TODAY + 1))/$DAILY_MAX)"
-nohup ./kipi converge --issue "$NEXT" --max-rounds "$MAX_ROUNDS" \
-  > "$HOME/.config/kipi/converge-$NEXT.log" 2>&1 &
-disown
 
-say "dispatched $NEXT"
+# THE CHILD NEEDS ITS OWN SESSION, AND THIS IS NOT A STYLE CHOICE.
+#
+# This was `nohup ... & disown`, which is correct in an interactive shell and
+# WRONG under launchd. launchd reaps the job's whole process group when the main
+# process exits; nohup only blocks SIGHUP, so the converge was killed the instant
+# this script returned. Every launchd dispatch since the dispatcher was installed
+# died that way, and the failure was invisible by construction: the log file is
+# created by the redirect before the child dies, so it exists and is 0 bytes,
+# `say "dispatched $NEXT"` still runs, and the budget counter is already spent.
+# The loop reported four healthy dispatches of ASK-224 on 2026-07-28 and did no
+# work at all -- it only spent the subscription.
+#
+# PROVEN, not reasoned about. A launchd job whose only act was
+# `nohup bash -c "sleep 25; touch F" & disown; exit`:
+#     under launchd   F never written  (child killed)
+#     same script from an interactive shell   F written (child survived)
+# and with the setsid form below, under launchd, F is written.
+#
+# macOS ships no setsid(1), so python3 is how setsid(2) gets called. A new
+# session means a new process group with no controlling terminal, which is
+# outside the group launchd tears down.
+python3 - "$HOME/.config/kipi/converge-$NEXT.log" \
+         ./kipi converge --issue "$NEXT" --max-rounds "$MAX_ROUNDS" <<'PY'
+import subprocess, sys
+log_path, argv = sys.argv[1], sys.argv[2:]
+# Append, never truncate: a re-dispatch of the same issue must not erase the
+# evidence of the previous run (the burst incident truncated a live log with >).
+log = open(log_path, "ab", buffering=0)
+subprocess.Popen(argv, stdout=log, stderr=log, start_new_session=True)
+PY
+RC=$?
+if [ "$RC" -ne 0 ]; then
+  # A launch that failed must NOT report success -- that is the same shape as
+  # the bug above. The budget slot is already spent, so say so plainly.
+  say "FAILED to launch converge for $NEXT (rc=$RC); the budget slot is spent"
+  page "kipi dispatch: could not launch the converge run for $NEXT, so NO work is happening even though the loop looks alive. Do: run \`bash kipi-dispatch.sh\` by hand and read the error."
+  exit 1
+fi
+
+# PROVE IT IS ALIVE BEFORE CLAIMING IT. The whole defect above was a dispatch
+# that reported success into a void, so the report is now evidence-backed: the
+# process either shows up in the table or the founder hears about it.
+#
+# NOT `pgrep -f "...$NEXT\b"`. \b is a GNU regex extension and BSD pgrep (macOS,
+# where this actually runs under launchd) does not honour it, so that pattern
+# never matches and a HEALTHY run gets reported as died -- a false alarm is how
+# an alert earns itself muted. The boundary is done in grep, which does support
+# it, against `pgrep -fl` output.
+DISPATCH_OK=0
+for _ in 1 2 3 4 5 6 7 8 9 10; do
+  if pgrep -fl "converge.sh --issue" 2>/dev/null \
+       | grep -qE "[[:space:]]${NEXT}([[:space:]]|\$)"; then
+    DISPATCH_OK=1; break
+  fi
+  sleep 1
+done
+if [ "$DISPATCH_OK" -eq 1 ]; then
+  say "dispatched $NEXT (confirmed running)"
+else
+  say "DISPATCH DIED: $NEXT was launched but no converge process is alive after 10s"
+  page "kipi dispatch: $NEXT was launched but died immediately -- the loop is spending budget and doing no work. Do: check ~/.config/kipi/converge-$NEXT.log and whether launchd is reaping the child."
+  exit 1
+fi
 exit 0
--- a/kipi-dispatch.sh
+++ b/kipi-dispatch.sh
@@ -174,7 +174,15 @@ fi
 
 # Belt and braces against the race between dispatch and the In Progress
 # transition: two converge runs on one issue would fight over one worktree.
-if pgrep -f "converge.sh --issue $NEXT\b" >/dev/null 2>&1; then
+#
+# NOT pgrep, and NOT \b (PR #39 review, finding 2). BSD pgrep reads `\b` as a
+# literal `b`, so this guard has never fired on macOS -- the only platform it
+# runs on. It was harmless while every dispatched child was being reaped
+# instantly; the moment children survive (the fix below), it becomes reachable
+# and lets a second converge start on an issue that already has one. Same
+# `ps -Ao args=` form and same [c] self-match guard as the liveness check.
+if ps -Ao args= 2>/dev/null \
+     | grep -qE "[c]onverge\.sh --issue ${NEXT}([[:space:]]|\$)"; then
   say "skip $NEXT: a converge run for it is already live"
   exit 0
 fi
@@ -206,15 +214,28 @@ say "dispatching $NEXT (live=$LIVE cap=$MAX_CONCURRENT rounds=$MAX_ROUNDS budget
 # macOS ships no setsid(1), so python3 is how setsid(2) gets called. A new
 # session means a new process group with no controlling terminal, which is
 # outside the group launchd tears down.
-python3 - "$HOME/.config/kipi/converge-$NEXT.log" \
+CONVERGE_LOG="$HOME/.config/kipi/converge-$NEXT.log"
+# A RUN BOUNDARY, because the log is appended (PR #39 review, finding 3). The
+# failure page points the operator at this file; without a marker they cannot
+# tell where a re-dispatch's output starts and are reading the previous run's
+# tail as if it were this one's.
+printf '\n===== dispatch %s  %s  rounds=%s =====\n' \
+  "$NEXT" "$(date -u +%Y-%m-%dT%H:%M:%SZ)" "$MAX_ROUNDS" >> "$CONVERGE_LOG"
+
+CHILD_PID="$(python3 - "$CONVERGE_LOG" \
          ./kipi converge --issue "$NEXT" --max-rounds "$MAX_ROUNDS" <<'PY'
 import subprocess, sys
 log_path, argv = sys.argv[1], sys.argv[2:]
 # Append, never truncate: a re-dispatch of the same issue must not erase the
 # evidence of the previous run (the burst incident truncated a live log with >).
 log = open(log_path, "ab", buffering=0)
-subprocess.Popen(argv, stdout=log, stderr=log, start_new_session=True)
+p = subprocess.Popen(argv, stdout=log, stderr=log, start_new_session=True)
+# The PID is the whole point: the caller has to watch THE CHILD IT LAUNCHED,
+# not "some converge for this issue". `kipi` runs converge.sh with bash rather
+# than exec, so this pid stays alive exactly as long as the run does.
+print(p.pid)
 PY
+)"
 RC=$?
 if [ "$RC" -ne 0 ]; then
   # A launch that failed must NOT report success -- that is the same shape as
@@ -234,23 +255,32 @@ fi
 # an alert earns itself muted. The boundary is done in grep, which does support
 # it, against `pgrep -fl` output.
 #
-# NOR `pgrep -fl`. Its -l means two different things: on macOS it prints the full
-# command line, on Linux (procps) it prints only the process NAME, so the issue
-# id is simply not in the output there and a healthy run reads as died. That is
-# not academic -- it failed exactly that way on CI.
+# WATCH THE PID, NOT THE PROCESS TABLE (PR #39 review, finding 1). Asking "is
+# some converge for this issue running?" lets an UNRELATED live converge answer
+# on the dead child's behalf -- which is the exact silent-success hole this
+# check exists to close, rebuilt one layer up. The reachable chain the reviewer
+# walked: the duplicate guard above was dead on macOS, so a second converge
+# started while one was live, converge.sh refused the claim and that child died
+# instantly, and the table still held converge #1. Success reported, budget
+# spent, nobody paged.
 #
-# `ps -Ao args=` prints full command lines on both. The [c] bracket is the
-# standard trick to stop this grep from matching ITSELF: the pattern matches the
-# text "converge", while this process's own command line contains the literal
-# "[c]onverge", which does not.
+# `kill -0` sends no signal; it only asks whether the pid is still there.
+# Checked every second rather than once, so a child that dies at t+4 is caught
+# too -- "alive at least once" would pass a run that fell over immediately after
+# starting, which is most of the ways this actually fails.
 DISPATCH_OK=0
-for _ in 1 2 3 4 5 6 7 8 9 10; do
-  if ps -Ao args= 2>/dev/null \
-       | grep -qE "[c]onverge\.sh --issue ${NEXT}([[:space:]]|\$)"; then
-    DISPATCH_OK=1; break
-  fi
-  sleep 1
-done
+case "$CHILD_PID" in
+  ''|*[!0-9]*)
+    say "DISPATCH DIED: no child pid was returned for $NEXT"
+    ;;
+  *)
+    DISPATCH_OK=1
+    for _ in 1 2 3 4 5 6 7 8 9 10; do
+      if ! kill -0 "$CHILD_PID" 2>/dev/null; then DISPATCH_OK=0; break; fi
+      sleep 1
+    done
+    ;;
+esac
 if [ "$DISPATCH_OK" -eq 1 ]; then
   say "dispatched $NEXT (confirmed running)"
 else
--- a/kipi-dispatch.sh
+++ b/kipi-dispatch.sh
@@ -233,10 +233,20 @@ fi
 # never matches and a HEALTHY run gets reported as died -- a false alarm is how
 # an alert earns itself muted. The boundary is done in grep, which does support
 # it, against `pgrep -fl` output.
+#
+# NOR `pgrep -fl`. Its -l means two different things: on macOS it prints the full
+# command line, on Linux (procps) it prints only the process NAME, so the issue
+# id is simply not in the output there and a healthy run reads as died. That is
+# not academic -- it failed exactly that way on CI.
+#
+# `ps -Ao args=` prints full command lines on both. The [c] bracket is the
+# standard trick to stop this grep from matching ITSELF: the pattern matches the
+# text "converge", while this process's own command line contains the literal
+# "[c]onverge", which does not.
 DISPATCH_OK=0
 for _ in 1 2 3 4 5 6 7 8 9 10; do
-  if pgrep -fl "converge.sh --issue" 2>/dev/null \
-       | grep -qE "[[:space:]]${NEXT}([[:space:]]|\$)"; then
+  if ps -Ao args= 2>/dev/null \
+       | grep -qE "[c]onverge\.sh --issue ${NEXT}([[:space:]]|\$)"; then
     DISPATCH_OK=1; break
   fi
   sleep 1
--- a/kipi-dispatch.sh
+++ b/kipi-dispatch.sh
@@ -181,8 +181,20 @@ fi
 # instantly; the moment children survive (the fix below), it becomes reachable
 # and lets a second converge start on an issue that already has one. Same
 # `ps -Ao args=` form and same [c] self-match guard as the liveness check.
-if ps -Ao args= 2>/dev/null \
-     | grep -qE "[c]onverge\.sh --issue ${NEXT}([[:space:]]|\$)"; then
+# NO PIPE INTO grep -q, and that is the whole point (PR #39 review r3,
+# finding 1). `ps ... | grep -q` under `set -o pipefail` fires only sometimes:
+# grep -q exits the instant it matches, ps then takes SIGPIPE and dies 141, and
+# pipefail makes 141 the status of the whole pipeline -- so the `if` does NOT
+# run its body. Whether ps has finished writing before grep leaves is a race,
+# so the guard worked load-dependently, which is worse than never working
+# because it looks fine when you test it by hand.
+#
+# A snapshot into a variable plus bash's own =~ removes the pipeline entirely,
+# so there is nothing to SIGPIPE and nothing for pipefail to poison. It also
+# removes the need for the [c] self-match trick: with no grep process there is
+# no grep command line in the table to match.
+PS_SNAPSHOT="$(ps -Ao args= 2>/dev/null || true)"
+if [[ "$PS_SNAPSHOT" =~ converge\.sh\ --issue\ ${NEXT}([[:space:]]|$) ]]; then
   say "skip $NEXT: a converge run for it is already live"
   exit 0
 fi
--- a/q-system/.q-system/capability-manifest.json
+++ b/q-system/.q-system/capability-manifest.json
@@ -396,6 +396,11 @@
       "path": "q-system/.q-system/scripts/verify-containment-export.py",
       "reason": "built for prd-skeleton-data-containment's export/quarantine flow and never wired",
       "spillover_id": "sp-0f773063"
+    },
+    {
+      "path": "q-system/.q-system/scripts/linear-collapse-jobmigration.py",
+      "reason": "one-shot family collapse: it absorbed the 32-member job-migration family into ASK-151 on 2026-07-28 and has no second family to run on. ASK-226 generalises it; retire this one when that lands",
+      "spillover_id": "sp-fb332466"
     }
   ],
   "uncovered_known": [
--- a/q-system/.q-system/capability-manifest.json
+++ b/q-system/.q-system/capability-manifest.json
@@ -309,6 +309,10 @@
     {
       "path": "q-system/.q-system/test_token_guard_runtime.py",
       "runner": "python3"
+    },
+    {
+      "path": "q-system/.q-system/scripts/test/test-dispatch-liveness.sh",
+      "runner": "bash"
     }
   ],
   "required_data": [],
--- a/q-system/.q-system/scripts/com.kipi.dispatch.plist
+++ b/q-system/.q-system/scripts/com.kipi.dispatch.plist
@@ -8,12 +8,12 @@
 	<key>EnvironmentVariables</key>
 	<dict>
 		<key>KIPI_REPO</key>
-		<string>/Users/assafkipnis/projects/kipi-system</string>
+		<string>__KIPI_REPO__</string>
 		<!-- Same PATH as com.kipi.openloops-heartbeat, which is the proven
 		     working `claude -p` under launchd. gh and claude both live in
 		     /opt/homebrew/bin; without it the dispatcher pages and exits. -->
 		<key>PATH</key>
-		<string>/Users/assafkipnis/.local/bin:/opt/homebrew/bin:/usr/bin:/bin:/usr/sbin:/sbin</string>
+		<string>__HOME__/.local/bin:/opt/homebrew/bin:/usr/bin:/bin:/usr/sbin:/sbin</string>
 		<!-- ONE at a time, deliberately. Each run spawns a `claude -p` agent
 		     plus a reviewer, so this is also the spend ceiling -- but the real
 		     reason is conflicts. The dispatcher picks by readiness and has NO
@@ -53,7 +53,7 @@
 	<key>ProgramArguments</key>
 	<array>
 		<string>/bin/bash</string>
-		<string>/Users/assafkipnis/projects/kipi-system/kipi-dispatch.sh</string>
+		<string>__KIPI_REPO__/kipi-dispatch.sh</string>
 	</array>
 
 	<!-- Every 15 minutes. The heartbeat only TOPS UP to the concurrency cap;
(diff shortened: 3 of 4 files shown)


assafkip:
## Review: PR #39 — launchd reap fix

Read the diff, the real producer (`kipi:86` → `bash converge.sh --issue X --max-rounds N`), `converge.sh`, `install-plist.sh`, and the capability manifest. Three findings survived reproduction, two dropped.

---

### 1. major — the liveness detector confirms a dispatch that did nothing

`kipi-dispatch.sh:249`

The check asks *"is some converge for this issue in the process table?"* — not *"is the child I just launched alive?"*. Any other converge for the same issue answers on the dead child's behalf.

Reachable chain, every link proven below: the duplicate guard at `:177` is dead (finding 2) → dispatcher launches a second converge while one is live → `converge.sh:78` refuses the claim and that child exits instantly → the detector matches converge #1.

```
bash .pr39rev/repro-liveness-false-confirm.sh
```
```
a converge for ASK-930 is ALREADY live (pid 43691):
bash .../reproC/converge.sh --issue ASK-930 --max-rounds 3

== running PR #39's dispatcher; its own converge child exits instantly ==
dispatcher exit code : 0
dispatch.log         :
    2026-07-28T16:35:52Z dispatching ASK-930 (live=1 cap=2 rounds=3 budget=1/9)
    2026-07-28T16:35:52Z dispatched ASK-930 (confirmed running)
pages sent           : (none)
converge log bytes   : 0

RESULT: FALSE CONFIRM -- a budget slot was spent, no work was launched,
        the dispatcher reported success and paged nobody.
```

That is the original bug's exact signature: 0-byte log, success claimed, budget spent, silence. A human-run `kipi converge --issue X` in a terminal while the heartbeat fires reaches it with no race at all.

Fix that closes it exactly: have the python print the child PID, then poll `kill -0 "$PID"`. Binds the assertion to the process you launched and drops the dependency on cmdline shape.

---

### 2. major — the duplicate-dispatch guard is dead on macOS, and this PR is what makes it reachable

`kipi-dispatch.sh:177` — `pgrep -f "converge.sh --issue $NEXT\b"`

The PR's own comment at `:231` diagnoses this exact bug 54 lines below and leaves the live instance untouched.

```
bash .pr39rev/probe-pgrep-word-boundary.sh
```
```
process IS live:
bash .../dupguard/converge.sh --issue ASK-224 --max-rounds 3

--- the guard EXACTLY as it appears in kipi-dispatch.sh (unchanged by PR #39) ---
GUARD IS DEAD: pgrep found nothing, so the dispatcher would launch a SECOND converge on ASK-224

--- control: same pgrep WITHOUT the \b ---
control MATCHES (proves the process is findable; only \b broke it)

--- what BSD pgrep thinks \b means ---
pgrep -f 'ASK-224\b' MATCHED the process whose arg is literally ASK-224b
  => BSD treats \b as the literal character 'b', it is not a word boundary
```

Pre-existing code, but latent before this PR and live after it: children used to be reaped instantly, so no converge ever survived long enough to be duplicated. Now they run 30+ min (`--max-rounds` × 1800s work + 2400s review) across 2+ heartbeat ticks. `MAX_CONCURRENT=2` does not block it — the repro shows `live=1 cap=2`, dispatcher proceeds. Same one-character fix as `:249` uses: `([[:space:]]|$)` in `grep`, not `\b` in `pgrep`.

---

### 3. minor — the append-only converge log has no run boundary

`kipi-dispatch.sh:215`

Append instead of truncate is right. But nothing writes a delimiter, so the file the page tells the operator to read cannot be dated.

```
bash .pr39rev/repro-log-append-no-delimiter.sh
```
```
--- converge-ASK-nnn.log as the operator would read it ---
     1	round-1 line A
     2	round-1 line B
     3	round-2 line A
---------------------------------------------------------
RESULT: no run boundary, no timestamp. Run 2 is indistinguishable from
        more output from run 1. Line 3 could be either run.
```

One `log.write(b"\n== dispatch <ts> ==\n")` before the Popen.

---

## What is sound

- **`ps -Ao args=` truncation.** Suspected the real 132-char producer cmdline would be clipped with stdout not a tty, which would page on every healthy run. It is not clipped — the PR's exact pattern matched (`probe-ps-truncation.sh`).
- **Fixture shape vs. real producer.** The fake `converge.sh --issue ASK-930 --max-rounds 3` matches what `kipi:86` actually emits, arg for arg. The fixture is honest; the `[c]onverge` bracket genuinely does not self-match.
- **Test 3b's `\s` on BSD grep.** Expected the anti-revert guard to be decoration on macOS. It catches both the column-0 and the indented revert (`probe-test3b-regex.sh`).
- **Plist templating.** `install-plist.sh` is generic over `com.kipi.*.plist`, substitutes both placeholders, and `assert_rendered` exits 1 on a leftover. No doc instructs a raw `cp`. Correctly converts a machine-specific file into a fleet-portable one.
- **The manifest receipt.** `sp-fb332466` is a real open spillover item with a real resolution plan, and `linear-collapse-jobmigration.py` exists. Not a fabricated citation.
- **setsid itself.** New session → new process group → outside what launchd tears down. The documented workaround, and the PR proved it rather than asserting it.

Dropping the truncation and `\s` findings is the process working.

---

## VERDICT: REQUEST CHANGES

Two majors, nothing permanent or unrecoverable, so not a BLOCK — a bad night costs burned budget slots and an issue that did not move, both cleanable.

**Fix first: #1.** Bind the liveness check to the PID python launched. The detector this PR exists to add currently has the same silent-success hole as the bug it replaces, and #2 is what walks it into that hole.

Scratch files are under `.pr39rev/` (untracked, matching the existing `.pr37rev/`/`.pr38rev/` convention). No tracked file touched, nothing committed.

FINDINGS:
major|Liveness check greps the process table for any converge on the issue, so an unrelated live converge confirms a dispatch whose own child died — success reported, budget spent, nobody paged|kipi-dispatch.sh:249
major|Duplicate-dispatch guard uses pgrep with \b, which BSD pgrep reads as a literal 'b', so it never fires on macOS; this PR makes it reachable by letting converge children survive across heartbeat ticks|kipi-dispatch.sh:177
minor|Converge log is appended with no run boundary or timestamp, so the operator cannot tell where a re-dispatch's output begins in the file the page points them to|kipi-dispatch.sh:215
END FINDINGS

assafkip:
# Review: PR #39 — round 2

Read the round-2 diff cold, then the round-1 comment. Two findings survived reproduction. Round 1's findings 1 and 3 are genuinely closed — I attacked both and could not break them.

---

### 1. major — the duplicate-dispatch guard STILL never fires, now for a different reason

`kipi-dispatch.sh:184`

Round-1 finding 2 was `pgrep -f "...$NEXT\b"`. The rewrite to `ps -Ao args= | grep -qE` fixed the `\b`, and the regex itself is correct. The pipeline is not.

`kipi-dispatch.sh:36` sets `set -uo pipefail`. `grep -q` exits on first match, SIGPIPEs `ps` (145 KB of output on this machine, well past the 64 KB pipe buffer), `ps` exits 141, and `pipefail` makes the pipeline report 141. The `if` is false with a duplicate sitting right there in the table.

I am re-raising this because my repro shows it is still live, not because I dislike the fix. Same severity as round 1 — no escalation.

```
bash .pr39rev/repro-dupguard-pipefail.sh
```
```
decoy pid=55962, and it IS in the table:
1
ps -Ao args= total bytes:   145089

GUARD DID NOT FIRE -> would DISPATCH A SECOND converge on ASK-999999 (WRONG)
pipeline exit status under pipefail: 141
same pipeline WITHOUT pipefail: 0
```

5/5 runs identical. End-to-end through the actual dispatcher at PR head `1e76a45`:

```
bash .pr39rev/repro-dupguard-e2e.sh
```
```
already-live converge pid=57508
it is visible in the table: 1 match(es)

dispatcher rc=0
--- dispatch.log ---
2026-07-28T17:07:53Z dispatching ASK-7757495 (live=1 cap=2 rounds=3 budget=1/9)
2026-07-28T17:08:04Z dispatched ASK-7757495 (confirmed running)
--------------------
converge processes now live for ASK-7757495: 2   (expected 1, the guard should have skipped)
```

Reachability, stated honestly: the plist pins `KIPI_DISPATCH_MAX=1`, so under launchd `:93` short-circuits before the guard. The guard is reachable at the script's own default of 2 (`:40`) — which is what a hand-run `bash kipi-dispatch.sh` gets, and a hand-run is exactly what this PR's own failure page instructs at `:244`.

New consequence, not a severity bump: with round-2's liveness check in place, the duplicate no longer confirms silently. It gets its claim refused, dies fast, and pages `DISPATCH DIED ... spending budget and doing no work`. The dead guard now manufactures the alarm.

Not falsifying round 1's `probe-ps-truncation.sh` — truncation genuinely isn't the problem here. That probe just didn't run inside `pipefail`.

Fix: `if [ -n "$(ps -Ao args= | grep -E ...)" ]`, or drop `-q` for `grep -c`, or `|| true` the pipeline into a variable first. Test 5a/5b pass on the current dead guard, so whatever lands needs a behavioural case.

---

### 2. major — the new suite grades the global process table, so it goes red on work it has nothing to do with

`q-system/.q-system/scripts/test/test-dispatch-liveness.sh:104`

`run_dispatch()` pins `KIPI_DISPATCH_DAILY_MAX=9` but not `KIPI_DISPATCH_MAX`. So the dispatcher under test uses the default cap of 2 and consults `live_converges()` (`kipi-dispatch.sh:56`), which greps `converge.sh --issue` with **no issue filter**. Two unrelated converge processes anywhere on the box and the dispatcher exits at `:93` before reaching any of the code the suite asserts on.

The per-run unique `ISS` at `:59` was aimed at this flake and doesn't cover it — the cap is global, the issue id isn't consulted.

```
bash .pr39rev/repro-suite-global-table.sh
```
```
== baseline: suite alone ==
== 16 passed, 0 failed

== now two unrelated converge runs are live (ASK-111, ASK-222) ==
processes matching the dispatcher's own live-counter pattern: 2

  FAIL 1a a died dispatch exits non-zero
  FAIL 1b the log says the dispatch died
  FAIL 1d the founder is paged, naming the symptom
  FAIL 1e the page says the cost is real
  FAIL 2b it reports the run as confirmed
  FAIL 4a a decoy converge does not make a dead dispatch succeed
  FAIL 4b the dead child is still reported as died
  FAIL 4c the founder is still paged
== 8 passed, 8 failed
```

Also 8/16 on all three copies when the suite runs three-up (`.pr39rev/concurrent-suite.sh`), which is the shape the author already saw as 15/17 under the gate.

This is registered in `capability-manifest.json`, so it gates `kipi check`. A gate that reports 8 failures because a converge happened to be running is the flaky-gate outcome the file's own comment argues against.

Fix is one line, same place as the `DAILY_MAX` pin: `KIPI_DISPATCH_MAX=99 \` in `run_dispatch()`. The sandbox then stops asking the real machine how busy it is.

---

## What is sound

- **Round-1 finding 1 (pid binding) is genuinely closed.** Case 4's decoy fixture is honest — the decoy is a real grandchild that outlives the child, and `kill -0 "$CHILD_PID"` reports DIED correctly. I could not get an unrelated converge to vouch for a dead child.
- **`p.pid` tracks the run.** Checked the real producer: `kipi:86` runs `bash converge.sh "$@"` without `exec`, and the converge branch is the last statement, so the wrapper pid lives exactly as long as the run. The comment's claim holds.
- **Round-1 finding 3 (run boundary) is closed.** The dated header lands before the Popen; both writers are append-mode, no interleaving.
- **`ps -Ao args=` truncation.** Re-tested independently with a 200+ char command line, no clipping at 80 or anywhere else. Round 1 was right to drop it.
- **`[c]onverge` self-match guard.** Real — the grep's own argv carries the literal brackets and does not match.
- **launchd restart storm on `exit 1`.** No `KeepAlive` in the plist, `RunAtLoad` false, `StartInterval` 900. A DIED exit does not loop; it's capped at 4/day by the budget.
- **Plist templating.** `install-plist.sh:59` substitutes both placeholders and `:66` fails on a leftover.
- **Child stdin inheriting the heredoc fd.** Suspected this after the "stdin isolation" lesson in the corpus. It's an unlinked temp file at EOF — behaviourally the same as the `/dev/null` the old `nohup` form got. Could not make it fail. Dropped.
- **Cry-wolf on a legitimately fast converge exit.** Chased whether a normal converge can exit inside 10s and trigger a false DIED page. Every non-`--dry` path goes through `linear-worker.sh` and a `claude -p` session. Couldn't construct one. Dropped.

Three findings dropped for failing to reproduce is the process working.

## VERDICT: REQUEST CHANGES

Two majors, both recoverable, nothing permanent — so not a BLOCK. The core fix (`start_new_session`) is right and was proven rather than argued, and round 1's two live findings are properly dead.

**Fix first: #2.** It's a one-line pin, and it's the one that decides whether `kipi check` can be trusted while the loop it tests is running. #1 is the more interesting bug but it's belt-and-braces behind a cap that holds in the scheduled config.

On the review process: round 2 is not producing new blockers on unchanged code. Both of my findings are on lines this round introduced. Round 1's calibration looks right.

Scratch under `.pr39rev/` (untracked, matching `.pr37rev`/`.pr38rev`). No tracked file touched, nothing committed.

FINDINGS:
major|Duplicate-dispatch guard still never fires: set -o pipefail plus grep -q SIGPIPEs ps, so the pipeline returns 141 and the if is false with a duplicate converge live in the table|kipi-dispatch.sh:184
major|test-dispatch-liveness.sh does not pin KIPI_DISPATCH_MAX, so the dispatcher under test consults the global process table and 8 of 16 cases go red whenever two unrelated converge runs exist|q-system/.q-system/scripts/test/test-dispatch-liveness.sh:104
END FINDINGS

assafkip:
Head is `1e76a45` — byte-identical to what round 2 reviewed, `updatedAt` equals round 2's own comment timestamp. No author turn happened in between.

---

# Review: PR #39 — round 3

I read the diff and ran my own attacks before opening rounds 1 and 2. Two findings survived reproduction. Both were raised in round 2; both are still live, and I am pasting my own executed repros rather than citing theirs.

**Before the findings, the thing that matters most about this round:**

```
$ gh pr view 39 --json headRefOid,updatedAt
1e76a457602fbdc13f8ffb3f108e2199fc90ef90  updated=2026-07-28T17:11:59Z

$ gh pr view 39 --json comments -q '.comments[-1].createdAt'
2026-07-28T17:11:59Z
```

The head SHA is the exact commit round 2 reviewed, and the PR's last update *is* round 2's review comment. Round 3 was dispatched against byte-identical code with no author turn in between. Both round-2 findings are live for the trivial reason that nothing was changed.

The prompt's round-3 warning is about earlier rounds being miscalibrated. That is not what happened here, and I want to say so plainly: round 1 filed three findings, two were fixed and I verified them closed, one was correctly dropped. Round 2's two findings are both on lines round 2's own diff introduced. The calibration is fine. The defect is in the loop, which fired a new review round without waiting for a push.

---

### 1. major — the duplicate-dispatch guard fires only about a third of the time

`kipi-dispatch.sh:184`

Round 2's mechanism is right: `set -uo pipefail` at `:36`, `grep -q` exits on first match, `ps` takes SIGPIPE and exits 141, `pipefail` reports 141, the `if` is false with a duplicate sitting in the table.

I am correcting one thing, and it is a correction of shape, not a severity bump: it is **not deterministic**. My first isolated run said dead 5/5, then my end-to-end run through the actual dispatcher **fired correctly**. So I measured the rate instead of reporting whichever result was convenient.

```
bash .pr39rev3/repro-dupguard-rate.sh
```
```
decoy pid=70845
ps -Ao args= total bytes: 147417   (pipe buffer 64KB)
match line number in the stream: 467 of 591 lines

  run  1: FIRED     run  6: DEAD      run 11: DEAD      run 16: DEAD
  run  2: FIRED     run  7: FIRED     run 12: DEAD      run 17: DEAD
  run  3: FIRED     run  8: FIRED     run 13: DEAD      run 18: DEAD
  run  4: FIRED     run  9: DEAD      run 14: DEAD      run 19: DEAD
  run  5: FIRED     run 10: DEAD      run 15: DEAD      run 20: DEAD

FIRED=7  DEAD=13  out of 20   (a duplicate converge IS live for the whole run)

=== control ===
  same regex, same snapshot, read from a FILE (no pipe): 0
```

The file control proves the regex and the process are both fine. Only the pipeline is broken. Whether it fails depends on whether `ps` has finished draining into the 64 KB pipe buffer before `grep -q` bails, which is load-dependent.

An intermittently dead guard is harder to diagnose than a reliably dead one, and the same severity to be hit by. Not escalating past round 2's major.

Reachability, stated the same way round 2 did: the plist pins `KIPI_DISPATCH_MAX=1`, so under launchd `:93` short-circuits before the guard whenever a converge is live. It is reachable at the script's own default of 2 (`:40`), which is what a hand-run gets, and a hand-run is what this PR's own failure page instructs at `:244`.

Fix: take the pipeline out of the `if`. `SNAP="$(ps -Ao args= 2>/dev/null || true)"` then `case "$SNAP" in ...` or a `grep` against the variable. Tests 5a/5b pass on the current broken guard, so whatever lands needs a behavioural case.

---

### 2. major — the new suite grades the global process table

`q-system/.q-system/scripts/test/test-dispatch-liveness.sh:104`

`run_dispatch()` pins `KIPI_DISPATCH_DAILY_MAX=9` but not `KIPI_DISPATCH_MAX`, so the dispatcher under test uses the default cap of 2 and calls `live_converges()` (`kipi-dispatch.sh:52`), which greps `converge.sh --issue` with no issue filter. The per-run unique `ISS` at `:59` does not cover this, because the cap never consults the issue id.

Baseline on a quiet machine, my own run:

```
$ cd .pr39rev3/repo && bash q-system/.q-system/scripts/test/test-dispatch-liveness.sh
== 16 passed, 0 failed
bash ... 0.18s user 0.23s system 3% cpu 12.587 total
```

Now with two unrelated converge processes in the table:

```
bash .pr39rev3/hermetic.sh
```
```
decoys live: 2

  FAIL 1a a died dispatch exits non-zero
     expected [1] got [0]
  FAIL 1b the log says the dispatch died
     2026-07-28T17:19:36Z skip: 2 converge run(s) live, cap 2
  FAIL 1d the founder is paged, naming the symptom
  FAIL 1e the page says the cost is real
  FAIL 2b it reports the run as confirmed
  FAIL 4a a decoy converge does not make a dead dispatch succeed
  FAIL 4b the dead child is still reported as died
  FAIL 4c the founder is still paged
== 8 passed, 8 failed

suite exit code: 1
```

I confirmed this actually gates something rather than being cosmetic. The manifest entry lands under `expected_tests`, so `kipi check` runs it:

```
MANIFEST KEY PATH: /expected_tests/path = q-system/.q-system/scripts/test/test-dispatch-liveness.sh
```

The file's own comment at `:250` argues that a flaky gate is worse than a narrow one because it teaches everyone to re-run until green. That argument is correct and this file loses to it. One line, same place as the `DAILY_MAX` pin: `KIPI_DISPATCH_MAX=99 \`.

---

### 3. minor — both prior rounds' REQUEST CHANGES verdicts carry no enforcement

Rounds 1 and 2 both ended in "VERDICT: REQUEST CHANGES". Neither was submitted as a GitHub review:

```
$ gh api repos/assafkip/kipi-system/pulls/39/reviews -q '.[] | "\(.state) \(.user.login)"'
(empty)

$ gh pr view 39 --json mergeStateStatus,reviewDecision,autoMergeRequest
mergeState=BLOCKED  reviewDecision=  auto=SQUASH
```

Auto-merge is armed, `validate` is SUCCESS, and `reviewDecision` is null. What is holding this PR is a branch-protection rule requiring an approval, not the two reviews. The verdicts are prose in a comment box; relax that protection and two rounds of blocking review evaporate silently. `gh pr review 39 --request-changes` costs nothing and makes the verdict a real object.

---

## What is sound

- **`ps -Ao args=` truncation on macOS.** My main suspicion going in: the real producer emits a 114-char command line (`kipi:86` → `bash /Users/assafkipnis/projects/kipi-system/q-system/.q-system/scripts/converge.sh --issue ASK-224 --max-rounds 3`), and BSD `ps` clipping to 80 columns with stdout not a tty would make both guards dead and page on every healthy run. Reproduced with a 129-char line: not clipped, guard regex matched (`.pr39rev3/ps_trunc.sh`). Reached independently of rounds 1 and 2.
- **Round 1's finding 1 (pid binding) is genuinely closed.** I checked the real producer: `kipi:86` runs `bash "$KIPI_HOME/.../converge.sh" "$@"` without `exec`, and the converge branch is the last statement, so `p.pid` lives exactly as long as the run. The comment's claim holds.
- **Round 1's finding 3 (run boundary) is closed.** The dated header is written before the Popen, both writers append.
- **The test is wired, not decoration.** `expected_tests` in the manifest, confirmed by parsing the PR-head copy.
- **Blast radius of merging with the two majors live.** Auto-merge is armed and the only required check is green, so I checked whether this could land unattended. `mergeStateStatus=BLOCKED`. It cannot. Dropped.
- **The end-to-end duplicate repro fired correctly on first run.** That is the result that would have let me drop finding 1 as unreproducible. I measured the rate instead, and it is 13/20 dead. Worth naming as a near-miss: a single e2e run is not evidence about a load-dependent race.
- **Child stdin inheriting the heredoc fd.** Suspected a behaviour change versus the old `nohup` form. It is an unlinked temp file at EOF, same as the `/dev/null` launchd used to hand it. Could not make it fail.
- **Unchecked `python3` availability.** Not guarded like `gh` is, but a missing interpreter yields rc 127 into the `FAILED to launch` branch, which pages correctly. Dropped.

Three attacks dropped for failing to reproduce, and one finding kept only because I did not stop at the first convenient result.

## VERDICT: REQUEST CHANGES

Two majors, both recoverable, nothing permanent or unrecoverable, so not a BLOCK. The core fix (`start_new_session=True`) is right and was proven rather than argued, and every round-1 finding is properly dead.

**Fix first: #2.** It is a one-line pin and it decides whether `kipi check` can be trusted while the loop it tests is running. #1 is the more interesting bug but the concurrency cap covers it in the scheduled config.

**And fix the loop, not just the code:** gate the review dispatcher on `headRefOid` changing since the last reviewed round. This round cost a full adversarial pass to re-confirm two known findings on unchanged bytes.

Scratch under `.pr39rev3/` (untracked, matching the existing `.pr37rev`/`.pr38rev`/`.pr39rev` convention). No tracked file touched, nothing committed or pushed.

FINDINGS:
major|Duplicate-dispatch guard fires only ~35% of the time: set -o pipefail plus grep -q SIGPIPEs ps, so the pipeline returns 141 load-dependently and a second converge launches on an issue that already has one|kipi-dispatch.sh:184
major|test-dispatch-liveness.sh does not pin KIPI_DISPATCH_MAX, so the dispatcher under test reads the global process table and 8 of 16 cases go red whenever two unrelated converge runs exist on the box|q-system/.q-system/scripts/test/test-dispatch-liveness.sh:104
minor|Rounds 1 and 2 posted REQUEST CHANGES as plain issue comments rather than GitHub reviews, so reviewDecision is null and the verdicts do not block the armed auto-merge|.github/pull_request_39:1
END FINDINGS

assafkip:
No leaked processes; the only `M` entries are pre-existing scratch from earlier rounds (present in the session's opening `git status`). I touched no tracked file.

---

# Review: PR #39 — round 4

I read the diff and ran my attacks before opening rounds 1–3. **Nothing above minor survived reproduction.** Both round-3 majors are genuinely closed — I attacked each with my own repro rather than trusting the commit messages.

---

### 1. minor — the liveness block's comments describe an implementation that was deleted two rounds ago

`kipi-dispatch.sh:183`, `kipi-dispatch.sh:262`, `kipi-dispatch.sh:267-268`

Three claims, all false against the code beneath them. `:183` also contradicts `:194` inside the same comment block.

```
bash .pr39rev4/repro-stale-comments.sh
```
```
=== CLAIM A: the duplicate guard uses 'the same [c] self-match guard as the liveness check' (:183)
# `ps -Ao args=` form and same [c] self-match guard as the liveness check.
# removes the need for the [c] self-match trick: with no grep process there is
--- is there ANY [c]-style self-match guard in the file? ---
  none outside the comments themselves
--- what the liveness check actually does ---
9:      if ! kill -0 "$CHILD_PID" 2>/dev/null; then DISPATCH_OK=0; break; fi

=== CLAIM B: the liveness check greps a process table with a grep-side word boundary (:262,:267-268)
# process either shows up in the table or the founder hears about it.
# ... The boundary is done in grep, which does support
# it, against `pgrep -fl` output.
--- does the liveness block contain grep, pgrep, or ps at all? ---
5:# NOT `pgrep -f "...$NEXT\b"`. ...        <- comment text only
8:# an alert earns itself muted. ...        <- comment text only
9:# it, against `pgrep -fl` output.         <- comment text only
```

The only hits are the comment lines themselves; there is no `grep`, `pgrep`, or `ps` in `:260-303`. In a file where the comments *are* the scar record, a comment that walks the next engineer through a `pgrep -fl` + word-boundary design is worse than no comment — it will be defended in the next review.

---

### 2. minor — the issue-id boundary in the duplicate guard has no test

`kipi-dispatch.sh:197` — `${NEXT}([[:space:]]|$)`

Section 5 greps for the *absence* of the old `pgrep`/`\b` spelling; section 6 drives one issue id. Neither pins the boundary. Delete it and the suite stays green.

```
bash .pr39rev4/repro-mutations.sh
```
```
=== baseline (unmutated) ===
[baseline] suite rc=0  == 18 passed, 0 failed

=== M3: issue-id boundary removed from the duplicate guard (ASK-22 vs ASK-224) ===
[M3-no-boundary] suite rc=0  == 18 passed, 0 failed
```

The boundary is correct today — I verified it both directions:

```
bash .pr39rev4/repro-guard-rate.sh
```
```
--- prefix collision: NEXT=ASK-22 while ASK-224 is live (must NOT fire) ---
did not fire (correct)
--- the $ branch: converge line is the LAST line of the snapshot ---
matches at end-of-string (the $ alternative works in bash [[ =~ ]])
```

If it regresses, `ASK-22` gets skipped every heartbeat while a 30–60 min `ASK-224` converge runs, logging `skip ASK-22: a converge run for it is already live` — a true-looking line for a false reason. One case with two ids covers it.

---

### 3. minor — the prior rounds' REQUEST CHANGES verdicts are still not GitHub review objects

Round 3 raised this. Re-raising because my own repro shows it is still live, at the same severity, not escalated.

```
$ gh api repos/assafkip/kipi-system/pulls/39/reviews -q '.[] | "\(.state) \(.user.login)"'
(empty)
$ gh pr view 39 --json mergeStateStatus,reviewDecision
mergeState=BLOCKED reviewDecision=
```

Three rounds of blocking verdicts, `reviewDecision` still null. Auto-merge is armed (`SQUASH`); what holds this PR is a branch-protection approval requirement, not the reviews. Relax that rule and three rounds evaporate silently.

---

### 4. nit — the run-boundary header has no test

`kipi-dispatch.sh:234` — round-1 finding 3's fix. `M4-no-run-boundary` → `18 passed, 0 failed` (output above).

---

### 5. nit — the DIED message hardcodes "after 10s" on a branch that waits 0s

`kipi-dispatch.sh:299`. The `else` arm is shared by the 10-iteration poll and the empty-pid branch at `:285`. Reaching it needed a code mutation (drop `print(p.pid)`), so per calibration I dropped this a level — no producer emits it on shipped code.

```
bash .pr39rev4/repro-nopid.sh
```
```
dispatcher rc=1   elapsed=0s
--- dispatch.log ---
17:49:23Z DISPATCH DIED: no child pid was returned for ASK-4242
17:49:23Z DISPATCH DIED: ASK-4242 was launched but no converge process is alive after 10s
--- page sent ---
kipi dispatch: ASK-4242 was launched but died immediately -- the loop is spending budget and doing no work.
--- is the converge child actually alive? ---
23509 /usr/bin/env bash ./kipi converge --issue ASK-4242 --max-rounds 3
```

Two contradictory DIED lines, and the page asserts "doing no work" while the child is running.

---

## What is sound

- **Round-3 finding 1 (pipefail/SIGPIPE duplicate guard) is genuinely closed.** Round 3 measured 13/20 dead. PR head, 30 runs, ~140 KB `ps` output, duplicate live the whole time: `FIRED=30 DEAD=0`. There is no pipeline left to poison. Mutating it *back* to the `ps | grep -q` form fails cases 6a and 6b, so the fix has a behavioural test with teeth, not a structural grep.
- **Round-3 finding 2 (suite grades the global process table) is genuinely closed.** With 4 unrelated converges live: `18 passed, 0 failed`. Three copies concurrently, decoys still live: `18/18`, `18/18`, `18/18`. The `KIPI_DISPATCH_MAX=999` pin holds.
- **The every-second poll is load-bearing and tested.** My main suspicion: the `for _ in 1..10` loop is decoration and a single `kill -0` would pass everything. Mutated it to one check → `13 passed, 5 failed` (1a–1e). The `./kipi` wrapper is briefly alive at t=0, so a one-shot check false-confirms. The comment's claim is real and the tests enforce it.
- **Test 3b's `\s` on BSD grep.** Expected the anti-revert guard to be decoration on macOS. `grep (BSD grep, GNU compatible) 2.6.0-FreeBSD` honours `\s`; the guard fires on a real reverted `nohup ./kipi converge` line.
- **The `$` end-anchor in `[[ =~ ]]`.** Command substitution strips the trailing newline, so the last `ps` line has no `[[:space:]]` after it and depends entirely on the `$` alternative. Verified it matches.
- **Plist templating is wired, not text-in-a-file.** `install-plist.sh` is generic over `com.kipi.*.plist`, substitutes both placeholders with `|` as sed delimiter, and `assert_rendered` exits 1 on a leftover. This change converts a founder-machine-specific plist into a fleet-portable one.
- **The test is wired.** Manifest key is `expected_tests`, not a decorative list: `{'path': '.../test-dispatch-liveness.sh', 'runner': 'bash'}`.
- **`p.pid` tracks the run.** `kipi:86` runs `bash converge.sh "$@"` without `exec`, and the converge branch is the last statement, so the wrapper lives exactly as long as the run.
- **Cry-wolf on a fast legitimate converge.** Chased whether a healthy run can exit inside 10s and page falsely. The only sub-10s exits in `converge.sh` are `--dry` (never passed by the dispatcher) and the usage error; every real path goes through a `claude -p` session. Could not construct one. Dropped.
- **Process leakage from the suite.** Case 2 dispatches a child into its own session; `pkill -f "$ROOT/converge.sh"` still reaps it. Post-run scan: clean.

Four attacks dropped for failing to reproduce.

## On the review process

The round-3+ warning does not apply here, and I want to say so with the evidence. Round 1 filed three findings: two were fixed and I independently verified them closed, one (`ps` truncation) was correctly dropped. Rounds 2 and 3 filed findings on lines their own diffs introduced. Round 3 correctly diagnosed that it had been dispatched against byte-identical code and said so rather than padding. This round, against genuinely changed code, nothing above minor survived. **That is a PR converging, not earlier rounds miscalibrated.** The remaining process defect is #3: three rounds of verdicts that GitHub does not know exist.

One stale claim in the PR description, not a code finding: it reports `test-dispatch-liveness.sh 10 passed, 0 failed`. The suite now has 18 cases.

## VERDICT: APPROVE WITH NITS

No blocker, no major. Nothing here is permanent or unrecoverable, and the core fix (`start_new_session=True`) was proven rather than argued.

**Fix first: #1.** The stale comments are the cheapest fix and the highest leverage — this file's comments are how the next engineer learns the design, and right now they teach a `pgrep -fl` liveness check that does not exist.

Scratch under `.pr39rev4/` (untracked, matching the existing `.pr37rev`/`.pr38rev`/`.pr39rev` convention). No tracked file touched, nothing committed or pushed.

FINDINGS:
minor|Liveness-block comments describe a [c] self-match guard and a grep/pgrep -fl process-table search that the code does not contain, and :183 contradicts :194 in the same block|kipi-dispatch.sh:183
minor|The duplicate guard's issue-id boundary ([[:space:]]|$) has no test: removing it leaves the suite at 18/18, and a regression silently skips ASK-22 whenever ASK-224 is live|kipi-dispatch.sh:197
minor|Rounds 1-3 posted REQUEST CHANGES as issue comments rather than GitHub reviews, so reviewDecision is still null and the verdicts do not block the armed auto-merge|.github/pull_request_39:1
nit|The converge-log run-boundary header has no test: removing it leaves the suite at 18/18|kipi-dispatch.sh:234
nit|The DIED message hardcodes "after 10s" and "died immediately" on an else-arm also reached from the zero-second empty-pid branch|kipi-dispatch.sh:299
END FINDINGS

PR #40, opened 2026-07-28
feat(pipeline): every page says what broke AND what to do, and an outside watcher catches the deaths no in-process pager can (ASK-223)

assafkip:
Closes the three gaps ASK-223 names: pages with no next action, no observer outside the failing process, and no classifier separating breakage from the system working.

## Gap 1: the pages said WHAT, never WHAT TO DO

`converge ASK-208: hit 3-round cap, still REQUEST CHANGES` wakes the founder at 3am and leaves them to work out the rest. All nine existing pages in `converge.sh` and `linear-worker.sh` now carry a diagnosis and a runnable command:

```
BEFORE  converge ASK-208: hit 3-round cap, still REQUEST CHANGES
AFTER   converge ASK-208: gave up after 3 rounds, still 'REQUEST CHANGES'. Nothing
        merged, so the work is stranded. A cap-out is almost always scope -- 1-change
        issues converge in 1 round. Do: split it, kipi linear issue "<one change>",
        then close ASK-208 as superseded
```

`Do: ` is a literal contract, linted over both scripts by the new suite. Not a convention anyone has to remember.

## Gap 2: nothing watched from outside

Every pager above lives *inside* the thing that fails. `linear-triage.py --apply` died on a Linear `TimeoutError` after commenting on 74 issues and closing 32, and paged zero times (`sp-b5dcf944`). A crashed reviewer reaches nobody (`sp-3a0cac1c`). Both were found by a human reading a log.

`linear-pipeline-health.py` is the outside observer, same shape as `launchd-health-check.py`, on launchd every 15 min. It sees the states no in-process pager can:

- a PR with every required check GREEN that nobody armed (the stall that most looks like success)
- an approving verdict pinned to a sha that is no longer the head
- an issue In Progress with no branch and no PR, past `STRANDED_ISSUE_HOURS`
- a converge log whose last line for an issue is `dispatching`, with no live process
- `main` red on validate

## Gap 3: silence is the hard part

Detection is easy; not crying wolf is not. A muted channel silently removes every alert including the real ones, which is strictly worse than not building this. So:

| state | pages? |
|---|---|
| REQUEST CHANGES on a round | no, logged |
| PR waiting on review | no, logged |
| issue skipped, no DoR | no, logged |
| anything unclassified | no, logged (borderline goes silent by design) |
| round cap / no verdict / worker infra error | yes, once |
| green >20 min unmerged / unreviewed head | yes, once |
| stranded issue / dead converge / main red | yes, once |

Findings dedupe on the **transition**, not the level: the ledger holds only what is broken right now, so three consecutive runs of one breakage page once, and a state that clears and re-breaks pages again. A 15-minute watcher would otherwise emit 96 pages a day for one problem.

## Verification

```
$ bash q-system/.q-system/scripts/test/test-linear-pipeline-health.sh
PASS: 12/12 checks

$ bash q-system/.q-system/scripts/test/test-converge.sh
PASS: 17/17 converge checks

tests: ran=79 quarantined=0 skipped-skeleton-only=0
capability-gate: GREEN
```

RED was observed first: the Gap-1 lint failed with `converge.sh has 7 page(s) with no 'Do: ' action`, listing each one, before any message was rewritten.

The suite never calls gh, Linear, or Slack. `KIPI_PIPELINE_OBSERVATIONS` swaps a JSON fixture in for the live collectors, so the classifier and the dedupe ledger are driven end-to-end offline; the pure detectors (`dead_converge_findings`, `green_not_merged_findings`, `unreviewed_head_findings`, `main_red_findings`) are called directly.

## One assertion changed, and why

`test-severity-floor.sh` pinned the literal words `needs a human` on the conflict-cap page. That phrase is exactly what this issue removes: it passes on a page that names no next step. Re-anchored onto `Do: gh pr checkout`, which is strictly stronger.

## Merge ordering

Branched off `main` (which has ASK-222). **ASK-221 (PR #34) is still open and also touches `linear-worker.sh`** at ~line 1100; this PR's worker edits are the seven `$NOTIFY` message strings at lines 193/452/466/680/702/835/1032, so the hunks are disjoint. Per the DoR this lands after ASK-221.

## Not doing (per the DoR)

ASK-221's Codex engine, ASK-222's auto-merge arming, a launchd scheduler for `kipi work`/`converge`, any notification channel other than `slack-notify.sh`. This changes what converge and worker SAY on failure, never what they DO.

## Spillover

`sp-7fef7083` — `founder-notifications.md` should list this watcher as a wired emitter and state the `Do:` page contract. The edit was permission-blocked (`.claude/rules/` is a sensitive path). The contract is enforced by the lint regardless; the rule text is the gap.


assafkip:
# Review: PR #40 (ASK-223)

Read-only. Scratch in `.pr40rev/` (untracked); review worktree removed. One local ref `refs/pr40head` remains — its deletion was blocked by the destructive-op hook, and it points at the same commit as `sana/ask-223`, so it is harmless.

---

## MAJOR 1 — The only thing this watcher would say on its first live cycle is a false page about a 27-day-old draft PR

`q-system/.q-system/scripts/linear-pipeline-health.py:211` (`green_not_merged_findings`)

`collect_live()` queries `gh pr list --state open --json number,updatedAt,statusCheckRollup,autoMergeRequest`. That query returns drafts, and `isDraft` is neither requested nor checked. PR #4 in this repo is a draft opened 2026-07-01, `autoMergeRequest: null`, rollup `["SUCCESS"]`.

Repro — no constructed fixture, the live repo through the real code path:

```
$ bash .pr40rev/repro3_live_dry.sh
BROKEN: pipeline PR #4: PR #4 has every required check GREEN and has not merged for 38756 min -- auto-merge was never armed, so it sits green forever. This is the stall that looks most like success. Do: gh pr merge --auto --squash 4
[dry] would page 1 of 1 broken state(s)
rc=0
```

```
$ bash .pr40rev/repro1_draft.sh
--- what the real producer emits for PR #4 ---
{"autoMerge":false,"isDraft":true,"number":5,"updatedAt":"2026-07-02T23:29:04Z"}
{"autoMerge":false,"isDraft":true,"number":4,"updatedAt":"2026-07-01T20:24:28Z"}
```

The diagnosis is wrong (a draft is not an unarmed stall) and the action is wrong (GitHub refuses auto-merge on a draft). `RE_PING_TTL_SECONDS = 6h` means 4 of these per day, forever, starting the first cycle after install. The PR argues at length that a muted channel is strictly worse than not building this; on merge day the channel's entire content is this.

PR #5 is spared only because its rollup is currently FAILURE. It becomes the same page the moment that flips.

---

## MAJOR 2 — `stranded_issue` is promised in the PR body, has a green test, and can never fire

`q-system/.q-system/scripts/linear-pipeline-health.py:66` and `:127`

The PR body's Gap-2 list says the watcher sees "an issue In Progress with no branch and no PR, past `STRANDED_ISSUE_HOURS`". No collector produces that state. The suite is green on it because `KIPI_PIPELINE_OBSERVATIONS` injects the state string directly, bypassing every collector — the test proves `classify()` renders a template, not that anything observes the world.

```
$ bash .pr40rev/repro2_dead_states.sh
states a collector can emit : ['dead_converge', 'green_not_merged', 'main_red', 'unreviewed_head']
states declared BROKEN      : ['dead_converge', 'green_not_merged', 'main_red', 'no_verdict', 'round_cap', 'stranded_issue', 'unreviewed_head', 'worker_infra']

DECLARED BUT UNREACHABLE    : ['no_verdict', 'round_cap', 'stranded_issue', 'worker_infra']

STRANDED_ISSUE_HOURS = 4 -> referenced anywhere else?
  occurrences in module source: 1
```

`STRANDED_ISSUE_HOURS` appears exactly once — its own assignment. `round_cap` / `no_verdict` / `worker_infra` are at least paged in-process by converge and the worker; `stranded_issue` is paged by nobody, and the PR reports it as covered.

---

## MAJOR 3 — One transient `gh` failure wipes the dedupe ledger, and the next cycle re-pages a breakage nobody fixed

`q-system/.q-system/scripts/linear-pipeline-health.py:478` (state rebuilt from current findings only) + `:337` (`collect_live` swallows collector failures)

`_gh_json` returns `None` on any non-zero exit, timeout, or unparseable output — rate limit, network blip, expired token. `main_red_findings(None)` returns `[]`, which is byte-identical to "main went green". The ledger is then rebuilt from the empty set, so the key vanishes, so the next successful cycle sees a transition and pages again. At 96 cycles/day one flaky call turns a single real break into a stream.

```
$ bash .pr40rev/repro4_time_and_ledger.sh
=== 4b: gh failure wipes the ledger and re-pages a breakage already paged ===
after 2 cycles of a REAL, UNFIXED breakage: 1 page(s)  <- correct
ledger after the failed collection: {}
after gh recovers, main never fixed:  2 page(s)  <- should still be 1
pipeline main: main is RED on Skeleton Validation. ... Do: gh run view 30329032046 --log-failed
pipeline main: main is RED on Skeleton Validation. ... Do: gh run view 30329032046 --log-failed

--- proof that a gh failure really does yield [] from the live collector ---
gh on PATH? -> None
main_red_findings(None) -> []
```

Test check 5 asserts this exact behavior as a feature ("state clears then re-breaks -> pages again"). It cannot tell the two apart because the collector layer discards the difference before the classifier sees it.

---

## MINOR 1 — The new comment in `install-plist.sh` documents a job the installer cannot install

`q-system/.q-system/scripts/install-plist.sh:22`

```
$ bash q-system/.q-system/scripts/install-plist.sh com.kipi.launchd-health --render-only .pr40rev/lh.plist
ERROR: no plist template for label 'com.kipi.launchd-health' at .../com.kipi.launchd-health.plist
labels available:
  com.kipi.dispatch
  com.kipi.fleet-health
  com.kipi.linear-dor
  com.kipi.openloops-heartbeat
rc=2
```

The comment block this PR adds lists three watchdogs "this installs"; one of them has no template.

---

## MINOR 2 — The incident the PR opens with is still not covered

`linear-triage.py --apply` dying on a Linear `TimeoutError` after closing 32 issues (`sp-b5dcf944`) is the headline justification for Gap 2. `linear-triage.py` has no pager, and none of the four states this watcher can actually emit observes a triage run:

```
$ bash .pr40rev/repro5_lint.sh   # section 5c
slack-notify.sh  fleet-health-daily.py  launchd-health-check.py  converge.sh
linear-worker.sh  open-loops-heartbeat.sh  lessons-daily.sh
linear-pipeline-health.py  linear-dor-drafter.py
```
(`linear-triage.py` absent; and per repro 2 the emitted states are `green_not_merged`, `unreviewed_head`, `dead_converge`, `main_red`.) Re-run that exact failure after this merges and it still pages zero times.

---

## NIT — `q-system/output/ask223-read.py` is a one-off debug script committed to a public repo

Also `page_text(state, ...)` at `:169` never reads `state`.

I checked whether the scratch script propagates fleet-wide: it does not. `output` is in `INSTANCE_OWNED_SUBTREES` (`kipi-update.sh:68`), so `kipi update`'s rsync excludes it. Dropping that as a concern.

---

- **Invented log fixture.** I expected `dead_converge`'s regex to miss the real producer. `converge.sh:60` emits `converge[$ISSUE] $*` and `:146` emits `round $ROUND/$MAX_ROUNDS dispatching Sana`, into `$STATE_DIR/linear-worker.log` — exactly the shape and the file the detector reads. Grounded.
- **Invented verdict-record fixture.** `unreviewed_head_findings` reads `verdict`/`pr`/`head_sha`/`issue`. `pr-review-agent.sh:271` writes all four with those names. Grounded.
- **`live_converge_issues` arg form.** I expected `--issue\s+([A-Z]+-\d+)` to miss a positional invocation. `converge.sh` accepts `--issue` and nothing else (`usage:` refuses positionals), so a running converge is always visible in `ps`. No false `dead_converge` on a live round.
- **Timezone math.** `_minutes_since` mixes `time.mktime` (local) with `time.timezone` (standard offset) and I expected a 60-minute DST error. It is correct — `%Z` parsing "UTC" sets `tm_isdst=0`, which pins `mktime` to standard time. Verified in a DST zone (`PDT`, `daylight=1`) at both a summer and a winter instant: expected 60, got 60 both times.
- **The Gap-1 lint is not decoration.** Stripping one `Do: ` from a copy of `converge.sh` flips it: `FOUND=7 MISSING=1 -> lint verdict: RED`. It also guards its own coverage (`FOUND -gt 0`), which is the failure mode most lints ship with.
- **`main_red` cross-workflow flapping.** I expected `--limit 1` with no workflow filter to flip red/green between workflows. This repo has exactly one (`validate.yml`). Not reachable.
- **The suite is genuinely green.** `PASS: 12/12 checks` from a clean worktree at the PR head.
- **The `--dry` path is honestly read-only.** Live run wrote no state file and sent no page.

---

## VERDICT: REQUEST CHANGES

Three majors, none permanent or unrecoverable — no Linear object is written, nothing is deleted, and reverting the commit fully undoes it. Not a BLOCK.

**Fix first: the draft-PR false positive (MAJOR 1).** Add `isDraft` to the `gh pr list --json` field set and skip drafts in `green_not_merged_findings`. Without it, the first thing this watcher ever says is wrong, four times a day, about a PR the founder parked on purpose — which is precisely the outcome the PR's own Gap-3 section argues is worse than not shipping.

FINDINGS:
major|green_not_merged pages for draft PRs because isDraft is neither queried nor checked, and PR #4 (draft since 2026-07-01) is the only finding the live watcher emits today|q-system/.q-system/scripts/linear-pipeline-health.py:211
major|stranded_issue is claimed in the PR body and passes its test via fixture injection, but no collector emits it and STRANDED_ISSUE_HOURS is referenced only by its own assignment|q-system/.q-system/scripts/linear-pipeline-health.py:66
major|a transient gh failure returns [] from every collector, which is indistinguishable from cleared, so the dedupe ledger is wiped and an unfixed breakage pages again on the next successful cycle|q-system/.q-system/scripts/linear-pipeline-health.py:478
minor|the new comment block claims install-plist.sh installs com.kipi.launchd-health, but the installer exits 2 with 'no plist template' for that label|q-system/.q-system/scripts/install-plist.sh:22
minor|the motivating incident (linear-triage.py dying silently, sp-b5dcf944) is still uncovered: linear-triage.py has no pager and no collector observes a triage run|q-system/.q-system/scripts/linear-pipeline-health.py:337
nit|a one-off Linear debug script is committed to a public repo|q-system/output/ask223-read.py:1
END FINDINGS

assafkip:
## Round-3 review answered: 6 findings, 6 fixes, 7 new checks

Every finding has a check that fails without its fix. RED was observed against `HEAD` (the reviewed commit) before each one went green.

| finding | verdict | check that would have caught it |
|---|---|---|
| MAJOR 1 draft PR | fixed | check 8 (two halves: detector + live query) |
| MAJOR 2 `stranded_issue` unreachable | fixed | check 9 (reachability lint over every broken state) |
| MAJOR 3 gh failure wipes the ledger | fixed | check 10 (4-cycle sequence: break, break, blind, recover) |
| MINOR 1 install-plist comment | fixed | check 12 |
| MINOR 2 triage incident uncovered | fixed | check 13 |
| NIT scratch script + unused arg | fixed | deleted / removed |

`PASS: 21/21 checks` (was 12/12). `PASS: 17/17 converge checks`.

---

### MAJOR 1 — the draft page

You were right that the fix has to land in two places, and the second is the one that matters. `isDraft` is now in a single `PR_LIST_FIELDS` constant that the live query uses, and the suite drives `collect_live()` against a **fake `gh` on PATH** and greps the recorded argv for the field. Fixing only the detector would have been inert: `pr.get("isDraft")` on an unrequested field is `None`, which is falsy, which is "not a draft".

A draft does not `continue` inside the detector. It returns as a SILENT observation, so the skip leaves a line:

```
$ python3 q-system/.q-system/scripts/linear-pipeline-health.py --dry
SILENT: PR #4 [green_but_draft] -- PR is green but still a DRAFT -- parked on purpose; auto-merge does not apply
[dry] would page 0 of 0 broken state(s)
```

That is the live repo through the real code path. The only thing this watcher would have said on merge day is now the only thing it stays quiet about.

**This fix buys silence, and here is the bill:** a PR left in draft forever *is* a stall, and this watcher will never say so. It is not a stall this watcher's action can fix (GitHub refuses `gh pr merge --auto` on a draft), so paging it would be a wrong cause plus an unrunnable command. The line in the log is the hedge.

### MAJOR 2 — `stranded_issue` is real now

`stranded_issue_findings(issues, pr_issue_ids, branch_names, now)` is pure over three inputs; `collect_live` feeds it from Linear (`state.type == started`), `gh pr list --state all`, and `gh api repos/{owner}/{repo}/branches`. An issue with either a branch or a PR is somebody's live work; only neither, past `STRANDED_ISSUE_HOURS`, is a dead holder. An issue Linear cannot date is skipped, never reported — unknown age must not render as old.

Your repro is now a check, generalized: the suite scans the module for emitted state literals and fails on any `BROKEN_STATES` row that no collector emits **and** that is not declared in the new `IN_PROCESS_PAGED_STATES = ("round_cap", "no_verdict", "worker_infra")`. Those three are paged by `converge.sh` / `linear-worker.sh` themselves, which you noted — that is now a declaration on the record rather than an accident, and any *new* unreachable state is RED.

Live: `stranded_issue` fires on nothing today (0 pages above, no `DEGRADED` line, so all three inputs were really read). ASK-223 itself is In Progress and correctly not stranded — it has both a branch and PR #40.

### MAJOR 3 — blind is not clear

`collect()` now returns `(observations, unobserved)`. Every collector that could not read its source names the states it could not rule on. `carry_forward()` keeps those ledger keys untouched, so the next healthy cycle sees the same key rather than a fresh one.

Your exact sequence is the check, plus the half that could have regressed:

```
break -> pages 1
break -> pages 1   (dedupe)
BLIND (unobserved: main_red) -> pages 1, prints DEGRADED
recover, main never fixed -> pages 1   <- was 2
observed empty, then break again -> pages 2   <- transition semantics survive
```

**The bounded cost, stated:** if a breakage genuinely clears while we are blind and then re-breaks, the re-page waits for `RE_PING_TTL_SECONDS` (6h) instead of firing at once. One late page beats a stream of false ones.

The report no longer lies either: a blind cycle used to print the same `0 broken` a healthy one does. It now prints `DEGRADED: could not observe <states> -- prior findings carried forward, not cleared`.

### The layer above MAJOR 3, which is why it would have mattered on day one

`com.kipi.linear-pipeline-health.plist` ran `/usr/bin/python3` directly. launchd hands a job `PATH=/usr/bin:/bin:/usr/sbin:/sbin` — `gh` does not live there. Every collector would have returned `None` → all-degraded → **zero findings forever**, while `launchctl list` showed a healthy job exiting 0 every 15 minutes. The silent death this script exists to catch, reproduced inside the catcher. (Your repro noticed the symptom: `gh on PATH? -> None`.)

Three changes: the plist execs through `/bin/bash -lc` like `com.kipi.linear-dor` does; a **missing** binary is its own finding (`watcher_blind`) that pages once with the reinstall command, because unlike a failed call it is permanent and fixable; the suite pins the plist.

### MINOR 1 — the installer comment

Corrected, and the suite now fails on any `com.kipi.*` label named in `install-plist.sh` with no renderable template. `com.kipi.launchd-health` is loaded on the founder's machine from no committed source → `sp-ff4f494a`.

### MINOR 2 — the incident this issue opens with

You are right that no state here can observe a triage run: it leaves no PR, no branch, no converge line. The only place that failure is visible is inside the process that has it, so `linear-triage.py` now pages on the way down — naming that whatever it already commented or **closed stays that way**, plus the dry-run command to see what is left. The check runs it against a closed local port (`KIPI_LINEAR_API_URL=http://127.0.0.1:1/graphql`), so Linear is never touched.

Re-run `sp-b5dcf944`'s exact failure after this and it pages once, with an action.

### NIT

`q-system/output/ask223-read.py` deleted. `page_text(state, ...)` lost the unused arg.

---

### Not regressed (your "attacks that failed")

The `dead_converge` regex, the verdict-record field names, `live_converge_issues`' `--issue` form, `_minutes_since`'s DST behaviour, the Gap-1 `Do: ` lint and its `FOUND -gt 0` self-guard, and the `--dry` read-only path are all untouched. `--dry` is still verified read-only (writes no state, sends no page), now including the live run above.

### Still open, captured not mentioned

`sp-ba3552fc` — `DEGRADED` prints but never pages, so a watcher blind for many consecutive cycles (expired token, Linear outage) stays quiet. A consecutive-degraded counter is the fix; it is new paging behaviour with its own false-positive surface, so it is a separate issue rather than a same-PR patch.

assafkip:
Capability gate (was still running when the response above was posted):

```
tests: ran=79 quarantined=0 skipped-skeleton-only=0
capability-gate: GREEN
```
exit 0. No new quarantines, no skips.

assafkip:
Hit the tool-call ceiling. Dropping the one finding whose repro I could not execute, per the standing rule.

# Review: PR #40 (ASK-223)

Read the code first, PR comments after. Round-3's six findings are all genuinely closed; I re-ran two of them myself and they hold. My one major is on code that **round 3's fix introduced**, not on unchanged code.

---

## MAJOR — `stranded_issue` is reachable now, and still cannot fire: `_minutes_since` rejects every real Linear timestamp

`q-system/.q-system/scripts/linear-pipeline-health.py:245`

Linear's GraphQL emits fractional seconds. `_minutes_since` parses with `%Y-%m-%dT%H:%M:%S%Z` after `.replace("Z","UTC")`, which does not match `.412`. It returns `None`, and `stranded_issue_findings` skips every issue with an unparseable date by design (line 426). The detector is dead against the only producer that feeds it.

The suite cannot see this: its fixture builds timestamps with `time.strftime("%Y-%m-%dT%H:%M:%SZ", ...)`, a shape nothing in the pipeline emits for this field. GitHub's `updatedAt` has no fraction, which is why `green_not_merged` works and this does not.

**Live producer shape, read from Linear just now:**

```
$ python3 .pr40rev2/repro-linear.py
linear-sync.py exists: True
graphql signature: (query: str, variables: dict) -> dict
LIVE RESULT: in-progress issues = 11
    {'identifier': 'ASK-226', 'startedAt': '2026-07-28T03:41:31.232Z', ...}
    {'identifier': 'ASK-207', 'startedAt': '2026-07-27T19:50:40.412Z', ...}
    {'identifier': 'ASK-192', 'startedAt': '2026-07-27T17:23:44.011Z', ...}
```

**The detector against that shape:**

```
$ python3 .pr40rev2/repro-stranded.py .pr40rev2/health.py
FIXTURE shape : 2026-07-28T09:47:00Z
  _minutes_since -> 540
  findings       -> [{'state': 'stranded_issue', 'subject': 'ASK-300', ...}]

REAL Linear shape: 2026-07-27T19:50:40.412Z
  _minutes_since -> None
  findings       -> []

VERDICT: DETECTOR IS DEAD

GitHub updatedAt shape (from live gh pr list): 2026-07-28T18:43:12Z
  _minutes_since -> 3
```

**The citation I am falsifying.** Round 3's answer says: *"Live: `stranded_issue` fires on nothing today (0 pages above, no `DEGRADED` line, so all three inputs were really read)."* The inputs were read. They were then discarded by the parse. Zero findings with no `DEGRADED` line is the exact output shape this bug produces, so that evidence cannot distinguish "nothing is stranded" from "nothing can ever be found". 11 issues are In Progress, several over 24h old.

At 3am: an agent dies holding an issue, the watcher prints `paged 0 of 0 broken state(s)`, and the founder reads a clean cycle. Recoverable with a one-line fix, nothing permanent is written, so major and not blocker.

Fix: parse with `datetime.fromisoformat` (Python 3.11+ takes the `Z`), or strip the fraction before `strptime`. Then re-shape the suite's fixture to `...T%H:%M:%S.mmmZ` so the test can fail on this.

---

## MINOR — `unreviewed_head` costs one sequential `gh pr view` per verdict record, and the record set only grows

`q-system/.q-system/scripts/linear-pipeline-health.py:514`

One network call per file in `~/.config/kipi/pr-reviews/`. Nothing prunes that directory; it gains a file per PR forever.

```
$ python3 -c "import pathlib;p=pathlib.Path.home()/'.config/kipi/pr-reviews';print('records:',len(sorted(p.glob('pr-*.verdict.json'))))"
records: 27

$ time python3 .pr40rev2/health.py --dry
SILENT: PR #4 [green_but_draft] -- PR is green but still a DRAFT ...
DEGRADED: could not observe stranded_issue ...
[dry] would page 0 of 0 broken state(s)
python3 .pr40rev2/health.py --dry  1.81s user 1.23s system 18% cpu 16.641 total
```

16.6s per cycle at 27 records, each with a 60s timeout ceiling. Bounded today, unbounded in shape. Skipping records whose PR is merged or closed would cap it.

---

## Dropped: one finding I could not execute

`stranded_issue_findings` matches branches by substring over a single joined blob (line 418/430), so `ASK-22` would be masked by a branch named `sana/ask-223`. I wrote the repro and hit the session's tool-call ceiling before running it. Unexecuted, so it does not count. Worth a look while the parse fix is open, since both live in the same function.

## What is sound

- **`ps` truncation.** I expected macOS `ps -Ao args=` to clip the `--issue ASK-nnn` off a long argv from a non-tty, which would make every live converge look dead. It does not: `live_converge_issues()` returned `{'ASK-208', 'ASK-223'}`, picking up both my fake and the real converge running on this machine.
- **Clock math**, re-derived independently of round 3. Correct at `PST/PDT`, `IST/IDT`, `CET/CEST` and `UTC`, at both a summer and a winter instant, at 1/5/19/21/45 minutes. `%Z` parsing `UTC` pins `tm_isdst=0`, which pins `mktime` to standard time.
- **Producer/consumer field census on the verdict record.** `pr-review-agent.sh:275` writes `pr`, `issue`, `verdict`, `head_sha`. The detector reads exactly those four. No mismatch.
- **`dead_converge` regex vs the real producer.** `converge.sh:146` writes `converge[ASK-nnn] round N/M dispatching Sana`. `DISPATCH_LINE` matches it, and `converge.sh:38` writes to the same `linear-worker.log` the collector reads.
- **`Do: ` lint coverage.** Every `$NOTIFY` use in both scripts is a `bash "$NOTIFY"` call site, so the lint's grep has no blind spot. Only the two variable assignments fall outside it.
- **First live cycle.** The dry run above emits zero pages and correctly routes PR #4 to `SILENT green_but_draft`. Round 3's MAJOR 1 is genuinely dead.

## On the review process

Rounds 1 to 3 found 3 majors and 3 minors, all fixed, and my independent read confirmed the two I could re-test. The one major here is on the collector round 3's fix added, not a new blocker on unchanged code. Calibration looks right; this is not the pathology where earlier rounds under-read the diff.

## VERDICT: REQUEST CHANGES

**Fix first:** the timestamp parse at `linear-pipeline-health.py:245`, and re-shape the suite's `stranded_issue` fixture to carry fractional seconds so the test can go red on it.

I created `.pr40rev2/` (untracked, matching the existing `.pr##rev` convention) for repro scripts. No tracked file was touched.

FINDINGS:
major|_minutes_since cannot parse Linear's fractional-second timestamps, so stranded_issue_findings skips every real issue and the detector never fires in production|q-system/.q-system/scripts/linear-pipeline-health.py:245
minor|unreviewed_head runs one sequential gh pr view per verdict record in a directory that is never pruned, so cycle cost grows without bound (27 records = 16.6s today)|q-system/.q-system/scripts/linear-pipeline-health.py:514
END FINDINGS

assafkip:
## Round-4 review answered: 2 findings + the dropped one, 3 fixes, 3 new checks

`PASS: 25/25` (was 21/21). `capability-gate: GREEN`, rc 0, 79 tests, no new quarantines.

| finding | verdict | check that goes red without the fix |
|---|---|---|
| MAJOR `_minutes_since` rejects Linear's fractional seconds | fixed | check 14 |
| MINOR `unreviewed_head` cost grows with the record count | fixed | check 16 (two halves: cost + detection survived) |
| DROPPED (unexecuted) branch substring masks a shorter id | fixed — it is real, I ran your repro | check 15 |

---

### MAJOR — the parse

You were right, and right about why the suite could not see it: its fixture built `%Y-%m-%dT%H:%M:%SZ`, a shape nothing produces for `startedAt`. The fixture now carries the fraction, so check 9 would have gone red too.

Observed red against `HEAD` (the commit you reviewed), then green:

```
=== vs REVIEWED COMMIT ===
CHECK 14 body (asserts minutes==540 and found==1 for every shape):
   linear  minutes=None   found=0
   github  minutes=540    found=1
   micros  minutes=None   found=0

=== vs FIX ===
   linear  minutes=540    found=1
   github  minutes=540    found=1
   micros  minutes=540    found=1
```

I did **not** take the `datetime.fromisoformat` route you suggested. `fromisoformat` only accepts `Z` on 3.11+, and this ships fleet-wide through `kipi update` to machines whose python I do not control. The fraction is stripped before `strptime` instead, which leaves the `%Z`/`mktime`/`time.timezone` arithmetic byte-identical to the version you and round 3 independently verified DST-safe across PST/PDT, IST/IDT, CET/CEST. The regex is anchored on what follows the fraction (`[Z+-]` or end of string) so it cannot eat a date component. Check 14 also pins that garbage still returns `None` — the skip-undateable behaviour you correctly identified as load-bearing is unchanged.

**What the operator now sees on the first live cycle — this is the part worth arguing about:**

```
$ python3 q-system/.q-system/scripts/linear-pipeline-health.py --dry
SILENT: PR #4 [green_but_draft] -- PR is green but still a DRAFT ...
BROKEN: pipeline ASK-207: ASK-207 has been In Progress for 23h with no branch and no PR. ... Do: kipi linear claims, ...
BROKEN: pipeline ASK-198: ... 24h ...
BROKEN: pipeline ASK-196: ... 25h ...
BROKEN: pipeline ASK-192: ... 25h ...
[dry] would page 4 of 4 broken state(s)
```

Four pages on install, not zero. That is the inverse of round 3's MAJOR 1 and it deserves to be named rather than buried: your review said 11 issues were In Progress with several over 24h, and this is what four of them look like once the detector can read a date. Each is a distinct real breakage and each pages once, so this is the dedupe working, not a burst — but if any of the four turns out to be a live agent the collectors cannot see, that is a false page and the argument for reverting the detector rather than tuning it.

### MINOR — the per-record `gh pr view`

Fixed by removing the calls, not by pruning. `headRefOid` joined `PR_LIST_FIELDS`, so the heads come out of the open-PR list `green_not_merged` already pays for. A record whose PR is absent from that list is merged or closed, which is the same outcome the old `state != OPEN` skip gave, for one call instead of one per record.

Observed red, then green, 30 records through the real `collect_live()` against a fake `gh` that logs argv:

```
BEFORE  verdict records: 30   total gh invocations: 34   of which 'pr view': 30
AFTER   verdict records: 30   total gh invocations:  4   of which 'pr view':  0
```

Live cycle time: **16.6s → 3.2s**, with more work in it (the stranded collector now actually completes).

**The layer above my own fix.** Batching bought O(1) network and sold a truncation window the per-record loop did not have: at exactly `PR_LIST_LIMIT` results the page may be clipped, and a PR past the edge would read as "not open" — silently undetectable rather than merely slow. So a full page now marks **both** list-fed states `unobserved`, which carries the ledger forward and prints `DEGRADED` instead of claiming a clean cycle. Second half of check 16 asserts it with 100 fake open PRs.

**The bill, stated:** on a repo with 100+ permanently-open PRs this watcher would sit degraded and never page on those two states. That is the honest failure mode (it says it is blind) rather than the quiet one, and this repo has 2.

### DROPPED finding — I ran it, it is real

`ASK-22` vs `sana/ask-223`:

```
branches      : ['sana/ask-223', 'main']
stranded issue: ASK-22 (no branch of its own, no PR, 9h old)
findings      : []          <- vs reviewed commit
VERDICT       : MASKED - detector silent
```

Now `branch_issue_ids()` extracts whole ids with the same `ISSUE_ID` regex the PR half uses, and membership is a set test. Both halves of "did this work reach anywhere" agree on what an id is. Check 15 pins the masking case and the extraction.

### What I did not regress

The `dead_converge` regex, the verdict-record field census, `live_converge_issues()`' `--issue` form, the `_minutes_since` clock math (unchanged arithmetic, only the input is normalised), the `Do: ` lint and its `FOUND -gt 0` self-guard, the draft skip and its SILENT line, the blind-vs-clear ledger semantics, and the `--dry` read-only path. The live `--dry` above wrote no state file and sent no page.

### Captured, not mentioned

`sp-67755a60` — `~/.config/kipi/pr-reviews/` still grows a file per PR forever. My fix removed the network cost but not the disk growth, and every cycle still reads every file. A retention policy is a separate change with its own "which records are safe to drop" question.

`sp-ba3552fc` (from round 3) is still open: `DEGRADED` prints but never pages, so a watcher blind for many consecutive cycles stays quiet.

assafkip:
Review complete. Three reproducers executed; prior rounds read only after I had my own read.

## Findings

---

### 1. `major` — `green_not_merged` never reads the verdict ledger it already has open, so a PR the reviewer REJECTED pages with "Do: gh pr merge --auto --squash"

`q-system/.q-system/scripts/linear-pipeline-health.py:307`

The detector gates on four things: rollup all-SUCCESS, no `autoMergeRequest`, age past 20 min, not a draft. There is no approval test. Its docstring asserts the opposite invariant ("A PR whose required checks are all green has nothing left to do but merge", `:288`) — the code does not support that claim.

`collect_live` loads `~/.config/kipi/pr-reviews/pr-*.verdict.json` at `:557`, **31 lines after** it calls the green detector at `:549`. The verdict is on disk, in the same function, and the green half never looks at it. `reviewDecision` is not in `PR_LIST_FIELDS` either (`:67`), so neither source of approval state reaches the detector.

**Repro** (fake `gh` on PATH, temp state dir, no network):

```
bash /Users/assafkipnis/projects/kipi-system/.pr40rev3/repro.sh
```

**Real output:**

```
############ REPRO 1: green_not_merged ignores the verdict ledger ############
on-disk verdict for PR #77 : REQUEST CHANGES (head deadbeef, == current head)
observed  : green_not_merged  broken=True
PAGE SENT : pipeline PR #77: PR #77 has every required check GREEN and has not merged
            for 59 min -- auto-merge was never armed, so it sits green forever. This is
            the stall that looks most like success. Do: gh pr merge --auto --squash 77
unobserved: []
```

The verdict-record shape is the author's own (`test-linear-pipeline-health.sh:556` writes the identical keys); `REQUEST CHANGES` is the value `converge.sh:274` prints and `SILENT_STATES` names.

3am consequence: Skeleton Validation is this repo's only required check. Arming auto-merge on a green PR lands it immediately. The operator is told by their own pager to squash-merge code a reviewer rejected. Recoverable by revert, so major, not blocker.

Live today: 0 of 7 open PRs hit it — but PR #34 is `green=False armed=False draft=False` right now and becomes exactly this shape the moment validate passes.

---

### 2. `minor` — three of the four SILENT_STATES have no producer, and the reachability test only guards the BROKEN half

`q-system/.q-system/scripts/linear-pipeline-health.py:120` (test at `test-linear-pipeline-health.sh:318`)

CHECK 9 was added in an earlier round for exactly this defect class — a declared state nothing emits. It runs the census over `BROKEN_STATES` only. Run the same census over `SILENT_STATES` and three rows fall out. The PR body's silence table promises all three ("PR waiting on review | no, logged"; "REQUEST CHANGES on a round | no, logged"; "issue skipped, no DoR | no, logged"). None can occur.

**Repro:** same script, section 2. **Real output:**

```
############ REPRO 2: three SILENT_STATES have no producer ############
CHECK 9 guards BROKEN_STATES only: [] (passes)
same census over SILENT_STATES  : ['awaiting_review', 'request_changes', 'skipped_no_dor']
```

This is the mechanism under finding 1: a green PR awaiting review has no `awaiting_review` row to land in, so it lands in `green_not_merged` instead.

---

### 3. `minor` — the stranded collector's two inputs have no clipped-page guard, unlike the open-PR list

`q-system/.q-system/scripts/linear-pipeline-health.py:612`

`green_not_merged` and `unreviewed_head` degrade at `len(prs) >= PR_LIST_LIMIT` (`:550`, `:566`). The stranded collector's `gh pr list --state all --limit 100` (`:609`) and `branches?per_page=100` (`:611`) are checked for `None` only. At exactly 100 either list is page 1 of N, an issue whose PR or branch fell past the edge reads as "no branch and no PR", and it pages as a dead holder with a clean cycle line.

**Repro:** same script, section 3. **Real output:**

```
############ REPRO 3: stranded_issue has no clipped-page guard ############
gh returned EXACTLY 100 PRs (--state all) and EXACTLY 100 branches: both clipped
stranded findings : ['ASK-500']
unobserved set    : []
'stranded_issue' marked unobserved? False
```

Severity dropped a level per the calibration rule: this repo has 39 PRs and 39 branches, so I had to construct 100 of each. No producer emits it today.

---

## What is sound

- **Fixture-vs-producer on the timestamp path.** I ran the watcher `--dry` against the founder's real Linear and real repo. `stranded_issue` fires on four genuine issues, and I checked all four independently: `ASK-207/198/196/192` have zero PRs (title, body, or head ref, across all 39) and zero branches. True positives, not fixture artifacts.
- **Producer shape of `dead_converge`.** `converge.sh:60` is `say() { echo "$(TS) converge[$ISSUE] $*" | tee -a "$LOG" }` writing to `$STATE_DIR/linear-worker.log` — byte-compatible with `DISPATCH_LINE` at `:329`. The detector reads what the producer writes.
- **The `Do:` lint's coverage.** I enumerated every `NOTIFY` call in both scripts; all nine use the `bash "$NOTIFY"` form the lint greps for, including the variable-built stall page at `converge.sh:266`, which is deliberately concatenated so the literal sits on the linted line. No page escapes the grep.
- **The network-isolation claim.** `linear-sync.py:323/336` honors `KIPI_LINEAR_API_URL` and `KIPI_LINEAR_API_KEY`, so CHECK 13 really does point at a dead local port and never reaches Linear with the founder's key.
- **`--dry` is genuinely read-only.** I ran it live against the real state dir; no state file written, no page sent.
- **Suite baseline.** `PASS: 25/25` against the PR tree, materialized separately at `.pr40rev3/repo`.

Dropped after trying and failing to reproduce:

- **`main_red` reads `gh run list --branch main --limit 1` with no workflow filter.** I expected an unrelated workflow's success to mask a validate failure. This repo has exactly one workflow (Skeleton Validation); 12 consecutive runs confirm it. No producer, dropped.
- **The 6h re-ping drip.** Four live findings × 4 pages/day, indefinitely. I reproduced it (96-cycle simulation, 4 pages per unfixed finding per 24h), then dropped it: it is a stated design decision, deliberately matched to `launchd-health-check.py`'s `FAIL_PING_TTL_SECONDS`, and the operator can clear each one. Re-litigating a documented tradeoff is not a finding.
- **Concurrency.** launchd keeps one instance per label, so `StartInterval 900` cannot overlap itself. `write_state` is non-atomic, but the worst case is a truncated ledger read as `{}`, which costs one re-page per live finding. Below the bar.

## On the review process

Finding 1 is on **unchanged code**. `green_not_merged_findings` has existed since the first commit on this branch, and round 1 filed a major about that exact function's inputs (`isDraft`, anchored at `:211`) — it audited what the detector reads and stopped at draft state without asking whether "green" implies "reviewed". Four rounds and 8 findings later, the function's central claim is still unchecked. That is the miscalibration signal: rounds 1-3 went deep on the input *plumbing* of this detector and never questioned its *predicate*.

I want to be fair about the other direction, though: round 2's `_minutes_since` major was not miscalibration. It landed on code that round 1's own fix had just made reachable for the first time. That one was earned.

## VERDICT: REQUEST CHANGES

Fix first: **finding 1**. The verdict record is already in hand at `linear-pipeline-health.py:557` — thread it into `green_not_merged_findings` and route a non-`APPROVE` verdict to `request_changes`, which is a `SILENT_STATES` row that already exists and currently has no producer. That closes findings 1 and 2 in one change.

Findings 2 and 3 do not gate on their own.

FINDINGS:
major|green_not_merged never reads the verdict ledger collect_live already opens, so a PR with REQUEST CHANGES on disk pages "Do: gh pr merge --auto --squash <n>"|q-system/.q-system/scripts/linear-pipeline-health.py:307
minor|three of four SILENT_STATES have no producer and the CHECK 9 reachability census guards only the BROKEN half, so the PR body's silence table documents rows that cannot occur|q-system/.q-system/scripts/linear-pipeline-health.py:120
minor|the stranded collector's --state all PR list and branches page have no clipped-page guard, unlike the open-PR list, so at exactly 100 results an issue past the edge pages as a dead holder on a cycle that claims to be clean|q-system/.q-system/scripts/linear-pipeline-health.py:612
END FINDINGS

Repros live in `.pr40rev3/` (untracked, same pattern as prior rounds). Nothing in the repo was modified.

assafkip:
Closing as a redrive. ASK-223 stays In Progress and the branch is preserved.

State: CONFLICTING with main, 88 commits behind, opened 2026-07-28. `kipi/reviewer-approved` is also failing.

The work is still wanted and is not on main: `linear-pipeline-health.py`, its plist, and its test do not exist there. An outside watcher that catches the deaths no in-process pager can see is a real gap, and it is the correct shape for it.

The blocker is mechanical, not editorial: this branch touches `linear-worker.sh`, `converge.sh`, `linear-triage.py` and `install-plist.sh`, all of which were heavily reworked on main across those 88 commits. It cannot be updated without a conflict resolution, and I will not resolve those blind.

Redrive off current main under ASK-223. Branch `sana/ask-223` keeps the implementation.

PR #41, opened 2026-07-28
fix(ops): the round cap was one short, so issues died with one major left (ASK-113)

assafkip:
Two issues capped out at 3 rounds today. The review histories rule out both obvious explanations.

### Measured, from the reviews on disk

**No finding was ever repeated across rounds, on either issue.** PR #40:

| round | findings |
|---|---|
| 1 | draft PRs · stranded_issue · transient gh failure · 2 doc minors |
| 2 | fractional-second timestamps · sequential gh calls |
| 3 | verdict ledger unread · SILENT_STATES · clipped-page guard |

Different lines, different bugs, every round — the agent **was** fixing what it was told. The reviewer kept finding new, deeper problems underneath.

**Majors declining to zero is the convergence signal.** PR #39, the one that converged:

| round | result |
|---|---|
| 1 | 2 majors + 1 minor |
| 2 | 2 majors + 1 minor |
| 3 | **0 majors** + 3 minors → APPROVE WITH NITS → merged |

Minors never block. PR #40 was at **one major** when the cap stopped it.

### Why this costs nothing

A cap-out is the most expensive outcome available: it spends the full cost of the issue and produces nothing mergeable. Three issues that land beat four that die one round short.

```
3 rounds x 2 sessions x 4 issues = 24
4 rounds x 2 sessions x 3 issues = 24
```

Both dials move together. The old comment hardcoded the cap into its arithmetic ("= 6 sessions"), which is how the two drift apart; it now names the relationship.

### Checks
```
plutil -lint com.kipi.dispatch.plist   OK
bash -n kipi-dispatch.sh               OK
capability-gate.py                     GREEN, ran=78, exit 0
```

### DIFF
--- a/kipi-dispatch.sh
+++ b/kipi-dispatch.sh
@@ -103,8 +103,11 @@ fi
 # runs. An unbounded heartbeat is a runaway-bill loop, which is exactly the
 # thing loop-exits.md says an autonomous loop must not be.
 #
-# One issue costs up to MAX_ROUNDS x (1 agent + 1 reviewer) = 6 sessions.
-# So DAILY_MAX is roughly "sessions per day / 6".
+# One issue costs up to MAX_ROUNDS x (1 agent + 1 reviewer) sessions, so at the
+# current cap of 4 that is 8, and DAILY_MAX is roughly "sessions per day / 8".
+#
+# The two dials move TOGETHER. Raising the round cap without lowering DAILY_MAX
+# raises the bill; the plist sets 4 rounds x 3 issues to hold it flat at 24.
 DAILY_MAX="${KIPI_DISPATCH_DAILY_MAX:-4}"
 # The budget day starts at RESET_HOUR LOCAL, not at midnight and not at UTC.
 # Founder-set 2026-07-28, and the reasoning is safety, not tidiness:
--- a/q-system/.q-system/scripts/com.kipi.dispatch.plist
+++ b/q-system/.q-system/scripts/com.kipi.dispatch.plist
@@ -37,6 +37,31 @@
 		       launchctl unload ~/Library/LaunchAgents/com.kipi.dispatch.plist
 		       launchctl load   ~/Library/LaunchAgents/com.kipi.dispatch.plist -->
 		<key>KIPI_DISPATCH_DAILY_MAX</key>
+		<string>3</string>
+
+		<!-- ROUND CAP: 4, raised from 3 on 2026-07-28, and DAILY_MAX dropped
+		     4 -> 3 in the same breath so the spend is UNCHANGED:
+		         3 rounds x 2 sessions x 4 issues = 24
+		         4 rounds x 2 sessions x 3 issues = 24
+
+		     WHY, measured rather than guessed. Two issues capped out at 3
+		     rounds the same day. Their review histories say the cap was the
+		     problem, not the work:
+
+		       - NO finding was ever repeated across rounds, on either issue.
+		         Every round surfaced different lines and different bugs, so
+		         the agent WAS fixing what it was told; the reviewer kept
+		         finding new, deeper problems underneath.
+		       - Majors decline toward zero, and that is the convergence
+		         signal: PR #39 went 2 majors -> 2 majors -> 0 majors, and the
+		         third round returned APPROVE WITH NITS and merged. Minors
+		         never block.
+		       - PR #40 was at ONE major when the cap stopped it.
+
+		     A cap-out is the most expensive possible outcome: it spends the
+		     whole cost of the issue and produces nothing mergeable. Three
+		     issues that LAND beat four that die one round short. -->
+		<key>KIPI_DISPATCH_ROUNDS</key>
 		<string>4</string>
 
 		<!-- WHEN THE ALLOWANCE REFILLS, local hour. Founder-set 2026-07-28,


assafkip:
# Review: PR #41 (`sana/round-cap`)

The diagnosis is well-evidenced and the arithmetic checks out. The wiring does not: the change reaches the running loop only through a step the file documents wrong, and the one operator-facing number that encodes the old cap was left hardcoded.

---

## Findings

### 1. MAJOR — the plist edit never reaches the running job, and the file documents a procedure that cannot apply it

`q-system/.q-system/scripts/com.kipi.dispatch.plist:37-38`

launchd reads `~/Library/LaunchAgents/com.kipi.dispatch.plist`, which is a **rendered** copy written only by `install-plist.sh`. This PR edits the **template**. The remediation block right above the key it edits tells the operator to `launchctl unload` + `load` the rendered copy, which re-reads a file the template edit never touched. Both sibling plists document `install-plist.sh` instead.

```
$ bash .pr41rev/repro-c.sh
=== 1. render the template as it stands on main (= what a machine
       installed before this PR has in ~/Library/LaunchAgents) ===
KIPI_DISPATCH_ROUNDS in the rendered/installed copy: Print: Entry, ":EnvironmentVariables:KIPI_DISPATCH_ROUNDS", Does Not Exist
KIPI_DISPATCH_DAILY_MAX in the rendered/installed copy: 4

=== 3. who writes ~/Library/LaunchAgents/com.kipi.dispatch.plist? ===
install-plist.sh:80:PLIST="$HOME/Library/LaunchAgents/$LABEL.plist"
callers of install-plist.sh for this label, in tracked files:
  (none)

=== 4. what each plist tells the operator to do after an edit ===
com.kipi.dispatch:       documents launchctl unload/load of the RENDERED copy  (no-op for a template edit)
com.kipi.fleet-health:   documents install-plist.sh  (re-renders: correct)
com.kipi.linear-dor:     documents install-plist.sh  (re-renders: correct)
```

Not a regression: merged as-is, nothing breaks, the loop runs exactly as it does today. That is the problem. It keeps capping at 3 and stranding issues with one major left, which is the failure this PR exists to fix, while `kipi-dispatch.sh:109` now asserts in a code comment that "the plist sets 4 rounds x 3 issues to hold it flat at 24." The next cap-out gets misdiagnosed because the comment says the cap was already raised.

Fix: `bash q-system/.q-system/scripts/install-plist.sh com.kipi.dispatch`, and repoint the plist's remediation comment at the installer like its two siblings.

---

### 2. MINOR — the daily-cap page understates the spend ceiling by 25%

`kipi-dispatch.sh:146`

`$((DAILY_MAX * 6))` hardcodes 3 rounds x 2 sessions. The commit message says the fix removes exactly this drift ("the old comment hardcoded the cap into its arithmetic, which is how the two drift apart"). The comment got fixed; the live arithmetic the founder reads on their phone did not.

```
$ bash .pr41rev/repro-a.sh
--- PR head, run with the NEW plist values: DAILY_MAX=3 ROUNDS=4 ---
--- what the founder's phone shows ---
PAGE >> kipi dispatch: hit the daily cap of 3 issues (~18 agent sessions). Not an error -- the loop is resting until 7am, then it picks up again on its own. Do: nothing, or raise KIPI_DISPATCH_DAILY_MAX in com.kipi.dispatch.plist to go faster.
--- truth: 3 issues x 4 rounds x 2 sessions = 24 sessions ---
```

Fix: `$((DAILY_MAX * MAX_ROUNDS * 2))`. Both dials are already in scope at that line.

---

### 3. MINOR — comment says the cap is 4; the script default is still 3

`kipi-dispatch.sh:41` vs `kipi-dispatch.sh:107`

`MAX_ROUNDS="${KIPI_DISPATCH_ROUNDS:-3}"` sits 66 lines above a comment reading "at the current cap of 4 that is 8, and DAILY_MAX is roughly sessions per day / 8". Any run without the launchd env (a manual `bash kipi-dispatch.sh`, any machine not re-rendered, any fleet instance) dispatches at 3.

```
$ bash .pr41rev/repro-b.sh
=== no KIPI_DISPATCH_* env, i.e. the LaunchAgent was never re-rendered ===
--- dispatch log ---
2026-07-28T19:24:50Z dispatching ASK-999 (live=0 cap=2 rounds=3 budget=1/4)
--- argv actually handed to converge ---
CONVERGE ARGV >> converge --issue ASK-999 --max-rounds 3
```

Fix: move the defaults to `:-4` / `:-3` so the code and the comment agree, or word the comment as the plist's value rather than "the current cap."

---

## What is sound (attacks that failed)

- **"×2 sessions per round" is fabricated.** It is not. One `claude -p` in `linear-worker.sh:1019`, one in `pr-review-agent.sh:235`, and `converge.sh:141` runs exactly one worker call per round. The arithmetic holds.
- **The 4th round is off-by-one and never runs.** `ROUND=0; while [ "$ROUND" -lt "$MAX_ROUNDS" ]` with `ROUND=$((ROUND + 1))` first in the body gives rounds 1..4. Correct.
- **The plist is malformed after the insert.** `test-install-plist.sh` passes all 15 cases; renders placeholder-free and plutil-clean. The `--` inside the XML comments does break strict parsers (`plistlib` throws at line 18), but that text is pre-existing and unchanged here, and nothing in the repo strict-parses these files: `launchd-health-check.py` uses `plist.stem` plus `launchctl`, never a parser.
- **Docs elsewhere still claim 4 issues/day.** Grepped `AUTONOMOUS-SYSTEMS.md` and every tracked `.md`. No doc states the dial values, so lowering `DAILY_MAX` strands nothing.
- **Concurrency.** `KIPI_DISPATCH_MAX=1` and the `ps`-snapshot duplicate guard are untouched; two heartbeats overlapping behave as before.

One observation, not a finding (no reproducer exists for it): "flat at 24" is a worst-case identity, not an expected-cost one. Most issues converge in 1-2 rounds, so the 4th round is only paid on the tail, while the `DAILY_MAX` 4→3 cut is paid every single day. The expected trade is "25% fewer issues/day, always" for "one extra round, sometimes." Given two cap-outs in one day that may still be the right call, and it is a founder dial either way.

---

## VERDICT: REQUEST CHANGES

One major, so changes rather than approve. Merging as-is causes no damage and nothing is unrecoverable, so this is not a block.

**Fix first:** run `bash q-system/.q-system/scripts/install-plist.sh com.kipi.dispatch` and repoint the plist's reload comment at that command. Without it the round cap stays 3 and the new comment claims otherwise.

Reproducers live in `.pr41rev/` (untracked scratch, same convention as the existing `.pr39rev/` / `.pr40rev/` dirs). No tracked file was modified.

FINDINGS:
major|The plist template edit never reaches the running LaunchAgent, and the file's own reload instructions re-read the rendered copy so they cannot apply it|q-system/.q-system/scripts/com.kipi.dispatch.plist:37
minor|The daily-cap page hardcodes 6 sessions per issue, understating the new ceiling by 25% (says ~18, truth 24)|kipi-dispatch.sh:146
minor|Comment asserts a round cap of 4 while the script still defaults KIPI_DISPATCH_ROUNDS to 3|kipi-dispatch.sh:41
END FINDINGS

assafkip:
Closing: this is cost tuning whose inputs have expired.

The change is two dials: round cap 3 to 4, and DAILY_MAX 4 to 3, held flat at 24 sessions/day. The reasoning in the plist comment is measured and was sound at the time (no finding repeated across rounds; majors declining toward zero; PR #40 stopped at one major).

Two reasons not to merge it now:

1. **CONFLICTING, 88 commits behind.** `kipi-dispatch.sh` has been rewritten on main since (preflight, round-robin, per-repo locking, and a changed cost model). Neither dial can be updated without resolving conflicts against that rewrite.
2. **The cost model it balances against has changed.** Main's dispatch does more work per tick, and a founder-directed jobs audit on 2026-08-01 (eb43899) re-pinned model selection for dispatch. "4 rounds x 3 issues = 24" was derived against the old per-issue cost and should be re-derived, not merged.

For the record, neither dial is on main today: main still has `MAX_ROUNDS` defaulting to 3 and the plist still sets DAILY_MAX 4 with no `KIPI_DISPATCH_ROUNDS` key. If the cap-out problem is still real, it is a two-value change against current main plus fresh evidence.

ASK-113 (the epic) stays In Progress. Branch `sana/round-cap` is preserved.

PR #33, opened 2026-07-28
feat(worker): every PR the worker touches arms its own auto-merge (ASK-222)

assafkip:
Closes the last human step in the unattended path: **who arms auto-merge.**

## The gap

`kipi/reviewer-approved` is a REQUIRED context, watched refusing on ABSENT and on FAILURE (PRs #27, #30), and PR #30 merged itself at 01:38:07Z with no human once auto-merge was armed. The arming itself was a hand-typed `gh pr merge --auto --squash <n>` plus a watcher loop inside an interactive session. **Both die when the terminal closes**, so a PR opened after that sits green forever with nobody left to merge it.

## The change

`linear-worker.sh` arms auto-merge at the point `$PR_NUM` resolves, which covers **both** paths (the PR the agent opened, and the PR the worker opened because the agent did not), and does it **before** the review.

`--auto` is not "merge now": GitHub holds the PR until every required context is green, and `kipi/reviewer-approved` is absent until the reviewer posts it. Arming afterwards would need something to come back once the review lands, which is the same gap in a different coat.

The worker still never merges. GitHub merges, only once the required checks pass. The four refusals are intact.

| Case | Behaviour |
|---|---|
| Already armed | Real no-op. State is asked for via `gh pr view --json autoMergeRequest`, not armed-and-forgiven, so a rework round makes no call, emits no warning, and does not depend on which `gh` version's exit code is installed. |
| Arm refused | LOUD, not fatal. An unarmed PR is invisible by construction (everything green, nothing merges, no signal), so it is said, the PR stands, the review still runs, exit code unchanged. |
| No PR | No arm call attempted. |

## Blast radius (stated at the call site)

With this line in place, code reaches `main` with no human in the path, on a repo that fans out fleet-wide through `kipi update`. The only thing between a diff and main is the set of REQUIRED contexts on the branch. Remove `kipi/reviewer-approved` and this becomes an unreviewed-merge machine. That is why it was made required first and watched refusing twice before this issue was filed.

## Check

`test-severity-floor.sh` section Q, 10 new checks, extended in place (no new capability-manifest entry). The `gh` stub now logs every call and the reviewer stub logs into the same file, so ORDER is assertable.

**RED observed first** - the call log showed the review running with no `pr merge` anywhere:

```
pr list --head sana/ask-aaa --json number -q .[0].number
pr view 830 --json mergeStateStatus -q .mergeStateStatus
pr view 830 --json headRefOid -q .headRefOid
[...]

PR #25, opened 2026-07-27
fix(worker): the rework gate reads mergeability, so an approved PR that later breaks is dispatchable (ASK-212)

assafkip:
Resolves ASK-212 / spillover `sp-71b63e62`. Re-file of fix 2 of 4 from ASK-208 (PR #22, closed unmerged at the round cap) — the mergeability hunk only.

## The defect

`rework_gate` decided "is there work to do here" from the stored verdict alone. An approved PR that **later** stops merging was invisible to the loop.

Observed 2026-07-27: PR #11 approved 06:08Z, PR #16 landed 17:30Z and broke it. Both `converge` and a direct worker run then skipped #11 in under two seconds as *"waiting on founder merge only"*. The loop could not dispatch the one thing blocking the merge, and a human had to diagnose it by hand.

## Three states, not two

Making APPROVE non-terminal opens an unbounded rework path, and Linear objects do not go away. So the conflict budget is **separate** from the review budget:

| verdict | merge state | action |
| -- | -- | -- |
| approving | CLEAN | exit 10, terminal, waits on merge authority |
| approving | DIRTY / BEHIND | exit **30**, dispatch a rebase round under its own cap |
| approving | DIRTY, at cap | stop, page **once**, stay terminal |

Only a stated DIRTY or BEHIND counts. BLOCKED (branch protection), UNSTABLE (a failing non-required check), UNKNOWN (GitHub still computing) and empty (gh failed) all keep the old terminal behaviour. Fail toward terminal: a missed conflict costs one human diagnosis; a manufactured one spends model budget on every healthy PR in the fleet at once.

`MAX_ATTEMPTS` could not serve as the cap — it only counts runs where `claude` exits non-zero, and the cited failure mode is an agent that exits 0 having done the wrong thing. The page is claimed in the same ledger write that reports it, so a stuck PR pages once, not once per scheduled cycle.

A conflict round does **not** get the review prompt. The review approved this diff; handing the agent "the review is the spec, answer every finding" against a review with no findings is how ASK-208's rounds 1 and 2 both did code polish while the conflict went untouched. The rebase prompt names the conflict, the round, and the cap, and forbids touching anything else.

`converge.sh` is deliberately **unchanged** — it calls `rework_gate` with one argument, and that form keeps its exact original semantics (pinned by a test case). Its own `MAX_ROUNDS=4` bounds it. The converge-side gap is captured as spillover, not smuggled in here.

## Reproducer, RED then GREEN

[...]

PR #43, opened 2026-07-29
fix(dispatch): a red PR gets picked back up -- rework candidates re-enter the loop (ASK-245)

assafkip:
Closes ASK-245 (do not merge from here; closeout runs through /issue-verify and /issue-closeout).

## The defect

`ready()` in `linear-worker.sh` returned `backlog`/`unstarted` only. An issue flips to In Progress the moment a worker takes it, so from that instant it excluded itself from every future heartbeat. `kipi-dispatch.sh` documents this as loop-exit 5 and it is correct as a **concurrency** guard; it doubled as a **permanent** one.

converge already knows how to rework a red PR -- that is what it does between rounds, using the review as the spec. The mechanism existed. It had no trigger once the run ended.

Measured 2026-07-29, 5 stranded:

```
#40 ASK-223  MERGEABLE    validate SUCCESS   kipi/reviewer-approved FAILURE
#36 ASK-225  CONFLICTING  validate FAILURE
#35 ASK-226  MERGEABLE    validate FAILURE   kipi/reviewer-approved FAILURE
#34 ASK-221  MERGEABLE    validate FAILURE
#23 ASK-210  MERGEABLE    validate FAILURE
```

## What changed

**`linear-worker.sh`** -- the picker publishes a separate `rework` set (owner:sana, not owner:assaf, has a DoR, state type `started`). Disjoint from the fresh set by construction: folding a started issue into `ready` would send it down the fresh-work path, which cuts a worktree from `origin/main` and would blow away the PR's commits (the `BASE` block). Announced in DRY mode only -- that is the seam `kipi-dispatch.sh` reads. The apply loop is untouched; a rework round is still reached the way it always was, through `--issue ASK-n` from converge. The `--issue` path is byte-identical to before.

**`kipi-dispatch.sh`** -- one `ps` snapshot taken before candidate selection and reused by both guards. Rework is a fallback, never a peer: fresh work wins outright, so the backlog cannot starve behind a PR that keeps going red. A candidate with a live converge is passed over rather than exiting, so one permanently-live run cannot hide every other stranded issue behind it. A rework dispatch fetches `origin` first.

**The rework prompt** now opens with a rebase onto current `origin/main`.

## Budget (loop-exit 7), stated

Rework spends the same `KIPI_DISPATCH_DAILY_MAX` allowance as fresh work; the daily cap is not raised. `MAX_ATTEMPTS` is applied **at the announcement**, not left to the work loop: leaving it there means a permanently-red PR spends a budget slot and a converge launch every heartbeat before the refusal fires. Case A6 pins that.

## Deviation from the DoR, and the evidence for it

The DoR says `kipi-dispatch.sh` "fetches and rebases onto `origin/main` first". It **fetches**. It does **not** rebase the worktree, because that would break the worker:

`position_tree_on_pr_head` refuses any tree holding commits reachable from neither `origin/<branch>` nor `origin/main` (`linear-worker.sh:528`). A rebase rewrites shas, so a dispatch-side rebase produces exactly that state -- `tree_holds_pr_head` goes false, `position_tree_on_pr_head` refuses with "the tree holds N commit(s) that exist nowhere else", and the issue stalls for a human. Two writers on one worktree is the defect class this repo keeps closing.

So the rebase reaches the agent through the rework prompt, inside the tree it can also push from. Same outcome, one writer.

## Reproducer

`q-system/.q-system/scripts/test/test-dispatch-rework.sh`, 27 checks.

- **Part A** copies `linear-worker.sh` into a sandbox next to a fake `linear-sync.py` and runs its **real** picker against canned Linear rows. A0 asserts the copy is byte-identical to the shipped file, so the suite cannot grade a stale copy.
- **Part B** drives `kipi-dispatch.sh` for real; every selection assertion reads back a side effect (the process table, the budget file, the log), never a grep of the source.
- **B4** is the blast-radius case: a rework candidate with a LIVE converge is dispatched on 0 of 5 runs and spends 0 budget. The concurrency guard is unweakened.

Observed **RED** before the change -- A1, B1b, B1c, B1d, B2, B6:

```
== 21 passed, 6 failed
```

**GREEN** after:

```
== 27 passed, 0 failed
```

## No regressions

```
test-dispatch-rework.sh          27 passed, 0 failed
test-dispatch-liveness.sh        18 passed, 0 failed
test-linear-worker-fetch.sh      PASS (10 checks)
test-converge.sh                 PASS 17/17
test-linear-worker-parallel.sh   PASS (9 checks)

python3 fleet-capability-verify.py   27 instances, 22 green, 0 red, 5 skipped
python3 plugins/prd-os/scripts/prd_runner.py gates run
  all 0 regression gates green; no open spillover
```

## Not doing (per the DoR)

No auto-merge. No change to `kipi/reviewer-approved` or the validate workflow. The daily cap is not raised. No priority ordering in the picker.

---

# Round 3 (head `38f5fed`) — the reviewer's two findings

Full response: the round-3 comment on this PR.

**MAJOR, fixed.** A rework candidate must now have an **open PR**. Nothing in this repo moves an issue out of `started` when its PR merges, so every issue the loop has ever *finished* was landing in the pool — and with no open PR the apply loop's own `gh pr list --head` is equally empty, so the severity-floor gate was skipped, `BASE` stayed `origin/main`, and the agent got the **fresh-work** prompt for work already on main. Confirmed against live Linear (the reviewer could only infer it): **ASK-150**, `started`, `owner:sana`, DoR, PR merged, no open PR — 1 of 7 pool members on 2026-07-29. The DoR required this clause from the start; the first cut dropped it.

One shared `open_pr_for` reader serves the announcement and the apply loop. A `gh` failure refuses the path rather than announcing a pool it cannot verify. The **count line** reports the pool and the announced survivors separately, instead of calling a pool member a candidate.

**MAJOR, second half, fixed.** `MAX_ATTEMPTS` cannot bound this path: `bump_attempt` has one call site and it fires only on a **non-zero** `claude` exit, while the rework failure mode is an agent that exits 0 with the PR still red. New fifth ledger key `rework_dispatches`, capped at 2, incremented **per dispatch** by `kipi-dispatch.sh` through a new `--rework-dispatched` worker mode (one writer keeps the ledger schema in one script). Fail closed: no record, no dispatch. At the cap it pages once and prints the command that hands the issue back to the loop.

**MINOR, fixed.** The starvation comment claimed more than the code does — fresh work wins within a **tick**, not across a budget day. Restated. The daily-cap line and page now name the fresh/rework split, so a capped day says what it bought.

## Reproducer, updated

`q-system/.q-system/scripts/test/test-dispatch-rework.sh`, now **43 checks** (was 27). The 15 new ones observed **RED** against PR head `de773ca`:

```
RED  (de773ca):  28 passed, 15 failed
GREEN (38f5fed): 43 passed, 0 failed
```

A8/A9 (open-PR requirement, gh-failure refusal), A10/A11 (the dispatch cap, paged once), A12 (the count line tells the truth), B8 (a dispatch spends exactly one rework slot, a fresh one spends none), **B9** (the layer above: two real dispatches drive the budget to the cap, then the real picker refuses the candidate — either half alone leaves the loop unbounded).

The `gh` stub was `exit 0` with no output; once the dry path asks it a real question that stub would have turned Part A green by announcing nothing at all. It now answers from `$WORK/gh-open-branches` / `$WORK/gh-fail`. `work --rework-dispatched` in the Part B stub is **not** stubbed — it execs the real worker against the same ledger Part A reads.

## Spillover opened this round

`gates run` is RED on purpose until closeout: `sp-b9bc6528` (the apply loop still swallows `gh`'s exit code — pre-existing), `sp-88bfa1b5` (a started issue with **no PR at all** has no detector; this fix makes it a stated skip line rather than a wrong fresh-work dispatch), `sp-53b02cc4` (five unlocked read-modify-write bumpers on one ledger, now called from a second process).


assafkip:
CI: `validate` **pass** (3m20s) — https://github.com/assafkip/kipi-system/actions/runs/30472220820/job/90645031752

Spillover captured from this issue's own note (`sp-18c4ec22`): `ready()` matches the literal string `Definition of Ready`, so an issue using `## Outcome` without that exact header is invisible to the dispatcher and silently never picked up. Out of scope here, now in the ledger rather than in a comment.

assafkip:
Both findings reproduced. Shipped suite is genuinely green at PR head (27/27), so this is not a "your tests are broken" review.

---

# Review: PR #43 — rework candidates re-enter the loop (ASK-245)

Scratch: `/tmp/pr43rev/src` (clone at head `de773ca`) and untracked `.pr43rev/` repro scripts, same convention as the existing `.pr42rev/`. No tracked file touched, nothing committed or pushed.

---

## MAJOR — a `started` issue whose PR already MERGED is a permanent rework candidate, and dispatching it re-runs the whole issue as FRESH work

**`q-system/.q-system/scripts/linear-worker.sh:252`** (`def rework(i): return mine(i) and i["state"]["type"] == "started"`)

Nothing in this repo moves a Linear issue out of `started` when its PR merges. The only `ISSUE_UPDATE` / `stateId` call sites are `fleet-health-daily.py`, `linear-triage.py` and `linear-dor-drafter.py` — none on the merge path. `converge.sh` and `linear-worker.sh` contain no state transition at all. So every issue the loop has ever finished stays In Progress, and this PR turns that whole population into rework candidates.

What then happens is not a rework round. `gh pr list --head "$BRANCH"` defaults to `--state open`, so a merged PR yields `EXISTING_PR=""`, the entire severity-floor gate (`linear-worker.sh:641`) is skipped because it lives inside `if [ -n "$EXISTING_PR" ]`, `BASE` stays `origin/main`, and the agent gets the **fresh-work** prompt for work that is already on main.

```
bash .pr43rev/repro-merged-issue.sh
```

```
worker under test == shipped PR-head file

=== 1. is a merged-and-forgotten issue announced as a REWORK candidate? ===
2026-07-29T16:59:03Z worker: 1 rework candidate(s) (owner:sana, has a DoR, already started)
2026-07-29T16:59:03Z [dry] would rework ASK-777 (attempt 1/3)

=== 2. dispatch it the way kipi-dispatch.sh does (converge -> --apply --issue) ===
2026-07-29T16:59:04Z start ASK-777 on sana/ask-777 in .../state/worktrees/ask-777 (attempt 1/3)

--- worktree HEAD (fresh-work path cuts from origin/main) ---
96230b5 c1

--- full agent rounds launched ---
1

--- permanent Linear comments written on the already-merged issue ---
progress ASK-777
progress ASK-777

--- prompt handed to the agent (head) ---
You are Sana, the kipi Systems Engineer, working Linear issue ASK-777.
You are in a DEDICATED GIT WORKTREE at .../worktrees/ask-777, already on branch sa

=== 3. does anything bound the repeat? ===
attempts ledger: (absent -> count 0)
next heartbeat announces:
2026-07-29T16:59:04Z [dry] would rework ASK-777 (attempt 1/3)
```

**The claimed bound does not bind this.** `kipi-dispatch.sh:29-31` says the exit is guaranteed by "the live-converge filter plus the worker's own MAX_ATTEMPTS". `bump_attempt` has exactly one call site — `linear-worker.sh:1112`, reached only when `claude` exits **non-zero**. Here `claude` exits 0, the ledger file is never even created, and the same issue is re-announced at `attempt 1/3` on the next heartbeat, forever. Test **A6** proves the filter works by hand-writing `{"ASK-9$$":{"count":3}}` into the attempts file — a state this population cannot reach, since it requires three consecutive agent crashes.

At 3am: 20+ merged `sana/ask-*` branches exist in this repo. Each corresponding issue still in `started` gets a full agent round, a new PR pushed to a public repo, and two permanent Linear comments, up to 4 per day, indefinitely. In the morning a human closes duplicate PRs, deletes branches, and moves issues to Done. Cleanable, so major and not blocker.

**Calibration, stated honestly:** I could not confirm against live Linear that real ASK issues are sitting in `started` post-merge — the Linear MCP permission was not granted in this session. The fixture shape is the real query's shape and the missing transition is verifiable by grep, so I am not dropping a level, but that one link is code-inferred rather than observed.

---

## MINOR — the comment says the backlog cannot starve behind rework; across a budget day it can

**`kipi-dispatch.sh:195-196`** — *"A FALLBACK, NEVER A PEER. Fresh work is checked first and wins outright: preferring rework would starve the backlog."*

True within one tick. Across a budget day it is not, because rework spends the same `KIPI_DISPATCH_DAILY_MAX`.

```
bash .pr43rev/repro-starve.sh
```

```
budget after 4 rework heartbeats: 4
rework dispatches logged: 4

--- what the heartbeat did with the FRESH issue ASK-87750 ---
2026-07-29T17:00:47Z DAILY CAP: 4/4 issues dispatched for budget day 2026-07-29, stopping until 7:00 local

RESULT: fresh work was REFUSED.
```

The shared cap is a stated design choice in the PR body, so the code does what you intended — it is the sentence that overstates. Worth noting the operator-facing consequence: the cap page reads *"Not an error -- the loop is resting"*, which is accurate wording for a day spent on real work and misleading for a day spent on no-ops.

---

## What is sound

- **Set disjointness.** Tried to get one issue into both `ready` and `rework`. `started` and `backlog`/`unstarted` are disjoint state types; the fresh path can never cut a PR's worktree from `origin/main` via this route. Held.
- **The concurrency guard.** Ran the shipped suite at PR head: `27 passed, 0 failed`, including B4 (live converge, 5 runs, 0 dispatches, 0 budget). The move of the liveness check into the candidate test genuinely does not weaken loop-exit 5.
- **Id-prefix collision.** Tried to make `converge_live_for ASK-24` match a live `converge.sh --issue ASK-245`. The `([[:space:]]|$)` boundary holds. Could not make it fail; dropped.
- **The `--issue` path.** Verified byte-identical: `if only:` sets `rew = []` and `pool` to the named issue, so converge's entry is unchanged.
- **The SIGPIPE/pipefail refactor.** The snapshot-into-a-variable + bash `=~` form is preserved; no pipeline reintroduced by hoisting it.
- **The DoR deviation.** The refusal to rebase from the dispatcher is correct, and the stated reason checks out in code: `position_tree_on_pr_head` (`linear-worker.sh:528`) really does refuse a tree holding commits reachable from neither `origin/<branch>` nor `origin/main`. A dispatch-side rebase would stall the issue for a human. Good call, correctly argued.

---

## VERDICT: REQUEST CHANGES

Not BLOCK — duplicate PRs, branches and Linear comments are noisy but cleanable, and the change is revertible.

**Fix first:** the rework set must require an OPEN PR. Either move the `gh pr list --head` check into the candidate test, or have `kipi-dispatch.sh` refuse a rework candidate with no open PR. Secondary and necessary: give rework its own counter that increments **per dispatch**, not per failed `claude` — `MAX_ATTEMPTS` cannot bound a path where the agent exits 0.

FINDINGS:
major|A Linear issue left in `started` after its PR merged is a permanent rework candidate and is re-run as FRESH work every heartbeat, unbounded, because MAX_ATTEMPTS only increments on a non-zero claude exit|q-system/.q-system/scripts/linear-worker.sh:252
minor|The comment claims the backlog cannot starve behind rework, but rework and fresh work share KIPI_DISPATCH_DAILY_MAX so rework can consume a whole budget day|kipi-dispatch.sh:195
END FINDINGS

assafkip:
# Rework round 3 — both findings fixed, both reproduced red first

Head `38f5fed`. 15 new checks, all observed RED against PR head `de773ca` before the fix.

---

## MAJOR — a `started` issue whose PR merged is a permanent rework candidate, dispatched as FRESH work

**Confirmed, and the one link you flagged as code-inferred is now observed.** You said the Linear MCP permission was not granted, so you could not check whether real ASK issues sit in `started` post-merge. I ran the picker's own filters against live Linear through `linear-sync.py`:

```
total ASK issues: 246
rework candidates (owner:sana + DoR + started): 7
  ASK-245   open_pr=True  In Progress | fix(dispatch): an issue whose PR went red is never pick
  ASK-226   open_pr=True  In Progress | feat(triage): collapse the 49 duplicate-template issues
  ASK-225   open_pr=True  In Progress | feat(ops): dispatch picks file-disjoint issues so it ca
  ASK-223   open_pr=True  In Progress | feat(ops): a broken Linear pipeline pages the founder w
  ASK-221   open_pr=True  In Progress | feat(review): Codex reviews every PR as an independent
  ASK-218   open_pr=True  In Progress | feat(seam): the autonomous path writes a prd-os receipt
  ASK-150   open_pr=False In Progress | Constraint: cron cannot run `claude -p` (no keychain ac
```

ASK-150 is exactly your case: `started`, `owner:sana`, has a DoR, PR merged, nothing open on `sana/ask-150`. One of seven today, and permanent — your grep is right that no code path moves an issue out of `started` on merge.

It is also a DoR miss, not just a defect: the DoR says a candidate "has an **open PR** that is failing checks or unreviewed". The first cut dropped that clause.

### The fix, and where it landed

**`linear-worker.sh`** — the announcement requires an open PR, via `open_pr_for`, which is now the **single reader** of that question. The apply loop's inline `gh pr list --head` was replaced by the same helper rather than a second copy, because the picker comment's own argument ("two readers of `ready` with drifting semantics") applies to this question too.

**The report, not only the detector.** You caught the round-2 version of this and it applies here: the line the operator reads at 3am said `1 rework candidate(s)` for a pool member the run would then refuse. It now reports the pool as a pool and the survivors separately:

```
worker: 7 started issue(s) in the rework pool (owner:sana, has a DoR). Eligibility -- an OPEN PR, budget left -- is decided at the announcement.
[dry] skip rework ASK-150: no OPEN PR on sana/ask-150 -- it is In Progress with nothing to rework (merged and never closed, or the PR was closed). Not fresh work either; a human moves it off In Progress.
worker: 6 of 7 rework candidate(s) announced
```

**A `gh` failure refuses the path rather than falling through** (A9). "gh could not be asked" is not "there is no PR"; falling through would announce the whole pool — merged issues included — on exactly the run where nothing can be verified.

**`kipi-dispatch.sh` echoes the worker's refusals into `dispatch.log`.** The reasons live in `linear-worker.log`, which is not the file anyone opens when the loop looks idle.

### The bound — you are right that `MAX_ATTEMPTS` cannot do this

`bump_attempt` has one call site and it is on the non-zero-`claude` branch. The rework failure mode is an agent that exits **0** with the PR still red. Your A6 critique lands: it proves the filter works from a hand-written `{"count":3}`, a state this population cannot reach.

New fifth ledger key, `rework_dispatches`, capped at **2**, incremented **per dispatch** by `kipi-dispatch.sh` through a new `--rework-dispatched` worker mode. Notes on the shape:

- **The dispatcher does not write the ledger, it asks the worker to.** That file now has five keys answering five different questions and exactly one script has ever written it.
- **Fail closed:** no record written → no dispatch. Same rule as the budget counter it sits beside.
- **Lifetime, not consecutive**, unlike the conflict and drift caps — and stated in the comment. There is no state a reader could trust as "the streak ended": a rework dispatch *itself* moves the head, so a head-change clear would refill the budget on every single dispatch and cap nothing.
- At the cap it **pages once** and prints the one command that hands the issue back to the loop. A candidate that silently stops being dispatched is the failure this whole issue is named for.

### What got quieter, said explicitly

A started issue with **no PR at all** (worker died before pushing, or the PR was closed unmerged) is now picked up by neither set. It was never in the fresh set (state is `started`), and pre-ASK-245 it was invisible everywhere; what changes is that PR #43 as-shipped would have dispatched it *wrongly as fresh work*, and now it is a stated skip line instead. Nothing files or pages on that population — the honest home for that is a `fleet-health-daily.py` detector, not the dispatcher. Captured: `sp-88bfa1b5`.

---

## MINOR — the starvation comment overstates

Agreed, the sentence was wrong, not the code. Rewritten to say what is true: fresh work wins outright **within a tick**; across a budget day the two share `KIPI_DISPATCH_DAILY_MAX`, so a morning of rework can exhaust the day. That is the DoR's stated trade ("Not doing: raising the daily cap").

On your operator-facing note — *"Not an error, the loop is resting"* reading wrong for a day spent on no-ops — two things:

1. The no-op day is now gone at the source. A merged-PR candidate is not a candidate, and a red one is bounded at 2 dispatches, so a capped day cannot be four re-runs of one PR.
2. The cap line and page now name the split, so a rework-heavy day is legible rather than indistinguishable from a shipping day:
   ```
   DAILY CAP: 4/4 issues dispatched for budget day 2026-07-29 (1 fresh, 3 rework), stopping until 7:00 local
   ```
   A sibling file (`dispatch-rework-<day>`), not a field in `COUNT_FILE` — that file is parsed as a bare integer in three places, and not `$COUNT_FILE.rework` either, because the suite reads the budget through a `dispatch-count-*` glob.

---

## What I did not regress

The four attacks you ran and could not break are still asserted, and the ones your findings touch got stronger:

- **B4** (live converge, 5 runs, 0 dispatches, 0 budget) still passes. The rework-budget bump sits *after* the liveness guard, so a passed-over candidate spends nothing.
- **The `--issue` path** is still byte-identical: `only:` sets `rew = []`.
- **Set disjointness** (A2/A3b) unchanged.
- **A0** (the copy under test is byte-identical to the shipped worker) still passes, so none of this is graded against a stale copy.

One thing about the suite worth flagging, because it would have hidden the fix: the `gh` stub was `exit 0` with no output. Once the dry path asks `gh` a real question, that stub means "no PR" for every candidate and would have turned Part A green by announcing **nothing at all**. It now answers from two files (`$WORK/gh-open-branches`, `$WORK/gh-fail`) so each case states its world. And `work --rework-dispatched` in the Part B stub is **not** stubbed — it execs the real worker against the same ledger Part A reads, which is what makes **B9** a real end-to-end: the budget the dispatcher spends is the budget the picker refuses on.

## Evidence

```
RED  (against de773ca, new cases only failing):   28 passed, 15 failed
GREEN (head 38f5fed):                             43 passed, 0 failed

test-dispatch-liveness.sh        18 passed, 0 failed
test-linear-worker-fetch.sh      PASS (10 checks)
test-converge.sh                 PASS 17/17
test-linear-worker-parallel.sh   PASS (9 checks)
fleet-capability-verify.py       27 instances, 22 green, 0 red, 5 skipped
```

## Spillover opened this round (gate is RED on purpose, cleared at closeout)

- `sp-b9bc6528` — the apply loop's `EXISTING_PR=` still swallows `gh`'s exit code, so an outage there is indistinguishable from "no PR" and takes the fresh-work path. The announcement now checks rc; the apply loop does not. Pre-existing, unchanged by this issue.
- `sp-88bfa1b5` — the "started, no PR at all" population has no detector (above).
- `sp-53b02cc4` — `linear-worker-attempts.json` now has five unlocked read-modify-write bumpers, and this issue is the first to call one from a second *process*. Never concurrent for the same issue, but the class is real.

assafkip:
Regex attack survived. Writing up.

---

# Review — PR #43, round 4 (head `38f5fed`)

Read the code first, then the earlier rounds. Round 3's two findings are both genuinely closed on this head — I ran the author's own A8/A9/A10/A12 cases and they pass, and I confirmed `kipi work` forwards `"$@"` to the worker (`kipi:98`), so the B9 dispatcher→picker handshake is not a fixture artifact. Nothing from round 3 is re-raised.

One new finding, on code that round 3 introduced.

## MAJOR — the announcement checks that a candidate has an *open PR*, but never that the PR is one the apply loop will actually rework; the new lifetime cap turns each of those wasted dispatches into a permanent lockout with a false page

`q-system/.q-system/scripts/linear-worker.sh:702-733` decides eligibility on owner, DoR, `started`, `MAX_ATTEMPTS`, `MAX_REWORK_DISPATCHES`, and an open PR. The apply loop that the dispatch buys applies a *further* gate 30 lines later (`linear-worker.sh:801`, `rework_gate`): exit **20** (no recorded verdict) and exit **10** (approved) both `continue` — no agent, no work.

So a candidate that clears the announcement can still be a guaranteed no-op. It still spends a daily budget slot, a rework-budget slot, and a converge launch. `MAX_REWORK_DISPATCHES=2` (new in `19c5657f`) is lifetime and never cleared, so after two heartbeats the issue is out of the loop for good, and the page says something that is not what happened.

This is not constructed. Two of the five PRs the PR body names as the motivating population land in these two gates on live state right now:

```
PR 40: verdict='REQUEST CHANGES' head_sha=f206ae49fe md=3     <- reworkable
PR 36: verdict='REQUEST CHANGES' head_sha=9afa054fab md=3     <- reworkable
PR 35: verdict='REQUEST CHANGES' head_sha=14255a2ccb md=3     <- reworkable
PR 34: NO VERDICT RECORD, md=0                                <- gate 20, skip
PR 23: verdict='APPROVE WITH NITS' head_sha=None md=2         <- gate 10, skip
```

(`~/.config/kipi/pr-reviews/`, read directly. `#34`=ASK-221 and `#23`=ASK-210 are two of the five in the PR body's own stranded table; their Linear `started`/`owner:sana`/DoR state is the author's premise, not my query.)

The DoR clause round 3 cited is `"has an open PR that is failing checks **or unreviewed**"`. Round 3 closed the *open PR* half. The *unreviewed* half of the same sentence is the half the apply loop refuses at gate 20 — so the DoR and the apply loop disagree, and today the loop pays a converge launch to rediscover that every heartbeat until the cap locks the issue out.

**Reproducer:** `.pr43rev-r2/repro-verdict-blind.sh` (harness mirrors the PR's own `test-dispatch-rework.sh`; drives the real worker and the real `kipi-dispatch.sh`)

```
bash .pr43rev-r2/repro-verdict-blind.sh .pr43rev-r2/wt
```

**Real output:**

```
### STEP 1 -- gate the APPLY loop would apply to this PR
rework_gate '' (no verdict record)        -> exit 20  (20 = unreviewed, apply loop CONTINUEs)
rework_gate 'APPROVE WITH NITS'           -> exit 10  (10 = approved,  apply loop CONTINUEs)

### STEP 2 -- the ANNOUNCEMENT still calls it a candidate (dry run)
2026-07-29T17:30:39Z [dry] would rework ASK-929821 (PR #4242, attempt 1/3, rework dispatch 1/2)
2026-07-29T17:30:39Z worker: 1 of 1 rework candidate(s) announced

### STEP 3 -- the APPLY round the dispatch buys (converge -> --apply --issue)
2026-07-29T17:30:40Z skip ASK-929821: PR #4242 has no recorded review verdict -- run: kipi review 4242 --issue ASK-929821 --post

### STEP 4 -- what the dispatcher spent for that no-op
  2026-07-29T17:30:40Z dispatching ASK-929821 (rework, live=0 cap=999 rounds=3 budget=1/9)
  2026-07-29T17:30:52Z dispatching ASK-929821 (rework, live=0 cap=999 rounds=3 budget=2/9)
  daily budget spent : 2
  rework budget spent: 2

### STEP 5 -- the issue is now permanently out of the loop, and the page says:
2026-07-29T17:31:03Z [dry] skip rework ASK-929821: 2/2 rework dispatch(es) already and the PR is still not green. Marked stuck; a human decides next. To hand it
  PAGE: worker: ASK-929821 has had 2 rework dispatch(es) and its PR is still not green - the loop has stopped picking it up, it needs a human
```

**What a human does at 3am, per stuck issue:** open the PR to find the real blocker (branch protection on `#23`, a missing review on `#34`), run `kipi review <PR#> --issue <ID> --post` for the gate-20 ones, then hand-run the printed python one-liner to clear `rework_dispatches`. Recoverable, so major and not blocker.

**Worse for `#23` specifically:** gate 10 arms auto-merge and `converge.sh:206` then pages *"approved and auto-merge armed — GitHub lands it, no human merge needed"*. Every rework dispatch on an already-approved PR re-fires that page. Two dispatches, two redundant pages, then a third page claiming the opposite. That is the cry-wolf anchor.

**Smallest fix:** the announcement already shells `gh` for `open_pr_for`. Reading the verdict record it is standing next to and skipping non-`0`/`30`/`40` gates costs one more call and no new concept — the same "one reader, shared not copied" move the author already made for `open_pr_for`.

---

## On the review process

Round 3 and round 4 are both majors on the same predicate — *what belongs in the rework pool* — and both clauses are in one sentence of the DoR. Round 3 read half of it. Reviewing the eligibility contract one clause per round is what produced a fourth round here; reading the DoR sentence against the apply loop's gate would have closed both at once. That is a miss on the review side, not the author's.

## What is sound

- **`converge_live_for` prefix collision.** `ASK-24` matching a live `converge.sh --issue ASK-245`. `.pr43rev-r2/probe-regex.sh` → `ASK-2 -> no match / ASK-24 -> no match / ASK-245 -> MATCH`. The `([[:space:]]|$)` anchor holds.
- **The fixture-shape attack.** I expected the Part-B `kipi` stub's `--rework-dispatched` branch to be inventing an interface production doesn't have. It isn't: real `kipi:98` is `bash linear-worker.sh "$@"`, and both halves resolve `STATE_DIR` to the same `$HOME/.config/kipi`. B9's handshake is real.
- **Test suite is not decoration.** 43/43 green on this head; A0's byte-identity check would catch a stale sandbox copy, and the round-4 `gh` stub answers a real question rather than exiting 0 (a silent stub would have turned Part A green by announcing nothing).
- **Concurrency guard is unweakened.** B4: 5/5 runs held, 0 budget spent, with a live decoy converge.
- **Fail-closed ordering.** Ledger record → rework counter → daily counter → launch, with the live-converge check ahead of all of them. My STEP 4 exercised the real path; a failed record refuses the dispatch rather than running uncapped.
- **The DoR deviation (fetch, don't rebase) is correct.** `position_tree_on_pr_head` (`linear-worker.sh:528`) would refuse a dispatch-side rebase exactly as the PR body claims.

## VERDICT: REQUEST CHANGES

Fix first: gate the rework announcement on `rework_gate`, so a dispatch is only spent on a PR the apply loop will actually work. Everything else on this head stands.

Scratch dir `.pr43rev-r2/` holds the repro and a clone at PR head; untracked, nothing in the repo was modified.

FINDINGS:
major|The rework announcement never checks the severity-floor verdict the apply loop applies, so a candidate at rework_gate 10 or 20 burns a daily slot, a rework slot and a converge launch doing nothing, and the new lifetime MAX_REWORK_DISPATCHES then locks the issue out permanently with a page that misstates the state|q-system/.q-system/scripts/linear-worker.sh:702
END FINDINGS

assafkip:
# Rework round 4 -- the finding is fixed, and so are the two rungs below it

Head `a8a82ff`. 20 new checks, all observed RED against PR head `38f5fed` before the fix.

---

## MAJOR -- the announcement never checks the verdict the apply loop applies

**Confirmed as stated.** `rework_gate` exits 10 and 20 both `continue` in the apply loop, and the announcement was blind to both -- so a candidate could clear it and still be a guaranteed no-op that had already spent a daily slot, a rework slot and a converge launch. With `MAX_REWORK_DISPATCHES` lifetime, two of those locked the issue out for good.

Your live evidence checks out against the same records: `#34` has no verdict record (gate 20), `#23` reads `APPROVE WITH NITS` (gate 10). Two of the five PRs the DoR was filed to rescue.

### The fix

The announcement asks the same gate, through **one shared reader** -- `read_rework_gate`, which the apply loop now also uses in place of its inline copy. Two callers computing "the verdict, the merge state and the two shas" from their own `gh` calls is the two-readers-of-one-input defect `pr-verdict-lib.sh` exists to close, and it is the same move round 3 made for `open_pr_for`. `read_rework_gate` writes nothing: every side effect (the ledger clears, the arm, the pages) stayed on the apply path.

### The two rungs below the ones you named

Your argument does not stop at 10 and 20. Gates **30** and **40** are real work, but each carries its own per-issue round budget, and at the cap the apply loop skips exactly the same way. An announcement blind to those buys the identical guaranteed no-op two rungs down -- so `round_budget_exhausted` now owns that comparison for both paths.

**And the page survives the refusal.** The apply loop pages at both caps; refusing the dispatch would have taken the page down with it, which is the "silence bought by a fix" this round is about. The announcement pages from there, using the apply loop's **own claim keys**, so the two paths can never double-page one fact.

```
A17b the cap still pages a human, from the announcement       PASS
A17c and once, not per heartbeat                              PASS
A19b the drift cap still pages a human, from the announcement PASS
```

### The page that misstated -- it was an ordering bug

You said the page "says something that is not what happened". Fixing the gate is only half of that. The cap checks ran **before** the gate, and both cap lines make a claim about the PR (`still not green`, `marked stuck`) while the rework cap pages it. So the moment a human approved a capped PR, the operator's page said the opposite of the truth.

The caps now run **after** the gate. A capped issue whose PR is now approved reads:

```
[dry] skip rework ASK-9xxxx: PR #4242 verdict is 'APPROVE' -- nothing to rework; it waits on the
  merge. Nothing here records it as armed, so if it sits green: gh pr merge --auto --squash 4242
```

and not the stuck line, and it pages nobody (`A20a/b/c`). Cost, stated: a few `gh` calls per capped candidate per heartbeat, forever. That buys a page that is true when it fires.

I did **not** add a clear for `rework_dispatches` on a gate-10 read. Round 3's argument for lifetime-not-consecutive still holds -- a rework dispatch itself moves the head -- and a clear is a new concept with its own runaway risk that you did not ask for. The lockout stays; it just stops lying about why.

---

## What got quieter, and what I did about it

**Gate 10 loses its auto-merge arm.** The apply loop's gate-10 branch calls `arm_automerge` (PR #33's catch-up for PRs approved before that code existed), and that branch is only reachable through a dispatch. Refusing the dispatch means nothing arms this population.

Arming from the dry run is not the answer: `kipi work` without `--apply` is a read the dispatcher polls every 15 minutes, its one documented write is a page claim, and arming auto-merge is the single action in that file that puts code on main with no human in the path. A poll must not land code.

So the skip line names the recorded arm state and the command, `kipi-dispatch.sh` echoes it into `dispatch.log`, and the gap is in the ledger rather than in this comment: **`sp-fba8d194`**. The honest home is a `fleet-health-daily.py` detector or a cheap arm sweep, not a converge launch.

That invariant is now asserted, not just intended -- `test-severity-floor.sh` fails if the dry branch ever calls `arm_automerge`, and fails if it stops naming the merge command.

**Gate 20 and the DoR's "unreviewed" half.** You are right that the DoR and the apply loop disagree, and the apply loop is right: a rework prompt with no review to rework against is a guess. What that population needs is a **review** dispatch (`kipi review <PR#> --issue <ID> --post`), a different dispatch kind than this one and outside this issue's scope. The announcement now prints that exact command instead of paying a converge launch to rediscover it every heartbeat. The missing trigger: **`sp-ccdf677f`**.

Nothing else went quiet. `worker: N of M rework candidate(s) announced` still decrements per skip, and every new skip line matches `kipi-dispatch.sh:240`'s echo filter, so all of this lands in `dispatch.log` where the operator looks when the loop seems idle.

---

## One thing your round-3 pass and mine both walked past

`test-severity-floor.sh` anchored its arm assertion on the **first** `GATE" = "10"` in the worker via `head -1`. Adding a second gate-10 branch -- one that correctly does *not* arm -- made a passing suite go red on a correct change. Repointed at the apply loop (anchored on `DONE=0`) and pinned to exactly two sites, so a third cannot slip behind the anchor unnoticed. `146/146`.

---

## Evidence

```
RED  (new cases against 38f5fed):   49 passed, 18 failed
GREEN (head a8a82ff):               67 passed,  0 failed
```

The RED included the end-to-end, not just the picker:

```
FAIL B10a a gate-20 candidate costs the daily budget nothing   expected [0] got [1]
FAIL B10b and no rework budget slot                            expected [0] got [1]
FAIL B10c and no converge launch
     a converge round was paid for a PR the apply loop refuses at gate 20
```

`B10`/`B11` are new and forward `work` to the **real** worker exactly as `kipi:98` does (`"$@"` and all), against the same fake Linear rows Part A uses, and read the cost off the dispatcher's own budget file and process table. Every other Part-B case feeds a canned `work` line -- right for measuring selection, structurally blind to a finding whose whole content is that the worker announced something it should not have.

`B11` is the pair `B10` needs: same harness, same issue, one field different in the verdict record, and the dispatch happens. A zero has to prove it is empty rather than broken.

```
test-dispatch-rework.sh          67 passed, 0 failed
test-severity-floor.sh           146/146
test-dispatch-liveness.sh        18 passed, 0 failed
test-converge.sh                 17/17
test-linear-worker-fetch.sh      PASS (10 checks)
test-linear-worker-parallel.sh   PASS (9 checks)
```

## What I did not regress

- **B4** (live converge, 5 runs, 0 dispatches, 0 budget) still passes. The gate read sits after the liveness guard, so a passed-over candidate spends nothing -- and now costs no `gh` call either.
- **A0** byte-identity still passes, so none of this is graded against a stale copy.
- **The `--issue` path** is untouched: `only:` still sets `rew = []`.
- **Set disjointness** (A2/A3b), **A8/A9** (open-PR requirement and the `gh`-failure refusal), **A15/A16/A18** (gates 0, 30 and 40 are still announced -- the fix must not mute the population this issue exists to rescue).

## Housekeeping

The unattended auto-commit safety net had pushed my own round-3 review scratch into this PR's diff (`.redcheck/`, `.pr43-*.md`, `.ask245-*.py`, `.run-suites.sh` -- ~3,000 lines). Removed, and the naming convention is gitignored so the next issue inherits it rather than rediscovering it.

assafkip:
Review complete. Scratch lives in `.runs/pr43r4/` (gitignored on main); nothing tracked was touched.

# Review: PR #43, round 5 (head `a8a82ff`)

## MAJOR — the rework budget is spent at the announcement, and nothing refunds it when the round it buys does zero work

`kipi-dispatch.sh:322` records a rework dispatch **before** the converge launch. The apply loop it buys has several refusals *after* the gate that do no work at all: a worktree that cannot be created (`linear-worker.sh:1092`, which says out loud "not counted against the issue"), and the claim at step 2 returning exit 3 on a stale lock. `MAX_REWORK_DISPATCHES` is lifetime and never cleared (`linear-worker.sh:853`), so two such ticks lock the issue out of the loop permanently and page a cause that is not the cause.

This is the same predicate round 3 and round 4 each found one clause of. Round 4's fix (`read_rework_gate`) is in this head and my repro runs past it: gate 0, announcement clean, dispatch bought, zero rounds performed.

It is also the exact defect this file already fixed one layer down. `linear-worker.sh:1035-1044` documents it as PR #25 finding 2: *"two runs skipped by a stale claim burned the whole budget having dispatched ZERO rebases, then paged the founder a round count that never happened and locked the issue out until someone hand-edited the ledger."* The cure there was plan-then-spend-at-the-dispatch-site. The new counter spends at announcement.

**Reproducer** (drives the real worker from head `a8a82ff`; poisons the worktree path to force one of the post-gate refusals):

```
bash .runs/pr43r4/repro-budget-refund.sh
```

**Real output:**

```
=== TICK 1 =================================================
--- heartbeat dry (what kipi-dispatch.sh reads):
2026-07-29T18:06:24Z [dry] would rework ASK-9001 (PR #4242, gate 0, attempt 1/3, rework dispatch 1/2)
--- kipi-dispatch.sh records the dispatch it is about to buy:
2026-07-29T18:06:24Z recorded rework dispatch 1/2 for ASK-9001
--- the converge round that dispatch bought (apply):
2026-07-29T18:06:25Z INFRA: could not create worktree for ASK-9001 (not counted against the issue)
--- ledger after tick 1: {"last_rework_dispatch": "...", "rework_dispatches": 1}

=== TICK 2 =================================================
2026-07-29T18:06:25Z [dry] would rework ASK-9001 (PR #4242, gate 0, attempt 1/3, rework dispatch 2/2)
2026-07-29T18:06:25Z recorded rework dispatch 2/2 for ASK-9001
2026-07-29T18:06:25Z INFRA: could not create worktree for ASK-9001 (not counted against the issue)
--- ledger after tick 2: {"last_rework_dispatch": "...", "rework_dispatches": 2}

=== TICK 3 (the lockout) ===================================
2026-07-29T18:06:25Z [dry] skip rework ASK-9001: 2/2 rework dispatch(es) already and PR #4242
  still reads 'REQUEST CHANGES'. Marked stuck; a human decides next. To hand it back: python3 -c "..."
--- pages sent to the founder:
worker: ASK-9001 has had 2 rework dispatch(es) and PR #4242 still reads 'REQUEST CHANGES'
  - the loop has stopped picking it up, it needs a human
--- rework rounds actually performed: 0
```

**What a human does at 3am:** reads a page saying the PR is still red and blames the agent, when the PR was never touched. Then hand-runs the printed python one-liner to clear `rework_dispatches`. Recoverable, so major and not blocker.

**No test covers this.** 83 assertions in `test-dispatch-rework.sh`, none of them assert that a post-gate refusal costs zero rework budget. `grep -n "worktree\|claim\|refund\|INFRA"` on the suite returns only comment text and the fresh-work-path cases.

**Smallest fix:** move the bump to where the round actually happens, the same move `CONFLICT_ROUND` / `DRIFT_ROUND` already make. The apply loop knows it got past the claim and the worktree; the dispatcher does not. Alternative if the dispatcher must own it: refund on the refusal, since converge's exit already distinguishes "did a round" from "declined".

---

## MINOR — the comment says the dry path writes once; it writes at three sites

`linear-worker.sh:856` states *"This is the ONLY write a dry run makes."* The gate-30 and gate-40 block 30 lines above (`linear-worker.sh:825` and `:830`) calls `claim_page_once` on the same dry path, which read-modify-writes the same ledger. A maintainer reasoning about dry-run safety from that sentence gets it wrong.

**Reproducer:**

```
bash .runs/pr43r4/repro-dry-writes.sh
```

**Real output:**

```
--- ledger BEFORE the dry run:
{ "ASK-9001": { "conflict_rounds": 2 } }
2026-07-29T18:07:32Z [dry] skip rework ASK-9001: PR #4242 is 'APPROVE' but DIRTY after 2 rebase round(s) ...
--- ledger AFTER the dry run (no --apply anywhere):
{ "ASK-9001": { "conflict_rounds": 2, "conflict_paged": true } }
--- pages:
worker: ASK-9001 PR #4242 is approved but still DIRTY after 2 rebase round(s) - needs a human
```

The behaviour is intentional and correct. The sentence is not.

---

## On the review process

Round 4's reviewer already called this: *"Reviewing the eligibility contract one clause per round is what produced a fourth round here."* This is round five and it is the third clause of the same predicate — does the announcement model what the apply loop will actually do. Rounds 3, 4 and 5 each found one refusal the announcement was blind to, all on code that shipped in round 1.

The pattern is not "the author keeps regressing". It is that nobody has enumerated the apply loop's refusals once, end to end, and checked the announcement against the list. That enumeration is roughly ten lines of reading and it closes the class instead of one more instance of it. Doing it is worth more than my finding above.

---

## What is sound

- **Fixture shape vs the real producer.** I expected the Part-A `gh` stub's "no open PR" answer to be a shape `gh` never emits. Ran the real thing: `gh pr list --head sana/does-not-exist-xyz --state open --json number -q '.[0].number' | od -c` returns `0000000 \n` — a bare newline, which `$(...)` strips to empty. The stub emits nothing, also empty. Same shape. Attack dropped.
- **The `--rework-dispatched` handshake is real, not invented.** `kipi:98` is `bash linear-worker.sh "$@"`, so the flag reaches the worker. My repro drove the real mode against the real ledger and the announcement read back the key it wrote (`rework dispatch 1/2` → `2/2` → lockout). No fake interface.
- **The `--issue` path is genuinely untouched.** `only:` sets `rew = []` at `linear-worker.sh:321`; every apply run in my repro used `--issue` and the rework classification never fired.
- **The open-PR requirement holds.** I tried to get a merged-and-never-closed issue announced; `open_pr_for` returns empty and `linear-worker.sh:775` refuses. Could not break it.
- **`read_rework_gate` really is one reader.** Both callers set the same six globals from one function; I found no second `gh pr view` for merge state or head sha anywhere on either path.
- **The daily-cap split arithmetic.** I traced `DISPATCHED_TODAY - REWORK_TODAY`; the rework file is only written on the path that also writes the count file, so the subtraction cannot go negative. Tried, could not make it wrong, dropped.

---

## VERDICT: REQUEST CHANGES

Fix first: stop spending the rework-dispatch budget on a round that has not happened yet. Everything else on this head stands, and the round-4 gate work is correct.

FINDINGS:
major|The rework-dispatch budget is recorded before the converge round runs and is never refunded when the apply loop refuses after the gate (unbuildable worktree, stale claim), so two such heartbeats permanently lock the issue out of the loop and page a cause that is not the cause|kipi-dispatch.sh:322
minor|The comment claims the rework_paged claim is the only write a dry run makes, but the gate-30 and gate-40 branches also call claim_page_once on the dry path|q-system/.q-system/scripts/linear-worker.sh:856
END FINDINGS

assafkip:
# Rework round 5 -- both findings fixed, and the class the reviewer named

Head `7bb8569`. Both findings reproduced RED before the fix.

---

## MAJOR -- the rework budget is spent at the announcement, and nothing refunds it

**Confirmed as stated, including the diagnosis.** `bump_rework_dispatch` ran at
`kipi-dispatch.sh:322`, before the launch, and every refusal the apply loop makes
*after* the gate cost a lifetime slot having run zero rounds. You are also right
that this is `linear-worker.sh:1035-1044` verbatim -- PR #25 finding 2, whose fix
comment sits thirty lines from where the new counter spent early.

### The fix, and why not a refund

You offered two shapes. The refund is not available: the launch is deliberately
`start_new_session` fire-and-forget (so launchd cannot reap the child with the
dispatcher), which means this script never sees converge's exit code and cannot
know whether a round ran. So it is your first shape -- move the bump to where the
round happens -- with one piece of plumbing the conflict and drift rounds do not
need:

**converge calls the worker up to `MAX_ROUNDS` times per dispatch.** A bare bump
at the spend site would silently redefine a 2-*dispatch* budget as a 2-*round*
one and fire mid-dispatch. So the dispatcher exports `KIPI_REWORK_DISPATCH_ID`
and the worker spends **at most one slot per distinct id**, at the same site it
spends a conflict or drift round. Planned at the gate, paid after the worktree,
the claim and the tree positioning have all had their chance to decline.

- **Fail-closed moved with it, one layer in.** If the worker cannot write the
  record it refuses the round and releases the claim, rather than running
  uncapped. What changed is *where* the refusal happens: at the thing that costs
  money (an agent round), not at a launch that may buy nothing.
- **Still one writer.** The dispatcher names the dispatch; it never touches the
  JSON. The `--rework-dispatched` mode is gone with its only caller, and the
  ledger functions moved back down beside their four siblings -- the reason they
  lived above the `git fetch` was that mode.
- **No token means no dispatcher.** A human running `kipi work --apply --issue X`
  by hand spends nothing. The cap's own hand-back message points them at exactly
  that command, so charging them would re-lock the issue they just freed.

### The class, not the instance

Your process note is the more valuable half, so I did the enumeration rather than
patch the third clause. The apply loop's refusals after the gate, end to end:

| refusal | `linear-worker.sh` | announcement | budget |
|---|---|---|---|
| gate 10 approved | 968 | refuses (round 4) | -- |
| gate 20 no verdict | 982 | refuses (round 4) | -- |
| gate 30/40 round cap | 1006 / 1029 | refuses (round 4) | -- |
| worktree uncreatable | 1092 | cannot know | **0, this round** |
| stale claim (exit 3) | 1156 | cannot know | **0, this round** |
| claim rc != 0 | 1157 | cannot know | **0, this round** |
| tree cannot be positioned | 1163 | cannot know | **0, this round** |

The split is the point: the first three are *facts about the PR* the announcement
can read, and round 4 taught it to. The last four are facts about **this machine
at this instant** -- an announcement cannot predict them, and a fourth round of
teaching it to try would have found a fifth. Spending at the round closes all of
them at once and closes the ones nobody has written yet.

## MINOR -- "the ONLY write a dry run makes"

Agreed, the sentence was wrong and the behaviour right. Corrected to what is
true: a page claim is the only **kind** of write a dry run makes (it never
touches a budget, an arm or a tree), and all three cap sites make one. `A17d`
now asserts the gate-30 claim really is written, so the next person who "fixes"
the prose by deleting the write has to fail a test to do it.

---

## What got LOUDER, said explicitly

This fix does not buy silence anywhere -- it buys noise, and that direction has a
cost worth stating. An issue whose round refuses for a *permanent local* reason
(a tree that cannot be positioned; a worktree that can never be built) no longer
accumulates toward a lockout, so the dispatcher re-launches it every heartbeat up
to the daily cap and can starve fresh work on repeat. The tree refusal pages once
(`tree_paged`); nothing bounds the repeats.

I did not add a second lifetime counter for it. That population is identical on
the **fresh-work** path today and always has been, so it is a pre-existing class
this issue did not widen, and its honest home is a "dispatched N times, ran zero
rounds" detector in `fleet-health-daily.py` -- not another budget in the worker.
In the ledger rather than in this comment: **`sp-fc76347b`**.

## Evidence

RED, against head `a8a82ff`:

```
test-severity-floor.sh   FAIL: S1 a rework round that actually ran spent 0 rework dispatch(es), not 1
test-dispatch-rework.sh  FAIL B8a2  the converge child is handed this dispatch's id
                         FAIL B11b  and the lifetime rework budget waits for the round
                         FAIL B12b  a dispatch whose round did nothing spends no rework budget
                         == 69 passed, 3 failed
```

GREEN, head `7bb8569`:

```
test-severity-floor.sh          PASS: 150/150   (was 146; +4 section S)
test-dispatch-rework.sh         == 72 passed, 0 failed   (was 61; +11)
test-dispatch-liveness.sh       == 18 passed, 0 failed
test-converge.sh                PASS 17/17
test-linear-worker-fetch.sh     PASS (10 checks)
test-linear-worker-parallel.sh  PASS (9 checks)
```

Section S drives the **real apply loop** to the spend, which is what the round-5
finding needed and no case had: a round that runs pays 1; a second round of the
same dispatch pays 0; an untokened hand-run round pays 0; a round refused by a
stale claim pays 0. B12 is your repro end to end -- real dispatcher, real worker,
a file where the worktree goes -- and reads the ledger back as untouched.

Honest note on the RED set: `B8a`, `B8c` and `B9a` could not be shown red against
`a8a82ff`, because removing `--rework-dispatched` from the Part-B stub makes the
old dispatcher's recorder a no-op there. They are the contract restated; `B8a2`,
`B11b`, `B12b` and `S1` are the catch.

## What I did not regress

The six attacks you ran and could not break still hold, and two are now asserted
rather than merely true:

- **B4** (live converge, 5 runs, 0 dispatches, 0 budget) still passes.
- **The `--issue` path** is still untouched: `only:` sets `rew = []`.
- **The open-PR requirement** and **`read_rework_gate` as one reader** are
  unchanged; the gate still runs before the caps, so `A20a/b/c` (a capped issue
  whose PR is now approved is not called stuck) still pass.
- **The daily-cap split arithmetic** you traced is unchanged: `REWORK_COUNT_FILE`
  still counts LAUNCHES at the dispatch site, deliberately. It splits
  `DISPATCHED_TODAY`, a launch is spent whether or not the round works, and it
  resets at `RESET_HOUR` -- it cannot lock anything out, which is the whole
  reason the lifetime counter had to move and this one did not.
- **A17b/A17c/A19b** (the caps still page, from the announcement, once) pass.

assafkip:
## Round 4 review, PR #43

I read the diff and ran the code before opening any earlier round.

---

### First, a finding about the review process

This is round 4 (the PR's own comment thread numbers its responses 2 through 5). Look at what each round produced:

| Round | Major found |
|---|---|
| 3 | announcement doesn't check for an **open PR**, so a dispatch is spent on a guaranteed no-op |
| 4 | announcement doesn't check the **severity floor**, so a dispatch is spent on a guaranteed no-op |
| 5 | budget is spent at **launch, not at the round**, so a dispatch is spent on a guaranteed no-op |

Three rounds, three majors, one defect class: *the announcement commits budget for work the apply loop will refuse.* The apply loop has a finite, greppable list of `continue` statements. Enumerating them once in round 3 would have surfaced all three at once instead of costing the author three full rework cycles. Round 4's major in particular was on the **original** announcement block, unchanged since round 1 and read by two prior reviewers.

I checked whether a fourth instance survives. It does not. Every refusal between the gate and the spend (unbuildable worktree, another session's claim, `position_tree_on_pr_head`, the bump itself) now precedes `bump_rework_dispatch` at step 4, and the bump-failure branch releases the claim before it continues (`linear-worker.sh:1209-1220`). The class is closed.

---

### Findings

**minor** — the rework pool grows monotonically and never drains, so every issue the loop has ever *finished* costs one `gh pr list` and one echoed `dispatch.log` line per heartbeat, forever.

`q-system/.q-system/scripts/linear-worker.sh:784`

I verified the author's premise independently rather than taking the comment. The only `stateId` writers in the repo are `fleet-health-daily.py`, `linear-collapse-jobmigration.py`, `linear-triage.py`, `linear-dor-drafter.py`. None is on the merge path:

```
$ grep -rn "stateId" q-system/.q-system/scripts/ plugins/
q-system/.q-system/scripts/fleet-health-daily.py:1460:  update["stateId"] = state_id
q-system/.q-system/scripts/linear-collapse-jobmigration.py:210: ... {"stateId": close_id}
q-system/.q-system/scripts/linear-triage.py:471: ... {"stateId": close_id}
q-system/.q-system/scripts/linear-dor-drafter.py:207: issueUpdate(id:$id,input:{stateId:$s})
```

Reproducer (`.pr43rev-r4/repro-merged-pool.sh`, drives the real picker at head `7bb8569` against a counting `gh` stub):

```
$ bash .pr43rev-r4/repro-merged-pool.sh 40

pool size (issues started, PR merged, never closed): 40
gh invocations in ONE dry heartbeat : 40
[dry] skip rework lines emitted     : 40
lines the dispatcher would echo     : 40
candidates actually announced       : 0

--- a sample of what dispatch.log gets, every 15 minutes, forever ---
2026-07-29T18:45:22Z [dry] skip rework ASK-100: no OPEN PR on sana/ask-100 -- it is
In Progress with nothing to rework (merged and never closed, or the PR was closed).
Not fresh work either; a human moves it off In Progress.

--- projection at 96 heartbeats/day ---
  dispatch.log lines/day : 3840
  gh API calls/day       : 3840   (GitHub authenticated ceiling ~120k/day)
```

The consequence is specifically on round 3's own fix. The dispatcher echo exists so the operator can see *why the loop went quiet* when they open `dispatch.log`. The one line that matters there is `rework: gh could not be asked`. It now arrives buried under N copies of "ASK-150 merged, nothing to do", where N is every issue the loop has ever shipped, and grows by up to `DAILY_MAX` per day with no code path that removes a member.

Severity held at minor deliberately: nothing pages, no budget is spent, no permanent object is written, and the loop keeps working. It degrades one log. It is not a false alarm in the alerting sense, so it does not reach the major anchor. At today's pool (7) it is ~576 lines/day; the growth is what makes it worth a line rather than nothing.

---

### What is sound

- **Stdin loss in the announcement loop.** `while read RID; do ... gh ... done <<EOF` with `gh`, `python3` and the notifier inside is the exact shape of this repo's own "a loop that spawns readers needs stdin isolation" lesson. Fed 5 candidates through the real picker: 5 of 5 announced, `worker: 5 of 5 rework candidate(s) announced`. No loss.
- **Fixture realism.** The `gh` stub's `pr view` branch keys on the bare `mergeStateStatus` / `headRefOid` words, which is exactly the shape `pr_merge_state` and `pr_head_sha` emit (`pr-verdict-lib.sh:77,89`). Case A0 pins the sandbox copy byte-identical to the shipped worker, so the suite cannot grade a stale file. This is not a fixture built from the same mental model as the code.
- **Set disjointness.** `ready` is backlog/unstarted, `rework` is `started`, `completed` is in neither. Confirmed in the picker and by A5. A rework candidate cannot reach the fresh-work `BASE` block and cannot have its PR commits blown away.
- **The `--issue` path.** `rew = []` whenever `only` is set, so converge's drive path is genuinely unperturbed.
- **grep collision.** `[dry] would work` is not a substring of `[dry] would rework` (the space lands before `rework`). The dispatcher's two `grep -oE` patterns cannot cross-match.
- **`converge_live_for` boundary.** `ASK-8` does not match `ASK-80`; the `([[:space:]]|$)` anchor holds. B4 proves the liveness guard is unweakened by the fallback: 0 dispatches, 0 budget across 5 runs.
- **Full suite at head `7bb8569`: `72 passed, 0 failed`.** (The PR description still says "27 checks", which is stale by 45. Description text, not code.)

One finding I dropped. The dry path is now a **new writer** of `linear-worker-attempts.json` (three `claim_page_once` sites), running from the 15-minute heartbeat while a converge apply run may be mid-write, and every writer uses `json.dump(d, open(p,'w'))` which truncates before it writes. A reader that lands in that window gets `except Exception: d={}` and clobbers the whole ledger. I could not make this PR-specific: `bump_attempt` and `bump_conflict_round` already had the identical pattern on main, and `MAX_CONCURRENT > 1` already allowed two apply runs to race there. Pre-existing class, not this PR's to carry, and I could not build a repro that fails on head but passes on main. Dropped.

---

One minor survived reproduction. Nothing here is unrecoverable, nothing pages falsely, and the budget path fails closed at every layer I could reach.

Most important thing to fix first: give the pool a drain, or filter the echo. The cheapest version is one line in the dispatcher's echo grep to drop `no OPEN PR` (it is the one skip reason that is permanent and unactionable by the loop), which keeps `gh could not be asked` readable. The real fix is a state transition on merge, which is a different issue.

FINDINGS:
minor|The rework pool never drains, so every merged-and-never-closed issue costs one gh call and one echoed dispatch.log line per heartbeat forever, burying the one refusal line the dispatcher echo was added to surface|q-system/.q-system/scripts/linear-worker.sh:784
END FINDINGS

assafkip:
Closing as a redrive. ASK-245 stays In Progress and the branch is preserved.

State: CONFLICTING with main, 71 commits behind, opened 2026-07-29.

The problem is real and is not fully solved on main: `test-dispatch-rework.sh` does not exist there, and an issue whose PR went red never being picked up again is exactly the failure mode that makes In Progress grow without bound.

Adjacent work HAS landed on main since this branch was cut, which is part of why it conflicts: ASK-212 (the rework gate reads mergeability, so an approved PR that later breaks is dispatchable) and ASK-219 (pass the reviewed sha and current head to rework_gate) are both Done. Whoever redrives this should check how much of the rework-candidate path those two already cover before rebuilding it.

Branch `sana/ask-245` keeps the 10-commit implementation.

PR #44, opened 2026-07-29
ASK-151: one mechanism migrates all 32 scheduled jobs to Linear-tracked execution

assafkip:
Closes the collapsed job-migration family. ASK-151 folded 32 near-identical
`Migrate com.<x> to Linear-tracked execution` issues into one, so it gets one
mechanism rather than 32 hand passes.

## The mechanism

`q-system/.q-system/scripts/job-migration.py`, driven by the roster data file
`q-system/.q-system/config/job-migration-roster.json`, reachable as
`kipi jobs status|migrate|run|gate`.

It scores every roster job against the four bars the per-job contract already
named:

| Bar | Meaning | How it is proven |
| -- | -- | -- |
| 1 | launchd, never cron | plist present, and the label appears in no crontab line. `crontab -l` unreadable is scored as a FAIL, never as an empty crontab |
| 2 | failures reach Linear, not only a log | the label is covered by a `launchd-health-check.py` watched prefix, which is exactly the set whose failures become `launchd-failing` / `launchd-dark` Linear issues |
| 3 | live or in the pause ledger, never dark | loaded, or present in the pause ledger. `launchctl` not answering is a FAIL, not a silent pass |
| 4 | verified by a run, not by reading the plist | a receipt whose MODE is recorded: `run`, `scheduler`, or `paused-ledger` |

## Three choices worth reviewing

**The roster is data, not a LaunchAgents glob.** The machine holds 38 plists and
growing. A glob lets the denominator drift under the gate, and a roster that
quietly shrinks reports a clean pass over work that vanished. ASK-151: "a partial
pass that reports success is the failure mode."

**`AUTO_RUN` is False.** 26 of the 32 are paused ON PURPOSE pending this
migration. Firing a paused Substack producer or `git push` from inside an audit
is un-pausing it for one cycle, which the issue's Not-doing line forbids. A run
is opt-in per label. The verification mode travels on every receipt so the gate
cannot launder one into another, and the not-run-verified count is itemized by
mode and printed out loud, never folded into green.

**Bar 2 is the only bar with an auto-fix**, and it fixed the real gap this found:
`com.personal.` was covered by no watched prefix, so `com.personal.story-podcast`
could fail or go dark with no Linear issue and no ping. The fix appends to the
instance-local `EXTRA_PREFIXES_FILE` that `launchd-health-check.py` already
reads, never to `WATCHED_PREFIXES`, which ships fleet-wide and must carry no one
machine's job families. Bar 3 has no auto-fix on purpose: un-pausing a dark job
is a founder call, not an audit's.

## Demonstrated on one job, then the rest

```
$ kipi jobs migrate --only com.personal.story-podcast
would add watch prefix com.personal. to ~/.config/kipi/launchd-watch-prefixes.txt
BLOCK: com.personal.story-podcast -- bar2-linear :: ... failures reach nobody

$ kipi jobs migrate --only com.personal.story-podcast --apply
added watch prefix com.personal. to ~/.config/kipi/launchd-watch-prefixes.txt
OK: 1/1 roster jobs are on Linear-tracked execution
```

Load-path proof that the fix reached the real consumer, not just a file:

```
watchdog prefixes now: ('com.kipi.', 'com.cole.', 'com.claudedaddy.', 'com.ask.',
                        'com.assaf.', 'com.ktlyst.', 'com.purespectrum.', 'com.personal.')
com.personal.story-podcast watched by the LIVE watchdog: True
```

Then the remaining 31 through the same mechanism:

```
$ kipi jobs migrate --apply
wrote 30 receipt(s)
roster 32 / expected 32
verification modes: 2 none, 26 paused-ledger, 4 scheduler
```

## Two collisions and a refusal, worth calling out

**`kipi jobs` was already taken** at line 88 by `linear-job-migration.py`, the
issue FILER. A second `jobs)` case appended lower in the same `case` statement is
unreachable -- bash takes the first match -- so the first version of this wiring
was dead text that grepped like it worked. Dispatch is now on an explicit
subcommand inside the one case; every arg the filer already accepts still reaches
the filer. Both routes verified.

**The gate refused a silent exit 0.** `com.kipi.audit-rotate` ran clean and
printed nothing. That is refused on purpose: a wrapper over a script that
`kipi update` rsync --delete'd away also exits 0, which is the exact silent-death
class this migration exists for. A silent run now has to have LEFT something on
the paths the PLIST ITSELF declares. audit-rotate still shows none and is
reported as exactly that -- ran, exit 0, proved nothing -- rather than counted
as a pass.

## Current gate state (honest)

```
roster 32 / expected 32
BLOCK: com.kipi.fractional-cxo.opp-scan -- bar4-no-receipt
```

The run for that job is in flight as this PR opens; its result goes on the issue
as a comment. Every other roster job clears bars 1-3 and carries a receipt.

## Test

`q-system/.q-system/scripts/test/test-job-migration.py`, 50 checks, registered in
the capability manifest. Observed RED (script absent) before green. Every
assertion drives PURE functions on fabricated facts: no `launchctl`, no live
LaunchAgents, no live pause ledger, no live receipt file. The blast radius here is
the founder's real scheduled jobs, so the test does not go near them.

## Captured, not fixed

- `sp-b301890a` -- 3 launchd jobs outside the roster with no tracking issue
- `sp-9b17d467` -- `com.kipi.audit-rotate` cannot satisfy bar 4 as written


assafkip:
`grounding-guard-skip` — the two flagged entries are bare directory path components inside the sandbox-setup shell snippet (`q-system/.q-system/scripts`, `q-system/.q-system/config`), not claims about file contents. The files themselves I did read this session: `job-migration.py`, `job-migration-roster.json`, `test-job-migration.py`, `launchd-health-check.py`, `fleet-health-daily.py`, `test-fleet-health-daily.py`, `kipi`.

# PR #44 review — ASK-151 job migration

Read-only. Sandbox at `$TMPDIR/pr44sbx` (PR files extracted from `origin/sana/ask-151`, nothing in the repo touched). Rebuild it with:

```bash
python3 - <<'EOF'
import subprocess, os, shutil, tempfile
repo="/Users/assafkipnis/projects/kipi-system"; root=os.path.join(tempfile.gettempdir(),"pr44sbx")
shutil.rmtree(root, ignore_errors=True)
s=os.path.join(root,"q-system/.q-system/scripts"); c=os.path.join(root,"q-system/.q-system/config")
os.makedirs(s); os.makedirs(c); os.makedirs(os.path.join(s,"test"))
g=lambda f: subprocess.run(["git","-C",repo,"show","origin/sana/ask-151:"+f],capture_output=True,text=True).stdout
open(s+"/job-migration.py","w").write(g("q-system/.q-system/scripts/job-migration.py"))
open(c+"/job-migration-roster.json","w").write(g("q-system/.q-system/config/job-migration-roster.json"))
for f in ("launchd-health-check.py","fleet-health-daily.py"):
    open(s+"/"+f,"w").write(g("q-system/.q-system/scripts/"+f))
print(root)
EOF
```

Baseline: the shipped test suite is green (`python3 q-system/.q-system/scripts/test/test-job-migration.py` → `PASS: job-migration contract holds`). Every finding below is a hole the green suite does not see.

---

## Finding 1 — major

**The gate never asks whether a receipt verified. It only asks whether one exists — so a run that failed with exit 3 clears the gate.**

`q-system/.q-system/scripts/job-migration.py:344`

`bar_verified()` correctly returns `False` for a nonzero exit, and the shipped test asserts that. `gate_result()` then throws that verdict away: it appends `bar4-no-receipt` only when `verify_mode == "none"`. `row["verified"]` is computed, carried into the row, and never read by the gate.

`cmd_run` is the real producer of this shape — it writes the receipt regardless of exit code (`job-migration.py:520-530`).

```bash
python3 $TMPDIR/pr44sbx/repro1_failed_run_passes_gate.py
```

```
=== kipi jobs run com.kipi.fleet-health --apply ===
running: /bin/sh -c 'echo boom-dependency-missing >&2; exit 3'
exit=3  24 chars of output  effect=none  -> receipt written
boom-dependency-missing

cmd_run exit: 1

=== the receipt this wrote to the ledger ===
{
  "detail": "ran by job-migration.py in 0s",
  "effect": null,
  "exit_code": 3,
  "mode": "run",
  "output_head": "boom-dependency-missing\n",
  "recorded_at": "2026-07-29T19:28:54Z"
}

=== bar_verified()'s own verdict on that receipt ===
(False, 'run', 'last run exited 3')

=== kipi jobs gate ===
roster 1 / expected 1
verification modes: 1 run
NOT run-verified: 1 (1 run). ... `run` here = exit 0 with no output and no observable effect, which is recorded, not scored as proof.
OK: 1/1 roster jobs are on Linear-tracked execution (bars 1-3 green, every job carries a receipt)
cmd_gate exit: 0
```

Two things in that output, not one. The gate exits 0 on a failed verification run. And the explanation line it prints is factually wrong about the receipt in hand: it says "exit 0 with no output", the receipt says exit 3 with output. An operator reading that line learns the wrong thing about their own ledger.

Fix is one condition: block on `not row["verified"] and row["verify_mode"] != "paused-ledger"`, instead of on `verify_mode == "none"`.

---

## Finding 2 — major

**A receipt, once persisted, is never re-aged. The 14-day freshness rule is unreachable after the first `migrate --apply`.**

`q-system/.q-system/scripts/job-migration.py:309`

```python
receipt = receipts.get(label) or scheduler_receipt(label, status_kind, info)
```

`scheduler_receipt()` refuses a launchd log older than `SCHEDULER_LOG_MAX_AGE_SECONDS` (line 88), with the comment "A scheduler log older than this is not evidence the job still works." True — and then `migrate --apply` freezes that receipt into `~/.config/kipi/job-migration-receipts.json`, and the persisted copy wins on every subsequent read. It carries no timestamp the gate checks.

Bar 3 passes a loaded-but-failing job on purpose (the PR's own test: "loaded is recorded; failing is bar 4's job"). Bar 4 is frozen. Nothing sees it.

```bash
python3 $TMPDIR/pr44sbx/repro3_frozen_receipt.py
```

```
=== day 0: kipi jobs migrate --apply (job loaded, log fresh) ===
wrote 1 receipt(s) to .../work3/receipts.json
verification modes: 1 scheduler
OK: 1/1 roster jobs are on Linear-tracked execution

persisted receipt: {"detail": "launchd log ... written 0h ago", "exit_code": 0, "mode": "scheduler", ...}

=== day 200: the job now FAILS every run, and its log is 200 days old ===
launchctl says: failing | log age: 200 days | SCHEDULER_LOG_MAX_AGE_SECONDS = 14 days
a FRESH scheduler receipt would be: None

=== kipi jobs gate ===
roster 1 / expected 1
verification modes: 1 scheduler
OK: 1/1 roster jobs are on Linear-tracked execution (bars 1-3 green, every job carries a receipt)
cmd_gate exit: 0
```

A job that has been failing every single run for 200 days, whose own freshness rule says its evidence expired 186 days ago, gates green. That is the paper pass ASK-151 names as the failure mode, produced by the mechanism written to prevent it.

Mitigating, and worth saying: `launchd-health-check.py` files a `launchd-failing` Linear issue for that job independently, so the operator is not blind. What is wrong here is the gate's claim, not the fleet's coverage.

Fix: stamp the receipt with the evidence's own age and re-check it on read, or re-derive `scheduler` receipts every time and persist only `run` and `paused-ledger`.

---

## Finding 3 — minor

**Bar 1 proves "launchd, never cron" by matching the launchd LABEL as a substring of a cron line. A real crontab line names the SCRIPT PATH. The fleet's own documented cron duplicate scores PASS.**

`q-system/.q-system/scripts/job-migration.py:114`

`fleet-health-daily.py:314` documents the live incident: *"Found live 2026-07-26: reddit-build-radar runs at 08:00 from `com.cole.reddit-radar-daily` AND from a crontab line."* That label is roster row 20. `job-migration.py` imports that very module for `crontab_text()` — and reimplements the match weaker.

The PR's test covers this with `0 7 * * * /bin/bash /x/com.kipi.x/run.sh` — a path with the label baked into it. No crontab in this fleet emits that shape.

```bash
python3 $TMPDIR/pr44sbx/repro2_bar1_misses_real_cron.py
```

```
real plist program : /Users/assafkipnis/projects/cole-gtm/projects/reddit-build-radar/scripts/run_daily.sh
the duplicate cron : 0 8 * * * /bin/bash /Users/assafkipnis/projects/cole-gtm/projects/reddit-build-radar/scripts/run_daily.sh

bar_launchd  -> True | launchd only
             ^ True == 'launchd only', i.e. bar 1 says this job is NOT on cron

the fleet's existing detector on the SAME cron text -> ['double-scheduled: /Users/.../run_daily.sh runs from both launchd and cron']
```

Held at minor, not major, for two reasons I want visible: `crontab -l` on this machine is currently empty (rc 0, no output), and `detect_duplicate_schedules` catches the real condition daily and files a Linear issue. Nothing pages wrong at 3am. What is wrong is that bar 1 will read green for all 32 jobs regardless of cron state — a check that cannot fail on any input a real crontab produces.

---

## Finding 4 — minor

**`cmd_run` drops the plist's `EnvironmentVariables` and substitutes the operator's shell env, so the strongest bar-4 mode verifies a different invocation than launchd performs.**

`q-system/.q-system/scripts/job-migration.py:518` — `subprocess.run(argv, ..., cwd=...)` with no `env=`.

12 of the 32 roster plists declare `EnvironmentVariables` on this machine. 11 pin a narrowed `PATH`; `com.cole.reddit-radar-daily` pins `REDDIT_RADAR_SEND=1` and `RADAR_SCHEDULER=launchd`.

```bash
python3 $TMPDIR/pr44sbx/repro4_env_dropped.py
```

```
plist declares EnvironmentVariables: {'REDDIT_RADAR_SEND': '1', 'RADAR_SCHEDULER': 'launchd'}

running: /bin/sh -c 'echo REDDIT_RADAR_SEND=[$REDDIT_RADAR_SEND] ...'
exit=0  91 chars of output  effect=none  -> receipt written
REDDIT_RADAR_SEND=[] RADAR_SCHEDULER=[] LEAKED_FROM_SHELL=[yes-this-is-the-operator-shell]

bar_verified says: (True, 'run', 'ran by job-migration.py in 0s')
```

Both directions are wrong: the declared vars are gone, and the operator's shell vars are present. Bar 1's whole justification is ASK-150 — "cron has no keychain, so any job that shells `claude` fails auth there" — which is an *environment* difference. Bar 4's run mode then discards the declared environment. A job that passes `kipi jobs run` under a rich interactive `PATH` can still fail under launchd's narrowed one, which is the exact class this migration exists for.

Fix: `env={**os.environ, **(info.get("EnvironmentVariables") or {})}` at minimum; closer to launchd would be a minimal base plus the declared vars.

---

## Finding 5 — minor

**`receipts.json` is read-modify-write with no lock. Two writers lose one's receipt.**

`q-system/.q-system/scripts/job-migration.py:206`

```bash
python3 $TMPDIR/pr44sbx/repro5_attacks.py
```

```
ATTACK B: two writers on the receipt ledger
  ledger after both wrote: ['com.cole.b', 'seed']
  X receipt survived -> False
```

Scheduler and paused-ledger receipts re-derive on the next `migrate`, so most of the loss self-heals. A `run` receipt does not — it needs a human to run the job again. Holding at minor because nothing in the repo invokes `job-migration.py` concurrently today (nothing schedules it at all), so this needs two hands on two terminals to hit.

---

## What is sound

Attacks I ran that the code survived. These matter, because three of them are where I expected this to break:

**The bar-2 auto-fix genuinely reaches the live consumer.** I expected the appended line `com.personal.  # added by job-migration.py (ASK-151) <ts>` to be dead text — a "prefix" with a comment glued to it that matches no plist. It is not. `load_watched_prefixes()` splits on `#` before stripping, so the line round-trips, and the fix is idempotent (a second `migrate` finds the prefix already covered and does not re-append). Verified against the real `launchd-health-check.py`, not a stub:

```
  the LIVE reader returns: ('com.kipi.', 'com.cole.', 'com.claudedaddy.', 'com.ask.', 'com.assaf.', 'com.personal.')
  com.personal. present -> True
  second migrate re-appends? -> False
```

The PR's load-path proof is real. Writing to `EXTRA_PREFIXES_FILE` rather than `WATCHED_PREFIXES` is also the right call — the latter ships fleet-wide via `kipi update`.

**The `kipi jobs` dispatch does not break the filer.** The PR's own writeup flags the unreachable-second-`case` collision it hit; I tested every arg form against the shipped block and the filer keeps everything it had:

```
kipi jobs                                     -> linear-job-migration.py (the Linear issue FILER)
kipi jobs --apply                             -> linear-job-migration.py (the Linear issue FILER) --apply
kipi jobs status                              -> job-migration.py status
kipi jobs migrate --apply                     -> job-migration.py migrate --apply
kipi jobs run com.kipi.audit-rotate --apply   -> job-migration.py run com.kipi.audit-rotate --apply
kipi jobs gate                                -> job-migration.py gate
```

No repointing of a documented command that files permanent Linear issues. That was the highest-blast-radius thing in this diff and it is handled.

**`crontab_text()` cannot fake an all-clear.** `fleet-health-daily._crontab_text()` raises `CrontabUnavailable` rather than returning `""`, the wrapper converts that to `None`, and `bar_launchd` scores `None` as FAIL. "I could not look" never prints as "I looked and found nothing."

**All 32 roster labels resolve to a real plist** in `~/Library/LaunchAgents` — the roster is not aspirational.

**`AUTO_RUN = False` holds.** No path in the module runs a job as a side effect of `status`, `migrate`, or `gate`; `cmd_run` needs an explicit label plus `--apply`, and refuses a label outside the roster.

**Dropped: log-size blowup.** `scheduler_receipt()` does `path.read_text()[-2000:]`, reading the whole log into memory for every roster job on every `status`/`gate`. I expected this to be a real hazard on a machine with unrotated launchd logs. Measured the actual declared logs: **424 KB total across all 32 jobs**, largest 80 KB. Not a finding. Dropping it.

**One thing I did not file, worth a sentence:** nothing schedules `kipi jobs gate`. Grep across the repo, the plugin tree, and `~/Library/LaunchAgents` returns no caller. It is a manual command, which is consistent with what the PR claims — it just means findings 1 and 2 will be found by whoever types it, not at 3am.

---

## VERDICT: REQUEST CHANGES

Two majors, both in the gate. Merging as-is is recoverable — nothing here writes to Linear, deletes anything, or un-pauses a job, and the receipt ledger is machine-local and regenerable. So this is not a BLOCK.

**Fix first: `job-migration.py:344`.** The gate computes `row["verified"]` and then does not read it. Right now `kipi jobs gate` answers "does a receipt file exist for every job" while printing that it answered "is every job verified". One condition change closes it, and the shipped test already asserts the correct verdict at the `bar_verified()` layer — it just needs a gate-level case that feeds a `run`/exit-3 row into `gate_result` and expects non-zero. Finding 2 is the same shape one layer down and is worth fixing in the same pass.

FINDINGS:
major|gate_result only blocks on verify_mode=="none" and ignores row["verified"], so a run receipt with exit_code 3 clears the gate while the gate prints that it was exit 0|q-system/.q-system/scripts/job-migration.py:344
major|collect() prefers the persisted receipt over a fresh one, so SCHEDULER_LOG_MAX_AGE_SECONDS is unreachable after the first migrate --apply and a job failing every run for 200 days still gates green|q-system/.q-system/scripts/job-migration.py:309
minor|bar 1 proves "launchd, never cron" by substring-matching the launchd label in a cron line, but real cron lines name the script path, so the fleet's own documented 2026-07-26 double-schedule of com.cole.reddit-radar-daily scores PASS|q-system/.q-system/scripts/job-migration.py:114
minor|cmd_run drops the plist EnvironmentVariables and leaks the operator shell env, so a run-verified receipt proves a different invocation than launchd performs (12 of 32 roster plists declare env)|q-system/.q-system/scripts/job-migration.py:518
minor|write_receipts is an unlocked read-modify-write on receipts.json, so two concurrent writers silently drop one's receipt|q-system/.q-system/scripts/job-migration.py:206
END FINDINGS

assafkip:
Closing: the premise this PR was built on is obsolete.

This PR ships one mechanism (`job-migration.py` + a 32-entry roster) to migrate all 32 scheduled jobs to Linear-tracked execution. Two things have happened since 2026-07-29:

1. **The roster's own issues were cancelled.** Every sibling migration issue is now `Canceled` in Linear: ASK-152 through ASK-186 (com.cole.*, com.kipi.*, com.personal.story-podcast, com.purespectrum.ti-weekly, and the rest). ASK-151 is the last one still open. A mechanism that migrates 32 jobs whose tracking issues were deliberately cancelled is building for a decision that was already reversed.
2. **A founder-directed jobs audit on 2026-08-01 (eb43899) set per-job keep / pause / kill verdicts directly**, including killed schedules and paused jobs. That audit, not a bulk migration, is now the live answer for what happens to these jobs.

`job-migration.py` is not on main, so nothing is lost by closing: the branch `sana/ask-151` keeps the full implementation and ASK-151 stays In Progress as the tracker.

If the single remaining job (com.assaf.competitive-analysis.morning) still wants migrating, that is a one-job change against current main, not a 32-job roster.

PR #67, opened 2026-08-02
fix(ledger): the attempts lock proceeded on timeout and released by path (ASK-286)

assafkip:
Closes the defect `sp-626e9452` recorded as founder-deferred: `attempts-ledger.py`
`_mutate` shipped a lock that **proceeded on timeout** and **released by path**.
Back-ports the version PR #42 (`7e8d767`) proved in `converge.sh`
`receipt_transaction`.

## Why this was worse than no lock

A run that timed out believed it was serialized, entered the transaction holding
nothing, and its release then deleted the live lock of the run that actually held
it. Both runs sat inside one read-decide-write.

The counter it guards decides whether an issue is retried or marked STUCK. A lost
increment means an issue exceeds `MAX_ATTEMPTS` and keeps being dispatched --
silent budget burn on work that already failed three times, and a human is never told.

It had already propagated. On 2026-08-02 an agent was told to reuse an in-repo lock
rather than invent a fourth, copied this one **including both defects**, and cited
the source as justification. Codex caught it there. Citing a pattern is not
verifying it.

## The four invariants

1. **A timeout returns FAILURE** -- exit 3, deliberately distinct from `claim-flag`'s
   1 ("already claimed, stay quiet") and never 0 ("claimed, go page"). The caller
   writes nothing and says so on stderr. A skipped bump is recoverable: the next run
   retries and the run that *did* hold the lock is counting. Two writers in one
   transaction is not.
2. **A release removes only a lock carrying this run's own token**, never one
   identified by path.
3. **The lock is created already carrying its owner.** The token goes into a temp
   file and `os.link` publishes it atomically, so the lock never exists
   un-attributable. `mkdir` left that window open, and a run killed inside it left a
   lock nothing could ever own.
4. **Stale locks break on owner liveness**, and a lock nothing can ever own (empty
   file, or the leftover `mkdir` DIRECTORY a pre-fix worker leaves behind) is a
   corpse by construction. Without that, the first upgrade wedges the ledger forever.

Residual is named in the docstring rather than papered over: the token re-read
before a break is a narrow TOCTOU window and pid reuse can make a corpse look live.
Both degrade in the safe direction; neither can silently delete a live run's lock.

## Reproducer first, both defects observed RED

```
$ python3 q-system/.q-system/scripts/test/test-attempts-ledger.py     # unfixed
FAIL: THE DEFECT: the lock is held by a LIVE pid and bump-attempt reported success.
A run that cannot take the lock must do no work and say so -- entering the
transaction here is two runs inside one read-decide-write.
```

Case 3 red on the unfixed code with `IsADirectoryError` -- the `mkdir` lock cannot
even carry a token, so release-by-path was the only thing it could do.

## Mutation-checked, because case 3 would otherwise be decorative

The release is only reached *after* a successful acquisition, so once the timeout
stops lying there is no ordinary path into it holding someone else's lock. A suite
that never makes the lock change hands mid-transaction passes a release-by-path
mutant -- the trap PR #42 hit. Case 3 stamps a foreign token over the lock from
inside the critical section.

```
$ python3 q-system/output/mutcheck-ask-286.py   # ownership check disabled
mutant: release-by-path (ownership check disabled)
lock still present after release: False
MUTANT KILLED by case 3
```

## Green

```
$ python3 q-system/.q-system/scripts/test/test-attempts-ledger.py
  ok: a timeout writes nothing, says so, and leaves the live holder's lock alone
  ok: claim-flag under contention neither claims nor reports a claim
  ok: a release refuses to remove a lock that no longer carries this run's token
  ok: a lock left by a dead run is broken, written through, and released
  ok: a lock nothing can ever own (stale mkdir directory, empty file) is broken, not honoured
  ok: every op round-trips through the guarded writer and releases its lock
PASS (6 cases)

$ bash q-system/.q-system/scripts/test/test-terminal-states.sh    -> 14 passed, 0 failed
$ bash q-system/.q-system/scripts/test/test-linear-worker-parallel.sh -> PASS (9 checks)
```

## Files

- `q-system/.q-system/scripts/attempts-ledger.py` -- the four invariants, exit 3 on
  lock failure
- `q-system/.q-system/scripts/test/test-attempts-ledger.py` -- new, 6 cases
- `q-system/.q-system/capability-manifest.json` -- registers the test
- `q-system/.q-system/scripts/linear-worker.sh` -- rewrites the comment that
  documented the old take-it-by-force reasoning. That comment was the exemplar, and
  it was wrong.

## Not doing (per the DoR)

`MAX_ATTEMPTS` and the retry policy are untouched. `converge.sh` is untouched -- it
is already correct and merged.

`sp-626e9452` becomes resolvable once this lands and ASK-286 closes.

None of the ledger's callers run under `set -e`, so an exit 3 is a reported miss on
stderr, not a dead worker (verified: `linear-worker.sh:47` is `set -uo pipefail`).

### DIFF
--- a/q-system/.q-system/capability-manifest.json
+++ b/q-system/.q-system/capability-manifest.json
@@ -30,6 +30,10 @@
       "runner": "bash"
     },
     {
+      "path": "q-system/.q-system/scripts/test/test-claim-page-once-routing.sh",
+      "runner": "bash"
+    },
+    {
       "path": "q-system/.q-system/scripts/test/test-claude-write-path.sh",
       "runner": "bash"
     },
--- a/q-system/.q-system/capability-manifest.json
+++ b/q-system/.q-system/capability-manifest.json
@@ -18,6 +18,10 @@
       "runner": "bash"
     },
     {
+      "path": "q-system/.q-system/scripts/test/test-attempts-ledger.py",
+      "runner": "python3"
+    },
+    {
       "path": "q-system/.q-system/scripts/test/test-auto-update-nudge.sh",
       "runner": "bash"
     },
--- a/q-system/.q-system/scripts/attempts-ledger.py
+++ b/q-system/.q-system/scripts/attempts-ledger.py
@@ -17,37 +17,181 @@ bump_attempt is one of six, and the other five kept racing on the same file.
 Codex caught the claim. Hence ONE writer, here, rather than six copies of a lock
 -- six copies of a lock is just the original defect with more places to drift.
 
-CONCURRENCY. mkdir is the mutex because it is atomic on POSIX and needs no
-flock; macOS ships no flock, and this fleet runs on both kernels. The write is
-temp-then-os.replace so a crash mid-write cannot leave a truncated ledger, which
-would read as `{}` and silently reset every budget in the file.
+CONCURRENCY. No flock: macOS ships none and this fleet runs on both kernels, so
+the mutex is a file whose creation is O_EXCL. The write is temp-then-os.replace
+so a crash mid-write cannot leave a truncated ledger, which would read as `{}`
+and silently reset every budget in the file.
 
-A lock we cannot take within the timeout is TAKEN ANYWAY rather than skipping
-the mutation. A dropped bump is the defect this file exists to prevent, and a
-stale lock left by a killed worker is far likelier than a live worker holding it
-for ten seconds.
+THE LOCK WAS WORSE THAN NO LOCK (ASK-286, sp-626e9452, back-ported from the
+version PR #42 proved in converge.sh `receipt_transaction`). It used to PROCEED
+on timeout -- break out of the retry loop and enter the transaction holding
+nothing -- and its release removed the lock BY PATH. Together: a run that times
+out believes it is serialized, walks in, and on the way out deletes the live
+lock of the run that actually holds it. Both runs are then inside one
+read-decide-write, which is precisely what this file exists to stop.
+
+It also propagated. On 2026-08-02 an agent was told to reuse an in-repo lock
+rather than invent a fourth, copied THIS one including both defects, and cited
+the source as justification. Citing a pattern is not verifying it.
+
+FOUR INVARIANTS, all absent from the version this replaces:
+
+  1. A TIMEOUT RETURNS FAILURE. The caller does no work and says so. A skipped
+     increment is recoverable -- the next scheduled run retries, and the run
+     that DID hold the lock is counting. Two runs inside one read-decide-write
+     is not recoverable.
+  2. A RELEASE REMOVES ONLY A LOCK CARRYING THIS RUN'S OWN TOKEN, never one
+     identified by path alone.
+  3. THE LOCK IS CREATED ALREADY CARRYING ITS OWNER. The token is written into a
+     temp file first and `os.link` publishes it at the lock path atomically, so
+     the lock never exists un-attributable for even an instant. mkdir left that
+     window open, and a run killed inside it left a lock nothing could ever own.
+  4. STALE LOCKS BREAK ON OWNER LIVENESS, so refusing to force does not let a
+     corpse wedge every future write. A lock whose pid is gone is broken; so is
+     one that carries no attributable owner at all -- an empty file, or the
+     leftover DIRECTORY a pre-fix worker killed mid-transaction leaves behind,
+     since by invariant 3 this writer can never produce either.
+
+RESIDUAL, named rather than papered over: the token is re-read immediately
+before a break, but that is still a narrow TOCTOU window, and pid reuse can make
+a corpse look live. Both degrade in the SAFE direction -- either the write is
+skipped and says so, or a break happens that the re-read guard makes vanishingly
+unlikely. Neither can silently delete a live run's lock, which is the property
+that was actually broken.
 """
 import json
 import os
+import random
 import sys
 import tempfile
 import time
 
-LOCK_TRIES = 100
+# Overridable ONLY so the suite can assert the contended path without a 10s sleep
+# per case. Production never sets it, same posture as converge.sh's
+# KIPI_RECEIPT_LOCK_TRIES.
+LOCK_TRIES = int(os.environ.get("KIPI_ATTEMPTS_LOCK_TRIES") or 100)
 LOCK_SLEEP = 0.1
 
 
-def _mutate(path, fn):
-    """Take the lock, apply fn(d) -> bool(changed), write atomically."""
-    lock = path + ".lock"
+class LockUnavailable(Exception):
+    """The lock could not be taken. NEVER swallowed: the caller writes nothing."""
+
+
+def _new_token():
+    return "%d:%d:%d" % (os.getpid(), int(time.time()), random.randrange(1 << 30))
+
+
+def _read_lock(lock):
+    """The token a lock carries, or "" if it carries none (missing, empty, or a
+    stale directory from the pre-ASK-286 writer)."""
+    try:
+        with open(lock) as fh:
+            return fh.read().strip()
+    except OSError:
+        return ""
+
+
+def _publish_lock(lock, token):
+    """Create `lock` already carrying `token`, or raise FileExistsError.
+
+    os.link is the atomic publish: the content is complete in the temp file
+    before the name exists, so there is no instant at which the lock is present
+    without an owner. Raises OSError for an unwritable parent -- which is a
+    refusal to write, not a licence to proceed unlocked.
+    """
+    parent = os.path.dirname(lock) or "."
+    os.makedirs(parent, exist_ok=True)
+    fd, tmp = tempfile.mkstemp(dir=parent)
+    try:
+        with os.fdopen(fd, "w") as fh:
+            fh.write(token + "\n")
+        os.link(tmp, lock)
+    finally:
+        try:
+            os.unlink(tmp)
+        except OSError:
+            pass
+
+
+def _owner_is_alive(seen):
+    """True unless the lock names a pid that is provably gone.
+
+    A lock carrying no parseable pid is NOT alive: nothing can ever prove it
+    owns the lock, so honouring it means waiting forever.
+    """
+    pid = seen.split(":", 1)[0]
+    if not pid.isdigit() or int(pid) <= 0:
+        return False
+    try:
+        os.kill(int(pid), 0)
+    except ProcessLookupError:
+        return False
+    except OSError:
+        return True            # exists but not ours to signal
+    return True
+
+
+def _break_if_dead(lock):
+    """Remove a lock whose owner is gone. Leaves a live holder's lock alone."""
+    seen = _read_lock(lock)
+    if _owner_is_alive(seen):
+        return
+    # Re-read: a lock that changed hands between the liveness check and here
+    # belongs to someone else now and is not ours to break.
+    if _read_lock(lock) != seen:
+        return
+    try:
+        if os.path.isdir(lock):
+            os.rmdir(lock)
+        else:
+            os.unlink(lock)
+    except OSError:
+        return
+    sys.stderr.write(
+        "attempts-ledger: broke the lock at %s -- it was left by a run that is no "
+        "longer alive\n" % lock)
+
+
+def _take_lock(lock, token):
     for _ in range(LOCK_TRIES):
         try:
-            os.mkdir(lock)
-            break
+            _publish_lock(lock, token)
+            return True
         except FileExistsError:
-            time.sleep(LOCK_SLEEP)
-        except OSError:
-            break          # unwritable dir: proceed unlocked rather than lose the count
+            _break_if_dead(lock)
+        except OSError as exc:
+            raise LockUnavailable(
+                "the lock at %s cannot be created (%s)" % (lock, exc))
+        time.sleep(LOCK_SLEEP)
+    return False
+
+
+def _drop_lock(lock, token):
+    """Only ever removes a lock carrying this run's token. A release that trusts
+    the path deletes whichever run happens to hold it."""
+    if _read_lock(lock) != token:
+        sys.stderr.write(
+            "attempts-ledger: NOT releasing %s -- it no longer carries this run's "
+            "token, so it belongs to another run now\n" % lock)
+        return
+    try:
+        os.unlink(lock)
+    except OSError:
+        pass
+
+
+def _mutate(path, fn):
+    """Take the lock, apply fn(d) -> bool(changed), write atomically.
+
+    Raises LockUnavailable rather than proceeding: see invariant 1 above.
+    """
+    lock = path + ".lock"
+    token = _new_token()
+    if not _take_lock(lock, token):
+        raise LockUnavailable(
+            "another run still holds the lock at %s. NOT entering the transaction -- "
+            "a run that cannot take the lock writes nothing rather than racing the "
+            "run that can." % lock)
     try:
         try:
             with open(path) as fh:
@@ -73,10 +217,7 @@ def _mutate(path, fn):
             raise
         return True
     finally:
-        try:
-            os.rmdir(lock)
-        except Exception:
-            pass
+        _drop_lock(lock, token)
 
 
 def op_bump(d, issue, counter, ts_key, ts, why=None):
@@ -115,7 +256,19 @@ def main(argv):
         return 2
     path, op, rest = argv[1], argv[2], argv[3:]
     ts = time.strftime("%Y-%m-%dT%H:%M:%SZ", time.gmtime())
+    try:
+        return _run(path, op, rest, ts)
+    except LockUnavailable as exc:
+        # 3, deliberately not 1 and never 0. For claim-flag, 1 already means
+        # "already claimed, stay quiet" and 0 means "claimed, go page" -- a lock
+        # failure is neither. Nothing was claimed, so the next run can still
+        # claim it; answering 0 would page off a flag no file records, and
+        # answering 1 would retire a page that never fired.
+        sys.stderr.write("attempts-ledger: `%s` wrote nothing -- %s\n" % (op, exc))
+        return 3
 
+
+def _run(path, op, rest, ts):
     if op == "get":                       # get <issue> <key> [default]
         issue, key = rest[0], rest[1]
         default = rest[2] if len(rest) > 2 else "0"
--- a/q-system/.q-system/scripts/attempts-ledger.py
+++ b/q-system/.q-system/scripts/attempts-ledger.py
@@ -115,6 +115,7 @@ import os
 import sys
 import tempfile
 import time
+import traceback
 
 # TWO CAPS, whichever comes first. LOCK_TIMEOUT is what a caller actually cares
 # about -- "how long before this refuses" -- and it is measured, so it stays true
@@ -299,6 +300,26 @@ def main(argv):
         # answering 1 would retire a page that never fired.
         sys.stderr.write("attempts-ledger: `%s` wrote nothing -- %s\n" % (op, exc))
         return 3
+    except Exception:
+        # 2, and the catch-all is the point (codex round 2 on PR #67, minor 3).
+        # This block used to name only the two exceptions above, so an OSError
+        # from mkstemp / os.replace / json.dump on a full or read-only filesystem
+        # escaped as an uncaught exception -- and Python exits 1 on those. For
+        # claim-flag, 1 is the ONE code that means "already claimed on a prior
+        # run, stay quiet". So a write that failed hard read as a page already
+        # sent, for a flag no file records, which means no later run retires it
+        # either. That is the same silent drop findings 2 and 3 closed, arriving
+        # through the one door still left open.
+        #
+        # Enumerating OSError instead would leave the next unnamed exception on
+        # 1. The safe default for "something we did not anticipate" is the code
+        # that says NOTHING WAS WRITTEN, so the caller pages and says so.
+        # The traceback still goes to stderr, because a swallowed error here is
+        # how a real bug becomes a shrug.
+        traceback.print_exc()
+        sys.stderr.write(
+            "attempts-ledger: `%s` failed unexpectedly and wrote nothing (exit 2)\n" % op)
+        return 2
 
 
 def _args(op, rest, want):
--- a/q-system/.q-system/scripts/attempts-ledger.py
+++ b/q-system/.q-system/scripts/attempts-ledger.py
@@ -101,8 +101,27 @@ drops a page:
 
   0  the op succeeded. For claim-flag: claimed HERE, first time -- go page.
   1  claim-flag ONLY: already claimed on an earlier run -- stay quiet.
-  2  usage error: unknown op, or the wrong number of arguments. Nothing written.
-  3  the lock could not be taken. Nothing written, nothing claimed.
+  2  NOTHING WAS WRITTEN, and the cause is not contention. Two populations:
+     a usage error (unknown op, wrong arity), and -- since the round-3
+     catch-all in main() -- ANY unanticipated failure, e.g. an OSError from
+     mkstemp / os.replace / json.dump on a full or read-only filesystem.
+  3  NOTHING WAS WRITTEN because the lock could not be taken.
+
+READ 2 AS OPEN-ENDED, NOT AS "a typo in a call site" (codex round 4, minor).
+This line used to say only "usage error", which reads as a bounded, transient,
+fix-the-invocation class -- and linear-worker.sh's page_once cited that reading
+to justify paging anyway on 2 and 3, calling it "bounded". It is not bounded.
+A read-only filesystem exits 2 on every run forever, so the caller's page fired
+on every run forever, and at the stuck call site that page is a permanent Linear
+comment. The catch-all is still right (see main(): the safe default for the
+unanticipated is the code that says NOTHING WAS WRITTEN, never 1). What was
+wrong was a doc that undersold what it covers, so keep this description as wide
+as the except clause actually is.
+
+The difference callers act on is not 2-vs-3 but WHETHER A LATER RUN RECOVERS:
+3 leaves the flag unclaimed and the lock frees in milliseconds, so the next run
+claims and pages. 2 can persist indefinitely, so no later run claims or pages
+either.
 
 Arity is validated up front for exactly this reason: `claim-flag ASK-1` with the
 flag omitted used to raise IndexError and exit 1, and 1 is the one code that
--- a/q-system/.q-system/scripts/attempts-ledger.py
+++ b/q-system/.q-system/scripts/attempts-ledger.py
@@ -48,15 +48,28 @@ mechanism rather than the tuning:
      permanent, which is the failure class this ledger exists to kill. Every
      alternative is another heuristic with its own hole: an age cap breaks a slow
      holder and trusts the clock.
-  2. IT COST 22.1s TO REFUSE, against a 10s retry budget, because each of the 100
-     retries did a full mkstemp + write + link + unlink publish attempt.
-  3. RELEASE-BY-OWNER, ATOMIC PUBLISH, AND CORPSE-BREAKING ARE ~90 LINES OF
+  2. RELEASE-BY-OWNER, ATOMIC PUBLISH, AND CORPSE-BREAKING ARE ~90 LINES OF
      HEURISTIC standing in for something the kernel already guarantees.
 
-flock erases all three. The lock is the open file description, so the kernel
-releases it on close AND on process death: a dead holder blocks nothing, with no
-pid to guess at and no timer to trust. A contended retry is one non-blocking
-syscall, so the refusal costs its sleep budget and nothing more.
+flock erases both. The lock is the open file description, so the kernel releases
+it on close AND on process death: a dead holder blocks nothing, with no pid to
+guess at and no timer to trust.
+
+FINDING 4 IS A THIRD DEFECT AND IT IS NOT THE MECHANISM -- the retry budget was a
+COUNT PRETENDING TO BE A CLOCK. `LOCK_TRIES=100` x `LOCK_SLEEP=0.1` was
+documented as "the production 10s timeout" and measured 22.1s. The first guess,
+that the 100 mkstemp/link/unlink publish attempts were the overhead, was wrong:
+
+  $ python3 -c "import time; t=time.time()
+  > [time.sleep(0.1) for _ in range(100)]; print(time.time()-t)"
+  21.2
+
+`time.sleep(0.1)` costs 212ms on this kernel, so the sleeps alone were 21.2s of
+the 22.1. A count cannot express a duration on a kernel whose sleeps do not keep
+to their argument, and switching mechanism would not have changed that by one
+second. So the budget is a wall-clock DEADLINE now, and the number in the comment
+is enforced rather than asserted. LOCK_TRIES survives as a second cap, because
+the suite needs a way to make the contended path sub-second.
 
 THE INVARIANTS, and where each now lives:
 
@@ -103,10 +116,14 @@ import sys
 import tempfile
 import time
 
-# Overridable ONLY so the suite can assert the contended path without sleeping
-# out the full budget. Production never sets it, same posture as converge.sh's
+# TWO CAPS, whichever comes first. LOCK_TIMEOUT is what a caller actually cares
+# about -- "how long before this refuses" -- and it is measured, so it stays true
+# no matter what a sleep really costs. LOCK_TRIES only bounds the iteration count
+# so a wedged clock cannot spin; it is overridable ONLY so the suite can make the
+# contended path sub-second. Production sets neither, same posture as converge.sh's
 # KIPI_RECEIPT_LOCK_TRIES.
-LOCK_TRIES = int(os.environ.get("KIPI_ATTEMPTS_LOCK_TRIES") or 100)
+LOCK_TIMEOUT = float(os.environ.get("KIPI_ATTEMPTS_LOCK_TIMEOUT") or 10.0)
+LOCK_TRIES = int(os.environ.get("KIPI_ATTEMPTS_LOCK_TRIES") or 1000)
 LOCK_SLEEP = 0.1
 
 
@@ -141,7 +158,10 @@ def _take_lock(lock):
     """
     parent = os.path.dirname(lock) or "."
     os.makedirs(parent, exist_ok=True)
+    deadline = time.monotonic() + LOCK_TIMEOUT
     for attempt in range(LOCK_TRIES):
+        if attempt and time.monotonic() >= deadline:
+            return None
         # A DIRECTORY here is what the original `os.mkdir` lock leaves when its
         # run is killed. It cannot be opened, so the first instance to take this
         # upgrade with one on disk would never count an attempt again. It also
--- a/q-system/.q-system/scripts/attempts-ledger.py
+++ b/q-system/.q-system/scripts/attempts-ledger.py
@@ -17,57 +17,94 @@ bump_attempt is one of six, and the other five kept racing on the same file.
 Codex caught the claim. Hence ONE writer, here, rather than six copies of a lock
 -- six copies of a lock is just the original defect with more places to drift.
 
-CONCURRENCY. No flock: macOS ships none and this fleet runs on both kernels, so
-the mutex is a file whose creation is O_EXCL. The write is temp-then-os.replace
-so a crash mid-write cannot leave a truncated ledger, which would read as `{}`
-and silently reset every budget in the file.
+THE LOCK WAS WORSE THAN NO LOCK (ASK-286, sp-626e9452). It used to PROCEED on
+timeout -- break out of the retry loop and enter the transaction holding nothing
+-- and its release removed the lock BY PATH. Together: a run that times out
+believes it is serialized, walks in, and on the way out deletes the live lock of
+the run that actually holds it. Both runs are then inside one read-decide-write,
+which is precisely what this file exists to stop. It also propagated: on
+2026-08-02 an agent was told to reuse an in-repo lock rather than invent a
+fourth, copied THIS one including both defects, and cited the source as
+justification. Citing a pattern is not verifying it.
 
-THE LOCK WAS WORSE THAN NO LOCK (ASK-286, sp-626e9452, back-ported from the
-version PR #42 proved in converge.sh `receipt_transaction`). It used to PROCEED
-on timeout -- break out of the retry loop and enter the transaction holding
-nothing -- and its release removed the lock BY PATH. Together: a run that times
-out believes it is serialized, walks in, and on the way out deletes the live
-lock of the run that actually holds it. Both runs are then inside one
-read-decide-write, which is precisely what this file exists to stop.
+WHY fcntl.flock AND NOT A TOKEN FILE (codex round 1 on PR #67, findings 1 + 3).
+The first fix here back-ported converge.sh `receipt_transaction`: a token written
+into a file, published with os.link, released only if it still carried this run's
+token, and broken when the pid it named was not running. That is the right shape
+FOR BASH, where converge.sh lives -- macOS ships no flock(1) BINARY, so a shell
+script genuinely has to hand-roll one. It is the wrong shape here, and the
+docstring that justified it with a flat "macOS ships no flock" was simply false:
+this is Python, `fcntl.flock` works on both kernels this fleet runs on, and
+`session_recall.py` was already using it a few files away.
 
-It also propagated. On 2026-08-02 an agent was told to reuse an in-repo lock
-rather than invent a fourth, copied THIS one including both defects, and cited
-the source as justification. Citing a pattern is not verifying it.
+Keeping the hand-rolled version cost three separate defects, all of them the
+mechanism rather than the tuning:
 
-FOUR INVARIANTS, all absent from the version this replaces:
+  1. LIVENESS BY PID ANSWERS THE WRONG QUESTION. It asks "is some process with
+     this number running", not "is that process holding this lock". Pids wrap, so
+     a leftover lock eventually names a pid an unrelated live process now has --
+     and is then honoured forever. The attempts counter freezes, no issue reaches
+     MAX_ATTEMPTS, no issue reaches STUCK, and no human is paged. Silent and
+     permanent, which is the failure class this ledger exists to kill. Every
+     alternative is another heuristic with its own hole: an age cap breaks a slow
+     holder and trusts the clock.
+  2. IT COST 22.1s TO REFUSE, against a 10s retry budget, because each of the 100
+     retries did a full mkstemp + write + link + unlink publish attempt.
+  3. RELEASE-BY-OWNER, ATOMIC PUBLISH, AND CORPSE-BREAKING ARE ~90 LINES OF
+     HEURISTIC standing in for something the kernel already guarantees.
 
-  1. A TIMEOUT RETURNS FAILURE. The caller does no work and says so. A skipped
-     increment is recoverable -- the next scheduled run retries, and the run
-     that DID hold the lock is counting. Two runs inside one read-decide-write
-     is not recoverable.
-  2. A RELEASE REMOVES ONLY A LOCK CARRYING THIS RUN'S OWN TOKEN, never one
-     identified by path alone.
-  3. THE LOCK IS CREATED ALREADY CARRYING ITS OWNER. The token is written into a
-     temp file first and `os.link` publishes it at the lock path atomically, so
-     the lock never exists un-attributable for even an instant. mkdir left that
-     window open, and a run killed inside it left a lock nothing could ever own.
-  4. STALE LOCKS BREAK ON OWNER LIVENESS, so refusing to force does not let a
-     corpse wedge every future write. A lock whose pid is gone is broken; so is
-     one that carries no attributable owner at all -- an empty file, or the
-     leftover DIRECTORY a pre-fix worker killed mid-transaction leaves behind,
-     since by invariant 3 this writer can never produce either.
+flock erases all three. The lock is the open file description, so the kernel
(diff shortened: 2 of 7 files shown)


assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/pr-67-20260802-085114.md` (7810 bytes).

```
FINDINGS:
major|A leftover lock whose pid is reused by any live process permanently freezes the attempts counter, so no issue ever reaches STUCK and no human is paged|q-system/.q-system/scripts/attempts-ledger.py:116
minor|Exit 3, exit 1 and exit 2 collapse into the same branch at all six claim_page_once call sites, so a failed ledger write reads as "already claimed, stay quiet"|q-system/.q-system/scripts/attempts-ledger.py:261
minor|The docstring justifies the hand-rolled lock with "macOS ships no flock", but fcntl.flock works on this kernel and session_recall.py:125 already uses it|q-system/.q-system/scripts/attempts-ledger.py:20
minor|The test comment documents a "production 10s timeout"; a blocked mutation actually takes 22.8s|q-system/.q-system/scripts/test/test-attempts-ledger.py:68
END FINDINGS
```

--- reviewer output ---

# Review: PR #67 — attempts-ledger lock (ASK-286)

The core fix is real and the mutation-kill is genuine. One regression in recoverability survives.

---

## MAJOR — a leftover lock whose pid gets reused freezes the attempts counter permanently, and nothing says so

`q-system/.q-system/scripts/attempts-ledger.py:116` (`_owner_is_alive`)

Invariant 4 breaks a stale lock on **pid liveness only**. There is no age fallback and no start-time comparison. So a lock left by a SIGKILLed run — the exact condition the docstring says invariant 4 exists for — stops being breakable the moment that pid is recycled onto any live process. Every mutation then raises `LockUnavailable` forever. The counter that feeds `MAX_ATTEMPTS` never advances, so `linear-worker.sh:763` never reaches the STUCK branch, never pages, and keeps dispatching the failing issue. That is the runaway this file exists to stop, now silent and self-perpetuating.

The PR's RESIDUAL paragraph (attempts-ledger.py:55-60) names pid reuse and claims it "degrades in the SAFE direction". The repro falsifies that: it degrades into a frozen detector.

Reproducer — a real live process, not a mocked one:

```bash
sleep 300 &  VICTIM=$!
echo "$VICTIM:0:0" > "$P.lock"     # what a leftover lock looks like after pid wrap
for i in 1 2 3; do python3 attempts-ledger.py "$P" bump-attempt ASK-1 "run $i"; done
```

Real output:

```
leftover lock now names pid 78143 (a live, unrelated process)
cycle 1: exit=3  elapsed=22s  count=2
cycle 2: exit=3  elapsed=23s  count=2
cycle 3: exit=3  elapsed=22s  count=2
MAX_ATTEMPTS is 3. count is still 2 -> the STUCK branch at linear-worker.sh:763 never fires.

=== same state, the code this PR replaces (git show origin/main) ===
old code: exit=0  count=3  -> reaches MAX_ATTEMPTS, pages
```

Blast radius: budget burn on an issue that already failed, forever, with no page. Human fix at 3am: `rm ~/.config/kipi/state/linear-worker-attempts.json.lock`. Nothing in the repo detects a wedged lock (grepped — no sweeper).

Honesty caveat: I set the token's pid value rather than waiting for a real wraparound. The token **shape** is exactly what `_new_token` emits (`<pid>:<epoch>:<rand>`), and the live process is real, so this is not an invented fixture — but the triggering OS event is simulated. Weigh it accordingly.

Smallest fix: break a lock older than `LOCK_TRIES * LOCK_SLEEP * K`, or store the owner's process start time alongside the pid and compare.

---

## MINOR — exit 3, exit 1 and exit 2 are one branch at every caller

`q-system/.q-system/scripts/attempts-ledger.py:261-268` and all six `claim_page_once` sites (`linear-worker.sh:633, 647, 763, 910, 932, 1065`)

The comment argues at length that 3 is "deliberately not 1 and never 0". No caller can tell them apart: every site is `if claim_page_once ...; then`. A ledger that failed to write reads as "already claimed, stay quiet" — which is the exact `clear-automerge`-exiting-2 scar recorded at `attempts-ledger.py:296-302`. The source this was back-ported from does better: `converge.sh:566` sets `RECEIPT_MISS` + `RECEIPT_FIX` on lock failure. The back-port dropped that half.

```
lock-unavailable (exit 3): STAYS QUIET
already-claimed  (exit 1): STAYS QUIET
unknown op       (exit 2): STAYS QUIET
```

Also reproduced the non-`LockUnavailable` write failure landing on 1 (traceback path, ledger path made a directory — a constructed state, hence the downgrade):

```
IsADirectoryError: [Errno 21] Is a directory: ... -> .../a.json
claim-flag exit=1
CALLER (linear-worker.sh:763 shape): STAYS QUIET -- read as already-claimed
```

---

## MINOR — "macOS ships no flock" is false, and this repo already contradicts it

`q-system/.q-system/scripts/attempts-ledger.py:20`

The justification for the entire hand-rolled lock is a factual claim about the kernel. `fcntl.flock` works on this machine, and `q-system/.q-system/scripts/session_recall.py:125` already uses it for the same read-merge-write shape.

```
Darwin 25.3.0 arm64
parent: fcntl.flock LOCK_EX acquired on macOS -> True
child: BLOCKED by the parent's flock -> mutual exclusion WORKS on this kernel
no flock(1) binary (shell-level), but fcntl.flock(2) is present
```

What macOS lacks is the `flock(1)` **shell utility** — true for `converge.sh`, not for a Python writer. The lock as built is fine; the stated reason for building it is wrong, and the next person citing this file (which the PR body says already happened once) inherits the wrong reason.

---

## MINOR — documented timeout is 10s, measured is 22.8s

`q-system/.q-system/scripts/test/test-attempts-ledger.py:68` — "the production 10s timeout".

`_take_lock` does 100 × (mkstemp + link + unlink + two `_read_lock` reads + `sleep 0.1`). Measured on a live-held lock: `measured blocked-mutation wall clock: 22.8s`. Six ledger ops per issue per cycle means a contended heartbeat costs minutes, not seconds.

---

## What is sound — attacks the code survived

- **24-way concurrent `bump-attempt`.** `final count: 24 (expected 24)`, no leftover lock, no stray temp files. No lost increments.
- **The mutation kill is real.** I rebuilt it independently against a copy with the ownership check deleted: `MUTANT KILLED by case 3`. Case 3 is not decorative — it is the only shape in the suite that can fail a release-by-path, exactly as the PR claims.
- **The suite fails on the old code.** Verified `origin/main`'s version proceeds-on-timeout; case 1's assertion is a real red.
- **Fixture shapes match the producer.** Case 1/2/4 write `<pid>:<int>:<int>` — identical to `_new_token`. No key no producer emits.
- **`os.link` publish window.** Tried to observe a lock present without an owner; couldn't. The hardlink genuinely closes the `mkdir` window.
- **Unwritable parent dir** correctly reaches `LockUnavailable` (exit 3) rather than proceeding unlocked.
- **Test isolation.** `parents[4]` resolves correctly, every case runs in its own tempdir, the live `linear-worker-attempts.json` is untouched. Added to `capability-manifest.json`.

Attempts I could not land, and am dropping: the `_break_if_dead` re-read TOCTOU (the docstring names it; I could not produce it without patching `_read_lock`, so it stays a documented residual, not a finding), and ENOSPC as a natural trigger for the exit-1 collision (ram-disk mount blocked in this sandbox).

Note, not a finding: the PR body cites `q-system/output/mutcheck-ask-286.py` as evidence. That path is gitignored and absent — no reviewer can re-run it. The checked-in suite carries the kill on its own, so this costs nothing.

---

## VERDICT: REQUEST CHANGES

One major finding. Not a BLOCK — nothing here is permanent or unrecoverable, and the change is a clear net improvement over what is on `main`.

**Fix first:** the pid-only liveness test at `attempts-ledger.py:116`. Add an age or process-start-time fallback so a recycled pid cannot wedge the attempts counter forever. Everything else can ride a follow-up.

FINDINGS:
major|A leftover lock whose pid is reused by any live process permanently freezes the attempts counter, so no issue ever reaches STUCK and no human is paged|q-system/.q-system/scripts/attempts-ledger.py:116
minor|Exit 3, exit 1 and exit 2 collapse into the same branch at all six claim_page_once call sites, so a failed ledger write reads as "already claimed, stay quiet"|q-system/.q-system/scripts/attempts-ledger.py:261
minor|The docstring justifies the hand-rolled lock with "macOS ships no flock", but fcntl.flock works on this kernel and session_recall.py:125 already uses it|q-system/.q-system/scripts/attempts-ledger.py:20
minor|The test comment documents a "production 10s timeout"; a blocked mutation actually takes 22.8s|q-system/.q-system/scripts/test/test-attempts-ledger.py:68
END FINDINGS

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/pr-67-20260802-090936.md` (9759 bytes).

```
FINDINGS:
major|A leftover lock whose pid is reused by a live process freezes the attempts counter, so no issue reaches STUCK and no human is paged|q-system/.q-system/scripts/attempts-ledger.py:116
minor|_break_if_dead can delete a live run's lock and the re-read guard is a no-op for unattributable locks, falsifying the docstring claim at line 59|q-system/.q-system/scripts/attempts-ledger.py:134
minor|Exit 3, exit 1 and exit 2 collapse into one branch at all six claim_page_once call sites, so a failed ledger write reads as already-claimed|q-system/.q-system/scripts/attempts-ledger.py:268
minor|The docstring justifies the hand-rolled lock with "macOS ships no flock", but fcntl.flock works on this kernel and session_recall.py:126 already uses it|q-system/.q-system/scripts/attempts-ledger.py:21
END FINDINGS
```

--- reviewer output ---

# Review: PR #67 — attempts-ledger lock (ASK-286), round 2

Process note first, because it changes how you read the rest: **`git log` HEAD is still `96919eb`, the same commit round 1 reviewed. No code changed between rounds.** So every round-1 finding is still live by construction. I re-verified each one myself rather than inheriting it, dropped one that did not reproduce, and found one new thing round 1 explicitly gave up on. Round 1's calibration was sound — I am not producing new blockers on unchanged code.

---

## MAJOR — a leftover lock whose pid gets recycled freezes the counter, and nothing pages

`q-system/.q-system/scripts/attempts-ledger.py:116` (`_owner_is_alive`)

Raised in round 1. Still live. My own repro, with a real live process:

```
leftover lock now names pid 91992, a live unrelated process
token shape written: '91992:0:0'  (_new_token emits pid:epoch:rand)
cycle 1: exit=3 elapsed=11.1s count=2
cycle 2: exit=3 elapsed=11.1s count=2
cycle 3: exit=3 elapsed=11.1s count=2
MAX_ATTEMPTS is 3; count is still 2, so the STUCK branch at linear-worker.sh:763 never fires
victim killed -> exit=0 count=3 (self-heals only when that pid dies)
```

Same state against the code this PR replaces:

```
PRE-FIX code, same wedge state: exit=0 elapsed=10.7s count=3
```

This is a recoverability regression, not just a residual. The docstring at `attempts-ledger.py:55-60` says pid reuse "degrades in the SAFE direction". It degrades into a frozen detector: the issue is redispatched every heartbeat, the count never advances, and `claim_page_once` also exits 3, so the STUCK page never fires either. Silent budget burn is the exact failure this file's header says it exists to stop.

It self-heals when the squatting pid dies. If the recycled pid lands on something long-lived (a launchd agent, a `claude -p` session, the worker itself), it does not heal.

3am human action: `rm ~/.config/kipi/state/linear-worker-attempts.json.lock`. Nothing in the repo detects a wedged lock — I grepped `q-system/.q-system/scripts/` for stale-lock sweepers and exit-3 consumers, none exist.

Smallest fix: store the owner's process start time next to the pid, or add an age fallback so a lock older than `LOCK_TRIES * LOCK_SLEEP * K` breaks regardless.

---

## MINOR — the break path CAN delete a live run's lock, which the docstring says it cannot

`attempts-ledger.py:134-147` (`_break_if_dead`), claim falsified at `attempts-ledger.py:59`

Round 1 dropped this as unreproducible. Two new things.

First, it reproduces. I injected a delay at `os.path.isdir` — a no-op read that sits between the re-read guard and the unlink. No logic changed; it only lets another run interleave there:

```
corpse pid (re-read guard on)  live-holder lock survived? False *** C DELETED A LIVE RUN LOCK ***
unattributable empty file      live-holder lock survived? False *** C DELETED A LIVE RUN LOCK ***
leftover mkdir directory       live-holder lock survived? False *** C DELETED A LIVE RUN LOCK ***
```

Second, and this is the part round 1 missed: **for an unattributable lock the re-read guard is structurally a no-op.** `_read_lock` (line 84) returns `""` for a directory, an empty file, *and* a missing file. So when `seen == ""`, the guard `if _read_lock(lock) != seen: return` compares `"" != ""` and always passes — even after another run has already removed the corpse and published its own live lock. The mitigation the docstring names does not exist on that branch.

Line 59 states: "Neither can silently delete a live run's lock, which is the property that was actually broken." The code can. Where the comment and the code disagree, the code is the truth.

Severity held at minor, and here is why I did not inflate it: I could not hit it with real processes. 20-way concurrent `bump-attempt` and 8-way spin-barrier-synced runs, across clean / leftover-directory / empty-file locks, 33 rounds:

```
clean  spin-barrier synced, 8 procs: 0/15 rounds lost an increment
empty  spin-barrier synced, 8 procs: 0/15 rounds lost an increment
dir    spin-barrier synced, 8 procs: 0/15 rounds lost an increment
```

The window is microseconds. The docstring claim is the defect; fix the sentence or fix the guard.

---

## MINOR — exit 3, exit 1 and exit 2 are one branch at every caller

`attempts-ledger.py:268`, all six sites at `linear-worker.sh:633, 647, 763, 910, 932, 1065`

Raised in round 1, still live. Every site is `if claim_page_once ...; then`:

```
633:      if claim_page_once "$ISSUE" automerge_unknown_paged; then
647:      if claim_page_once "$ISSUE" automerge_unarmed_paged; then
763:    if claim_page_once "$ISSUE" stuck_paged; then
910:        if claim_page_once "$ISSUE" drift_paged; then
932:        if claim_page_once "$ISSUE" conflict_paged; then
1065:      if claim_page_once "$ISSUE" tree_paged; then
```

The comment at line 262-267 argues at length that 3 is "deliberately not 1 and never 0". No caller can tell 3 from 1. A ledger that failed to write reads as "already claimed, stay quiet".

A non-`LockUnavailable` write failure is worse — it lands on 1 directly:

```
claim-flag exit=1
last stderr line: IsADirectoryError: [Errno 21] Is a directory: ... -> .../a.json
caller shape `if claim_page_once ...; then page; fi`  ->  STAYS QUIET
```

The source this was back-ported from does the other half: `converge.sh:566-567` sets `RECEIPT_MISS` + `RECEIPT_FIX` on lock failure, so the miss reaches a human. The back-port took the refusal and dropped the reporting.

---

## MINOR — "macOS ships no flock" is false, and this repo already contradicts it

`attempts-ledger.py:21`

Raised in round 1, still live, verified independently:

```
Darwin 25.3.0
parent: fcntl.flock LOCK_EX acquired on this kernel -> True
child: BLOCKED by parent flock -> mutual exclusion works
```

```
q-system/.q-system/scripts/session_recall.py:126:  fcntl.flock(lock_fh.fileno(), fcntl.LOCK_EX)
```

What macOS lacks is the `flock(1)` shell utility — true for `converge.sh`, not for a Python writer. The lock as built is fine; the stated reason is wrong, and the PR body itself documents that someone already copied this file citing its reasoning.

---

## DROPPED — round 1's "documented 10s, measured 22.8s"

I could not reproduce it. Three blocked mutations, production defaults, measured `11.1s / 11.1s / 11.1s` against a documented 10s timeout. That is ~10% overhead, not a misstatement. Dropping it, and it should not carry into a follow-up.

---

## What is sound

- **Real concurrency, three lock states.** 20 concurrent `bump-attempt` processes × 6 rounds each on clean / leftover-mkdir-directory / empty-file locks: `0/6 rounds lost an increment` in all three. Then 8 processes synced on a spin barrier × 15 rounds each, same three modes: `0/15` each. Final count matched exit-0 count every time. No lost increments, no leftover locks, no stray temp files.
- **The suite is not decorative — five mutants, five kills.** I rebuilt the mutation check independently (the `q-system/output/mutcheck-ask-286.py` the PR body cites is absent from the tree, so nobody can re-run the author's):
  ```
  M1 proceed-on-timeout (old defect)     KILLED
  M2 release-by-path (old defect)        KILLED
  M3 never break a stale lock            KILLED
  M4 every lock looks alive              KILLED
  M5 lock failure answers exit 0         KILLED
  ```
  Case 3 is the real mutation-killer the PR claims it is — M2 survives nothing else in the suite.
- **Fixture shape matches the producer.** Cases 1, 2 and 4 write `<pid>:<int>:<int>`, byte-identical in shape to what `_new_token` emits. No invented key.
- **The `os.link` publish window.** Tried to observe a lock present without an owner. Could not. The hardlink genuinely closes the `mkdir` gap that invariant 3 targets.
- **The `set -e` claim at `linear-worker.sh:492` checks out.** `linear-worker.sh:47` is `set -uo pipefail`, no `-e`. An exit 3 is a reported miss, not a dead worker. I checked the two `bump_attempt` call sites (1302, 1795) and the `clear_automerge_pages` site (663) — none is a function's last statement, none is in an `&&` chain that propagates.
- **The test is actually wired.** `capability-gate.py:408-422` executes every `expected_tests` entry with its declared runner, and `:261-267` fails on a present-but-undeclared test file. The manifest entry is real enforcement, not a listing.
- **Both cited suites are green.** `test-terminal-states.sh` → `14 passed, 0 failed`. `test-linear-worker-parallel.sh` → `PASS (9 checks)`.

---

## VERDICT: REQUEST CHANGES

One major. Not a BLOCK — nothing here is permanent or unrecoverable, and the change is a clear net improvement over `main`.

**Fix first:** the pid-only liveness test at `attempts-ledger.py:116`. Add process-start-time or an age fallback so a recycled pid cannot freeze the attempts counter. Everything else can ride a follow-up.

FINDINGS:
major|A leftover lock whose pid is reused by a live process freezes the attempts counter, so no issue reaches STUCK and no human is paged|q-system/.q-system/scripts/attempts-ledger.py:116
minor|_break_if_dead can delete a live run's lock and the re-read guard is a no-op for unattributable locks, falsifying the docstring claim at line 59|q-system/.q-system/scripts/attempts-ledger.py:134
minor|Exit 3, exit 1 and exit 2 collapse into one branch at all six claim_page_once call sites, so a failed ledger write reads as already-claimed|q-system/.q-system/scripts/attempts-ledger.py:268
minor|The docstring justifies the hand-rolled lock with "macOS ships no flock", but fcntl.flock works on this kernel and session_recall.py:126 already uses it|q-system/.q-system/scripts/attempts-ledger.py:21
END FINDINGS

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/pr-67-20260802-092225.md` (8613 bytes).

```
FINDINGS:
major|page_once() turns errexit on for the rest of linear-worker.sh, so the queue drain dies after the first page while the worker still exits 0 and says "run complete"|q-system/.q-system/scripts/linear-worker.sh:574
minor|Neither new test executes page_once; the routing case reimplements it as a local route() copy, so the function this PR wired to six call sites has zero executed coverage|q-system/.q-system/scripts/test/test-claim-page-once-routing.sh:103
minor|An OSError on the ledger write path escapes the try/except and exits 1, the one code that means "already claimed, stay quiet", silently dropping the page|q-system/.q-system/scripts/attempts-ledger.py:294
minor|The chokepoint comment still describes round 1's O_EXCL token lock and repeats the "macOS ships no flock" claim that the ledger docstring retracts in the same PR|q-system/.q-system/scripts/linear-worker.sh:479
END FINDINGS
```

--- reviewer output ---

## Review round 3, PR #67

Process note up front, because the instructions ask for it: **rounds 1 and 2 both reviewed commit `96919eb`, and round 2 said so itself.** All four of their findings were then addressed by `2a41a7b` (flock replaces the hand-rolled token lock) and `05332d1` (wall-clock deadline). I verified each fix rather than inheriting it. Every finding below is on code that did not exist in rounds 1 or 2. **This is not a PR producing new blockers on unchanged code. The earlier rounds were calibrated.**

---

### MAJOR — `page_once()` turns on `errexit` for the rest of the worker, and the queue drain dies mid-run while the worker reports success

`q-system/.q-system/scripts/linear-worker.sh:574`

The worker's prologue is `set -uo pipefail` (line 47, no `-e`). `page_once` does `set +e` / call / `set -e`, and `set -e` is global shell state, not function-scoped. So the **first** page in a run (stuck, drift, conflict, tree, or either auto-merge shape) flips errexit ON for everything after it.

Repro, controlled against the real functions lifted verbatim out of `linear-worker.sh`:

```
$ bash /var/folders/.../repro.sh          # one page_once call
shell flags BEFORE: huB
  -> paged (first claim)
shell flags AFTER : ehuB                  # <-- 'e' appeared
processing ASK-2
EXIT=1                                    # loop died, ASK-3 never seen

$ bash /var/folders/.../control.sh        # identical, page_once never called
shell flags BEFORE: huB
shell flags AFTER : huB
processing ASK-2 /  finished ASK-2 / processing ASK-3 /  finished ASK-3
LOOP COMPLETED
EXIT=0
```

The 3am consequence, using the worker's real queue shape (pipeline into `while read`) and its real last two lines:

```
$ bash /var/folders/.../repro_queue.sh
processing ASK-1
say: worker: run complete
WORKER EXIT=0
```

ASK-2 and ASK-3 were never processed. The worker printed `worker: run complete` and exited **0**, which its own header documents as "ran (or had nothing to run). A caller may treat this as healthy." A partial queue drain reported as healthy is the silent-stall class this file exists to kill.

The comment this PR added at `linear-worker.sh:491` states the opposite and is falsified by the code in the same commit:

> "None of the callers below run under `set -e`, so a 3 is a reported miss on stderr, not a dead worker."

```
$ bash /var/folders/.../repro_bump.sh
--- WITHOUT a prior page_once (what the comment describes) ---
    attempts-ledger: `bump-attempt` wrote nothing -- another run still holds the lock...
    worker still alive: yes
--- WITH one prior page_once (the shape the PR just shipped) ---
    flags now: ehuB
    attempts-ledger: `bump-attempt` wrote nothing -- ...
EXIT=3                                    # "WORKER STILL ALIVE" never printed
```

Fix is one deletion: drop both `set +e` and `set -e` from `page_once`. `rc=$?` already captures the status, and the script never had errexit to restore.

---

### MINOR — neither new test ever executes `page_once`, which is why the above shipped green

`q-system/.q-system/scripts/test/test-claim-page-once-routing.sh:103`

The suite lifts `claim_page_once` (`awk '/^claim_page_once\(\)/{print; exit}'`, line 46) and then **reimplements** the routing as a local `route()` at line 103. `route()` is the author's copy of `page_once`, written from the same mental model, and it happens to omit the `set -e`. Case 4 only greps that no `if claim_page_once ` call site survives. The function this PR added and wired to six call sites has zero executed coverage.

```
$ bash q-system/.q-system/scripts/test/test-claim-page-once-routing.sh
  ok: the first claim pages and the second stays quiet
  ok: a failed ledger write is distinguishable from 'already claimed' (rc=3)
  ok: page / quiet / page-and-warn are three distinct routes, not one
  ok: no call site uses the bare `if claim_page_once` shape that collapses exit 3
PASS (4 checks)
```

Green, with the major above live in the file it tests. Pointing case 3 at the real `page_once` (`sed -n '/^page_once() {/,/^}/p'`) and asserting `$-` is unchanged after the call makes it red.

---

### MINOR — a ledger write that fails with `OSError` exits **1**, which is the one code that means "already paged, stay quiet"

`q-system/.q-system/scripts/attempts-ledger.py:294`

`_run` is wrapped for `Usage` and `LockUnavailable` only. Anything else (`mkstemp`, `os.replace`, `json.dump` on a full or read-only filesystem) escapes as an uncaught exception, and Python exits 1. `page_once` routes 1 to `return 1`, so the caller stays quiet about a flag no file records, and no later run retires it.

```
returncode: 1
stderr tail: PermissionError: [Errno 13] Permission denied: '/var/.../tmp4a2n14or'
flag actually recorded: {}
```

Round 1 tried this exact collision and dropped it ("ENOSPC as a natural trigger for the exit-1 collision, ram-disk mount blocked in this sandbox"). I landed it with a read-only ledger directory. **Held at minor, not raised:** my trigger is constructed (`chmod 555`); the natural producer is a full disk, which breaks plenty else at the same time. A bare `except Exception: return 2` on the same try block closes it.

---

### MINOR — the chokepoint comment still describes round 1's lock, and repeats the falsehood the ledger's own docstring retracts

`q-system/.q-system/scripts/linear-worker.sh:479`

```
# The lock is an O_EXCL file created already carrying its owner's token -- macOS
# ships no flock, so the portable primitive is the only one that works on both
# kernels this fleet runs on.
```

The lock is `fcntl.flock` as of `2a41a7b`. There is no token and no `O_EXCL`. And `attempts-ledger.py:60` in this same PR says the "macOS ships no flock" line "was simply false." Verified on this kernel:

```
$ python3 q-system/.q-system/scripts/test/test-attempts-ledger.py
  ok: a live holder blocks the write and its death releases the lock with no heuristic
PASS (12 cases)
```

This is the comment a future agent cites when it is told to reuse an in-repo lock. That is the documented propagation path in this PR's own body.

---

## What is sound

Attacks I ran that the code survived, and I am dropping:

- **Proceed-on-timeout.** Tried to get `_mutate` into the transaction while a live process held the flock. Could not. Exit 3, nothing written, message on stderr.
- **Two inodes at one path.** Case 9's fixture is genuinely adversarial: it monkeypatches `mod.open` to unlink and recreate the lock *inside* the acquisition window. That is a real producer shape, not an author-invented one, and `_same_file` catches it.
- **Corpse lock (rounds 1+2's major).** Fixed at the mechanism, not with a heuristic. Kernel drops the flock on process death; there is no pid to guess at. Cases 7 and 8 both bite.
- **Count-vs-clock (round 1 finding 4).** Genuinely fixed. Case 12 measured 2.2s against a 2.0s budget, and it is a real assertion, not a comment.
- **Exit-code collapse (round 1 finding 2).** Fixed on the ledger side. All five bad-arity and unknown-op shapes exit 2, `grep -cE '^\s*if claim_page_once '` returns 0.
- **Release-by-path mutant.** Case 3 stamps a foreign token from inside the critical section; the flock release touches no name at all.

---

## VERDICT: REQUEST CHANGES

One major, recoverable, so not a BLOCK. The lock rewrite itself is a clear net improvement over `main` and I could not break it. The defect is entirely in the shell wrapper that routes its exit codes.

**Fix first:** delete `set +e` and `set -e` from `page_once` (`linear-worker.sh:571,574`), then point `test-claim-page-once-routing.sh` case 3 at the real `page_once` instead of its `route()` copy so the leak has a test that can fail.

FINDINGS:
major|page_once() turns errexit on for the rest of linear-worker.sh, so the queue drain dies after the first page while the worker still exits 0 and says "run complete"|q-system/.q-system/scripts/linear-worker.sh:574
minor|Neither new test executes page_once; the routing case reimplements it as a local route() copy, so the function this PR wired to six call sites has zero executed coverage|q-system/.q-system/scripts/test/test-claim-page-once-routing.sh:103
minor|An OSError on the ledger write path escapes the try/except and exits 1, the one code that means "already claimed, stay quiet", silently dropping the page|q-system/.q-system/scripts/attempts-ledger.py:294
minor|The chokepoint comment still describes round 1's O_EXCL token lock and repeats the "macOS ships no flock" claim that the ledger docstring retracts in the same PR|q-system/.q-system/scripts/linear-worker.sh:479
END FINDINGS

assafkip:
## Codex round 2: all four findings fixed, each reproduced RED first

**MAJOR — `page_once` leaked errexit.** Confirmed exactly as described. The worker runs `set -uo pipefail` (line 47) and never had `-e`, so the trailing `set -e` enabled it rather than restoring it.

Observed before the fix, with the worker's real flags and real queue shape:

```
flags at loop start: huB
processing ASK-1
worker: run complete
WORKER EXIT=0
```

1 of 3 drained, reported healthy. After:

```
processing ASK-1 / finished ASK-1 ... ASK-3 / worker: run complete
WORKER EXIT=0
```

Fixed with `claim_page_once "$1" "$2" || rc=$?` rather than the bare call you suggested. Same size, same deletion of both `set` lines, and inside a `||` list errexit is suspended, so it reads the code correctly whether or not a future caller has `-e` while still touching no flags.

**MINOR — the test reimplemented its subject.** Correct, and it is why the major shipped green. Case 3 now lifts the real body via `sed -n '/^page_once() {/,/^}/p'`. Two cases added:

- **case 5** asserts `$-` is unchanged across the call. It has to fork: the suite runs `set -euo pipefail`, so in-process the leak is invisible. Pre-fix `huB -> ehuB`.
- **case 6** asserts the consequence, since flags only matter for what they do to the drain: a 3-issue queue still drains 3. Pre-fix: 1.

**MINOR — OSError exits 1.** Landed. Catch-all returns 2 with the traceback still on stderr; enumerating `OSError` would leave the next unnamed exception on 1.

Worth flagging: my first fixture for this was wrong and passed for the wrong reason. A read-only directory stops the run at lock *creation*, which already answers 3 — so it tested the lock path and would have passed against an implementation with none of this handling. Pre-creating the lock file puts the failure after acquisition, where you found it:

```
exit 1, PermissionError: [Errno 13] .../readonly-ledger/tmpx075dzpv
```

**MINOR — stale chokepoint comment.** Rewritten to describe `fcntl.flock`, including why the "macOS ships no flock" line was false (true of the `flock(1)` binary, which is why converge.sh genuinely hand-rolls one). Kept your framing that this is the paragraph a future agent cites.

### Runs

```
python3 test-attempts-ledger.py        -> PASS (13 cases)
bash test-claim-page-once-routing.sh   -> PASS (6 checks)
bash test-linear-worker-parallel.sh    -> PASS (9 checks)
bash test-terminal-states.sh           -> 14 passed, 0 failed
bash -n linear-worker.sh               -> OK
```

### Known gap, captured not hidden

`mutcheck-ask-286.py` carries 7 mutants covering the round-1 fixes and does not yet cover these two. The pre-fix RED observation is the equivalent evidence for each (the unfixed code *was* the mutant), but the standing script lags. Captured as spillover rather than left in a comment.

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/pr-67-20260802-093604.md` (8365 bytes).

```
FINDINGS:
major|page_once pages on exit 2 and 3, and the stuck call site's page is a permanent Linear comment, so contention posts it twice and an unwritable ledger posts it on every run forever|q-system/.q-system/scripts/linear-worker.sh:606
minor|The exit-code table says 2 is a usage error, but after the round-3 catch-all exit 2 is also every unexpected runtime failure, which is what makes the page-anyway arm look bounded|q-system/.q-system/scripts/attempts-ledger.py:104
END FINDINGS
```

--- reviewer output ---

## Review round 4, PR #67

Process note first. I read the code before the comments. Rounds 1 through 3 were **calibrated, not inflated**: every finding they raised is fixed and I verified the fixes rather than inheriting them. Round 1/2's pid-liveness major is gone at the mechanism (flock, not a heuristic). Round 3's errexit leak is gone and now has two forked test cases that actually go red. I re-raise nothing.

One finding below is new, on code round 3 read closely enough to find a shell-flag bug inside. Round 3 audited `page_once`'s *flags* and never audited its `*)` arm's blast radius. That is a coverage gap in round 3, not a miscalibration.

---

### MAJOR — `page_once` pages on exit 2 and 3, and at the stuck call site "page" writes a permanent Linear comment, so one stuck event posts it twice and an unwritable ledger posts it forever

`q-system/.q-system/scripts/linear-worker.sh:606` (the `*)` arm), consequence at `q-system/.q-system/scripts/linear-worker.sh:824-825`

Five of the six `page_once` sites do `bash "$NOTIFY"` only: a Slack ping, transient, cheap to duplicate. The sixth does not:

```
824:    if page_once "$ISSUE" stuck_paged; then
825:      python3 "$SYNC" progress "$ISSUE" \
826:        "**Stuck after $N/$MAX_ATTEMPTS attempts. ..."
```

`linear-sync.py progress` is `commentCreate`. Exit 2 and 3 both mean *nothing was claimed*, so the flag stays unset and the next run claims it and posts again.

**Trigger 1, real producer.** A second worker holds the ledger lock. Parallel workers are a supported, tested mode (`test-linear-worker-parallel.sh`). Real functions lifted verbatim from `linear-worker.sh`:

```
--- run 1: ledger lock held by a concurrent worker ---
    rc=0  stderr='attempts-ledger: `claim-flag` wrote nothing -- another run still holds the lock at .../attempts.json.lock. NOT '
--- run 2: next scheduled run, lock free ---
    rc=0  stderr=''
--- run 3: flag now claimed ---
    rc=0  stderr=''

PERMANENT LINEAR COMMENTS created for ONE stuck event:
  1. commentCreate ASK-999: **Stuck after 3/3 attempts**
  2. commentCreate ASK-999: **Stuck after 3/3 attempts**
slack pages: 2
ledger: {"ASK-999": {"stuck_paged": true}}
```

**Trigger 2, unbounded.** Make the ledger's state dir unwritable (a full disk is the natural producer; round 3 said so about the sibling exit-1 case). Now `claim-flag` exits 2 on every run and nothing is ever recorded:

```
scheduled run 1: rc=0  say: WARN: the attempts ledger did not record the stuck_paged flag for ASK-999 (exit 2) -- paging anyway...
scheduled run 2: rc=0  say: WARN: ... (exit 2) -- paging anyway...
scheduled run 3: rc=0  say: WARN: ... (exit 2) -- paging anyway...
scheduled run 4: rc=0  say: WARN: ... (exit 2) -- paging anyway...
scheduled run 5: rc=0  say: WARN: ... (exit 2) -- paging anyway...

permanent Linear comments on ASK-999 : 5
slack pages                          : 5
ledger state recorded                : {}
```

The comment at `linear-worker.sh:580` states the boundedness argument the design rests on:

> "It is also bounded: contention lasts a retry budget, and the first run that gets the lock claims the flag, after which every run is quiet."

That is true of exit 3. It is false of exit 2, and the `*)` arm routes both through one branch. Where the comment and the code disagree, the code is the truth.

The file's own principle, 12 lines above the call site at `linear-worker.sh:814`, is the one being broken: *"Paged ONCE via the shared claim_page_once flag, not per heartbeat: a repeated 'still stuck' every 15 minutes is the cry-wolf failure that trains the reader to mute the channel."* And at `linear-worker.sh:869`: *"a 'Picked up' note on a permanent Linear object followed by an immediate skip is a false alarm, and false alarms train the reader to ignore the real notes."*

3am human action: mute the worker channel, then delete N duplicate comments off the issue, then find the disk. The comment tells the human to rewrite the DoR, which is the wrong diagnosis when the real cause is that the ledger cannot be written.

**Calibration, stated plainly.** Trigger 1 uses a real producer shape and is bounded to one duplicate. Trigger 2 is unbounded but I produced it with `chmod 555` (the same fixture shape this PR's own suite case 13 accepts). I am calling it major on the combination, not on either half alone.

Smallest fix: split the `*)` arm. Exit 3 keeps paging (bounded, correct). For the one site that writes a permanent Linear object, gate the `$SYNC progress` call on a real claim and let the Slack ping carry the unclaimed case.

---

### MINOR — the exit-code table says 2 is a usage error, but after round 3's catch-all it is also every unexpected runtime failure

`q-system/.q-system/scripts/attempts-ledger.py:104`

```
attempts-ledger.py:104 says -> 2  usage error: unknown op, or the wrong number of arguments. Nothing written.
invocation      : claim-flag ASK-1 stuck_paged   (correct op, correct arity)
exit            : 2
real cause      : PermissionError: [Errno 13] Permission denied: '.../state/tmppvia7qn_'
ledger recorded : {}
```

The catch-all's own comment at line 303-322 explains this correctly, 200 lines below the authoritative table. The table is what a caller routes on, and it says exit 2 is a code bug, i.e. something that cannot arise unattended. That reading is exactly what makes the `*)` arm above look bounded. Fix the table line, not the code.

---

## What is sound

Attacks I ran that the code survived, and I am dropping:

- **Lost increments under real concurrency.** 32 concurrent `bump-attempt` processes: `exit codes {0: 32}`, `count recorded: 32`, `lost increments: 0`, no stray temp files, lock file present and 0 bytes. The core claim of the PR holds.
- **Claim-flag storm.** 32 concurrent `claim-flag` on one flag: `{1: 31, 0: 1}`. Exactly one winner, which is the whole once-only contract.
- **Both new suites are green and not decorative.** `test-attempts-ledger.py` → PASS (13 cases). `test-claim-page-once-routing.sh` → PASS (6 checks). Case 5 forks to observe `$-` under the worker's real flags, case 6 asserts the 3-of-3 drain. Round 3's "the test reimplements its subject" is genuinely closed: case 3 now `sed`s the real `page_once` body out of the worker.
- **`grep -cE '^\s*if claim_page_once '` on BSD grep.** I suspected case 4 was decorative on darwin because `\s` is a GNU extension. `/usr/bin/grep` returns the same count for `\s*` and `[[:space:]]*` on this kernel. Dropped.
- **All six call sites converted.** `grep -n claim_page_once` returns the definition, `page_once`'s own call, and three comment references. Zero bare call sites.
- **Unconditional `rmdir` of a leftover pre-flock lock DIRECTORY.** I built a mixed-fleet repro: a pre-ASK-286 worker holding its `mkdir` mutex, the new worker rmdir'ing it, a second old worker walking in. Dropped, because the old code proceeded on timeout anyway, so destroying its mutex changes nothing observable. The comment at `attempts-ledger.py:170` ("It also cannot be held by anything") is loose, but nothing turns on it.
- **`except Exception: d = {}` on the ledger read.** Real, but byte-identical on `origin/main`. Outside this diff.
- **Env-var parse at import** (`KIPI_ATTEMPTS_LOCK_TIMEOUT=abc` → ValueError before `main`'s try → exit 1 → "already claimed"). No producer sets these outside the suite. Not raising it.

---

## VERDICT: REQUEST CHANGES

One major, one minor. **Not a BLOCK** — nothing here is permanent in the unrecoverable sense (Linear comments are deletable), and the lock rewrite is a clear net improvement over `main` that I could not break in 32-way concurrency.

**Fix first:** `linear-worker.sh:606`. The `*)` arm is correct for the five Slack-only sites and wrong for `stuck_paged` at line 824, which turns an unrecorded page into a permanent Linear comment on every run.

FINDINGS:
major|page_once pages on exit 2 and 3, and the stuck call site's page is a permanent Linear comment, so contention posts it twice and an unwritable ledger posts it on every run forever|q-system/.q-system/scripts/linear-worker.sh:606
minor|The exit-code table says 2 is a usage error, but after the round-3 catch-all exit 2 is also every unexpected runtime failure, which is what makes the page-anyway arm look bounded|q-system/.q-system/scripts/attempts-ledger.py:104
END FINDINGS

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/pr-67-20260802-095401.md` (6787 bytes).

```
FINDINGS:
minor|The Slack alert in ledger_fault uses one template for exit 2 and exit 3, so benign lock contention pages the founder with "the ledger is not answering" and "check the ledger file is writable"|q-system/.q-system/scripts/linear-worker.sh:634
minor|bump_attempt's exit code is dropped, so an unwritable ledger freezes the attempts counter below MAX_ATTEMPTS and the new ledger_fault alarm can never fire for the caller the PR's rationale is about|q-system/.q-system/scripts/linear-worker.sh:1406
END FINDINGS
```

--- reviewer output ---

Read the code first, ran my own attacks, then read rounds 1-4. Both new suites are green on my machine (13 cases / 9 checks) and the mutation check is real.

## Findings

### MINOR — the Slack line the founder actually gets says the ledger is broken and names a fix, when the cause was benign lock contention

`q-system/.q-system/scripts/linear-worker.sh:634` (the `bash "$NOTIFY"` line in `ledger_fault`)

`ledger_fault` builds a correct per-code `detail` and puts it in `say` (the run log). The Slack line ignores `detail` and uses one template for rc=2 and rc=3. On rc=3 all three of its claims are false.

Reproducer (real functions lifted verbatim out of the worker, real flock holder, `$NOTIFY` stubbed to a log):

```
=== rc=3: a concurrent worker holds the lock. The ledger file itself: ===
    -rw-r--r--@ 1 assafkipnis  staff  3 Aug  2 10:00 .../attempts.json
    [ -w attempts.json ] -> WRITABLE
SAY>   WARN: ... (exit 3) -- the lock was contended, so nothing was written. The flag is
       still UNCLAIMED and the next run will claim it and page -- this defers a page by
       one cycle, it does not drop it.
page_once rc=1
SLACK> kipi worker: the attempts ledger at .../attempts.json is not answering (exit 3,
       first seen on ASK-CONTEND/stuck_paged). Once-only pages cannot be de-duplicated
       while this holds, so the worker is staying quiet on them rather than re-posting.
       Do: check the ledger file is writable.
```

Same text, verbatim, for rc=2 in the second half of the run. On rc=3: the ledger *did* answer; de-dup *does* still work (case 7 of the suite proves the next cycle posts exactly one comment); the file *is* writable, so "check the ledger file is writable" sends the founder after nothing at 3am.

Nothing in the suite asserts message content, so mutating either string ships green.

**Calibration, stated:** minor, not major. It does not cry wolf on every run, only on genuine contention, and I had to construct the 30s holder window to land it. Anchor: "log or help text that misstates what the code does."

---

### MINOR — the fault alarm this PR adds is wired to `page_once` only, so the failure the PR's own rationale is about stays silent

`q-system/.q-system/scripts/linear-worker.sh:1406` (`bump_attempt "$ISSUE" "claude run failed rc=$rc"`, exit code dropped)

The PR body: *"A lost increment means an issue exceeds MAX_ATTEMPTS and keeps being dispatched — silent budget burn on work that already failed three times, and a human is never told."* That is exactly what an unwritable state dir still does, and `ledger_fault` cannot fire because the STUCK branch is what calls `page_once`, and the frozen counter is what stops the STUCK branch being reached.

Reproducer (`chmod 555` on the state dir, lock file pre-created so the failure lands after acquisition, same fixture shape the suite's own case 13 uses):

```
  cycle 1: bump_attempt exit=2   count on disk=2   (MAX_ATTEMPTS=3)
  cycle 2: bump_attempt exit=2   count on disk=2   (MAX_ATTEMPTS=3)
  cycle 3: bump_attempt exit=2   count on disk=2   (MAX_ATTEMPTS=3)
  cycle 4: bump_attempt exit=2   count on disk=2   (MAX_ATTEMPTS=3)
  cycle 5: bump_attempt exit=2   count on disk=2   (MAX_ATTEMPTS=3)
  cycle 6: bump_attempt exit=2   count on disk=2   (MAX_ATTEMPTS=3)

  last stderr line: attempts-ledger: `bump-attempt` failed unexpectedly and wrote nothing (exit 2)
```

Six failed agent runs, counter frozen at 2, zero alerts. The only trace is a stderr line in the launchd log, which is this repo's own `an output nobody reads is the same as no output`.

**Calibration, stated:** minor, and pre-existing. `origin/main` loses the same write (uncaught, exit 1). This PR does not make it worse. What is new is the asymmetry: round 4 built a fault channel and wired it to one of the two ledger callers. Same one-line shape (`|| ledger_fault "$ISSUE" bump_attempt $?`) closes it.

---

## What is sound

Attacks I ran that the code survived. Dropping all of them:

- **Lost increments, 40-way.** `exit codes: {0: 40} / successes: 40 / recorded count: 40 / LOST INCREMENTS: 0`. The central claim of the PR holds under real concurrency.
- **The mutation check is not decoration.** I ran `q-system/output/mutcheck-ask286-r4.sh`: baseline `contend=1 broken=0 alerts=1`; `fault arm returns 0` → `contend=2 broken=5` **KILLED**; `once-per-run guard removed` → `alerts=4` **KILLED**. Round 1 complained the author's mutcheck was absent from the tree. It is present and it bites.
- **Round 4's major is genuinely closed.** `page_once` returns 1 on both 2 and 3 in my own repro, so the caller writes nothing, so no permanent Linear comment. Not inherited from the author's claim; observed.
- **`LEDGER_FAULT_ALERTED` really is once-per-run.** The drain is a pipeline into `while read`, i.e. a subshell, and `arm_automerge` is called inline (`:977`, `:1764`), not in a command substitution. The flag survives the whole drain. Suite case 9 confirms 4 issues → 1 alert.
- **Fixture shape.** This is the fleet's known scar (a mutex whose fixture used a key no producer emits). Not present: contention is produced by a live process holding a real `flock`, and case 9 monkeypatches `mod.open` to swap the inode *inside* the acquisition window rather than before it. Both are producer shapes.
- **All six sites converted.** `grep -n claim_page_once` returns the definition, `page_once`'s own call, and comment references only.
- **Process calibration.** Rounds 1-4 were calibrated, and I checked rather than took their word. Every round's finding was on code the previous round's fix introduced (r2 flock rewrite → r3 errexit leak → r4 `*)` arm blast radius → r5 message text). Rounds 3 and 4 each said this about themselves and were right. The lock itself has produced no new finding since round 2; the churn is entirely in the shell wrapper, and at round 5 it has converged to text and coverage. No re-litigation from me.

---

## VERDICT: APPROVE WITH NITS

Two minors, both bounded, neither gates. The lock rewrite is a clear net improvement over `main` and I could not break it.

**Fix first:** pass `$detail` into the Slack line in `ledger_fault` (`linear-worker.sh:634`) so rc=3 stops telling the founder to go check a file that is writable.

FINDINGS:
minor|The Slack alert in ledger_fault uses one template for exit 2 and exit 3, so benign lock contention pages the founder with "the ledger is not answering" and "check the ledger file is writable"|q-system/.q-system/scripts/linear-worker.sh:634
minor|bump_attempt's exit code is dropped, so an unwritable ledger freezes the attempts counter below MAX_ATTEMPTS and the new ledger_fault alarm can never fire for the caller the PR's rationale is about|q-system/.q-system/scripts/linear-worker.sh:1406
END FINDINGS

PR #70, opened 2026-08-02
feat(claude-changes): replace rule text through the safe path, ratcheted on rule content (ASK-289)

assafkip:
Closes ASK-289 (do not merge from here; closeout runs through /issue-verify and /issue-closeout).

## The problem

`apply_claude_changes.py` was additive-only, so a wrong sentence in a `.claude/` rule could be **buried but never corrected**. The scar is visible in the repo right now: `.claude/rules/design-auto-invoke.md` carries its narrowing paragraph wedged **above its own H1**, because `insert_before` was the only expressible op (ASK-134). A false enforcement claim had no fix.

## What changed

`replace` is now an op. It is safe only because of the second half of this change.

**Pinned to rule TEXT** (`rule_text_only`) — three checks, each covering a hole the other two leave:
- *identity*: refuses `.claude/settings.json` / `settings-template.json` under any spelling (resolved by inode upstream, so a case variant or `./` spelling arrives as the same key)
- *shape*: refuses anything that is not `.claude/rules/*.md`, so `agents/` and `output-styles/` keep the additive-only guarantee they had before
- *realpath*: refuses a symlink parked at `.claude/rules/x.md` pointing elsewhere. `scoped_path` permits that today (it only asks the target stay under `.claude/`), so a string-shape check alone waves it through — the `settings.local.json` (round 1) and `./` (round 2) classes arriving through the new op

**The census now reads rule CONTENT** (`_rule_content_marks`), not just the `rules/` directory listing. A listing is blind to everything a replace does inside a file. Two countable enforcement tokens per rule:
- an `(ENFORCED` marker — demoting a rule to advisory is enforcement-weakening
- each named `.py`/`.sh` the rule routes a reader to — a rule whose pointer to its enforcer vanishes is a rule nobody can act on

Prose meaning is deliberately **not** counted. A ratchet that judged tone or length would refuse the honest corrections this op exists to enable.

**Still no delete** (the issue's Not-doing line): `insert` must be a non-empty, non-blank string, so "replace this paragraph with nothing" is refused at load. An anchor contained in its own replacement is refused too — it never converges, and would grow the file on every run while already-applied stayed undetectable.

**One ambiguity check, not two**: anchor arithmetic extracted to `_unique_anchor_hits` so `replace` and the insert ops share it rather than keeping two copies free to drift.

## Verification

`bash q-system/.q-system/scripts/test/test-apply-claude-changes.sh`

RED first — every new case refused with `op 'replace' is not additive`. Then:

```
passed: 81   failed: 0
```

Two new mutation cases prove the new guards are load-bearing:

```
PASS: MUTATION marks: census blinded -> (ENFORCED) silently stripped (rc=0), test goes RED as required
PASS: MUTATION scope: pin removed -> replace reached settings.json and broke a hook (rc=0), test goes RED as required
```

All five pre-existing mutation cases still pass.

## Blast radius

`apply_claude_changes.py` is the only write route into `.claude/`. The additive ops, the L2 ratchet, L3 verify+revert, the lock, and the permission surfaces are unchanged. The config surface is unreachable by `replace`.

## Honest boundary

The content census counts markers and executable references, not meaning. A replace that rewrites a rule into something vaguer while keeping its `(ENFORCED` marker and every script name passes the ratchet. That is the deliberate trade: the alternative refuses the corrections this exists to enable.

### DIFF
--- a/q-system/.q-system/scripts/apply_claude_changes.py
+++ b/q-system/.q-system/scripts/apply_claude_changes.py
@@ -39,10 +39,12 @@ THE THREE LAYERS
                       removes something as a side effect -- e.g. inserting
                       " || true" after a hook command, which deletes nothing but
                       makes the old exact command string vanish from the census.
-                      The census reads rule CONTENT, not just the rules/ listing
-                      (_rule_marks): an (ENFORCED marker or a pointer to a named
-                      enforcer may not vanish from a rule, and a rule's count of
-                      substantive lines may not shrink. Filenames alone were
+                      The census reads rule CONTENT at any depth under rules/,
+                      not just the rules/ listing (_rule_marks): an (ENFORCED
+                      marker may not lose an occurrence NOR leave the heading it
+                      sits on, a pointer to a named enforcer may not vanish or be
+                      rerouted, and a rule's count of substantive lines may not
+                      shrink. Filenames alone were
                       blind to everything a replace does inside a file, which is
                       exactly why replace could not exist before that census; the
                       line floor is there because 5 of 34 rules carry no token at
@@ -57,17 +59,21 @@ half-writes.
 
 OUT OF REACH BY DESIGN (say so, do not pretend otherwise): removing a hook,
 deleting a rule, narrowing a matcher, widening permissions.allow or defaultMode,
-stripping a rule's (ENFORCED marker or its pointer to a named enforcer, changing
-a rule's frontmatter (the keys that decide whether it loads at all), and
-shortening a rule's body, cannot be done through this path at all. Those need a
-different tool and a real conversation. See REFUSED_SURFACES.
+dropping an occurrence of a rule's (ENFORCED marker or moving it off the heading
+it sits on, dropping or rerouting a rule's pointer to a named enforcer, changing
+a rule's frontmatter (the keys that decide whether it loads at all) by ANY op
+including the additive ones, and shortening a rule's body, cannot be done through
+this path at all. Those need a different tool and a real conversation. See
+REFUSED_SURFACES.
 
 WHAT STILL GETS THROUGH (the honest residue, do not read the list above as
-wider than it is): the census counts markers, executable references and line
-COUNT -- never meaning. A replace that rewrites a rule into something vaguer
-while keeping its (ENFORCED marker, every script name and its line count passes.
-That is the deliberate trade: a ratchet that judged prose would refuse the
-honest corrections this op exists to enable.
+wider than it is): the census counts markers, the routes to executables, and
+line COUNT -- never meaning. A replace that rewrites a rule into something vaguer
+while keeping its (ENFORCED marker on the same heading, every script route and
+its line count passes. Marker placement is checked at heading granularity only:
+moving it between two headings of the same file is invisible here. That is the
+deliberate trade: a ratchet that judged prose would refuse the honest
+corrections this op exists to enable.
 """
 import json
 import os
@@ -96,6 +102,14 @@ RULES_DIR = os.path.join(".claude", "rules")
 # A rule's pointer to the thing that actually enforces it. Census member, so a
 # replace cannot quietly cut a reader's route to the enforcer.
 #
+# The DIRECTORY PREFIX is part of the mark, not decoration. Matching the
+# basename alone let a replace keep `existing-lint.py` while moving its route to
+# `q-system/retired/hooks/existing-lint.py` -- the census saw the same member and
+# the reader's route pointed at nothing (PR #70 round 4, minor). The route as the
+# rule WRITES it is the census member, so moving it is a removal. The leading
+# lookbehind stops the match from starting mid-token, which is what keeps a
+# prefixed route from also registering as its own bare basename.
+#
 # The trailing boundary is load-bearing and is NOT a plain \b. Without it,
 # rewriting `voice-lint.py` to `voice-lint.py.retired` left the mark
 # `voice-lint.py` intact -- the pattern simply matched the prefix -- while the
@@ -104,13 +118,21 @@ RULES_DIR = os.path.join(".claude", "rules")
 # "... is voice-lint.py.", and rejecting a following "." would quietly drop
 # every one of those from the census. So the lookaheads reject exactly two
 # things: more filename characters, and a dot that STARTS another extension.
-_EXEC_REF = re.compile(r"[A-Za-z0-9_][A-Za-z0-9_.-]*\.(?:py|sh)(?![A-Za-z0-9_-])(?!\.[A-Za-z0-9_])")
+_EXEC_REF = re.compile(r"(?<![A-Za-z0-9_./-])(?:[A-Za-z0-9_.-]+/)*[A-Za-z0-9_][A-Za-z0-9_.-]*\.(?:py|sh)(?![A-Za-z0-9_-])(?!\.[A-Za-z0-9_])")
+
+# A markdown heading line. Used only to ask WHERE an (ENFORCED marker sits; see
+# _rule_marks for why the location and not just the presence has to be counted.
+_HEADING = re.compile(r"[ ]{0,3}#{1,6}[ \t]")
 
 # A rule's frontmatter decides whether the rule LOADS. Narrowing paths:/globs:
 # leaves the body, the tokens and the line count all identical while switching
 # the rule off, so no content census can see it (PR #70 round 3, major).
-# `replace` therefore may not move this block at all -- same standing as removing
-# a hook: a different tool and a real conversation.
+# NO op may move this block -- same standing as removing a hook: a different tool
+# and a real conversation. Pinning it to `replace` alone was itself a hole:
+# insert_before wedged a never-matching paths: block ABOVE the H1 of any rule
+# that had no frontmatter yet, which deletes nothing and switches the rule off
+# just the same (PR #70 round 4, major). "Additive" is a claim about the text,
+# not about the effect.
 _FRONTMATTER = re.compile(r"\A---\n.*?\n---(?:\n|\Z)", re.S)
 
 ALLOWED_PROPOSAL_KEYS = {"schema_version", "slug", "reason", "requires", "edits"}
@@ -310,6 +332,22 @@ def resolve_special_keys(root, prop):
     return settings_key, template_key
 
 
+def is_rule_text(rel):
+    """The SHAPE half of "is this rule text", in one place.
+
+    Two callers ask it -- the `replace` scope pin and the frontmatter pin -- and
+    two spellings of one question is the drift class round 2 found in the path
+    normalization and round 3 found again between the token census and the
+    frontmatter nobody was reading.
+
+    Depth-permissive on purpose, and that is now safe: _rule_marks walks the
+    whole rules/ tree, so a rule in a subdirectory is inside the content census
+    and inside these pins together. It used to be inside replace's reach and
+    outside the census at the same time (PR #70 round 4, minor).
+    """
+    return rel.startswith(RULES_DIR + os.sep) and rel.endswith(".md")
+
+
 def rule_text_only(root, rel, settings_key, template_key):
     """`replace` is the only non-additive op, so its reach is pinned to rule TEXT.
 
@@ -331,7 +369,7 @@ def rule_text_only(root, rel, settings_key, template_key):
     """
     if rel == settings_key or rel == template_key:
         raise Refusal("replace may not target %s (rule text only)" % rel)
-    if not (rel.startswith(RULES_DIR + os.sep) and rel.endswith(".md")):
+    if not is_rule_text(rel):
         raise Refusal("replace is only permitted on %s%s*.md, got %s" % (RULES_DIR, os.sep, rel))
     rules_dir = os.path.realpath(os.path.join(root, RULES_DIR))
     real = os.path.realpath(os.path.join(root, rel))
@@ -416,26 +454,53 @@ def _unique_anchor_hits(content, anchor, rel):
     return hits
 
 
+def _guard_frontmatter(rel, before, after):
+    """A rule's frontmatter block may not move, whatever op moved it.
+
+    Compared before/after rather than "is the anchor inside the block": that
+    also catches an insert that ADDS or CLOSES a --- fence, which moves the
+    frontmatter boundary without the anchor ever sitting in it. Equality is
+    transitive, so checking each edit holds across a whole proposal.
+
+    Called for EVERY op with prior content, not just `replace`. The round-3 fix
+    lived inside the replace branch, which left insert_before free to wedge a
+    never-matching paths: block above the H1 of a rule that had no frontmatter
+    yet -- additive text, dead rule, rc=0 (PR #70 round 4, major).
+    """
+    if not is_rule_text(rel):
+        return
+    if _frontmatter(after) != _frontmatter(before):
+        raise Refusal(
+            "edit may not change the frontmatter of %s; the scoping keys "
+            "decide whether the rule loads at all" % rel)
+
+
 def apply_edit(content, edit, rel):
-    """Return (new_content, already_satisfied). Pure; never touches disk."""
+    """Return (new_content, already_satisfied). Pure; never touches disk.
+
+    Every op that has prior content leaves through the single _guard_frontmatter
+    call at the bottom. One exit on purpose: a per-branch guard is what round 4
+    found missing on three of the four branches.
+    """
     op = edit["op"]
     ins = edit["insert"]
 
-    if op == "append":
-        if content.endswith(ins):
-            return content, True
-        return content + ins, False
-
     if op == "create_file":
+        # No prior content, so there is no frontmatter to preserve. A brand-new
+        # rule declaring its own scope is an addition, not a narrowing.
         if content is None:
             return ins, False
         if content == ins:
             return content, True
         raise Refusal("create_file target already exists with different content: %s" % rel)
 
-    anchor = edit["anchor"]
+    if op == "append":
+        if content.endswith(ins):
+            return content, True
+        new = content + ins
 
-    if op == "replace":
+    elif op == "replace":
+        anchor = edit["anchor"]
         # An anchor that survives into its own replacement never converges: the
         # next run finds it again, inserts again, and the file grows forever
         # while "already applied" stays undetectable. Refused, not tolerated --
@@ -452,31 +517,24 @@ def apply_edit(content, edit, rel):
                 return content, True
             raise Refusal("anchor not found in %s: %r" % (rel, _snip(anchor)))
         new = content.replace(anchor, ins, 1)
-        # Compared before/after rather than "is the anchor inside the block":
-        # that also catches an insert that ADDS or CLOSES a --- fence, which
-        # would move the frontmatter boundary without the anchor ever sitting in
-        # it. Equality is transitive, so checking each edit is enough to hold
-        # across a whole proposal.
-        if _frontmatter(new) != _frontmatter(content):
-            raise Refusal(
-                "replace may not change the frontmatter of %s; the scoping keys "
-                "decide whether the rule loads at all" % rel)
-        return new, False
 
-    if _unique_anchor_hits(content, anchor, rel) == 0:
-        raise Refusal("anchor not found in %s: %r" % (rel, _snip(anchor)))
+    else:
+        anchor = edit["anchor"]
+        if _unique_anchor_hits(content, anchor, rel) == 0:
+            raise Refusal("anchor not found in %s: %r" % (rel, _snip(anchor)))
+        pos = content.index(anchor)
+        if op == "insert_after":
+            cut = pos + len(anchor)
+            if content[cut:cut + len(ins)] == ins:
+                return content, True
+            new = content[:cut] + ins + content[cut:]
+        else:  # insert_before
+            if content[max(0, pos - len(ins)):pos] == ins:
+                return content, True
+            new = content[:pos] + ins + content[pos:]
 
-    pos = content.index(anchor)
-    if op == "insert_after":
-        cut = pos + len(anchor)
-        if content[cut:cut + len(ins)] == ins:
-            return content, True
-        return content[:cut] + ins + content[cut:], False
-
-    # insert_before
-    if content[max(0, pos - len(ins)):pos] == ins:
-        return content, True
-    return content[:pos] + ins + content[pos:], False
+    _guard_frontmatter(rel, content, new)
+    return new, False
 
 
 def _snip(text, n=60):
@@ -565,21 +623,47 @@ def _rule_marks(rules_dir):
     line_marks = set()
     if not os.path.isdir(rules_dir):
         return token_marks, line_marks
-    for name in sorted(os.listdir(rules_dir)):
-        if name.startswith(".") or not name.endswith(".md"):
-            continue
-        try:
-            with open(os.path.join(rules_dir, name)) as fh:
-                text = fh.read()
-        except (OSError, UnicodeDecodeError):
-            continue
-        for ref in set(_EXEC_REF.findall(text)):
-            token_marks.add("%s|exec|%s" % (name, ref))
-        if "(ENFORCED" in text:
-            token_marks.add("%s|enforced" % name)
-        substantive = sum(1 for line in text.splitlines() if line.strip())
-        for index in range(substantive):
-            line_marks.add("%s|line|%d" % (name, index))
+    # os.walk, not os.listdir. rule_text_only permits a rule at any depth under
+    # rules/, so a one-level census left a subdirectory rule inside replace's
+    # reach and outside every content check at once (PR #70 round 4, minor). The
+    # mark key is the path relative to rules/, which is identical to the old
+    # filename for every top-level rule, so nothing that was censused stops being.
+    for dirpath, dirnames, filenames in os.walk(rules_dir):
+        dirnames[:] = sorted(d for d in dirnames if not d.startswith("."))
+        for filename in sorted(filenames):
+            if filename.startswith(".") or not filename.endswith(".md"):
+                continue
+            full = os.path.join(dirpath, filename)
+            name = os.path.relpath(full, rules_dir)
+            try:
+                with open(full) as fh:
+                    text = fh.read()
+            except (OSError, UnicodeDecodeError):
+                continue
+            for ref in set(_EXEC_REF.findall(text)):
+                token_marks.add("%s|exec|%s" % (name, ref))
+            # (ENFORCED is counted twice, and the second count is the one that
+            # closes the hole. A whole-file presence boolean let a replace lift
+            # the marker off the heading that DECLARES the rule enforced and park
+            # the token in a sentence saying the rule is advisory -- the token was
+            # still in the file, so the census reported clean (PR #70 round 4).
+            #   occurrences      : total, so losing one anywhere is visible.
+            #   enforced-heading : how many HEADING lines carry it, so moving it
+            #                      out of a heading and into prose drops a mark.
+            # Counts encoded as sets, never positions: rewording a heading and
+            # adding a section above it both have to stay free, and only a count
+            # survives both. Same encode-a-count-as-a-set trick as line marks,
+            # for the same reason -- ratchet_check compares sets.
+            for index in range(text.count("(ENFORCED")):
+                token_marks.add("%s|enforced|%d" % (name, index))
+            lines = text.splitlines()
+            headings = sum(1 for line in lines
+                           if _HEADING.match(line) and "(ENFORCED" in line)
+            for index in range(headings):
+                token_marks.add("%s|enforced-heading|%d" % (name, index))
+            substantive = sum(1 for line in lines if line.strip())
+            for index in range(substantive):
+                line_marks.add("%s|line|%d" % (name, index))
     return token_marks, line_marks
 
 
--- a/q-system/.q-system/scripts/apply_claude_changes.py
+++ b/q-system/.q-system/scripts/apply_claude_changes.py
@@ -40,10 +40,13 @@ THE THREE LAYERS
                       " || true" after a hook command, which deletes nothing but
                       makes the old exact command string vanish from the census.
                       The census reads rule CONTENT, not just the rules/ listing
-                      (_rule_content_marks): an (ENFORCED marker or a pointer to a
-                      named enforcer may not vanish from a rule. Filenames alone
-                      were blind to everything a replace does inside a file, which
-                      is exactly why replace could not exist before that census.
+                      (_rule_marks): an (ENFORCED marker or a pointer to a named
+                      enforcer may not vanish from a rule, and a rule's count of
+                      substantive lines may not shrink. Filenames alone were
+                      blind to everything a replace does inside a file, which is
+                      exactly why replace could not exist before that census; the
+                      line floor is there because 5 of 34 rules carry no token at
+                      all, so tokens alone left them free to be gutted.
   L3 verify+revert  : run the gate suite before and after. Any gate that goes
                       pass -> fail auto-restores the backup. The founder must
                       never be the one who notices a regression.
@@ -54,9 +57,17 @@ half-writes.
 
 OUT OF REACH BY DESIGN (say so, do not pretend otherwise): removing a hook,
 deleting a rule, narrowing a matcher, widening permissions.allow or defaultMode,
-and stripping a rule's (ENFORCED marker or its pointer to a named enforcer,
-cannot be done through this path at all. Those need a different tool and a real
-conversation. See REFUSED_SURFACES.
+stripping a rule's (ENFORCED marker or its pointer to a named enforcer, changing
+a rule's frontmatter (the keys that decide whether it loads at all), and
+shortening a rule's body, cannot be done through this path at all. Those need a
+different tool and a real conversation. See REFUSED_SURFACES.
+
+WHAT STILL GETS THROUGH (the honest residue, do not read the list above as
+wider than it is): the census counts markers, executable references and line
+COUNT -- never meaning. A replace that rewrites a rule into something vaguer
+while keeping its (ENFORCED marker, every script name and its line count passes.
+That is the deliberate trade: a ratchet that judged prose would refuse the
+honest corrections this op exists to enable.
 """
 import json
 import os
@@ -84,7 +95,23 @@ RULES_DIR = os.path.join(".claude", "rules")
 
 # A rule's pointer to the thing that actually enforces it. Census member, so a
 # replace cannot quietly cut a reader's route to the enforcer.
-_EXEC_REF = re.compile(r"[A-Za-z0-9_][A-Za-z0-9_.-]*\.(?:py|sh)")
+#
+# The trailing boundary is load-bearing and is NOT a plain \b. Without it,
+# rewriting `voice-lint.py` to `voice-lint.py.retired` left the mark
+# `voice-lint.py` intact -- the pattern simply matched the prefix -- while the
+# reader's route to the enforcer was dead (PR #70 round 3, minor). But a plain
+# boundary over-corrects the other way: rules routinely end a sentence with
+# "... is voice-lint.py.", and rejecting a following "." would quietly drop
+# every one of those from the census. So the lookaheads reject exactly two
+# things: more filename characters, and a dot that STARTS another extension.
+_EXEC_REF = re.compile(r"[A-Za-z0-9_][A-Za-z0-9_.-]*\.(?:py|sh)(?![A-Za-z0-9_-])(?!\.[A-Za-z0-9_])")
+
+# A rule's frontmatter decides whether the rule LOADS. Narrowing paths:/globs:
+# leaves the body, the tokens and the line count all identical while switching
+# the rule off, so no content census can see it (PR #70 round 3, major).
+# `replace` therefore may not move this block at all -- same standing as removing
+# a hook: a different tool and a real conversation.
+_FRONTMATTER = re.compile(r"\A---\n.*?\n---(?:\n|\Z)", re.S)
 
 ALLOWED_PROPOSAL_KEYS = {"schema_version", "slug", "reason", "requires", "edits"}
 ALLOWED_EDIT_KEYS = {"file", "op", "anchor", "insert", "reason"}
@@ -424,7 +451,17 @@ def apply_edit(content, edit, rel):
             if content.count(ins) == 1:
                 return content, True
             raise Refusal("anchor not found in %s: %r" % (rel, _snip(anchor)))
-        return content.replace(anchor, ins, 1), False
+        new = content.replace(anchor, ins, 1)
+        # Compared before/after rather than "is the anchor inside the block":
+        # that also catches an insert that ADDS or CLOSES a --- fence, which
+        # would move the frontmatter boundary without the anchor ever sitting in
+        # it. Equality is transitive, so checking each edit is enough to hold
+        # across a whole proposal.
+        if _frontmatter(new) != _frontmatter(content):
+            raise Refusal(
+                "replace may not change the frontmatter of %s; the scoping keys "
+                "decide whether the rule loads at all" % rel)
+        return new, False
 
     if _unique_anchor_hits(content, anchor, rel) == 0:
         raise Refusal("anchor not found in %s: %r" % (rel, _snip(anchor)))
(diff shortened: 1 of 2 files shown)


assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/pr-70-20260802-110654.md` (8053 bytes).

```
FINDINGS:
major|replace deletes a rule's entire body (security.md, 475 chars) while the ratchet reports rule-marks 113->113 and the tool exits OK applied; 5 of 34 rules carry zero census marks|q-system/.q-system/scripts/apply_claude_changes.py:476
major|replace can narrow a rule's frontmatter paths:/globs: so the rule never loads, and the content census sees no change because it only reads the body|q-system/.q-system/scripts/apply_claude_changes.py:502
minor|_EXEC_REF has no trailing boundary, so rewriting `voice-lint.py` to `voice-lint.py.retired` preserves the exec mark while the rule's pointer to its enforcer is dead|q-system/.q-system/scripts/apply_claude_changes.py:87
END FINDINGS
```

--- reviewer output ---

# Review: PR #70 — `replace` op, ratcheted on rule content

Read the diff, read the engine, ran the suite (`passed: 81 failed: 0`, 3.8s), then attacked the new op against a copy of the real repo tree.

---

## Findings

### 1. major — `replace` deletes a rule's whole body and the tool reports `OK applied ... gates held`

`q-system/.q-system/scripts/apply_claude_changes.py:476` (`_rule_content_marks`)

The census counts `(ENFORCED` + `.py`/`.sh` tokens. **5 of 34 live rules carry zero marks**, so their entire content is invisible to L2. `security.md` is one of them.

```
python3 -c "import sys; sys.path.insert(0,'q-system/.q-system/scripts'); import apply_claude_changes as a, glob, os, collections
per=collections.Counter(x.split('|')[0] for x in a._rule_content_marks('.claude/rules'))
print([os.path.basename(f) for f in sorted(glob.glob('.claude/rules/*.md')) if per[os.path.basename(f)]==0])"
```
```
rules: 34  rules with ZERO census marks: 5
    auto-detection.md content-output.md marketing-system.md security.md sycophancy-core.md
```

Repro (fresh copy of this repo at `/tmp/pr70c/repo`, one `replace` whose anchor is the 475-char body of `security.md`):
```
cd /tmp/pr70c/repo && python3 q-system/.q-system/scripts/apply_claude_changes.py /tmp/pr70c/gut.json --root /tmp/pr70c/repo
```
```
OK applied gut-security-rule: 1 edit(s), 1 file(s), hooks 38->38, gates held
RC=0
--- security.md is now:
# Security Rules

See the team handbook.
--- ratchet line:
  ratchet ok: hooks 38->38, rules 34->34, rule-marks 113->113, deny 21->21
```

Every line of the security rule ("Never read, edit, or write `.env`", "Never include API keys… in committed files", "Never run `rm -rf` on root or dot directories") is gone. Ratchet says clean.

The code contradicts its own docstring at `:29` ("There is still no delete") and `:55-56` ("OUT OF REACH BY DESIGN … deleting a rule … cannot be done through this path at all"). The non-blank-`insert` check is not a no-delete guarantee; it costs one character to bypass.

3am answer: an agent lands a "tidy-up" proposal, the log says `gates held`, nobody looks. Human fix = `git diff .claude/rules/` and revert from `q-system/output/claude-changes/.backups/`.

---

### 2. major — frontmatter `paths:` narrowing disables a rule's load path; the census only reads the body

`q-system/.q-system/scripts/apply_claude_changes.py:502`

29 of 34 rules gate their loading on a `paths:`/`globs:` frontmatter block. `replace` reaches it, and the census counts only body tokens — so a rule with a full mark set can be scoped into never loading while the ratchet reports no change.

Repro on `linear-first.md` (the fleet's declared top rule, `paths:` `**/*.py`, `**/*.sh`, `**/*.md`, `**/*.json`, `**/*.yml`):
```
cd /tmp/pr70b/repo && python3 q-system/.q-system/scripts/apply_claude_changes.py /tmp/pr70b/narrow.json --root /tmp/pr70b/repo
```
```
OK applied narrow-linear-first: 1 edit(s), 1 file(s), hooks 38->38, gates held
RC=0
  ratchet ok: hooks 38->38, rules 34->34, rule-marks 113->113, deny 21->21
--- rule frontmatter now:
paths:
  - "q-system/output/.disabled-never-matches/**"
```

This one survives any fix to finding 1 that just counts more body tokens — the scope block carries no marks by construction.

---

### 3. minor — `_EXEC_REF` has no trailing boundary, so a rule's pointer to its enforcer can be redirected while the mark survives

`q-system/.q-system/scripts/apply_claude_changes.py:87`

```
python3 -c "import sys; sys.path.insert(0,'q-system/.q-system/scripts')
from apply_claude_changes import _EXEC_REF
for s in ['run voice-lint.py now','run voice-lint.py.disabled now','run voice-lint.pyc now']: print(repr(s),'->',_EXEC_REF.findall(s))"
```
```
'run voice-lint.py now' -> ['voice-lint.py']
'run voice-lint.py.disabled now' -> ['voice-lint.py']
'run voice-lint.pyc now' -> ['voice-lint.py']
```

End-to-end on `voice-enforcement.md`:
```
cd /tmp/pr70d/repo && python3 q-system/.q-system/scripts/apply_claude_changes.py /tmp/pr70d/redirect.json --root /tmp/pr70d/repo
```
```
OK applied redirect-enforcer: 1 edit(s), 1 file(s), hooks 38->38, gates held
  ratchet ok: hooks 38->38, rules 34->34, rule-marks 113->113, deny 21->21
26:Enforced by `voice-lint.py.retired` and `voice-substance-lint.py.retired` ...
```

The docstring at `:494` says "A rule whose pointer to its enforcer vanishes is a rule nobody can act on." The pointer vanished; the mark did not. Constructed input, no producer emits it today, so minor.

---

## What is sound

Attacks that held, each run against a real tree:

- **Symlink `.claude/rules/sneak.md -> ../settings.json`** — refused (`replace may not target`), settings.json byte-identical. The inode/identity + realpath layering does what the docstring claims.
- **Symlink out of `rules/` to an agent file** — refused (`resolves outside`).
- **`replace` on `.claude/settings.json` and on `.claude/agents/*.md`** — refused by identity and by shape respectively.
- **Empty and whitespace-only `insert`** — both refused at load.
- **Anchor contained in its own replacement** — refused (`never converges`). Correct call; the alternative is unbounded file growth on a re-run.
- **Ambiguous / absent anchor** — shared `_unique_anchor_hits` behaves identically for `replace` and the insert ops. Consolidating the two copies was the right move.
- **The two new mutation cases are genuinely load-bearing.** Both assert the mutant *writes* the bad state, not merely that a return code changed. Blinding the census does strip `(ENFORCED)`; removing the scope pin does put `"type": "cmd"` into settings.json.
- **No cry-wolf on ordinary work.** A benign `append` mentioning a script: `rule-marks 113->114`, applied. The new census key does not make the tool refuse the additive proposals it has always accepted.
- **The motivating case actually works.** I built the real un-wedge for `design-auto-invoke.md` (move the narrowing paragraph below its own H1) and ran it against a copy of this repo: `OK applied unwedge-design-rule: 1 edit(s), 1 file(s), gates held`. The feature delivers the thing ASK-289 exists for.

**Dropped after failing to reproduce:** a false `already-applied` when a replace's anchor is absent but its `insert` is coincidentally present once (`apply_edit` line 424). I could only reach it with an input I invented; no proposal producer in this repo emits that shape, so it does not survive the bar.

**Not pursued:** the lock and the write/rollback path are untouched by this PR (ASK-286, `#67`), so I did not re-litigate concurrency.

---

## VERDICT: REQUEST CHANGES

Not BLOCK. Every repro is git-tracked, a backup lands in `q-system/output/claude-changes/.backups/`, and the log records the edit. Nothing here is unrecoverable.

**Fix first: finding 2.** The frontmatter scope block is machine-read metadata that decides whether a rule loads at all, and `replace` reaches it while the census watches only the body. The cheapest correct move is to pin `replace` to the region *below* the closing `---` and refuse any anchor overlapping the frontmatter. That also shrinks finding 1's blast radius, though finding 1 still needs its own answer for the 5 mark-free rules — the honest-boundary paragraph in the PR body says "rewrites a rule into something vaguer"; what it actually permits is deleting the body outright.

FINDINGS:
major|replace deletes a rule's entire body (security.md, 475 chars) while the ratchet reports rule-marks 113->113 and the tool exits OK applied; 5 of 34 rules carry zero census marks|q-system/.q-system/scripts/apply_claude_changes.py:476
major|replace can narrow a rule's frontmatter paths:/globs: so the rule never loads, and the content census sees no change because it only reads the body|q-system/.q-system/scripts/apply_claude_changes.py:502
minor|_EXEC_REF has no trailing boundary, so rewriting `voice-lint.py` to `voice-lint.py.retired` preserves the exec mark while the rule's pointer to its enforcer is dead|q-system/.q-system/scripts/apply_claude_changes.py:87
END FINDINGS

assafkip:
## Round 3: three findings, three reproducers, all watched RED first

All three were real. Commit `c85f3e3`.

**major, body floor** (`apply_claude_changes.py:476`). Confirmed: 5 of this repo's 34 rules carry NEITHER token the census counts (`security.md`, `content-output.md`, `marketing-system.md`, `auto-detection.md`, `sycophancy-core.md` — no `(ENFORCED` marker, no named script). Replacing one's whole body reported `rule_marks 113 -> 113` and exited `OK applied`. The census now also emits one mark per substantive line index per rule — a count only becomes visible to `ratchet_check`'s set-difference when it is encoded as a set. **Lines, not words or characters:** rewording a line has to stay free, because that is the correction this op exists for; deleting lines must not. A case pins both directions (`gutting a zero-token rule refused` / `same-length reword still succeeds`).

**major, frontmatter** (`:502`). Confirmed: narrowing `paths:` leaves body, tokens and line count identical while the rule stops loading. `replace` now refuses any edit whose **result** changes the leading `---` block. Compared before/after rather than "is the anchor inside the block", so an insert that adds or closes a fence is caught too; equality is transitive, so the per-edit check holds across a whole proposal. The body of a frontmattered rule is still editable — a case pins that, so the fix is a pin and not a blanket refusal.

**minor, exec-ref boundary** (`:87`). Confirmed: `existing-lint.py.retired` contains `existing-lint.py`, so the mark survived a dead pointer. The boundary is deliberately **not** a plain `\b` — rules routinely end a sentence with "... is `voice-lint.py`.", and rejecting a following period would have silently dropped every one of those from the census, which is worse than the bug. The two lookaheads reject exactly two things: more filename characters, and a dot that *starts another extension*. A new case (`a ref ending a sentence is still a census member`) pins that this did not over-correct.

**One reader, not two.** `_rule_content_marks` became `_rule_marks` and returns both mark sets from one walk and one read per file. Two functions opening the same rules is two readers free to drift — the class round 2 found in the path normalization, and round 3 found again between the token census and the frontmatter it never looked at.

## Verification

`bash q-system/.q-system/scripts/test/test-apply-claude-changes.sh`

RED first, against the round-2 engine:

```
FAIL: gutting a zero-token rule refused (exit 0, wanted 2) :: OK applied gut-advisory
FAIL: narrowing frontmatter refused (exit 0, wanted 2) :: OK applied unload-rule
FAIL: renaming the enforcer to a dead name refused (exit 0, wanted 2) :: OK applied retire-enforcer
passed: 85   failed: 7
```

After:

```
passed: 95   failed: 0
```

Three new mutation cases, one per new guard:

```
PASS: MUTATION lines: floor blinded -> a zero-token rule's whole body vanished (rc=0), test goes RED as required
PASS: MUTATION frontmatter: pin dropped -> the rule was switched off via paths: (rc=0), test goes RED as required
PASS: MUTATION execref: boundary removed -> the rule now points at a name that does not run (rc=0), test goes RED as required
```

All five pre-existing mutations still pass.

## Honest boundary (updated, replaces the one in the PR body)

The census counts markers, executable references and line **count** — never meaning. A replace that rewrites a rule into something vaguer while keeping its `(ENFORCED` marker, every script name and its line count still passes. That is the deliberate trade, and it is now stated in the module docstring under `WHAT STILL GETS THROUGH` rather than only in this thread.

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/pr-70-20260802-112252.md` (9246 bytes).

```
FINDINGS:
major|the frontmatter pin only runs for `replace`, so insert_before adds a never-matching paths: block to any of the 11 no-frontmatter rules and switches an ENFORCED rule off at rc=0|q-system/.q-system/scripts/apply_claude_changes.py:460
minor|_rule_marks reads one directory level while rule_text_only permits any depth, so a rule in a rules/ subdir is outside the content census and replace gutted an ENFORCED one at rc=0|q-system/.q-system/scripts/apply_claude_changes.py:568
minor|_EXEC_REF marks the basename only, so repointing a rule's enforcer at a nonexistent directory keeps the census mark while the reader's route is dead|q-system/.q-system/scripts/apply_claude_changes.py:107
minor|the docstring claims the (ENFORCED marker cannot be stripped, but the census is a whole-file presence boolean, so replace removed it from the H1 and parked the token in a sentence saying the rule is advisory|q-system/.q-system/scripts/apply_claude_changes.py:578
END FINDINGS
```

--- reviewer output ---

## Review: PR #70, round 2 (PR history shows this is the 4th pass)

Ran the suite first: `passed: 95   failed: 0`. Then attacked the engine against fresh fixtures built from the PR's own `mk_fixture` (suite lines 47-133), and against throwaway copies of this repo's real `.claude/`. Repo left untouched, `git status --porcelain` empty at the end.

I read the prior rounds only after forming my own read. Round 3's three findings are all genuinely fixed for the op they were filed against. My major is the unclosed half of round 3 finding 2, not a new class on unchanged code.

---

### 1. major — the frontmatter pin guards `replace` only, so `insert_before` still switches an ENFORCED rule off

`q-system/.q-system/scripts/apply_claude_changes.py:460`

The round-3 fix (`_frontmatter(new) != _frontmatter(content)`) sits inside the `op == "replace"` branch of `apply_edit`. The additive ops return at lines 466-479 and never reach it. 11 of this repo's 34 rules carry no frontmatter today, so they load unconditionally; one `insert_before` gives them a `paths:` block that matches nothing.

Reproducer (throwaway copy of this repo's `.claude/`, real rule `wiring-check.md`):

```bash
python3 q-system/.q-system/scripts/apply_claude_changes.py "$C/off.json" --root "$C/repo"
# edit: {"file": ".claude/rules/wiring-check.md", "op": "insert_before",
#        "anchor": "# Definition of Done: Fully Wired (ENFORCED)",
#        "insert": "---\ndescription: Wiring\npaths:\n  - \"q-system/output/__never__/**\"\n---\n\n"}
```

Real output:

```
=== BEFORE: first 2 lines of wiring-check.md (no frontmatter -> always loads) ===
# Definition of Done: Fully Wired (ENFORCED)

OK applied scope-wiring-check-to-nowhere: 1 edit(s), 1 file(s), hooks 38->38, gates held
rc=0
=== AFTER: first 7 lines ===
---
description: Wiring
paths:
  - "q-system/output/__never__/**"
---

# Definition of Done: Fully Wired (ENFORCED)
=== ratchet line from the log ===
  ratchet ok: hooks 38->38, rules 34->34, rule-marks 113->113, deny 21->21
```

The 11 rules in this shape: `dogfood-gate.md`, `fable-discipline-auto-invoke.md`, `founder-notifications.md`, `model-allocation.md`, `quick-plan.md`, `rca-mode.md`, `no-orphan-findings.md`, `self-healing-retry.md`, `skill-hook-pairing.md`, `social-reaction-gate.md`, `wiring-check.md` (`grep -L -- '^---$' .claude/rules/*.md`).

3am answer: an agent lands an `insert_before` proposal, the log reads `gates held`, and the wiring gate silently stops loading. Human fix is `git diff .claude/rules/` plus the backup under `q-system/output/claude-changes/.backups/`. Recoverable, so major and not blocker. Severity matches round 3's frontmatter finding; not escalated.

Also falsifies the module docstring at `:58-62`: "changing a rule's frontmatter (the keys that decide whether it loads at all) ... cannot be done through this path at all."

---

### 2. minor — a rule one directory down is outside the content census, and `replace` is allowed there

`q-system/.q-system/scripts/apply_claude_changes.py:568` (`os.listdir`, not a walk) against `:334` (`rel.startswith(RULES_DIR + os.sep)`, which accepts any depth)

`rule_text_only` permits `.claude/rules/team/secrets.md`; `_rule_marks` and `_dir_names` both read one level only, so that file contributes zero token marks and zero line marks.

Reproducer output:

```
=== ATTACK 2: a rule in a rules/ SUBDIR is outside the content census ===
--- BEFORE ---
# Secrets Rule (ENFORCED)

The gate is `secret-scan.py`, PreToolUse on Bash.

Never write a credential into a committed file.

Never echo a token into a log the fleet ships.
OK applied gut-subdir-rule: 1 edit(s), 1 file(s), hooks 1->1, gates held
rc=0
--- AFTER ---
# Secrets

Use your judgement.
```

`find .claude/rules -type d` returns only `.claude/rules` today, so no producer emits this shape. Dropped a level for that, per the calibration rule.

---

### 3. minor — the exec-ref mark is basename-only, so repointing the enforcer at a dead directory keeps the mark

`q-system/.q-system/scripts/apply_claude_changes.py:107`

The `_EXEC_REF` char class excludes `/`, so the mark is `existing-lint.py` regardless of the path in front of it. The round-3 boundary closed the `.retired` suffix; the directory variant is the same "reader's route is dead, mark survives" class and is still live.

```
=== ATTACK 3: repoint the enforcer to a dead PATH, same basename ===
--- BEFORE ---
7:The gate is q-system/.q-system/scripts/existing-lint.py and it runs on write.
OK applied repoint-enforcer: 1 edit(s), 1 file(s), hooks 1->1, gates held
rc=0
--- AFTER ---
7:The gate is plugins/retired/does-not-exist/existing-lint.py and it runs on write.
```

Constructed input, same severity round 3 gave it. Not escalated.

---

### 4. minor — the docstring says the `(ENFORCED` marker cannot be stripped; `"(ENFORCED" in text` is a whole-file presence boolean

`q-system/.q-system/scripts/apply_claude_changes.py:578`, claim at `:60-61`

A replace can take the marker out of the H1 as long as the token appears anywhere, including in a sentence saying the opposite:

```
=== ATTACK 4: demote the H1 while keeping the (ENFORCED token somewhere ===
--- BEFORE ---
# Sample Rule (ENFORCED)
OK applied demote-keep-token: 1 edit(s), 1 file(s), hooks 1->1, gates held
rc=0
--- AFTER ---
# Sample Rule

This rule is advisory. It is no longer (ENFORCED) and nothing blocks on it.
```

Close to the stated residue at `:65-70`, which is why it is a minor and not more: the honest thing is to narrow the `OUT OF REACH BY DESIGN` sentence to match what the boolean actually holds.

---

## What is sound

- **Concurrency.** Two applies fired at the same target simultaneously: `REFUSED: another apply is already running (lock held); nothing was read or written` on one, `OK applied conc-1` on the other, and the file carried exactly one of the two appends. No last-writer-wins.
- **The whole existing suite, 95 cases including 8 mutation cases, green in one run.** The three new mutation cases are real: each asserts the mutant *wrote the bad state*, not that a return code moved. Blinding the line floor does gut a zero-token rule; dropping the frontmatter pin does write `__never__`; removing the exec-ref boundary does write the dead pointer.
- **Pushing text ABOVE an existing frontmatter block with `insert_before`.** It works (`frontmatter still parses: False`), and I dropped it as a finding: a rule with no frontmatter loads unconditionally in this repo, so destroying the block widens the rule rather than weakening it. Not an enforcement regression. Reporting the drop because it is the shape finding 1 grew out of.
- **`_unique_anchor_hits` consolidation.** One ambiguity check shared by `replace` and the insert ops is the right call, and the duplicate-anchor and absent-anchor cases behave identically across ops.
- **The line-floor / same-length-reword pair.** The floor refuses gutting while still allowing the reword the op exists for. That pin is what keeps the fix from being a blanket refusal.

## Process note

Rounds 1 through 3 were not miscalibrated: every finding was real, every fix landed with a RED-first reproducer and a mutation case. The repeating shape is narrower than that. Each round's guard is pinned to the *op that round attacked* rather than to the invariant, and the next round finds the same invariant reachable through a different op. Round 1: `settings.local.json`. Round 2: the `./` spelling. Round 3: frontmatter via `replace`. This round: frontmatter via `insert_before`. A guard placed on the invariant (frontmatter equality checked for every op, after all edits stage) closes the class instead of the instance.

## VERDICT: REQUEST CHANGES

Not BLOCK. Every reproducer is git-tracked, a backup lands in `q-system/output/claude-changes/.backups/`, and `apply.log` records the edit. Nothing is unrecoverable.

**Fix first: finding 1.** Move the `_frontmatter` equality check out of the `op == "replace"` branch and run it once per staged file against its pre-edit content, so it covers `insert_before`, `insert_after`, `append` and `create_file` too. That is a smaller change than the round-3 fix was, and it turns the pin into an invariant instead of an op-local guard.

FINDINGS:
major|the frontmatter pin only runs for `replace`, so insert_before adds a never-matching paths: block to any of the 11 no-frontmatter rules and switches an ENFORCED rule off at rc=0|q-system/.q-system/scripts/apply_claude_changes.py:460
minor|_rule_marks reads one directory level while rule_text_only permits any depth, so a rule in a rules/ subdir is outside the content census and replace gutted an ENFORCED one at rc=0|q-system/.q-system/scripts/apply_claude_changes.py:568
minor|_EXEC_REF marks the basename only, so repointing a rule's enforcer at a nonexistent directory keeps the census mark while the reader's route is dead|q-system/.q-system/scripts/apply_claude_changes.py:107
minor|the docstring claims the (ENFORCED marker cannot be stripped, but the census is a whole-file presence boolean, so replace removed it from the H1 and parked the token in a sentence saying the rule is advisory|q-system/.q-system/scripts/apply_claude_changes.py:578
END FINDINGS

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/pr-70-20260802-113924.md` (7977 bytes).

```
FINDINGS:
major|a case-variant spelling of .claude/settings.json creates a second staged key, so permission_surface_check inspects the first while the second is what lands, widening permissions.allow at rc=0 OK applied|q-system/.q-system/scripts/apply_claude_changes.py:1006
END FINDINGS
```

--- reviewer output ---

## Review: PR #70, round 3 (my own read first, prior rounds read after)

I ran the suite before touching anything: **114 passed, 0 failed**. Then I built the PR's own motivating example against a full copy of this repo and it worked, which is the thing the PR claims to enable.

---

## Finding 1

**major** — a case-variant spelling of `.claude/settings.json` produces a *second* `staged` key; `permission_surface_check` only inspects the first, so an additive edit on the second spelling widens `permissions.allow` and exits `0 OK applied`.

`q-system/.q-system/scripts/apply_claude_changes.py:1006`

This falsifies a citation made in this PR's body and repeated in new code. `rule_text_only`'s docstring (`:356-358`) says:

> Already collapsed to one key by inode upstream (resolve_special_keys), so a case variant or a ./ spelling arrives here as the same key rather than as a new name.

`resolve_special_keys` (`:326-332`) does not collapse anything. It **picks the first** edit whose inode matches and returns that key. Both spellings stay in `staged` as separate entries, each read from disk independently, each written back to the same inode in `sorted()` order (`:1041`). The lowercase spelling sorts last, so its content is what lands; the uppercase spelling is what `permission_surface_check` reads.

### Reproducer (run verbatim, from the repo root)

```python
python3 - <<'PY'
import json, os, subprocess, sys, tempfile
ENGINE = os.path.abspath("q-system/.q-system/scripts/apply_claude_changes.py")
r = tempfile.mkdtemp(prefix="pr70-case-")
os.makedirs(os.path.join(r, ".claude/rules")); os.makedirs(os.path.join(r, "q-system/.q-system/scripts"))
settings = dict(
    permissions=dict(allow=["Bash(ls:*)"], deny=["Read(.env)"], defaultMode="acceptEdits"),
    hooks=dict(PostToolUse=[dict(matcher="Edit|Write", hooks=[
        dict(type="command", command='python3 "$CLAUDE_PROJECT_DIR/q-system/.q-system/scripts/existing-lint.py"')])]))
open(os.path.join(r, ".claude/settings.json"), "w").write(json.dumps(settings, indent=2))
open(os.path.join(r, "settings-template.json"), "w").write(json.dumps(settings, indent=2))
open(os.path.join(r, "q-system/.q-system/scripts/existing-lint.py"), "w").write("print(1)\n")
open(os.path.join(r, ".claude/rules/x.md"), "w").write("# X (ENFORCED)\n\nBody line.\n")
e1 = dict(file=".claude/Settings.json", op="insert_after", anchor='"deny": [',
          insert='\n      "Read(.secret)",', reason="decoy that becomes settings_key")
e2 = dict(file=".claude/settings.json", op="insert_after", anchor='"allow": [',
          insert='\n      "Bash(:*)",', reason="THE ATTACK: widen allow, never permission-checked")
prop = dict(schema_version=1, slug="case-variant-bypass", reason="two spellings of settings.json",
            requires=dict(template_pairs=["existing-lint.py"]), edits=[e1, e2])
pf = os.path.join(r, "p.json"); open(pf, "w").write(json.dumps(prop, indent=2))
p = subprocess.run([sys.executable, ENGINE, pf, "--root", r], capture_output=True, text=True)
print("RC=%d" % p.returncode); print((p.stdout or p.stderr).strip())
live = json.load(open(os.path.join(r, ".claude/settings.json")))
print("FINAL permissions.allow =", live["permissions"]["allow"])
print("FINAL permissions.deny  =", live["permissions"]["deny"])
PY
```

### Real output

```
RC=0
OK applied case-variant-bypass: 2 edit(s), 2 file(s), hooks 1->1, gates held (log: .../apply.log)
FINAL permissions.allow = ['Bash(:*)', 'Bash(ls:*)']
FINAL permissions.deny  = ['Read(.env)']
```

Filesystem check, same machine, same run:

```
case-insensitive FS: True
samefile: True
```

### Two things go wrong in one shot

1. `permissions.allow` gained `Bash(:*)` through the path whose stated contract is `permissions.allow / defaultMode may not move` (`:720`). Everything reported clean: ratchet held, gates held, one line, exit 0.
2. Edit 0 asked for `Read(.secret)` in `deny`. `deny` is unchanged. The tool printed **"2 edit(s), 2 file(s)"** for one edit's worth of content. Success reported for work that did not happen.

### At 3am

An unattended agent grants itself a `Bash` allow entry and the operator's only signal is `OK applied`. Recoverable — `settings.json` is tracked, and the backup sits in `q-system/output/claude-changes/.backups/` — so this is a major, not a blocker. The cleanup is `git diff .claude/settings.json`, which nobody runs because the line said OK.

### Cheapest correct fix

Collapse at the boundary, where `canonical_rel` already lives: after `load_proposal`, refuse any proposal whose edit paths resolve to the same inode under two spellings. One name per file in `staged`, or no run. Anything that keeps two keys keeps two readers, which is the exact defect class round 2 wrote `canonical_rel` to kill — it just stopped at string normalization, and `samefile` sees a spelling `normpath` cannot.

---

Every one of these I ran; each was refused:

| Attack | Result |
|---|---|
| `replace` on `.claude/Rules/x.md` (case-variant rules dir) | `REFUSED: replace is only permitted on .claude/rules/*.md` |
| `replace` on `.claude/rules/../settings.json` | `REFUSED` (normpath collapses it, pairing check fires) |
| `replace` rerouting `existing-lint.py` → `retired/existing-lint.py` | `REFUSED: anchor is contained in the insert, ... never converges` |
| `replace` lifting `(ENFORCED` off the H1 into an "advisory" sentence | `REFUSED: enforcement ratchet: 1 rule_marks entr(ies) would disappear, first: x.md\|enforced-heading\|0` |
| Non-UTF-8 locale blinding the census (`LC_ALL=C LANG=C`) | Dead. Python auto-enables UTF-8 mode under C/POSIX (`locale enc: US-ASCII utf8mode: 1`, `READ OK`), so the 15 non-ASCII rules stay censused. Dropped. |
| Symlink write-through during the copy-tree staging phase | Real in principle (`open(dest,"w")` follows a symlink copied with `symlinks=True`), but `find .claude -type l` returns nothing and no producer creates one. Could not reproduce against this repo. Dropped. |

And the PR does the job it says it does. The motivating scar, fixed for real against a full copy of this repo:

```
OK applied fix-design-auto-invoke-order: 1 edit(s), 1 file(s), hooks 38->38, gates held
```

The `replace` op, the rule-content census, the line floor, the frontmatter pin on all four ops, and the shared `_unique_anchor_hits` all held against everything I threw at them. The mutation cases are not decoration — I watched each guard's mutant go RED in the suite run.

---

## Process note (worth more than another nit)

Rounds 1-4 produced 3 majors and 4 minors, **all seven inside the `replace` / census surface this PR added**. Round 2's review says explicitly: *"the lock and the write/rollback path are untouched by this PR, so I did not re-litigate concurrency."* Reasonable scoping — except this PR's own docstring makes an affirmative safety claim about the additive/config path (`:356-358`, `:60-67`) and the PR body repeats it as a bullet. A claim the PR asserts is in scope even when the code carrying it is not.

The result: four rounds of escalating scrutiny on the new op, while the old op still widens `permissions.allow` at exit 0. That is not miscalibrated severity — it is miscalibrated *aim*. Worth checking claims against code the PR did not touch when the PR is the thing making the claim.

---

## VERDICT: REQUEST CHANGES

One major, recoverable, so this does not BLOCK.

**Fix first:** collapse same-inode spellings to a single `staged` key at load, so `permission_surface_check` reads the text that actually lands. Then either delete the "arrives here as the same key" sentence from `rule_text_only`'s docstring or make it true.

FINDINGS:
major|a case-variant spelling of .claude/settings.json creates a second staged key, so permission_surface_check inspects the first while the second is what lands, widening permissions.allow at rc=0 OK applied|q-system/.q-system/scripts/apply_claude_changes.py:1006
END FINDINGS

assafkip:
## Round 5 fix pushed — 2f9b2cc

Finding accepted. Your reproducer, run verbatim against the pre-fix engine on this branch:

```
RC=0
OK applied case-variant-bypass: 2 edit(s), 2 file(s), hooks 1->1, gates held
FINAL permissions.allow = ['Bash(:*)', 'Bash(ls:*)']
FINAL permissions.deny  = ['Read(.env)']
```

Same reproducer, after:

```
RC=2
REFUSED: edits name one file under two spellings (.claude/Settings.json and
.claude/settings.json resolve to the same file); use one spelling per file
FINAL permissions.allow = ['Bash(ls:*)']
```

### What changed

`refuse_duplicate_spellings(root, prop)`, called before `resolve_special_keys`. It takes your prescription: collapse at the boundary, one name per file in `staged` or no run. Identity is `st_dev`/`st_ino` where the file exists, `realpath` where it does not.

The fix went there and not into `permission_surface_check` because the defect was not that a particular guard read the wrong key. It was that two keys existed. Every downstream check is already correct once the input holds one name per file — which is the same reason round 2 put `canonical_rel` at the boundary rather than adding a second `normpath` call beside the first. It just stopped at string normalization, and `samefile` sees what `normpath` cannot.

### The docstring

Corrected, not deleted. `resolve_special_keys` does match by inode rather than by string, so the identity check `rule_text_only` describes is real — it just never merged two keys into one, which is what the sentence claimed. The new text says which function owns which half and cites the round that found the gap.

### Coverage

Suite: **122 passed, 0 failed** (114 before).

Case 15p, four assertions:

```
PASS: a symlinked second spelling of settings.json refused
PASS: permissions.allow untouched by the refused proposal
PASS: two-spellings refusal emits exactly 1 line
PASS: a case-variant spelling of settings.json refused
PASS: two spellings of one rule file refused
PASS: two edits on one spelling still succeed
PASS: both chained edits landed
```

The symlink spelling carries the deterministic case because it holds on any filesystem; the case-variant case is probed for (`[ -f "$T/.claude/SETTINGS.json" ]`) rather than assumed, since on a case-sensitive FS those genuinely are two files and the assertion would be wrong. The rule-file case is there because this is not only a config-surface defect — two names for one rule file loses an edit and still reports `2 edit(s)`. The last two assertions are the over-blocking guard: chaining two edits onto one file under **one** spelling is the supported case and stays supported.

New mutation case, watched RED:

```
PASS: MUTATION spelling: collapse removed -> the unchecked spelling widened
      permissions.allow (rc=0), test goes RED as required
```

All prior mutation cases still pass.

### Honest boundary

Two `create_file` edits, both case-variant spellings of a path that exists under **neither** case, are not caught: no inode yet, and `realpath` preserves basename case. That is a reporting defect (one file created, `2 edit(s)` reported), not an enforcement one — nothing pre-existing can be weakened by it, and every guarded file this engine knows about exists. Stated in the function's docstring and captured as spillover rather than left in a comment thread.

### On the process note

Taken. Rounds 1-4 all landed inside the surface this PR added while the claim the PR itself made about the untouched path went unchecked — and the claim was false. The docstring asserting a property of `resolve_special_keys` is exactly the thing that should have been read against `resolve_special_keys`. Noted as the aim problem it is, not a severity one.

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/pr-70-20260802-115347.md` (8139 bytes).

```
FINDINGS:
minor|_EXEC_REF still lets a `~` suffix keep the census mark while the rule's route to its enforcer is dead, the third character-level variant of the same class|q-system/.q-system/scripts/apply_claude_changes.py:121
minor|a malformed live .claude/settings.json makes permission_surface_check raise an uncaught JSONDecodeError, giving a traceback and exit 1 instead of the documented one-line 0/2/3 contract|q-system/.q-system/scripts/apply_claude_changes.py:773
minor|create_file births .claude/settings.json with defaultMode bypassPermissions at rc=0 because _same_file requires the target to exist, so settings_key is None and three guards never fire|q-system/.q-system/scripts/apply_claude_changes.py:311
nit|the PR body's verification block reports passed: 81 while the suite now reports passed: 122|q-system/.q-system/scripts/test/test-apply-claude-changes.sh:1
END FINDINGS
```

--- reviewer output ---

I have my own read plus repros. Reading prior rounds after the fact confirmed nothing I found was already answered.

## Review: PR #70, round 4 (my read first, prior rounds read after)

Ran the suite before touching anything: **`passed: 122   failed: 0`, 5.2s**, 12 mutation cases green. Then attacked the engine with throwaway fixtures under `$TMPDIR`. Repo untouched, read-only throughout.

---

### 1. minor — the "mark survives, route dies" class is still open; `~` is a third spelling the two lookaheads let through

`q-system/.q-system/scripts/apply_claude_changes.py:121` (`_EXEC_REF`)

Round 3 closed `.py.retired`, round 4 closed the directory-prefix variant. `~` is in neither rejected class, so `existing-lint.py~` still yields the mark `existing-lint.py` while the reader's route points at nothing.

```
python3 -c "..."  # fixture rule + one replace, engine run with --root on a temp tree
```
Real output:
```
rc=0
OK applied route-tilde: 1 edit(s), 1 file(s), hooks 1->1, gates held
--- rule after ---
# Sample Rule (ENFORCED)

The deterministic half is `existing-lint.py~`, retired 2026-08-02.
...
route file exists? False
```

Rated minor, matching rounds 3 and 4. **Not escalated.** The useful part is not the character: it is that this invariant is on its third character-level patch while the frontmatter invariant got a real single-exit fix (`_guard_frontmatter`, :505). The equivalent move here is an existence check or an explicit enumerated boundary, not a fourth lookahead.

---

### 2. minor — a malformed live `settings.json` crashes the engine with a traceback and exit 1, outside the documented 0/2/3 contract

`q-system/.q-system/scripts/apply_claude_changes.py:773` (`json.load` on the live file, uncaught)

The module premise is "Output is ONE line" (:11) and the wrapper documents "0 applied, 2 refused, 3 applied-then-reverted".

```
# fixture with .claude/settings.json = "{ this is not json", one append proposal
```
Real output:
```
rc=1
  ...
  File ".../apply_claude_changes.py", line 1058, in main
    permission_surface_check(root, staged[settings_key])
  File ".../apply_claude_changes.py", line 773, in permission_surface_check
    live = json.load(fh)
json.decoder.JSONDecodeError: Expecting property name enclosed in double quotes: line 1 column 3 (char 2)
--- settings.json on disk after ---
{ this is not json
```

Fail-closed (nothing written, confirmed above), which is why it is minor and not more. But a caller checking `rc == 2` for "refused" reads exit 1 as neither refused nor applied, and the operator gets 20 lines instead of one. `gate_settings_parses` already anticipates this state existing; this reader does not.

---

### 3. minor — `create_file` births `.claude/settings.json` with `bypassPermissions` and reports `gates held`; the docstring says that is impossible

`q-system/.q-system/scripts/apply_claude_changes.py:311` (`_same_file` requires the target to *exist*), claim falsified at `:60-62`

`_same_file` returns False when `.claude/settings.json` does not exist, so `settings_key` is None, so `touches_settings` is False. Three guards then never fire: the settings/template both-or-neither pairing (:996), `permission_surface_check` (:1057), and the paired-write scoping.

```
# tree with .claude/rules/x.md and NO settings.json; one create_file edit
```
Real output:
```
rc=0
OK applied birth-wide: 1 edit(s), 1 file(s), hooks 0->0, gates held
exists: True
{
  "permissions": {
    "allow": [ "Bash(:*)" ],
    "defaultMode": "bypassPermissions"
  }
}
```

The docstring at :60-62 says "widening permissions.allow or defaultMode … cannot be done through this path at all." For the create case it can. Pre-existing (`resolve_special_keys` is unchanged from base `ddadfff`, verified), and it needs a tree with no `settings.json`, which no real instance has — so **dropped a level per the calibration rule**. The cheap correct fix is to resolve the guarded keys by *intended path* as well as by inode, or to refuse `create_file` on the two guarded targets outright.

---

### 4. nit — the PR body's verification block is stale

PR description says `passed: 81   failed: 0`; the suite now reports `passed: 122   failed: 0`. Three commits of fixes landed after that block was written.

---

## What is sound

Attacks I ran or reasoned through that the code survived:

- **`replace` + `append` line-count laundering.** Replaced three enforcement sentences with "Use your judgement." and padded the count back with an `append` in the same proposal: `rc=0, OK applied launder-lines`. I dropped it — the module docstring's `WHAT STILL GETS THROUGH` names exactly this ("a rewrite that keeps the shape and hollows out the meaning"). Reporting a documented trade as a defect is inflation.
- **Symlink divergence between the ratchet copy and the live write.** The copy writes *through* a symlink (`shutil.copytree(symlinks=True)` then `open(dest,"w")`), the live path replaces the symlink itself (`os.replace`, :1099). Different resulting trees. Not exploitable: the live write touches strictly *less* than the copy, so any mark the copy preserved the live tree also preserves. No repro produced, dropped.
- **A symlink and its target named as two edits.** `os.stat` follows symlinks, so both hash to one `(dev, ino)` and `refuse_duplicate_spellings` (:316) refuses. The round-5 fix covers the symlink case, not only the case variant.
- **A symlinked `rules/` subdirectory.** Pointing outside `rules/` is refused by the realpath check (:424); pointing inside is walked by `os.walk` under its real name, and the copy and live writes agree. Consistent both ways.
- **Non-UTF-8 locale crash in the unguarded write loop (:1097).** `except (OSError, IOError)` would not catch a `UnicodeEncodeError`, and 15 of 34 rules carry non-ASCII. Measured: `LC_ALL=C python3 -c "..."` → `utf-8 1`, PEP 540 auto-enables UTF-8 mode, so the crash does not occur. **Dropped, could not reproduce.**
- **The suite.** 122 cases, 12 mutation cases, each asserting the mutant *wrote the bad state* rather than that a return code moved. That is the reason to trust the other 110.

## Process note (worth more than finding 1)

Rounds 1-5 were **not** miscalibrated — every finding was real and every fix landed RED-first with a mutation case. But round 4's own process note diagnosed the repeating shape (op-local guards instead of invariant-local guards) and the author acted on it for exactly one invariant: frontmatter now leaves through a single `_guard_frontmatter` exit for every op. The exec-ref route invariant did not get the same treatment, and finding 1 is the predictable third instance. Six review rounds on one PR is the signal that the remaining items are below the bar to gate on, not that the code is unsafe.

## VERDICT: APPROVE WITH NITS

Nothing found is unrecoverable, nothing writes on the failure paths, and no producer in this repo emits the shapes in findings 1 and 3. Using REQUEST CHANGES to log three minors on a PR that has already absorbed five rounds would wedge work that should ship.

**Fix first:** finding 1 — close the exec-ref class with an enumerated boundary or a route-exists check, rather than a fourth character patch.

FINDINGS:
minor|_EXEC_REF still lets a `~` suffix keep the census mark while the rule's route to its enforcer is dead, the third character-level variant of the same class|q-system/.q-system/scripts/apply_claude_changes.py:121
minor|a malformed live .claude/settings.json makes permission_surface_check raise an uncaught JSONDecodeError, giving a traceback and exit 1 instead of the documented one-line 0/2/3 contract|q-system/.q-system/scripts/apply_claude_changes.py:773
minor|create_file births .claude/settings.json with defaultMode bypassPermissions at rc=0 because _same_file requires the target to exist, so settings_key is None and three guards never fire|q-system/.q-system/scripts/apply_claude_changes.py:311
nit|the PR body's verification block reports passed: 81 while the suite now reports passed: 122|q-system/.q-system/scripts/test/test-apply-claude-changes.sh:1
END FINDINGS

PR #76, opened 2026-08-03
The review gate cannot go green on a review that never ran (ASK-312)

assafkip:
## The defect

`kipi/reviewer-approved` is the one required check standing in for human review across this fleet. It went green **twice** on PR #74 for a review that never ran.

codex answered *"Reply `OK` and I'll run the review exactly as planned"* and echoed the prompt's findings **template** back: `FINDINGS:`, the column legend, `END FINDINGS`, zero rows. Structurally perfect, so the severity ladder fell through to `APPROVE`, and the script preferred that derived APPROVE over the reviewer's own stated `REQUEST CHANGES`.

| SHA | wall clock | artifact | posted |
|---|---|---|---|
| `73a8870` | ~8 min | 345 KB | `failure / REQUEST CHANGES` (correct) |
| `b6af3e0` | ~20 s | 8 KB | **`success / APPROVE`** |
| `1e4c748` | ~57 s | 38 KB | **`success / APPROVE`** |

It fails **open**, posts on the exact head SHA where it looks maximally legitimate, and the loop merges its own PRs.

## The fix, and the wrong turn on the way

The obvious move is to stop an empty findings block deriving APPROVE. **That was wrong.** It collides with a deliberate contract — `test-severity-floor.sh` pins *"an empty findings block must derive APPROVE"* by name, and `test-findings-block-reader.sh` case 2 depends on it so a round 2 that refutes every prior finding can still land. Two existing suites went red. The change was wrong, not the suites.

"Reviewed, found nothing" and "never started" are byte-identical *inside* the block. The discriminator has to come from outside it: what the reviewer itself said.

`resolve_verdict()` ranks verdicts and takes the harsher of stated and derived. The severity floor still overrides a reviewer that logs a blocker then writes APPROVE — its original job. What it can no longer do is let silence overrule a reviewer that said stop.

## Verification

Fixtures are the three real artifacts byte for byte, not imitations. A fixture I invent tests my model of the bug; these test the parser.

```
test-review-gate-no-fake-green   16/16
test-findings-block-reader       10/10
test-severity-floor            198/198
```

Replayed through the fixed path: the 345 KB real review still derives REQUEST CHANGES; both declines now resolve to REQUEST CHANGES instead of APPROVE.

[...]

PR #87, opened 2026-08-03
The review gate cannot be filled by a review that never ran, and the reviewer stops wearing our own config (ASK-274)

assafkip:
`q-system/.q-system/scripts/pr-review-agent.sh` owns `kipi/reviewer-approved`, a
REQUIRED context on main. It was wrong in both directions: shipping unreviewed
code AND wedging good code. Two defects, one commit each, both reproduced first.

## Defect 1 — false APPROVE (sp-df1a458f) — `787b6f5`

On Alice PR #1 round 2, codex answered with a PLAN and no review ("Reply `OK` and
I'll execute exactly that plan.") in 12 seconds. Its stdout carries the echoed
prompt, and the prompt ends with a literal `severity|one-sentence claim|file:line`
template between the two markers. That echo was the only COMPLETE block in the
stream:

    has_complete_findings_block -> true      (REVIEW_UNUSABLE never set)
    verdict_from_findings       -> APPROVE   (no severities means nothing was wrong)

...and `kipi/reviewer-approved=success` was posted on code nobody read.

Measured on `origin/main` before touching anything:

    findings_block           -> FINDINGS:\nseverity|one-sentence claim|file:line\nEND FINDINGS
    has_complete_findings... -> TRUE (usable)
    verdict_from_findings    -> [APPROVE]
    extract_verdict          -> [REQUEST CHANGES]   <- from the ECHOED RULE TEXT

That last line is the both-directions part: the "stated" verdict is the prompt's
own grading rule, not the reviewer's. A fix that merely prefers the harsher side
turns this into a false BLOCK and wedges a PR nobody reviewed. The correct answer
is UNSTATED.

Two independent holes, both closed:

- **(a)** `findings_block` skips a block whose rows exist but contain no gradeable
  severity. An EMPTY block stays legitimate and still derives APPROVE — that is
  the round-2-refutes-everything shape, pinned by `test-severity-floor.sh` and
  `test-findings-block-reader.sh` case 2. The discriminator is
  rows-that-are-not-findings, never row-counting.
- **(b)** `review_declined_to_start`: a plan-and-await answer is not a review even
  when a parseable block IS present. From round 2 the stream carries prior-round
  findings, so (a) alone would grade findings the reviewer never re-proved. It is
  anchored AFTER the last block (the prompt orders the block last), so a genuine
  review that QUOTES plan prose still lands.

`review_is_usable` is the one predicate both dispatch sites gate on — the codex
path and the Opus fallback, where this class hid last time (ASK-221).

## Defect 2 — root cause (sp-cc9955db) — `98cf566`

Why codex answered with a plan at all: `codex exec` loaded THIS FLEET'S config
into the reviewer's session — assaf-voice, audhd-executive-function,
fable-discipline, SessionStart and UserPromptSubmit hooks — and then applied the
founder's own "state your planned approach and wait for OK" rule to its own
review. The existing `-C $REVIEW_ROOT` worktree isolation does not reach it; the
config resolves from the user home.

Proven by observation, same prompt, neutral cwd, codex v0.146.0:

    without --ignore-user-config:  12 `hook: ` lines
    with    --ignore-user-config:   0 `hook: ` lines

**Stated narrowly on purpose.** This is not total isolation. The "Skill
descriptions were shortened to fit the 2% skills context budget" warning appears
in BOTH runs, so codex still sees the skill catalogue with the flag set. The hooks
were the layer carrying the plan-and-await instruction and they are gone; claiming
the flag severs skills would be an overclaim. Captured as its own item, not
bundled here.

## Verification

    test-review-plan-echo-gate.sh   before: FAIL at case 4    after: PASS (8 checks)
    KIPI_TEST_LIB_REF=origin/main   FAILS at case 4 (the test can fail)
    test-findings-block-reader.sh   PASS 9/9
    test-severity-floor.sh          PASS 198/198

The reproducer runs three NEGATIVE self-tests first (ordinary review; legitimately
empty block; genuine review quoting plan prose), so a predicate hard-wired to
"unusable" fails there instead of passing the defect cases.

No commit status was posted to any PR during this work.

## Overlap to resolve before merge

PR #76 (ASK-312) touches the same region with `resolve_verdict`. It is
complementary, not duplicate — it fixes stated-vs-derived disagreement, which does
not fire when a plan-only answer states no verdict at all. Whichever lands second
will need a small merge in `pr-review-agent.sh`'s verdict block.

### DIFF
--- a/q-system/.q-system/capability-manifest.json
+++ b/q-system/.q-system/capability-manifest.json
@@ -62,6 +62,10 @@
       "runner": "bash"
     },
     {
+      "path": "q-system/.q-system/scripts/test/test-review-plan-echo-gate.sh",
+      "runner": "bash"
+    },
+    {
       "path": "q-system/.q-system/scripts/test/test-repo-preflight.sh",
       "runner": "bash"
     },
--- a/q-system/.q-system/scripts/pr-review-agent.sh
+++ b/q-system/.q-system/scripts/pr-review-agent.sh
@@ -509,9 +509,46 @@ run_engine() {   # run_engine <claude|codex> <destination-file>
     claude) run_bounded "$TIMEOUT_SECONDS" bash -c \
               "cd '$REVIEW_ROOT' && claude -p --model '$CLAUDE_MODEL' \"\$1\" </dev/null > '$2' 2>&1" _ "$PROMPT" ;;
     codex)  run_bounded "$TIMEOUT_SECONDS" bash -c \
-              "codex exec --skip-git-repo-check --model '$CODEX_MODEL' -C '$REVIEW_ROOT' \"\$1\" </dev/null > '$2' 2>&1" _ "$PROMPT" ;;
+              "codex exec --ignore-user-config --skip-git-repo-check --model '$CODEX_MODEL' -C '$REVIEW_ROOT' \"\$1\" </dev/null > '$2' 2>&1" _ "$PROMPT" ;;
   esac
 }
+
+# --ignore-user-config KEEPS OUR OWN AGENT CONFIG OUT OF THE REVIEWER (sp-cc9955db).
+# Without it, `codex exec` loads THIS FLEET'S config into the reviewer's session.
+# The 2026-08-03 artifact shows it announcing "I'm using the assaf-voice,
+# audhd-executive-function, and fable-discipline skills", firing SessionStart and
+# UserPromptSubmit hooks, and then applying the founder's own "state your planned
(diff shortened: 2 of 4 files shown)


assafkip:
@codex review

assafkip:
@codex review

assafkip:
@codex review

assafkip:
@codex review

assafkip:
@codex review

assafkip:
@codex review

assafkip:
@codex review

PR #103, opened 2026-08-04
Wire the judge runner: the producer that never existed (ASK-363)

assafkip:
**Stacked on #102.** Base retargets to `main` automatically when #102's auto-merge lands. The diff here is one commit.

## The gap

The Judgment Compiler's entire evaluation half was unreachable. `--judge-run`, `_load_judge_run` and `evaluate` all existed — but **no production code ever wrote a judge run**, because `/prd-triage` never passed the flag. `evaluate` counts a calibration case only when a receipt carries **both** `judge` and `human`, so every triage produced a human-only receipt, `judged` stayed empty forever, and all four release gates (50 cases, 88% agreement, kappa 0.80, per-class recall) were unreachable by construction.

Roughly half the engine had no wired input. ~90 tests passed on that path because **every one of them hand-built the judge run** — the recurring "a gate's input needs a production producer, not just a test" class.

## Reproducer

Drive the real production path end to end and ask `evaluate` what it scored:

```
before: judged_receipts = 0
after:  judged_receipts = 1, cases = 1
```

## What landed

`kipi judgment judge --prd <id> --finding <id> --output <f>` assembles the packet, calls one LLM via `claude -p`, validates the reply against the existing `validate_judge_output`, and writes a judge run. `/prd-triage` calls it and passes `--judge-run`.

**Tools are OFF, and a test asserts it.** `duplicates[].source` is a filesystem path to a findings file and `prior_receipts` lists receipt ids — a judge that can open files reads prior *human* dispositions straight out of both, and the calibration set stops measuring prediction and starts measuring leakage. `source` is additionally dropped from the prompt text (not from the packet, so the hash binding is untouched), so the judge never even sees a pointer to the labels. A separate test asserts the real prompt text carries no disposition or rationale after one exists to leak.

**The packet binding.** I verified the plan's claim at the source before building on it: `packet_hash` excludes `assembled_at` and `packet_sha256`, so the judge's independent assembly hashes identically to `capture`'s. A test asserts the emitted `input_sha256` against a *separately assembled* packet, so if that ever stops holding, this fails loudly instead of silently making judged receipts impossible.

**Bounded retry, then a loud failure.** 3 attempts per the self-healing-retry contract, no fallback disposition, and no run file written on failure — a fabricated prediction poisons the calibration set worse than a missing one. Negative self-test: a transient malformed reply recovers within the cap (exactly 3 calls, correct output).

[...]

PR #102, opened 2026-08-04
Judgment receipts: required at approval (integrity), reported on drift (ASK-363)

assafkip:
Supersedes #101. Same goal — make judgment receipts REQUIRED at `advance approved` rather than merely written — but #101's review history said the change was two different changes wearing one coat.

## Blast radius, measured against the real 36 PRDs

The gate is unconditional: every dispositioned finding in a PRD being approved needs a receipt, with no date logic anywhere. `_findings_gate` runs only under `if target == "approved"`, so PRDs already past that point can never re-reach it:

| PRD status (spec frontmatter) | count | can still reach this gate? |
|---|---|---|
| archived | 21 | no |
| approved | 13 | no |
| **in-review** | **1** (`prd-build-craft-2026-06-15`) | **yes — 13 dispositioned, 0 pending** |
| no spec file | 1 (`prd-notify-ledger-isolation-2026-08-04`) | yes — 0 dispositioned, 8 pending |

### Migration cost, stated plainly

**`prd-build-craft-2026-06-15` will BLOCK at `advance approved`** until each of its 13 dispositioned findings (8 accepted, 4 rejected, 1 deferred) carries a receipt. That is correct and intended — those decisions were made before receipts were enforced, and the gate's whole point is that an approval cannot bless decisions it has no record of.

It is satisfiable without special tooling: re-run `findings_writer.py set-disposition` for each finding, which mints the receipt as a side effect. Bounded, one-time, self-service.

This is called out because an earlier version of this PR body claimed the gate "blocks nothing today". That claim came from measuring only the *active* PRD and reading an inert gate as a safe one. The honest statement is the one above: it blocks one in-review PRD until 13 re-dispositions mint their receipts.

**Note on reading these numbers:** `prd_runner.py status` reports only the *active* PRD (`.claude/state/active-prd.json`), which is `prd-notify-ledger-isolation-2026-08-04` with 0 dispositioned findings. The 13 belong to `prd-build-craft-2026-06-15`, which is not active but is `in-review` and therefore still loadable and advanceable. Per-PRD status lives in each spec's frontmatter under `.prd-os/prds/`, not in the runner state.

**The `issue/` subtree does not inflate this.** `cross_check_findings` uses `rglob`, which sweeps 114 files rather than the 36 top-level ones. All 78 subtree records carry `prd_id = null`, so the gate's `prd_id` scoping excludes every one of them: measured contribution is zero.

## Why this is a split, not round 9

[...]

PR #104, opened 2026-08-05
Publish four cross-instance lessons from the Judgment Compiler build (ASK-363)

assafkip:
Documentation only. Four HOW-only lessons written to `q-system/lessons/` via the skeleton-publisher path in `lessons/README.md`, all passing `lessons-validator.py`.

Closes the gap left when the ASK-363 run ended at the merge rather than at closeout: the reusable output existed only in Linear comments and code comments, not in the corpus other instances read from.

- **a check must be able to fail for the reason you care about** — twelve instances in one session, split evenly between author and reviewer. Carries two non-obvious practices: mutation-test the *guard* (twice a guard could have been deleted with the whole suite green), and distrust the convenient explanation for an intermittent failure (one timed run refuted "contention" on a test 10% over its cap).
- **kill a defect class by deleting the mechanism, not by guarding it** — three successive hardenings of a date inference that a code comment had already admitted was undecidable. Deleting the mechanism retired all three at once.
- **a safety property enforced at N sites is not a chokepoint** — blindness was preserved at four seams and three were wrong. Single-writer discipline applies to invariants, not only data paths.
- **an artifact must carry its own provenance, never inherit it** — a stale runner produced a review of the wrong tree, filed under the wrong engine's name, and was indistinguishable from a real review by reading it.

No executable behavior changes. 🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/q-system/lessons/a-check-must-be-able-to-fail-for-the-reason-you-care-abou.md
+++ b/q-system/lessons/a-check-must-be-able-to-fail-for-the-reason-you-care-abou.md
@@ -1,0 +1,21 @@
+---
+id: a-check-must-be-able-to-fail-for-the-reason-you-care-abou
+kind: pattern
+title: A check must be able to fail for the reason you care about, or it is decoration
+date: 2026-08-05
+---
+
+A test, gate, metric or diagnostic can pass while being structurally incapable of detecting the thing it was written to detect. It then reads as evidence of correctness and is worse than having no check at all, because it retires the question. The defect is not carelessness: checking a proxy for a property is easier than checking the property, and the proxy usually agrees, so nothing feels wrong.
+
+Observed twelve times in a single engineering session, split roughly evenly between the author and the reviewer. A test asserting a permission flag when the property was availability, controlled by a different flag. A table of malformed inputs where a coarse range guard already rejected every row, so the fine-grained parse could be deleted with the suite green. A binding test run against the one fixture shape structurally unable to exercise the binding. A measurement whose number equally supported "this is safe" and "this does nothing", read as the first. A process-matching pattern that could not distinguish the process to kill from the process to protect. A rate metric that by construction reports only on one subset of decisions, read as reporting on all of them. A pipeline whose exit status came from its final formatting command rather than the gate it wrapped, reporting success while the gate was red.
+
+How to apply:
+
+1. Before trusting a green result, name the concrete input that would make it RED for the reason you care about. If you cannot construct that input, the check is decoration. Delete it or rewrite it; do not leave it passing.
+2. Mutation-test the GUARD, not only the code it guards. Delete or invert the guard and confirm a test dies. A guard that can be removed with every test still green is not enforcing anything, and this is invisible to ordinary review.
+3. For races, timing and anything intermittent, a single run proves nothing in either direction: it can pass on broken code and fail on correct code. Iterate to N red before the fix and N green after, and use elapsed time as independent corroboration that the run actually exercised the path.
+4. Distrust the convenient explanation for an intermittent failure. "Contention", "flake", "transient" each cost one measured run to refute. Skipping that run converts a real defect into a false claim about a gate's state, which is worse than a wrong conclusion because it looks settled.
+5. Watch for the asymmetry of recomputing someone else's number while waving through your own. Both occurred in one session from the same engineer, who correctly named it afterward.
+6. Grade the INPUT before grading the source. A reviewer that read nothing is not weak evidence; it is not evidence at any strength, and its output should be discarded unread rather than triaged.
+
+The general contract: a check earns trust from the failure it can produce, never from the pass it did produce. State the red-making input first, and the pass becomes meaningful.
--- a/q-system/lessons/a-safety-property-enforced-at-n-sites-is-not-a-chokepoint.md
+++ b/q-system/lessons/a-safety-property-enforced-at-n-sites-is-not-a-chokepoint.md
@@ -1,0 +1,22 @@
+---
+id: a-safety-property-enforced-at-n-sites-is-not-a-chokepoint
+kind: pattern
+title: A safety property enforced at N sites is not a chokepoint; apply single-writer to invariants too
+date: 2026-08-05
+---
+
+Single-writer discipline is usually applied to data paths: one function owns the write, so the write cannot be inconsistent. The same reasoning applies to INVARIANTS, and it is easy to miss because an invariant does not look like a resource.
+
+Observed on a component required to stay blind to a value it was being measured against. Blindness was preserved at four independent seams: a runtime flag controlling capability, the content of a rendered prompt, what got printed to an interactive session, and how references were validated. Three of the four were wrong, each discovered in a separate review round, because each seam was a separate opportunity to get it right and nothing forced them to agree. The fix was one function constructing the entire view the component was permitted to see, plus one test asserting that view carries no forbidden field. Three of the four seams then became ordinary code that structurally could not leak.
+
+The same shape appeared in a two-phase write whose consistency was maintained by careful ordering plus rollback. Three rounds each moved a boundary and each left a different inconsistency reachable, because rollback cannot undo an append. Recovering forward from a single irreversible step, with everything after it recomputable, removed the class rather than guarding it.
+
+How to apply:
+
+1. When a property must hold, count the places that preserve it. More than one means more than one chance to be wrong, and review will find them one round at a time.
+2. Build one constructor that establishes the property, and write ONE test asserting the property of its output. Prefer an allowlist over deleting known-bad fields: a field a future contributor adds is then excluded by default rather than leaking until someone remembers to exclude it.
+3. Ask what the component may CITE or emit, not only what it may perceive. Both are the same invariant and both fail independently.
+4. In a multi-step write, make exactly one step irreversible. Everything before it is validated and reversible; everything after it must be derivable from what is already durable, so a late failure is recovered by recomputation rather than by an impossible rollback.
+5. A late-stage failure of a derived artifact must never escalate into refusing an operation that already succeeded. Report it, repair it inline if bounded, and let a separate verifier own the residual state.
+
+The general contract: if you would apply single-writer to a data path, apply it to the invariant. A property maintained by N cooperating precautions is a property waiting for the round that finds the N+1th seam.
--- a/q-system/lessons/an-artifact-must-carry-its-own-provenance-not-inherit-it.md
+++ b/q-system/lessons/an-artifact-must-carry-its-own-provenance-not-inherit-it.md
@@ -1,0 +1,22 @@
+---
+id: an-artifact-must-carry-its-own-provenance-not-inherit-it
+kind: pattern
+title: An artifact must carry its own provenance, never inherit it from where it was filed
+date: 2026-08-05
+---
+
+An output file whose identity comes from its location, its filename, or the intent of the script that produced it will eventually describe something it did not do. Nothing in the artifact itself contradicts the label, so the mismatch is invisible to every later reader.
+
+Observed when an automated review was run from a stale copy of a runner left behind in a scratch directory. The stale copy hardcoded three things: an output directory named after one engine while invoking a different one, a review root pointing at a parent directory that was not a repository at all, and a process name identical to the canonical runner's. The result was a file that looked like a review by one tool, was produced by another, contained a verdict formed with no diff available, and could not be distinguished from a real review by reading it. A process-level attempt to stop it would have matched the legitimate run instead.
+
+The same class explains a family of quieter failures: a prompt asserting that references are resolved when they are only pattern-matched, a note instructing a reader not to be shown a value printed directly beside it, and a validator whose documentation names a helper that has since been deleted. In each, an artifact asserted something about the system that the system did not do, and each satisfied whatever checker was watching.
+
+How to apply:
+
+1. Every generated artifact records, inside itself, what produced it: the tool actually invoked, the root it actually read, and the resolved identifier of the input it examined. A reader must never have to inspect an exited process to learn what a file means.
+2. Treat a directory name, a filename convention, or a script's stated intent as a hint, never as provenance. Where a label and the content disagree, the content is the truth and the label is a finding.
+3. Neutralise stale copies of executable tooling, and make surviving copies distinguishable at the process level, not merely absent from filesystem searches. Sharing a process name with the canonical runner makes targeted operations unsafe.
+4. Prose adjacent to a value does not constrain the value. A warning not to display something, printed next to the thing itself, has already failed. Move the enforcement into the code path that emits.
+5. When a comment names a helper, a flag, or a guarantee, it is a claim about the code and it decays. Prefer deriving documentation from the structures themselves, and treat a comment contradicted by the code as a defect rather than as stale text.
+
+The general contract: an artifact that inherits its identity from its context is a claim nobody verified. Make it self-describing, and a wrong run becomes self-evident instead of requiring forensics.
--- a/q-system/lessons/kill-a-defect-class-by-deleting-the-mechanism-not-by-guar.md
+++ b/q-system/lessons/kill-a-defect-class-by-deleting-the-mechanism-not-by-guar.md
@@ -1,0 +1,22 @@
+---
+id: kill-a-defect-class-by-deleting-the-mechanism-not-by-guar
+kind: pattern
+title: Kill a defect class by deleting the mechanism, not by adding a guard to it
+date: 2026-08-05
+---
+
+When successive review rounds keep finding NEW defects of the SAME class in the same code, each round is guarding one more instance of a class rather than removing the class. The signal is not "these are hard bugs"; it is that the mechanism producing them should not exist.
+
+Observed on a rule that decided an exemption by inferring a date from mutable, strippable state. Three separate hardenings shipped: first on one mutable field, then on a different identifier's embedded date, then on the format of that date. Each fix was correct about the instance it addressed and each left the class alive. A code comment written during the second attempt openly recorded that the inference was undecidable, and the third attempt was built on top of that admission. The resolution was to delete the exemption mechanism entirely so the rule consulted no date at all, which retired all three hardenings at once.
+
+The counter-case matters as much: deleting a mechanism is only correct when the thing it protected turns out to be unnecessary or obtainable another way. Measure that before deleting, rather than assuming.
+
+How to apply:
+
+1. Track defects by CLASS across rounds, not by count. Three findings that share a root cause are one signal about a surface; three unrelated findings are ordinary review. Only the first justifies restructuring.
+2. Distinguish self-inflicted from independent. A round finding a defect INTRODUCED by the previous round's fix means the loop is generating work faster than it removes it, and grinding is the wrong response. A round finding another pre-existing defect means review is working through a dense surface, which is review functioning.
+3. Before deleting a mechanism, measure what it actually protected. Enumerate the cases it currently exempts or handles, and confirm they are either unreachable or covered elsewhere. A deletion justified by argument rather than measurement is a new defect.
+4. When you do delete, delete the whole apparatus: the helper, its constants, its tests' reliance on it. Leave a test asserting the names stay gone, so a later change cannot quietly reintroduce the mechanism.
+5. Decide the restructure BEFORE the next round's verdict arrives. Deciding after means the verdict shapes the decision, and a clean round is not evidence of sufficiency for a class that has produced a defect every round so far.
+
+The general contract: a guard added to a bad mechanism inherits the mechanism's failure modes. Removing the mechanism is the only fix that cannot be followed by another instance of the same class.


assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/pr-104-20260805-085944.md` (190603 bytes).

```
FINDINGS:
minor|The lesson falsely claims one constructor defines the judge's entire permitted view even though tool availability remains an independent blindness boundary|q-system/lessons/a-safety-property-enforced-at-n-sites-is-not-a-chokepoint.md:10
END FINDINGS
```

--- reviewer output, last 54735 bytes of 190603 (full review at the path above) ---

+        # issue in the repo "proved" a duplicate (round 3). Membership is
+        # strictly stronger than both: a ref that is in the set necessarily
+        # exists, because the view was assembled from real state.
         #
-        # Unresolvable refs are DROPPED rather than retried: a model that
-        # cannot cite something real will not do better on attempt 2, and the
-        # honest outcome is an unsupported disposition. Dropping them leaves
-        # `evidence_gate_errors` to convert the disposition to needs-human,
-        # which is the gate WORKING and is already counted as
-        # `converted_to_needs_human`.
-        refs = list(output.get("evidence_refs") or [])
-        if refs:
-            unresolved = set()
-            for problem in resolve_evidence_refs(cfg, refs):
-                for ref in refs:
-                    if ref in problem:
-                        unresolved.add(ref)
-            output["evidence_refs"] = [r for r in refs if r not in unresolved]
-        # RELEVANCE, on top of prefix + existence. Scoped to `duplicate`
-        # deliberately: other codes legitimately cite refs the packet does not
-        # enumerate (`commit:` for already-remediated, `scope:` for
-        # scope-removed), and rejecting those would convert every disposition
-        # to needs-human and score nothing -- the same failure in the opposite
-        # direction, which is what the paired negative self-test guards.
-        if output["workflow_reason_code"] == "duplicate":
-            allowed = _packet_duplicate_refs(packet)
-            output["evidence_refs"] = [
-                r for r in output["evidence_refs"] if r in allowed]
+        # Non-citable refs are DROPPED rather than retried: a model that cannot
+        # cite something it was shown will not do better on attempt 2, and the
+        # honest outcome is an unsupported disposition. Dropping leaves
+        # `evidence_gate_errors` to convert it to needs-human, which is the
+        # gate WORKING and is already counted as `converted_to_needs_human`.
+        # The negative self-test over all nine prefixes is what stops this
+        # becoming "refuse everything", which would pass a membership test
+        # trivially while scoring zero calibration cases.
+        output["evidence_refs"] = [r for r in (output.get("evidence_refs") or [])
+                                   if r in citable]
         return {
             "model": model,
             "prompt_sha256": JUDGE_PROMPT_SHA256,
             "input_sha256": packet["packet_sha256"],
             "output": output,
         }
     raise ValidationError(
         f"judge produced no valid output in {JUDGE_MAX_ATTEMPTS} attempts:\n  "
         + "\n  ".join(failures)
         + "\nRefusing to write a run. A fabricated prediction is worse than a "
           "missing one: it enters the calibration set as if a judge had made "
           "it.")
 
 
 def cmd_judge(cfg: Config, args: argparse.Namespace) -> int:
     packet = assemble_packet(cfg, args.prd, args.finding)
     model = args.model or os.environ.get("KIPI_JUDGE_MODEL") \
         or JUDGE_MODEL_DEFAULT
-    run = run_judge(cfg, packet, model=model)
+    run = run_judge(packet, model=model)
     # Written only on success, and only after validation, so a failed judge
     # leaves no partial file for a later `--judge-run` to pick up.
     destination = Path(args.output)
     destination.parent.mkdir(parents=True, exist_ok=True)
     destination.write_text(json.dumps(run, indent=2, sort_keys=True),
                            encoding="utf-8")
     # The PREDICTION IS DELIBERATELY ABSENT from this summary (Codex major, PR
     # #103 round 2). `/prd-triage` runs this command inside the founder's
     # interactive session, so anything printed here lands in the transcript
     # they read BEFORE setting a disposition. A founder who sees the prediction
     # and agrees inflates measured agreement, and the calibration set stops
     # measuring anything.
     #
     # The first version printed `workflow_disposition` alongside a `note`
     # saying not to show it. That is prose doing a job that belongs to code:
     # the value was already on screen by the time anyone read the warning. The
     # prediction lives in the run file, which `set-disposition --judge-run`
     # consumes without displaying. Paired test:
     # TestTheJudgeSummaryWithholdsThePrediction.
     print(json.dumps({
         "path": str(destination),
         "model": run["model"],
         "prompt_sha256": run["prompt_sha256"],
         "input_sha256": run["input_sha256"],
         "prediction": "withheld until the disposition is set",
     }, indent=2))
     return 0
 
 
 def main(argv: list[str]) -> int:
     parser = argparse.ArgumentParser(description=__doc__)
     parser.add_argument("--selftest", action="store_true")
     parser.add_argument("--repo-root", help="override repo root discovery")
     sub = parser.add_subparsers(dest="cmd")
 
     p_assemble = sub.add_parser("assemble")
     p_assemble.add_argument("--prd", required=True)
     p_assemble.add_argument("--finding", required=True)
     p_assemble.add_argument("--output")
     p_assemble.set_defaults(func=cmd_assemble)
diff --git a/plugins/prd-os/tests/test_judgment_compiler.py b/plugins/prd-os/tests/test_judgment_compiler.py
index 39e17adb..f388c69b 100644
--- a/plugins/prd-os/tests/test_judgment_compiler.py
+++ b/plugins/prd-os/tests/test_judgment_compiler.py
@@ -153,93 +153,109 @@ def make_judge_run(packet: dict, *, disposition: str = "fix-now",
         "workflow_disposition": disposition,
         "workflow_reason_code": reason_code,
         "evidence_refs": evidence or [],
         "missing_context": [],
         "confidence": confidence,
     }
     if extra_output:
         output.update(extra_output)
     return {
         "model": "test-model",
         "prompt_sha256": "0" * 64,
         "review_run_id": "run-1",
         "input_sha256": input_sha256 or packet["packet_sha256"],
         "output": output,
     }
 
 
 # ---------------------------------------------------------------------------
 # Gap repros (PRD section 6, run RED against the pre-change tree)
 # ---------------------------------------------------------------------------
 
 
 class TestGapRepros:
     def test_g1_decision_time_context_survives_prd_change(self, judgment_repo):
         """G-1: adjudicating freezes decision-time PRD state; editing the PRD
         afterwards does not invalidate or mutate the receipt."""
         packet_path, packet = assemble_packet(judgment_repo)
         proc = capture(judgment_repo, packet_path)
         assert proc.returncode == 0, proc.stderr
         before = read_ledger(judgment_repo)
         assert len(before) == 1
         frozen_sha = before[0]["prd_state"]["sha256"]
         assert frozen_sha not in ("", "unknown")
 
         write_prd_spec(judgment_repo, status="approved")  # PRD moves on
         after = read_ledger(judgment_repo)
         assert after == before  # receipt untouched
         proc = run_judgment(judgment_repo, "verify")
         assert proc.returncode == 0, proc.stderr
 
-    def test_g2_technical_validity_and_workflow_disposition_are_separate(self, judgment_repo):
-        """G-2: a technically valid finding recorded as a workflow duplicate."""
+    def test_g2_technical_validity_and_workflow_disposition_are_separate(
+            self, judgment_repo, run_findings_writer):
+        """G-2: a technically valid finding recorded as a workflow duplicate.
+
+        The cited duplicate is a REAL second finding, produced by
+        findings_writer, so the packet actually contains the candidate. The
+        original version cited finding-1 as its own duplicate against a
+        zero-duplicate packet; `judge_view`'s citable set now (correctly)
+        refuses that, and the test was only ever green because relevance went
+        unchecked (ASK-363 judge_view refactor)."""
+        assert run_findings_writer(
+            judgment_repo, "add", PRD_ID, "--source", "codex-review",
+            stdin_text=json.dumps([{"severity": "major",
+                                    "body": "the fixture gate can be bypassed"}]),
+        ).returncode == 0
         packet_path, packet = assemble_packet(judgment_repo)
+        assert packet["duplicates"], "fixture must carry a duplicate candidate"
+        dupe = packet["duplicates"][0]
+        ref = f"finding:{dupe['prd_id']}/{dupe['finding_id']}"
         judge = make_judge_run(
             packet, disposition="duplicate", reason_code="duplicate",
-            evidence=[f"finding:{PRD_ID}/finding-1"],
+            evidence=[ref],
         )
         judge_path = judgment_repo / "judge.json"
         judge_path.write_text(json.dumps(judge))
         proc = capture(
             judgment_repo, packet_path, "--judge-run", str(judge_path),
             "--reason-code", "duplicate",
-            "--evidence", f"finding:{PRD_ID}/finding-1",
+            "--evidence", ref,
             disposition="rejected",
         )
         assert proc.returncode == 0, proc.stderr
         rec = read_ledger(judgment_repo)[-1]
         assert rec["judge"]["output"]["technical_validity"] == "valid"
         assert rec["judge"]["output"]["workflow_disposition"] == "duplicate"
         assert rec["human"]["disposition"] == "rejected"
         assert rec["human"]["reason_code"] == "duplicate"
 
     def test_g3_duplicate_without_reference_is_refused(self, judgment_repo):
         """G-3 / N-1 (human half): evidence-requiring disposition with no
         stable reference fails hard before anything is written."""
         packet_path, _ = assemble_packet(judgment_repo)
         proc = capture(judgment_repo, packet_path,
                        "--reason-code", "duplicate", disposition="rejected")
         assert proc.returncode == 2
         assert "duplicate" in proc.stderr
         assert read_ledger(judgment_repo) == []
 
     def test_g4_judge_run_reproducible_from_hashes(self, judgment_repo):
         """G-4: the stored packet re-hashes to the receipt's recorded basis;
         a corrupted packet is detected."""
         packet_path, packet = assemble_packet(judgment_repo)
         judge_path = judgment_repo / "judge.json"
         judge_path.write_text(json.dumps(make_judge_run(packet)))
         proc = capture(judgment_repo, packet_path, "--judge-run", str(judge_path))
         assert proc.returncode == 0, proc.stderr
         receipt_id = read_ledger(judgment_repo)[-1]["receipt_id"]
 
         proc = run_judgment(judgment_repo, "verify", "--packet", str(packet_path),
                             "--receipt-id", receipt_id)
         assert proc.returncode == 0, proc.stderr
 
         corrupted = json.loads(packet_path.read_text())
         corrupted["finding"]["body"] = "tampered"
         packet_path.write_text(json.dumps(corrupted))
         proc = run_judgment(judgment_repo, "verify", "--packet", str(packet_path),
                             "--receipt-id", receipt_id)
         assert proc.returncode == 2
 
@@ -2263,101 +2279,295 @@ class TestDuplicateClaimsMustComeFromThePacket:
         issues.mkdir(parents=True, exist_ok=True)
         (issues / f"{name}.md").write_text("# an unrelated real issue\n")
         return f"issue:{name}"
 
     def test_an_unrelated_real_issue_cannot_prove_a_duplicate(
             self, judgment_repo, tmp_path, run_findings_writer, judge_stub):
         ref = self._issue(judgment_repo)
         payload = json.dumps({
             "technical_validity": "valid",
             "technical_reason": "this looks like something we already have",
             "workflow_disposition": "duplicate",
             "workflow_reason_code": "duplicate",
             "evidence_refs": [ref],
             "missing_context": [],
             "confidence": 0.91,
         })
         run = tmp_path / "judge-run.json"
         assert run_judgment(
             judgment_repo, "judge", "--prd", PRD_ID, "--finding", "finding-1",
             "--output", str(run),
             env_extra={"KIPI_JUDGE_CMD": judge_stub,
                        "STUB_JUDGE_OUTPUT": payload}).returncode == 0
         emitted = json.loads(run.read_text())
         assert emitted["output"]["evidence_refs"] == [], (
             "an issue that is not a duplicate candidate in the packet was "
             f"accepted as proof of duplication: {emitted['output']['evidence_refs']}")
         assert run_findings_writer(
             judgment_repo, "set-disposition", PRD_ID, "finding-1", "rejected",
             "--rationale", "dupe", "--judge-run", str(run)).returncode == 0
         receipt = read_ledger(judgment_repo)[-1]
         assert receipt["judge"]["converted_to_needs_human"] is True
 
     def test_a_real_packet_duplicate_candidate_is_still_accepted(
             self, judgment_repo, tmp_path, run_findings_writer, judge_stub):
         """Negative self-test. Over-restricting would convert every duplicate
         to needs-human and score nothing, which fails the same way in the
         opposite direction."""
         jc = _load_module("jc_dupe_ok", "judgment_compiler.py")
         packet = jc.assemble_packet(_cfg_for(judgment_repo), PRD_ID, "finding-1")
         assert packet["duplicates"] == [], "fixture unexpectedly has duplicates"
-        # Synthesise the packet-derived ref shape and prove it survives.
-        refs = jc._packet_duplicate_refs(
-            {"duplicates": [{"prd_id": "prd-x-2026-01-01",
-                             "finding_id": "finding-7", "similarity": 0.9}]})
-        assert refs == {"finding:prd-x-2026-01-01/finding-7"}
+        # `_packet_duplicate_refs` is subsumed by `judge_view`, which derives
+        # the citable set for ALL nine prefixes from one view (ASK-363).
+        packet["duplicates"] = [{"prd_id": "prd-x-2026-01-01",
+                                 "finding_id": "finding-7", "similarity": 0.9,
+                                 "source": "some/ledger.jsonl"}]
+        _, citable = jc.judge_view(packet)
+        assert "finding:prd-x-2026-01-01/finding-7" in citable
 
 
 class TestBindingSurvivesTheDispositionWrite:
     """Codex MAJOR, PR #103 round 4. The plan's load-bearing design claim was
     "packet_hash excludes assembled_at and packet_sha256, so the judge can
     assemble independently and still bind". The exclusion is real; the
     conclusion did not follow.
 
     `findings_xref.cross_reference` computes candidates ONLY for findings whose
     disposition is currently `pending` (findings_xref.py:186-188). So the judge
     assembles while the finding is pending and sees cross-PRD duplicates;
     `_write_all` then sets the disposition; `capture_from_triage` reassembles,
     the candidates are gone, the hash moves, and `_load_judge_run` refuses the
     run as stale. Every finding with a cross-PRD duplicate candidate is
     therefore un-capturable with a judge run.
 
     The earlier binding test passed because the fixture had NO duplicates -- it
     asserted the property on the one input that could not exercise it.
     """
 
     def _prior_prd_with_a_matching_finding(self, repo, body):
         prior = "prd-prior-2026-07-01"
         path = repo / FINDINGS_REL / f"{prior}-findings.jsonl"
         path.parent.mkdir(parents=True, exist_ok=True)
         path.write_text(json.dumps({
             "id": "finding-1", "prd_id": prior, "source": "codex-review",
             "severity": "major", "body": body, "disposition": "rejected",
             "rationale": "we decided against this before",
             "created_at": "2026-07-01T00:00:00Z",
             "resolved_at": "2026-07-02T00:00:00Z",
         }) + "\n")
         return prior
 
     def test_a_cross_prd_duplicate_candidate_does_not_break_capture(
             self, judgment_repo, tmp_path, run_findings_writer, judge_stub):
         _, rows = _findings_rows(judgment_repo)
         self._prior_prd_with_a_matching_finding(judgment_repo, rows[0]["body"])
 
         jc = _load_module("jc_bind", "judgment_compiler.py")
         packet = jc.assemble_packet(_cfg_for(judgment_repo), PRD_ID, "finding-1")
         assert packet["duplicates"], (
             "fixture failed to produce a cross-PRD duplicate candidate, so "
             "this test cannot exercise the defect")
 
         run = tmp_path / "judge-run.json"
         assert run_judgment(
             judgment_repo, "judge", "--prd", PRD_ID, "--finding", "finding-1",
             "--output", str(run),
             env_extra={"KIPI_JUDGE_CMD": judge_stub}).returncode == 0
         disp = run_findings_writer(
             judgment_repo, "set-disposition", PRD_ID, "finding-1", "accepted",
             "--judge-run", str(run))
         assert disp.returncode == 0, (
             "a finding with a cross-PRD duplicate candidate could not be "
             f"captured with its judge run:\n{disp.stderr}")
         assert len(read_ledger(judgment_repo)) == 1
+
+
+# ---------------------------------------------------------------------------
+# judge_view: ONE constructor for the judge's view + its citable set
+# (ASK-363 / PR #103 refactor, replacing four independent seams)
+# ---------------------------------------------------------------------------
+
+# Field names that carry a triage LABEL, or point at a file that carries one.
+# Table-driven so reintroducing any of them is caught by name, wherever it is
+# spliced into the packet.
+LABEL_BEARING_FIELDS = (
+    "disposition", "rationale", "resolved_at", "triaged_at", "triaged_by",
+    "human", "judge", "workflow_disposition",
+)
+# (block, key) pairs where a `source` value is a filesystem PATH into a ledger
+# that carries other findings' dispositions. `review.source` ("codex-review")
+# and `scope.source` ("<prd>.md#Scope") are NOT paths into a label store and
+# are deliberately kept: `scope.source` is the only citable proof of scope.
+LABEL_POINTER_ROWS = ("duplicates", "remediation")
+
+# The verified nine-prefix classification. `closed` names the view field the
+# refs come from; the two refused kinds have no source in the packet at all.
+CITABLE_TABLE = {
+    "finding:": "duplicates",
+    "judgment:": "prior_receipts",
+    "prd:": "related_prds",
+    "issue:": "issue_state.issue_id",
+    "receipt:": "remediation[].issue_id",
+    "commit:": "remediation[].commit_sha",
+    "scope:": "scope.source",
+}
+REFUSED_PREFIXES = ("spillover:", "test:")
+
+
+def _rich_packet(repo):
+    """A packet with EVERY citable source populated.
+
+    Rows use the exact shapes `_assemble_duplicates` / `_assemble_remediation`
+    emit (read from the assemblers, not invented). A zero-duplicate fixture is
+    how an earlier hash-binding test passed while testing nothing.
+    """
+    jc = _load_module(f"jc_rich_{id(repo)}", "judgment_compiler.py")
+    packet = jc.assemble_packet(_cfg_for(repo), PRD_ID, "finding-1")
+    packet["duplicates"] = [{
+        "prd_id": "prd-other-2026-01-01", "finding_id": "finding-9",
+        "similarity": 0.91,
+        "source": "POINTER-DUPES-LEDGER.jsonl"}]
+    packet["related_prds"] = ["prd-other-2026-01-01"]
+    packet["prior_receipts"] = ["jr-00000001"]
+    packet["issue_state"]["issue_id"] = "ASK-555"
+    packet["remediation"] = [{
+        "issue_id": "ASK-556", "finding_id": "finding-1",
+        "closed_at": "2026-01-01T00:00:00Z", "commit_sha": "a" * 40,
+        "source": "POINTER-RECEIPTS-LEDGER.jsonl:7"}]
+    packet["repo_state"]["commit_sha"] = "b" * 40   # HEAD, must NOT be citable
+    packet["scope"] = {"source": ".prd-os/prds/x.md#Scope", "sha256": "c" * 64}
+    return jc, packet
+
+
+def _expected_ref(prefix, packet):
+    return {
+        "finding:": "finding:prd-other-2026-01-01/finding-9",
+        "judgment:": "judgment:jr-00000001",
+        "prd:": "prd:prd-other-2026-01-01",
+        "issue:": "issue:ASK-555",
+        "receipt:": "receipt:ASK-556",
+        "commit:": "commit:" + "a" * 40,
+        "scope:": "scope:.prd-os/prds/x.md#Scope",
+    }[prefix]
+
+
+class TestJudgeViewIsTheOnlyConstructorOfTheJudgesWorld:
+    """PR #103 rounds 1-4 found five majors, every one in the same dimension:
+    the judge's blindness and citation integrity. Blindness was enforced at
+    four independent seams (tool availability, prompt content, stdout, evidence
+    refs) and three failed. A property enforced at N sites is not a chokepoint,
+    so a fifth review round would only have told us the reviewer had not yet
+    found the next insufficient guard. `judge_view` is the single writer.
+    """
+
+    # --- Property 1: the perceivable view ---------------------------------
+
+    @pytest.mark.parametrize("field", LABEL_BEARING_FIELDS)
+    def test_no_label_bearing_field_survives_into_the_view(
+            self, judgment_repo, field):
+        """Splice the label into every block and prove the ALLOWLIST drops it.
+
+        Red at 85d4a46: the old builder deep-copied the packet and popped one
+        known pointer, so `packet["finding"]["rationale"]` reached the prompt
+        verbatim. A findings ledger record really does carry that field."""
+        jc, packet = _rich_packet(judgment_repo)
+        marker = f"LEAKED-{field.upper()}"
+        for block in ("finding", "review", "repo_state", "prd_state",
+                      "issue_state", "scope"):
+            packet[block][field] = marker
+        for block in ("duplicates", "remediation"):
+            for row in packet[block]:
+                row[field] = marker
+        view, _ = jc.judge_view(packet)
+        rendered = json.dumps(view, sort_keys=True)
+        assert marker not in rendered, f"{field!r} reached the judge"
+        assert f'"{field}"' not in rendered, f"key {field!r} reached the judge"
+
+    @pytest.mark.parametrize("block", LABEL_POINTER_ROWS)
+    def test_no_row_source_pointer_survives_into_the_view(
+            self, judgment_repo, block):
+        """`duplicates[].source` and `remediation[].source` are paths into
+        ledgers that carry OTHER findings' human dispositions. A judge that
+        could open one reads the labels straight out."""
+        jc, packet = _rich_packet(judgment_repo)
+        view, _ = jc.judge_view(packet)
+        rendered = json.dumps(view, sort_keys=True)
+        assert "POINTER-" not in rendered, rendered
+        assert all("source" not in row for row in view[block]), view[block]
+
+    def test_the_view_spec_covers_every_field_the_assembler_emits(
+            self, judgment_repo):
+        """Self-enumerating guard. An allowlist silently drops a NEW packet
+        field, which is the safe direction, but it must be a decision someone
+        made rather than an omission nobody noticed."""
+        jc, packet = _rich_packet(judgment_repo)
+        assert set(jc.JUDGE_VIEW_SPEC) == set(packet), (
+            "packet fields and JUDGE_VIEW_SPEC have diverged: "
+            f"{set(packet) ^ set(jc.JUDGE_VIEW_SPEC)}")
+
+    # --- Property 2: the citable set --------------------------------------
+
+    @pytest.mark.parametrize("prefix", sorted(CITABLE_TABLE))
+    def test_a_ref_outside_the_citable_set_is_refused(
+            self, judgment_repo, prefix):
+        """One case per CLOSED prefix: a well-formed ref of the right kind that
+        the view does not contain. Red at 85d4a46 for every prefix whose
+        resolver merely checked existence -- any real PRD, issue, path or
+        commit in the repo satisfied the gate."""
+        jc, packet = _rich_packet(judgment_repo)
+        _, citable = jc.judge_view(packet)
+        outsider = prefix + "not-in-the-view-at-all"
+        assert outsider not in citable
+        assert jc.evidence_gate_errors(None, None, [outsider], citable) == \
+            jc.evidence_gate_errors(None, None, [], citable)
+
+    @pytest.mark.parametrize("prefix", REFUSED_PREFIXES)
+    def test_the_two_refused_kinds_are_never_citable(
+            self, judgment_repo, prefix):
+        """Principled, not a default: no spillover block exists in the packet
+        and nothing enumerates test paths, so the judge cannot honestly cite
+        either. Nothing is stranded -- `duplicate` keeps `finding:`/`issue:`
+        and `already-remediated` keeps `receipt:`/`commit:`."""
+        jc, packet = _rich_packet(judgment_repo)
+        _, citable = jc.judge_view(packet)
+        assert not any(r.startswith(prefix) for r in citable), citable
+
+    def test_head_is_not_citable_but_a_remediation_commit_is(
+            self, judgment_repo):
+        """`repo_state.commit_sha` is the CURRENT commit and always exists, so
+        an existence check let a judge cite HEAD as proof of the very
+        remediation under review. Only `remediation[].commit_sha` counts."""
+        jc, packet = _rich_packet(judgment_repo)
+        _, citable = jc.judge_view(packet)
+        assert "commit:" + "a" * 40 in citable, "remediation commit must count"
+        assert "commit:" + "b" * 40 not in citable, "HEAD must not be citable"
+
+    # --- The negative self-test (acceptance criterion) --------------------
+
+    @pytest.mark.parametrize("prefix", sorted(CITABLE_TABLE))
+    def test_every_closed_kind_survives(self, judgment_repo, prefix):
+        """THE acceptance criterion, table-driven over all nine prefixes (seven
+        here, two in the refused test above).
+
+        With the carve-out empty, "refuse everything" would pass every
+        membership test trivially while converting every disposition to
+        needs-human and scoring ZERO calibration cases -- the same failure in
+        the opposite direction. Each closed kind must be proven to SURVIVE."""
+        jc, packet = _rich_packet(judgment_repo)
+        _, citable = jc.judge_view(packet)
+        assert _expected_ref(prefix, packet) in citable, (
+            f"{prefix} was derivable from the view but is not citable; "
+            "over-refusal scores zero cases")
+
+    def test_the_survival_check_can_actually_fail(self, judgment_repo):
+        """Negative self-test OF the negative self-test. An empty citable set
+        is exactly what "refuse everything" produces; prove the survival
+        assertion above goes red against it rather than passing vacuously."""
+        _, packet = _rich_packet(judgment_repo)
+        empty: frozenset = frozenset()
+        for prefix in CITABLE_TABLE:
+            assert _expected_ref(prefix, packet) not in empty
+        # and the same mutant must be caught by the gate, not silently allowed
+        jc = _load_module("jc_mutant_gate", "judgment_compiler.py")
+        assert jc.evidence_gate_errors(
+            "duplicate", None, ["finding:prd-other-2026-01-01/finding-9"],
+            empty), "an empty citable set must fail the duplicate gate"
a07eabe2 fix(prd-os): delete the date exemption; the lock must span the cross-check (ASK-363)
 plugins/prd-os/.claude-plugin/plugin.json      |   2 +-
 plugins/prd-os/CHANGELOG.md                    |  32 ++++++++
 plugins/prd-os/scripts/prd_runner.py           | 108 +++++++++----------------
 plugins/prd-os/tests/test_judgment_compiler.py |  88 ++++++++++++--------
 4 files changed, 124 insertions(+), 106 deletions(-)
diff --git a/plugins/prd-os/tests/test_judgment_compiler.py b/plugins/prd-os/tests/test_judgment_compiler.py
index 1a38d16d..6cad9a0a 100644
--- a/plugins/prd-os/tests/test_judgment_compiler.py
+++ b/plugins/prd-os/tests/test_judgment_compiler.py
@@ -1679,78 +1679,89 @@ def _rewrite_findings(path: Path, rows: list[dict]) -> None:
 
 
 class TestGateUsesAPrdLevelFloor:
     """Rounds 2, 7 and 8 were ONE defect class: inferring "was this decided
     after the floor" from `resolved_at`, a mutable strippable field on a
     hand-editable file. The round-8 code admitted the inference was undecidable
     ("undateable AND unclaimed: cannot judge, do not guess") and left a
     documented hole: capture off, strip the date, invisible. The gate runs for
     ONE named PRD whose id carries its creation date, so the floor is read from
     the id instead and no `resolved_at` is parsed at all.
     """
 
     def test_stripping_resolved_at_no_longer_hides_a_missing_receipt(
             self, judgment_repo, run_findings_writer):
         """THE round-8 hole, executed: the documented invisible case."""
         assert run_findings_writer(
             judgment_repo, "set-disposition", PRD_ID, "finding-1", "accepted",
             env_extra={"KIPI_JUDGMENT_CAPTURE": "0"}).returncode == 0
         path, rows = _findings_rows(judgment_repo)
         for row in rows:
             row.pop("resolved_at", None)
         _rewrite_findings(path, rows)
         runner = _load_module("pr_floor_a", "prd_runner.py")
         code, message = runner._judgment_receipt_gate(
             _cfg_for(judgment_repo), PRD_ID)
         assert code == 2, (
             "a dispositioned finding with the date stripped and no receipt was "
             "waved through: the kill-switch-plus-strip hole")
         assert "no judgment receipt" in message
 
-    def test_a_prd_created_before_the_floor_stays_exempt(
+    def test_an_old_prd_does_not_exempt_a_decision_made_today(
             self, fake_repo, write_config, run_findings_writer):
-        """Pre-compiler decisions can never have receipts. A gate that cannot
-        be satisfied gets switched off, and an off gate protects nothing."""
+        """Codex BLOCKER on PR #102: a PRD-creation-date floor exempted every
+        FUTURE decision on a pre-floor PRD. 35 of 36 real PRDs predate the
+        floor, so the gate was a near-permanent no-op -- the opposite of
+        "receipts are required from here on".
+
+        Measured before removing the exemption: of the 36 real PRDs, 21 are
+        archived and 13 approved, so they can never reach this gate again. ONE
+        (in-review, 13 dispositioned findings) can, and its remedy is one
+        `set-disposition` re-run per finding, which mints the receipt. So the
+        exemption bought almost nothing and cost the entire guarantee."""
         legacy = "prd-legacy-2026-07-01"
         write_config(fake_repo, {"config_schema_version": 1})
         write_prd_spec(fake_repo, legacy)
         assert run_findings_writer(
             fake_repo, "add", legacy, "--source", "codex-review",
             stdin_text=json.dumps(
                 [{"severity": "major", "body": "a legacy finding"}])
         ).returncode == 0
         assert run_findings_writer(
             fake_repo, "set-disposition", legacy, "finding-1", "accepted",
             env_extra={"KIPI_JUDGMENT_CAPTURE": "0"}).returncode == 0
         runner = _load_module("pr_floor_b", "prd_runner.py")
-        code, _ = runner._judgment_receipt_gate(_cfg_for(fake_repo), legacy)
-        assert code == 0, "the floor must not block every legacy PRD forever"
+        code, message = runner._judgment_receipt_gate(_cfg_for(fake_repo), legacy)
+        assert code == 2, (
+            "an old PRD id must not buy an exemption for a decision being "
+            "approved today")
+        assert "no judgment receipt" in message
 
     def test_an_undateable_prd_id_fails_closed(
             self, fake_repo, write_config, run_findings_writer):
         """No code enforces a prd_id format, so an id whose date cannot be
         parsed must be treated as post-floor, never exempted."""
         weird = "prd-no-date-at-all"
         write_config(fake_repo, {"config_schema_version": 1})
         write_prd_spec(fake_repo, weird)
         assert run_findings_writer(
             fake_repo, "add", weird, "--source", "codex-review",
             stdin_text=json.dumps(
                 [{"severity": "major", "body": "an undateable finding"}])
         ).returncode == 0
         assert run_findings_writer(
             fake_repo, "set-disposition", weird, "finding-1", "accepted",
             env_extra={"KIPI_JUDGMENT_CAPTURE": "0"}).returncode == 0
         runner = _load_module("pr_floor_c", "prd_runner.py")
         code, message = runner._judgment_receipt_gate(_cfg_for(fake_repo), weird)
         assert code == 2, "an unparseable prd_id must fail CLOSED, not exempt"
         assert "no judgment receipt" in message
 
     def test_a_post_floor_prd_with_its_receipt_passes(self, judgment_repo,
                                                       run_findings_writer):
         """Negative self-test: the gate is not simply always-blocking."""
         assert run_findings_writer(judgment_repo, "set-disposition", PRD_ID,
                                    "finding-1", "accepted").returncode == 0
         runner = _load_module("pr_floor_d", "prd_runner.py")
         code, message = runner._judgment_receipt_gate(
             _cfg_for(judgment_repo), PRD_ID)
         assert code == 0, message
@@ -1847,59 +1858,70 @@ class TestPersistPathIsOneCriticalSection:
 
     def test_no_reader_can_observe_a_disposition_without_its_receipt(
             self, judgment_repo, tmp_path, monkeypatch):
         import argparse
 
         writer = _load_module("fw_window", "findings_writer.py")
         observer = tmp_path / "lock_observer.py"
         observer.write_text(_LOCK_OBSERVER)
         seen: list[str] = []
         real_write_all = writer._write_all
 
         def watched_write_all(path, records):
             real_write_all(path, records)
             proc = subprocess.run(
                 [sys.executable, str(observer), str(judgment_repo), PRD_ID],
             seen.append((proc.stdout + proc.stderr).strip())
 
         monkeypatch.setattr(writer, "_write_all", watched_write_all)
         args = argparse.Namespace(
             prd_id=PRD_ID, finding_id="finding-1", disposition="accepted",
             rationale="", covered_by="", reason_code=None, evidence=[],
             actor="founder", judge_run=None)
         assert writer.cmd_set_disposition(_cfg_for(judgment_repo), args) == 0
         assert seen, "the persist path never wrote the findings file"
         assert seen[0] == "LOCK_HELD", (
             "an outside reader observed the findings file mid-persist: "
             f"{seen[0]}")
 
 
-@pytest.mark.parametrize("prd_id,requires_receipt", [
-    ("prd-real-2026-08-05", True),      # post-floor: enforced
-    ("prd-legacy-2026-06-01", False),   # genuinely pre-floor: exempt
-    ("prd-nodate", True),               # no date at all: fail closed
-    ("prd-evade-0000-00-00", True),     # date-SHAPED, no calendar produces it
-    ("prd-evade-1970-01-01", True),     # a real date, absurd for this project
-    ("prd-evade-2026-13-45", True),     # month 13: rejected by shape alone
-    # IN-RANGE and sorts before the floor, so ONLY the real-date parse rejects
-    # it. Added after a mutation run: deleting the strptime guard left every
-    # other row green, because the plausibility floor happens to subsume
-    # 0000-00-00 and string ordering happens to subsume 2026-13-45. Without
-    # this row the guard could be deleted and no test would notice.
-    ("prd-evade-2026-02-30", True),     # February 30th does not exist
-    ("prd-evade-2026-06-31", True),     # June has 30 days
-])
-def test_only_a_real_plausible_date_earns_an_exemption(prd_id, requires_receipt):
-    """The exemption must need a date that a calendar can produce AND that this
-    project could plausibly have used.
-
-    The whole reason the floor moved off `resolved_at` was that a strippable
-    field handed out a free exemption. Matching a date-SHAPED suffix and then
-    string-comparing it relocates that defect rather than killing it:
-    `prd-evade-0000-00-00` is not a date any calendar produces, and
-    `1970-01-01` parses fine but predates this project by decades. Both were
-    exempt. `strptime` alone does not close it -- the epoch date is valid --
-    so a plausibility floor is required too, and both branches fail closed.
+# The date-shape table that used to live here is gone with the mechanism it
+# guarded. `_prd_predates_floor` parsed a date out of the prd_id; Codex's
+# blocker showed the whole date-based exemption was wrong, so the function was
+# deleted rather than hardened a third time. A class dies best by deleting the
+# mechanism, not by adding a guard to it.
+
+
+def test_no_date_parsing_survives_in_the_receipt_gate():
+    """Executable proof the class is gone, not merely unused."""
+    source = (SCRIPTS_DIR / "prd_runner.py").read_text()
+    for dead in ("_prd_predates_floor", "_PRD_ID_DATE", "_PRD_DATE_EARLIEST"):
+        assert dead not in source, (
+            f"{dead} is back: the receipt gate must not infer eligibility "
+            "from any date")
+
+
+def test_the_cross_check_runs_under_the_writer_lock(judgment_repo, monkeypatch):
+    """Codex MAJOR on PR #102: the gate read the ledger under `ledger_lock`,
+    RELEASED it, and only then cross-checked the findings files. A concurrent,
+    perfectly valid triage landing in that gap writes a disposition the gate's
+    stale ledger snapshot cannot see, so approval false-blocks on a missing
+    receipt that does exist. The lock has to span the comparison, not just the
+    read that feeds it.
+
+    `ledger_lock` is re-entrant per thread, so depth > 0 is exactly the
+    assertion "a lock is held right now" without deadlocking the prober.
     """
-    runner = _load_module(f"pr_shape_{prd_id}", "prd_runner.py")
-    assert runner._prd_predates_floor(prd_id) is not requires_receipt
+    runner = _load_module("pr_lockspan", "prd_runner.py")
+    import judgment_compiler as jc
+
+    seen = []
+    real = jc.cross_check_findings
+    monkeypatch.setattr(jc, "cross_check_findings",
+                        lambda *a, **k: (seen.append(
+                            getattr(jc._LOCK_DEPTH, "value", 0)) or real(*a, **k)))
+    runner._judgment_receipt_gate(_cfg_for(judgment_repo), PRD_ID)
+    assert seen, "the cross-check never ran"
+    assert seen[0] >= 1, (
+        "cross_check_findings ran with the writer lock released; a concurrent "
+        "triage in that window false-blocks approval")

exec
/bin/zsh -lc "git diff --name-only -z origin/main...HEAD | while IFS= read -r -d '' f; do printf '{\"tool_input\":{\"file_path\":\"%s\"}}\\n' \""'$PWD/$f" | python3 q-system/.q-system/scripts/lessons-validator.py; printf '"'validator_exit=%s file=%s\\n' \""'$?" "$f"; done
python3 -c '"'import importlib.util,json,pathlib,subprocess; root=pathlib.Path(\".\"); spec=importlib.util.spec_from_file_location(\"scrub\",root/\"q-system/.q-system/scripts/lessons_scrub.py\"); m=importlib.util.module_from_spec(spec); spec.loader.exec_module(m); terms=m.codenames_from_registry(root/\"instance-registry.json\"); files=subprocess.check_output([\"git\",\"diff\",\"--name-only\",\"origin/main...HEAD\"],text=True).splitlines(); [(lambda p,text,fm: print(f\"{p}: id_matches_filename={fm[chr(105)+chr(100)]==p.stem} scrub_hits={m.find_client_data(text,terms)}\"))(root/f,(root/f).read_text(),{line.split(\":\",1)[0].strip():line.split(\":\",1)[1].strip() for line in (root/f).read_text().split(\"---\",2)[1].strip().splitlines()}) for f in files for p in [pathlib.Path(f)]]'
python3 -c 'import ast,pathlib; p=pathlib.Path(\"plugins/prd-os/scripts/judgment_compiler.py\"); t=ast.parse(p.read_text()); fs={n.name:n for n in t.body if isinstance(n,(ast.FunctionDef,ast.AsyncFunctionDef))}; print(\"judge_view_calls_in__judge_argv=\",[n.func.id for n in ast.walk(fs[\"_judge_argv\"]) if isinstance(n,ast.Call) and isinstance(n.func,ast.Name) and n.func.id==\"judge_view\"]); print(\"_judge_argv_source=\",ast.get_source_segment(p.read_text(),fs[\"_judge_argv\"]))'" in /Users/assafkipnis/.config/kipi/review-trees/pr-104
 succeeded in 41ms:
q-system/lessons/a-check-must-be-able-to-fail-for-the-reason-you-care-abou.md: id_matches_filename=True scrub_hits=[]
q-system/lessons/a-safety-property-enforced-at-n-sites-is-not-a-chokepoint.md: id_matches_filename=True scrub_hits=[]
q-system/lessons/an-artifact-must-carry-its-own-provenance-not-inherit-it.md: id_matches_filename=True scrub_hits=[]
q-system/lessons/kill-a-defect-class-by-deleting-the-mechanism-not-by-guar.md: id_matches_filename=True scrub_hits=[]
judge_view_calls_in__judge_argv= []
_judge_argv_source= def _judge_argv(model: str) -> list[str]:
    """The real invocation: `claude -p`, no tools.

    TOOLS OFF is a dataset requirement, not a hardening nicety. The packet's
    `duplicates[].source` is a filesystem path to a findings file and
    `prior_receipts` lists receipt ids; a judge that can open files reads prior
    HUMAN dispositions straight out of both, and the calibration set stops
    measuring prediction and starts measuring leakage.

    `claude -p` is the established LLM client in this fleet (no API key, rides
    the founder's subscription). The model is pinned and recorded on the run:
    unpinned headless jobs here silently rode a cheaper model and burned 3% of
    a weekly budget in an hour, and an unrecorded model change would confound a
    kappa shift with a capability shift.
    """
    # `--tools ""` is the AVAILABILITY control; `claude --help` states: "Use
    # "" to disable all tools". The first version used `--allowedTools ""`,
    # which is a permission ALLOWLIST and does not remove availability -- the
    # judge kept its tools and the paired test asserted the wrong flag, so the
    # test encoded the bug instead of catching it (Codex, PR #103 round 1).
    return ["claude", "-p", "--model", model, "--tools", ""]

exec
/bin/zsh -lc "nl -ba q-system/lessons/a-safety-property-enforced-at-n-sites-is-not-a-chokepoint.md | sed -n '1,24p'
python3 -c 'import ast,pathlib; p=pathlib.Path(\"plugins/prd-os/scripts/judgment_compiler.py\"); s=p.read_text(); funcs={n.name:n for n in ast.parse(s).body if isinstance(n,ast.FunctionDef)}; f=funcs[\"_judge_argv\"]; calls=[n.func.id for n in ast.walk(f) if isinstance(n,ast.Call) and isinstance(n.func,ast.Name)]; ret=next(n for n in ast.walk(f) if isinstance(n,ast.Return)); print(\"_judge_argv calls judge_view:\",\"judge_view\" in calls); print(\"_judge_argv return:\",ast.literal_eval(ret.value.elts[0]),ast.literal_eval(ret.value.elts[1]),ast.literal_eval(ret.value.elts[2]),\"<model>\",ast.literal_eval(ret.value.elts[4]),repr(ast.literal_eval(ret.value.elts[5])))'" in /Users/assafkipnis/.config/kipi/review-trees/pr-104
 succeeded in 0ms:
     1	---
     2	id: a-safety-property-enforced-at-n-sites-is-not-a-chokepoint
     3	kind: pattern
     4	title: A safety property enforced at N sites is not a chokepoint; apply single-writer to invariants too
     5	date: 2026-08-05
     6	---
     7	
     8	Single-writer discipline is usually applied to data paths: one function owns the write, so the write cannot be inconsistent. The same reasoning applies to INVARIANTS, and it is easy to miss because an invariant does not look like a resource.
     9	
    10	Observed on a component required to stay blind to a value it was being measured against. Blindness was preserved at four independent seams: a runtime flag controlling capability, the content of a rendered prompt, what got printed to an interactive session, and how references were validated. Three of the four were wrong, each discovered in a separate review round, because each seam was a separate opportunity to get it right and nothing forced them to agree. The fix was one function constructing the entire view the component was permitted to see, plus one test asserting that view carries no forbidden field. Three of the four seams then became ordinary code that structurally could not leak.
    11	
    12	The same shape appeared in a two-phase write whose consistency was maintained by careful ordering plus rollback. Three rounds each moved a boundary and each left a different inconsistency reachable, because rollback cannot undo an append. Recovering forward from a single irreversible step, with everything after it recomputable, removed the class rather than guarding it.
    13	
    14	How to apply:
    15	
    16	1. When a property must hold, count the places that preserve it. More than one means more than one chance to be wrong, and review will find them one round at a time.
    17	2. Build one constructor that establishes the property, and write ONE test asserting the property of its output. Prefer an allowlist over deleting known-bad fields: a field a future contributor adds is then excluded by default rather than leaking until someone remembers to exclude it.
    18	3. Ask what the component may CITE or emit, not only what it may perceive. Both are the same invariant and both fail independently.
    19	4. In a multi-step write, make exactly one step irreversible. Everything before it is validated and reversible; everything after it must be derivable from what is already durable, so a late failure is recovered by recomputation rather than by an impossible rollback.
    20	5. A late-stage failure of a derived artifact must never escalate into refusing an operation that already succeeded. Report it, repair it inline if bounded, and let a separate verifier own the residual state.
    21	
    22	The general contract: if you would apply single-writer to a data path, apply it to the invariant. A property maintained by N cooperating precautions is a property waiting for the round that finds the N+1th seam.
_judge_argv calls judge_view: False
_judge_argv return: claude -p --model <model> --tools ''

exec
/bin/zsh -lc "sed -n '1,130p' q-system/hooks/lessons-index.py; printf '{\"cwd\":\"%s\"}\\n' \""'$PWD" | python3 q-system/hooks/lessons-index.py | rg '"'A check must|A safety property|An artifact must|Kill a defect'" in /Users/assafkipnis/.config/kipi/review-trees/pr-104
 succeeded in 0ms:
#!/usr/bin/env python3
"""SessionStart consumer for the cross-instance lessons corpus.

Reads q-system/lessons/*.md frontmatter titles and injects them (titles only,
capped, bodies on demand) via hookSpecificOutput.additionalContext so every
instance sees the available lessons. Part of PRD prd-cross-instance-learning-2026-06-19.

Fail-closed and never-blocks: any error -> emit nothing, exit 0. The QROOT
resolution mirrors session-start.py get_qroot (flat + nested q-system/q-system/
subtree layout); the additionalContext shape mirrors voice-dna-loader.py:152 /
token-guard.py:336 (hookSpecificOutput.additionalContext); the SessionStart form additionally sets hookEventName. stdlib only.
"""
import json
import os
import sys
from pathlib import Path

CAP = 20


def get_qroot(project_dir):
    nested = Path(project_dir) / "q-system" / "q-system" / "canonical"
    if nested.exists():
        return Path(project_dir) / "q-system" / "q-system"
    return Path(project_dir) / "q-system"


def frontmatter(path):
    try:
        text = path.read_text()
    except Exception:
        return {}
    if not text.startswith("---"):
        return {}
    end = text.find("\n---", 3)
    if end == -1:
        return {}
    fm = {}
    for line in text[3:end].splitlines():
        line = line.strip()
        if not line or line.startswith("#") or ":" not in line:
            continue
        k, _, v = line.partition(":")
        fm[k.strip()] = v.strip().strip('"').strip("'")
    return fm


def main():
    try:
        project_dir = os.environ.get("CLAUDE_PROJECT_DIR", os.getcwd())
        lessons = get_qroot(project_dir) / "lessons"
        if not lessons.is_dir():
            sys.exit(0)
        items = []
        for f in sorted(lessons.glob("*.md")):
            if f.name == "README.md":
                continue
            fm = frontmatter(f)
            title = fm.get("title")
            if title:
                items.append((fm.get("date", ""), title))
        if not items:
            sys.exit(0)
        items.sort(key=lambda x: x[0], reverse=True)  # most recent date first
        titles = [t for _, t in items[:CAP]]
        body = "# Cross-instance lessons (titles only; read q-system/lessons/<file> for detail)\n" + \
            "\n".join("- " + t for t in titles)
        print(json.dumps({"hookSpecificOutput": {"hookEventName": "SessionStart", "additionalContext": body}}))
        sys.exit(0)
    except Exception:
        sys.exit(0)  # never block session start


if __name__ == "__main__":
    main()
{"hookSpecificOutput": {"hookEventName": "SessionStart", "additionalContext": "# Cross-instance lessons (titles only; read q-system/lessons/<file> for detail)\n- A check must be able to fail for the reason you care about, or it is decoration\n- A safety property enforced at N sites is not a chokepoint; apply single-writer to invariants too\n- An artifact must carry its own provenance, never inherit it from where it was filed\n- Kill a defect class by deleting the mechanism, not by adding a guard to it\n- A retry changes the attempt, a different model changes the reasoning\n- A supervised restart runs in a stripped environment, and auth errors will lie about it\n- A zero result must prove it is empty, not broken\n- An output nobody reads is the same as no output\n- Every stage needs its own trigger\n- Name an ownership boundary for reads, not just writes\n- Prove a negative with a live probe\n- Prove delivery with receipts, not proxies\n- Silent zero-output in gated automation\n- Split the act path from the verify path\n- The last path insert wins: order your module search path deliberately\n- Consolidating duplicate answers changes behaviour exactly where the copies disagreed\n- A cross-process handshake needs one source and a test that pins both ends\n- A dedup ledger must be append-only, single-writer, and floor-checked\n- A freshness deadman must live off the machine it watches\n- A gate is only real if production writes what it reads"}}

codex
## Finding 1

**MINOR:** The lesson claims one constructor defines the judge’s entire permitted view, but tool availability remains an independent blindness boundary outside `judge_view`.

**Location:** `q-system/lessons/a-safety-property-enforced-at-n-sites-is-not-a-chokepoint.md:10`

**Reproducer command:**

```bash
python3 -c 'import ast,pathlib; p=pathlib.Path("plugins/prd-os/scripts/judgment_compiler.py"); s=p.read_text(); funcs={n.name:n for n in ast.parse(s).body if isinstance(n,ast.FunctionDef)}; f=funcs["_judge_argv"]; calls=[n.func.id for n in ast.walk(f) if isinstance(n,ast.Call) and isinstance(n.func,ast.Name)]; ret=next(n for n in ast.walk(f) if isinstance(n,ast.Return)); print("_judge_argv calls judge_view:","judge_view" in calls); print("_judge_argv return:",ast.literal_eval(ret.value.elts[0]),ast.literal_eval(ret.value.elts[1]),ast.literal_eval(ret.value.elts[2]),"<model>",ast.literal_eval(ret.value.elts[4]),repr(ast.literal_eval(ret.value.elts[5])))'
```

**Real output:**

```text
_judge_argv calls judge_view: False
_judge_argv return: claude -p --model <model> --tools ''
```

The constructor controls the prompt payload. `--tools ""` separately controls what the judge can perceive outside that payload. Removing either boundary breaks blindness. The lesson should describe these as two required chokepoints instead of claiming one function constructs the entire permitted view.

## What is sound

- All four files passed the real `lessons-validator.py`.
- Every frontmatter `id` matches its filename.
- The production scrubber, using codenames from the real instance registry, reported `scrub_hits=[]` for all four files.
- The SessionStart hook successfully surfaced all four titles.
- No malformed frontmatter, client token, absolute path, email, URL, or registry codename survived.
- I attempted the repository’s temp-backed lesson test. The read-only sandbox rejected `mktemp`, so that suite could not execute. The direct production-path checks above required no repository writes.

Fix first: narrow the single-constructor claim so it acknowledges the independent tool-availability boundary.

FINDINGS:
minor|The lesson falsely claims one constructor defines the judge's entire permitted view even though tool availability remains an independent blindness boundary|q-system/lessons/a-safety-property-enforced-at-n-sites-is-not-a-chokepoint.md:10
END FINDINGS
tokens used
66,007
## Finding 1

**MINOR:** The lesson claims one constructor defines the judge’s entire permitted view, but tool availability remains an independent blindness boundary outside `judge_view`.

**Location:** `q-system/lessons/a-safety-property-enforced-at-n-sites-is-not-a-chokepoint.md:10`

**Reproducer command:**

```bash
python3 -c 'import ast,pathlib; p=pathlib.Path("plugins/prd-os/scripts/judgment_compiler.py"); s=p.read_text(); funcs={n.name:n for n in ast.parse(s).body if isinstance(n,ast.FunctionDef)}; f=funcs["_judge_argv"]; calls=[n.func.id for n in ast.walk(f) if isinstance(n,ast.Call) and isinstance(n.func,ast.Name)]; ret=next(n for n in ast.walk(f) if isinstance(n,ast.Return)); print("_judge_argv calls judge_view:","judge_view" in calls); print("_judge_argv return:",ast.literal_eval(ret.value.elts[0]),ast.literal_eval(ret.value.elts[1]),ast.literal_eval(ret.value.elts[2]),"<model>",ast.literal_eval(ret.value.elts[4]),repr(ast.literal_eval(ret.value.elts[5])))'
```

**Real output:**

```text
_judge_argv calls judge_view: False
_judge_argv return: claude -p --model <model> --tools ''
```

The constructor controls the prompt payload. `--tools ""` separately controls what the judge can perceive outside that payload. Removing either boundary breaks blindness. The lesson should describe these as two required chokepoints instead of claiming one function constructs the entire permitted view.

## What is sound

- All four files passed the real `lessons-validator.py`.
- Every frontmatter `id` matches its filename.
- The production scrubber, using codenames from the real instance registry, reported `scrub_hits=[]` for all four files.
- The SessionStart hook successfully surfaced all four titles.
- No malformed frontmatter, client token, absolute path, email, URL, or registry codename survived.
- I attempted the repository’s temp-backed lesson test. The read-only sandbox rejected `mktemp`, so that suite could not execute. The direct production-path checks above required no repository writes.

Fix first: narrow the single-constructor claim so it acknowledges the independent tool-availability boundary.

FINDINGS:
minor|The lesson falsely claims one constructor defines the judge's entire permitted view even though tool availability remains an independent blindness boundary|q-system/lessons/a-safety-property-enforced-at-n-sites-is-not-a-chokepoint.md:10
END FINDINGS

PR #111, opened 2026-08-05
P0: refuse a kipi update whose --delete would remove founder data (ASK-402)

assafkip:
Split out of #110 so the P0 does not wait on a design question.

## The bug, reproduced

A real `kipi update` deleted `my-project/current-state.md` and `memory/last-handoff.md` from an instance.

The excludes that protect instance data are ANCHORED to the rsync transfer root, so they only line up when the destination IS the instance's `q-system` dir. With a null `subtree_prefix` the destination is the instance ROOT, and the data sits one level below where the excludes point:

```
*deleting q-system/my-project/current-state.md
*deleting q-system/memory/last-handoff.md
```

`reddit-build-radar` is a real registry entry with a null prefix.

## The fix

Re-anchoring fixes that variant, not the class: any future layout/prefix/registry drift reopens the same hole and the failure mode is deleted founder data with no error.

So `kipi-update-deletion-guard.py` reads the deletions rsync **actually plans**, from rsync's own itemized dry-run output, and refuses if any touch instance-owned names at any depth. The excludes stay as the first line; this is the fail-closed backstop.

## Evidence

- **14/14** in `test-kipi-update-owned-deletion-guard.sh`, including an end-to-end case that runs a real `kipi update` and asserts the instance data survived.
- **6/6 mutants killed**, including deleting the guard's call site from `kipi-update.sh` — the end-to-end test then reports *"a real kipi update DESTROYED instance-owned data"*. The wiring is tested, not assumed.
- Fail-closed verified two ways: a missing guard script makes `python3` exit 2 (abandons the instance), and `set -o pipefail` means a failed rsync fails the pipeline rather than handing the guard empty stdin.
- Case-folding is deliberate — macOS is case-insensitive, so `My-Project` and `my-project` are one directory; a case-exact match would let a real deletion through on the platform this runs on.
- The other six kipi-update suites and `validate-separation.py` pass on this branch off `main`.

## Why split from #110

Codex called this guard sound in six consecutive review rounds. The receipt-contract rework on that branch drew a new major in four of them, and the remaining one ("what is machine evidence that a review ran") is a design decision, not a patch — a clean review legitimately writes no findings, so there is nothing to bind the receipt to without redesigning how a review attests itself. That work stays on #110.

## Known limitation (captured, not fixed)

[...]

PR #115, opened 2026-08-06
The deletion guard stops claiming coverage it never computed (sp-5e857e2a)

assafkip:
The module docstring said it "also covers sp-10cf4f76 (a tracked instance-only script inside the synced tree) because a deletion is a deletion whatever git thinks of the file". It does not. `owned_hits()` matches seven instance-owned **data-directory** names; a script is not one of them.

Reproduced on this branch before touching anything:

```
*deleting q-system/.q-system/scripts/income-scan.sh   -> exit 0
*deleting q-system/my-project/current-state.md        -> exit 2
```

That first path is how the fractional-cxo income scanners died silently for six days. Reading the false sentence almost caused sp-10cf4f76 to be voided as already-fixed. A guard that RECORDS a claim it never COMPUTED is worse than an obvious gap, because the next reader trusts it.

## Docstring only. No behaviour change. Deliberate.

The narrow fix that would close the reproducer (add `scripts` to `INSTANCE_OWNED`) is **not safe**, and that is measured rather than argued:

| | |
|---|---|
| skeleton-owned files in `q-system/.q-system/scripts/` | **141** |
| files the skeleton has deleted from that directory historically | **6** |

Those six are legitimate propagation. A name-based entry for `scripts` would refuse every one of them on all 24 instances — the **PR #111 failure mode** this file already carries a scar for, where "any depth" matching refused the sync fleet-wide over `agent-pipeline/templates/deck/output/`. A guard that halts unattended updates gets switched off, and a gate that is off protects nothing.

## The design call

Instance-only and skeleton-owned files are **interleaved in the same directories**, so no path name can separate them. The allowlist is not merely lagging; it is the wrong kind of predicate.

The discriminator has to be **provenance**: did the skeleton ever ship this file (legitimate removal) or did the instance create it (data loss)? The skeleton's git history answers that.

I did not ship it here. Its blast radius is every `kipi update` on 24 instances, and its false-positive rate is **unmeasured** — instances that generate transient files inside the synced tree would start refusing every update. Shipping an unmeasured fleet-wide refusal is the exact mistake `design-auto-invoke.md` warns about. The prerequisite is a read-only `kipi update --dry` census across the fleet, which is a measurement, not a fix.

**So the reproducer still exits 0.** The bar asked me to flip it to 2. I am reporting that I did not, and why, rather than flipping it by a route that breaks unattended updates fleet-wide. `sp-10cf4f76` stays OPEN and now says so in the code that was misdescribing it.

## Verified

- guard suite `test-kipi-update-owned-deletion-guard.sh` — 17/17 green
- both reproducer lines unchanged (0 and 2)
- PR #111 regression path (`deck/output`) still exits 0
- `ast.parse` clean

Codex is out of credits until Aug 9, so the claude reviewer fills PRIMARY and posts DEGRADED.


assafkip:
reopening to re-fire the workflow after the Actions outage

PR #117, opened 2026-08-06
--dry stops manufacturing a failure out of a directory named build (sp-f6733ee3)

assafkip:
`model_rsync_excludes()` strips any directory whose **basename** is a build-cache name, then `.git` is copied verbatim. A **tracked** file under such a directory is absent from the model worktree but still in the model index — so git reports a deletion, the tracked-changes predicate reads the tree as dirty, and a healthy instance is abandoned.

Measured 2026-08-06 on `interview-coach`: the model printed `D design-room/build/gate-report.md` while live that file is **tracked, present, and the tree has zero tracked changes**. `design-room/build/` is authored gate output, not a build cache. The fleet census reported `Failed: 4` when only 3 were real.

Second time `--dry` has invented a failure. The disk-pressure one is already recorded three lines above in the same function.

## Model-only, not data loss

Checked in the *reassuring* direction on purpose — two claims earlier in this session were wrong in the alarming direction, which makes a comfortable conclusion the one most worth re-testing. Four ways:

1. `MODEL_EXCLUDES` appears in exactly two rsyncs, and `$path/` is the **source** in both
2. the real sync uses `rsync_owned_excludes` — a different function
3. `rsync_owned_excludes` contains **no** build-cache name
4. the real sync's destination is `$path/q-system/`; `design-room/` sits at instance root, outside it

## The fix asks git

A directory holding tracked files is not a build cache whatever it is named. The function's own comment already stated the real predicate — *"regenerable by its own toolchain AND gitignored"* — it just was not being checked.

This cannot reintroduce the 8.7G `src-tauri/target` problem that motivated the excludes: tracked content is committed content, and that tree is gitignored, so it stays excluded.

## Case 2 is the one that matters

Deleting the dirty-tree predicate outright would satisfy case 1. So case 2 holds the other side: an instance with a **real** tracked modification must still be refused.

| | Result |
|---|---|
| Control | 2/2 green |
| Tracked-check removed (pre-fix) | case 1 **RED**, printing the exact phantom `D design-room/build/gate-report.md`; case 2 still passes |

The mutant validates its own anchor before running.

## Third instance of one defect class

This is the same mistake as `INSTANCE_OWNED` (a name allowlist deciding what is founder data) and `[ "$DRY_RUN" != "1" ]` (a comparison that can never be false): **a name treated as a fact about what something is.** Written up as a lesson in #116 rather than patched a fourth time.

## Verified

Local only. **CI is down; nothing here is CI-verified.** 2/2 green, `bash -n` clean, manifest parses.

### DIFF
--- a/kipi-update.sh
+++ b/kipi-update.sh
@@ -242,9 +242,29 @@ model_rsync_excludes() {
   # and the instance reported FAILED in --dry while the real run had updated it
   # fine. So --dry manufactured a false failure out of disk pressure alone.
   # Every path here is regenerable by its own toolchain and gitignored.
+  # A BASENAME IS NOT A FACT ABOUT WHAT A DIRECTORY IS (sp-f6733ee3). The names
+  # below are a heuristic for "regenerable build cache", and it is right nearly
+  # every time -- which is why the exception stayed invisible. Measured
+  # 2026-08-06 on interview-coach: `design-room/build/gate-report.md` is AUTHORED
+  # gate output living under a directory that happens to be called `build`. The
+  # rsync skipped it while `.git` was copied verbatim, so the model saw a tracked
+  # DELETION, read the tree as dirty, and reported a healthy instance FAILED.
+  # Second time --dry has manufactured a false failure (the disk-pressure one
+  # above was the first), and a fleet health check driven off --dry would page on
+  # instances that are fine.
+  #
+  # SO ASK GIT, WHICH KNOWS. A directory holding tracked files is not a build
+  # cache whatever it is named -- the comment above already states the real
+  # predicate ("every path here is regenerable by its own toolchain AND
+  # GITIGNORED"), it just was not being checked. Tracked content is committed
+  # content, so copying it cannot reintroduce the 8.7G src-tauri/target problem:
+  # that tree was gitignored and stays excluded.
   MODEL_EXCLUDES=(--exclude=".git")
   local cache_dir
   for cache_dir in target node_modules .venv venv __pycache__ .next dist build .pytest_cache .mypy_cache .ruff_cache; do
+    if [ -n "$(git -C "$instance_root" ls-files -- "$cache_dir/*" "*/$cache_dir/*" 2>/dev/null | head -1)" ]; then
+      continue
+    fi
     MODEL_EXCLUDES+=(--exclude="$cache_dir/")
   done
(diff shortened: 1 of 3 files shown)


assafkip:
## Closing: superseded on main, and the test here cannot go red

Measured today while draining the DIRTY backlog, not inferred.

### The fix is already on main, by another route

`model_rsync_excludes()` on main carries the same guard against the same defect, tagged with the same spillover id:

```bash
case "$tracked_nl" in
  *$'\n'"$cache_dir"/*|*"/$cache_dir"/*)
    # Tracked content lives here. Excluding it would fake a deletion.
    continue
    ;;
esac
```

Behaviourally equivalent to this branch's `git ls-files` version. Proof: I resolved the merge taking main's side and ran **this branch's own test** against it. 2 passed, 0 failed.

### But that test cannot fail

With main's implementation in place I deleted the tracked-content check outright, restoring the exact defect, and re-ran this branch's test:

```
== 1. THE DEFECT: tracked content under build/ must NOT fail a clean instance ==
  PASS: a clean instance with tracked design-room/build/ content is NOT refused
== 2. THE OTHER SIDE: a genuinely dirty instance must STILL be refused ==
  PASS: a real tracked modification is still refused (the guard still guards)
-------- 2 passed, 0 failed --------
```

Green with the fix, green without it. Case 1 asserts only that the run was **not** refused, which any run that gets nowhere also satisfies — the case-1 log shows the model reaching `MODELED_FINAL_DIFF_END` with no dirty-tree evaluation at all. The `refused` grep itself is fine; case 2 depends on it matching and case 2 discriminates.

### Main's reproducer does discriminate

`test-kipi-update-cache-exclusion.sh` on main, same `sp-f6733ee3`:

```
=== the UNTRACKED cache must still be excluded (control) ===
  OK: untracked node_modules/ still stripped
  OK: build/ not excluded, because git tracks something there
=== negative self-test: revert the fix, the check must go RED ===
  OK [mutant]: reverting the fix re-excludes build/, so the check can fail
PASS
```

An untracked control **and** an in-test mutant that proves the assertion can fail. Strictly stronger than what this branch adds.

### So

The code change is redundant and the test is decoration, so merging this would add false confidence rather than coverage. Closing rather than merging. Reopen if I have read either side wrong — the whole check is two commands and both are quoted above.

The finding worth keeping is the test, not the fix: it is another instance of a check that passes for a reason nobody named. Captured as spillover.

https://claude.ai/code/session_01YWxNWmEobPKcJ6HqqUBMTw

PR #114, opened 2026-08-06
The verdict record names the model that actually wrote the review (ASK-287)

assafkip:
Revives an orphan commit from 2026-08-02 (ASK-287) that was never opened as a PR and never reached main.

## Why now

Codex is out of credits until Aug 9. Probed today with a real billed `codex exec`:

```
ERROR: You've hit your usage limit ... try again at Aug 9th, 2026 6:53 PM.
```

So for the next three days **every** review in this repo runs the Opus fallback as PRIMARY and posts DEGRADED. This is precisely the commit that makes the verdict record say so. Unmerged, it protected nothing.

## The defect

The human-facing surfaces were already honest: the commit-status description and the Linear comment both say "DEGRADED (codex down, Opus fallback)". The machine-readable record that `converge.sh` and `linear-worker.sh` gate on said `"engine": "codex"` for reviews codex never produced. Measured 2026-08-02: PR #66 and #67 both carried `engine=codex` on Opus-written reviews.

Truthful prose over a lying record is worse than an obvious gap, because the wrong reader trusts it silently.

## The change

Additive, record-only, no gate moves:
- `reviewed_by` — the model that produced the prose. `engine` stays the flag the run was invoked with; on the fallback they disagree, and that disagreement IS the record of the outage.
- `degraded` — bool.

Derived, never branched. The fallback trigger is the one path in this script where a wrong edit posts an unearned green, so this reads existing state and adds nothing to the control flow.

## What changed vs the orphan

Main moved under it. `ASK-352` added a 13th positional arg (`$REVIEW_USABLE`) to the same `python3` call, so the cherry-pick conflicted on exactly those lines. Resolved by keeping both: the record now carries `usable` + `reviewed_by` + `degraded` and unpacks `argv[1:16]`.

Two invisible placement constraints are now why-commented at the derivation:
- **`set -e` is OFF here** (line 64 is `set -uo pipefail`) and these lines depend on it. Under `-e`, a false `[ ... ] && assign` is an AND-list returning 1, which would abort every healthy codex review right before its record is written.
- **The derivation sits below `if review_is_usable` on purpose.** The test extracts it by awk range and executes it; moving it up pulls an undefined function call into the extracted block.

## Watched RED before trusted GREEN

The test is 180 lines nobody had ever seen fail. It drives the **shipped** writer (extracted from the script at run time), not a reimplementation. Four external mutants, each run against a copy of the tree, each killing a different assertion:

| Mutant | Result |
|---|---|
[...]

PR #119, opened 2026-08-06
spec: review independence is derived, not the degraded cause-flag

assafkip:
**Spec only, no code.** #114 is under re-review and the consumer fix needs its fields to exist first, so this is written to be buildable the moment #114 lands.

`.prd-os/issues/review-independence-is-derived-not-declared.md`

## The defect

`degraded` encodes a **cause** — "codex was asked and failed" — while every consumer reads it for a **property**: "is this an independent second opinion?" Those come apart on a path that already exists.

**Writer half** (reviewer finding 3 on #114): a deliberate `--engine claude` run records `degraded: false` because codex was never asked, so nothing failed. But an Opus review is not a second lab's opinion regardless of *why* it happened, and the consumer reference maps `degraded == false` to independent. `--engine claude` is a live path, and with `KIPI_REVIEW_PRIMARY_ENGINE=claude` it writes the **root gating record**.

**Reader half** (`sp-cc2de280`): `verify-codex-review-live.sh` selects on `engine != "codex" -> continue`, then reports *"a codex-engine review really ran"*. A DEGRADED Opus record also carries `engine: codex` — so the one check whose entire job is proving codex is alive can be satisfied by a review codex never wrote.

Same defect from both ends, so one issue: **derive the property at the writer, fix the reader to consult it.**

## What the spec carries

- **Both reproducers, red first**, driven against the shipped code rather than a reimplementation
- **Mutation bar**: each assertion killed by its own mutant, neither masking the other, every mutant validated as applied, plus the anchor-survival check (a suite exiting 1 from its own extraction guard is not a case-level kill)
- **Fail-safe direction**: a record with no `reviewed_by` is UNKNOWN, never independent — same direction the existing `invoker` handling takes
- **`independent` is DERIVED AND RECOMPUTED AT READ TIME, never stored.** A stored independence flag is the same defect one turn later: a field asserting a property nothing recomputed. `reviewed_by` is an observation; independence is a judgement *about* it, and judgements go stale when the primary engine changes.

## On `bypass_check`

It runs **both suites in full, with no selector**, deliberately. The other Sana just found a `bypass_check` selecting 6 of 23 tests with the security-property test among the 17 omitted — and an unquoted `-k` in this repo once meant zero tests ran while the harness reported perfect survival. Any future narrowing must be proven to still include both cases by deleting each and watching it go red.

## Time sensitivity

Codex is out until 2026-08-09. Every review until then runs the fallback, and **until this lands the live-proof check will certify reviews codex never wrote.**

## Verified

Spec structure only: 3 deliverable checkboxes matching `deliverables_count`, all `allowed_files` paths exist. No code changed, so nothing to test.

### DIFF
--- a/.prd-os/issues/review-independence-is-derived-not-declared.md
+++ b/.prd-os/issues/review-independence-is-derived-not-declared.md
@@ -1,0 +1,124 @@
+---
+id: review-independence-is-derived-not-declared
+title: Review independence is a derived property, not the degraded cause-flag
+status: open
+priority: p1
+parent_prd: null
+blocked_by: "PR #114 (ASK-287) must land first: reviewed_by/degraded do not exist on main yet"
+allowed_files:
+  - q-system/.q-system/scripts/pr-review-agent.sh
+  - q-system/.q-system/scripts/verify-codex-review-live.sh
+  - q-system/.q-system/scripts/test/test-review-degraded-provenance.sh
+  - q-system/.q-system/scripts/test/test-review-independence-property.sh
+  - q-system/.q-system/capability-manifest.json
+disallowed_files:
+  - q-system/.q-system/scripts/converge.sh
+  - q-system/.q-system/scripts/linear-worker.sh
+  - .prd-os/**
+  - plugins/**
+  - instance-registry.json
+required_checks:
+  - bash q-system/.q-system/scripts/test/test-review-independence-property.sh
+  - bash q-system/.q-system/scripts/test/test-review-degraded-provenance.sh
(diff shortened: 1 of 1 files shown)


assafkip:
Closing as superseded, branch kept. #319 (2026-09-06) made claude the gating engine and codex advisory, so check 8 in verify-codex-review-live.sh now selects engine == claude and the engine == codex reader this spec fixes no longer exists. The write-time property this spec wanted lives on as sp-bd35985f (a gating boolean on the verdict record). Reopen if that item wants this text as its starting point.

PR #98, opened 2026-08-04
Stop the fleet updater blocking on its own exhaust, and name the instances it skips (ASK-363)

assafkip:
## The defect

`kipi update` silently stopped updating part of the fleet. Measured 2026-08-04: **6 of 23 instances sat 2–6 prd-os versions behind**, and each run reached a *different* subset — their `plugin.json` mtimes clustered at 07-30 and 08-01.

Root cause is at `kipi-update.sh` — the dirty-tree guard refuses an instance on **any** tracked modification. That is correct for founder work and it is the reason the guard exists ("must never package unrelated founder edits into an infra commit"). But it cannot distinguish founder work from the system's **own exhaust**. Four of the six were blocked *solely* by files this fleet writes:

- `q-system/memory/.sycophancy-monthly-stamp` — written by the monthly sycophancy hook (5 instances)
- `.claude/state/stop-gate-firings.json` — hook state
- `plugins/memory-lifecycle/` — shipped into travel-agent by an **earlier `kipi update`** and left staged, which then blocked every update after it

The updater was blocked by its own output, and the summary printed `Failed: 6` with no names, so nobody could see which instances were drifting.

## Two changes

1. **`SYSTEM_OWNED_PATHS`** — a narrow, explicit allowlist committed immediately before the guard runs. Each path is committed only if actually dirty. Anything not on the list is still treated as founder work and still refuses. The founder-protection intent is unchanged.
2. **The summary names every instance it did not update.** A count is not a report.

## Verification

- `bash -n kipi-update.sh` clean.
- The named-failure report immediately surfaced a real problem I could not see before: `accountant` fails its dry-run copy with `No space left on device` (disk 93%, and that instance carries an 8.7G `src-tauri/target/debug` tree the dry model has no reason to copy). Captured separately as `sp-b2f16971` — **not** fixed here, and not by deleting anyone's build cache.
- Fleet is now **22/22 governed instances at prd-os 0.12.0** (`reddit-build-radar` is `skeleton_managed: false` by design, ASK-117).

## Honest limits

- This unblocks the recurrence for the three named paths only. A *new* system-written file will re-block the fleet the same way, and the allowlist is hand-maintained.
- The four instances unblocked today were fixed by hand-committing their churn; this change prevents the next occurrence, it did not cause today's recovery.
- I did not touch the dry-run rsync scope, so `--dry` can still report false failures under disk pressure.

[...]

PR #124, opened 2026-08-07
auto-commit stops sweeping source code into generic commits (ASK-498)

assafkip:
`FALLBACK_AREA` caught every path not named in `AREA_MAP`, so an instance's own source tree went into an unattended `chore: update project files` commit with no description and no issue id.

## Measured

Three sweeps on the consulting instance in a single session (`d96e621`, `7a252f4`, `f0a3183`) carried real feature work onto `main`. Two also **raced the agent writing the files** — a `git add` of new files reported success and staged nothing, because the hook had already committed them, so the agent's own commit silently held half the change.

## The distinction

The hook is a safety net for **generated state** (canonical, session memory, marketing content) that nobody would otherwise commit. Source code is the opposite: it is exactly what an agent commits deliberately, with a real message and a Linear id. Sweeping it is a second writer to the same branch.

- No fallback. `classify()` returns `SKIP_UNCLASSIFIED`; `group_files` never commits it.
- `report_skipped()` prints what was left. Silence would be the same defect in reverse: work uncommitted with nobody told.
- Declared skips (`q-system/output`) stay quiet.
- The `[no-issue:]` hatch is unchanged, so real auto-commits stay countable.

Uncommitted is not lost. The files are on disk.

## Evidence

9 new tests in `q-system/.q-system/tests/test_auto_commit.py`, **6 of 6 mutants killed**.

One mutant exposed an accidental shield: `get_changed_files` pre-filters `q-system/output`, so the declared-skip branch was unreachable through the CLI and its test passed for the wrong reason. It now drives `group_files` directly.

https://claude.ai/code/session_01X77DsK9jLSLAiA4vsQ652E

### DIFF
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -65,6 +65,18 @@ jobs:
 
       - name: Run prd-os tests
         run: pytest plugins/prd-os/tests/ -q
+
+      # The skeleton's own harnesses had NO runner (adversarial review finding-5):
+      # test_auto_commit, test_token_guard and test_fable_escalation all sat unrun,
+      # so "a test locks the property" was an unenforced claim. wiring-check.md is
+      # explicit that a Python harness needs a caller.
+      - name: Run skeleton hook tests
+        # SCOPED past separation/ deliberately: that subdir has 6 failures that
+        # predate this change and are untouched by it. Widening CI onto someone
+        # else's red is how a gate starts paging the founder for a bug the PR did
+        # not cause. Tracked separately; this step covers the hook harnesses.
+        run: pytest q-system/.q-system/tests/ -q
+             --ignore=q-system/.q-system/tests/separation
         env:
           CI: true
 
--- a/.github/workflows/validate.yml
+++ b/.github/workflows/validate.yml
@@ -71,12 +71,13 @@ jobs:
       # so "a test locks the property" was an unenforced claim. wiring-check.md is
       # explicit that a Python harness needs a caller.
       - name: Run skeleton hook tests
-        # SCOPED past separation/ deliberately: that subdir has 6 failures that
-        # predate this change and are untouched by it. Widening CI onto someone
-        # else's red is how a gate starts paging the founder for a bug the PR did
-        # not cause. Tracked separately; this step covers the hook harnesses.
-        run: pytest q-system/.q-system/tests/ -q
-             --ignore=q-system/.q-system/tests/separation
+        # SCOPED TO THIS FILE deliberately. The sibling harnesses in that directory
+        # were unrun for a reason: separation/ has 6 failures predating this branch,
+        # and test_token_guard.py passes locally but fails 11 in CI (it depends on
+        # git identity and /tmp state the runner does not provide). Widening the gate
+        # onto either is how CI starts paging the founder for a bug this PR did not
+        # cause. Both are tracked; unignore them as they are fixed.
+        run: pytest q-system/.q-system/tests/test_auto_commit.py -q
         env:
           CI: true
 
--- a/q-system/.q-system/tests/test_auto_commit.py
+++ b/q-system/.q-system/tests/test_auto_commit.py
@@ -223,3 +223,25 @@ def test_classify_covers_skeleton_and_instance_alike():
     assert mod.classify("q-thaena/my-project/x.json") == ("content", "update project state")
     assert mod.classify("q-consult/output/x.json") == mod.SKIP_DECLARED
     assert mod.classify("q-consult/pipeline/x.py") == mod.SKIP_UNCLASSIFIED
+
+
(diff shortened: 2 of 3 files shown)


assafkip:
**Blocked, not abandoned.** `validate` is green (all 12 steps). The merge is held by the other required context, `kipi/reviewer-approved`, which has never been posted on this head.

That context comes from a Codex verdict. Probed just now from a trusted dir with stdin closed:

```
ERROR: You've hit your usage limit ... try again at Aug 9th, 2026 6:53 PM.
```

So the reviewer cannot run until **Aug 9 18:53**.

I am deliberately **not** using `--admin` to merge past it. `enforce_admins` is false, so it would work, but this PR changes a Stop hook that runs in every instance in the fleet — that is exactly the change the review gate exists for. Merging around it would be working around a gate that is doing its job.

Note for whoever picks this up: `gh pr checks` shows only `validate` and reports everything green. It does not surface a required context that was never posted, so the PR looks mergeable when it is not. `gh api repos/.../commits/<sha>/status` is what shows the empty context list.

**Consequence meanwhile:** the fix is not live in the consulting instance. The old hook swept source into generic `chore: update project files` commits twice more during the session it was written in.

assafkip:
## Adversarial review — Claude (Opus), 2026-08-07

Codex is rate-limited until Aug 9 18:53, so per founder instruction ("stop waiting on codex - use claude") the reviewer changed. **The review did not become a formality.** It ran adversarially against the diff with a mandate to find defects, and it found five, two of them blocking. Every one was reproduced independently before being fixed.

### finding-1 — CRITICAL, fixed in `e869dbcc`
`git commit -m` with **no pathspec commits the entire index**. A file an agent had staged but not committed was swept in anyway, while `report_skipped` printed that it had NOT been committed.

Reproduced: the hook printed `NOT committed q-consult/pipeline/repo_links.py` and the resulting commit contained that exact file. A false report is worse than the silence it replaced — it tells the next session the file is still theirs. This is the same race the original incidents describe, so the PR as first written did not close the incident it cites. `kipi-update.sh` had already fixed this identical defect once (its PR #98 note says so verbatim); it returned through a different door.

Now `git commit -m <msg> -- <files>`.

### finding-2 — CRITICAL, fixed in `e869dbcc`
Every `AREA_MAP` prefix starts `q-system/`, which is the **skeleton**. An instance keeps its real content one segment over. Measured on the consulting instance: **1047 of 2099 tracked files unclassified**, including `my-project` (the system of record), `canonical` and `marketing`.

Removing the fallback without fixing this would have disabled the safety net for exactly the generated state it exists to protect — the outcome inverted from the stated intent. `INSTANCE_AREAS` now matches the path with its first segment stripped, so one row covers every instance with no registry lookup. Source trees stay excluded; `output/` is a declared skip.

After: unclassified 1047 → 719, and every remaining one is source, tests or build state.

### finding-4 — MAJOR, fixed in `42bd2fed`
`report_skipped` print()ed into a stream nobody reads: the hook is wired `async` and the fleet template appends `2>/dev/null`. The entire replacement for the removed fallback was a message the operator never receives — the silently-dropped-notification scar `founder-notifications.md` exists for. Now routed through `slack-notify.sh`, which no-ops when unconfigured and swallows every error.

### finding-5 — MAJOR, fixed in `42bd2fed` + the follow-up
Nothing ran the test file. `validate.yml` ran only `plugins/prd-os/tests/`. CI now runs `test_auto_commit.py`.

Scoped to that one file deliberately: the sibling harnesses in that directory fail for reasons this PR did not cause — `separation/` has 6 pre-existing failures, and `test_token_guard.py` passes locally (53 green) but fails 11 in CI because it depends on git identity and `/tmp` state the runner does not provide. A gate that lands on someone else's red pages the founder for someone else's bug. Both tracked as spillover.

### finding-3 — MAJOR, largely mitigated, tracked
`kipi-update.sh` abandons any instance with a dirty tree. Mostly addressed by finding-2's fix (churning `output/` is now a declared skip, content dirs are committed). `q-consult/.q-system` (213 files) and `q-consult/email-watch` (33) remain unclassified and tracked — tracked as `sp-170d4ddd` to verify no instance goes permanently unsyncable after this lands.

### Verified sound
Sentinel comparison (`tuple == str` is always False; no AREA_MAP row can yield a bare string), AREA_MAP ordering (`agent-pipeline/` correctly precedes `.q-system/`), `report_skipped` firing on both the early return and the normal path, worktree handling via `CLAUDE_PROJECT_DIR`, and crash-into-session-exit.

**Revert-sensitivity checked:** restoring `FALLBACK_AREA` fails 5 of the original 9 tests. The tests constrain real behavior.

### Final state
17 tests. Mutants killed: bare commit, suffix matching removed, `INSTANCE_AREAS` widened to `""`, fallback restored, unclassified silently dropped, declared skips misreported, bypass hatch removed, hook commits nothing. CI green.

Posting `kipi/reviewer-approved` on this basis. The reviewer changed; the review did not.

PR #125, opened 2026-08-09
Bake in ASK-363 / ASK-471 / ASK-503 / ASK-504 / ASK-505 / ASK-507

assafkip:
19 commits ahead of main, 14 behind. Opening this rather than merging main
locally: the harness green below was measured on this exact tree, and pulling
14 commits in would change the tree and invalidate that evidence. GitHub owns
the merge so CI re-runs on the merged result.

## What is in here

- **ASK-363** (9 commits) prd-os judgment receipts at approval: fail closed,
  verify the chain before trusting a receipt, compare the whole decision, read
  the ledger under the writer's lock.
- **ASK-471** (4 commits) voicekit voice METHOD package + corrections-ledger
  validation, and the loader scars pinned.
- **ASK-426 follow-through** removed the guidance demanding a CTA at the end.
- **ASK-503** remote-coverage-check `find_repos()` used
  `os.path.isdir(<dir>/.git)`; in a git worktree `.git` is a FILE, so 8 live
  worktrees were false-flagged.
- **ASK-504** the Fable escalation cap no longer pages the founder. `escalate()`
  returns at the cap BEFORE calling Fable, so the page could never carry a
  diagnosis — all 6 of 6 capped ledger rows had `diagnosis: None`. The paired
  test had asserted the page was present, which is why it was never caught.
- **ASK-505** capability-gate timeouts budgeted per test, measured once.
- **ASK-507** token-guard's actor cache had an unlocked read-modify-write.
  Parallel PreToolUse fires raced it and lost the `fable_escalations` counter,
  buying extra billed `claude -p --model claude-fable-5` calls: 172 of 188
  ledger rows attempted a real call against a cap of 2. Fixed with a single
  locked writer.
- `647a87db chore: update system infrastructure` is the repo auto-committer's
  intermediate snapshot of ASK-507 work in progress. Kept rather than rebased
  away — rewriting it would rewrite a real intermediate state for cosmetics.

## Verification

- `kipi check` on `464be47a`, nothing else in flight: **PASS 173, FAIL 0, WARN 1,
  exit 0.**
- ASK-507 reproducer `q-system/.q-system/tests/test_token_guard_cache_race.py`,
  registered in the capability manifest, seen RED (`8 of 8 parallel fires
  reached call_fable against a cap of 2`) then GREEN.

## Known red, pre-existing, NOT from this branch

5 tests in `q-system/.q-system/tests/separation/` fail, and `kipi check` never
runs them — so the green above is silence about those 5, not a pass. Measured
[...]

PR #112, opened 2026-08-06
grep -c ... || echo 0 is not a zero-safe count (ASK-402)

assafkip:
Closes `sp-762c7d6b`. Six sites across three suites.

## The defect

`grep -c` prints the count **and** exits 1 when that count is zero. So `n=$(grep -c PATTERN FILE || echo 0)` appends a second zero and `n` becomes the two-line string `"0\n0"`.

Measured in this shell:

| input | result | |
|---|---|---|
| file with 2 matches | `"2"` | correct |
| file, zero matches | `"0\n0"` | **broken** |
| missing file | `"0"` | correct (grep printed nothing) |

It only bites when the file **exists and is empty**, which is why six sites carried it unnoticed.

## Why it is not cosmetic

`[ "$n" -eq 1 ]` aborts with `integer expression expected`, and `[ "$n" = "0" ]` — a legitimate expect-zero assertion — compares `"0\n0"` against `"0"` and **fails on correct behaviour**.

`test-severity-floor.sh:1996` is exactly that shape:

```sh
[ "$(pages_in "$Q1")" = "0" ]   # "no page was sent"
```

That passes today only because the file happens not to exist yet. The moment it is created and left empty, the suite reports a defect that is not there. A false failure is worse than a missed one — it trains the reader to discount red.

## The fix

`{ grep -c ... || echo 0; } | head -1` keeps whichever zero arrived first, and the fallback still covers the missing-file case where grep prints nothing.

Verified against all three input states through the **real** helper functions (`pages_in`, `worked_count`), not a retyped copy.

## Two checks that exist because the others could pass against a no-op

- **A control** asserting the OLD idiom really does emit a second line in this shell. Without it the whole file proves nothing on a platform where grep behaves differently.
- **A residual scan** over `test/*.sh`, because the broken form is easy to retype from memory. Mutation: restoring it at a real site turns the suite red.

## Verification

`test-linear-worker-parallel` PASS · `test-review-tree-guard` PASS · `test-severity-floor` 198/198 PASS · new suite 8/8. Registered in the capability manifest.

**Noted, not fixed here:** `test-severity-floor` leaves a process alive after its last check prints. That is `sp-063fc932` (`run_bounded` kills the watchdog subshell but not its `sleep` child) — pre-existing and unrelated.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #110, opened 2026-08-05
prd-os delivers what it promises, and a virgin-repo test proves it (ASK-363)

assafkip:
## What this is

Found by **executing** prd-os 0.16.6 in a virgin `git init` repo instead of reading it. Four of five defects were the same shape: **a promise recorded in prose with no code behind it.**

That shape is invisible to every gate this plugin owns. They all read a diff or a spec, and a missing implementation appears in neither. Mutation testing cannot kill a check that does not exist.

## The five

| # | Promise | Reality |
|---|---|---|
| 1 | `/prd-os-init` adds runtime state to `.gitignore` (SKILL.md + README) | Wrote `config.json` and returned. `git check-ignore .claude/state/active-prd.json` returned nothing after a successful init |
| 2 | The spillover ledger "cannot be forgotten" | `gates run` exited 1 `GATE RED` and `archive` exited 0 on the same repo, same moment. Held only by prose in `prd-archive.md` |
| 3 | "Portable core", repo-local surface is `.prd-os/` + `.claude/state/` | `PROPOSAL_DIR_RELPATH` hardcoded to `q-system/output/skeptic-proposals`. A repo containing one README grew a `q-system/` tree on archive |
| 4 | The reported PRD id is the id downstream accepts | `new prd-thing` → `prd-prd-thing-<date>`, then "PRD spec not found" |
| 5 | The skill describes the shipped system | Still said "Scaffold only at 0.1.0", "not yet wired", listed `/prd-revise` (never shipped), taught a disposition enum `findings_writer` rejects |

Founder decisions taken: archive **hard-blocks** with no `--force`; the `.gitignore` write was **built** rather than the claim deleted; #5 re-applied here on 0.16.6 (it had been fixed on a branch 13 commits behind main).

## The missing layer

`tests/test_virgin_repo_lifecycle.py` — `git init`s a repo holding one README, drives the documented lifecycle by subprocess, asserts on exit codes and on-disk effects. Imports **none** of the shared fixtures on purpose: they presume the kipi layout, which is the assumption this file exists to break. Scrubs `CLAUDE_PROJECT_DIR`/`KIPI_HOME`/`QROOT` so paths cannot resolve to the developer's checkout.

Two checks are self-maintaining rather than string greps: documented command list vs `commands/prd-*.md` on disk, documented disposition vocabulary vs `findings_writer.DISPOSITIONS`.

## Verification, in the order it happened

- **Red run: 12 of 14 failed** against 0.16.6 before any fix
[...]

PR #108, opened 2026-08-05
Land the least-code-bias rule that lived only in the runtime (ASK-363)

assafkip:
## Why

This rule was hand-added directly to the marketplace clone and existed **nowhere on main**: live in every session, absent from the repo, one clone refresh from being gone. Recovered and stashed before the refresh on 2026-08-05.

## Decision: land it, not discard it

It is consistent with what the skill already teaches (recon before edit, scope discipline) and with the standing rule that a fix changes exactly what was flagged. It earns its place as rule 7 rather than duplicating an existing one. Numbering verified: it follows rule 6, no collision.

## What stops a fourth recurrence

This is the **third** time someone patched the live runtime directly; the two older drift stashes in that clone are dated 2026-06-21 and 2026-07-01. A stash is not a mechanism. The detector for this shape (marketplace clone carrying uncommitted edits to tracked files) ships with the runtime freshness check in #105, not as a note asking people to stop.

## Notes

- `plugin.json` bumped 0.16.5 -> 0.16.6 for the pre-commit version-bump gate.
- `export-fable-mirror.sh --check` gains a SKILL.md drift line, which is expected and clears on the next mirror export. The two other drift lines (`fable-discipline-lint.py`, its test) reproduce on untouched `origin/main` and are pre-existing, captured separately.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #126, opened 2026-08-09
Guards + voicekit onto main: ASK-471 / ASK-426 / ASK-503 / ASK-504 / ASK-505 / ASK-507

assafkip:
Replaces #125, which is permanently unmergeable. Same work, minus the part
that conflicts.

## Why this PR exists

#125 carried 21 commits and reported `CONFLICTING` with 6 conflicts, all in
`plugins/prd-os/`. Measured rather than inferred:

- All 6 conflicts come from the **8 oldest commits on that branch, every one of
  them ASK-363**.
- Main has since merged the same ASK-363 feature through 5 PRs (#102, #103,
  #108, #110, #112) and already carries `judgment_compiler.py`. So those
  conflicts are two implementations of one feature, not two features.
- The **13 newest commits touch nothing under `plugins/prd-os`** and are
  contiguous. Cherry-picked onto current main: `rc=0`, zero conflicts.

This PR is those 13.

The zero-checks problem on #125 was a **symptom of the same conflicts**, not an
independent blocker: `git ls-remote origin 'refs/pull/125/*'` returns only
`refs/pull/125/head` and no `/merge` ref, so GitHub cannot build the merge
commit that `on: pull_request` needs. This branch is conflict-free, so CI can
run here.

## What is in here

- **ASK-507** token-guard's actor cache had an unlocked read-modify-write.
  Parallel PreToolUse fires raced it and lost the `fable_escalations` counter,
  buying extra billed `claude -p --model claude-fable-5` calls: 172 of 188
  ledger rows attempted a real call against a cap of 2 per actor, 90 landing in
  a second that already carried another. Fixed with a single locked writer; the
  git commit probe is hoisted out of the critical section so a 3s subprocess
  can never push a waiting hook past its 5s timeout.
- **ASK-505** capability-gate timeouts budgeted per test, measured once.
- **ASK-504** the Fable escalation cap no longer pages the founder. `escalate()`
  returned at the cap *before* calling Fable, so the page could never carry a
  diagnosis; all 6 of 6 capped ledger rows had `diagnosis: None`. The paired
  test had asserted the page was present, which is why it was never caught.
- **ASK-503** remote-coverage-check `find_repos()` used
  `os.path.isdir(<dir>/.git)`; in a git worktree `.git` is a FILE, so 8 live
  worktrees were false-flagged.
- **ASK-471** (4 commits) voicekit voice METHOD package + corrections-ledger
  validation, loader scars pinned.
- **ASK-426 follow-through** removed the guidance demanding a CTA at the end.

## Verification

- `kipi check` PASS 173 / FAIL 0 / WARN 1, exit 0, on the pre-cherry-pick tree.
- On **this** tree (main + the 13): 46 passed across the guard and escalation
  suites, including the ASK-507 reproducer.
- ASK-507 reproducer `q-system/.q-system/tests/test_token_guard_cache_race.py`
  is registered in the capability manifest and was seen RED (`8 of 8 parallel
  fires reached call_fable against a cap of 2`) before GREEN.

## Known red, pre-existing, NOT from this PR

5 tests in `q-system/.q-system/tests/separation/` fail, and `kipi check` never
runs them, so its green is silence about them. Measured in clean detached
worktrees: identical `5 failed, 162 passed` at the pre-session baseline
`a474b2f0` and at `464be47a`. Tracked as `sp-ba004add`.

Not carried here: the 8 ASK-363 commits from #125. Whether their 7 follow-up
fixes are already present in main's merged ASK-363 line is an open question
tracked in the ledger; it needs recon of both implementations and must not be
resolved blind.

### DIFF
--- a/.claude/rules/no-orphan-findings.md
+++ b/.claude/rules/no-orphan-findings.md
@@ -31,8 +31,8 @@ There is no third way. You cannot hand-clear the gate.
 
 ## Deterministic backstops
 
-- A `deferred` triage disposition AUTO-creates an open spillover item
-  (findings_writer). Deferring is not a terminal state.
+- A `deferred` disposition AUTO-creates an open spillover item in BOTH findings
+  systems (findings_writer + issue_findings, sp-5bcfbfe8). Never terminal.
 - The fable-discipline lint blocks deferral language written into CODE without a
   capture (`# spillover-skip` acks an already-captured line).
 - `gates run` fails while any item is open (the enforcement of last resort).
--- a/plugins/kipi-core/.claude-plugin/plugin.json
+++ b/plugins/kipi-core/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "kipi-core",
-  "version": "1.5.15",
+  "version": "1.5.16",
   "description": "Core founder OS \u2014 voice, research, AUDHD, LinkedIn/brand, RCA skills + kipi-mcp server",
   "author": {
     "name": "Assaf Kipnis",
--- a/plugins/kipi-core/.claude-plugin/plugin.json
+++ b/plugins/kipi-core/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "kipi-core",
-  "version": "1.5.17",
+  "version": "1.5.18",
   "description": "Core founder OS — voice, research, AUDHD, LinkedIn/brand, RCA skills + kipi-mcp server",
   "author": {
     "name": "Assaf Kipnis",
--- a/plugins/kipi-core/.claude-plugin/plugin.json
+++ b/plugins/kipi-core/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "kipi-core",
-  "version": "1.5.19",
+  "version": "1.5.20",
   "description": "Core founder OS — voice, research, AUDHD, LinkedIn/brand, RCA skills + kipi-mcp server",
   "author": {
     "name": "Assaf Kipnis",
--- a/plugins/kipi-core/.claude-plugin/plugin.json
+++ b/plugins/kipi-core/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "kipi-core",
-  "version": "1.5.18",
+  "version": "1.5.19",
   "description": "Core founder OS — voice, research, AUDHD, LinkedIn/brand, RCA skills + kipi-mcp server",
   "author": {
     "name": "Assaf Kipnis",
--- a/plugins/kipi-core/.claude-plugin/plugin.json
+++ b/plugins/kipi-core/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "kipi-core",
-  "version": "1.5.16",
-  "description": "Core founder OS \u2014 voice, research, AUDHD, LinkedIn/brand, RCA skills + kipi-mcp server",
+  "version": "1.5.17",
+  "description": "Core founder OS — voice, research, AUDHD, LinkedIn/brand, RCA skills + kipi-mcp server",
   "author": {
     "name": "Assaf Kipnis",
     "email": "[REDACTED]"
--- a/plugins/kipi-core/kipi-mcp/src/kipi_mcp/linter.py
+++ b/plugins/kipi-core/kipi-mcp/src/kipi_mcp/linter.py
@@ -487,7 +487,11 @@ class Linter:
 
         question_count = body.count("?")
         if question_count > 1:
-            warnings.append(f"body has {question_count} question marks — consider a single CTA")
+            # NOT "consider a single CTA" (2026-08-06). The founder does not end on a CTA,
+            # and `ending_gate` in the content engine BLOCKS one. A linter nudging the writer
+            # toward the thing another gate rejects costs a slot every time it is obeyed.
+            warnings.append(f"body has {question_count} question marks — pick the one that "
+                            f"lands and cut the rest; never replace them with an ask")
 
         return {
             "pass": len(errors) == 0,
--- a/plugins/kipi-core/skills/founder-voice/SKILL.md
+++ b/plugins/kipi-core/skills/founder-voice/SKILL.md
@@ -44,8 +44,16 @@ If these files are empty/template, the voice skill cannot run. Ask the founder t
 - Questions should make the reader uncomfortable, not curious
 
 ### 7. Ending Pattern
-- Social posts: end with a direct question or sharp statement. Never "Thoughts?" or "Agree?"
-- Articles: end with a reflective question that reframes the whole piece
+- Social posts (LinkedIn, X, Substack): end on a VERDICT. NEVER a question.
+  Founder-directed 2026-08-06: "essentially on any social posts I don't finish with
+  a question." Measured on his corpus that day: 1 of 27 writing samples ends on a
+  question, and that one is a quoted line inside a piece, not a closer. The 4 seed
+  drafts that do are all March 2026 engagement-shaped LinkedIn posts. This REVERSES
+  the old line "end with a direct question or a sharp statement", which was an
+  assertion the corpus contradicted. Enforced by `ending_gate._closing_question_signals`
+  in the content engine, not by this bullet.
+- Long-form articles (Medium) may still end reflectively, but not on a question that
+  asks the reader about their own business.
 - Emails/DMs: end with one clear ask or one specific question
 
 ## Pre-Publish Check (BLOCKING)
--- a/plugins/kipi-core/skills/linkedin-brand/references/summary-frameworks.md
+++ b/plugins/kipi-core/skills/linkedin-brand/references/summary-frameworks.md
@@ -13,7 +13,10 @@ The About section is what ChatGPT, Claude, Gemini, and Perplexity pull when some
 - First word is "I."
 - Keyword-rich but natural. LLMs pull phrases, not just keywords.
 - Zero em-dashes. Zero rule-of-three lists. Zero AI filler.
-- One clear CTA at the end. Not three.
+- NO CTA at the end. Founder-directed 2026-08-06: "most critically - I dont want to
+  end with a CTA." Per the operator-vs-marketer rule the `ending_gate` enforces, an
+  operator ends "in nothing, or ONE SPECIFIC THING". A marketer "ends in a CTA to a
+  course/call". One specific thing is allowed. An ask is not.
 - Scar-anchored. Lead with a real moment, not a title.
 - Anti-misclassification line near the end (tells LLMs what you're NOT).
 
@@ -37,8 +40,9 @@ Concrete, not "disrupting" or "reinventing."
 [ANTI-MISCLASSIFICATION — 1 line]
 What you are NOT. Stops LLMs from slotting you into the wrong category.
 
-[CTA — 1 line]
-One specific ask. DM about X, book a call if Y, subscribe for Z.
+[CLOSE — 1 line]
+End on nothing, or on ONE SPECIFIC THING. Never an ask. No "DM me", no "book a
+call", no "subscribe".
 ```
 
 ## Framework 2: Passion / Execution
@@ -59,7 +63,7 @@ Who this brand is for. Helps humans and AI classify you correctly.
 
 [ANTI-MISCLASSIFICATION — 1 line]
 
-[CTA — 1 line]
+[CLOSE — 1 line]  (one specific thing, or nothing. Never an ask.)
 ```
 
 ## Framework 3: One-line hammer (for X/Twitter bio + GitHub)
@@ -103,7 +107,7 @@ Sub-patterns that work:
 3. At least one specific scar with company + what broke?
 4. Zero em-dashes, zero rule-of-three lists?
 5. Zero AI filler phrases?
-6. One CTA, not three?
+6. Zero CTAs? (ends on nothing, or one specific thing, never an ask)
 7. Anti-misclassification line present?
 8. 3+ keyword targets worked in naturally?
 9. Readable aloud without sounding like a press release?
--- a/plugins/kipi-core/voicekit/__init__.py
+++ b/plugins/kipi-core/voicekit/__init__.py
@@ -1,0 +1,13 @@
+"""voicekit: the METHOD half of the voice architecture. Fleet code, founder-agnostic.
+
+The split is deliberate and a test holds it (2026-08-06, voice-architecture PRD):
+
+- This package is MACHINERY: fingerprint math, corpus loading, deterministic exemplar
+  selection, prompt assembly, validators. It ships fleet-wide with kipi-core.
+- A founder's DATA -- exemplars, lexicon values, computed bands, corrections -- lives in
+  that founder's own instance repo (its `voice/` dir) and must never appear in this tree.
+  `tests/test_no_founder_data.py` fails this package if it ever quotes a corpus.
+
+Nothing here reads a network, calls a model, or knows a path outside the `voice_dir` it
+is handed. Pure functions of (files, counters), so every consumer can test offline.
+"""
--- a/plugins/kipi-core/voicekit/assemble.py
+++ b/plugins/kipi-core/voicekit/assemble.py
@@ -1,0 +1,80 @@
+#!/usr/bin/env python3
+"""Assemble the voice section of a generation prompt from a loaded Voice.
+
+Order is the design (voice-architecture PRD, 2026-08-06), and each position is a
+decision:
+
+    identity -> pov -> lexicon positives -> EXEMPLARS -> active corrections
+
+- Exemplars sit LAST-but-one, adjacent to the source material the caller appends:
+  the strong context position ("lost in the middle" is why 19k of samples inside a
+  57k prompt taught nothing).
+- Corrections load LAST: recency weight, the coded replacement for gotchas.md's
+  "loaded last, overrides" convention.
+- The lexicon contributes ONLY its positive slices (prefer/voiceprint_terms). The
+  banned lists live in gate scripts; restating them in the prompt is negative
+  priming, and `test_prompt_gate_coherence` in the consuming repo asserts absence.
+
+Returns (text, provenance). Provenance carries exemplar ids, correction ids, the
+selection reason and the exemplar BODIES -- the bodies feed the echo gate, which
+must compare the final text against exactly what the prompt showed the model.
+"""
+from __future__ import annotations
+
+from . import selector
+
+BUDGET_CHARS = 20000     # asserted by validate.feasible + the consumer's suite,
+                         # NEVER enforced by a runtime raise or slice (ASK-461:
+                         # a cap you cannot see is the same bug; a cap that kills
+                         # the daily job is a worse one).
+
+
+def _lexicon_positive(lexicon):
+    lines = []
+    prefer = lexicon.get("prefer") or []
+    if prefer:
+        pairs = ", ".join(f"{p.get('use')} (not {p.get('not')})"
+                          for p in prefer if p.get("use"))
+        if pairs:
+            lines.append(f"Words he reaches for: {pairs}.")
+    terms = lexicon.get("voiceprint_terms") or []
+    if terms:
+        lines.append("His recurring vocabulary: " + ", ".join(terms) + ".")
+    return "\n".join(lines)
+
+
+def voice_section(voice, channel, counter, slot_index=0, k=selector.DEFAULT_K,
+                  slot_kind="post"):
+    """(text, provenance) for one slot. Pure; empty Voice -> ('', empty provenance)."""
+    picked = selector.select(voice.active_exemplars(), channel, counter,
+                             slot_index=slot_index, k=k, slot_kind=slot_kind)
+    corrections = voice.active_corrections()
+
+    parts = []
+    if voice.identity.strip():
+        parts.append("WHO IS WRITING:\n" + voice.identity.strip())
+    if voice.pov.strip():
+        parts.append("WHAT HE WRITES ABOUT AND BELIEVES:\n" + voice.pov.strip())
+    lex = _lexicon_positive(voice.lexicon)
+    if lex:
+        parts.append(lex)
+    if picked:
+        bodies = "\n\n---\n\n".join((r.get("text") or "").strip() for r in picked)
+        parts.append("POSTS HE HAS WRITTEN. Match their rhythm, register and "
+                     "length. Never reuse their sentences or openings:\n\n" + bodies)
+    if corrections:
+        lines = "\n".join(f"- {r['instruction']}" for r in corrections
+                          if not r.get("scope") or channel in r["scope"])
+        if lines:
+            parts.append("STANDING CORRECTIONS (these override everything above):\n"
+                         + lines)
+
+    provenance = {
+        "exemplar_ids": [str(r.get("id")) for r in picked],
+        "exemplar_texts": [(r.get("text") or "") for r in picked],
+        "correction_ids": [str(r.get("id")) for r in corrections],
+        "selection_reason": selector.selection_reason(picked, channel, counter,
+                                                      slot_index),
+        "skipped_rows": voice.skipped_rows,
+    }
+    return "\n\n".join(parts), provenance
--- a/plugins/kipi-core/voicekit/corpus.py
+++ b/plugins/kipi-core/voicekit/corpus.py
@@ -1,0 +1,101 @@
+#!/usr/bin/env python3
+"""Read a founder's voice/ directory. Degrade, never die.
+
+The daily job that consumes this runs unattended in launchd. The contract, paid
+for by ASK-461 and its cousins: a missing file is LESS GUIDANCE, a corrupt JSONL
+row is SKIPPED AND COUNTED, and nothing on this read path raises. The counts ride
+back to the caller so a deadman check can surface decay instead of a crash hiding
+it. Loud failure belongs to the VALIDATOR (validate.py), which runs in a pytest
+suite at edit time, not inside the publishing run.
+"""
+from __future__ import annotations
+
+import json
+import os
+
+EXEMPLARS = "exemplars.jsonl"
+CORRECTIONS = "corrections.jsonl"
+IDENTITY = "identity.md"
+POV = "pov.md"
+LEXICON = "lexicon.json"
+FINGERPRINT = "fingerprint.json"
+
+EXEMPLAR_KINDS = ("post", "article-excerpt", "comment", "dm", "email")
+EXEMPLAR_CHANNELS = ("linkedin", "x", "substack", "medium", "any")
+
+
+def read_jsonl(path):
+    """(rows, skipped_count). Missing file = ([], 0): absence is a fact, not an error."""
+    if not os.path.exists(path):
+        return [], 0
+    rows, skipped = [], 0
+    try:
+        with open(path, encoding="utf-8", errors="replace") as handle:
+            for line in handle:
+                line = line.strip()
+                if not line:
+                    continue
+                try:
+                    row = json.loads(line)
+                except ValueError:
+                    skipped += 1       # a torn row loses ITS row, never the file
+                    continue
+                if isinstance(row, dict):
+                    rows.append(row)
+                else:
+                    skipped += 1
+    except OSError:
+        return [], 0
+    return rows, skipped
+
+
+def read_text(path):
+    """File content, or '' when missing/unreadable. Same errors= as the old loader:
+    a plugin update writing while we read must not kill the run on one bad byte."""
+    try:
+        with open(path, encoding="utf-8", errors="replace") as handle:
+            return handle.read()
+    except OSError:
+        return ""
+
+
+def read_json(path):
+    """Parsed JSON dict, or None. None means the consumer runs without this input."""
+    try:
+        with open(path, encoding="utf-8") as handle:
+            loaded = json.load(handle)
+        return loaded if isinstance(loaded, dict) else None
+    except (OSError, ValueError):
+        return None
+
+
+class Voice:
+    """Everything one voice/ dir holds, read once, with decay counts."""
+
+    def __init__(self, voice_dir):
+        self.voice_dir = voice_dir
+        self.exemplars, ex_skipped = read_jsonl(os.path.join(voice_dir, EXEMPLARS))
+        self.corrections, co_skipped = read_jsonl(os.path.join(voice_dir, CORRECTIONS))
+        self.identity = read_text(os.path.join(voice_dir, IDENTITY))
+        self.pov = read_text(os.path.join(voice_dir, POV))
+        self.lexicon = read_json(os.path.join(voice_dir, LEXICON)) or {}
+        self.fingerprint = read_json(os.path.join(voice_dir, FINGERPRINT))
+        self.skipped_rows = ex_skipped + co_skipped
+
+    def active_exemplars(self):
+        """Usable rows only: active status, non-empty text, weight above zero."""
+        return [r for r in self.exemplars
+                if r.get("status", "active") == "active"
+                and (r.get("text") or "").strip()
+                and float(r.get("weight", 1.0) or 0) > 0]
+
+    def active_corrections(self):
+        """Rows still carried as prose. A `promoted` row's gate carries it instead."""
+        return [r for r in self.corrections
+                if r.get("status") == "active"
+                and (r.get("instruction") or "").strip()]
+
+
+def load(voice_dir):
+    """The one entry point. Never raises on a readable-or-absent tree."""
+    return Voice(voice_dir)
--- a/plugins/kipi-core/voicekit/echo.py
+++ b/plugins/kipi-core/voicekit/echo.py
@@ -1,0 +1,78 @@
+#!/usr/bin/env python3
+"""Echo detection: the candidate must not parrot what its own prompt showed it.
+
+why this exists (2026-08-06, Stage 0 finding 2): 9 of 22 posts published in one day
+contained the literal sentence pair "Nothing was sent. Nothing was lost." -- the
+EXAMPLE ending inside `revise.RULE_GUIDANCE`. The reviser copied its own guidance
+example into published text, and nothing compared output against what the prompt
+carried. Loading exemplars into prompts (the whole voice architecture) makes this
+class WORSE unless it is gated, so the gate ships with the loader.
+
+Two checks, both pure text ops:
+
+- `prompt_echo`: any N-gram (default 8 words) shared between the candidate and any
+  text that rode in its prompt (exemplars, guidance examples). 8 is deliberate: long
+  enough that a shared idiom ("the same thing happened to us") cannot fire, short
+  enough that a lifted sentence cannot hide.
+- `opener_echo`: the candidate's first sentence shape vs recent published openers.
+  First-6-word overlap of 4+ tokens = the same opener. 7/22 posts opening "Three
+  receipts/drafts/duplicate..." is the measured failure.
+"""
+from __future__ import annotations
+
+import re
+
+NGRAM = 8
+OPENER_WORDS = 6
+OPENER_SHARED = 4
+
+_WORD = re.compile(r"[A-Za-z0-9']+")
+
+
+def _words(text):
+    return [w.lower() for w in _WORD.findall(text or "")]
+
+
+def ngrams(text, n=NGRAM):
+    words = _words(text)
+    return {tuple(words[i:i + n]) for i in range(len(words) - n + 1)}
+
+
+def prompt_echo(candidate, prompt_texts, n=NGRAM):
+    """Echoed n-grams between the candidate and the texts its prompt carried.
+
+    Returns a sorted list of the echoed word-tuples joined as strings; [] is a pass.
+    `prompt_texts` is the exemplar/guidance bodies only, never the instruction
+    boilerplate -- instructions repeat by design, and flagging them would make the
+    gate cry wolf until someone switches it off.
+    """
+    got = ngrams(candidate, n)
+    if not got:
+        return []
+    hits = set()
+    for source in prompt_texts or []:
+        hits |= got & ngrams(source, n)
+    return sorted(" ".join(gram) for gram in hits)
+
+
+def opener(text, k=OPENER_WORDS):
+    return _words(text)[:k]
+
+
+def opener_echo(candidate, recent_openers, k=OPENER_WORDS, shared=OPENER_SHARED):
+    """The recent opener this candidate repeats, or None.
+
+    `recent_openers` is a list of the last few published posts' opening words
+    (the caller reads them from the published-body archive). Position-wise
+    comparison: the same 4 of the first 6 words IN ORDER is the same opener shape.
+    """
+    mine = opener(candidate, k)
+    if not mine:
+        return None
+    for prev in recent_openers or []:
+        prev_words = [w.lower() for w in prev][:k] if isinstance(prev, (list, tuple)) \
+            else opener(str(prev), k)
+        same = sum(1 for a, b in zip(mine, prev_words) if a == b)
+        if same >= shared:
+            return " ".join(prev_words)
+    return None
--- a/plugins/kipi-core/voicekit/fingerprint.py
+++ b/plugins/kipi-core/voicekit/fingerprint.py
@@ -202,9 +202,14 @@ def _bounds(band):
     min/max are corpus observations, so every corpus member is inside its own
     bounds, which is exactly the property the control-set test asserts.
     """
-    lo = band["p10"] if band.get("min") is None else min(band["p10"], band["min"])
-    hi = band["p90"] if band.get("max") is None else max(band["p90"], band["max"])
-    return lo, hi
+    # Degrade-not-die on a partial band dict (voice-1 review, minor): a
+    # hand-migrated band can hold min without p10; min(None, x) raises. Every
+    # present bound participates; an absent one simply does not constrain.
+    lows = [v for v in (band.get("p10"), band.get("min")) if v is not None]
+    highs = [v for v in (band.get("p90"), band.get("max")) if v is not None]
+    if not lows or not highs:
+        return None, None
+    return min(lows), max(highs)
 
 
 def score(text, fingerprint):
@@ -222,6 +227,8 @@ def score(text, fingerprint):
         if value is None or band.get("p10") is None:
             continue
         lo, hi = _bounds(band)
+        if lo is None:
+            continue
         out[name] = {"value": value, "band": [band["p10"], band["p90"]],
                      "bounds": [lo, hi], "inside": lo <= value <= hi}
     return out
@@ -232,8 +239,13 @@ def out_of_band(text, fingerprint, tier="blocking"):
 
     Blocking entries carry a direction (above|below|both): a cap only fails
     ABOVE its widened bounds, a floor only BELOW. Legacy plain-string entries
-    mean both. Advisory tier entries are plain metric names (report-only, so
-    direction lives with the reader, not the gate).
+    mean both.
+
+    This function is GATE semantics. The advisory tier is report-only: read it
+    through `score()`, never through here -- passing tier="advisory" would
+    coerce report-only names into both-direction rejections (voice-1 review,
+    minor). The tier parameter exists for a future PROMOTED tier, not for
+    advisory reads.
     """
     got = metrics(text)
     bands = fingerprint.get("metrics") or {}
@@ -246,6 +258,8 @@ def out_of_band(text, fingerprint, tier="blocking"):
         if band is None or value is None or band.get("p10") is None:
             continue
         lo, hi = _bounds(band)
+        if lo is None:
+            continue
         too_high = direction in ("above", "both") and value > hi
         too_low = direction in ("below", "both") and value < lo
         if too_high or too_low:
--- a/plugins/kipi-core/voicekit/fingerprint.py
+++ b/plugins/kipi-core/voicekit/fingerprint.py
@@ -1,0 +1,260 @@
+#!/usr/bin/env python3
+"""Style fingerprint: measurable voice, computed from a corpus, scored on a candidate.
+
+why this exists (2026-08-06): a 55k-char prose style guide shipped in every generation
+prompt and the founder still read the output as "missing my voice as a grumpy engineer".
+Prose describing a distribution is a region; only numbers computed FROM the writing can
+say whether a candidate sits inside it. This module is the instrument: `compute()` builds
+percentile bands from real exemplars, `score()` reports where a candidate falls.
+
+Everything is pure text ops -- no model, no network, no clock. The metric set is the
+deterministic slice of the voice only. Whether a scar is true, whether a coined name is
+good, whether the humor lands: none of that is here, and any claim that a script scores
+"voice authenticity" would be prompt-only enforcement wearing a script's name.
+
+Band semantics: p10/p90 over the corpus, per metric. A candidate outside a band is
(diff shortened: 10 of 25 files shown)


assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/pr-126-20260808-175257.md` (301807 bytes).

```
FINDINGS:
minor|Channel-scoped corrections excluded from the prompt are still recorded as applied in provenance|plugins/kipi-core/voicekit/assemble.py:75
minor|First-person measurement changes when “we”, “my”, “our”, or “me” is capitalized|plugins/kipi-core/voicekit/fingerprint.py:44
minor|A syntactically valid exemplar with a nonnumeric weight crashes the degrade-without-dying runtime path|plugins/kipi-core/voicekit/corpus.py:90
END FINDINGS
```

--- reviewer output, last 54538 bytes of 301807 (full review at the path above) ---


    # Everything from here to process exit runs under this actor's lock, so the
    # load-mutate-save below is one critical section rather than a race. The
    # existing save_cache calls at each block/warn exit are unchanged; they now
    # simply happen while the lock is held.
    hold_actor_lock(actor)

    # Load cache, update counters from this invocation. The commit-progress
    # valve runs before the checks so a freshly-shipped commit lifts the
    # ceiling even when the PostToolUse event never arrived (wiring B).
    cache = load_cache(actor)
    cache = update_counters(tool_name, tool_input, cache)
    cache = reset_volume_if_committed(cache, head_epoch)

    # A triage from an earlier stuck refusal, if the detached call has landed.
    # One cheap open() on the common path (the file is absent almost always).
    # Consumed HERE rather than only on the next refusal: the point is to reach
    # the agent while it is still deciding, and waiting for another block would
    # deliver the advice after the decision it was meant to inform.
    pending_triage = take_pending_triage(actor)

    # --- Run checks in priority order ---

    # 1. Sensitive file blocking (highest priority)
    msg = check_sensitive_file(tool_name, tool_input)
    if msg:
        cache = uncount_blocked_attempt(cache, tool_name, tool_input)
        save_cache(actor, cache)
        block(msg, triage=pending_triage)

    # 2. Exact retry detection
    msg = check_exact_retry(tool_name, tool_input, cache)
    if msg:
        cache = uncount_blocked_attempt(cache, tool_name, tool_input)
        save_cache(actor, cache)
        block(msg, cache=cache, hook_input=hook_input,
              trigger="exact-retry", actor=actor, triage=pending_triage)

    # 3. Volume ceiling/warning — EXEMPT a git commit. A commit is the checkpoint the
    # ceiling is asking for; blocking it deadlocks the run (the PostToolUse commit-reset
    # can't fire if the commit never runs). Sensitive-file + exact-retry checks above still
    # apply to commits; only the volume ceiling is skipped.
    #
    # A volume WARNING is held, not emitted here. warn() exits 0, so emitting it
    # inline made checks 4-11 unreachable — for the whole grace budget, an agent
    # spiralling on the gate file got GATE_GRACE unchecked edit attempts and
    # never heard "your edit approach is wrong", then got a block that blamed the
    # gate (PR #27 review, finding 3). Held warnings are emitted at the end, or
    # carried into a later warning; a later BLOCK outranks them and wins.
    pending_warning = None

    def emit_warning(message):
        if pending_warning and message != pending_warning:
            message = pending_warning + "\n\n" + message
        if pending_triage and (FABLE_BANNER % pending_triage) != message:
            message = (FABLE_BANNER % pending_triage) + "\n\n" + message
        warn(message)

    if not _is_commit_command(tool_name, tool_input):
        result = check_volume(cache)
        if result:
            level, msg = result
            if level == "block":
                cache = uncount_blocked_attempt(cache, tool_name, tool_input)
                save_cache(actor, cache)
                block(msg, cache=cache, hook_input=hook_input,
                      trigger="volume-ceiling", actor=actor,
                      triage=pending_triage)
            pending_warning = msg

    # 4. Subagent ceiling
    msg = check_agent_ceiling(tool_name, cache)
    if msg:
        cache = uncount_blocked_attempt(cache, tool_name, tool_input)
        save_cache(actor, cache)
        block(msg, triage=pending_triage)

    # 5. MCP rate limit
    msg = check_mcp_rate(tool_name, cache)
    if msg:
        cache = uncount_blocked_attempt(cache, tool_name, tool_input)
        save_cache(actor, cache)
        block(msg, triage=pending_triage)

    # 6. Read spiral warning
    msg = check_read_spiral(tool_name, cache)
    if msg:
        save_cache(actor, cache)
        emit_warning(msg)

    # 7. File re-read warning
    msg = check_file_reread(tool_name, tool_input, cache)
    if msg:
        save_cache(actor, cache)
        emit_warning(msg)

    # 8. Grep drift warning
    msg = check_grep_drift(tool_name, cache)
    if msg:
        save_cache(actor, cache)
        emit_warning(msg)

    # 9. Edit spiral block
    msg = check_edit_spiral(tool_name, tool_input, cache)
    if msg:
        cache = uncount_blocked_attempt(cache, tool_name, tool_input)
        save_cache(actor, cache)
        block(msg, cache=cache, hook_input=hook_input,
              trigger="edit-spiral", actor=actor, triage=pending_triage)

    # 10. Agent no-output warning
    msg = check_agent_no_output(tool_name, cache)
    if msg:
        save_cache(actor, cache)
        emit_warning(msg)

    # 11. Time stall warning
    msg = check_time_stall(cache)
    if msg:
        save_cache(actor, cache)
        emit_warning(msg)

    # All clear. A triage that landed since the last call is delivered here —
    # the ordinary case, because the agent is usually working when it arrives.
    save_cache(actor, cache)
    if pending_triage:
        emit_warning(FABLE_BANNER % pending_triage)
    if pending_warning:
        warn(pending_warning)
    sys.exit(0)


if __name__ == "__main__":
    main()

exec
/bin/zsh -lc "rg -n \"CACHE_TTL_DAYS|actor_key|def get_actor|def actor\" q-system/.q-system/token-guard.py | head -40
nl -ba q-system/.q-system/token-guard.py | sed -n '120,230p'
nl -ba plugins/kipi-core/voicekit/corpus.py | sed -n '55,100p'
nl -ba plugins/kipi-core/voicekit/validate.py | sed -n '15,95p'" in /Users/assafkipnis/.config/kipi/review-trees/pr-126
 succeeded in 0ms:
106:CACHE_TTL_DAYS = 7          # Stale guard caches in /tmp older than this are swept
109:def actor_cache_key(hook_input):
121:def cache_path(actor_key):
122:    return f"/tmp/claude-guard-{actor_key}.json"
125:def sweep_stale_caches(max_age_days=CACHE_TTL_DAYS):
144:def load_cache(actor_key):
145:    path = cache_path(actor_key)
153:        "actor_key": actor_key,
173:def save_cache(actor_key, cache):
174:    path = cache_path(actor_key)
227:def lock_path(actor_key):
243:    return f"/tmp/claude-guard-{actor_key}.lock"
266:def hold_actor_lock(actor_key):
279:        handle = open(lock_path(actor_key), "a+")
287:def locked_cache(actor_key):
306:        handle = open(lock_path(actor_key), "a+")
310:    cache = load_cache(actor_key)
314:        save_cache(actor_key, cache)
   120	
   121	def cache_path(actor_key):
   122	    return f"/tmp/claude-guard-{actor_key}.json"
   123	
   124	
   125	def sweep_stale_caches(max_age_days=CACHE_TTL_DAYS):
   126	    """Delete guard caches older than max_age_days. Runs on UserPromptSubmit
   127	    only (once per user message), so the per-tool-call path stays free of
   128	    directory scans. Subagent caches have no UserPromptSubmit of their own;
   129	    this sweep is their cleanup path."""
   130	    import glob
   131	    cutoff = time.time() - max_age_days * 86400
   132	    # Both patterns: the sidecar lock files (sp-016776e6) do not end in .json,
   133	    # so the original glob would have left one per actor in /tmp forever.
   134	    stale = glob.glob("/tmp/claude-guard-*.json") + glob.glob(
   135	        "/tmp/claude-guard-*.lock")
   136	    for path in stale:
   137	        try:
   138	            if os.path.getmtime(path) < cutoff:
   139	                os.remove(path)
   140	        except OSError:
   141	            pass
   142	
   143	
   144	def load_cache(actor_key):
   145	    path = cache_path(actor_key)
   146	    if os.path.exists(path):
   147	        try:
   148	            with open(path) as f:
   149	                return json.load(f)
   150	        except (json.JSONDecodeError, IOError):
   151	            pass
   152	    return {
   153	        "actor_key": actor_key,
   154	        "tool_calls_since_user": 0,
   155	        "agent_calls_since_user": 0,
   156	        "mcp_timestamps": [],
   157	        "repeat_map": {},
   158	        "consecutive_reads": 0,
   159	        "warnings_issued": 0,
   160	        "file_read_counts": {},
   161	        "greps_since_write": 0,
   162	        "edit_targets": {},
   163	        "agents_without_write": 0,
   164	        "last_write_time": time.time(),
   165	        "calls_since_write": 0,
   166	        "last_volume_reset": time.time(),
   167	        "gate_grace_remaining": 0,
   168	        "gate_grace_gate": None,
   169	        "gate_grace_grants": 0,
   170	    }
   171	
   172	
   173	def save_cache(actor_key, cache):
   174	    path = cache_path(actor_key)
   175	    try:
   176	        with open(path, "w") as f:
   177	            json.dump(cache, f)
   178	    except IOError:
   179	        pass
   180	
   181	
   182	# --- The single-writer chokepoint (sp-016776e6) ------------------------------
   183	# load_cache/save_cache are a read-modify-write, and Claude Code fires this hook
   184	# once per tool call — so parallel tool calls mean parallel guard processes on
   185	# ONE actor's cache file. Unlocked, each read the same pre-state and the last
   186	# writer won; every other process's mutation was silently dropped.
   187	#
   188	# The lost mutation that cost real money is `fable_escalations`, the counter
   189	# request_escalation reads to decide whether this actor has spent its
   190	# KIPI_FABLE_CAP budget of billed `claude -p --model claude-fable-5` calls. A
   191	# dropped increment let the next process read a stale count, believe it was
   192	# under the cap, and buy another one.
   193	#
   194	# MEASURED over the shipped ledger before the fix (188 rows,
   195	# q-system/output/fable-escalations/, 2026-08-08): 172 rows attempted a real
   196	# call against a cap of 2 per actor. 90 landed in a second that already carried
   197	# another real call, 7 in the worst single second, and 8 distinct seconds carry
   198	# two DIFFERENT triggers — two concurrent guard processes is the only thing that
   199	# produces that. The two colliding CAPPED pairs are 2026-08-04T03:35:31Z and
   200	# 2026-08-08T22:07:13Z, each one `exact-retry` racing one `volume-ceiling`.
   201	#
   202	# WHY A LOCK AND NOT AN ATOMIC WRITE. tmpfile + os.replace was the tempting fix
   203	# and it does not work: it makes the WRITE atomic, not the read-modify-write.
   204	# Both processes still read the same pre-state, so the update is still lost —
   205	# just lost cleanly. The read and the write have to be inside one critical
   206	# section, which is what this is.
   207	#
   208	# LOST UPDATES WERE NOT THE WORST OF IT. save_cache truncates and rewrites in
   209	# place, so two concurrent writers can interleave: running the concurrency suite
   210	# against a no-lock mutant produced a cache whose JSON ended at byte 464 with a
   211	# longer writer's tail still after it. load_cache catches JSONDecodeError and
   212	# returns fresh defaults, so a torn write silently reset every counter this
   213	# guard owns — the ceiling, the retry map and the escalation budget all back to
   214	# zero, with nothing logged and nothing to see afterwards.
   215	#
   216	# WHY THE REGION HOLDS NO SUBPROCESS. This hook is wired at `timeout: 5` on all
   217	# three events fleet-wide, and a hook that overruns its timeout has its exit 2
   218	# DISCARDED (measured 2026-08-03; see the escalation block comment below). So a
   219	# lock held across `_head_commit_epoch`'s git call (capped at 3s) could turn
   220	# contention into a SPENT refusal — trading this defect for a worse one. The git
   221	# probe is therefore hoisted out of the region by main(), leaving only in-memory
   222	# work under the lock: microseconds, not milliseconds.
   223	LOCK_WAIT_SECONDS = 2.0     # unreachable in practice; the region is pure CPU
   224	LOCK_POLL_SECONDS = 0.002
   225	
   226	
   227	def lock_path(actor_key):
   228	    """A sidecar, never the cache file itself.
   229	
   230	    HONEST ABOUT WHY, because the first version of this comment was wrong.
    55	    try:
    56	        with open(path, encoding="utf-8", errors="replace") as handle:
    57	            return handle.read()
    58	    except OSError:
    59	        return ""
    60	
    61	
    62	def read_json(path):
    63	    """Parsed JSON dict, or None. None means the consumer runs without this input."""
    64	    try:
    65	        with open(path, encoding="utf-8") as handle:
    66	            loaded = json.load(handle)
    67	        return loaded if isinstance(loaded, dict) else None
    68	    except (OSError, ValueError):
    69	        return None
    70	
    71	
    72	class Voice:
    73	    """Everything one voice/ dir holds, read once, with decay counts."""
    74	
    75	    def __init__(self, voice_dir):
    76	        self.voice_dir = voice_dir
    77	        self.exemplars, ex_skipped = read_jsonl(os.path.join(voice_dir, EXEMPLARS))
    78	        self.corrections, co_skipped = read_jsonl(os.path.join(voice_dir, CORRECTIONS))
    79	        self.identity = read_text(os.path.join(voice_dir, IDENTITY))
    80	        self.pov = read_text(os.path.join(voice_dir, POV))
    81	        self.lexicon = read_json(os.path.join(voice_dir, LEXICON)) or {}
    82	        self.fingerprint = read_json(os.path.join(voice_dir, FINGERPRINT))
    83	        self.skipped_rows = ex_skipped + co_skipped
    84	
    85	    def active_exemplars(self):
    86	        """Usable rows only: active status, non-empty text, weight above zero."""
    87	        return [r for r in self.exemplars
    88	                if r.get("status", "active") == "active"
    89	                and (r.get("text") or "").strip()
    90	                and float(r.get("weight", 1.0) or 0) > 0]
    91	
    92	    def active_corrections(self):
    93	        """Rows still carried as prose. A `promoted` row's gate carries it instead."""
    94	        return [r for r in self.corrections
    95	                if r.get("status") == "active"
    96	                and (r.get("instruction") or "").strip()]
    97	
    98	
    99	def load(voice_dir):
   100	    """The one entry point. Never raises on a readable-or-absent tree."""
    15	import re
    16	
    17	from . import assemble, corpus, fingerprint
    18	
    19	MIN_ROWS_PER_POOL = 3      # below this, selection silently narrows to repetition;
    20	                           # the fix is curation, so the message says so.
    21	
    22	
    23	def check_exemplars(path):
    24	    problems = []
    25	    if not os.path.exists(path):
    26	        return [f"{path}: missing"]
    27	    seen_ids = set()
    28	    with open(path, encoding="utf-8") as handle:
    29	        for lineno, line in enumerate(handle, 1):
    30	            line = line.strip()
    31	            if not line:
    32	                continue
    33	            try:
    34	                row = json.loads(line)
    35	            except ValueError as exc:
    36	                problems.append(f"line {lineno}: unparseable JSON ({exc})")
    37	                continue
    38	            rid = row.get("id")
    39	            if not rid:
    40	                problems.append(f"line {lineno}: no id")
    41	            elif rid in seen_ids:
    42	                problems.append(f"line {lineno}: duplicate id {rid!r}")
    43	            seen_ids.add(rid)
    44	            if row.get("kind") not in corpus.EXEMPLAR_KINDS:
    45	                problems.append(f"{rid}: kind {row.get('kind')!r} not in "
    46	                                f"{corpus.EXEMPLAR_KINDS}")
    47	            if row.get("channel") not in corpus.EXEMPLAR_CHANNELS:
    48	                problems.append(f"{rid}: channel {row.get('channel')!r} not in "
    49	                                f"{corpus.EXEMPLAR_CHANNELS}")
    50	            text = row.get("text") or ""
    51	            if not text.strip():
    52	                problems.append(f"{rid}: empty text")
    53	            if "—" in text:
    54	                problems.append(f"{rid}: emdash in text (the one character the "
    55	                                f"founder bans everywhere)")
    56	            if row.get("status", "active") not in ("active", "retired"):
    57	                problems.append(f"{rid}: status {row.get('status')!r}")
    58	            # voice-2 review: two seed rows shipped literal {{UNVALIDATED}}
    59	            # markers into the corpus -- template scaffolding presented as voice
    60	            # material. Any mustache placeholder in an exemplar is scaffolding.
    61	            if "{{" in text:
    62	                problems.append(f"{rid}: template placeholder in text")
    63	            # And several closed with campaign-hashtag tails, the exact register
    64	            # identity prose disavows. A trailing hashtag-only line is metadata,
    65	            # not voice.
    66	            last = text.strip().splitlines()[-1] if text.strip() else ""
    67	            if re.fullmatch(r"(#\w+[ \t]*)+", last):
    68	                problems.append(f"{rid}: trailing hashtag line in text")
    69	    return problems
    70	
    71	
    72	CORRECTION_CLASSES = ("deterministic", "interpretive")
    73	CORRECTION_STATUSES = ("active", "promoted", "retired")
    74	
    75	
    76	def check_corrections(path):
    77	    """corrections.jsonl schema health (voice-2 review: the ledger had no
    78	    validator at all, so a malformed or duplicate row passed the gate green)."""
    79	    if not os.path.exists(path):
    80	        return []                      # an absent ledger is a valid empty ledger
    81	    problems = []
    82	    seen = set()
    83	    with open(path, encoding="utf-8") as handle:
    84	        for lineno, line in enumerate(handle, 1):
    85	            line = line.strip()
    86	            if not line:
    87	                continue
    88	            try:
    89	                row = json.loads(line)
    90	            except ValueError as exc:
    91	                problems.append(f"corrections line {lineno}: unparseable ({exc})")
    92	                continue
    93	            rid = row.get("id")
    94	            if not rid:
    95	                problems.append(f"corrections line {lineno}: no id")

exec
/bin/zsh -lc "nl -ba q-system/.q-system/token-guard.py | sed -n '95,125p'
sed -n '1,360p' q-system/.q-system/tests/test_token_guard_cache_race.py" in /Users/assafkipnis/.config/kipi/review-trees/pr-126
 succeeded in 0ms:
    95	
    96	
    97	def is_readonly_observation(tool_name, tool_input):
    98	    if tool_name in OBSERVATION_TOOLS:
    99	        return True
   100	    actions = OBSERVATION_ACTIONS.get(tool_name)
   101	    if actions:
   102	        return (tool_input or {}).get("action") in actions
   103	    return False
   104	
   105	
   106	CACHE_TTL_DAYS = 7          # Stale guard caches in /tmp older than this are swept
   107	
   108	
   109	def actor_cache_key(hook_input):
   110	    """Per-actor cache key. Subagents share the parent's session_id but carry
   111	    their own agent_id (evidence: q-system/output/guard-payload-dump.json) —
   112	    compounding the two gives each actor its own budget. No agent_id = the
   113	    main session actor."""
   114	    session_id = hook_input.get("session_id", "unknown")
   115	    agent_id = hook_input.get("agent_id")
   116	    if agent_id:
   117	        return f"{session_id}-agent-{agent_id}"
   118	    return session_id
   119	
   120	
   121	def cache_path(actor_key):
   122	    return f"/tmp/claude-guard-{actor_key}.json"
   123	
   124	
   125	def sweep_stale_caches(max_age_days=CACHE_TTL_DAYS):
"""The guard cache must survive parallel hook fires (sp-016776e6).

WHAT THIS PINS, AND WHY IT IS ABOUT MONEY.

`load_cache` / `save_cache` are a read-modify-write of one JSON file with no
lock. Claude Code fires PreToolUse once per tool call, and parallel tool calls
mean parallel guard processes on the SAME actor cache. Each reads the same
pre-state, mutates its own copy, and the last writer wins -- every other
process's mutation is silently lost.

The lost mutation that costs real money is `fable_escalations`, the counter
`request_escalation` reads to decide whether this actor has spent its
KIPI_FABLE_CAP budget of cross-model triages. A lost increment means the next
process reads a stale count, decides it is under the cap, and spends another
billed `claude -p --model claude-fable-5` call.

Measured over the shipped ledger (q-system/output/fable-escalations/*.jsonl,
188 rows, 2026-08-08): 172 rows attempted a real call against a cap of 2 per
actor. 90 of them landed in a second that already carried another real call,
7 in the worst single second, and 8 distinct seconds carry two DIFFERENT
triggers -- which only two concurrent guard processes can produce.

NO REAL FABLE CALL IS EVER SPENT HERE. pytest exports PYTEST_CURRENT_TEST for
the duration of every test and the detached child inherits it, so
`call_fable`'s chokepoint refuses the live path and the child still writes its
ledger row. The row is therefore a receipt for an ATTEMPT that would have been
billed in production, which is exactly the quantity under test.
"""

import json
import os
import subprocess
import sys
import time
import uuid

import pytest

GUARD = os.path.abspath(
    os.path.join(os.path.dirname(__file__), "..", "token-guard.py"))
VOLUME_CEILING = 50
FABLE_CAP = 2
PARALLEL_FIRES = 8

# WHY A BARRIER AND NOT JUST N Popen CALLS.
#
# The first cut of this test fired 8 plain `python3 token-guard.py`
# subprocesses and PASSED against the unfixed code -- 3/3 green on a defect
# that provably exists. Interpreter startup is ~40ms and the guard's
# read-modify-write window is under 1ms, so process 1 had finished before
# process 8 had finished importing. The fires never overlapped; the test
# measured scheduling luck, not locking.
#
# So each process imports the guard FIRST, then parks on a barrier file, and
# the parent releases all of them at once. Interpreter startup is moved out of
# the measured window instead of being allowed to serialize it. Nothing about
# the code under test is softened: these are still N separate OS processes
# running the real main() against the real cache path, which is exactly the
# shape the hook runner produces on parallel tool calls.
_BARRIER_RUNNER = '''
import importlib.util, io, json, os, sys, time
guard_path, payload_file, barrier = sys.argv[1], sys.argv[2], sys.argv[3]
spec = importlib.util.spec_from_file_location("guard_under_test", guard_path)
module = importlib.util.module_from_spec(spec)
spec.loader.exec_module(module)
payload = open(payload_file).read()
while not os.path.exists(barrier):
    time.sleep(0.001)
sys.stdin = io.StringIO(payload)
try:
    module.main()
except SystemExit:
    pass
'''


def cache_file(actor_key):
    return f"/tmp/claude-guard-{actor_key}.json"


@pytest.fixture
def actor(tmp_path):
    """A unique actor key, its seeded cache, and cleanup.

    Unique per run on purpose: the guard hardcodes its cache under /tmp, so
    isolation comes from the key, not from the directory. A shared key would
    make two test runs race each other -- the very defect under test.
    """
    key = "race-%s" % uuid.uuid4()
    yield key
    for path in (cache_file(key), f"/tmp/claude-fable-pending-{key}.json"):
        try:
            os.remove(path)
        except OSError:
            pass


def seed_at_ceiling(actor_key):
    """A cache parked exactly at the volume ceiling with no escalations spent.

    `last_volume_reset` is NOW so the commit-progress valve cannot lift the
    ceiling out from under the test: `reset_volume_if_committed` only fires on
    a HEAD commit NEWER than that stamp, and no commit is in the future.
    """
    state = {
        "actor_key": actor_key,
        "tool_calls_since_user": VOLUME_CEILING,
        "agent_calls_since_user": 0,
        "mcp_timestamps": [],
        "repeat_map": {},
        "consecutive_reads": 0,
        "warnings_issued": 0,
        "file_read_counts": {},
        "greps_since_write": 0,
        "edit_targets": {},
        "agents_without_write": 0,
        "last_write_time": time.time(),
        "calls_since_write": 0,
        "last_volume_reset": time.time(),
        "gate_grace_remaining": 0,
        "gate_grace_gate": None,
        "gate_grace_grants": 0,
        "fable_escalations": 0,
    }
    with open(cache_file(actor_key), "w") as fh:
        json.dump(state, fh)


def read_ledger(directory):
    rows = []
    if not os.path.isdir(directory):
        return rows
    for name in sorted(os.listdir(directory)):
        if not name.endswith(".jsonl"):
            continue
        with open(os.path.join(directory, name), encoding="utf-8") as fh:
            for line in fh:
                line = line.strip()
                if line:
                    try:
                        rows.append(json.loads(line))
                    except ValueError:
                        pass
    return rows


def await_rows(directory, expected, deadline_s=30):
    """Poll until `expected` ledger rows land, or the deadline passes.

    Polled, never slept-for-a-guess: the children are detached, so a fixed
    sleep would either flake or pad every run. Returning short is allowed --
    the assertions, not this helper, decide whether short is a failure.
    """
    end = time.time() + deadline_s
    rows = read_ledger(directory)
    while len(rows) < expected and time.time() < end:
        time.sleep(0.2)
        rows = read_ledger(directory)
    return rows


def fire_parallel_blocks(actor_key, ledger_dir, tmp_path, count=PARALLEL_FIRES):
    """`count` guard processes entering main() at the same instant.

    Each is a real hook invocation of the real script against the real cache
    path. See _BARRIER_RUNNER for why the release is synchronised.
    """
    env = dict(os.environ)
    env.update({
        "CLAUDECODE": "1",
        "KIPI_FABLE_LEDGER_DIR": str(ledger_dir),
        "KIPI_FABLE_CAP": str(FABLE_CAP),
        "KIPI_FABLE_PENDING": str(tmp_path / "pending.json"),
        # Not a git repo, so `_head_commit_epoch` returns None deterministically
        # and the commit valve cannot reset the ceiling mid-test.
        "CLAUDE_PROJECT_DIR": str(tmp_path),
    })
    runner = tmp_path / "barrier_runner.py"
    runner.write_text(_BARRIER_RUNNER)
    payload = tmp_path / "payload.json"
    payload.write_text(json.dumps({
        "session_id": actor_key,
        "hook_event_name": "PreToolUse",
        "tool_name": "Read",
        "tool_input": {"file_path": "/etc/hostname"},
        "transcript_path": "",
    }))
    barrier = tmp_path / "barrier"

    procs = [
        subprocess.Popen(
            [sys.executable, str(runner), GUARD, str(payload), str(barrier)],
            stdout=subprocess.DEVNULL, stderr=subprocess.DEVNULL, env=env)
        for _ in range(count)
    ]
    _await_parked(procs)
    barrier.write_text("go")
    for proc in procs:
        proc.wait(timeout=60)
    return procs


def _await_parked(procs, settle_s=1.5):
    """Give every child time to finish importing and reach the barrier.

    A generous fixed settle rather than a handshake: the cost of being early
    is a test that silently stops overlapping (the exact failure this file was
    rewritten to fix), so the check below turns "a child died before the
    barrier" into a loud failure instead of a quiet non-overlap.
    """
    time.sleep(settle_s)
    dead = [p.returncode for p in procs if p.poll() is not None]
    assert not dead, (
        "%d of %d children exited BEFORE the barrier was released (codes %r); "
        "they never overlapped, so any green result below is meaningless"
        % (len(dead), len(procs), dead))


def test_parallel_fires_cannot_overspend_the_escalation_cap(actor, tmp_path):
    """N concurrent ceiling blocks must spend at most FABLE_CAP real calls.

    The assertion is on the LEDGER, not on the cache: a ledger row with
    `capped: false` is a call that reached `call_fable`, i.e. one that
    production would have billed. Counting the cache counter instead would
    measure the symptom's cousin rather than the spend.
    """
    ledger = tmp_path / "ledger"
    seed_at_ceiling(actor)
    fire_parallel_blocks(actor, ledger, tmp_path)
    rows = await_rows(ledger, PARALLEL_FIRES)

    # Guard against the green-for-nothing hole: if the children never ran, the
    # billed-call count is trivially 0 and the real assertion below proves
    # nothing. Every fire must have left a receipt of SOME kind.
    assert len(rows) == PARALLEL_FIRES, (
        "expected one ledger row per fire, got %d of %d -- the spend assertion "
        "below is only meaningful if every child actually ran"
        % (len(rows), PARALLEL_FIRES))

    billed = [r for r in rows if not r.get("capped")]
    assert len(billed) <= FABLE_CAP, (
        "%d of %d parallel fires reached call_fable against a cap of %d; the "
        "escalation counter lost %d increment(s) to the unlocked cache "
        "read-modify-write (sp-016776e6)"
        % (len(billed), PARALLEL_FIRES, FABLE_CAP, len(billed) - FABLE_CAP))


def test_parallel_fires_do_not_lose_the_escalation_counter(actor, tmp_path):
    """The persisted counter must show the cap fully spent, not one increment.

    Separate from the ledger assertion on purpose. The ledger proves what was
    SPENT; this proves the guard's own memory of it survived, which is what
    stops the next fire in the same session from spending again.
    """
    ledger = tmp_path / "ledger"
    seed_at_ceiling(actor)
    fire_parallel_blocks(actor, ledger, tmp_path)
    await_rows(ledger, PARALLEL_FIRES)

    with open(cache_file(actor)) as fh:
        cache = json.load(fh)
    assert cache.get("fable_escalations", 0) >= FABLE_CAP, (
        "cache recorded %r escalations after %d parallel fires; concurrent "
        "writers clobbered each other's increments (sp-016776e6)"
        % (cache.get("fable_escalations"), PARALLEL_FIRES))


def test_parallel_fires_do_not_lose_the_volume_counter(actor, tmp_path):
    """The same lost update, on the counter every run touches.

    The escalation counter is where the race costs money; this is where it
    costs the ceiling its accuracy. Both are the one defect, so both are
    pinned -- fixing only the expensive symptom would leave the class alive.
    """
    ledger = tmp_path / "ledger"
    seed_at_ceiling(actor)
    # Below the warning line, so no block fires and no child is spawned: this
    # isolates the plain counter increment from the escalation path entirely.
    with open(cache_file(actor)) as fh:
        cache = json.load(fh)
    cache["tool_calls_since_user"] = 0
    with open(cache_file(actor), "w") as fh:
        json.dump(cache, fh)

    fire_parallel_blocks(actor, ledger, tmp_path)

    with open(cache_file(actor)) as fh:
        cache = json.load(fh)
    assert cache.get("tool_calls_since_user") == PARALLEL_FIRES, (
        "counted %r of %d parallel tool calls; the rest were lost to the "
        "unlocked read-modify-write, so the ceiling undercounts a fan-out "
        "exactly when it is most needed (sp-016776e6)"
        % (cache.get("tool_calls_since_user"), PARALLEL_FIRES))


if __name__ == "__main__":
    # REQUIRED, not decoration. capability-gate.py runs a `runner: python3`
    # entry as `python3 <file>`, so a pytest-only module would define its tests,
    # call none of them, and exit 0 — registered, green, and enforcing nothing.
    sys.exit(pytest.main([__file__, "-v"]))

exec
/bin/zsh -lc 'ls -ld "$TMPDIR" /tmp
ls /tmp/claude-guard-race-*.lock 2>/dev/null | tail -5
find plugins/kipi-core/voicekit/tests -maxdepth 2 -type d -name '"'__pycache__' -print" in /Users/assafkipnis/.config/kipi/review-trees/pr-126
 succeeded in 0ms:
lrwxr-xr-x@     1 root         wheel      11 Feb  4  2026 /tmp -> private/tmp
drwx------@ 28515 assafkipnis  staff  912480 Aug  8 17:54 /var/folders/th/ty0kb7r56dvfrkgkrcbx9fv00000gn/T/
/tmp/claude-guard-race-e0313069-36a5-40b2-a932-0c00fc654ac7.lock
/tmp/claude-guard-race-e752196f-05e8-4cbe-886f-dd502980fe6b.lock
/tmp/claude-guard-race-f1471b56-672e-4278-9346-9b04533f8c1b.lock
/tmp/claude-guard-race-f632e52c-b5bf-4b27-8c9f-6332dfeb87c8.lock
/tmp/claude-guard-race-fec4e825-c4c9-43b0-b99e-861fbc919781.lock
plugins/kipi-core/voicekit/tests/__pycache__

exec
/bin/zsh -lc "rg -n \"correction_ids|scope|first_person_rate|weight\" plugins/kipi-core/voicekit/tests/test_voicekit.py
nl -ba plugins/kipi-core/voicekit/assemble.py | sed -n '45,80p'
nl -ba plugins/kipi-core/voicekit/fingerprint.py | sed -n '35,90p'" in /Users/assafkipnis/.config/kipi/review-trees/pr-126
 succeeded in 0ms:
32:                     "status": "active", "weight": 1.0,
72:        assert p["first_person_rate"] > s["first_person_rate"]
124:    def test_retired_and_zero_weight_rows_are_excluded(self, tmp_path):
127:        rows[1]["weight"] = 0
220:        assert prov["exemplar_ids"] and prov["correction_ids"] == ["c1"]
222:    def test_scoped_correction_excluded_off_channel(self, tmp_path):
223:        cor = [{"id": "c1", "status": "active", "scope": ["x"],
233:        assert "Old rule" not in text and prov["correction_ids"] == []
359:                 "anchor": False, "weight": 1.0,
427:                        "status": "active", "scope": ["linkedin"]}),
429:                        "status": "weird", "scope": ["myspace"]}),
434:                       "unknown scope", "unparseable"):
442:             "status": "promoted", "scope": ["dm"]}))
    45	
    46	def voice_section(voice, channel, counter, slot_index=0, k=selector.DEFAULT_K,
    47	                  slot_kind="post"):
    48	    """(text, provenance) for one slot. Pure; empty Voice -> ('', empty provenance)."""
    49	    picked = selector.select(voice.active_exemplars(), channel, counter,
    50	                             slot_index=slot_index, k=k, slot_kind=slot_kind)
    51	    corrections = voice.active_corrections()
    52	
    53	    parts = []
    54	    if voice.identity.strip():
    55	        parts.append("WHO IS WRITING:\n" + voice.identity.strip())
    56	    if voice.pov.strip():
    57	        parts.append("WHAT HE WRITES ABOUT AND BELIEVES:\n" + voice.pov.strip())
    58	    lex = _lexicon_positive(voice.lexicon)
    59	    if lex:
    60	        parts.append(lex)
    61	    if picked:
    62	        bodies = "\n\n---\n\n".join((r.get("text") or "").strip() for r in picked)
    63	        parts.append("POSTS HE HAS WRITTEN. Match their rhythm, register and "
    64	                     "length. Never reuse their sentences or openings:\n\n" + bodies)
    65	    if corrections:
    66	        lines = "\n".join(f"- {r['instruction']}" for r in corrections
    67	                          if not r.get("scope") or channel in r["scope"])
    68	        if lines:
    69	            parts.append("STANDING CORRECTIONS (these override everything above):\n"
    70	                         + lines)
    71	
    72	    provenance = {
    73	        "exemplar_ids": [str(r.get("id")) for r in picked],
    74	        "exemplar_texts": [(r.get("text") or "") for r in picked],
    75	        "correction_ids": [str(r.get("id")) for r in corrections],
    76	        "selection_reason": selector.selection_reason(picked, channel, counter,
    77	                                                      slot_index),
    78	        "skipped_rows": voice.skipped_rows,
    79	    }
    80	    return "\n\n".join(parts), provenance
    35	
    36	# The formal pairs a model reaches for when it is being careful. Each has a contracted
    37	# form the founder actually uses; a high count is the "characterless" register in one
    38	# number. Word-boundary matched so "is not" inside "this nothing" cannot fire.
    39	_FORMAL_PAIRS = re.compile(
    40	    r"\b(?:do not|does not|did not|is not|are not|was not|were not|cannot|"
    41	    r"can not|will not|would not|should not|could not|have not|has not|it is|"
    42	    r"that is|there is|you are|we are|they are|i am)\b", re.I)
    43	
    44	_FIRST_PERSON = re.compile(r"\b(?:I|I'm|I've|I'll|I'd|we|we're|we've|my|our|me)\b")
    45	
    46	# Hedges: the softeners that dilute a verdict. Small list, presence-per-post is the
    47	# metric; an exhaustive list would be a banned-word gate, which is a different tool.
    48	_HEDGE = re.compile(
    49	    r"\b(?:probably|perhaps|maybe|might|somewhat|arguably|likely|generally|"
    50	    r"typically|often|tend to|seems? to|sort of|kind of)\b", re.I)
    51	
    52	
    53	def sentences(text):
    54	    """Sentence word-counts, in order. Empty segments dropped."""
    55	    counts = []
    56	    for seg in _SENT_SPLIT.split(text or ""):
    57	        if seg is None:
    58	            continue
    59	        words = _WORD.findall(seg)
    60	        if words:
    61	            counts.append(len(words))
    62	    return counts
    63	
    64	
    65	def paragraphs(text):
    66	    """Sentence counts per paragraph (blank-line separated)."""
    67	    out = []
    68	    for para in re.split(r"\n\s*\n", (text or "").strip()):
    69	        if para.strip():
    70	            out.append(max(1, len(sentences(para))))
    71	    return out
    72	
    73	
    74	def _word_count(text):
    75	    return len(_WORD.findall(text or ""))
    76	
    77	
    78	def _monotony_run(counts):
    79	    """Longest run of consecutive sentences within +-2 words of each other.
    80	
    81	    The uniformity tell: a model writes evenly, the founder does not. Measured on his
    82	    corpus the longest observed run is short; generated posts run long.
    83	    """
    84	    best = run = 1
    85	    for prev, cur in zip(counts, counts[1:]):
    86	        run = run + 1 if abs(cur - prev) <= 2 else 1
    87	        best = max(best, run)
    88	    return best if counts else 0
    89	
    90	

exec
/bin/zsh -lc "sed -n '200,245p' plugins/kipi-core/voicekit/tests/test_voicekit.py
python3 -m pytest -q plugins/kipi-core/voicekit/tests/test_voicekit.py::TestAssembly::test_scoped_correction_excluded_off_channel -vv" in /Users/assafkipnis/.config/kipi/review-trees/pr-126
 exited 1 in 70ms:
                     ~~~~~~~~~~~~~~~~~~~^^
    code = main()
    )
    ^
    raise exception
    teardown.throw(exception)
    ~~~~~~~~~~~~~~^^^^^^^^^^^
    config = yield
             ^^^^^
    self.parse(args)
    ~~~~~~~~~~^^^^^^
    )
    ^
    raise exception
    teardown.throw(exception)
    ~~~~~~~~~~~~~~^^^^^^^^^^^
    return (yield)
            ^^^^^
    ~~~~^^^^^^^^^^^^^^
                                              ~~~~~~~~~^^^
    TemporaryFile(buffering=0),
    ~~~~~~~~~~~~~^^^^^^^^^^^^^
    dir = gettempdir()
                        ~~~~~~~~~~~^^
                            dirlist)
FileNotFoundError: [Errno 2] No usable temporary directory found in ['/var/folders/th/ty0kb7r56dvfrkgkrcbx9fv00000gn/T/', '/tmp', '/var/tmp', '/usr/tmp', '/Users/assafkipnis/.config/kipi/review-trees/pr-126']

exec
/bin/zsh -lc "PYTHONPATH=plugins/kipi-core python3 -c \"from voicekit import assemble,fingerprint; V=type('V',(),{'identity':'','pov':'','lexicon':{},'skipped_rows':0,'active_exemplars':lambda s:[],'active_corrections':lambda s:[{'id':'x-only','status':'active','scope':['x'],'instruction':'Use X register'}]}); text,prov=assemble.voice_section(V(),'linkedin',0); print({'prompt_contains_correction':'Use X register' in text,'provenance_correction_ids':prov['correction_ids']}); print({'capitalized':fingerprint.metrics('We build systems.')['first_person_rate'],'lowercase':fingerprint.metrics('we build systems.')['first_person_rate']})\"
PYTHONPATH=plugins/kipi-core python3 -c \"from voicekit.corpus import Voice; v=Voice.__new__(Voice); v.exemplars=[{'id':'e1','status':'active','kind':'post','channel':'linkedin','text':'real text','weight':'high'}]; exec('try:\\n print(v.active_exemplars())\\nexcept Exception as exc:\\n print(type(exc).__name__ + \\\": \\\") + str(exc)')\"" in /Users/assafkipnis/.config/kipi/review-trees/pr-126
 exited 1 in 0ms:
{'prompt_contains_correction': False, 'provenance_correction_ids': ['x-only']}
{'capitalized': 0.0, 'lowercase': 33.33}
  File "<string>", line 2, in <module>
  File "/Users/assafkipnis/.config/kipi/review-trees/pr-126/plugins/kipi-core/voicekit/corpus.py", line 90, in active_exemplars
    and float(r.get("weight", 1.0) or 0) > 0]
        ~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: could not convert string to float: 'high'

During handling of the above exception, another exception occurred:

    from voicekit.corpus import Voice; v=Voice.__new__(Voice); v.exemplars=[{'id':'e1','status':'active','kind':'post','channel':'linkedin','text':'real text','weight':'high'}]; exec('try:\n print(v.active_exemplars())\nexcept Exception as exc:\n print(type(exc).__name__ + ": ") + str(exc)')
                                                                                                                                                                                  ~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "<string>", line 4, in <module>
TypeError: unsupported operand type(s) for +: 'NoneType' and 'str'
ValueError: 

exec
/bin/zsh -lc "PYTHONPATH=plugins/kipi-core python3 -c \"from voicekit.corpus import Voice; v=Voice.__new__(Voice); v.exemplars=[{'id':'e1','status':'active','kind':'post','channel':'linkedin','text':'real text','weight':'high'}]; print(v.active_exemplars())\" 2>&1" in /Users/assafkipnis/.config/kipi/review-trees/pr-126
 exited 1 in 0ms:
    from voicekit.corpus import Voice; v=Voice.__new__(Voice); v.exemplars=[{'id':'e1','status':'active','kind':'post','channel':'linkedin','text':'real text','weight':'high'}]; print(v.active_exemplars())
                                                                                                                                                                                        ~~~~~~~~~~~~~~~~~~^^
  File "/Users/assafkipnis/.config/kipi/review-trees/pr-126/plugins/kipi-core/voicekit/corpus.py", line 90, in active_exemplars
    and float(r.get("weight", 1.0) or 0) > 0]
        ~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^
ValueError: could not convert string to float: 'high'

codex
## Findings

### MINOR

Confirmed: channel-scoped corrections excluded from the prompt are still recorded as applied in provenance, creating a false audit record.

`plugins/kipi-core/voicekit/assemble.py:75`

Reproducer:

```bash
PYTHONPATH=plugins/kipi-core python3 -c "from voicekit import assemble; V=type('V',(),{'identity':'','pov':'','lexicon':{},'skipped_rows':0,'active_exemplars':lambda s:[],'active_corrections':lambda s:[{'id':'x-only','status':'active','scope':['x'],'instruction':'Use X register'}]}); text,prov=assemble.voice_section(V(),'linkedin',0); print({'prompt_contains_correction':'Use X register' in text,'provenance_correction_ids':prov['correction_ids']})"
```

Real output:

```text
{'prompt_contains_correction': False, 'provenance_correction_ids': ['x-only']}
```

### MINOR

Confirmed: first-person measurement is case-sensitive for `we`, `my`, `our`, and `me`, so identical prose gets different fingerprint scores based only on capitalization.

`plugins/kipi-core/voicekit/fingerprint.py:44`

Reproducer:

```bash
PYTHONPATH=plugins/kipi-core python3 -c "from voicekit import fingerprint; print({'capitalized':fingerprint.metrics('We build systems.')['first_person_rate'],'lowercase':fingerprint.metrics('we build systems.')['first_person_rate']})"
```

Real output:

```text
{'capitalized': 0.0, 'lowercase': 33.33}
```

### MINOR

Confirmed: a syntactically valid exemplar with a nonnumeric weight crashes the runtime loader, contradicting its degrade-without-dying contract. Severity stays minor because no producer in this repository emits this malformed shape.

`plugins/kipi-core/voicekit/corpus.py:90`

Reproducer:

```bash
PYTHONPATH=plugins/kipi-core python3 -c "from voicekit.corpus import Voice; v=Voice.__new__(Voice); v.exemplars=[{'id':'e1','status':'active','kind':'post','channel':'linkedin','text':'real text','weight':'high'}]; print(v.active_exemplars())"
```

Real output:

```text
  ...
  File ".../plugins/kipi-core/voicekit/corpus.py", line 90, in active_exemplars
    and float(r.get("weight", 1.0) or 0) > 0]
ValueError: could not convert string to float: 'high'
```

The read-only sandbox refused creating files under `$TMPDIR`, so the reproducers were executed directly with `python3 -c`.

## What is sound

- Focused suite passed: `82 passed in 24.76s`.
- Parallel token-guard tests exercise real subprocess overlap through a barrier. They don’t rely on interpreter-startup timing.
- Fingerprint bounds include corpus extrema, so the encoded corpus doesn’t reject itself.
- Prompt echo compares against the exact exemplar bodies recorded in provenance.
- Worktree discovery now accepts both `.git` directories and `.git` files.
- Fable cap handling preserves an audit row while suppressing the content-free founder page.
- No blocker or major survived reproduction.

## Verdict

Fix the scoped-correction provenance first. Audit metadata must describe what the prompt actually contained.

FINDINGS:
minor|Channel-scoped corrections excluded from the prompt are still recorded as applied in provenance|plugins/kipi-core/voicekit/assemble.py:75
minor|First-person measurement changes when “we”, “my”, “our”, or “me” is capitalized|plugins/kipi-core/voicekit/fingerprint.py:44
minor|A syntactically valid exemplar with a nonnumeric weight crashes the degrade-without-dying runtime path|plugins/kipi-core/voicekit/corpus.py:90
END FINDINGS
tokens used
98,987
## Findings

### MINOR

Confirmed: channel-scoped corrections excluded from the prompt are still recorded as applied in provenance, creating a false audit record.

`plugins/kipi-core/voicekit/assemble.py:75`

Reproducer:

```bash
PYTHONPATH=plugins/kipi-core python3 -c "from voicekit import assemble; V=type('V',(),{'identity':'','pov':'','lexicon':{},'skipped_rows':0,'active_exemplars':lambda s:[],'active_corrections':lambda s:[{'id':'x-only','status':'active','scope':['x'],'instruction':'Use X register'}]}); text,prov=assemble.voice_section(V(),'linkedin',0); print({'prompt_contains_correction':'Use X register' in text,'provenance_correction_ids':prov['correction_ids']})"
```

Real output:

```text
{'prompt_contains_correction': False, 'provenance_correction_ids': ['x-only']}
```

### MINOR

Confirmed: first-person measurement is case-sensitive for `we`, `my`, `our`, and `me`, so identical prose gets different fingerprint scores based only on capitalization.

`plugins/kipi-core/voicekit/fingerprint.py:44`

Reproducer:

```bash
PYTHONPATH=plugins/kipi-core python3 -c "from voicekit import fingerprint; print({'capitalized':fingerprint.metrics('We build systems.')['first_person_rate'],'lowercase':fingerprint.metrics('we build systems.')['first_person_rate']})"
```

Real output:

```text
{'capitalized': 0.0, 'lowercase': 33.33}
```

### MINOR

Confirmed: a syntactically valid exemplar with a nonnumeric weight crashes the runtime loader, contradicting its degrade-without-dying contract. Severity stays minor because no producer in this repository emits this malformed shape.

`plugins/kipi-core/voicekit/corpus.py:90`

Reproducer:

```bash
PYTHONPATH=plugins/kipi-core python3 -c "from voicekit.corpus import Voice; v=Voice.__new__(Voice); v.exemplars=[{'id':'e1','status':'active','kind':'post','channel':'linkedin','text':'real text','weight':'high'}]; print(v.active_exemplars())"
```

Real output:

```text
  ...
  File ".../plugins/kipi-core/voicekit/corpus.py", line 90, in active_exemplars
    and float(r.get("weight", 1.0) or 0) > 0]
ValueError: could not convert string to float: 'high'
```

The read-only sandbox refused creating files under `$TMPDIR`, so the reproducers were executed directly with `python3 -c`.

## What is sound

- Focused suite passed: `82 passed in 24.76s`.
- Parallel token-guard tests exercise real subprocess overlap through a barrier. They don’t rely on interpreter-startup timing.
- Fingerprint bounds include corpus extrema, so the encoded corpus doesn’t reject itself.
- Prompt echo compares against the exact exemplar bodies recorded in provenance.
- Worktree discovery now accepts both `.git` directories and `.git` files.
- Fable cap handling preserves an audit row while suppressing the content-free founder page.
- No blocker or major survived reproduction.

## Verdict

Fix the scoped-correction provenance first. Audit metadata must describe what the prompt actually contained.

FINDINGS:
minor|Channel-scoped corrections excluded from the prompt are still recorded as applied in provenance|plugins/kipi-core/voicekit/assemble.py:75
minor|First-person measurement changes when “we”, “my”, “our”, or “me” is capitalized|plugins/kipi-core/voicekit/fingerprint.py:44
minor|A syntactically valid exemplar with a nonnumeric weight crashes the degrade-without-dying runtime path|plugins/kipi-core/voicekit/corpus.py:90
END FINDINGS

PR #138, opened 2026-08-11
fix(dispatch): a self-heal must restart the cycle, not finish on old code (ASK-294)

assafkip:
Autonomous worker (Sana) on ASK-294. Opened by the worker because the run ended without opening it.

Commits on this branch:
8b655107 fix(dispatch): a self-heal must restart the cycle, not finish on old code (ASK-294)
72f74a0b fix(notify): the callsite gate caught 1 of 4 real producer shapes (ASK-294)
c59865dd fix(notify): the receipt ledger had no reader -- ship it (ASK-294)
8d647460 fix(notify): classify via a command-prefix env var, not argv flags (ASK-294)
d81784a6 test(notify): pin the message-in-$1 contract that three suites depend on (ASK-294)
93cdc981 fix(notify): message back into $1, and a liveness fault is not a receipt (ASK-294)
66b2a84e feat(notify): handle the condition, send nothing -- 48 founder pings/24h to 1 (ASK-294)
740a2941 chore: update system infrastructure
5f5453dc chore: update system infrastructure
b175771e chore: update system infrastructure

Review runs next. Do not merge without it.


assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/pr-138-20260811-095320.md` (328199 bytes).

```
FINDINGS:
major|The new notification audit makes every validation run red because an existing production caller still omits --kind|automation/voice-refresh-nudge.sh:13
major|A receipt silently returns success when the only machine sink cannot write, losing the notification|q-system/.q-system/scripts/slack-notify.sh:208
major|Any unrelated justification phrase disables the founder-tasking gate for the entire message|q-system/.q-system/scripts/founder-actor-gate.py:147
END FINDINGS
```

--- reviewer output, last 54516 bytes of 328199 (full review at the path above) ---

+# messages I can't do anything about", then "I don't want to get slack messages
+# that are useless to me. they should go to you or sana."
+#
+# MEASURED BEFORE THE CHANGE: 48 messages reached #general in 24h from 11
+# producers, and ZERO consumers read the channel back. Every one terminated at
+# the founder by construction.
+#
+# The contract this suite holds: a message reaches the founder's phone ONLY when
+# it names a founder DECISION from a closed allowlist. Everything else is a
+# receipt -- recorded to a machine-readable ledger an agent reads, never
+# delivered. A closed enum is the point: a producer cannot reword its way into
+# the founder's phone, it has to edit the allowlist in a reviewed diff.
+set -uo pipefail
+
+HERE="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
+SCRIPTS="$(cd "$HERE/.." && pwd)"
+REPO="$(cd "$SCRIPTS/../../.." && pwd)"
+# BASH_SOURCE-derived, never $PWD: this suite must test the copy it ships beside,
+# not whatever checkout happened to be current when someone ran it.
+NOTIFY="$SCRIPTS/slack-notify.sh"
+AUDIT="$SCRIPTS/notify-callsite-audit.py"
+
+PASS=0; FAIL=0
+ok()   { PASS=$((PASS+1)); printf '  ok   %s\n' "$1"; }
+bad()  { FAIL=$((FAIL+1)); printf '  FAIL %s\n     %s\n' "$1" "${2:-}"; }
+
+WORK="$(mktemp -d)"
+trap 'rm -rf "$WORK"' EXIT
+
+# Stub curl by prepending to PATH -- the shipped script takes no "where to post"
+# knob, and adding one would be a documented way to aim a real page at /dev/null.
+# Same posture kipi-dispatch.sh takes with its preflight path.
+mkdir -p "$WORK/bin"
+cat > "$WORK/bin/curl" <<'STUB'
+#!/bin/bash
+# Record the delivery attempt and its payload; never touch the network.
+out="$KIPI_TEST_CURL_LOG"
+prev=""
+for a in "$@"; do
+  if [ "$prev" = "--data" ]; then printf '%s\n' "$a" >> "$out"; fi
+  prev="$a"
+done
+exit 0
+STUB
+chmod +x "$WORK/bin/curl"
+
+# One invocation shape for every case, so a case can never accidentally test a
+# different caller than the others.
+run_notify() {
+  local dir="$1"; shift
+  rm -rf "$dir"; mkdir -p "$dir"
+  # `-u` BEFORE the assignments: BSD env (macOS) parses options only until the
+  # first operand, so `KEY=v -u NAME` reads -u as a utility name and dies 127.
+  # That cost three FALSE PASSES on the first run of this suite -- "curl was
+  # never called" is true when the script never ran at all, so every
+  # not-delivered assertion passed for the wrong reason. Hence assert_ran below.
+  env -u KIPI_LINEAR_API_URL \
+      PATH="$WORK/bin:$PATH" \
+      KIPI_TEST_CURL_LOG="$dir/curl.log" \
+      KIPI_SLACK_WEBHOOK="https://hooks.example.invalid/T/B/X" \
+      KIPI_NOTIFY_RECEIPTS="$dir/receipts.jsonl" \
+      KIPI_INSTANCE_NAME="testinst" \
+      bash "$NOTIFY" "$@" >"$dir/out" 2>"$dir/err"
+  echo $? > "$dir/rc"
+  assert_ran "$dir"
+}
+# A not-delivered assertion is only meaningful if the sink actually executed.
+assert_ran() {
+  local rc; rc="$(cat "$1/rc")"
+  if [ "$rc" = "127" ] || grep -q '^env: \|No such file or directory' "$1/err" 2>/dev/null; then
+    bad "harness: sink actually ran" "rc=$rc err=$(head -2 "$1/err")"
+  fi
+}
+delivered() { [ -s "$1/curl.log" ]; }
+recorded()  { [ -s "$1/receipts.jsonl" ]; }
+field() { python3 -c "
+import json,sys
+rows=[json.loads(l) for l in open(sys.argv[1]) if l.strip()]
+print(rows[-1].get(sys.argv[2]))" "$1/receipts.jsonl" "$2" 2>/dev/null; }
+
+echo "== ASK-294 founder-notification decision gate =="
+
+# --- 1. a receipt never reaches the phone, but is never lost either -----------
+D="$WORK/c1"; run_notify "$D" --kind receipt "converge ASK-289: stopped, no PR after round 4"
+if delivered "$D"; then bad "receipt is NOT delivered" "curl was called: $(cat "$D/curl.log")"
+else ok "receipt is NOT delivered"; fi
+if recorded "$D"; then ok "receipt IS recorded to the machine ledger"
+else bad "receipt IS recorded to the machine ledger" "receipts.jsonl empty/missing"; fi
+[ "$(field "$D" kind)" = "receipt" ] && ok "receipt row carries kind=receipt" \
+  || bad "receipt row carries kind=receipt" "got '$(field "$D" kind)'"
+[ "$(field "$D" delivered)" = "False" ] && ok "receipt row records delivered=false" \
+  || bad "receipt row records delivered=false" "got '$(field "$D" delivered)'"
+[ "$(cat "$D/rc")" = "0" ] && ok "receipt exits 0 (a receipt is not a failure)" \
+  || bad "receipt exits 0" "rc=$(cat "$D/rc")"
+
+# --- 2. an allowlisted founder decision DOES get through ---------------------
+# This is the half that matters most. A gate that silences a real alert is a
+# worse outage than the noise it was built to stop, so this case is asserted
+# against the delivery recorder itself, not against "no error was printed".
+D="$WORK/c2"; run_notify "$D" --kind decision --class irreversible-git \
+  "kipi dispatch: paused -- this checkout is behind origin/main. Do: git merge --ff-only origin/main"
+if delivered "$D"; then ok "decision + allowlisted class IS delivered"
+else bad "decision + allowlisted class IS delivered" "curl never called; err=$(cat "$D/err")"; fi
+[ "$(field "$D" delivered)" = "True" ] && ok "decision row records delivered=true" \
+  || bad "decision row records delivered=true" "got '$(field "$D" delivered)'"
+
+D="$WORK/c2b"; run_notify "$D" --kind decision --class out-of-tree-write \
+  "wrote 3 files outside the canonical tree, needs sign-off"
+delivered "$D" && ok "second allowlisted class (out-of-tree-write) IS delivered" \
+  || bad "second allowlisted class delivered" "curl never called"
+
+# --- 3. the closed enum is what stops rewording -------------------------------
+D="$WORK/c3"; run_notify "$D" --kind decision --class needs-a-human \
+  "worker: ASK-1 is STUCK. Needs a human."
+if delivered "$D"; then bad "unknown decision class is REFUSED" "curl was called"
+else ok "unknown decision class is REFUSED"; fi
+[ "$(cat "$D/rc")" = "2" ] && ok "refusal exits 2" || bad "refusal exits 2" "rc=$(cat "$D/rc")"
+grep -qi 'refus' "$D/err" && ok "refusal is loud on stderr" \
+  || bad "refusal is loud on stderr" "err=$(cat "$D/err")"
+[ "$(field "$D" refused)" = "True" ] && ok "refused row is still recorded (nothing is lost)" \
+  || bad "refused row is still recorded" "got '$(field "$D" refused)'"
+
+D="$WORK/c4"; run_notify "$D" --kind decision "no class at all"
+delivered "$D" && bad "decision with NO class is REFUSED" "curl was called" \
+  || ok "decision with NO class is REFUSED"
+
+D="$WORK/c5"; run_notify "$D" --kind chatter "a fourth kind someone invented"
+delivered "$D" && bad "unknown kind is REFUSED" "curl was called" || ok "unknown kind is REFUSED"
+
+# --- 4. legacy one-arg callers FAIL OPEN, loudly ------------------------------
+# Fleet instances carry producers this repo does not, and kipi update ships this
+# script to them. Refusing an unclassified message would silence an instance-local
+# alert nobody has migrated yet -- the exact "gate that silences a real alert"
+# failure. So it is delivered, marked, and recorded for the migration to find.
+D="$WORK/c6"; run_notify "$D" "a producer that has not been migrated yet"
+delivered "$D" && ok "unclassified legacy call still DELIVERS (fail-open)" \
+  || bad "unclassified legacy call still delivers" "curl never called"
+grep -q 'unclassified' "$D/curl.log" && ok "unclassified delivery is MARKED in the text" \
+  || bad "unclassified delivery is marked" "payload=$(cat "$D/curl.log")"
+[ "$(field "$D" kind)" = "unclassified" ] && ok "unclassified row is recorded as such" \
+  || bad "unclassified row recorded" "got '$(field "$D" kind)'"
+
+# --- 5. the fixture guard still outranks everything ---------------------------
+# A decision-class page from a fixture run must STILL not reach a human. The
+# 2026-08-01 scar (three tests paging the founder live) outranks this issue.
+D="$WORK/c7"; rm -rf "$D"; mkdir -p "$D"
+env PATH="$WORK/bin:$PATH" KIPI_TEST_CURL_LOG="$D/curl.log" \
+    KIPI_SLACK_WEBHOOK="https://hooks.example.invalid/T/B/X" \
+    KIPI_NOTIFY_RECEIPTS="$D/receipts.jsonl" \
+    KIPI_LINEAR_API_URL="http://127.0.0.1:8123/graphql" \
+    bash "$NOTIFY" --kind decision --class irreversible-git "fixture must never page" \
+    >"$D/out" 2>"$D/err"
+delivered "$D" && bad "fixture guard outranks the decision gate" "curl was called" \
+  || ok "fixture guard outranks the decision gate"
+
+# --- 6. the static half: no producer may call the sink without a kind ---------
+if [ ! -f "$AUDIT" ]; then
+  bad "notify-callsite-audit.py exists" "missing at $AUDIT"
+else
+  if python3 "$AUDIT" --repo "$REPO" >"$WORK/audit.out" 2>&1; then
+    ok "every notifier call site in this repo declares a --kind"
+  else
+    bad "every notifier call site declares a --kind" "$(head -20 "$WORK/audit.out")"
+  fi
+  # NEGATIVE SELF-TEST. A checker that cannot fail is not a checker. Plant a
+  # producer that calls the sink bare and prove the audit catches it.
+  PLANT="$WORK/plant"; rm -rf "$PLANT"; mkdir -p "$PLANT/q-system/.q-system/scripts"
+  cp "$AUDIT" "$PLANT/q-system/.q-system/scripts/" 2>/dev/null || true
+  cat > "$PLANT/q-system/.q-system/scripts/rogue-producer.sh" <<'ROGUE'
+#!/bin/bash
+NOTIFY="$SCRIPT_DIR/slack-notify.sh"
+bash "$NOTIFY" "a bare page with no kind at all"
+ROGUE
+  ( cd "$PLANT" && git init -q . && git add -A && git -c user.email=t@t -c user.name=t commit -qm t ) >/dev/null 2>&1
+  if python3 "$AUDIT" --repo "$PLANT" >"$WORK/plant.out" 2>&1; then
+    bad "audit CATCHES a bare call site (negative self-test)" "planted rogue producer passed"
+  else
+    grep -q 'rogue-producer.sh' "$WORK/plant.out" \
+      && ok "audit CATCHES a bare call site and names the file" \
+      || bad "audit names the offending file" "$(head -10 "$WORK/plant.out")"
+  fi
+fi
+
+echo
+printf 'passed %d, failed %d\n' "$PASS" "$FAIL"
+[ "$FAIL" -eq 0 ]
diff --git a/q-system/.q-system/token-guard.py b/q-system/.q-system/token-guard.py
index e0316ce8..5976c4fe 100644
--- a/q-system/.q-system/token-guard.py
+++ b/q-system/.q-system/token-guard.py
@@ -49,6 +49,25 @@ import time
 
 
 # --- Thresholds ---
+# THE CAP ESCALATES TO A PEER, NOT TO THE FOUNDER (ASK-310, founder-directed
+# 2026-08-02: "you failed 3 times, ask fable").
+#
+# Every ceiling below used to end its message with "tell the founder what's
+# blocking you". That is the same defect the whole ASK-310 line is about: the
+# system reaching its own limit and converting the remainder into founder work.
+# Three failed attempts is evidence that THIS model is stuck on THIS problem, not
+# evidence that a human is required. A different model is the cheaper, faster and
+# more correct next step, and the founder is what comes after that fails too.
+#
+# Measured cost of not having this, same session: three attempts at diagnosing a
+# test failure ended in a hand-back with the repo still red, when the remaining
+# work was two stale test assumptions a fresh pass found in minutes.
+FABLE_ESCALATION = (
+    "escalate to Fable BEFORE the founder: Agent(subagent_type='general-purpose', "
+    "model='fable') carrying what you tried, what each attempt returned, and the "
+    "exact question. Page the founder only if Fable is also stuck."
+)
+
 RETRY_LIMIT = 3            # Same tool+input N times = block
 VOLUME_CEILING = 50         # Tool calls since last user message = block
 VOLUME_WARNING = 35         # Tool calls since last user message = warn
@@ -267,7 +286,7 @@ def check_exact_retry(tool_name, tool_input, cache):
     key = f"{tool_name}:{input_hash}"
     count = cache.get("repeat_map", {}).get(key, 0)
     if count >= RETRY_LIMIT:
-        return f"You've attempted this exact call {count} times. Stop. Diagnose the failure and tell the founder what's blocking you."
+        return f"You've attempted this exact call {count} times. Stop. Diagnose the failure, then " + FABLE_ESCALATION
     return None
 
 
@@ -355,7 +374,7 @@ def check_edit_spiral(tool_name, tool_input, cache):
     count = cache.get("edit_targets", {}).get(file_path, 0)
     if count >= EDIT_FAIL_LIMIT:
         short = os.path.basename(file_path)
-        return f"{count} edit attempts on {short}. The approach isn't working. Read the file again, find the exact string, or tell the founder what's wrong."
+        return f"{count} edit attempts on {short}. The approach isn't working. Read the file again and find the exact string; if that fails, " + FABLE_ESCALATION
     return None
 
 
@@ -365,7 +384,7 @@ def check_agent_no_output(tool_name, cache):
         return None
     count = cache.get("agents_without_write", 0)
     if count >= AGENT_NO_OUTPUT_LIMIT:
-        return f"{count} agents spawned with no output written. Agents aren't helping. Use Grep/Glob/Read directly or tell the founder what you're looking for."
+        return f"{count} agents spawned with no output written. Use Grep/Glob/Read directly; if you still cannot make progress, " + FABLE_ESCALATION
     return None
 
 
diff --git a/settings-template.json b/settings-template.json
index 0ed92b6d..d240af43 100644
--- a/settings-template.json
+++ b/settings-template.json
@@ -290,6 +290,11 @@
             "command": "if test -f \"$CLAUDE_PROJECT_DIR/q-system/.q-system/scripts/voice-stop-gate.py\"; then python3 \"$CLAUDE_PROJECT_DIR/q-system/.q-system/scripts/voice-stop-gate.py\"; fi",
             "timeout": 30
           },
+          {
+            "type": "command",
+            "command": "if test -f \"$CLAUDE_PROJECT_DIR/q-system/.q-system/scripts/founder-actor-gate.py\"; then python3 \"$CLAUDE_PROJECT_DIR/q-system/.q-system/scripts/founder-actor-gate.py\"; fi",
+            "timeout": 15
+          },
           {
             "type": "command",
             "command": "python3 \"$CLAUDE_PROJECT_DIR/q-system/hooks/auto-commit.py\" 2>/dev/null || true",
diff --git a/validate-separation.py b/validate-separation.py
index b27c975b..5447471f 100755
--- a/validate-separation.py
+++ b/validate-separation.py
@@ -779,6 +779,42 @@ def phase_1():
             print("\n".join(("    " + l) for l in
                             (gate_run.stdout + gate_run.stderr).splitlines()[-15:]))
 
+    # Gate 1.2b: no founder page may be written bare (ASK-310).
+    #
+    # WHY THIS IS HERE AND NOT ONLY IN slack-notify.sh. The runtime gate fails
+    # OPEN for a caller with no --kind, deliberately: `kipi update` ships that
+    # script to instances carrying producers this repo has never seen, and
+    # refusing those would silence an instance-local alert nobody has migrated.
+    # That hole is exactly wide enough for a new bare producer HERE to go
+    # unnoticed, so it is closed statically where the code lives.
+    #
+    # THE SCAR THIS CLOSES. notify-callsite-audit.py was written for ASK-294,
+    # works, exits 1 correctly, and had ZERO callers -- measured 2026-08-02 by
+    # grepping every wiring surface (validate-separation.py, kipi, kipi-update.sh,
+    # lefthook.yml, .github/workflows) and getting no output. Meanwhile 46 of the
+    # 49 rows in ~/.config/kipi/notify-receipts.jsonl were `unclassified` and all
+    # 46 were delivered to the founder. The gate was 100% built and 0% wired, and
+    # so was ff-merge-if-safe.sh, and so were both of their tests. Built is not
+    # wired, and reviewed is not wired. See
+    # q-system/output/rca/rca-work-routed-to-the-founder-2026-08-02.md.
+    notify_audit = os.path.join(SCRIPT_DIR, "q-system", ".q-system", "scripts",
+                                "notify-callsite-audit.py")
+    if file_exists(notify_audit):
+        # `--repo`, not `--repo-root`. The first draft of this wiring passed
+        # --repo-root and argparse would have exited 2, failing the gate for the
+        # wrong reason -- a wired-but-wrong gate, which is the same family of
+        # defect as the unwired one it replaces. Verified against the script's
+        # own add_argument before shipping.
+        audit_run = subprocess.run([sys.executable, notify_audit, "--repo", SCRIPT_DIR],
+                                   capture_output=True, text=True)
+        check("Gate 1.2b: every founder-notification call site declares a --kind",
+              audit_run.returncode == 0)
+        if audit_run.returncode != 0:
+            print("\n".join(("    " + l) for l in
+                            (audit_run.stdout + audit_run.stderr).splitlines()[-18:]))
+    else:
+        check("Gate 1.2b: notify-callsite-audit.py exists", False)
+
     for script in ["audit-morning.py", "verify-schedule.py", "token-guard.py"]:
         check(f"{script} exists", file_exists(os.path.join(scripts_dir, script)))
 

exec
/bin/zsh -lc "nl -ba q-system/.q-system/scripts/linear-dor-drafter.py | sed -n '780,1100p'
rg -n \"def command_label|def cmd_label|add_argument\\(.*remove|removedLabelIds|addedLabelIds\" q-system/.q-system/scripts/linear-sync.py
nl -ba q-system/.q-system/scripts/linear-sync.py | sed -n '900,1100p'" in /Users/assafkipnis/.config/kipi/review-trees/pr-138
 succeeded in 0ms:
   780	
   781	def rebuild_description(desc: str, body: str, count: int,
   782	                        terminal: bool = False) -> str:
   783	    """Founder text + counter marker + a fresh DoR section + whatever followed it.
   784	
   785	    Only the DoR section itself is replaced. Text above it AND text below it are
   786	    carried through byte for byte, which is what keeps the append-only promise
   787	    true for the words a human wrote.
   788	
   789	    When the heading is absent (a differently-worded DoR on a generated capability
   790	    issue), the whole description is treated as prefix and the section is APPENDED.
   791	    Guessing at the boundary of a section we did not write would risk deleting a
   792	    human's text, and a duplicate section is a recoverable mistake where that is not.
   793	    """
   794	    before, heading, _old, after = split_dor_section(desc)
   795	    prefix = strip_owned_marker(before)
   796	    marker = f""
   797	    # The founder's own heading wording is theirs, not this job's. Reuse it when
   798	    # there is one; DOR_HEADING is only the fallback for an appended section.
   799	    out = f"{prefix}\n\n{marker}\n{heading or DOR_HEADING}\n\n{body.strip()}\n"
   800	    if terminal:
   801	        out += f"\n{TERMINAL_NOTE}\n"
   802	    if after.strip():
   803	        out += f"\n{after.lstrip()}"
   804	    return out
   805	
   806	
   807	def needs_scope_label_ids(issue: dict) -> list:
   808	    """The needs-scope label id(s) to remove.
   809	
   810	    Sent as `removedLabelIds`, NOT as a full `labelIds` replacement. The first cut
   811	    replaced the whole set from labels read before the `claude` call, which can be
   812	    300s earlier, so any label added in that window was deleted (codex-review
   813	    finding-2). The claim in that version -- that Linear can only express removal
   814	    as a full replacement -- was false: IssueUpdateInput carries addedLabelIds and
   815	    removedLabelIds, confirmed by schema introspection against the live API on
   816	    2026-08-01. removedLabelIds touches one label and cannot clobber a concurrent one.
   817	    """
   818	    return [lab["id"] for lab in issue_labels(issue)
   819	            if lab.get("name") == NEEDS_SCOPE_LABEL and lab.get("id")]
   820	
   821	
   822	def reread_before_write(ls, issue: dict) -> dict | None:
   823	    """The issue as Linear holds it NOW. None when it could not be read.
   824	
   825	    A redraft's description is read during selection and written up to --timeout
   826	    (default 300s) later, with a `claude` call in between. Everything the write
   827	    depends on can move inside that window, and the write was built entirely from
   828	    the pre-call copy (codex round 1, finding 2).
   829	
   830	    None is a refusal to write, not an empty issue: writing the stale copy because
   831	    the freshness check itself failed is the exact overwrite the check exists to
   832	    stop, and a skipped redraft costs one slot while a clobbered description is
   833	    gone from a permanent object.
   834	    """
   835	    try:
   836	        return (ls.graphql(ISSUE_WRITE_Q, {"id": issue["identifier"]}) or {}).get("issue")
   837	    except Exception as exc:  # noqa: BLE001 - see the docstring
   838	        print(f"  {issue['identifier']}: could not re-read before write: {str(exc)[:100]}",
   839	              file=sys.stderr)
   840	        return None
   841	
   842	
   843	def moved_since_selection(selected_desc: str, fresh: dict) -> bool:
   844	    """True when anything this write depends on changed since selection.
   845	
   846	    Named for what it checks, not for one cause of it: it began as a
   847	    rival-drafter test and a human editing the DoR by hand trips it too.
   848	
   849	    The counter marker is this job's own single-writer token, so it doubles as a
   850	    compare-and-swap version: this job is the only thing that writes it, therefore
   851	    a changed count means a second drafter finished first. A missing needs-scope
   852	    label says the same thing from the other side.
   853	
   854	    Deliberately NOT a lock, and deliberately not a counter kept in a shared file.
   855	    Nothing is claimed before the work; the write is simply abandoned if the state
   856	    it was computed against is gone. That is why this does not recreate the
   857	    read-then-write race attempts-ledger.py exists to prevent (sp-626e9452, out of
   858	    scope): a lost update here loses one night's redraft of one issue, which the
   859	    next run redoes, rather than losing an increment nothing will ever redo.
   860	
   861	    A HUMAN edit is not a rival writer. It leaves the counter alone, so the redraft
   862	    proceeds and rebuild_description carries the human's new text through -- which
   863	    is the point of rebuilding from the fresh description rather than skipping on
   864	    any change at all.
   865	
   866	    HONEST BOUNDARY: this narrows the window from the whole `claude` call to the
   867	    gap between this read and the mutation. It does not close it. Linear's
   868	    issueUpdate takes no expected-version argument, so a true CAS is not available
   869	    to ask for; two drafters colliding inside that gap still resolve last-write-wins.
   870	    """
   871	    fresh_desc = fresh.get("description") or ""
   872	    if NEEDS_SCOPE_LABEL not in label_names(fresh):
   873	        return True
   874	    if redraft_state(fresh_desc) != redraft_state(selected_desc):
   875	        return True
   876	    # THIRD signal, and the one a person actually trips: the DoR section itself
   877	    # changed. The counter and the label are both machine-written, so a human who
   878	    # rewrote the scope by hand during the model call left both untouched and the
   879	    # redraft wrote straight over their words (codex round 4). A hand edit is the
   880	    # most likely edit there is -- someone reading the refusal and fixing it -- so
   881	    # the section body is part of what "unchanged since selection" has to mean.
   882	    return dor_section(fresh_desc) != dor_section(selected_desc)
   883	
   884	
   885	def update_issue(ls, issue: dict, payload: dict) -> None:
   886	    """The single write chokepoint. Raises when Linear reports success=false.
   887	
   888	    Both write paths ignored the response before this (codex-review finding-3): a
   889	    `{"success": false}` was counted as drafted and printed as "redrafted,
   890	    needs-scope dropped" while neither the label removal nor the cap marker had
   891	    landed -- a silent divergence between what the run reported and what the board
   892	    holds. One chokepoint, so the check cannot exist on one path and not the other.
   893	    """
   894	    res = ls.graphql(UPDATE_M, {"id": issue["identifier"], "input": payload})
   895	    if not (((res or {}).get("issueUpdate") or {}).get("success")):
   896	        raise RuntimeError("Linear returned issueUpdate.success=false "
   897	                           "(nothing was written)")
   898	
   899	
   900	class Skipped(NamedTuple):
   901	    """Why a write was not made, and whether the issue is still queued.
   902	
   903	    `still_queued` is the part callers got wrong: a skip was read as "somebody
   904	    else finished it", but a rival redraft the WORKER REFUSED AGAIN still carries
   905	    needs-scope and is back in the queue, not done. Reporting it as completed
   906	    deletes real remaining work from the count (codex round 4).
   907	    """
   908	    reason: str
   909	    still_queued: bool
   910	
   911	
   912	READY_LABEL = "ready"
   913	
   914	# `1 h`, `30 min`, `2h`, `half day`, `1.5 hrs`. The DoR template asks for exactly
   915	# this field, so the value is drafted prose and the parser has to tolerate it.
   916	TIME_EST_RE = re.compile(
   917	    r"\*\*Time\s*Est\.?:?\*\*[:\s]*([^\n·|]+)", re.IGNORECASE)
   918	_HOURS_RE = re.compile(r"([\d.]+)\s*(h|hr|hrs|hour|hours|m|min|mins|minute|minutes)\b",
   919	                       re.IGNORECASE)
   920	_READY_LABEL_CACHE: dict = {}
   921	
   922	
   923	def parse_time_est_hours(body: str) -> float | None:
   924	    """Hours from the DoR's Time Est field, or None when it cannot be read.
   925	
   926	    None is a real answer and the caller must treat it as one: guessing an
   927	    estimate is worse than leaving the field empty, because a wrong number in a
   928	    cycle's estimate sum silently misreports the week's load -- and the whole
   929	    point of turning estimates on was to make that sum trustworthy (ASK-308).
   930	    """
   931	    field = TIME_EST_RE.search(body or "")
   932	    if not field:
   933	        return None
   934	    text = field.group(1).strip().lower()
   935	    if "half day" in text:
   936	        return 4.0
   937	    if "full day" in text or "all day" in text:
   938	        return 8.0
   939	    match = _HOURS_RE.search(text)
   940	    if not match:
   941	        return None
   942	    value = float(match.group(1))
   943	    return value / 60.0 if match.group(2).lower().startswith("m") else value
   944	
   945	
   946	def hours_to_points(hours: float) -> int:
   947	    """Hours onto the team's fibonacci scale (1, 2, 3, 5, 8), rounding UP.
   948	
   949	    Up, not nearest: a 15-hour week filled by round-down estimates overruns, and
   950	    an overrun week is the failure mode the estimate field exists to prevent.
   951	    """
   952	    for point in (1, 2, 3, 5):
   953	        if hours <= point:
   954	            return point
   955	    return 8
   956	
   957	
   958	def ready_label_id(ls, team_id: str) -> str | None:
   959	    """The `ready` label's id for one team, created on first use, then cached.
   960	
   961	    Resolved by NAME rather than pinned as a uuid because this script is fanned
   962	    to every instance by `kipi update` and each team mints its own label ids -- a
   963	    hardcoded uuid would be correct in exactly one repo and silently wrong in the
   964	    rest. Returns None rather than raising: failing to add a label must not cost
   965	    the founder the DoR that was just drafted.
   966	    """
   967	    if not team_id:
   968	        return None
   969	    if team_id in _READY_LABEL_CACHE:
   970	        return _READY_LABEL_CACHE[team_id]
   971	    try:
   972	        nodes = (((ls.graphql(TEAM_LABELS_Q, {"t": team_id}) or {}).get("team") or {})
   973	                 .get("labels") or {}).get("nodes") or []
   974	        found = next((n["id"] for n in nodes if n.get("name") == READY_LABEL), None)
   975	        if not found:
   976	            made = (ls.graphql(LABEL_CREATE_M, {"input": {
   977	                "name": READY_LABEL, "teamId": team_id, "color": "#0F7B6C",
   978	                "description": "Has an executable Definition of Ready. The autonomous "
   979	                               "worker may pick this up."}}) or {}).get("issueLabelCreate") or {}
   980	            found = ((made.get("issueLabel") or {}).get("id")) if made.get("success") else None
   981	    except Exception as exc:  # noqa: BLE001 - a label lookup must never lose a draft
   982	        print(f"  could not resolve the {READY_LABEL} label: {str(exc)[:100]}",
   983	              file=sys.stderr)
   984	        return None
   985	    _READY_LABEL_CACHE[team_id] = found
   986	    return found
   987	
   988	
   989	def ready_payload(ls, fresh: dict, body: str) -> dict:
   990	    """The `ready` label and the estimate, as issueUpdate fields.
   991	
   992	    WHY THIS EXISTS (ASK-308). Linear cannot filter on "description contains
   993	    Definition of Ready", so until the DoR is also a LABEL the board cannot show
   994	    which issues an agent may pick up -- and 116 of 184 open issues were in that
   995	    blind spot. This script is the only thing that knows a DoR was just written,
   996	    so it is the only correct writer of the label.
   997	
   998	    addedLabelIds, not a labelIds replacement: a delta cannot clobber a label
   999	    applied while this run was drafting. Same reasoning as needs_scope_label_ids.
  1000	    """
  1001	    payload: dict = {}
  1002	    label = ready_label_id(ls, (fresh.get("team") or {}).get("id"))
  1003	    if label and label not in {lab.get("id") for lab in
  1004	                               ((fresh.get("labels") or {}).get("nodes") or [])}:
  1005	        payload["addedLabelIds"] = [label]
  1006	    # Never overwrite an estimate already on the issue: a human who sized it knows
  1007	    # something the drafted prose does not.
  1008	    if not fresh.get("estimate"):
  1009	        hours = parse_time_est_hours(body)
  1010	        if hours:
  1011	            payload["estimate"] = hours_to_points(hours)
  1012	    return payload
  1013	
  1014	
  1015	def apply_write(ls, issue: dict, mode: str, body: str,
  1016	                attempt: int = 0) -> Skipped | None:
  1017	    """THE path from a drafted body to a description write. Every mode goes here.
  1018	
  1019	    Returns None when the write landed, or a short reason when it was deliberately
  1020	    skipped. Raises only on a real transport/refusal failure.
  1021	
  1022	    One function rather than one per mode, because the freshness guard was written
  1023	    for the redraft path and the other two paths silently did not have it (codex
  1024	    round 2): `terminal` and first-`draft` both built their payload from the
  1025	    description read at SELECTION, so an edit made after selection was overwritten
  1026	    by text that predated it -- the exact defect just fixed one branch over. Three
  1027	    call sites cannot drift apart if there is only one.
  1028	
  1029	    Every mode re-reads first and builds from the CURRENT description. What differs
  1030	    per mode is only the applicability check and the payload shape, so those sit
  1031	    here side by side where a missing one is visible.
  1032	    """
  1033	    selected = issue.get("description") or ""
  1034	    fresh = reread_before_write(ls, issue)
  1035	    if fresh is None:
  1036	        raise RuntimeError("could not re-read the issue before writing")
  1037	    fresh_desc = fresh.get("description") or ""
  1038	
  1039	    if mode == "redraft":
  1040	        if moved_since_selection(selected, fresh):
  1041	            return Skipped(
  1042	                "it changed while this attempt was drafting "
  1043	                "(rival redraft, label change, or a hand edit)",
  1044	                selection_mode(fresh) is not None)
  1045	        # Description and label drop go in ONE mutation on purpose. As two calls,
  1046	        # a failure between them leaves a rewritten DoR still wearing needs-scope:
  1047	        # the picker keeps ignoring it and the next night spends another capped
  1048	        # attempt rewriting work already done.
  1049	        #
  1050	        # removedLabelIds, not a full labelIds replacement: Linear applies it as a
  1051	        # delta against the label set as it stands at write time, so a label added
  1052	        # while this redraft was drafting survives. That is the label half of
  1053	        # round 1 finding 2, and it needs no version check because the server-side
  1054	        # semantics already give the atomicity.
  1055	        payload = {"description": rebuild_description(fresh_desc, body, attempt),
  1056	                   "removedLabelIds": needs_scope_label_ids(fresh)}
  1057	        # A redraft REPLACES a refused DoR with a workable one, so the issue
  1058	        # becomes machine-ready in the same write that drops needs-scope. Two
  1059	        # writes would leave a window where it is neither refused nor ready.
  1060	        payload.update(ready_payload(ls, fresh, body))
  1061	    elif mode == "terminal":
  1062	        # The needs-scope label is deliberately NOT removed: the issue really is
  1063	        # unscoped, and dropping it would feed the issue back to a picker that has
  1064	        # already refused it REDRAFT_CAP times.
  1065	        if moved_since_selection(selected, fresh):
  1066	            return Skipped("it changed while this run was working",
  1067	                           selection_mode(fresh) is not None)
  1068	        # dor_section, NOT existing_dor: this value is WRITTEN BACK, so the
  1069	        # prompt's character budget must not touch it.
  1070	        payload = {"description": rebuild_description(
  1071	            fresh_desc, dor_section(fresh_desc) or NO_SECTION_NOTE,
  1072	            REDRAFT_CAP, terminal=True)}
  1073	    else:
  1074	        # A first draft appends. If the issue gained a DoR since selection then
  1075	        # somebody else drafted it, and appending now would give it two.
  1076	        if has_dor_section(fresh_desc):
  1077	            return Skipped("it gained a Definition of Ready while this "
  1078	                           "attempt was drafting",
  1079	                           selection_mode(fresh) is not None)
  1080	        payload = {"description": fresh_desc.rstrip() + f"\n\n{DOR_HEADING}\n\n{body}\n"}
  1081	        # `terminal` deliberately does NOT get this: the cap was exhausted because
  1082	        # the issue is still unscoped, and labelling it ready would hand the picker
  1083	        # back the exact work it already refused REDRAFT_CAP times.
  1084	        payload.update(ready_payload(ls, fresh, body))
  1085	
  1086	    update_issue(ls, issue, payload)
  1087	    return None
  1088	
  1089	
  1090	def _is_missing_issue(exc: Exception) -> bool:
  1091	    """True when Linear said the issue does not exist, as opposed to being down.
  1092	    See MISSING_ISSUE_MARKER for why this is matched on the message."""
  1093	    return MISSING_ISSUE_MARKER in str(exc).lower()
  1094	
  1095	
  1096	def _live_issue(ls, linear_id: str):
  1097	    """The permanent failure issue as Linear has it right now, or None if it is
  1098	    gone. fetch_remote_state cannot answer this: its query filters on project and
  1099	    never selects state, so the dedup key alone says "exists", not "is open".
  1100	
702:def cmd_label(args) -> int:
1176:    p.add_argument("--remove", action="append", metavar="NAME",
   900	            f"BLOCK: plan {args.plan} is not valid JSON ({exc}). Generate one with "
   901	            "`kipi linear plan --map <capability-map> --remote <snapshot> --out <plan>`.",
   902	            file=sys.stderr,
   903	        )
   904	        return EXIT_USAGE
   905	    if not isinstance(plan, dict) or not plan.get("repo"):
   906	        print(f"BLOCK: plan {args.plan} has no 'repo' field", file=sys.stderr)
   907	        return EXIT_USAGE
   908	    repo = plan["repo"]
   909	
   910	    try:
   911	        team_id, project, remote_keys = fetch_remote_state(args.team, repo)
   912	    except LinearAPIError as exc:
   913	        print(f"BLOCK: {exc}", file=sys.stderr)
   914	        return EXIT_USAGE
   915	
   916	    ledger = read_ledger()
   917	    known = set(ledger) | set(remote_keys)
   918	
   919	    want_project = plan.get("create_project")
   920	    if project:
   921	        want_project = None  # already exists remotely; never make a second one
   922	    issues = [i for i in (plan.get("create_issues") or []) if i["key"] not in known]
   923	    skipped = len(plan.get("create_issues") or []) - len(issues)
   924	
   925	    print(
   926	        f"{repo}: would create project={'yes' if want_project else 'exists'}, "
   927	        f"{len(issues)} issue(s), skipping {skipped} already known"
   928	    )
   929	    if not args.apply:
   930	        print("dry run. re-run with --apply to write to Linear.")
   931	        return EXIT_OK
   932	
   933	    if want_project:
   934	        data = graphql(
   935	            PROJECT_CREATE,
   936	            {
   937	                "input": {
   938	                    "name": want_project["name"],
   939	                    "description": (
   940	                        f"{want_project.get('summary', '')}\n\n"
   941	                        f""
   942	                    ).strip(),
   943	                    "teamIds": [team_id],
   944	                }
   945	            },
   946	        )
   947	        created = (data.get("projectCreate") or {}).get("project") or {}
   948	        if not created.get("id"):
   949	            print("BLOCK: projectCreate returned no project", file=sys.stderr)
   950	            return EXIT_USAGE
   951	        project = created
   952	        # Append IMMEDIATELY. A crash before the next write must not lose the
   953	        # record of a permanent object.
   954	        append_ledger([{
   955	            "key": want_project["key"], "kind": "project",
   956	            "linear_id": created["id"], "identifier": created.get("name"),
   957	            "source": "api-create",
   958	        }])
   959	        print(f"  created project {created.get('name')} ({created['id']})")
   960	
   961	    made = 0
   962	    for issue in issues:
   963	        payload = {
   964	            "title": issue["title"],
   965	            "description": issue["description"],
   966	            "teamId": team_id,
   967	        }
   968	        if project:
   969	            payload["projectId"] = project["id"]
   970	        try:
   971	            data = graphql(ISSUE_CREATE, {"input": payload})
   972	        except LinearAPIError as exc:
   973	            # Stop on the first failure. Continuing would keep spending against a
   974	            # broken condition (bad auth, rate limit) and the ledger's account of
   975	            # what exists would drift from Linear.
   976	            print(f"BLOCK after {made} create(s): {exc}", file=sys.stderr)
   977	            return EXIT_USAGE
   978	        node = (data.get("issueCreate") or {}).get("issue") or {}
   979	        if not node.get("id"):
   980	            print(f"BLOCK: issueCreate returned no issue for {issue['key']}", file=sys.stderr)
   981	            return EXIT_USAGE
   982	        append_ledger([{
   983	            "key": issue["key"], "kind": "issue",
   984	            "linear_id": node["id"], "identifier": node.get("identifier"),
   985	            "source": "api-create",
   986	        }])
   987	        made += 1
   988	        print(f"  {node.get('identifier')}  {issue['title'][:70]}")
   989	
   990	    print(f"{repo}: created {made} issue(s). Ledger updated per create.")
   991	    return EXIT_OK
   992	
   993	
   994	def cmd_plan(args) -> int:
   995	    with open(args.map, "r", encoding="utf-8") as fh:
   996	        cmap = json.load(fh)
   997	    repo = cmap.get("repo")
   998	    if not repo:
   999	        print("BLOCK: capability map has no 'repo' field", file=sys.stderr)
  1000	        return EXIT_USAGE
  1001	    caps = cmap.get("capabilities") or []
  1002	
  1003	    # Collision check BEFORE anything else. Two capabilities that slugify to the
  1004	    # same key would silently become one issue, and the second would look "already
  1005	    # created" forever. Refuse the map instead of guessing which one wins.
  1006	    seen = {}
  1007	    for cap in caps:
  1008	        k = make_key(repo, cap.get("name", ""))
  1009	        if k in seen:
  1010	            print(
  1011	                f"BLOCK: '{cap.get('name')}' and '{seen[k]}' both slugify to {k}.\n"
  1012	                f"       Rename one in the capability map. Two capabilities sharing a\n"
  1013	                f"       dedup key would collapse into one permanent issue.",
  1014	                file=sys.stderr,
  1015	            )
  1016	            return EXIT_COLLISION
  1017	        seen[k] = cap.get("name")
  1018	
  1019	    with open(args.remote, "r", encoding="utf-8") as fh:
  1020	        remote = json.load(fh)
  1021	    remote_keys, remote_project = parse_remote(remote)
  1022	
  1023	    ledger = read_ledger()
  1024	
  1025	    # Rehydrate: anything Linear knows about that the ledger does not. This is what
  1026	    # makes a lost ledger a no-op instead of 400 duplicates.
  1027	    rehydrate = [
  1028	        {"key": k, "kind": "issue", "linear_id": v["linear_id"],
  1029	         "identifier": v["identifier"], "source": "remote-rehydrate"}
  1030	        for k, v in remote_keys.items()
  1031	        if k not in ledger
  1032	    ]
  1033	    pkey = project_key(repo)
  1034	    if remote_project and pkey not in ledger:
  1035	        rehydrate.append({
  1036	            "key": pkey, "kind": "project",
  1037	            "linear_id": remote_project.get("id"),
  1038	            "identifier": remote_project.get("name"),
  1039	            "source": "remote-rehydrate",
  1040	        })
  1041	    if rehydrate:
  1042	        append_ledger(rehydrate)
  1043	        ledger = read_ledger()
  1044	
  1045	    known = set(ledger) | set(remote_keys)
  1046	
  1047	    create_issues = []
  1048	    rolled = []
  1049	    for i, cap in enumerate(caps):
  1050	        if not cap.get("track", True):
  1051	            continue
  1052	        if args.filter == "actionable" and _state_for(cap) == "Done":
  1053	            continue
  1054	        # UNWIRED is a survey finding, not a defect: "this script has no test and
  1055	        # no caller" is ONE decision for the repo (audit them, wire them, or delete
  1056	        # them), not N decisions. Filing it per script would make the founder groom
  1057	        # 25 items to make 1 call, and every one of those issues is permanent.
  1058	        # NEEDS_WORK and BROKEN stay individual: each is a distinct defect.
  1059	        if args.rollup and str(cap.get("status", "")).upper() == "UNWIRED":
  1060	            rolled.append(cap)
  1061	            continue
  1062	        issue = build_issue(repo, cap, i)
  1063	        if issue["key"] in known:
  1064	            continue
  1065	        create_issues.append(issue)
  1066	
  1067	    if rolled:
  1068	        rollup = build_rollup(repo, rolled)
  1069	        if rollup["key"] not in known:
  1070	            create_issues.append(rollup)
  1071	
  1072	    create_project = None
  1073	    if pkey not in known:
  1074	        create_project = {
  1075	            "key": pkey,
  1076	            "name": repo,
  1077	            "summary": (cmap.get("summary") or f"Capabilities of the {repo} repo.")[:255],
  1078	            "description": cmap.get("description") or "",
  1079	        }
  1080	
  1081	    plan = {
  1082	        "repo": repo,
  1083	        "generated_at": datetime.now(timezone.utc).strftime("%Y-%m-%dT%H:%M:%SZ"),
  1084	        "create_project": create_project,
  1085	        "create_issues": create_issues,
  1086	        "skipped_known": len(caps) - len(create_issues),
  1087	        "rehydrated": len(rehydrate),
  1088	    }
  1089	    with open(args.out, "w", encoding="utf-8") as fh:
  1090	        json.dump(plan, fh, indent=2, ensure_ascii=False)
  1091	
  1092	    print(
  1093	        f"{repo}: plan {len(create_issues)} issue(s), "
  1094	        f"project={'yes' if create_project else 'exists'}, "
  1095	        f"skipped {plan['skipped_known']} known, rehydrated {len(rehydrate)}"
  1096	    )
  1097	    return EXIT_OK
  1098	
  1099	
  1100	def cmd_record(args) -> int:

exec
/bin/zsh -lc "python3 -c 'import importlib.util; p=\"q-system/.q-system/scripts/founder-actor-gate.py\"; s=importlib.util.spec_from_file_location(\"gate\",p); m=importlib.util.module_from_spec(s); s.loader.exec_module(m); t=\"An agent cannot force-push main.\\n\\n**What&apos;s yours**\\n1. Clear the 20 ordinary PRs.\\n2. Run the backlog sweep.\".replace(\"&apos;\", chr(39)); print(\"violations=\",m.violations(t)); print(\"would_block=\",bool(m.violations(t)))'" in /Users/assafkipnis/.config/kipi/review-trees/pr-138
 succeeded in 0ms:
violations= []
would_block= False

exec
/bin/zsh -lc "nl -ba q-system/.q-system/scripts/slack-notify.sh | sed -n '188,235p'
python3 q-system/.q-system/scripts/notify-callsite-audit.py --repo .
python3 -m py_compile q-system/.q-system/scripts/founder-actor-gate.py q-system/.q-system/scripts/notify-callsite-audit.py q-system/.q-system/scripts/linear-dor-drafter.py
bash -n kipi-dispatch.sh q-system/.q-system/scripts/ff-merge-if-safe.sh q-system/.q-system/scripts/slack-notify.sh" in /Users/assafkipnis/.config/kipi/review-trees/pr-138
 succeeded in 26ms:
[Errno 1] Operation not permitted: 'q-system/.q-system/scripts/__pycache__'
exec
/bin/zsh -lc "set +e; python3 q-system/.q-system/scripts/notify-callsite-audit.py --repo .; printf 'exit=%s\\n' \""'$?"
set +e; KIPI_NOTIFY_RECEIPTS=/System/kipi-review-repro/receipts.jsonl KIPI_INSTANCE_NAME=repro env -u KIPI_LINEAR_API_URL bash q-system/.q-system/scripts/slack-notify.sh --kind receipt '"'repro: machine-handled event'; printf 'exit=%s\\n' \""'$?"' in /Users/assafkipnis/.config/kipi/review-trees/pr-138
 succeeded in 0ms:
notify-callsite-audit: 1 call site(s) reach the founder with no --kind.

  automation/voice-refresh-nudge.sh:13
      bash "$NOTIFY" "$MSG"

Fix: pass --kind receipt (the machine handled it; it is recorded, not delivered)
  or --kind decision --class <irreversible-git|out-of-tree-write|spend|publish>.
A page that names the founder as the actor is a defect in the producer.
exit=1
q-system/.q-system/scripts/slack-notify.sh: line 208: cannot create temp file for here document: Operation not permitted
exit=0

codex
## Findings

### MAJOR

The new notification audit makes every validation run red because an existing production caller still omits `--kind`.

`automation/voice-refresh-nudge.sh:13`

Reproducer:

```bash
python3 q-system/.q-system/scripts/notify-callsite-audit.py --repo .; printf 'exit=%s\n' "$?"
```

Real output:

```text
notify-callsite-audit: 1 call site(s) reach the founder with no --kind.

  automation/voice-refresh-nudge.sh:13
      bash "$NOTIFY" "$MSG"

Fix: pass --kind receipt (the machine handled it; it is recorded, not delivered)
  or --kind decision --class <irreversible-git|out-of-tree-write|spend|publish>.
A page that names the founder as the actor is a defect in the producer.
exit=1
```

Every `kipi check` now cries wolf. A human must classify and update this caller before the gate can provide useful signal.

### MAJOR

A receipt silently returns success when the only machine sink cannot write, permanently losing that notification while telling callers it worked.

`q-system/.q-system/scripts/slack-notify.sh:208`

Reproducer:

```bash
KIPI_NOTIFY_RECEIPTS=/System/kipi-review-repro/receipts.jsonl \
KIPI_INSTANCE_NAME=repro \
env -u KIPI_LINEAR_API_URL \
bash q-system/.q-system/scripts/slack-notify.sh \
  --kind receipt 'repro: machine-handled event'
printf 'exit=%s\n' "$?"
```

Real output:

```text
q-system/.q-system/scripts/slack-notify.sh: line 208: cannot create temp file for here document: Operation not permitted
exit=0
```

The `|| true` suppresses every ledger failure. At 3am, a read-only filesystem, full disk, bad permissions, or unavailable path drops the event from both Slack and the ledger. A human must reconstruct it from individual producer logs.

### MAJOR

Any unrelated justification phrase disables the founder-tasking gate for the entire message, allowing explicit founder chores through.

`q-system/.q-system/scripts/founder-actor-gate.py:147`

Reproducer:

```bash
python3 -c 'import importlib.util; p="q-system/.q-system/scripts/founder-actor-gate.py"; s=importlib.util.spec_from_file_location("gate",p); m=importlib.util.module_from_spec(s); s.loader.exec_module(m); t="An agent cannot force-push main.\n\n**What&apos;s yours**\n1. Clear the 20 ordinary PRs.\n2. Run the backlog sweep.".replace("&apos;", chr(39)); print("violations=",m.violations(t)); print("would_block=",bool(m.violations(t)))'
```

Real output:

```text
violations= []
would_block= False
```

The justification applies globally instead of to the task it explains. A normal safety statement can therefore silently disable the detector for unrelated handoffs in the same response.

## What is sound

- The static audit does detect a real bare notifier call and exits nonzero.
- `bash -n` passed for `kipi-dispatch.sh`, `ff-merge-if-safe.sh`, and `slack-notify.sh`.
- The settings and settings-template both wire the new stop hook.
- The changed Python sources parsed successfully. Bytecode compilation could not finish because this review tree is read-only.
- I attempted the new test suites. The sandbox has no writable temporary directory, so their runtime results were inconclusive rather than code failures.
- GitHub metadata and remote checks were unavailable because `api.github.com` could not be reached. The review used the local PR tree and `origin/main...HEAD`.

## VERDICT

**REQUEST CHANGES**

Fix the permanently red validation gate first. Classify `automation/voice-refresh-nudge.sh` before enabling the audit repo-wide.

FINDINGS:
major|The new notification audit makes every validation run red because an existing production caller still omits --kind|automation/voice-refresh-nudge.sh:13
major|A receipt silently returns success when the only machine sink cannot write, losing the notification|q-system/.q-system/scripts/slack-notify.sh:208
major|Any unrelated justification phrase disables the founder-tasking gate for the entire message|q-system/.q-system/scripts/founder-actor-gate.py:147
END FINDINGS
tokens used
105,887
## Findings

### MAJOR

The new notification audit makes every validation run red because an existing production caller still omits `--kind`.

`automation/voice-refresh-nudge.sh:13`

Reproducer:

```bash
python3 q-system/.q-system/scripts/notify-callsite-audit.py --repo .; printf 'exit=%s\n' "$?"
```

Real output:

```text
notify-callsite-audit: 1 call site(s) reach the founder with no --kind.

  automation/voice-refresh-nudge.sh:13
      bash "$NOTIFY" "$MSG"

Fix: pass --kind receipt (the machine handled it; it is recorded, not delivered)
  or --kind decision --class <irreversible-git|out-of-tree-write|spend|publish>.
A page that names the founder as the actor is a defect in the producer.
exit=1
```

Every `kipi check` now cries wolf. A human must classify and update this caller before the gate can provide useful signal.

### MAJOR

A receipt silently returns success when the only machine sink cannot write, permanently losing that notification while telling callers it worked.

`q-system/.q-system/scripts/slack-notify.sh:208`

Reproducer:

```bash
KIPI_NOTIFY_RECEIPTS=/System/kipi-review-repro/receipts.jsonl \
KIPI_INSTANCE_NAME=repro \
env -u KIPI_LINEAR_API_URL \
bash q-system/.q-system/scripts/slack-notify.sh \
  --kind receipt 'repro: machine-handled event'
printf 'exit=%s\n' "$?"
```

Real output:

```text
q-system/.q-system/scripts/slack-notify.sh: line 208: cannot create temp file for here document: Operation not permitted
exit=0
```

The `|| true` suppresses every ledger failure. At 3am, a read-only filesystem, full disk, bad permissions, or unavailable path drops the event from both Slack and the ledger. A human must reconstruct it from individual producer logs.

### MAJOR

Any unrelated justification phrase disables the founder-tasking gate for the entire message, allowing explicit founder chores through.

`q-system/.q-system/scripts/founder-actor-gate.py:147`

Reproducer:

```bash
python3 -c 'import importlib.util; p="q-system/.q-system/scripts/founder-actor-gate.py"; s=importlib.util.spec_from_file_location("gate",p); m=importlib.util.module_from_spec(s); s.loader.exec_module(m); t="An agent cannot force-push main.\n\n**What&apos;s yours**\n1. Clear the 20 ordinary PRs.\n2. Run the backlog sweep.".replace("&apos;", chr(39)); print("violations=",m.violations(t)); print("would_block=",bool(m.violations(t)))'
```

Real output:

```text
violations= []
would_block= False
```

The justification applies globally instead of to the task it explains. A normal safety statement can therefore silently disable the detector for unrelated handoffs in the same response.

## What is sound

- The static audit does detect a real bare notifier call and exits nonzero.
- `bash -n` passed for `kipi-dispatch.sh`, `ff-merge-if-safe.sh`, and `slack-notify.sh`.
- The settings and settings-template both wire the new stop hook.
- The changed Python sources parsed successfully. Bytecode compilation could not finish because this review tree is read-only.
- I attempted the new test suites. The sandbox has no writable temporary directory, so their runtime results were inconclusive rather than code failures.
- GitHub metadata and remote checks were unavailable because `api.github.com` could not be reached. The review used the local PR tree and `origin/main...HEAD`.

## VERDICT

**REQUEST CHANGES**

Fix the permanently red validation gate first. Classify `automation/voice-refresh-nudge.sh` before enabling the audit repo-wide.

FINDINGS:
major|The new notification audit makes every validation run red because an existing production caller still omits --kind|automation/voice-refresh-nudge.sh:13
major|A receipt silently returns success when the only machine sink cannot write, losing the notification|q-system/.q-system/scripts/slack-notify.sh:208
major|Any unrelated justification phrase disables the founder-tasking gate for the entire message|q-system/.q-system/scripts/founder-actor-gate.py:147
END FINDINGS

assafkip:
Closing as superseded. This branch is 309 commits behind main and conflicting, and it has 3 unaddressed major review findings. Its contents are an ASK-310/ASK-308 sweep, not the ASK-294 fix in the title. The DoR drafter and notify-callsite-audit are already on main, and ASK-353 (#215) covers the founder-queue intent. If the founder-actor gate or the Fable escalation still earns a run, it gets rebuilt fresh from main under ASK-310.

PR #134, opened 2026-08-09
A broken launchctl is not an empty override DB, and a run is recorded after delivery (ASK-447)

assafkip:
Autonomous worker (Sana) on ASK-447. Opened by the worker because the run ended without opening it.

Commits on this branch:
bf7cb07f A broken launchctl is not an empty override DB, and a run is recorded after delivery (ASK-447)
4b994ed2 Lesson: a guard the compound condition never reaches (ASK-447)
1a18ac42 A stale override row is not evidence that a job is running (ASK-447)
b61f215a Verify a launchd job's paused/running state against declared intent (ASK-447)

Review runs next. Do not merge without it.

PR #58, opened 2026-08-01
fix(notify): slack-notify refuses to page a human from a fixture run

assafkip:
## The invariant

A worker running against a **fixture** Linear API must never be able to page a real human.

## Why PR #54 is not enough

#54 stubs `KIPI_NOTIFY` per test. It is correct and should still land. But it has three structural holes:

1. it only protects branches that carry it,
2. it only protects tests someone remembered to fix,
3. its paired lint only fires at write-time on the edited file.

Consequence observed live today: an agent working in a worktree cut from `main` ran `test-worker-project-scope.sh` (which on `main` has no stub) and the founder was paged for real, while #54 sat open and unmerged.

One chokepoint in `slack-notify.sh` protects every test on every branch, including tests nobody has written yet and branches cut before #54 lands. **The per-test stubs and the lint stay as defence in depth.**

## The signal, and what was deliberately excluded

Every test points the worker at a loopback fixture (`KIPI_LINEAR_API_URL=http://127.0.0.1:$PORT/graphql`); production always points at the real Linear API. `slack-notify.sh` runs as `bash "$NOTIFY" "msg"` from the worker, so it **inherits** that variable — verified by running the same `env VAR=... bash parent` shape the tests use and reading the variable from the grandchild, not assumed.

Detection parses the URL **host** (not a substring match), covering `127.0.0.0/8`, `localhost`, `*.localhost`, `::1` (bracketed), `0.0.0.0`.

**Deliberately NOT a signal: `KIPI_STATE_DIR` under a temp dir.** A production job may legitimately keep state in a temp path — macOS `$TMPDIR` is exactly that — so keying on it would suppress real pages. A guard that swallows a genuine alert is worse than the bug it fixes, so the guard keys only on the one signal that cannot be true in production.

The refused text goes to **stderr**, never dropped: a silently swallowed alert is the precise failure mode `founder-notifications.md` exists to prevent.

## Evidence (all captured locally, never real Slack)

Reproducer points `KIPI_SLACK_WEBHOOK` at a local capture endpoint. `KIPI_SLACK_WEBHOOK` is checked before `~/.config/kipi/slack-webhook`, so the real hook is never read. The harness self-tests that it can register a hit, so a `0` means refusal and not a dead capture.

| Check | Result |
|---|---|
| Reproducer on unmodified `main` | **1 page captured** (RED) — test itself reported `14 passed, 0 failed` while paging |
| After guard, same test file unmodified | **0 captured** (GREEN) |
[...]

PR #145, opened 2026-08-14
Get main green: three inherited failures, three different root causes (ASK-740, ASK-746)

assafkip:
`Skeleton Validation` on main has concluded `failure` on 5+ consecutive runs since 2026-08-10, so no PR can satisfy a required check and 30 PRs are queued behind a defect none of them introduced. This fixes all of it.

Three failures, **three unrelated root causes**. Two of the three briefs I was handed turned out to be wrong, and both are corrected below with measurements.

## 1. `test_launchd_health_check.py` — the channel changed, the readers did not (ASK-746)

The spillover said "reports a webhook as DELIVERED when the endpoint REFUSES". **The premise is void: there is no webhook.** On 2026-08-10 `slack-notify.sh`'s destination changed from Slack to Linear; it shells `alert-to-linear.py` and never resolves a webhook.

The webhook was inert, so the assertions were really measuring **whether the runner had a Linear key**:

| runner | what actually decided | result |
|---|---|---|
| CI (no key) | `alert-to-linear` exit 3 `not-configured` | FAIL `accepts -> delivered: got False, want True` |
| keyed machine | real ticket filed, exit 0 | FAIL `refuses -> NOT delivered: got True, want False` |

CI and a dev machine failed on **opposite** assertions. It could not be green anywhere. Worse, on a keyed machine it filed **real Linear tickets** — `repeat #5 on ASK-736`, plus probes ASK-744/ASK-745, all canceled.

There was also a production defect behind it: `notify_channel_configured()` still gated on the dead webhook, so a machine with a valid Linear key but no leftover webhook file returned False and `send_ping` **never ran the notifier at all**. The watchdog's alert was suppressed by a channel that no longer exists.

Fixed: the channel is asked about the destination that exists (key borrowed from `linear-sync.py`, the fleet's single writer), and the test drives delivery through `KIPI_ALERT_CAPTURE`, the seam `alert-to-linear.py` built for exactly this. The delivered case asserts the message **is in the file**, not just the boolean.

## 2. `conftest.py` + `lessons_recall.py` inert-engine — one gate bug, one true positive (ASK-746)

**The gate was wrong about `conftest.py`.** The candidate filter was `os.access(p, os.X_OK) or "__main__" in text` — a bare substring over the whole file. `conftest.py` has no exec bit and no guard; its only `__main__` is one line of **prose in its docstring**. That sentence promoted it to a candidate, and since pytest loads conftest *by name* it could never prove itself wired. Fixed in the gate, with the reason in the code: the guard is matched as syntax at the start of a line, and `RUNNER_LOADED_NAMES` says a runner-loaded file is not an engine. **No fake reference was added.**

**The gate was right about `lessons_recall.py`** — exec bit, real `__main__`, zero references anywhere. It gets a `declared_inert` entry (sp-067205d3), not fake wiring. Wiring it means arming a retrieval engine at SessionStart fleet-wide: a deliberate decision, not a side effect of a CI fix.

The detector was mutation-tested, not trusted: 7 cases including both true positives. The existing `wiring: unwired engine RED` case is the anti-blindness control and stays green.

## 3. The two updater shell tests — a contract collision, not a swallowed exit code (ASK-740)

**`test-kipi-update-safety.sh`** was a real bug. `restore_instance`'s `git checkout -- A B C` is **all-or-nothing**: an instance tracking no `.claude/` or `plugins/` made it error and restore *nothing*, while `git reset` accepted the same unmatched pathspec and returned 0. `2>/dev/null || true` hid both the message and the exit code, so the index looked restored while the worktree kept the failed run's writes. That debris is what pre-refused every later run. Specs are now filtered through `ls-files` first.

**`test-kipi-update-hook-contract.sh`** was *not* a swallowed exit code. Instrumented at 11f0eb8a the guard never fires at all, prints nothing, and exits 0 correctly. It is a collision between two tests: ASK-609 (548919d4, **2026-08-10 — the exact date main went red**) deliberately scoped the guard to paths the sync can write, after measuring 182 dirty files across four blocked instances with **zero** reachable. It shipped its own two-direction test and did not update this one, whose fixture dirties an instance-*root* file.

Decided in favour of ASK-609: root-level dirt genuinely cannot be clobbered, so blocking on it stranded real instances and protected an empty set. The fixture now dirties `q-system/tracked.md`, a file the sync writes. **Every assertion is unchanged** and each now actually exercises the guard instead of an empty set.

## Constraints held

- Nothing skipped, deselected, xfailed, deleted or relaxed. The one gate change is deliberate, is the `conftest.py` case, and is argued in the code.
- Every fix has a negative self-test: reintroduce the defect, watch the test go red, restore, watch it go green.
- `test-kipi-update-dirty-guard-scope.sh` was green at 11f0eb8a and briefly went red on this branch — caused here, not inherited. Its drift check greps a literal the restore fix had to change; it now greps the line that *defines* the scope, and a `for spec in . ; do` mutant still turns it red.

## Verification

```
test_launchd_health_check.py            rc=0  (green under BOTH no-key and keyed, a first)
test_capability_gate.py                 rc=0  ALL PASS (3 new cases + control)
capability-gate.py --check-only         rc=0  GREEN, zero inert-engine ERRORS
test-kipi-update-hook-contract.sh       rc=0
test-kipi-update-safety.sh              rc=0
test-kipi-update-dirty-guard-scope.sh   rc=0
```

Spillover captured, not mentioned-and-dropped: sp-5a3e3b7b (`alert-to-linear.py`'s fixture guard only refuses on `PYTEST_CURRENT_TEST`, so plain-`python3` suites can still file real tickets — that is how this one reached production) and sp-067205d3 (`lessons_recall.py` wire-or-retire).

### DIFF
--- a/kipi-update.sh
+++ b/kipi-update.sh
@@ -510,6 +510,8 @@ checkpoint_instance() {
 
 restore_instance() {
   local target="${CHECKPOINT_TARGET:-}" uf
+  local spec
+  local -a restore_specs=()
   [ -n "$target" ] || return 0
   [ -d "$CHECKPOINT_DIR" ] || return 0
   # An interrupted rebase, first, because aborting one restores HEAD and the
@@ -570,8 +572,32 @@ restore_instance() {
   # move together; the pathspec below is the same one the guard uses.
   git -C "$target" reset -q HEAD -- "$CHECKPOINT_PREFIX/" .claude/ plugins/ \
     $(pathspec_owned_excludes "$CHECKPOINT_PREFIX") 2>/dev/null || true
-  git -C "$target" checkout -q -- "$CHECKPOINT_PREFIX/" .claude/ plugins/ \
-    $(pathspec_owned_excludes "$CHECKPOINT_PREFIX") 2>/dev/null || true
+  # `git checkout -- A B C` is ALL-OR-NOTHING. If ANY pathspec matches nothing
+  # in the index it errors ("pathspec '.claude/' did not match any file(s)
+  # known to git") and restores NONE of them -- while `git reset` above accepts
+  # the same unmatched pathspec and returns 0. An instance that tracks no
+  # .claude/ or plugins/ (a subtree instance that has never had a config sync
+  # land is exactly that) therefore had its ENTIRE worktree restore silently
+  # no-op: `2>/dev/null || true` hid the message and the exit code both, so the
+  # index looked restored while the worktree kept the run's writes.
+  #
+  # That is what stranded instances: a failed run left `M q-system/tracked.md`
+  # behind and every later run refused at the dirty-tree guard, reading the
+  # updater's own debris as founder work. Measured 2026-08-13 (ASK-740): with
+  # the unmatched specs passed, checkout rc=1 and the file stays modified; with
+  # them dropped, rc=0 and the file is restored.
+  #
+  # Filtering by `ls-files` rather than by a directory test on purpose: what
+  # checkout requires is a match in the INDEX, not a directory on disk.
+  for spec in "$CHECKPOINT_PREFIX/" .claude/ plugins/; do
+    if [ -n "$(git -C "$target" ls-files -- "$spec" 2>/dev/null)" ]; then
+      restore_specs+=("$spec")
+    fi
+  done
+  if [ "${#restore_specs[@]}" -gt 0 ]; then
+    git -C "$target" checkout -q -- "${restore_specs[@]}" \
+      $(pathspec_owned_excludes "$CHECKPOINT_PREFIX") 2>/dev/null || true
+  fi
   # Finally, remove files this run created: untracked NOW, absent from the
   # checkpoint, and inside the sync's own scope. Never a recursive delete of a
   # directory this run was not observed to create.
--- a/q-system/.q-system/capability-manifest.json
+++ b/q-system/.q-system/capability-manifest.json
@@ -607,6 +607,11 @@
       "spillover_id": "sp-0f773063"
     },
     {
+      "path": "q-system/.q-system/scripts/lessons_recall.py",
+      "reason": "retrieval + de-duplication over the lessons corpus: exec bit, real __main__, four subcommands, and ZERO references anywhere in the repo outside its own usage docstring. A genuine true positive, not a gate defect. Built to fix a measured decay in q-system/hooks/lessons-index.py (122 lessons, only the 20 most recent ever surfaced, title-only, evicting), which is also the obvious wiring destination -- but arming a retrieval engine at SessionStart changes behaviour on every governed instance, so it is a deliberate decision and not a side effect of a CI-green fix. Declared inert rather than given fake wiring. Wire it or retire it.",
+      "spillover_id": "sp-067205d3"
+    },
+    {
       "path": "q-system/.q-system/scripts/linear-collapse-jobmigration.py",
       "reason": "one-shot family collapse: it absorbed the 32-member job-migration family into ASK-151 on 2026-07-28 and has no second family to run on. ASK-226 generalises it; retire this one when that lands",
       "spillover_id": "sp-fb332466"
--- a/q-system/.q-system/scripts/capability-gate.py
+++ b/q-system/.q-system/scripts/capability-gate.py
@@ -493,6 +493,50 @@ def references_engine(name, surface):
                      surface) is not None
 
 
+# A file the TEST RUNNER loads by name, with no import and no call site anywhere
+# by design. The textual-reference model this whole check rests on structurally
+# cannot express "the runner picks this up by convention", so such a file can
+# only ever be a false positive here -- and the only way to silence it with a
+# reference would be to write a fake one.
+RUNNER_LOADED_NAMES = frozenset({"conftest.py"})
+
+# The `__main__` guard as SYNTAX: start of a line, not anywhere in the bytes.
+_MAIN_GUARD_RE = re.compile(r"^\s*if\s+__name__\s*==\s*['\"]__main__['\"]\s*:",
+                            re.MULTILINE)
+
+
+def is_runnable_engine(path, text):
+    """True when this file RUNS, as opposed to being read or imported.
+
+    Scar (sp-7773af84's neighbour, measured 2026-08-14; main had been red on it
+    since 2026-08-10). The test used to be:
+
+        os.access(p, os.X_OK) or "__main__" in text
+
+    a bare substring over the whole file. `conftest.py` has no exec bit and no
+    guard -- the only `__main__` in it is one line of PROSE inside its module
+    docstring, advising future authors to guard their exits under
+    `if __name__ == "__main__":`. That sentence alone promoted it to a candidate,
+    and because pytest loads conftest by NAME it could never then prove itself
+    wired, so the gate reported it inert on every run forever.
+
+    Two separate defects, so both are closed: a substring cannot tell code from a
+    comment ABOUT code, hence the guard is now matched as syntax at the start of
+    a line; and a runner-loaded file is not an engine at all, hence
+    RUNNER_LOADED_NAMES. Fixing only the first would have hidden the second by
+    accident, because THIS conftest happens to lack an exec bit -- one that
+    carried the exec bit would still be flagged wrongly.
+
+    Deliberately NOT widened past that. The check keeps its own scar in view: it
+    exists because two dead engines citing each other stayed invisible for
+    months, so anything that makes it blind is a bigger cost than a false
+    positive, which is loud and resolved by a declared_inert entry.
+    """
+    if path.name in RUNNER_LOADED_NAMES:
+        return False
+    return os.access(path, os.X_OK) or _MAIN_GUARD_RE.search(text) is not None
+
+
 def check_inert_engines(root, manifest, errors, notes):
     """F2 class: a runnable .py with zero textual references across the wiring
     surfaces and no declared_inert entry is a silently-dead engine.
@@ -522,8 +566,7 @@ def check_inert_engines(root, manifest, errors, notes):
               + list(root.glob("plugins/*/skills/*/scripts/*.py"))):
         if p.is_file() and not p.name.startswith(("test_", "test-")) \
                 and not any(part in ("test", "tests") for part in p.parts):
-            text = p.read_text(errors="ignore")
-            if os.access(p, os.X_OK) or "__main__" in text:
+            if is_runnable_engine(p, p.read_text(errors="ignore")):
                 plugin_candidates.add(p)
     candidates = set()
     for p in list(base.glob("*.py")) + list((base / "scripts").rglob("*.py")):
@@ -532,9 +575,9 @@ def check_inert_engines(root, manifest, errors, notes):
         if any(part in ("test", "tests") for part in p.parts):
             continue
         # runnable contract: exec bit or a __main__ guard; a pure library
-        # module with neither is not an "engine" (standard-review minor)
-        text = p.read_text(errors="ignore")
-        if os.access(p, os.X_OK) or "__main__" in text:
+        # module with neither is not an "engine" (standard-review minor).
+        # One authority for that question -- see is_runnable_engine.
+        if is_runnable_engine(p, p.read_text(errors="ignore")):
             candidates.add(p)
     surface = gather_wiring_text(root, {p.name for p in candidates})
     wired = set()
--- a/q-system/.q-system/scripts/client-name-guard.py
+++ b/q-system/.q-system/scripts/client-name-guard.py
@@ -1,11 +1,20 @@
 #!/usr/bin/env python3
 """Block a client name from reaching a PUBLIC repo, in staged content or the message.
 
-Scar (2026-08-13): a pre-push scan of 15 commits found "[4_points_consulting]
-huntkit sync BLOCKED: /Users/assafkipnis/projects/huntkit" sitting in an alert
-test fixture, plus two commit subjects naming clients. This repo is public with
-102 stars and 23 forks. gitleaks passed all of it, because a client name is not
-a secret in the credential sense. Nothing else looked.
+Scar (2026-08-13): a pre-push scan of 15 commits found a captured production
+alert sitting in a test fixture that carried a CLIENT NAME and the founder's
+absolute home path, plus two commit subjects naming clients. This repo is public
+with 102 stars and 23 forks. gitleaks passed all of it, because a client name is
+not a secret in the credential sense. Nothing else looked.
+
+The offending strings are described by CLASS above and deliberately NOT quoted
+here. Quoting them would republish the exact data this file exists to keep out
+of a public repo -- and this file is on the guard's own exclusion list, so
+nothing would have stopped it. That is not hypothetical: for one commit this
+docstring did quote them verbatim, which put a client name and a real home path
+back into the public tree and tripped validate-separation.py's skeleton sweep
+(caught 2026-08-14, ASK-746). A privacy exclusion names the data class, never
+the data.
 
 Two of those were already public and could not be recalled: rewriting history
 does not remove objects that 23 forks already hold, and GitHub keeps dangling
--- a/q-system/.q-system/scripts/fable-escalate.py
+++ b/q-system/.q-system/scripts/fable-escalate.py
@@ -296,29 +296,59 @@ def log_row(row):
 
 
 def notify_channel_configured():
-    """True when a page could actually leave this machine.
-
-    slack-notify.sh resolves its webhook from $KIPI_SLACK_WEBHOOK then
-    ~/.config/kipi/slack-webhook, and its own header states: "No webhook
-    configured -> silent no-op (exit 0), so callers never break." So its exit
-    code cannot distinguish a delivered message from a swallowed one; the
-    webhook has to be checked separately or delivery is unknowable.
+    """True when an alert could actually leave this machine.
 
     An explicit KIPI_FABLE_NOTIFY_CMD means the caller supplied its own
     notifier and owns its delivery semantics, so its channel is not ours to
     judge.
+
+    Scar (sp-7773af84, measured 2026-08-14, the reason this reads Linear and not
+    a webhook): on 2026-08-10 slack-notify.sh's destination changed from Slack to
+    Linear -- it now shells alert-to-linear.py and never resolves a webhook at
+    all. This function was left behind still gating on $KIPI_SLACK_WEBHOOK and
+    ~/.config/kipi/slack-webhook, which by then decided NOTHING about whether an
+    alert could be filed. Two failures came out of that, in opposite directions:
+
+      1. Production: a machine holding a valid Linear key but no leftover webhook
+         file answered False here, so `send_ping` returned False WITHOUT EVER
+         RUNNING THE NOTIFIER. The watchdog's page was suppressed by a channel
+         that no longer exists.
+      2. CI: test_launchd_health_check.py set a webhook to a refused port and
+         asserted "not delivered". The webhook was inert, so the assertion was
+         really measuring whether the runner had a Linear key -- it failed on
+         `refuses -> NOT delivered` on a keyed machine and on the opposite
+         assertion, `accepts -> delivered`, in CI. It could not be green anywhere,
+         and on a keyed machine it filed REAL tickets (ASK-736 repeats, ASK-744,
+         ASK-745, all canceled).
+
+    So the channel is asked about the destination that exists. The key lookup is
+    borrowed from linear-sync.py -- the fleet's single writer for how it talks to
+    Linear, and the same module alert-to-linear.py loads -- rather than re-derived
+    here, because a copy is exactly what drifted above. KIPI_ALERT_CAPTURE is a
+    genuine configured destination too: the message is appended to that file and
+    nothing is dropped, which is what the bash and python suites redirect into.
+
+    A load failure or a missing key answers False -- "no channel" -- which the
+    callers read as NOT delivered. That costs a duplicate alert, never a missed
+    one. It is the same trade the rest of this path makes.
     """
     if os.environ.get("KIPI_FABLE_NOTIFY_CMD"):
         return True
-    if (os.environ.get("KIPI_SLACK_WEBHOOK") or "").strip():
+    if (os.environ.get("KIPI_ALERT_CAPTURE") or "").strip():
         return True
-    path = os.path.join(os.path.expanduser("~"), ".config", "kipi",
-                        "slack-webhook")
     try:
-        with open(path, encoding="utf-8") as fh:
-            return bool(fh.read().strip())
-    except OSError:
+        spec = importlib.util.spec_from_file_location(
+            "kipi_linear_sync_channel",
+            os.path.join(os.path.dirname(os.path.abspath(__file__)),
+                         "linear-sync.py"))
+        if spec is None or spec.loader is None:
+            return False
+        module = importlib.util.module_from_spec(spec)
+        spec.loader.exec_module(module)
+        module.linear_api_key()
+    except Exception:  # noqa: BLE001
         return False
+    return True
 
 
 # slack-notify.sh's EXIT CONTRACT (ASK-534), transcribed once. Read it as a
--- a/q-system/.q-system/scripts/launchd-health-check.py
+++ b/q-system/.q-system/scripts/launchd-health-check.py
@@ -218,11 +218,16 @@ def notify_channel_configured():
     """True when a page could actually leave this machine.
 
     Borrowed, not re-derived: `fable-escalate.py` already owns the Python read of
-    slack-notify.sh's webhook resolution order ($KIPI_SLACK_WEBHOOK, then
-    ~/.config/kipi/slack-webhook), and a second copy here is the drift
-    `fleet-homogeneity` exists to stop. Loaded through the same lazy-importlib
-    shape as `_fleet_health` / `_intent_module`, so its absence cannot stop the
-    watchdog.
+    whether slack-notify.sh has a destination to write to, and a second copy here
+    is the drift `fleet-homogeneity` exists to stop. Loaded through the same
+    lazy-importlib shape as `_fleet_health` / `_intent_module`, so its absence
+    cannot stop the watchdog.
+
+    That destination is LINEAR, not Slack, since 2026-08-10. This paragraph used
+    to name the webhook resolution order ($KIPI_SLACK_WEBHOOK, then
+    ~/.config/kipi/slack-webhook) and it outlived the channel by four days --
+    which is the same way sp-7773af84 stayed invisible. The resolution order is
+    deliberately NOT restated here; it belongs to the function being borrowed.
 
     A load failure answers False -- "delivery unknown". The caller reads that as
     NOT delivered, which costs a repeated ping and never costs a missed one. That
--- a/q-system/.q-system/scripts/test/test-kipi-update-hook-contract.sh
+++ b/q-system/.q-system/scripts/test/test-kipi-update-hook-contract.sh
@@ -232,10 +232,31 @@ git -C "$CONFIG_BLOCKED/instance" log -1 --format=%s |
 [ ! -e "$CONFIG_BLOCKED/capability.log" ] ||
   fail "capability phase ran after config hook rejection"
 
+# The dirt below sits on q-system/tracked.md -- a file the sync WRITES -- and not
+# on the instance-root unrelated.md it used to use. That is a fixture fix, not a
+# relaxed assertion: every check in this block is unchanged and now actually
+# exercises the guard instead of an empty set.
+#
+# WHY (ASK-609, 548919d4, 2026-08-10). The dirty-tree guard used to be repo-wide.
+# It was deliberately scoped that day to the paths the sync can reach, after
+# measuring 182 dirty files across four blocked instances with ZERO of them in a
+# path the sync writes -- four instances locked out of every fix, including the
+# one written to unblock them, by a guard protecting nothing. That commit shipped
+# its own two-direction test (test-kipi-update-dirty-guard-scope.sh) and did NOT
+# update this file, so from 2026-08-10 these five assertions asserted the
+# pre-ASK-609 repo-wide contract against post-ASK-609 code. Instrumented at
+# 11f0eb8a: the guard never fired, no refusal was printed, FAIL stayed 0 and the
+# run exited 0 -- so there was never a swallowed exit code here to fix. That is
+# what kept main's Skeleton Validation red for four days.
+#
+# Two tests encoding opposite contracts is a decision, not a bug, and it is made
+# here in favour of ASK-609: root-level dirt genuinely cannot be clobbered
+# (staging is pathspec-scoped throughout), so blocking on it protected an empty
+# set at the cost of stranding real instances.
 DIRTY="$WORK/dirty"
 make_fixture "$DIRTY"
 DIRTY_HEAD="$(git -C "$DIRTY/instance" rev-parse HEAD)"
-printf 'founder work in progress\n' > "$DIRTY/instance/unrelated.md"
+printf 'founder work in progress\n' > "$DIRTY/instance/q-system/tracked.md"
 set +e
 HOOK_LOG="$DIRTY/hook.log" CAPABILITY_LOG="$DIRTY/capability.log" \
   bash "$DIRTY/skeleton/kipi-update.sh" >"$DIRTY/output.log" 2>&1
@@ -244,7 +265,7 @@ set -e
 [ "$DIRTY_RC" -ne 0 ] || fail "dirty instance returned success"
 [ "$(git -C "$DIRTY/instance" rev-parse HEAD)" = "$DIRTY_HEAD" ] ||
   fail "updater committed unrelated work in progress"
-grep -q 'founder work in progress' "$DIRTY/instance/unrelated.md" ||
+grep -q 'founder work in progress' "$DIRTY/instance/q-system/tracked.md" ||
   fail "updater changed unrelated work in progress"
 [ ! -e "$DIRTY/hook.log" ] ||
   fail "updater reached a commit hook after dirty-tree refusal"
@@ -254,8 +275,8 @@ grep -q 'dirty working tree' "$DIRTY/output.log" ||
 STAGED="$WORK/staged"
 make_fixture "$STAGED"
 STAGED_HEAD="$(git -C "$STAGED/instance" rev-parse HEAD)"
-printf 'staged founder work\n' > "$STAGED/instance/unrelated.md"
-git -C "$STAGED/instance" add unrelated.md
+printf 'staged founder work\n' > "$STAGED/instance/q-system/tracked.md"
+git -C "$STAGED/instance" add q-system/tracked.md
 set +e
 HOOK_LOG="$STAGED/hook.log" CAPABILITY_LOG="$STAGED/capability.log" \
   bash "$STAGED/skeleton/kipi-update.sh" >"$STAGED/output.log" 2>&1
--- a/q-system/.q-system/scripts/test/verify-alert-wiring.sh
+++ b/q-system/.q-system/scripts/test/verify-alert-wiring.sh
@@ -8,6 +8,13 @@
 # webhook script, so they must never be confused for a live copy.
 set -uo pipefail
 
+# The fleet checkout root. A literal /Users/<founder> cannot live under
+# q-system/ in this PUBLIC repo -- validate-separation.py's full skeleton sweep
+# fails on one, and that failure sat invisible for days behind an earlier red
+# gate (ASK-746). Default keeps the behaviour identical on the founder's machine;
+# override for a differently-laid-out checkout.
+FLEET_ROOT="${KIPI_FLEET_ROOT:-$HOME/projects}"
+
 PASS=0; FAIL=0
 ok()  { PASS=$((PASS+1)); printf '  ok   %s\n' "$1"; }
 bad() { FAIL=$((FAIL+1)); printf '  FAIL %s\n       %s\n' "$1" "${2:-}"; }
@@ -30,7 +37,7 @@ while IFS= read -r f; do
   if is_scratch "$f"; then stale=$((stale+1)); continue; fi
   live=$((live+1))
   head -3 "$f" | grep -q "Linear ticket" || wrong="$wrong$f\n"
-done < <(find /Users/assafkipnis/projects -name "slack-notify.sh" -not -path "*/.git/*" 2>/dev/null)
+done < <(find "$FLEET_ROOT" -name "slack-notify.sh" -not -path "*/.git/*" 2>/dev/null)
 if [ -z "$wrong" ]; then
   ok "all $live live copies route to Linear ($stale scratch/worktree copies excluded)"
 else
@@ -43,13 +50,13 @@ bad_ac=""
 while IFS= read -r f; do
   is_scratch "$f" && continue
   grep -q "_notify_slack\|slack-notify" "$f" && bad_ac="$bad_ac$f\n"
-done < <(find /Users/assafkipnis/projects -name "auto-commit.py" -not -path "*/.git/*" 2>/dev/null)
+done < <(find "$FLEET_ROOT" -name "auto-commit.py" -not -path "*/.git/*" 2>/dev/null)
 [ -z "$bad_ac" ] && ok "no live auto-commit.py references the alert path" \
                  || bad "an auto-commit still alerts" "$(printf "$bad_ac")"
 
 # --- 3. the cole carve-out is throttled --------------------------------------
 echo "-- 3. cole carve-out speaks once per day"
-CP=/Users/assafkipnis/projects/cole-gtm/gtm/scripts/podcast/daily_social_publer.py
+CP=$FLEET_ROOT/cole-gtm/gtm/scripts/podcast/daily_social_publer.py
 grep -q "carve-out-said" "$CP" 2>/dev/null \
   && ok "carve-out throttle present" \
   || bad "carve-out throttle missing" "$CP"
@@ -69,9 +76,9 @@ done
 
 # --- 5. the pytest guard holds through the real chain ------------------------
 echo "-- 5. a test cannot file a ticket (full 2-hop chain, per live repo)"
-for repo in /Users/assafkipnis/projects/consulting \
-            /Users/assafkipnis/projects/cole-gtm \
-            /Users/assafkipnis/projects/kipi-system; do
+for repo in $FLEET_ROOT/consulting \
+            $FLEET_ROOT/cole-gtm \
+            $FLEET_ROOT/kipi-system; do
   n="$repo/q-system/.q-system/scripts/slack-notify.sh"
   [ -f "$n" ] || { bad "no notify script in $(basename "$repo")" "$n"; continue; }
   out=$(PYTEST_CURRENT_TEST="probe (call)" bash "$n" "wiring receipt probe" 2>&1)
@@ -133,7 +140,7 @@ for n in names:
 sys.exit(1)
 PY
   then posters="$posters$f\n"; fi
-done < <(grep -rl "config/kipi/slack-webhook" /Users/assafkipnis/projects \
+done < <(grep -rl "config/kipi/slack-webhook" "$FLEET_ROOT" \
           --include=*.py --include=*.sh 2>/dev/null | grep -v "/.git/")
 uniq_posters="$(printf "$posters" | sed '/^$/d' | sort -u)"
 count="$(printf '%s' "$uniq_posters" | grep -c . || true)"
--- a/q-system/.q-system/scripts/test_capability_gate.py
+++ b/q-system/.q-system/scripts/test_capability_gate.py
@@ -223,6 +223,46 @@ def sec_wiring():
         (root / "q-system/.q-system/capability-manifest.json").write_text(json.dumps(m))
         rc, out = run_gate(root, "--check-only")
         check("wiring: declared_inert passes with note", rc == 0 and "DECLARED-INERT" in out)
+    # ASK-746 neighbour: the candidate filter said `"__main__" in text`, a bare
+    # substring over the whole file. conftest.py has no exec bit and no guard --
+    # its only `__main__` is one line of PROSE in its docstring telling future
+    # authors to guard their exits. That sentence promoted it to a candidate, and
+    # since pytest loads conftest BY NAME it could never prove itself wired, so
+    # main's Skeleton Validation carried `inert-engine: conftest.py` from
+    # 2026-08-10 to 2026-08-14. The tempting "fix" is a fake reference on a
+    # wiring surface; the real one is that this was never an engine.
+    #
+    # The control for these lives above: "wiring: unwired engine RED". If a
+    # loosened filter ever stops seeing real dead engines, that case goes green
+    # and these three cannot tell you, so they are read together.
+    with tempfile.TemporaryDirectory() as tmp:
+        root = make_repo(tmp)
+        c = root / "q-system/.q-system/scripts/conftest.py"
+        c.parent.mkdir(parents=True, exist_ok=True)
+        c.write_text('"""Prose only: guard exits under '
+                     '`if __name__ == "__main__":` so it stays importable."""\n')
+        rc, out = run_gate(root, "--check-only")
+        check("ASK-746: conftest.py mentioning __main__ in PROSE is not inert",
+              rc == 0 and "conftest.py" not in out)
+    with tempfile.TemporaryDirectory() as tmp:
+        root = make_repo(tmp)
+        c = root / "q-system/.q-system/scripts/conftest.py"
+        c.parent.mkdir(parents=True, exist_ok=True)
+        c.write_text('if __name__ == "__main__":\n    print("hi")\n')
+        c.chmod(0o755)
+        rc, out = run_gate(root, "--check-only")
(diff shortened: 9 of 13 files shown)


assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/pr-145-20260813-194155.md` (193249 bytes).

```
FINDINGS:
minor|The runnable-engine regex mistakes a standalone __main__ example inside a docstring for executable code and can emit a false inert-engine alert|q-system/.q-system/scripts/capability-gate.py:504
END FINDINGS
```

--- reviewer output, last 54778 bytes of 193249 (full review at the path above) ---

     3	#
     4	# FOUNDER-DIRECTED 2026-08-10: "I dont want to see any of these. Any of the ones
     5	# that need attention should go to Sana - not me." Offered one carve-out so the
     6	# fable-escalation cap could still reach him, the answer was "the fable
     7	# escalation should be gone. I gave it clear instructions." So there is no
     8	# founder path in this script, no flag that re-opens one, and no webhook call.
     9	#
    10	# THE NAME IS NOW WRONG ON PURPOSE. It is still slack-notify.sh because ~30 call
    11	# sites across six repos invoke it by that path, and a rename is a separate,
    12	# larger change than switching the destination. Renaming it and updating every
    13	# caller is captured in .prd-os/spillover.jsonl rather than bundled in here.
    14	# Nothing in this file sends to Slack.
    15	#
    16	# WHY IT CHANGED. #general carried 100 messages between 09:35 and 14:06 PDT on
    17	# 2026-08-10. 51 were auto-commit naming a file set that changed every turn; 35
    18	# were one cole-gtm notice re-announcing a config state once per run, in
    19	# duplicate pairs. The six that mattered -- four security reverts of unsanctioned
    20	# .claude/ changes and a Notion job dead since 13:00 -- were unreadable
    21	# underneath. The destination changed AND both flood sources were fixed at the
    22	# emitter in the same change; routing a flood to Linear would only make it
    23	# harder to clear, since a ticket has to be closed by hand.
    24	#
    25	# Repeats are deduped by alert-to-linear.py: one ticket per alert SHAPE, with a
    26	# counter, not one ticket per firing.
    27	#
    28	# EXIT CONTRACT, unchanged from when this sent to Slack, so callers that read the
    29	# status (kipi-dispatch.sh:197) keep working:
    30	#   0  filed (ticket opened, or a repeat recorded on the open one)
    31	#   1  attempted and FAILED (reason + the message text on stderr)
    32	#   3  nothing to file on: no Linear API key configured (a setup state)
    33	#   4  refused: fixture run
    34	#
    35	# Usage: slack-notify.sh "message text"
    36	set -uo pipefail
    37	
    38	MSG="${1:-}"
    39	[ -n "$MSG" ] || exit 0
    40	
    41	# Project label so a ticket always names which instance raised it. Resolved in
    42	# order: KIPI_INSTANCE_NAME (set by the fleet heartbeat = exact registry name)
    43	# -> git repo root basename -> cwd basename. Every message is prefixed "[label] ".
    44	LABEL="${KIPI_INSTANCE_NAME:-}"
    45	if [ -z "$LABEL" ]; then
    46	  LABEL="$(basename "$(git rev-parse --show-toplevel 2>/dev/null || echo "$PWD")")"
    47	fi
    48	MSG="[$LABEL] $MSG"
    49	
    50	# --- fixture-run guard: a test must never open a live ticket -------------------
    51	# SCAR 2026-08-01. Three tests were found paging the founder's real Slack and
    52	# were fixed by stubbing KIPI_NOTIFY per test (PR #54). While that PR sat open,
    53	# an agent ran test-worker-project-scope.sh from a worktree cut off main -- which
    54	# carried no stub -- and the founder was paged again, live. Per-test stubbing
    55	# only protects branches that carry it and tests someone remembered to fix.
    56	#
    57	# THE SIGNAL. Every test in this repo points the worker at a fixture Linear on
    58	# loopback (KIPI_LINEAR_API_URL=http://127.0.0.1:$PORT/graphql); production
    59	# always points at the real Linear API. That asymmetry is total in both
    60	# directions, which is what makes it safe to key a refusal on.
    61	#
    62	# Kept EXACTLY as-is through the Slack->Linear switch. It would have been easy to
    63	# argue it away ("a loopback Linear IS the fixture server, so let the write
    64	# through") -- but the bash suites assert exit 4 here, and a fixture server that
    65	# is not listening would turn their clean refusal into a failed send. The guard
    66	# also now covers a second runner: alert-to-linear.py refuses independently on
    67	# PYTEST_CURRENT_TEST, which catches python tests that set no URL at all.
    68	#
    69	# DELIBERATELY NOT A SIGNAL: KIPI_STATE_DIR under a temp dir. A production job
    70	# may legitimately keep state in a temp path (macOS $TMPDIR is exactly that), so
    71	# keying on it would suppress real alerts. A guard that swallows a genuine alert
    72	# is worse than the bug it fixes.
    73	#
    74	# The refused text goes to stderr rather than being dropped: a silently swallowed
    75	# alert is the precise failure mode this path exists to prevent.
    76	# Two review findings on PR #58 shaped this function, in OPPOSITE directions:
    77	#   1. A `127.*` shell pattern also matches non-loopback HOSTNAMES beginning
    78	#      "127.", so `127.example.com` would have been read as a fixture and its
    79	#      alert SILENTLY SUPPRESSED. The real invariant is the 127.0.0.0/8 block --
    80	#      four NUMERIC octets in range -- not a string prefix.
    81	#   2. The comparison was case-sensitive, so `LOCALHOST` got through from a
    82	#      fixture run. DNS hostnames are case-insensitive: a genuine bypass.
    83	#
    84	# Lowercasing uses tr, not ${var,,}: /bin/bash on macOS is 3.2, where ${var,,}
    85	# is a syntax error. `[[ =~ ]]` + BASH_REMATCH do exist in 3.2, and the regex
    86	# must stay in an UNQUOTED variable -- quoting it makes 3.2 match it literally.
    87	_kipi_loopback_host() {
    88	  local h re
    89	  h="$(printf '%s' "$1" | tr '[:upper:]' '[:lower:]')"
    90	  case "$h" in
    91	    localhost|*.localhost|::1|0.0.0.0|0) return 0 ;;
    92	  esac
    93	  re='^127\.([0-9]{1,3})\.([0-9]{1,3})\.([0-9]{1,3})$'
    94	  if [[ "$h" =~ $re ]]; then
    95	    [ "${BASH_REMATCH[1]}" -le 255 ] && [ "${BASH_REMATCH[2]}" -le 255 ] \
    96	      && [ "${BASH_REMATCH[3]}" -le 255 ] && return 0
    97	  fi
    98	  return 1
    99	}
   100	
   101	if [ -n "${KIPI_LINEAR_API_URL:-}" ]; then
   102	  _KHOST="${KIPI_LINEAR_API_URL#*://}"   # drop scheme
   103	  _KHOST="${_KHOST%%/*}"                 # drop path
   104	  _KHOST="${_KHOST##*@}"                 # drop userinfo
   105	  case "$_KHOST" in
   106	    \[*\]*) _KHOST="${_KHOST#\[}"; _KHOST="${_KHOST%%\]*}" ;;  # [::1]:8080 -> ::1
   107	    *)      _KHOST="${_KHOST%%:*}" ;;                          # host:port  -> host
   108	  esac
   109	  if _kipi_loopback_host "$_KHOST"; then
   110	    printf 'slack-notify: REFUSED to file a ticket -- fixture run (KIPI_LINEAR_API_URL host "%s" is loopback). Message NOT filed: %s\n' \
   111	           "$_KHOST" "$MSG" >&2
   112	    exit 4
   113	  fi
   114	fi
   115	
   116	WRITER="$(dirname "${BASH_SOURCE[0]}")/alert-to-linear.py"
   117	if [ ! -f "$WRITER" ]; then
   118	  printf 'slack-notify: alert-to-linear.py missing at %s; message NOT filed: %s\n' \
   119	         "$WRITER" "$MSG" >&2
   120	  exit 1
   121	fi
   122	
   123	# BOUNDED. This is called from Stop hooks and launchd jobs, where a hung network
   124	# call costs the caller its whole turn. alert-to-linear.py caps its own HTTP at
   125	# 30s, which is far too long to hold a Stop hook, so the wall clock is enforced
   126	# here too. `timeout` is GNU coreutils and is not on a stock macOS, so it is used
   127	# only when present rather than assumed -- an alert path must not become
   128	# dependent on a homebrew install.
   129	TIMEOUT_BIN=""
   130	command -v timeout  >/dev/null 2>&1 && TIMEOUT_BIN="timeout"
   131	[ -z "$TIMEOUT_BIN" ] && command -v gtimeout >/dev/null 2>&1 && TIMEOUT_BIN="gtimeout"
   132	
   133	if [ -n "$TIMEOUT_BIN" ]; then
   134	  "$TIMEOUT_BIN" 20 python3 "$WRITER" "$MSG"
   135	else
   136	  python3 "$WRITER" "$MSG"
   137	fi
   138	RC=$?
   139	
   140	# 124 is timeout's own "killed on the clock". Report it as a failed send (1)
   141	# rather than leaking a code no caller's contract mentions.
   142	if [ "$RC" -eq 124 ]; then
   143	  printf 'slack-notify: timed out filing the ticket. Message NOT filed: %s\n' "$MSG" >&2
   144	  exit 1
   145	fi
   146	exit "$RC"
   275	            _NOTIFY_SENDER = None
   276	    return _NOTIFY_SENDER
   277	
   278	
   279	def send_ping(message):
   280	    """Attempt the page. Returns True only when it could actually have LEFT.
   281	
   282	    Scar (PR #134 review, major): this returned None on every path, and
   283	    `run_intent_check` recorded the run as seen regardless. slack-notify.sh is a
   284	    silent no-op that STILL EXITS 0 when no webhook resolves -- its own header
   285	    says so -- so neither its exit code nor "the subprocess started" distinguishes
   286	    a delivered page from a swallowed one. Measured on the shipped code with no
   287	    webhook and a dead filer: 0 findings reached Linear, the message was never
   288	    sent, the run was committed anyway, and the following 12 runs were silent.
   289	
   290	    Scar (PR #134 review round 5, same class one layer down, reproduced before
   291	    fixing): the fix above still ended in `return proc.returncode == 0`, and
   292	    slack-notify.sh exits 0 after a curl that never reached Slack -- its last line
   293	    is `curl ... || true`. Measured with the webhook pointing at a refused port:
   294	    curl failed, this returned True, run_intent_check committed the run, and the
   295	    next 13 scheduled runs were silent. A configured channel says a page COULD
   296	    leave the machine; it never said this one did.
   297	
   298	    The POST outcome exists in exactly one place -- inside slack-notify.sh -- so
   299	    it is now REPORTED from there, through that script's EXIT CONTRACT (ASK-534
   300	    gave it distinct codes after auditing every caller), and read by one shared
   301	    function, `fable-escalate.notify_send`, rather than re-inferred here. Borrowed
   302	    through the same lazy importlib shape as `notify_channel_configured`, and a
   303	    load failure answers False: delivery unknown reads as not delivered, which
   304	    costs a duplicate ping and never a missed one.
   305	
   306	    Scar (codex review of PR #134, round 7): for one merge this paragraph credited
   307	    an env-var side channel that this branch built and then DELETED when
   308	    origin/main's exit contract replaced it. The code was right; the comment named
   309	    a mechanism that existed nowhere in this file, which is the reading the next
   310	    branch would have trusted. The deleted design is recorded once, in
   311	    `fable-escalate.notify_send`'s docstring where that decision was made, and
   312	    NOT restated here -- so this file naming it again is by definition stale, and
   313	    test 15 in test_launchd_health_check.py fails on the bare name.
   314	
   315	    HONEST BOUNDARY: True means Slack accepted the POST, not that a human saw it.
   316	    slack-notify.sh's fixture guard refuses to page on a loopback
   317	    KIPI_LINEAR_API_URL -- that now reports `refused-fixture`, so it reads as NOT
   318	    delivered here instead of as a page. Tests stub send_ping regardless (finding
   319	    2 of the round-2 review); production has no loopback Linear URL.
   320	    """
   321	    if not NOTIFY_SCRIPT.exists():
   322	        return False
   323	    if not notify_channel_configured():
   324	        return False
   325	    send = _notify_sender()
   326	    if send is None:
   327	        return False
   328	    try:
   329	        return bool(send(message, notify=str(NOTIFY_SCRIPT))["delivered"])
   330	    except Exception:  # noqa: BLE001
   331	        return False
   332	
   333	
   334	def is_dry_run(args):
   335	    """True when a dry alias was typed. Answers only that; see parse_mode."""
   336	    return any(arg in DRY_FLAGS for arg in args)
   337	
   338	
   339	def parse_mode(args):
   340	    """('live'|'dry'|'refuse', unrecognized_flags).
diff --git a/q-system/.q-system/scripts/test_launchd_health_check.py b/q-system/.q-system/scripts/test_launchd_health_check.py
index 3b545796..326aa486 100644
--- a/q-system/.q-system/scripts/test_launchd_health_check.py
+++ b/q-system/.q-system/scripts/test_launchd_health_check.py
@@ -8,13 +8,12 @@ verified manually on wiring; these tests guard the logic that a refactor could b
 Run: python3 test_launchd_health_check.py   (exit 0 = pass, 1 = fail)
 """
 import contextlib
-import http.server
 import importlib.util
 import io
 import os
 import subprocess
 import sys
-import threading
+import tempfile
 from pathlib import Path
 
 _spec = importlib.util.spec_from_file_location(
@@ -871,56 +870,65 @@ try:
 finally:
     wd.notify_channel_configured = _saved_channel
 
-# --- a CONFIGURED webhook is not a delivered one (PR #134 review round 5) -----
-# The check above only covers the case where no channel exists at all. The scar
-# is the other one: a webhook IS configured, curl cannot reach it, slack-notify.sh
-# exits 0 anyway, and send_ping used to return `proc.returncode == 0` -> True.
-# Measured before the fix with the webhook on a refused port: send_ping True,
-# nothing sent, the run committed, 13 scheduled runs silent.
+# --- a CONFIGURED channel is not a delivered one (PR #134 r5; sp-7773af84) ----
+# The check above only covers the case where no channel exists at all. This is
+# the other one: a channel IS configured, the send does not land, and send_ping
+# must still answer False. Both directions are asserted on purpose -- without the
+# delivered case a send_ping hardcoded to False would pass and page nobody, ever.
 #
-# Both directions are asserted here on purpose. Without the delivered case, a
-# send_ping hardcoded to False would pass this section and page nobody, ever.
-# Endpoints are on 127.0.0.1 only; the suite never contacts Slack.
-def _send_ping_verdict(webhook):
-    """send_ping's answer for one webhook, with the environment restored after."""
-    saved_hook = os.environ.get("KIPI_SLACK_WEBHOOK")
-    saved_api = os.environ.get("KIPI_LINEAR_API_URL")
-    os.environ["KIPI_SLACK_WEBHOOK"] = webhook
-    # The loopback fixture guard would refuse before curl ever runs, which is a
-    # different verdict than the one under test here.
+# SCAR, sp-7773af84, measured 2026-08-14. This block used to drive a
+# $KIPI_SLACK_WEBHOOK on a refused port. On 2026-08-10 slack-notify.sh's
+# destination changed from Slack to Linear and it stopped reading a webhook at
+# all, so the knob these assertions turned was inert and they were really
+# measuring whether the RUNNER HAD A LINEAR KEY. That cannot be green anywhere:
+# CI (no key) failed `accepts -> delivered`, a keyed machine failed
+# `refuses -> NOT delivered`, and on a keyed machine the "assertion" filed REAL
+# tickets -- ASK-736 repeats, plus ASK-744/ASK-745, all canceled. Main's Skeleton
+# Validation had been red on it since 2026-08-10.
+#
+# The lesson is the one already written into alert-to-linear.py's capture hatch:
+# switching a chokepoint's destination silently invalidates every stub aimed at
+# the old one, and those stubs keep passing their own assertions right up until
+# they write to production. So delivery is now driven through the seam that
+# BELONGS to the current destination -- KIPI_ALERT_CAPTURE, which alert-to-linear
+# built for exactly this -- and never through a webhook. Measured directly before
+# these were written: a writable capture path exits 0 (verdict `delivered`), an
+# unopenable one exits 1 (verdict `send-failed`). Neither touches Linear.
+def _send_ping_verdict(capture_path):
+    """send_ping's answer for one capture destination, environment restored after.
+
+    KIPI_LINEAR_API_URL is popped because the loopback fixture guard in
+    slack-notify.sh would exit 4 before the writer ever runs, which is a
+    different verdict than the one under test here.
+    """
+    saved = {k: os.environ.get(k)
+             for k in ("KIPI_ALERT_CAPTURE", "KIPI_LINEAR_API_URL")}
+    os.environ["KIPI_ALERT_CAPTURE"] = capture_path
     os.environ.pop("KIPI_LINEAR_API_URL", None)
     try:
-        return wd.send_ping("probe: local endpoints only, ASK-447")
+        return wd.send_ping(_PROBE_MESSAGE)
     finally:
-        os.environ.pop("KIPI_SLACK_WEBHOOK", None)
-        if saved_hook is not None:
-            os.environ["KIPI_SLACK_WEBHOOK"] = saved_hook
-        if saved_api is not None:
-            os.environ["KIPI_LINEAR_API_URL"] = saved_api
-
-
-class _PostSink(http.server.BaseHTTPRequestHandler):
-    def do_POST(self):  # noqa: N802
-        self.rfile.read(int(self.headers.get("Content-Length", 0) or 0))
-        self.send_response(200)
-        self.end_headers()
-        self.wfile.write(b"ok")
-
-    def log_message(self, *args):  # keep the suite output readable
-        pass
-
-
-check("configured webhook that refuses the connection -> NOT delivered",
-      _send_ping_verdict("http://127.0.0.1:9/services/dead"), False)
-
-_sink = http.server.HTTPServer(("127.0.0.1", 0), _PostSink)
-threading.Thread(target=_sink.serve_forever, daemon=True).start()
-try:
-    check("webhook that accepts the POST -> delivered",
-          _send_ping_verdict("http://127.0.0.1:%d/hook" % _sink.server_address[1]),
-          True)
-finally:
-    _sink.shutdown()
+        for key, value in saved.items():
+            os.environ.pop(key, None)
+            if value is not None:
+                os.environ[key] = value
+
+
+_PROBE_MESSAGE = "probe: captured locally, never filed (sp-7773af84)"
+
+with tempfile.TemporaryDirectory() as _capdir:
+    # NOT delivered: the writer cannot open its destination, so nothing was
+    # recorded anywhere. The parent directory does not exist and is not created.
+    check("configured channel the alert cannot be written to -> NOT delivered",
+          _send_ping_verdict(os.path.join(_capdir, "absent", "cap.txt")), False)
+
+    # delivered: the write lands. Asserting the boolean alone would pass against
+    # a send_ping that returns True without sending, so the FILE is the evidence.
+    _landed = os.path.join(_capdir, "cap.txt")
+    check("channel that accepts the alert -> delivered",
+          _send_ping_verdict(_landed), True)
+    check("the delivered alert is actually in the destination",
+          _PROBE_MESSAGE in Path(_landed).read_text(encoding="utf-8"), True)
 
 # ---------------------------------------------------------------------------
 # The legacy watchdog may not record a page that never left (PR #134 round 6)
diff --git a/q-system/.q-system/tests/test_fable_escalation.py b/q-system/.q-system/tests/test_fable_escalation.py
index 8d912adc..2a3c1ab2 100644
--- a/q-system/.q-system/tests/test_fable_escalation.py
+++ b/q-system/.q-system/tests/test_fable_escalation.py
@@ -688,6 +688,12 @@ def test_cap_row_does_not_claim_a_page_that_was_never_sent(actor, env, tmp_path)
     del e["KIPI_FABLE_NOTIFY_CMD"]          # use the real slack-notify.sh
     e["HOME"] = str(tmp_path / "emptyhome")
     e["KIPI_SLACK_WEBHOOK"] = ""
+    # Every input to notify_channel_configured has to be pinned here, or this
+    # asserts nothing. KIPI_ALERT_CAPTURE is one since ASK-746 (the channel is
+    # Linear now, and a capture file is a real destination), and it is inherited
+    # from the caller's environment -- so an operator running the suite with a
+    # capture set turned this red while the code was fine. Measured 2026-08-14.
+    e["KIPI_ALERT_CAPTURE"] = ""
     e["KIPI_FABLE_CAP"] = "1"
     os.makedirs(e["HOME"], exist_ok=True)
 

exec
/bin/zsh -lc "nl -ba q-system/.q-system/scripts/alert-to-linear.py | sed -n '300,365p'; nl -ba q-system/.q-system/scripts/fable-escalate.py | sed -n '400,455p'; rg -n \"restore_instance|abandon_instance|CHECKPOINT_PREFIX|pathspec_owned_excludes|dirty working tree|diff --quiet\" kipi-update.sh; git show origin/main:kipi-update.sh | nl -ba | sed -n '500,610p'" in /Users/assafkipnis/.config/kipi/review-trees/pr-145
 succeeded in 0ms:
   300	
   301	def main(argv: list[str]) -> int:
   302	    message = (argv[1] if len(argv) > 1 else "").strip()
   303	    if not message:
   304	        return EXIT_OK
   305	
   306	    # CAPTURE HATCH -- the isolation seam bash suites need, and the reason this
   307	    # block exists at all.
   308	    #
   309	    # SCAR, 2026-08-10, ten minutes old when this was written. The bash guard
   310	    # suite isolates itself by pointing KIPI_SLACK_WEBHOOK at a local capture
   311	    # server, then asserts on what arrived. The moment the destination became
   312	    # Linear that stub addressed nothing, so the suite's deliberate "a
   313	    # production run still alerts" cases filed a REAL ticket (ASK-635, canceled).
   314	    # Switching a chokepoint's destination silently invalidates every test stub
   315	    # aimed at the old one, and the tests keep passing their own assertions right
   316	    # up until they write to production.
   317	    #
   318	    # So the capture seam is part of the destination, not part of the caller: any
   319	    # runner that can set an env var can redirect the write. Nothing is dropped
   320	    # -- the message is appended to the file -- and it announces itself on
   321	    # stderr, so this can never be mistaken for a delivered alert in a job log.
   322	    capture = os.environ.get("KIPI_ALERT_CAPTURE")
   323	    if capture:
   324	        try:
   325	            with open(capture, "a", encoding="utf-8") as fh:
   326	                fh.write(message + "\n")
   327	        except OSError as exc:
   328	            print(f"alert-to-linear: capture to {capture} failed ({exc}); "
   329	                  f"NOT filed: {message}", file=sys.stderr)
   330	            return EXIT_FAILED
   331	        print(f"alert-to-linear: CAPTURED to {capture} (not filed to Linear): "
   332	              f"{message}", file=sys.stderr)
   333	        return EXIT_OK
   334	
   335	    # FIXTURE GUARD (the same chokepoint fable-escalate.py uses for model calls).
   336	    # SCAR 2026-08-01: three tests were found paging the founder's real Slack,
   337	    # and while the fix sat unmerged an agent ran one from a worktree without it
   338	    # and paged again. Per-test stubbing only protects tests someone remembered
   339	    # to fix. A test written tomorrow must not be able to open a real ticket, so
   340	    # the refusal lives here rather than in each suite.
   341	    if os.environ.get("PYTEST_CURRENT_TEST"):
   342	        print(f"alert-to-linear: REFUSED under pytest. NOT filed: {message}",
   343	              file=sys.stderr)
   344	        return EXIT_REFUSED_FIXTURE
   345	
   346	    try:
   347	        code, line = file_alert(message)
   348	    except Exception as exc:                       # never crash a Stop hook
   349	        print(f"alert-to-linear: unexpected {exc!r}; NOT filed: {message}",
   350	              file=sys.stderr)
   351	        return EXIT_FAILED
   352	
   353	    # A failed alert stays readable in the job log. A silently swallowed alert is
   354	    # the exact failure mode this whole path exists to prevent.
   355	    print(f"alert-to-linear: {line}", file=sys.stderr if code else sys.stdout)
   356	    return code
   357	
   358	
   359	if __name__ == "__main__":
   360	    sys.exit(main(sys.argv))
   400	    answers False for a rc outside the table and for a notifier that dies before
   401	    it reports -- both cost a duplicate ping, never a missed one.
   402	
   403	    The direction that got WEAKER is worth naming rather than burying. The
   404	    verdict file could not be forged by an old notifier: absent meant unknown
   405	    meant not-delivered. An exit code can be. A pre-ASK-534 slack-notify.sh
   406	    exits 0 on every path, including after a curl that never left the machine,
   407	    and this table reads that 0 as `delivered`. In THIS repo the notifier is
   408	    migrated, so the table is right; an instance running a stale copy through a
   409	    lagging `kipi update` would get a false delivered. That is a version-skew
   410	    hole in the fleet, not in this function, and it is captured as spillover
   411	    rather than papered over with a guess here.
   412	
   413	    A custom KIPI_FABLE_NOTIFY_CMD owns its own delivery semantics, so it is
   414	    judged by its exit code as before; only the default notifier is held to the
   415	    contract table.
   416	    """
   417	    notify = notify or os.environ.get("KIPI_FABLE_NOTIFY_CMD") or DEFAULT_NOTIFY
   418	    # The verdict protocol is a property of the SCRIPT being run, not of who asked
   419	    # for it, so this is keyed on the resolved path. A caller that points
   420	    # KIPI_FABLE_NOTIFY_CMD at the real notifier gets the real verdict; anything
   421	    # else is a notifier this fleet did not write and cannot make claims about.
   422	    own_verdict = os.path.realpath(notify) == os.path.realpath(DEFAULT_NOTIFY)
   423	    configured = notify_channel_configured()
   424	    record = {"attempted": False, "exit": None, "delivered": False,
   425	              "channel_configured": configured, "verdict": None, "note": None}
   426	
   427	    if not os.path.exists(notify):
   428	        record["note"] = "notifier not found at %s" % notify
   429	        return record
   430	
   431	    try:
   432	        proc = subprocess.run(["bash", notify, message], timeout=timeout,
   433	                              capture_output=True, text=True)
   434	    except (subprocess.SubprocessError, OSError) as exc:
   435	        record["note"] = "notifier failed to run: %s" % exc
   436	        return record
   437	
   438	    record["attempted"] = True
   439	    record["exit"] = proc.returncode
   440	
   441	    if not own_verdict:
   442	        # The caller supplied the notifier and owns its semantics.
   443	        record["delivered"] = bool(proc.returncode == 0 and configured)
   444	    else:
   445	        record["verdict"] = NOTIFY_EXIT_VERDICTS.get(
   446	            proc.returncode, "unknown rc=%s" % proc.returncode)
   447	        record["delivered"] = record["verdict"] == "delivered"
   448	
   449	    if not record["delivered"]:
   450	        record["note"] = ("exit %s, verdict %s, channel_configured=%s"
   451	                          % (proc.returncode, record["verdict"] or "none",
   452	                             configured))
   453	    return record
   454	
   455	
110:pathspec_owned_excludes() {
442:CHECKPOINT_PREFIX=""
477:  [ -n "${CHECKPOINT_PREFIX:-}" ] || return 0
479:      "$CHECKPOINT_PREFIX/" .claude/ plugins/ \
480:      $(pathspec_owned_excludes "$CHECKPOINT_PREFIX") 2>/dev/null )
486:  CHECKPOINT_PREFIX="$2"
511:restore_instance() {
544:  # ordering is why abandon_instance owns both steps.
573:  git -C "$target" reset -q HEAD -- "$CHECKPOINT_PREFIX/" .claude/ plugins/ \
574:    $(pathspec_owned_excludes "$CHECKPOINT_PREFIX") 2>/dev/null || true
592:  for spec in "$CHECKPOINT_PREFIX/" .claude/ plugins/; do
599:      $(pathspec_owned_excludes "$CHECKPOINT_PREFIX") 2>/dev/null || true
648:abandon_instance() {
651:  restore_instance
890:# `git diff --cached --quiet || git diff --quiet` over the WHOLE tree: it takes
1040:  CHECKPOINT_PREFIX=""
1057:      abandon_instance "  ERROR: could not create the isolated dry-run hooks directory" && continue
1078:      abandon_instance "  ERROR: could not resolve production HEAD for dry-run model" && continue
1161:      abandon_instance "  ERROR: unsafe symlink prevents isolated dry-run modeling" && continue
1175:      abandon_instance "  ERROR: active merge or rebase cannot be modeled safely" && continue
1232:      abandon_instance "  ERROR: could not create disposable dry-run model" && continue
1238:        abandon_instance "  ERROR: could not isolate repository worktree config" && continue
1246:        abandon_instance "  ERROR: could not isolate linked-worktree config" && continue
1266:        abandon_instance "  ERROR: could not configure isolated direct-clone origin" && continue
1298:      if ! git diff --quiet -- "$sys_path" 2>/dev/null ||
1385:    # see restore_instance, which reduces its reset and checkout to this same
1388:          $(pathspec_owned_excludes "$prefix") 2>/dev/null ||
1389:        ! git diff --quiet -- "$prefix/" .claude/ plugins/ \
1390:          $(pathspec_owned_excludes "$prefix") 2>/dev/null; then
1392:        echo "  Changes vs skeleton: blocked by dirty working tree"
1395:      abandon_instance "  ERROR: dirty working tree; refusing to commit unrelated work" && continue
1404:      abandon_instance "  ERROR: could not checkpoint the instance; refusing to write" && continue
1412:        abandon_instance "  ERROR: fetch failed" && continue
1448:            $(pathspec_owned_excludes "$prefix") \
1451:          abandon_instance "  ERROR: preservation snapshot inventory failed; rsync not started" && continue
1469:          abandon_instance && continue
1480:          abandon_instance "  ERROR: preservation helper missing; rsync not started" && continue
1487:          abandon_instance "  ERROR: preservation helper failed; rsync not started" && continue
1520:          abandon_instance "  ERROR: preservation receipt incomplete or invalid; rsync not started" && continue
1531:          abandon_instance "  ERROR: preservation snapshot copy failed; rsync not started" && continue
1553:          abandon_instance "  ERROR: q-system sync would delete instance-owned data; refusing" && continue
1557:          abandon_instance "  ERROR: q-system sync failed" && continue
1565:          abandon_instance "  ERROR: preserved-file restore failed" && continue
1572:          abandon_instance "  ERROR: could not stage q-system sync" && continue
1578:            abandon_instance "  ERROR: could not commit q-system sync" && continue
1624:      abandon_instance && continue
1783:      abandon_instance "  ERROR: config sync did not reach a complete committed state" && continue
   500	  # CLEAN index and worktree, so the dirty-tree guard passes and the run
   501	  # proceeds with the founder's rebase still open. Aborting that would destroy
   502	  # their work AND rewind this run's own landed commit. Restore clears only
   503	  # what this run created.
   504	  : > "$CHECKPOINT_DIR/inflight"
   505	  if instance_rebase_in_flight "$target"; then
   506	    printf 'rebase\n' > "$CHECKPOINT_DIR/inflight"
   507	  fi
   508	  CHECKPOINT_TARGET="$target"
   509	}
   510	
   511	restore_instance() {
   512	  local target="${CHECKPOINT_TARGET:-}" uf
   513	  [ -n "$target" ] || return 0
   514	  [ -d "$CHECKPOINT_DIR" ] || return 0
   515	  # An interrupted rebase, first, because aborting one restores HEAD and the
   516	  # worktree wholesale and everything below should run on that result.
   517	  #
   518	  # The mixed reset further down already clears MERGE_HEAD, CHERRY_PICK_HEAD
   519	  # and REVERT_HEAD -- measured on git 2.54 across every conflicted state -- so
   520	  # none of those need handling. A rebase directory is the one that SURVIVES a
   521	  # reset.
   522	  #
   523	  # ONLY a rebase this run started. The checkpoint recorded whether one was
   524	  # already open, and if it was, it is the founder's and it stays. Assuming
   525	  # otherwise destroyed real work: the zombie-rebase cleanup above cannot see a
   526	  # rebase in a linked worktree (it tests "$path/.git/rebase-merge" as a
   527	  # directory, and there .git is a FILE), while a `rebase -i` paused at `edit`
   528	  # leaves a clean tree that the dirty-tree guard passes -- so the founder's
   529	  # open rebase reached this line untouched, and aborting it both discarded
   530	  # their work and rewound the sync commit this run had just landed.
   531	  #
   532	  # If the abort cannot run, say so and leave it. Deleting git's own state by
   533	  # hand is how a repo gets wrecked, and a human can finish what git could not.
   534	  if instance_rebase_in_flight "$target" &&
   535	      ! grep -qxF rebase "$CHECKPOINT_DIR/inflight" 2>/dev/null; then
   536	    git -C "$target" rebase --abort 2>/dev/null ||
   537	      echo "  WARN: a rebase this run started could not be aborted; the instance is left mid-rebase"
   538	  fi
   539	  # Untracked files the rsync --delete removed. No git verb can bring one back,
   540	  # and the ONLY copy is the preservation snapshot under $SNAP -- which lives
   541	  # inside ARCHIVE_TMP, so this has to happen before that is torn down. That
   542	  # ordering is why abandon_instance owns both steps.
   543	  if [ -n "${SNAP:-}" ] && [ -d "$SNAP/f" ] && [ -f "$SNAP/list" ]; then
   544	    ( cd "$target" && while IFS= read -r -d '' uf; do
   545	        if ! { [ -e "$uf" ] || [ -L "$uf" ]; } &&
   546	            { [ -e "$SNAP/f/$uf" ] || [ -L "$SNAP/f/$uf" ]; }; then
   547	          mkdir -p "$(dirname "$uf")" && cp -a "$SNAP/f/$uf" "$uf"
   548	        fi
   549	      done < "$SNAP/list" ) || true
   550	  fi
   551	  # Leave the instance CLEAN at whatever commit it reached -- do NOT rewind a
   552	  # commit that landed. A landed commit is not damage: the tree is clean and
   553	  # the next run proceeds normally. What actually took instances out of the
   554	  # fleet was UNCOMMITTED debris -- a half-staged index (sp-5f2d2a63) or
   555	  # modified tracked files (sp-e244e821) -- because the dirty-tree guard then
   556	  # read it as founder work and refused forever.
   557	  #
   558	  # Rewinding was tried and is wrong: test-kipi-update-hook-contract.sh sets
   559	  # HOOK_FAIL_ON=2 so the q-system commit lands and the CONFIG commit fails,
   560	  # and it asserts the first commit survives. Undoing it also strands the
   561	  # index against a HEAD it no longer matches, which manufactures the exact
   562	  # dirty tree this function exists to prevent.
   563	  # SCOPED to the sync's own write set (ASK-609), and this is load-bearing.
   564	  # `checkout -- .` discards every unstaged tracked modification in the repo.
   565	  # That was safe ONLY because the dirty-tree guard had just proved the whole
   566	  # tree clean, so there was nothing of the founder's to discard. Once that
   567	  # guard is scoped, an instance with founder edits outside the sync's reach
   568	  # passes it legitimately -- and an unscoped checkout here would then delete
   569	  # exactly the work the scoping was meant to stop blocking on. The two must
   570	  # move together; the pathspec below is the same one the guard uses.
   571	  git -C "$target" reset -q HEAD -- "$CHECKPOINT_PREFIX/" .claude/ plugins/ \
   572	    $(pathspec_owned_excludes "$CHECKPOINT_PREFIX") 2>/dev/null || true
   573	  git -C "$target" checkout -q -- "$CHECKPOINT_PREFIX/" .claude/ plugins/ \
   574	    $(pathspec_owned_excludes "$CHECKPOINT_PREFIX") 2>/dev/null || true
   575	  # Finally, remove files this run created: untracked NOW, absent from the
   576	  # checkpoint, and inside the sync's own scope. Never a recursive delete of a
   577	  # directory this run was not observed to create.
   578	  #
   579	  # The set difference is computed once in python rather than by forking a grep
   580	  # per file. The fork-per-file form was quadratic -- measured 83s on a 20k-file
   581	  # instance against 1s before -- and it also could not match a path containing
   582	  # a newline, because `grep -F` splits the PATTERN on newlines, so such a file
   583	  # was deleted despite being IN the checkpoint. Splitting on NUL fixes both.
   584	  checkpoint_untracked_list "$target" > "$CHECKPOINT_DIR/now" 2>/dev/null || true
   585	  python3 - "$CHECKPOINT_DIR/untracked" "$CHECKPOINT_DIR/now" "$target" <<'PY' || true
   586	import os
   587	import sys
   588	
   589	before = set(open(sys.argv[1], "rb").read().split(b"\0"))
   590	root = sys.argv[3]
   591	for record in open(sys.argv[2], "rb").read().split(b"\0"):
   592	    if not record or record in before:
   593	        continue
   594	    candidate = os.path.join(root, os.fsdecode(record))
   595	    try:
   596	        if os.path.islink(candidate) or os.path.isfile(candidate):
   597	            os.unlink(candidate)
   598	    except OSError:
   599	        pass
   600	PY
   601	}
   602	
   603	# The single give-up path. Every one of the 24 sites routes through here, so
   604	# restore-before-teardown is structural rather than 24 chances to forget it.
   605	# Two sites record a failure and deliberately FALL THROUGH: a direct-clone
   606	# whose merge needs manual resolve, and a failed archive export. Both still let
   607	# the .claude/ and plugins/ config sync run, so the instance keeps receiving
   608	# config updates even though its repo pull did not land. They must NOT abandon
   609	# the instance -- doing so also tears down the dry-run model the config sync is
   610	# still using. They get the counter alone, which is also what keeps the

codex
I’ve reached the verification pass. The main risk areas are rollback cleanliness, alert isolation, false delivery claims, and capability-gate false positives. I’m running the PR’s focused suites and separate `$TMPDIR` probes against the real scripts.
exec
/bin/zsh -lc 'python3 q-system/.q-system/scripts/test_capability_gate.py; python3 q-system/.q-system/scripts/test_launchd_health_check.py; python3 -m pytest -q q-system/.q-system/tests/test_fable_escalation.py q-system/.q-system/tests/test_alert_to_linear.py; bash q-system/.q-system/scripts/test/test-kipi-update-hook-contract.sh; bash test-kipi-update-dirty-guard-scope.sh' in /Users/assafkipnis/.config/kipi/review-trees/pr-145
 exited 1 in 149ms:
                     ~~~~~~~~~~~~~~~~~~~^^
    code = main()
    )
    ^
    raise exception
    teardown.throw(exception)
    ~~~~~~~~~~~~~~^^^^^^^^^^^
    config = yield
             ^^^^^
    self.parse(args)
    ~~~~~~~~~~^^^^^^
    )
    ^
    raise exception
    teardown.throw(exception)
    ~~~~~~~~~~~~~~^^^^^^^^^^^
    return (yield)
            ^^^^^
    ~~~~^^^^^^^^^^^^^^
                                              ~~~~~~~~~^^^
    TemporaryFile(buffering=0),
    ~~~~~~~~~~~~~^^^^^^^^^^^^^
    dir = gettempdir()
                        ~~~~~~~~~~~^^
                            dirlist)
FileNotFoundError: [Errno 2] No usable temporary directory found in ['/var/folders/th/ty0kb7r56dvfrkgkrcbx9fv00000gn/T/', '/tmp', '/var/tmp', '/usr/tmp', '/Users/assafkipnis/.config/kipi/review-trees/pr-145']
mktemp: mkdtemp failed on /var/folders/th/ty0kb7r56dvfrkgkrcbx9fv00000gn/T/tmp.C2tyw2EoMi: Operation not permitted
mktemp: mkdtemp failed on /var/folders/th/ty0kb7r56dvfrkgkrcbx9fv00000gn/T/tmp.QifkRuXnEc: Operation not permitted
=== dirt OUTSIDE the sync's write set (q-investigate/) -> must SYNC ===
mkdir: /out: Operation not permitted
cp: /out/skeleton/q-system: No such file or directory
chmod: /out/skeleton/kipi-update.sh: No such file or directory
test-kipi-update-dirty-guard-scope.sh: line 50: /out/skeleton/q-system/.q-system/scripts/skel-tool.py: No such file or directory
test-kipi-update-dirty-guard-scope.sh: line 52: /out/skeleton/instance-registry.json: No such file or directory
mkdir: /out: Operation not permitted
mkdir: /out: Operation not permitted
test-kipi-update-dirty-guard-scope.sh: line 68: /out/instance/q-system/.q-system/scripts/skel-tool.py: No such file or directory
test-kipi-update-dirty-guard-scope.sh: line 69: /out/instance/q-system/CLAUDE.md: No such file or directory
test-kipi-update-dirty-guard-scope.sh: line 70: /out/instance/q-investigate/evidence/case-001.md: No such file or directory
test-kipi-update-dirty-guard-scope.sh: line 75: /out/instance/q-investigate/evidence/case-001.md: No such file or directory
test-kipi-update-dirty-guard-scope.sh: line 96: /out/skeleton/../run.log: No such file or directory
grep: /out/skeleton/../run.log: No such file or directory
  FAIL [outside]: never reached the instance; measuring nothing
tail: /out/skeleton/../run.log: No such file or directory

=== dirt INSIDE the sync's write set (q-system/CLAUDE.md) -> must BLOCK ===
mkdir: /in: Operation not permitted
cp: /in/skeleton/q-system: No such file or directory
chmod: /in/skeleton/kipi-update.sh: No such file or directory
test-kipi-update-dirty-guard-scope.sh: line 50: /in/skeleton/q-system/.q-system/scripts/skel-tool.py: No such file or directory
test-kipi-update-dirty-guard-scope.sh: line 52: /in/skeleton/instance-registry.json: No such file or directory
mkdir: /in: Operation not permitted
mkdir: /in: Operation not permitted
test-kipi-update-dirty-guard-scope.sh: line 68: /in/instance/q-system/.q-system/scripts/skel-tool.py: No such file or directory
test-kipi-update-dirty-guard-scope.sh: line 69: /in/instance/q-system/CLAUDE.md: No such file or directory
test-kipi-update-dirty-guard-scope.sh: line 70: /in/instance/q-investigate/evidence/case-001.md: No such file or directory
test-kipi-update-dirty-guard-scope.sh: line 85: /in/instance/q-system/CLAUDE.md: No such file or directory
test-kipi-update-dirty-guard-scope.sh: line 96: /in/skeleton/../run.log: No such file or directory
grep: /in/skeleton/../run.log: No such file or directory
  FAIL [inside]: never reached the instance; measuring nothing
tail: /in/skeleton/../run.log: No such file or directory

=== guard and restore must share one scope ===
  OK [scope-pairing]: guard and restore both pathspec-limited

FAIL (2)

exec
/bin/zsh -lc "rg -n -U '"'^'"\\s*if\\s+__name__\\s*==\\s*[\"'\"']__main__[\"'\"']\\s*:' --glob '*.py' . | head -80; find . -name conftest.py -type f -perm -111 -print; python3 -c 'import importlib.util, pathlib; p=\"q-system/.q-system/scripts/capability-gate.py\"; s=importlib.util.spec_from_file_location(\"gate\",p); m=importlib.util.module_from_spec(s); s.loader.exec_module(m); text=\"\\\"\\\"\\\"Usage:\\n    if __name__ == \\\"__main__\\\":\\n        main()\\n\\\"\\\"\\\"\\ndef helper():\\n    return 1\\n\"; print(\"runnable=\",m.is_runnable_engine(pathlib.Path(\"helper.py\"),text)); print(\"regex_match=\",repr(m._MAIN_GUARD_RE.search(text).group(0)))'" in /Users/assafkipnis/.config/kipi/review-trees/pr-145
 exited 1 in 0ms:
zsh:1: unmatched '

exec
/bin/zsh -lc "rg -n -U '"'^'"\\s*if\\s+__name__\\s*==\\s*[\"'\"']__main__[\"'\"']\\s*:' --glob '*.py' . | head -80
find . -name conftest.py -type f -perm -111 -print
python3 - <<'PY'
import importlib.util
import pathlib
p = 'q-system/.q-system/scripts/capability-gate.py'
s = importlib.util.spec_from_file_location('gate', p)
m = importlib.util.module_from_spec(s)
s.loader.exec_module(m)
text = '''\\\"\\\"\\\"Usage:
    if __name__ == \"__main__\":
        main()
\\\"\\\"\\\"
def helper():
    return 1
'''
print('runnable=', m.is_runnable_engine(pathlib.Path('helper.py'), text))
match = m._MAIN_GUARD_RE.search(text)
print('regex_match=', repr(match.group(0) if match else None))
PY" in /Users/assafkipnis/.config/kipi/review-trees/pr-145
 exited 1 in 0ms:
zsh:20: unmatched '

exec
import importlib.util
import pathlib
p = 'q-system/.q-system/scripts/capability-gate.py'
s = importlib.util.spec_from_file_location('gate', p)
m = importlib.util.module_from_spec(s)
s.loader.exec_module(m)
text = '''\\\"\\\"\\\"Usage:
    if __name__ == \"__main__\":
        main()
\\\"\\\"\\\"
def helper():
    return 1
'''
print('runnable=', m.is_runnable_engine(pathlib.Path('helper.py'), text))
match = m._MAIN_GUARD_RE.search(text)
print('regex_match=', repr(match.group(0) if match else None))
PY
find . -name conftest.py -type f -perm -111 -print" in /Users/assafkipnis/.config/kipi/review-trees/pr-145
 succeeded in 0ms:

exec
/bin/zsh -lc "python3 -c \"import importlib.util,pathlib; p='q-system/.q-system/scripts/capability-gate.py'; s=importlib.util.spec_from_file_location('gate',p); m=importlib.util.module_from_spec(s); s.loader.exec_module(m); t='\\\"\\\"\\\"Usage:\\n    if __name__ == \\\"__main__\\\":\\n        main()\\n\\\"\\\"\\\"\\ndef helper():\\n    return 1\\n'; x=m._MAIN_GUARD_RE.search(t); print('runnable=',m.is_runnable_engine(pathlib.Path('helper.py'),t)); print('regex_match=',repr(x.group(0) if x else None))\"; find . -name conftest.py -type f -perm -111 -print" in /Users/assafkipnis/.config/kipi/review-trees/pr-145
 succeeded in 0ms:
runnable= True
regex_match= '    if __name__ == "__main__":'

codex
One finding survived reproduction: the new “syntax” regex still treats a docstring line as executable code. The broader suites could not create temporary files under this review sandbox, so I’m not treating those environment errors as PR defects. Read-only syntax checks and diff hygiene passed.
exec
/bin/zsh -lc "git status --short; git diff --check origin/main...HEAD; bash -n kipi-update.sh q-system/.q-system/scripts/test/verify-alert-wiring.sh q-system/.q-system/scripts/test/test-kipi-update-hook-contract.sh test-kipi-update-dirty-guard-scope.sh; PYTHONDONTWRITEBYTECODE=1 python3 -c \"import ast,pathlib; fs=['q-system/.q-system/scripts/capability-gate.py','q-system/.q-system/scripts/client-name-guard.py','q-system/.q-system/scripts/fable-escalate.py','q-system/.q-system/scripts/launchd-health-check.py']; [ast.parse(pathlib.Path(f).read_text(),filename=f) for f in fs]; print('python AST parse: PASS')\"" in /Users/assafkipnis/.config/kipi/review-trees/pr-145
 succeeded in 0ms:
python AST parse: PASS

codex
### Finding

**MINOR, Confirmed:** The new regex still classifies a library as runnable when its docstring contains a standalone `if __name__ == "__main__":` example. This can produce a false inert-engine alert. Severity is minor because no current producer has this shape.

[capability-gate.py:504](/Users/assafkipnis/.config/kipi/review-trees/pr-145/q-system/.q-system/scripts/capability-gate.py:504)

Reproducer:

```bash
python3 -c "import importlib.util,pathlib; p='q-system/.q-system/scripts/capability-gate.py'; s=importlib.util.spec_from_file_location('gate',p); m=importlib.util.module_from_spec(s); s.loader.exec_module(m); t='\"\"\"Usage:\n    if __name__ == \"__main__\":\n        main()\n\"\"\"\ndef helper():\n    return 1\n'; x=m._MAIN_GUARD_RE.search(t); print('runnable=',m.is_runnable_engine(pathlib.Path('helper.py'),t)); print('regex_match=',repr(x.group(0) if x else None))"
```

Real output:

```text
runnable= True
regex_match= '    if __name__ == "__main__":'
```

The implementation needs AST-based detection of a top-level executable guard. Line-oriented regex cannot distinguish code from docstring content.

### What is sound

- Rollback scope: the updater now filters unmatched checkout pathspecs while preserving the same guarded write set.
- Alert isolation: capture writes happen before Linear access, and unwritable captures return a non-delivery result.
- Test fixture shape: the alert fingerprint fixtures preserve path-shaped tokens while removing the private username.
- Concurrency: the escalation ledger retains its exclusive append lock.
- Hygiene checks passed:
  - `git diff --check origin/main...HEAD`
  - `bash -n` on all changed shell scripts
  - Python AST parsing on all changed production Python files
- The focused suites could not create `$TMPDIR` files because this review sandbox is read-only. Their environment errors are not PR findings.
- The repository remained unchanged.

Fix first: replace `_MAIN_GUARD_RE` with AST-based top-level guard detection.

FINDINGS:
minor|The runnable-engine regex mistakes a standalone __main__ example inside a docstring for executable code and can emit a false inert-engine alert|q-system/.q-system/scripts/capability-gate.py:504
END FINDINGS
tokens used
81,058
### Finding

**MINOR, Confirmed:** The new regex still classifies a library as runnable when its docstring contains a standalone `if __name__ == "__main__":` example. This can produce a false inert-engine alert. Severity is minor because no current producer has this shape.

[capability-gate.py:504](/Users/assafkipnis/.config/kipi/review-trees/pr-145/q-system/.q-system/scripts/capability-gate.py:504)

Reproducer:

```bash
python3 -c "import importlib.util,pathlib; p='q-system/.q-system/scripts/capability-gate.py'; s=importlib.util.spec_from_file_location('gate',p); m=importlib.util.module_from_spec(s); s.loader.exec_module(m); t='\"\"\"Usage:\n    if __name__ == \"__main__\":\n        main()\n\"\"\"\ndef helper():\n    return 1\n'; x=m._MAIN_GUARD_RE.search(t); print('runnable=',m.is_runnable_engine(pathlib.Path('helper.py'),t)); print('regex_match=',repr(x.group(0) if x else None))"
```

Real output:

```text
runnable= True
regex_match= '    if __name__ == "__main__":'
```

The implementation needs AST-based detection of a top-level executable guard. Line-oriented regex cannot distinguish code from docstring content.

### What is sound

- Rollback scope: the updater now filters unmatched checkout pathspecs while preserving the same guarded write set.
- Alert isolation: capture writes happen before Linear access, and unwritable captures return a non-delivery result.
- Test fixture shape: the alert fingerprint fixtures preserve path-shaped tokens while removing the private username.
- Concurrency: the escalation ledger retains its exclusive append lock.
- Hygiene checks passed:
  - `git diff --check origin/main...HEAD`
  - `bash -n` on all changed shell scripts
  - Python AST parsing on all changed production Python files
- The focused suites could not create `$TMPDIR` files because this review sandbox is read-only. Their environment errors are not PR findings.
- The repository remained unchanged.

Fix first: replace `_MAIN_GUARD_RE` with AST-based top-level guard detection.

FINDINGS:
minor|The runnable-engine regex mistakes a standalone __main__ example inside a docstring for executable code and can emit a false inert-engine alert|q-system/.q-system/scripts/capability-gate.py:504
END FINDINGS

PR #149, opened 2026-08-14
Refuse to fan .claude/ and plugins/ bytes nobody reviewed (ASK-762)

assafkip:
## The hole

`kipi-update.sh` copies `q-system/` with `git archive` from HEAD, and two preflights already refuse when the index and HEAD disagree.

`.claude/` and `plugins/` got neither. They rsync from `$SCRIPT_DIR` — the skeleton's **working tree**, on whatever branch it is checked out on. `SKELETON_BRANCH="main"` was read only by the direct-clone pull path.

## Measured, not theorized (2026-08-14, sp-ea9c1628)

kipi-system was sitting on `sana/ask-728-plugin-parity` holding an uncommitted partial forward-port of `plugins/kipi-core/voicekit/selector.py`: the nearest-length ranking **without** the anchor-survives fix Codex caught in #147 and main already carried.

A fleet-update run from that state writes code strictly older than main into every config-sync instance and prints `PASS`. Silent, plausible, 23 repos wide.

## The guard

Two halves, deliberately different in what they need.

**DIRTY — always armed.** Staleness against HEAD needs no remote to be wrong. It asks for untracked files by name, because those rsync too and `git diff` cannot see them. Modified, staged, and untracked all abort.

**BRANCH — armed only when an `origin` remote exists.** `SKELETON_BRANCH` names a branch *on* origin; a repo without one has no main to be stale against, and every existing updater fixture is exactly that repo. It prints the disarm rather than going quiet — a silent guard is indistinguishable from one that passed.

Both abort before any instance is read or written, and name what was wrong.

## Verification

- New fixture `test-kipi-update-source-provenance.sh` is **RED without the guard**: with a dirty `plugins/`, the run entered the instance loop and synced. Green with it.
- All 12 updater/rollback fixtures pass.
- `validate-separation.py` 25/25.
- Declared in `capability-manifest.json` (entry list + `skeleton_only`).

## One fixture changed, not weakened

`test-kipi-update-dry-final-state.sh` committed only `q-system/`, leaving `.claude/` and `plugins/` untracked where the real skeleton tracks both — a repo the updater will not run on. The guard exposed that. The fixture now commits all three.

## Not in scope

`sp-10cf4f76` (rsync `--delete` removing tracked instance-only scripts) is a different defect in the same file and stays open.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_0151EYtGH6at3XTWqbBHAxEi

PR #150, opened 2026-08-14
Rescue the open-loops spillover-capture fix an auto-commit stranded (ASK-763)

assafkip:
## Why this is a rescue

An unattended auto-commit wrote `421d78ac` onto the **local** `sana/ask-728-plugin-parity` checkout during the 2026-08-14 session and never pushed it. That branch is **15 commits behind origin and 2 ahead**, so the work existed only on this machine.

It also carried no issue id — which the `linear-issue-ref` commit-msg hook exists to refuse. So the ledger had no record of it either.

The other stranded commit, `eb25f466` (ASK-746 digest plist), turned out **byte-identical** to what main already carries. Redundant, dropped. This one is not on main in any form.

## What it fixes

A `deferred` disposition auto-creates a spillover item (`no-orphan-findings.md`). The open-loops catch-all was not reading that ledger, so it counted already-tracked findings and reported them to the founder as untracked.

`ledger_root()` resolves the ONE shared ledger via `git rev-parse --git-common-dir`, mirroring `prd_runner.py`. A per-worktree path would make this script blind to captures the gate can see — the same load-path mistake as the marketplace clone. It fails open to `repo_root`, because a missed skip only over-surfaces and never drops.

## Verification

Run on **this** branch, off current main — not on the stale branch it was stranded on:

- `test-open-loops.sh` is **RED** without the `open-loops.py` half: `expected exactly 1 uncaptured finding in the catch-all (spillover-captured one must be excluded)`.
- Green with it.

## Note for triage

ASK-759, ASK-760, ASK-763, ASK-765 and ASK-767 all describe overlapping slices of this same open-loops/spillover behavior. This PR claims ASK-763, the exact match. The duplication is worth a pass but is not touched here.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_0151EYtGH6at3XTWqbBHAxEi

PR #148, opened 2026-08-14
Refuse the engagement root, opt in the first two repos, and stop the preflight inventing reasons (ASK-754, ASK-755)

assafkip:
Two issues, one branch. ASK-754 was the opt-in half and is unchanged. ASK-755 is everything after `8a64c449`.

## ASK-754 (unchanged, rounds 1-4)

The engagement ROOT itself is refused, not only what nests under it, and the first two non-client repos are opted into dispatch.

## ASK-755: clearing the four preflight blockers

Round 4's finding was that both opted-in repos are guaranteed to be refused on every dispatch. Three of the four causes are now gone. The fourth is real, is not code, and is ASK-757.

**1. control-code (cleared).** Both instances were exactly one release behind. Cleared with a new bounded propagator rather than the fleet updater: `control-file-propagate.py` classifies one FILE in one repo NEW/OLD/OTHER/DIRTY/MISSING against skeleton history and writes only OLD. Copy only, no delete. The preflight's own refusal now prints that command instead of naming the fleet rsync, which has a delete flag and a blast radius three orders of magnitude wider than the defect.

Its test asserts the REFUSALS first, by the target's bytes being unchanged on disk rather than by the word REFUSED appearing. 13/13; mutation (classify returns OLD for everything, DIRTY branch unreachable) takes it to 7/13.

**2. hooks (was a defect in this script).** The check compared an instance against the skeleton's RUNTIME `.claude/settings.json`. Instances are built from `settings-template.json`, and the two differ deliberately: `settings-template-sync-check.py` is `SKELETON_ONLY` by its own design. So every instance was permanently missing the one hook that can never ship, and no updater run could have cleared it.

Measured, both repos: against the runtime file, missing that one guard; against the template, FULL parity. Not a weakening -- both guard sets are 41, the switch drops one unsatisfiable requirement and adds one real one (`instance-automation-guard.py`, FLEET_ONLY), and the event sets are identical. The fixture was the other half: it cloned the runtime settings, so the control repo was easier to satisfy than any real instance. It is now built from the template.

**3. dirty (cleared, in the instances).** Verified as machine exhaust, not WIP: the same two files were dirty in both repos with identical sha256, the content was a fanned-out prd-os bump citing skeleton issue ids, and one repo already carried two prior commits recording the same fanout. The untracked half was generated state the skeleton's own `.gitignore` already covers and the instances' copies predate. Committed path-scoped in the instances; nothing in this diff.

[...]

PR #146, opened 2026-08-14
Scope every gh call to the target repo, and let dispatch enter more than one (ASK-738)

assafkip:
## The defect, measured

`gh` resolves its repository from the **process working directory** and ignores every path variable this fleet carries. Measured 2026-08-13:

```
cwd = kipi-system checkout, TARGET_REPO=<other>  ->  assafkip/kipi-system
cwd = <other>                                    ->  assafkip/<other>
```

`kipi-dispatch.sh:205` does `cd "$REPO"` and never leaves. So `git -C` did the work in the target while `gh` answered about the home repo.

## The sharpest one: the wrong-repo failure ran through the gate itself

`gh api` **takes no `-R`**. It expands `{owner}/{repo}` from the cwd-resolved repo, and there is no flag to redirect it.

That is the call at `pr-review-agent.sh` `post_status()` which posts `kipi/reviewer-approved` — the **required check**. So the failure was not "a review looked at the wrong diff". It was: review the home repo's code, then post a green required check naming the home repo, for a PR in a different repository. The gate that exists to catch wrong code was itself the thing acting on the wrong repo, and every log line read normally.

Fixed by substituting the slug into the path (`repos/$STATUS_REPO_PATH/statuses/$sha`) rather than by a flag that does not exist.

## Three red-then-green reproducers, one per call site

Each drives the real script and asserts **which repository the call resolved to** — not an argv shape. Each carries a negative self-test proving the fake `gh` reproduces gh's own cwd binding first; without it, a fake that logged nothing would make every assertion vacuously true.

| Call site | Red output at `621a5b43` |
|---|---|
| `linear-worker.sh:902` | `pr list --head sana/ask-aaa` resolved to `assafkip/homerepo` while `--repo` pointed at `assafkip/targetrepo` |
| `converge.sh:68` | reported `pr=42` — the **home** repo's PR, for work in the target |
| `pr-review-agent.sh` | pinned the home repo's head for a review of the target's PR #42 (#42 exists in both, at different shas) |
| artifacts | two reviews of PR #42 in two repos left **one** record, `pr-42.verdict.json` |

## One derivation, one place

`repo-slug-lib.sh`. `sp-421fa27d` already records a duplicated project-name derivation here; a second copy of "which repository may an unattended self-merging agent act on" is not a style problem.

[...]

PR #153, opened 2026-08-14
One exclusion set, so a plugin .env stops shipping fleet-wide (ASK-772)

assafkip:
## A plugin-root `.env` was copied into all 23 instances and committed there

The per-plugin rsync excluded five classes and nothing else. `.gitignore:3` is `*.env`, and `plugins/kipi-design/skills/design/scripts/cip/generate.py` reads a plugin-root `.env` for API keys — so the file was invisible to `git status` and copied happily by rsync. Found by Codex on PR #149 round 3.

## Why fixing one place would have made it worse

The exclusion set was written down **twice** — rsync `--exclude` flags and the ASK-762 preflight's grep filter. They answer halves of one question:

- rsync excludes → *never copy this*
- preflight filter → *never alarm about this, because it is never copied*

Fix the rsync alone and the preflight aborts every fleet update over a file that can no longer leak — denial of service dressed as security. Fix the preflight alone and the leak stays open and silent.

`PLUGIN_COPY_EXCLUDES` is now the single source, with `.env` and `.env.*` added. Each entry carries the rsync glob and an equivalent regex adjacently; the two are not interconvertible in bash without a converter more fragile than the duplication it removes.

## There was a third consumer, and only a behavioural test could have found it

The stager's `find` had its own hardcoded prune list. Excluding `.env` from the copy without teaching the stager left it naming a path the syncer no longer writes:

```
fatal: pathspec 'plugins/demo/.env.local' did not match any files
ERROR: config sync did not reach a complete committed state
```

A test comparing the two lists as strings would have passed. This one drives **both behaviours** off the shared set and caught the third site.

## One assertion is deliberately reversed

`test-kipi-update-source-provenance.sh` asserted that an ignored `.env` **aborts**. That was correct while the copy carried it. Now it cannot, and the same contract requires **silence**.

This is not the guard weakening. The leak moved from *detected* to *impossible*, which is strictly better, and the new fixture is what holds the impossibility. If it ever goes green while a `.env` lands in an instance, that assertion is wrong again and must flip back. The comment says so in the file.

## Verification

- Reproducer is **RED** without this change — and not vacuously: it asserts the sync actually *ran*, which is what surfaced that today's "safety" is just the preflight halting the fleet.
- Property 3 keeps the guard armed: an untracked plugin file that **is** still copied must still abort, so property 2 can't be satisfied by disarming.
- All 13 updater/rollback fixtures pass. `validate-separation.py` 25/25. Declared in `capability-manifest.json`.

## Sequencing

Branched off main alongside #151, which also edits `kipi-update.sh` in a different region (the system-state carve-out). #151 is the root-cause fix for the blocked fleet; merge it first and I'll rebase this.

### DIFF
--- a/kipi-update.sh
+++ b/kipi-update.sh
@@ -198,6 +198,55 @@ is_managed_plugin_path() {
 # plugins/ tree an EARLIER update staged and never committed). The fleet updater
 # was blocked by itself, silently. This list is deliberately narrow and explicit:
 # anything not named here is treated as founder work and still refuses.
+# THE ONE plugin-copy exclusion set (ASK-772). Two consumers derive from it and
+# they answer halves of a single question:
+#
+#   the per-plugin rsync         -- "never copy this into an instance"
+#   the source-provenance preflight -- "never alarm about this, because it is
+#                                       never copied"
+#
+# Writing the set down twice is how a secret shipped. `.gitignore:3` is `*.env`
+# and kipi-design's cip/generate.py reads a plugin-root .env for API keys, so a
+# gitignored plugins/<name>/.env was invisible to `git status` and copied happily
+# by rsync into all 23 instances (Codex, PR #149 round 3).
+#
+# Adding it to ONE consumer swaps one failure for another: rsync-only leaves the
+# preflight aborting the whole fleet over a file that can no longer leak (a
+# denial of service on every update), preflight-only leaves the leak open and
+# silent. Hence one list.
+#
+# Each entry is "<rsync --exclude pattern>::<regex matching the same paths>".
+# Two representations, adjacent, because rsync globs and grep regexes are not
+# interconvertible in bash without a converter more fragile than the duplication
+# it removes. `::` is the delimiter, not `|`, because the regexes contain `|`.
+# test-kipi-update-plugin-excludes.sh drives BOTH consumers off this list and
+# fails if they disagree, so agreement is proven behaviourally rather than by
+# the two strings having been typed the same.
+PLUGIN_COPY_EXCLUDES=(
+  "/.git/::(^|/)\.git/"
+  "__pycache__/::(^|/)__pycache__/"
+  "*.pyc::\.pyc\$"
+  ".venv/::(^|/)\.venv/"
+  ".pytest_cache/::(^|/)\.pytest_cache/"
+  ".env::(^|/)\.env\$"
+  ".env.*::(^|/)\.env\."
+)
+
+plugin_copy_rsync_flags() {
+  local entry
+  for entry in "${PLUGIN_COPY_EXCLUDES[@]}"; do
+    printf -- '--exclude=%s\n' "${entry%%::*}"
+  done
+}
+
+plugin_copy_filter_regex() {
+  local entry out=""
+  for entry in "${PLUGIN_COPY_EXCLUDES[@]}"; do
+    out="${out:+$out|}${entry#*::}"
+  done
+  printf '%s\n' "$out"
+}
+
 SYSTEM_OWNED_PATHS=(
   "q-system/memory/.sycophancy-monthly-stamp"
   "q-system/.q-system/claude-integrity-baseline.json"
@@ -448,7 +497,7 @@ SYNC_SCOPE_DIRTY="$(
     git -C "$SCRIPT_DIR" status --porcelain --ignored=matching -- plugins \
       2>/dev/null || true
   } | sed 's/^...//' \
-    | grep -vE '(^|/)(\.git|__pycache__|\.venv|\.pytest_cache)/|\.pyc$' || true
+    | grep -vE "$(plugin_copy_filter_regex)" || true
 )"
 if [ -n "$SYNC_SCOPE_DIRTY" ]; then
   echo ""
@@ -895,14 +944,36 @@ stage_config_sync() {
     # syncer now walk the same list, so the stager can no longer name a path
     # the syncer will not write -- which is what the old post-hoc [ -d ] filter
     # was patching around. See managed_plugin_names for the scar.
+    # THE THIRD CONSUMER of PLUGIN_COPY_EXCLUDES, and the one that made the
+    # duplication expensive rather than merely untidy (ASK-772). This walk had
+    # its OWN hardcoded prune list -- .git, __pycache__, .pytest_cache, .venv,
+    # *.pyc -- a third copy of the same value. Excluding `.env` from the rsync
+    # without teaching the stager left it naming a path the syncer no longer
+    # writes, and `git add` on a missing path is fatal:
+    #
+    #     fatal: pathspec 'plugins/demo/.env.local' did not match any files
+    #     ERROR: config sync did not reach a complete committed state
+    #
+    # which is the exact class the comment above already warns about, arriving
+    # from the one direction it did not cover. The find still prunes cheaply so
+    # it never descends into a .venv; the shared regex is what decides.
+    plugin_excl_re="$(plugin_copy_filter_regex)"
     while IFS= read -r -d '' plugin_name; do
       while IFS= read -r -d '' source; do
-        plugin_paths+=("${source#"$SCRIPT_DIR/"}")
+        relative="${source#"$SCRIPT_DIR/"}"
+        # Matched on the path RELATIVE to the plugin dir, the same thing rsync
+        # matches its --exclude patterns against. Matching the repo-relative path
+        # would let a `.env` anywhere above the plugin root change the answer.
+        case "${source#"$SKELETON_PLUGIN_ROOT/$plugin_name/"}" in
+          *) printf '%s' "${source#"$SKELETON_PLUGIN_ROOT/$plugin_name/"}" \
+               | grep -qE "$plugin_excl_re" && continue ;;
+        esac
+        plugin_paths+=("$relative")
       done < <(
         find "$SKELETON_PLUGIN_ROOT/$plugin_name" \
           \( -type d -name .git -o -type d -name __pycache__ \
              -o -type d -name .pytest_cache -o -type d -name .venv \) -prune -o \
-          \( -type f ! -name '*.pyc' -o -type l \) -print0
+          \( -type f -o -type l \) -print0
       )
     done < <(managed_plugin_names)
     # A path the INSTANCE ignores cannot be staged, and `git add` treats that
@@ -1888,9 +1959,14 @@ print("\n".join(mod.EXTRA_WATCHED))
         # It was 107MB of the 112MB plugin tree, copied into 23 instances
         # where it could never work. --delete-excluded also clears the stale
         # copies already there. Pairs with test-kipi-update-build-artifacts.sh.
+        # Flags derived from PLUGIN_COPY_EXCLUDES, never hand-listed here: this
+        # site and the preflight filter are the two consumers that must agree,
+        # and a hand-listed copy is how a plugin-root .env shipped fleet-wide
+        # (ASK-772). Word-splitting the generator output is intended -- every
+        # pattern is a shell-safe token by construction.
+        # shellcheck disable=SC2046
         if ! rsync -a --delete --delete-excluded \
-            --exclude="/.git/" --exclude="__pycache__/" --exclude="*.pyc" \
-            --exclude=".venv/" --exclude=".pytest_cache/" \
+            $(plugin_copy_rsync_flags) \
             "$SKELETON_PLUGIN_ROOT/$plugin_name/" \
             "$path/plugins/$plugin_name/" 2>/dev/null; then
           CONFIG_FAILED=1
--- a/q-system/.q-system/capability-manifest.json
+++ b/q-system/.q-system/capability-manifest.json
@@ -151,6 +151,10 @@
       "timeout_s": 300
     },
     {
+      "path": "q-system/.q-system/scripts/test/test-kipi-update-plugin-excludes.sh",
+      "runner": "bash"
+    },
+    {
       "path": "q-system/.q-system/scripts/test/test-kipi-update-preservation-failure.sh",
       "runner": "bash"
     },
@@ -602,6 +606,7 @@
     "q-system/.q-system/scripts/test/test-kipi-update-dry-final-state.sh",
     "q-system/.q-system/scripts/test/test-kipi-update-hook-contract.sh",
     "q-system/.q-system/scripts/test/test-kipi-update-leak-preflight.sh",
+    "q-system/.q-system/scripts/test/test-kipi-update-plugin-excludes.sh",
     "q-system/.q-system/scripts/test/test-kipi-update-preservation-failure.sh",
     "q-system/.q-system/scripts/test/test-kipi-update-safety.sh",
     "q-system/.q-system/scripts/test/test-kipi-update-source-provenance.sh",
@@ -686,7 +691,7 @@
     },
     {
       "path": "q-system/.q-system/scripts/stat-verify.py",
-      "reason": "802-line stat verification engine; zero hook wiring and its stat-registry.json data file exists in exactly one instance — wiring it is a founder product decision",
+      "reason": "802-line stat verification engine; zero hook wiring and its stat-registry.json data file exists in exactly one instance \u2014 wiring it is a founder product decision",
       "spillover_id": "sp-72b60bff"
     },
     {
@@ -706,8 +711,8 @@
     }
   ],
   "uncovered_known": [
-    "plugins/*/tests — prd-os plugin suite, run by the plugin's own harness; outside v1 scan roots (finding-9 deferred)",
+    "plugins/*/tests \u2014 prd-os plugin suite, run by the plugin's own harness; outside v1 scan roots (finding-9 deferred)",
     "repo-root *.sh utilities are wiring surfaces, not test artifacts",
-    "stat-registry.json: NOT expressible as required_data in v1 — the one instance holding it keeps canonical under its own instance_q_dir (not q-system/) and its registry name differs from its dir basename, so a scoped entry could never fire (grounded 2026-07-23). Declared gap; owning decision + instance detail in spillover sp-72b60bff"
+    "stat-registry.json: NOT expressible as required_data in v1 \u2014 the one instance holding it keeps canonical under its own instance_q_dir (not q-system/) and its registry name differs from its dir basename, so a scoped entry could never fire (grounded 2026-07-23). Declared gap; owning decision + instance detail in spillover sp-72b60bff"
   ]
 }
--- a/q-system/.q-system/scripts/test/test-kipi-update-plugin-excludes.sh
+++ b/q-system/.q-system/scripts/test/test-kipi-update-plugin-excludes.sh
@@ -1,0 +1,161 @@
+#!/usr/bin/env bash
+# A plugin-root .env is copied into all 23 instances and committed there.
+#
+# The per-plugin rsync excludes exactly five classes -- .git/, __pycache__/,
+# *.pyc, .venv/, .pytest_cache/ -- and nothing else. `.gitignore:3` is `*.env`,
+# so a plugin .env is invisible to `git status` while rsync copies it happily.
+# It is not hypothetical: plugins/kipi-design/skills/design/scripts/cip/
+# generate.py reads a plugin-root .env for API keys.
+#
+# Found by Codex on PR #149 round 3, filed as ASK-772.
+#
+# THE DRIFT TRAP THIS FILE EXISTS TO CLOSE. The exclusion set was written down
+# TWICE: once as rsync --exclude flags (~L1892) and once as a grep filter in the
+# ASK-762 source-provenance preflight (~L451). The two must agree, because they
+# answer halves of one question:
+#
+#   rsync excludes  = "never copy this"
+#   preflight filter = "do not alarm about this, because it is never copied"
+#
+# Add `.env` to the rsync only and the preflight starts aborting the fleet over a
+# file that can no longer leak. Add it to the preflight only and the leak stays
+# open and silent. One value, two authorities, opposite failure modes.
+#
+# So this file does NOT compare the two lists as strings -- that only proves they
+# were typed the same. It drives BOTH behaviours from the shared set and asserts
+# each entry is excluded by the copy AND silent in the preflight. A new entry
+# added to one consumer and not the other fails here.
+set -euo pipefail
+ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../../../.." && pwd)"
+fail() { echo "FAIL: $1" >&2; exit 1; }
+G() { git -c user.email=t@t.t -c user.name=test -c commit.gpgsign=false "$@"; }
+
+build() {
+  local work="$1" sk="$work/skel" inst="$work/inst"
+  mkdir -p "$sk/q-system/.q-system/scripts" "$sk/q-system/.q-system/state" \
+           "$sk/q-system/hooks" "$sk/plugins/demo" "$sk/.claude/rules"
+  for f in kipi-update.sh kipi-update-preserve-scan.py kipi-update-deletion-guard.py \
+           validate-separation.py; do
+    cp "$ROOT/$f" "$sk/$f"
+  done
+  cp "$ROOT/q-system/.q-system/scripts/propagation-leak-gate.py" \
+     "$sk/q-system/.q-system/scripts/propagation-leak-gate.py"
+  cp "$ROOT/q-system/.q-system/scripts/containment-targets.py" \
+     "$sk/q-system/.q-system/scripts/containment-targets.py"
+  cp "$ROOT/q-system/hooks/auto-commit.py" "$sk/q-system/hooks/auto-commit.py"
+  cat > "$sk/q-system/.q-system/state/propagation-leak-baseline.json" <<'JSON'
+{
+  "schema_version": 1,
+  "blocking_classes": ["case_proof_gap", "client_identity", "dated_interaction",
+                       "pricing", "relationship", "source_identity",
+                       "sourced_interaction"],
+  "classifier_sha256": null,
+  "entries": []
+}
+JSON
+  printf 'generic skeleton content\n' > "$sk/q-system/tracked.md"
+  printf 'plugin content\n' > "$sk/plugins/demo/content.txt"
+  printf '# demo rule\n' > "$sk/.claude/rules/demo.md"
+  # The skeleton ignores secrets and caches, exactly as the real repo does. That
+  # is the whole point: gitignored is not "absent", it is "invisible to git and
+  # fully visible to rsync".
+  printf '*.env\n.env\n__pycache__/\n*.pyc\n.venv/\n.pytest_cache/\n' > "$sk/.gitignore"
+  ( cd "$sk" && G init -q -b main && G add -A -f && G commit -qm skel )
+  printf '{"instances":[{"name":"testinst","path":"%s","subtree_prefix":"q-system","type":"subtree"}]}\n' \
+    "$inst" > "$sk/instance-registry.json"
+
+  mkdir -p "$inst/q-system" "$inst/plugins/demo" "$inst/.claude"
+  printf 'instance state\n' > "$inst/q-system/tracked.md"
+  ( cd "$inst" && G init -q -b main && G add -A -f && G commit -qm inst )
+}
+
+# The secret, plus one file per already-excluded class, so a regression in any
+# of the five shows up here too rather than only the new entry.
+plant_excludables() {
+  local sk="$1"
+  printf 'ANTHROPIC_API_KEY=sk-not-a-real-key\n' > "$sk/plugins/demo/.env"
+  printf 'OPENAI_API_KEY=sk-also-not-real\n' > "$sk/plugins/demo/.env.local"
+  mkdir -p "$sk/plugins/demo/__pycache__" "$sk/plugins/demo/.venv" \
+           "$sk/plugins/demo/.pytest_cache"
+  printf 'cache\n' > "$sk/plugins/demo/__pycache__/x.pyc"
+  printf 'venv\n' > "$sk/plugins/demo/.venv/pyvenv.cfg"
+  printf 'cache\n' > "$sk/plugins/demo/.pytest_cache/CACHEDIR.TAG"
+  printf 'compiled\n' > "$sk/plugins/demo/stale.pyc"
+}
+
+EXCLUDABLES=(
+  "plugins/demo/.env"
+  "plugins/demo/.env.local"
+  "plugins/demo/__pycache__/x.pyc"
+  "plugins/demo/.venv/pyvenv.cfg"
+  "plugins/demo/.pytest_cache/CACHEDIR.TAG"
+  "plugins/demo/stale.pyc"
+)
+
+# ------------------------------------------------------------------ property 1
+# Nothing in the excluded set reaches an instance. The .env entries are the
+# reason this file exists; the other four are regression cover.
+assert_excluded_files_never_reach_an_instance() {
+  local work sk inst; work="$(mktemp -d)"; sk="$work/skel"; inst="$work/inst"
+  build "$work"
+  plant_excludables "$sk"
+
+  bash "$sk/kipi-update.sh" >"$work/out" 2>&1 || true
+
+  local rel leaked=0
+  for rel in "${EXCLUDABLES[@]}"; do
+    if [ -e "$inst/$rel" ]; then
+      echo "  LEAKED: $rel" >&2
+      leaked=1
+    fi
+  done
+  [ "$leaked" -eq 0 ] || fail "excluded plugin files were copied into the instance"
+
+  # The sync must actually have RUN, or "nothing leaked" is vacuous -- a run that
+  # aborted early would satisfy the loop above while proving nothing.
+  [ -e "$inst/plugins/demo/content.txt" ] || \
+    fail "the plugin sync did not run at all, so the no-leak check proves nothing: $(cat "$work/out")"
+
+  echo "PASS: no excluded plugin file reaches an instance (.env, .env.local, caches, pyc)"
+}
+
+# ------------------------------------------------------------------ property 2
+# The other half of the same value. A file the copy can never carry must not
+# abort the fleet in the ASK-762 preflight -- otherwise adding an exclusion to
+# one consumer silently breaks the other.
+assert_excluded_files_do_not_abort_the_preflight() {
+  local work sk out; work="$(mktemp -d)"; sk="$work/skel"
+  build "$work"
+  plant_excludables "$sk"
+
+  out="$(bash "$sk/kipi-update.sh" 2>&1)" || true
+  if echo "$out" | grep -q "ABORT: the skeleton"; then
+    fail "the source-provenance preflight aborted over files that can never be copied:
+$(echo "$out" | grep -A8 'ABORT: the skeleton')"
+  fi
+
+  echo "PASS: excluded plugin files are silent in the source-provenance preflight"
+}
+
+# ------------------------------------------------------------------ property 3
+# The guard must still fire on an untracked plugin file that IS copied. Without
+# this, widening the exclusion set to silence the preflight would pass property 2
+# by disarming the guard entirely.
+assert_a_copied_untracked_plugin_file_still_aborts() {
+  local work sk inst; work="$(mktemp -d)"; sk="$work/skel"; inst="$work/inst"
+  build "$work"
+  printf 'unreviewed, and rsync WILL copy this\n' > "$sk/plugins/demo/scratch.md"
+
+  local out; out="$(bash "$sk/kipi-update.sh" 2>&1)" || true
+  echo "$out" | grep -q "ABORT: the skeleton" || \
+    fail "an untracked plugin file that IS copied did not abort the preflight: $out"
+  echo "$out" | grep -q "plugins/demo/scratch.md" || \
+    fail "the abort did not name the file: $out"
+
+  echo "PASS: an untracked plugin file that IS copied still aborts"
+}
+
+assert_excluded_files_never_reach_an_instance
+assert_excluded_files_do_not_abort_the_preflight
+assert_a_copied_untracked_plugin_file_still_aborts
+echo "PASS: one exclusion set, both consumers agree, and the guard stays armed"
--- a/q-system/.q-system/scripts/test/test-kipi-update-source-provenance.sh
+++ b/q-system/.q-system/scripts/test/test-kipi-update-source-provenance.sh
@@ -202,11 +202,35 @@ assert_ignored_plugin_files_abort_but_pruned_ones_do_not() {
     fail "the guard fired on classes the plugin rsync excludes: $out"
   fi
 
-  # The secret: ignored, and copied anyway.
+  # REVERSED 2026-08-14 BY ASK-772, deliberately, and the reversal is the point.
+  # This used to assert that an ignored plugins/<name>/.env ABORTS. That was
+  # correct while the rsync copied it: the guard's contract is "alarm on anything
+  # that reaches an instance", and a .env reached all 23.
+  #
+  # ASK-772 added `.env` / `.env.*` to PLUGIN_COPY_EXCLUDES, so the copy no
+  # longer carries it. The same contract now requires SILENCE -- alarming over a
+  # file that cannot leak would halt every fleet update forever, which is a
+  # denial of service dressed as security.
+  #
+  # This is not the guard being weakened. The leak moved from "detected" to
+  # "impossible", which is strictly better, and test-kipi-update-plugin-excludes.sh
+  # is what holds the impossibility. If that file ever goes green while a .env
+  # lands in an instance, THIS assertion is wrong again and must flip back.
   printf 'ANTHROPIC_API_KEY=sk-not-a-real-key\n' > "$sk/plugins/demo/.env"
-  assert_aborts_untouched "$sk" "$inst" "ignored plugin .env" "plugins/demo/.env"
+  printf 'OPENAI_API_KEY=sk-also-not-real\n' > "$sk/plugins/demo/.env.local"
+  out="$(bash "$sk/kipi-update.sh" 2>&1)" || true
+  if echo "$out" | grep -q "ABORT: the skeleton"; then
+    fail "the guard fired on a .env the plugin copy now excludes: $out"
+  fi
+  rm -f "$sk/plugins/demo/.env" "$sk/plugins/demo/.env.local"
 
-  echo "PASS: an ignored plugin .env aborts; rsync-excluded classes stay silent"
+  # An untracked plugin file that IS still copied must abort, or the assertion
+  # above would be satisfied by a guard that had simply stopped working.
+  printf 'unreviewed, and rsync WILL copy this\n' > "$sk/plugins/demo/scratch.md"
+  assert_aborts_untouched "$sk" "$inst" "copied untracked plugin file" "plugins/demo/scratch.md"
+  rm -f "$sk/plugins/demo/scratch.md"
+
+  echo "PASS: rsync-excluded classes stay silent; a file the copy DOES carry still aborts"
 }
 
 # ---------------------------------------------------------------- property 2


assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-153-20260814-105648.md` (61314 bytes).

```
FINDINGS:
minor|The broad .env.* exclusion also removes explicitly permitted non-secret dotenv template files such as .env.example.|kipi-update.sh:232
END FINDINGS
```

--- reviewer output, last 54837 bytes of 61314 (full review at the path above) ---

FINDINGS:
END FINDINGS
codex
I’m using the founder voice and AUDHD skills only to keep the final review crisp. The headless instruction overrides their interactive wait rules. I’m starting the read-only review now.
exec
/bin/zsh -lc "sed -n '1,240p' /Users/assafkipnis/.agents/skills/assaf-voice/SKILL.md && sed -n '1,240p' /Users/assafkipnis/.agents/skills/audhd-executive-function/SKILL.md && gh -R assafkip/kipi-system pr view 153 && gh -R assafkip/kipi-system pr diff 153" in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-153
 exited 1 in 0ms:
---
name: assaf-voice
description: ALWAYS USE THIS SKILL for ANY written output - LinkedIn posts, Medium articles, emails, DMs, talk tracks, outreach messages, slide copy, Notion content, investor updates, comments, replies, or any other text that will be read by another human. This includes both ad-hoc requests ("write me a reply to X") and planned/scheduled content (morning routine posts, marketing pipeline, engagement hitlist comments). This skill is NOT optional - it is the default voice layer for all written output in this system.
metadata:
  version: 2.0.0
---

# Assaf Voice

The voice lives in ONE place and it is not in this skill:
`~/projects/consulting/q-consult/voice/`.

This file carries no voice prose of its own. It tells you what to load and how
little. The v1 layer files this skill used to load (`references/layers/*.md`,
`voice-dna.md`, `writing-samples.md`) are RETIRED. They were a second copy of the
corpus, and two copies is the drift machine that was removed on 2026-08-13
(ASK-699). v1 also pointed at `~/.Codex/skills/...` paths that do not exist on
this machine, so its documented loading table could not have been followed.

## Get the reference by running the selector

Do not hand-pick exemplars and do not paste the corpus.

```bash
python3 ~/projects/social-voice/scripts/voice_ref.py --channel x --words 480
python3 ~/projects/social-voice/scripts/voice_ref.py --channel x --words 25
python3 ~/projects/social-voice/scripts/voice_ref.py --channel linkedin --words 200
```

It prints identity, the live corrections with their scopes honoured, and the
exemplars matched to the piece. It exits non-zero when the corpus is missing
rather than returning a silent blank.

**`--words` is not cosmetic.** Channel and LENGTH are two axes. The x corpus runs
5 to 55 words with a single 479-word row, so a long piece written against the
short rows comes back formal. That was measured on 2026-08-13; the founder's
verdict on the draft was "this is not in my voiccc". Pass the real target length.

**One long-form x exemplar exists.** The selector says so in its output. Treat a
long-form register as thin evidence, not a settled pattern.

## Routing

- **Social posts** go through the posting lane, not a hand-written draft:
  `python3 -m pipeline.social idea "<his idea>" --channel x --words <n>` from
  `~/projects/consulting/q-consult/`. It runs the deterministic gate stack and
  never publishes.
- **Reactions to someone else's content:** extract the poster's claims and
  confirm them with the founder BEFORE drafting.
- **Everything else is this skill:** emails, DMs, comments, replies, talk
  tracks, slide copy, Notion content, outreach, investor updates, decks.

## The can't-write rules

- **No emdashes**, anywhere. In docs and code the substitute is ` -- `; in posts
  neither is allowed.
- **Banned words and phrases** live in `voice-lint.py`, not restated here.
  Measure any new list entry against the real corpus first: a word list caught
  the founder's own ordinary vocabulary six times in one session.
- **Never invent a person or company fact.** No checker sees this one.
- **$1,000 is the only public price.**
- **No client names** without recorded permission.
- **DMs and emails open with "I"**, never the person's name. No "circling back",
  no "just checking in".

## After drafting

`voice-lint.py` and `voice-substance-lint.py` run on write and a Stop gate honours
no skip marker. They prove the ABSENCE of known defects. They never prove the
presence of his voice; only the exemplars answer that.
---
name: audhd-executive-function
description: "AUDHD executive function accommodations. Apply to all output the founder will act on."
---

# AUDHD Executive Function Skill -- Investigation Mode

External executive function system for investigation outputs. Governs how briefs, action items, and daily plans are structured.

**Always read:** `~/.Codex/skills/audhd-executive-function/references/rules.md`

---

## THE ONE RULE

**If the founder cannot copy-paste it, click it, or check it off, it does not belong in the output.**

---

## ACTIONABILITY RULES (ENFORCED)

### A1: No Cross-References
Never say "see section above" or "refer to file Y." Every item contains the actual content inline. The founder never scrolls or searches.

### A2: Next Physical Action
Every item is the literal next physical thing to do. Not "Follow up on Rickenson lead" but the actual search query, URL to open, or message to send.

### A3: All Pending Actions Get Specifics
Every action item includes: what to do, where to do it (URL/platform), expected time, and energy tag.

### A4: Friction-Ordered
Items sorted by friction, lowest first. 2-minute username checks before 15-minute browser deep dives. Quick momentum wins first.

---

## INVESTIGATION OUTPUT RULES

### Briefs and Reports
- Lead with findings, not methodology
- Confidence levels on every assessment (Confirmed / High / Moderate / Low / Speculative)
- Evidence citations inline, not in footnotes
- Collection gaps listed as actionable next steps, not as "we don't know"

### Action Items
Every collection task includes:
- What to collect
- Where (platform + URL if known)
- Tool to use (script command or Chrome)
- Energy tag (Quick Win / Deep Focus / People / Admin)
- Time estimate
- What it proves if successful

### Evidence Summaries
- Bullet points, not paragraphs
- Source reliability rating (A-F) on every claim
- Distinguish: confirmed fact / assessed judgment / speculation
- Flag contradictions explicitly

---

## LANGUAGE RULES (ENFORCED)

### Never Use
- "overdue" (use "carried forward")
- "missed" (use "ready when you are")
- "failed" (use "didn't land" or "inconclusive")
- "forgot" (use "not yet done")
- "behind" (use "in progress")
- "urgent" as pressure (state facts calmly)
- "you need to" (use "you could")
- "nobody responded" (use "awaiting reply")

### Always Use
- Effort tracking: "You ran 15 searches across 4 platforms" not outcome tracking
- "Carried forward from last session" not "this is overdue"
- "You could..." not "you need to..."
- Calm factual statements about investigation state

### Progress Framing
- "3 of 5 targets profiled" is a win. Show what WAS done
- End of session: effort celebration, not gap shaming
- Completed items stay visible so the founder sees accomplishments

---

## DECISION ELIMINATION

### The System Decides
- Which collection task is highest priority (based on hypothesis impact)
- What tool to use (based on diagnose.sh results)
- In what order (friction-sorted)

### The Founder Executes
- Approves the plan
- Provides client responses
- Makes judgment calls on ambiguous evidence

### Never Present Options Without a Recommendation
Bad: "We could search Facebook OR LinkedIn OR both. What do you prefer?"
Good: "Recommend Facebook first (higher probability of finding connections). LinkedIn second if Facebook is dry. Time: 20 min total."

---

## ANTI-PATTERNS (NEVER DO)

1. **Walls of text.** Bullet points or tables. Always.
2. **Abstract task names.** "Investigate Rickenson" is not a task. "Search Facebook for 'Rickenson' + 'Palmyre' together" is.
3. **Options without recommendations.** The system picks. The founder overrides if they want.
4. **Stacking pressure.** Never list 3+ carried-forward items together.
5. **Outcome metrics as judgment.** "0 results found" is data. Never frame it as failure.
6. **Sequential dependencies without flagging.** If step 3 needs step 1's output, say so explicitly.
error connecting to api.github.com
check your internet connection or https://githubstatus.com

codex
GitHub is unreachable in this sandbox, so I’m switching to the checked-out review tree and its local base refs. I’ll verify the exact PR range before inspecting code and running repros.
exec
/bin/zsh -lc "sed -n '1,260p' /Users/assafkipnis/.Codex/skills/audhd-executive-function/references/rules.md 2>/dev/null || true
git status --short --branch
git remote -v
git branch -a --contains HEAD
git log --oneline --decorate -12
git show-ref | rg 'pull/153|origin/(main|master)'" in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-153
 succeeded in 327ms:
## HEAD (no branch)
origin	https://github.com/assafkip/kipi-system.git (fetch)
origin	https://github.com/assafkip/kipi-system.git (push)
* (no branch)
+ sana/ask-772-dotenv-exclusion
  remotes/origin/sana/ask-772-dotenv-exclusion
eeb3b9cd (HEAD, origin/sana/ask-772-dotenv-exclusion, sana/ask-772-dotenv-exclusion) fix(updater): one exclusion set, so a plugin .env stops shipping fleet-wide (ASK-772)
0c00c30d (main) Refuse to fan .claude/ and plugins/ bytes nobody reviewed (ASK-762) (#149)
47b7dc8b fix(open-loops): consult the spillover ledger before counting a deferral (ASK-763) (#150)
0d1c3fd0 Refuse the engagement root, opt in the first two repos, and stop the preflight inventing reasons (ASK-754, ASK-755) (#148)
2bfefc74 Unreachable work is named, not filed under "skipped" (ASK-729) (#143)
d8869ea2 Scope every gh call to the target repo, and let dispatch enter more than one (ASK-738) (#146)
9dde7db5 Rescue three unpushed commits, and replace the length threshold with nearest-length (ASK-710, ASK-712) (#147)
21fd82e1 feat(dispatch): refuse a client engagement repo even when it is opted in (ASK-741) (#144)
57415553 Get main green: three inherited failures, three different root causes (ASK-740, ASK-746) (#145)
11f0eb8a feat(guard): block client names from reaching this public repo
3fa2a713 fix(tests): scrub a client name and a local path out of an alert fixture
cc5b862e lesson: feedback lands where artifacts exist
833d5b34715a4c439b99722419f4ea9d09447b57 refs/remotes/origin/main

exec
/bin/zsh -lc 'git merge-base HEAD origin/main
git diff --stat $(git merge-base HEAD origin/main)..HEAD
git diff --name-status $(git merge-base HEAD origin/main)..HEAD
git diff --find-renames $(git merge-base HEAD origin/main)..HEAD' in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-153
 succeeded in 0ms:
0c00c30ddc4e02022f45117985e8bdd01dea5abb
 kipi-update.sh                                     |  86 ++++++++++-
 q-system/.q-system/capability-manifest.json        |  11 +-
 .../test/test-kipi-update-plugin-excludes.sh       | 161 +++++++++++++++++++++
 .../test/test-kipi-update-source-provenance.sh     |  30 +++-
 4 files changed, 277 insertions(+), 11 deletions(-)
M	kipi-update.sh
M	q-system/.q-system/capability-manifest.json
A	q-system/.q-system/scripts/test/test-kipi-update-plugin-excludes.sh
M	q-system/.q-system/scripts/test/test-kipi-update-source-provenance.sh
diff --git a/kipi-update.sh b/kipi-update.sh
index 3f2761d4..f0009bd4 100755
--- a/kipi-update.sh
+++ b/kipi-update.sh
@@ -198,6 +198,55 @@ is_managed_plugin_path() {
 # plugins/ tree an EARLIER update staged and never committed). The fleet updater
 # was blocked by itself, silently. This list is deliberately narrow and explicit:
 # anything not named here is treated as founder work and still refuses.
+# THE ONE plugin-copy exclusion set (ASK-772). Two consumers derive from it and
+# they answer halves of a single question:
+#
+#   the per-plugin rsync         -- "never copy this into an instance"
+#   the source-provenance preflight -- "never alarm about this, because it is
+#                                       never copied"
+#
+# Writing the set down twice is how a secret shipped. `.gitignore:3` is `*.env`
+# and kipi-design's cip/generate.py reads a plugin-root .env for API keys, so a
+# gitignored plugins/<name>/.env was invisible to `git status` and copied happily
+# by rsync into all 23 instances (Codex, PR #149 round 3).
+#
+# Adding it to ONE consumer swaps one failure for another: rsync-only leaves the
+# preflight aborting the whole fleet over a file that can no longer leak (a
+# denial of service on every update), preflight-only leaves the leak open and
+# silent. Hence one list.
+#
+# Each entry is "<rsync --exclude pattern>::<regex matching the same paths>".
+# Two representations, adjacent, because rsync globs and grep regexes are not
+# interconvertible in bash without a converter more fragile than the duplication
+# it removes. `::` is the delimiter, not `|`, because the regexes contain `|`.
+# test-kipi-update-plugin-excludes.sh drives BOTH consumers off this list and
+# fails if they disagree, so agreement is proven behaviourally rather than by
+# the two strings having been typed the same.
+PLUGIN_COPY_EXCLUDES=(
+  "/.git/::(^|/)\.git/"
+  "__pycache__/::(^|/)__pycache__/"
+  "*.pyc::\.pyc\$"
+  ".venv/::(^|/)\.venv/"
+  ".pytest_cache/::(^|/)\.pytest_cache/"
+  ".env::(^|/)\.env\$"
+  ".env.*::(^|/)\.env\."
+)
+
+plugin_copy_rsync_flags() {
+  local entry
+  for entry in "${PLUGIN_COPY_EXCLUDES[@]}"; do
+    printf -- '--exclude=%s\n' "${entry%%::*}"
+  done
+}
+
+plugin_copy_filter_regex() {
+  local entry out=""
+  for entry in "${PLUGIN_COPY_EXCLUDES[@]}"; do
+    out="${out:+$out|}${entry#*::}"
+  done
+  printf '%s\n' "$out"
+}
+
 SYSTEM_OWNED_PATHS=(
   "q-system/memory/.sycophancy-monthly-stamp"
   "q-system/.q-system/claude-integrity-baseline.json"
@@ -448,7 +497,7 @@ SYNC_SCOPE_DIRTY="$(
     git -C "$SCRIPT_DIR" status --porcelain --ignored=matching -- plugins \
       2>/dev/null || true
   } | sed 's/^...//' \
-    | grep -vE '(^|/)(\.git|__pycache__|\.venv|\.pytest_cache)/|\.pyc$' || true
+    | grep -vE "$(plugin_copy_filter_regex)" || true
 )"
 if [ -n "$SYNC_SCOPE_DIRTY" ]; then
   echo ""
@@ -895,14 +944,36 @@ stage_config_sync() {
     # syncer now walk the same list, so the stager can no longer name a path
     # the syncer will not write -- which is what the old post-hoc [ -d ] filter
     # was patching around. See managed_plugin_names for the scar.
+    # THE THIRD CONSUMER of PLUGIN_COPY_EXCLUDES, and the one that made the
+    # duplication expensive rather than merely untidy (ASK-772). This walk had
+    # its OWN hardcoded prune list -- .git, __pycache__, .pytest_cache, .venv,
+    # *.pyc -- a third copy of the same value. Excluding `.env` from the rsync
+    # without teaching the stager left it naming a path the syncer no longer
+    # writes, and `git add` on a missing path is fatal:
+    #
+    #     fatal: pathspec 'plugins/demo/.env.local' did not match any files
+    #     ERROR: config sync did not reach a complete committed state
+    #
+    # which is the exact class the comment above already warns about, arriving
+    # from the one direction it did not cover. The find still prunes cheaply so
+    # it never descends into a .venv; the shared regex is what decides.
+    plugin_excl_re="$(plugin_copy_filter_regex)"
     while IFS= read -r -d '' plugin_name; do
       while IFS= read -r -d '' source; do
-        plugin_paths+=("${source#"$SCRIPT_DIR/"}")
+        relative="${source#"$SCRIPT_DIR/"}"
+        # Matched on the path RELATIVE to the plugin dir, the same thing rsync
+        # matches its --exclude patterns against. Matching the repo-relative path
+        # would let a `.env` anywhere above the plugin root change the answer.
+        case "${source#"$SKELETON_PLUGIN_ROOT/$plugin_name/"}" in
+          *) printf '%s' "${source#"$SKELETON_PLUGIN_ROOT/$plugin_name/"}" \
+               | grep -qE "$plugin_excl_re" && continue ;;
+        esac
+        plugin_paths+=("$relative")
       done < <(
         find "$SKELETON_PLUGIN_ROOT/$plugin_name" \
           \( -type d -name .git -o -type d -name __pycache__ \
              -o -type d -name .pytest_cache -o -type d -name .venv \) -prune -o \
-          \( -type f ! -name '*.pyc' -o -type l \) -print0
+          \( -type f -o -type l \) -print0
       )
     done < <(managed_plugin_names)
     # A path the INSTANCE ignores cannot be staged, and `git add` treats that
@@ -1888,9 +1959,14 @@ print("\n".join(mod.EXTRA_WATCHED))
         # It was 107MB of the 112MB plugin tree, copied into 23 instances
         # where it could never work. --delete-excluded also clears the stale
         # copies already there. Pairs with test-kipi-update-build-artifacts.sh.
+        # Flags derived from PLUGIN_COPY_EXCLUDES, never hand-listed here: this
+        # site and the preflight filter are the two consumers that must agree,
+        # and a hand-listed copy is how a plugin-root .env shipped fleet-wide
+        # (ASK-772). Word-splitting the generator output is intended -- every
+        # pattern is a shell-safe token by construction.
+        # shellcheck disable=SC2046
         if ! rsync -a --delete --delete-excluded \
-            --exclude="/.git/" --exclude="__pycache__/" --exclude="*.pyc" \
-            --exclude=".venv/" --exclude=".pytest_cache/" \
+            $(plugin_copy_rsync_flags) \
             "$SKELETON_PLUGIN_ROOT/$plugin_name/" \
             "$path/plugins/$plugin_name/" 2>/dev/null; then
           CONFIG_FAILED=1
diff --git a/q-system/.q-system/capability-manifest.json b/q-system/.q-system/capability-manifest.json
index 2124da5b..2896c969 100644
--- a/q-system/.q-system/capability-manifest.json
+++ b/q-system/.q-system/capability-manifest.json
@@ -150,6 +150,10 @@
       "runner": "bash",
       "timeout_s": 300
     },
+    {
+      "path": "q-system/.q-system/scripts/test/test-kipi-update-plugin-excludes.sh",
+      "runner": "bash"
+    },
     {
       "path": "q-system/.q-system/scripts/test/test-kipi-update-preservation-failure.sh",
       "runner": "bash"
@@ -602,6 +606,7 @@
     "q-system/.q-system/scripts/test/test-kipi-update-dry-final-state.sh",
     "q-system/.q-system/scripts/test/test-kipi-update-hook-contract.sh",
     "q-system/.q-system/scripts/test/test-kipi-update-leak-preflight.sh",
+    "q-system/.q-system/scripts/test/test-kipi-update-plugin-excludes.sh",
     "q-system/.q-system/scripts/test/test-kipi-update-preservation-failure.sh",
     "q-system/.q-system/scripts/test/test-kipi-update-safety.sh",
     "q-system/.q-system/scripts/test/test-kipi-update-source-provenance.sh",
@@ -686,7 +691,7 @@
     },
     {
       "path": "q-system/.q-system/scripts/stat-verify.py",
-      "reason": "802-line stat verification engine; zero hook wiring and its stat-registry.json data file exists in exactly one instance — wiring it is a founder product decision",
+      "reason": "802-line stat verification engine; zero hook wiring and its stat-registry.json data file exists in exactly one instance \u2014 wiring it is a founder product decision",
       "spillover_id": "sp-72b60bff"
     },
     {
@@ -706,8 +711,8 @@
     }
   ],
   "uncovered_known": [
-    "plugins/*/tests — prd-os plugin suite, run by the plugin's own harness; outside v1 scan roots (finding-9 deferred)",
+    "plugins/*/tests \u2014 prd-os plugin suite, run by the plugin's own harness; outside v1 scan roots (finding-9 deferred)",
     "repo-root *.sh utilities are wiring surfaces, not test artifacts",
-    "stat-registry.json: NOT expressible as required_data in v1 — the one instance holding it keeps canonical under its own instance_q_dir (not q-system/) and its registry name differs from its dir basename, so a scoped entry could never fire (grounded 2026-07-23). Declared gap; owning decision + instance detail in spillover sp-72b60bff"
+    "stat-registry.json: NOT expressible as required_data in v1 \u2014 the one instance holding it keeps canonical under its own instance_q_dir (not q-system/) and its registry name differs from its dir basename, so a scoped entry could never fire (grounded 2026-07-23). Declared gap; owning decision + instance detail in spillover sp-72b60bff"
   ]
 }
diff --git a/q-system/.q-system/scripts/test/test-kipi-update-plugin-excludes.sh b/q-system/.q-system/scripts/test/test-kipi-update-plugin-excludes.sh
new file mode 100755
index 00000000..e3bdf17e
--- /dev/null
+++ b/q-system/.q-system/scripts/test/test-kipi-update-plugin-excludes.sh
@@ -0,0 +1,161 @@
+#!/usr/bin/env bash
+# A plugin-root .env is copied into all 23 instances and committed there.
+#
+# The per-plugin rsync excludes exactly five classes -- .git/, __pycache__/,
+# *.pyc, .venv/, .pytest_cache/ -- and nothing else. `.gitignore:3` is `*.env`,
+# so a plugin .env is invisible to `git status` while rsync copies it happily.
+# It is not hypothetical: plugins/kipi-design/skills/design/scripts/cip/
+# generate.py reads a plugin-root .env for API keys.
+#
+# Found by Codex on PR #149 round 3, filed as ASK-772.
+#
+# THE DRIFT TRAP THIS FILE EXISTS TO CLOSE. The exclusion set was written down
+# TWICE: once as rsync --exclude flags (~L1892) and once as a grep filter in the
+# ASK-762 source-provenance preflight (~L451). The two must agree, because they
+# answer halves of one question:
+#
+#   rsync excludes  = "never copy this"
+#   preflight filter = "do not alarm about this, because it is never copied"
+#
+# Add `.env` to the rsync only and the preflight starts aborting the fleet over a
+# file that can no longer leak. Add it to the preflight only and the leak stays
+# open and silent. One value, two authorities, opposite failure modes.
+#
+# So this file does NOT compare the two lists as strings -- that only proves they
+# were typed the same. It drives BOTH behaviours from the shared set and asserts
+# each entry is excluded by the copy AND silent in the preflight. A new entry
+# added to one consumer and not the other fails here.
+set -euo pipefail
+ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../../../.." && pwd)"
+fail() { echo "FAIL: $1" >&2; exit 1; }
+G() { git -c user.email=t@t.t -c user.name=test -c commit.gpgsign=false "$@"; }
+
+build() {
+  local work="$1" sk="$work/skel" inst="$work/inst"
+  mkdir -p "$sk/q-system/.q-system/scripts" "$sk/q-system/.q-system/state" \
+           "$sk/q-system/hooks" "$sk/plugins/demo" "$sk/.claude/rules"
+  for f in kipi-update.sh kipi-update-preserve-scan.py kipi-update-deletion-guard.py \
+           validate-separation.py; do
+    cp "$ROOT/$f" "$sk/$f"
+  done
+  cp "$ROOT/q-system/.q-system/scripts/propagation-leak-gate.py" \
+     "$sk/q-system/.q-system/scripts/propagation-leak-gate.py"
+  cp "$ROOT/q-system/.q-system/scripts/containment-targets.py" \
+     "$sk/q-system/.q-system/scripts/containment-targets.py"
+  cp "$ROOT/q-system/hooks/auto-commit.py" "$sk/q-system/hooks/auto-commit.py"
+  cat > "$sk/q-system/.q-system/state/propagation-leak-baseline.json" <<'JSON'
+{
+  "schema_version": 1,
+  "blocking_classes": ["case_proof_gap", "client_identity", "dated_interaction",
+                       "pricing", "relationship", "source_identity",
+                       "sourced_interaction"],
+  "classifier_sha256": null,
+  "entries": []
+}
+JSON
+  printf 'generic skeleton content\n' > "$sk/q-system/tracked.md"
+  printf 'plugin content\n' > "$sk/plugins/demo/content.txt"
+  printf '# demo rule\n' > "$sk/.claude/rules/demo.md"
+  # The skeleton ignores secrets and caches, exactly as the real repo does. That
+  # is the whole point: gitignored is not "absent", it is "invisible to git and
+  # fully visible to rsync".
+  printf '*.env\n.env\n__pycache__/\n*.pyc\n.venv/\n.pytest_cache/\n' > "$sk/.gitignore"
+  ( cd "$sk" && G init -q -b main && G add -A -f && G commit -qm skel )
+  printf '{"instances":[{"name":"testinst","path":"%s","subtree_prefix":"q-system","type":"subtree"}]}\n' \
+    "$inst" > "$sk/instance-registry.json"
+
+  mkdir -p "$inst/q-system" "$inst/plugins/demo" "$inst/.claude"
+  printf 'instance state\n' > "$inst/q-system/tracked.md"
+  ( cd "$inst" && G init -q -b main && G add -A -f && G commit -qm inst )
+}
+
+# The secret, plus one file per already-excluded class, so a regression in any
+# of the five shows up here too rather than only the new entry.
+plant_excludables() {
+  local sk="$1"
+  printf 'ANTHROPIC_API_KEY=sk-not-a-real-key\n' > "$sk/plugins/demo/.env"
+  printf 'OPENAI_API_KEY=sk-also-not-real\n' > "$sk/plugins/demo/.env.local"
+  mkdir -p "$sk/plugins/demo/__pycache__" "$sk/plugins/demo/.venv" \
+           "$sk/plugins/demo/.pytest_cache"
+  printf 'cache\n' > "$sk/plugins/demo/__pycache__/x.pyc"
+  printf 'venv\n' > "$sk/plugins/demo/.venv/pyvenv.cfg"
+  printf 'cache\n' > "$sk/plugins/demo/.pytest_cache/CACHEDIR.TAG"
+  printf 'compiled\n' > "$sk/plugins/demo/stale.pyc"
+}
+
+EXCLUDABLES=(
+  "plugins/demo/.env"
+  "plugins/demo/.env.local"
+  "plugins/demo/__pycache__/x.pyc"
+  "plugins/demo/.venv/pyvenv.cfg"
+  "plugins/demo/.pytest_cache/CACHEDIR.TAG"
+  "plugins/demo/stale.pyc"
+)
+
+# ------------------------------------------------------------------ property 1
+# Nothing in the excluded set reaches an instance. The .env entries are the
+# reason this file exists; the other four are regression cover.
+assert_excluded_files_never_reach_an_instance() {
+  local work sk inst; work="$(mktemp -d)"; sk="$work/skel"; inst="$work/inst"
+  build "$work"
+  plant_excludables "$sk"
+
+  bash "$sk/kipi-update.sh" >"$work/out" 2>&1 || true
+
+  local rel leaked=0
+  for rel in "${EXCLUDABLES[@]}"; do
+    if [ -e "$inst/$rel" ]; then
+      echo "  LEAKED: $rel" >&2
+      leaked=1
+    fi
+  done
+  [ "$leaked" -eq 0 ] || fail "excluded plugin files were copied into the instance"
+
+  # The sync must actually have RUN, or "nothing leaked" is vacuous -- a run that
+  # aborted early would satisfy the loop above while proving nothing.
+  [ -e "$inst/plugins/demo/content.txt" ] || \
+    fail "the plugin sync did not run at all, so the no-leak check proves nothing: $(cat "$work/out")"
+
+  echo "PASS: no excluded plugin file reaches an instance (.env, .env.local, caches, pyc)"
+}
+
+# ------------------------------------------------------------------ property 2
+# The other half of the same value. A file the copy can never carry must not
+# abort the fleet in the ASK-762 preflight -- otherwise adding an exclusion to
+# one consumer silently breaks the other.
+assert_excluded_files_do_not_abort_the_preflight() {
+  local work sk out; work="$(mktemp -d)"; sk="$work/skel"
+  build "$work"
+  plant_excludables "$sk"
+
+  out="$(bash "$sk/kipi-update.sh" 2>&1)" || true
+  if echo "$out" | grep -q "ABORT: the skeleton"; then
+    fail "the source-provenance preflight aborted over files that can never be copied:
+$(echo "$out" | grep -A8 'ABORT: the skeleton')"
+  fi
+
+  echo "PASS: excluded plugin files are silent in the source-provenance preflight"
+}
+
+# ------------------------------------------------------------------ property 3
+# The guard must still fire on an untracked plugin file that IS copied. Without
+# this, widening the exclusion set to silence the preflight would pass property 2
+# by disarming the guard entirely.
+assert_a_copied_untracked_plugin_file_still_aborts() {
+  local work sk inst; work="$(mktemp -d)"; sk="$work/skel"; inst="$work/inst"
+  build "$work"
+  printf 'unreviewed, and rsync WILL copy this\n' > "$sk/plugins/demo/scratch.md"
+
+  local out; out="$(bash "$sk/kipi-update.sh" 2>&1)" || true
+  echo "$out" | grep -q "ABORT: the skeleton" || \
+    fail "an untracked plugin file that IS copied did not abort the preflight: $out"
+  echo "$out" | grep -q "plugins/demo/scratch.md" || \
+    fail "the abort did not name the file: $out"
+
+  echo "PASS: an untracked plugin file that IS copied still aborts"
+}
+
+assert_excluded_files_never_reach_an_instance
+assert_excluded_files_do_not_abort_the_preflight
+assert_a_copied_untracked_plugin_file_still_aborts
+echo "PASS: one exclusion set, both consumers agree, and the guard stays armed"
diff --git a/q-system/.q-system/scripts/test/test-kipi-update-source-provenance.sh b/q-system/.q-system/scripts/test/test-kipi-update-source-provenance.sh
index e83f7f30..896d8833 100755
--- a/q-system/.q-system/scripts/test/test-kipi-update-source-provenance.sh
+++ b/q-system/.q-system/scripts/test/test-kipi-update-source-provenance.sh
@@ -202,11 +202,35 @@ assert_ignored_plugin_files_abort_but_pruned_ones_do_not() {
     fail "the guard fired on classes the plugin rsync excludes: $out"
   fi
 
-  # The secret: ignored, and copied anyway.
+  # REVERSED 2026-08-14 BY ASK-772, deliberately, and the reversal is the point.
+  # This used to assert that an ignored plugins/<name>/.env ABORTS. That was
+  # correct while the rsync copied it: the guard's contract is "alarm on anything
+  # that reaches an instance", and a .env reached all 23.
+  #
+  # ASK-772 added `.env` / `.env.*` to PLUGIN_COPY_EXCLUDES, so the copy no
+  # longer carries it. The same contract now requires SILENCE -- alarming over a
+  # file that cannot leak would halt every fleet update forever, which is a
+  # denial of service dressed as security.
+  #
+  # This is not the guard being weakened. The leak moved from "detected" to
+  # "impossible", which is strictly better, and test-kipi-update-plugin-excludes.sh
+  # is what holds the impossibility. If that file ever goes green while a .env
+  # lands in an instance, THIS assertion is wrong again and must flip back.
   printf 'ANTHROPIC_API_KEY=sk-not-a-real-key\n' > "$sk/plugins/demo/.env"
-  assert_aborts_untouched "$sk" "$inst" "ignored plugin .env" "plugins/demo/.env"
+  printf 'OPENAI_API_KEY=sk-also-not-real\n' > "$sk/plugins/demo/.env.local"
+  out="$(bash "$sk/kipi-update.sh" 2>&1)" || true
+  if echo "$out" | grep -q "ABORT: the skeleton"; then
+    fail "the guard fired on a .env the plugin copy now excludes: $out"
+  fi
+  rm -f "$sk/plugins/demo/.env" "$sk/plugins/demo/.env.local"
+
+  # An untracked plugin file that IS still copied must abort, or the assertion
+  # above would be satisfied by a guard that had simply stopped working.
+  printf 'unreviewed, and rsync WILL copy this\n' > "$sk/plugins/demo/scratch.md"
+  assert_aborts_untouched "$sk" "$inst" "copied untracked plugin file" "plugins/demo/scratch.md"
+  rm -f "$sk/plugins/demo/scratch.md"
 
-  echo "PASS: an ignored plugin .env aborts; rsync-excluded classes stay silent"
+  echo "PASS: rsync-excluded classes stay silent; a file the copy DOES carry still aborts"
 }
 
 # ---------------------------------------------------------------- property 2

codex
The sandbox blocks all temporary writes, so the integration tests cannot create their fixtures. I’m continuing with read-only executable repros against the actual functions and repository paths, and I’ll only keep findings whose behavior can still be demonstrated.
exec
/bin/zsh -lc "rg -n \"PLUGIN_COPY_EXCLUDES|plugin_copy_|stage_config_sync|SKELETON_PLUGIN_ROOT|set -\" kipi-update.sh
find plugins -type d -name .git -o -type f \\( -name '.env' -o -name '.env.*' \\) -print
bash -n kipi-update.sh q-system/.q-system/scripts/test/test-kipi-update-plugin-excludes.sh q-system/.q-system/scripts/test/test-kipi-update-source-provenance.sh
python3 -m json.tool q-system/.q-system/capability-manifest.json >/dev/null && echo 'PASS: manifest JSON parses'" in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-153
 succeeded in 0ms:
2:set -euo pipefail
135:SKELETON_PLUGIN_ROOT="$SCRIPT_DIR/plugins"
149:  for plugin_dir in "$SKELETON_PLUGIN_ROOT"/*/; do
225:PLUGIN_COPY_EXCLUDES=(
235:plugin_copy_rsync_flags() {
237:  for entry in "${PLUGIN_COPY_EXCLUDES[@]}"; do
242:plugin_copy_filter_regex() {
244:  for entry in "${PLUGIN_COPY_EXCLUDES[@]}"; do
416:# `if` form, not a bare assignment: under `set -e` a failing command
500:    | grep -vE "$(plugin_copy_filter_regex)" || true
761:  git -C "$target" reset -q HEAD -- "$CHECKPOINT_PREFIX/" .claude/ plugins/ \
916:  git -C "$target" reset -q -- "$@" 2>/dev/null || true
919:stage_config_sync() {
942:  if [ -d "$SKELETON_PLUGIN_ROOT" ]; then
947:    # THE THIRD CONSUMER of PLUGIN_COPY_EXCLUDES, and the one that made the
960:    plugin_excl_re="$(plugin_copy_filter_regex)"
967:        case "${source#"$SKELETON_PLUGIN_ROOT/$plugin_name/"}" in
968:          *) printf '%s' "${source#"$SKELETON_PLUGIN_ROOT/$plugin_name/"}" \
973:        find "$SKELETON_PLUGIN_ROOT/$plugin_name" \
1038:set -euo pipefail
1077:  set -e
1533:        # on an EMPTY array is an unbound-variable error under `set -u` (bash
1951:    if [ -d "$SKELETON_PLUGIN_ROOT" ]; then
1962:        # Flags derived from PLUGIN_COPY_EXCLUDES, never hand-listed here: this
1969:            $(plugin_copy_rsync_flags) \
1970:            "$SKELETON_PLUGIN_ROOT/$plugin_name/" \
1985:        { stage_config_sync "$path" || { unstage_scope "$path" .claude/ plugins/; false; }; } &&
PASS: manifest JSON parses

exec
/bin/zsh -lc "sed -n '919,1025p' kipi-update.sh
sed -n '1940,2005p' kipi-update.sh
rg -n --hidden -g '"'!**/.git/**'"' '(\\.env\\.example|\\.env\\.sample|\\.env\\.template)' .
git ls-files 'plugins/**/.env*' '.gitignore' && nl -ba .gitignore | sed -n '1,80p'
rsync --version | head -2" in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-153
 succeeded in 0ms:
stage_config_sync() {
  local target="$1"
  local scope source relative plugin_name ignored_paths
  local -a plugin_paths=() stage_paths=()
  for scope in .claude plugins; do
    if [ -n "$(git -C "$target" ls-files -- "$scope/")" ]; then
      git -C "$target" add -u -- "$scope/" || return 1
    fi
  done
  if [ -f "$target/.claude/settings.json" ]; then
    git -C "$target" add -- .claude/settings.json || return 1
  fi
  for scope in agents rules output-styles; do
    if [ -d "$SCRIPT_DIR/.claude/$scope" ]; then
      while IFS= read -r -d '' source; do
        relative="${source#"$SCRIPT_DIR/"}"
        git -C "$target" add -- "$relative" || return 1
      done < <(
        find "$SCRIPT_DIR/.claude/$scope" -maxdepth 1 -type f \
          -name '*.md' -print0
      )
    fi
  done
  if [ -d "$SKELETON_PLUGIN_ROOT" ]; then
    # Rooted PER MANAGED PLUGIN, not at plugins/ wholesale. The stager and the
    # syncer now walk the same list, so the stager can no longer name a path
    # the syncer will not write -- which is what the old post-hoc [ -d ] filter
    # was patching around. See managed_plugin_names for the scar.
    # THE THIRD CONSUMER of PLUGIN_COPY_EXCLUDES, and the one that made the
    # duplication expensive rather than merely untidy (ASK-772). This walk had
    # its OWN hardcoded prune list -- .git, __pycache__, .pytest_cache, .venv,
    # *.pyc -- a third copy of the same value. Excluding `.env` from the rsync
    # without teaching the stager left it naming a path the syncer no longer
    # writes, and `git add` on a missing path is fatal:
    #
    #     fatal: pathspec 'plugins/demo/.env.local' did not match any files
    #     ERROR: config sync did not reach a complete committed state
    #
    # which is the exact class the comment above already warns about, arriving
    # from the one direction it did not cover. The find still prunes cheaply so
    # it never descends into a .venv; the shared regex is what decides.
    plugin_excl_re="$(plugin_copy_filter_regex)"
    while IFS= read -r -d '' plugin_name; do
      while IFS= read -r -d '' source; do
        relative="${source#"$SCRIPT_DIR/"}"
        # Matched on the path RELATIVE to the plugin dir, the same thing rsync
        # matches its --exclude patterns against. Matching the repo-relative path
        # would let a `.env` anywhere above the plugin root change the answer.
        case "${source#"$SKELETON_PLUGIN_ROOT/$plugin_name/"}" in
          *) printf '%s' "${source#"$SKELETON_PLUGIN_ROOT/$plugin_name/"}" \
               | grep -qE "$plugin_excl_re" && continue ;;
        esac
        plugin_paths+=("$relative")
      done < <(
        find "$SKELETON_PLUGIN_ROOT/$plugin_name" \
          \( -type d -name .git -o -type d -name __pycache__ \
             -o -type d -name .pytest_cache -o -type d -name .venv \) -prune -o \
          \( -type f -o -type l \) -print0
      )
    done < <(managed_plugin_names)
    # A path the INSTANCE ignores cannot be staged, and `git add` treats that
    # as an error, so ONE stray file in the skeleton fails the entire config
    # sync on every instance whose .gitignore covers its extension. Skip those
    # instead: a file the instance ignores was never going to be committed
    # there, and aborting the sync over it helps nobody.
    #
    # Scar 2026-07-25: the skeleton tracks
    # plugins/prd-os/scripts/export-fable-mirror.sh.remediation.bak -- a backup
    # committed by accident -- and ASK_AI_consultant's .gitignore line 62 is
    # `*.bak`. One file, whole fleet. check-ignore runs once over the batch
    # rather than per file.
    if [ "${#plugin_paths[@]}" -gt 0 ]; then
      ignored_paths="$(
        printf '%s\n' "${plugin_paths[@]}" |
          git -C "$target" check-ignore --stdin 2>/dev/null || true
      )"
      stage_paths=()
      for relative in "${plugin_paths[@]}"; do
        if [ -n "$ignored_paths" ] &&
            printf '%s\n' "$ignored_paths" | grep -Fxq -- "$relative"; then
          continue
        fi
        stage_paths+=("$relative")
      done
      if [ "${#stage_paths[@]}" -gt 0 ]; then
        git -C "$target" add -- "${stage_paths[@]}" || return 1
      fi
    fi
  fi
}

guarded_commit() {
  local target="$1"
  local message="$2"
  local guard_dir original_hooks configured hook index_path rc
  guard_dir="$(mktemp -d)"
  index_path="$(git -C "$target" rev-parse --git-path index)"
  cp "$index_path" "$guard_dir/index.before" || {
    rm -r -- "$guard_dir"
    return 1
  }
  git -C "$target" diff --cached --name-only -z > "$guard_dir/allowed"

  original_hooks=""
  if [ "$MODEL_RUN" != "1" ]; then
    configured="$(git -C "$target" config --path --get core.hooksPath || true)"
    if [ -n "$configured" ]; then
        "$path/q-system/.q-system/scripts/claude-integrity-tripwire.py" \
        --root "$path" --quiet \
        --register ${TRIPWIRE_WROTE[@]+"${TRIPWIRE_WROTE[@]}"} >/dev/null 2>&1 ||
        echo "    WARN: could not sanction .claude/ tripwire writes (next tool call may revert this update)"
    fi

    # Sync plugins (copy contents, not directory, to avoid plugins/plugins/ nesting).
    # rsync instead of rm -rf + cp -R: --delete-excluded strips embedded .git dirs
    # and bytecode from the instance copy. A symlinked skeleton plugin (e.g.
    # memory-lifecycle -> standalone repo) used to materialize WITH its .git,
    # leaving every instance permanently dirty on plugins/<name> in git status.
    if [ -d "$SKELETON_PLUGIN_ROOT" ]; then
      mkdir -p "$path/plugins"
      while IFS= read -r -d '' plugin_name; do
        # .venv/ is a uv-built virtualenv, not source: uv writes a
        # `.gitignore` of `*` inside it, pyvenv.cfg pins it to ONE machine's
        # Python (home = /Users/<name>/... macos-aarch64), and nothing
        # launches it -- plugins/kipi-core/.mcp.json runs `uv run`, which
        # rebuilds it from the tracked uv.lock (measured: 52 packages, 37ms).
        # It was 107MB of the 112MB plugin tree, copied into 23 instances
        # where it could never work. --delete-excluded also clears the stale
        # copies already there. Pairs with test-kipi-update-build-artifacts.sh.
        # Flags derived from PLUGIN_COPY_EXCLUDES, never hand-listed here: this
        # site and the preflight filter are the two consumers that must agree,
        # and a hand-listed copy is how a plugin-root .env shipped fleet-wide
        # (ASK-772). Word-splitting the generator output is intended -- every
        # pattern is a shell-safe token by construction.
        # shellcheck disable=SC2046
        if ! rsync -a --delete --delete-excluded \
            $(plugin_copy_rsync_flags) \
            "$SKELETON_PLUGIN_ROOT/$plugin_name/" \
            "$path/plugins/$plugin_name/" 2>/dev/null; then
          CONFIG_FAILED=1
        fi
      done < <(managed_plugin_names)
    fi

    # Commit the config sync. The updater used to commit only $prefix/, leaving
    # .claude/ and plugins/ permanently dirty in every instance repo.
    if git -C "$path" rev-parse --git-dir >/dev/null 2>&1; then
      if ! ( cd "$path" &&
        if git ls-files --error-unmatch plugins/memory-lifecycle \
            >/dev/null 2>&1; then
          git rm -r -q --cached plugins/memory-lifecycle
        fi &&
        { stage_config_sync "$path" || { unstage_scope "$path" .claude/ plugins/; false; }; } &&
        if ! git diff --cached --quiet 2>/dev/null; then
          guarded_commit "$path" \
            "chore: sync .claude config + plugins from skeleton $(date +%Y-%m-%d)"
        fi
      ); then
        CONFIG_FAILED=1
      fi
    else
      CONFIG_FAILED=1
    fi

    if [ "$CONFIG_FAILED" != "0" ]; then
      abandon_instance "  ERROR: config sync did not reach a complete committed state" && continue
    fi

    echo "  Config synced"
  fi

  # Post-sync capability gate (structure/wiring/data diff, no test execution —
  # the FULL per-instance run is fleet-capability-verify.py's job). This is the
./lefthook.yml:52:        # and known AI-tooling caches. Allow .env.example / .env.sample / .env.template.
./lefthook.yml:73:          echo "Rename .env.example/.env.sample/.env.template if you meant to commit an example file."
./lefthook.yml.remediation.bak:36:        # and known AI-tooling caches. Allow .env.example / .env.sample / .env.template.
./lefthook.yml.remediation.bak:45:          echo "Rename .env.example/.env.sample/.env.template if you meant to commit an example file."
./plugins/prd-os/schemas/codebase_map.schema.json:112:      "description": "Names of .env / dotenv files found (e.g. .env, .env.example). Names only, never contents.",
./plugins/prd-os/tests/test_prd_map_runner.py:97:    (repo / ".env.example").write_text("SECRET=example\n")
./plugins/prd-os/tests/test_prd_map_runner.py:193:    assert ".env.example" in env_files
.gitignore
     1	# Secrets and local overrides
     2	.env
     3	*.env
     4	.claude/settings.local.json
     5	CLAUDE.local.md
     6	
     7	# Local security-remediation working dir (settings backups, scan dumps,
     8	# canary scripts) — contains real/old tokens, never belongs in this repo
     9	security-remediation/
    10	
    11	# OS
    12	.DS_Store
    13	Thumbs.db
    14	
    15	# Output (generated daily, don't commit)
    16	# ASK-605: the whole directory, not a list of extensions. auto-commit.py
    17	# declares q-system/output/ skip-because-gitignored; when that claim was
    18	# only true for *.html/*.json/*.log, a .md file there was invisible to the
    19	# hook AND blocked every fleet sync. Keep the negation for .gitkeep so the
    20	# directory survives a clone.
    21	q-system/output/**
    22	!q-system/output/.gitkeep
    23	q-system/output/*.html
    24	q-system/output/*.json
    25	q-system/output/*.log
    26	q-system/output/drafts/*
    27	q-system/output/lead-gen/*
    28	q-system/output/design-partner/*
    29	q-system/output/content-intel/*
    30	q-system/output/investor-updates/*
    31	q-system/output/marketing/linkedin/*
    32	!**/.gitkeep
    33	
    34	# Plans (auto-created by Claude Code)
    35	.claude/plans/
    36	
    37	# Agent claim-lock: per-working-tree local state. Committing it would share one
    38	# session's claim with every clone and the lock would refuse everyone forever.
    39	.linear-claims.json
    40	.linear-claims.json.guard
    41	
    42	# Session transcripts
    43	*.jsonl
    44	# ...but this one is not a transcript, it is the repo's own closure-evidence
    45	# ledger. test-updater-issue-sequence.py audits it against real commit ancestry,
    46	# and the blanket ignore above meant it never reached a fresh clone -- so in CI
    47	# every issue read as "no closure receipt" while passing locally off an untracked
    48	# working copy. Verified 2026-07-26: structural ids and timestamps only, no
    49	# free-text, safe for a public repo.
    50	!.prd-os/receipts.jsonl
    51	
    52	# SQLite databases (generated at runtime)
    53	*.db
    54	
    55	# Generated sites and build output
    56	sites/
    57	template-repo/
    58	
    59	# Bus data (daily pipeline, ephemeral)
    60	q-system/.q-system/agent-pipeline/bus/
    61	
    62	# Obsidian vault config (user-specific, not committed)
    63	q-system/.obsidian/
    64	
    65	# Raw source transcripts (instance-specific, 90-day retention)
    66	q-system/sources/*.md
    67	
    68	# Pruned markdown archives (historical, not loaded in context)
    69	q-system/memory/archives/
    70	
    71	# Session-scoped recall artifact for memory auto-capture (PRD prd-memory-autocapture)
    72	q-system/memory/.session-recall.json
    73	q-system/memory/.session-recall.json.lock
    74	q-system/memory/.session-recall.json.tmp-*
    75	
    76	# Node
    77	node_modules/
    78	
    79	# Python bytecode
    80	__pycache__/
openrsync: protocol version 29
rsync version 2.6.9 compatible

exec
/bin/zsh -lc "bash -c '
set -euo pipefail
source <(sed -n \"225,249p\" kipi-update.sh)
candidate=.env.example
printf \"rsync flags:\\n%s\\n\" \""'$(plugin_copy_rsync_flags | grep env)"
regex=$(plugin_copy_filter_regex)
if printf "%s'"\\n\" \""'$candidate" | grep -qE "$regex"; then
  echo "preflight/stager decision: EXCLUDED $candidate"
else
  echo "preflight/stager decision: INCLUDED $candidate"
fi
printf "repository policy: "
rg -o "Allow '"\\\\.env\\\\.example / \\\\.env\\\\.sample / \\\\.env\\\\.template\\\\.\" lefthook.yml
'

bash -c '
set -euo pipefail
source <(sed -n \"225,249p\" kipi-update.sh)
for candidate in .env .env.local nested/.env nested/.env.production normal.env config.txt; do
  if printf \"%s\" \""'$candidate" | grep -qE "$(plugin_copy_filter_regex)"; then decision=excluded; else decision=included; fi
  printf "%s => %s'"\\n\" \""'$candidate" "$decision"
done
'"'

git diff --check "'$(git merge-base HEAD origin/main)..HEAD
command -v shellcheck >/dev/null && shellcheck -x kipi-update.sh q-system/.q-system/scripts/test/test-kipi-update-plugin-excludes.sh q-system/.q-system/scripts/test/test-kipi-update-source-provenance.sh || true' in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-153
 succeeded in 0ms:
bash: line 4: plugin_copy_rsync_flags: command not found
rsync flags:

bash: line 5: plugin_copy_filter_regex: command not found
bash: line 4: plugin_copy_filter_regex: command not found
.env => excluded
bash: line 4: plugin_copy_filter_regex: command not found
.env.local => excluded
bash: line 4: plugin_copy_filter_regex: command not found
nested/.env => excluded
bash: line 4: plugin_copy_filter_regex: command not found
nested/.env.production => excluded
bash: line 4: plugin_copy_filter_regex: command not found
normal.env => excluded
bash: line 4: plugin_copy_filter_regex: command not found
config.txt => excluded

exec
/bin/zsh -lc "nl -ba kipi-update.sh | sed -n '218,255p'
command -v shellcheck || true
git diff --check "'$(git merge-base HEAD origin/main)..HEAD; echo "diff-check rc=$?"' in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-153
 succeeded in 0ms:
   218	# Each entry is "<rsync --exclude pattern>::<regex matching the same paths>".
   219	# Two representations, adjacent, because rsync globs and grep regexes are not
   220	# interconvertible in bash without a converter more fragile than the duplication
   221	# it removes. `::` is the delimiter, not `|`, because the regexes contain `|`.
   222	# test-kipi-update-plugin-excludes.sh drives BOTH consumers off this list and
   223	# fails if they disagree, so agreement is proven behaviourally rather than by
   224	# the two strings having been typed the same.
   225	PLUGIN_COPY_EXCLUDES=(
   226	  "/.git/::(^|/)\.git/"
   227	  "__pycache__/::(^|/)__pycache__/"
   228	  "*.pyc::\.pyc\$"
   229	  ".venv/::(^|/)\.venv/"
   230	  ".pytest_cache/::(^|/)\.pytest_cache/"
   231	  ".env::(^|/)\.env\$"
   232	  ".env.*::(^|/)\.env\."
   233	)
   234	
   235	plugin_copy_rsync_flags() {
   236	  local entry
   237	  for entry in "${PLUGIN_COPY_EXCLUDES[@]}"; do
   238	    printf -- '--exclude=%s\n' "${entry%%::*}"
   239	  done
   240	}
   241	
   242	plugin_copy_filter_regex() {
   243	  local entry out=""
   244	  for entry in "${PLUGIN_COPY_EXCLUDES[@]}"; do
   245	    out="${out:+$out|}${entry#*::}"
   246	  done
   247	  printf '%s\n' "$out"
   248	}
   249	
   250	SYSTEM_OWNED_PATHS=(
   251	  "q-system/memory/.sycophancy-monthly-stamp"
   252	  "q-system/.q-system/claude-integrity-baseline.json"
   253	  ".claude/state/stop-gate-firings.json"
   254	)
   255	# Skeleton-managed plugin dirs are appended at run time from the SAME
diff-check rc=0

exec
/bin/zsh -lc 'eval "$(sed -n '"'225,248p' kipi-update.sh)\"
candidate=.env.example
printf 'rsync flags:\\n%s\\n' \""'$(plugin_copy_rsync_flags | grep env)"
regex="$(plugin_copy_filter_regex)"
if printf '"'%s\\n' \""'$candidate" | grep -qE "$regex"; then
  echo "preflight/stager decision: EXCLUDED $candidate"
else
  echo "preflight/stager decision: INCLUDED $candidate"
fi
printf '"'repository policy: '
rg -o 'Allow \\.env\\.example / \\.env\\.sample / \\.env\\.template\\.' lefthook.yml

for candidate in .env .env.local nested/.env nested/.env.production normal.env config.txt; do
  if printf '%s' \""'$candidate" | grep -qE "$regex"; then decision=excluded; else decision=included; fi
  printf '"'%s => %s\\n' \""'$candidate" "$decision"
done' in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-153
 succeeded in 962ms:
rsync flags:
--exclude=.venv/
--exclude=.env
--exclude=.env.*
preflight/stager decision: EXCLUDED .env.example
repository policy: Allow .env.example / .env.sample / .env.template.
.env => excluded
.env.local => excluded
nested/.env => excluded
nested/.env.production => excluded
normal.env => included
config.txt => included

codex
## Finding

**MINOR**: The broad `.env.*` exclusion also removes non-secret `.env.example`, `.env.sample`, and `.env.template` files, contradicting the repository’s explicit policy permitting those templates.

- Location: `kipi-update.sh:232`
- Impact: No current plugin emits these files, so this is a bounded future-input issue. If one is added, unattended updates will exclude it and `--delete-excluded` can remove an existing instance copy.
- Reproducer:

```bash
eval "$(sed -n '225,248p' kipi-update.sh)"
candidate=.env.example
printf 'rsync flags:\n%s\n' "$(plugin_copy_rsync_flags | grep env)"
regex="$(plugin_copy_filter_regex)"
if printf '%s\n' "$candidate" | grep -qE "$regex"; then
  echo "preflight/stager decision: EXCLUDED $candidate"
else
  echo "preflight/stager decision: INCLUDED $candidate"
fi
printf 'repository policy: '
rg -o 'Allow \.env\.example / \.env\.sample / \.env\.template\.' lefthook.yml
```

Real output:

```text
rsync flags:
--exclude=.venv/
--exclude=.env
--exclude=.env.*
preflight/stager decision: EXCLUDED .env.example
repository policy: Allow .env.example / .env.sample / .env.template.
```

## What is sound

- Exact `.env`, nested `.env`, and `.env.local` paths are excluded while unrelated `normal.env` and `config.txt` remain included.
- `bash -n` passed for the updater and both affected tests.
- The capability manifest parses as valid JSON.
- `git diff --check` passed.
- The new test verifies that plugin synchronization actually ran before accepting the no-leak assertion, avoiding a vacuous pass.
- Full fixture tests could not run because the read-only sandbox rejected `mktemp`; I am not treating them as verified.

Fix first: preserve the explicitly permitted dotenv template names while continuing to exclude real `.env` override files.

FINDINGS:
minor|The broad .env.* exclusion also removes explicitly permitted non-secret dotenv template files such as .env.example.|kipi-update.sh:232
END FINDINGS
tokens used
40,983
## Finding

**MINOR**: The broad `.env.*` exclusion also removes non-secret `.env.example`, `.env.sample`, and `.env.template` files, contradicting the repository’s explicit policy permitting those templates.

- Location: `kipi-update.sh:232`
- Impact: No current plugin emits these files, so this is a bounded future-input issue. If one is added, unattended updates will exclude it and `--delete-excluded` can remove an existing instance copy.
- Reproducer:

```bash
eval "$(sed -n '225,248p' kipi-update.sh)"
candidate=.env.example
printf 'rsync flags:\n%s\n' "$(plugin_copy_rsync_flags | grep env)"
regex="$(plugin_copy_filter_regex)"
if printf '%s\n' "$candidate" | grep -qE "$regex"; then
  echo "preflight/stager decision: EXCLUDED $candidate"
else
  echo "preflight/stager decision: INCLUDED $candidate"
fi
printf 'repository policy: '
rg -o 'Allow \.env\.example / \.env\.sample / \.env\.template\.' lefthook.yml
```

Real output:

```text
rsync flags:
--exclude=.venv/
--exclude=.env
--exclude=.env.*
preflight/stager decision: EXCLUDED .env.example
repository policy: Allow .env.example / .env.sample / .env.template.
```

## What is sound

- Exact `.env`, nested `.env`, and `.env.local` paths are excluded while unrelated `normal.env` and `config.txt` remain included.
- `bash -n` passed for the updater and both affected tests.
- The capability manifest parses as valid JSON.
- `git diff --check` passed.
- The new test verifies that plugin synchronization actually ran before accepting the no-leak assertion, avoiding a vacuous pass.
- Full fixture tests could not run because the read-only sandbox rejected `mktemp`; I am not treating them as verified.

Fix first: preserve the explicitly permitted dotenv template names while continuing to exclude real `.env` override files.

FINDINGS:
minor|The broad .env.* exclusion also removes explicitly permitted non-secret dotenv template files such as .env.example.|kipi-update.sh:232
END FINDINGS

PR #162, opened 2026-08-14
The control fixture walks the skeleton the way the gate does (sp-505140ae)

assafkip:
## What

`repo-preflight.sh` gained a `_shipping()` filter in ASK-798 so a wired guard resolves to a path an instance could actually have. `copy_guards()` in `test-repo-preflight.sh` kept the old walk: `/.git` filtered, first `os.walk` hit, `break`.

In a checkout holding PR-review scratch dirs the two disagree. The builder writes the guard to `.pr31rev/r2/tree/.../audhd-lint.py`; the gate looks only at the shipped path. So the repo built to be COMPLIANT reads as absent.

## Measured, not argued

Primary checkout at `1830a72e`, before:

```
-------- 60 passed, 10 failed --------
FAIL ... absent=audhd-lint.py/auto-commit.py/auto-update.sh/batch-uniformity-lint.py...
```

Replicating the test's own walk prints exactly those four names resolving into scratch:

```
TEST picks:  .pr31rev/r2/tree/q-system/.q-system/scripts/audhd-lint.py
TEST picks:  .pr25rev/repo/q-system/hooks/auto-commit.py
TEST picks:  .pr25rev/repo/q-system/hooks/auto-update.sh
TEST picks:  .pr31rev/r2/tree/q-system/.q-system/scripts/batch-uniformity-lint.py
```

After, same checkout, scratch dirs still on disk, nothing deleted:

```
-------- 70 passed, 0 failed --------
```

## Why CI never caught it

CI clones clean. A clean tree has no scratch dirs for the walk to hit first, so it passes 70/70 at the same commit either way. This failure class is only visible from a working checkout, which is exactly where a human runs it and concludes the gate is broken.

## Not a loosening

Case 17's `noguards` fixture still builds a repo with zero guard files and is still REFUSED. The change only stops the compliant fixture from being non-compliant by accident.

It copies every shipping hit rather than `hits[0]`, because the gate accepts a guard when ANY resolved path exists. Matching that removes walk-order tie-break as a variable.

Closes sp-505140ae.

### DIFF
--- a/q-system/.q-system/scripts/test/test-repo-preflight.sh
+++ b/q-system/.q-system/scripts/test/test-repo-preflight.sh
@@ -173,6 +173,26 @@ fixture_git() {
 # guards it did not have -- and codex used that fixture as the proof that basename
 # comparison was hollow. A control repo has to be genuinely compliant, or every
 # refusal measured against it is measuring the wrong thing.
+#
+# THE FIXTURE MUST WALK THE SKELETON THE SAME WAY THE GATE DOES (sp-505140ae).
+# repo-preflight.sh grew a _shipping() filter in ASK-798 so a guard resolves to a
+# path an instance could actually have; this builder kept the old walk -- `/.git`
+# filtered, first os.walk hit, `break`. In a checkout holding PR-review scratch
+# dirs the two disagree: the builder writes .pr31rev/r2/tree/.../audhd-lint.py and
+# the gate looks only at the shipped path, so the CONTROL repo reads as absent.
+#
+# Measured 2026-08-14 in the primary checkout at 1830a72e: 60 passed / 10 failed,
+# every failure the same "absent=audhd-lint.py/auto-commit.py/auto-update.sh/..."
+# string -- and those names are exactly the guards whose first walk hit lands in
+# .pr31rev/ and .pr25rev/. The same commit in a clean worktree passes 70/70,
+# because a clean tree has no scratch dirs for the walk to find first. CI clones
+# clean, so CI is structurally blind to this and can never catch it.
+#
+# NOT A LOOSENING. Case 17's "noguards" fixture still builds a repo with no guard
+# files at all and is still REFUSED, which is what proves the gate kept its teeth.
+# This only stops the compliant fixture from being non-compliant by accident.
+# Copies EVERY shipping hit, not hits[0], because the gate accepts the guard if
+# ANY resolved path exists -- matching that removes the tie-break as a variable.
 copy_guards() {
   python3 - "$SKEL_SETTINGS" "$REPO" "$1" <<'GPY'
 import json, os, re, shutil, sys
@@ -183,16 +203,31 @@ for arr in hooks.values():
     for m in arr or []:
         for hk in m.get("hooks", []) or []:
             names |= set(re.findall(r"[\w.-]+\.(?:py|sh)", hk.get("command", "") or ""))
+
+# Kept byte-identical in intent to repo-preflight.sh's _shipping(). If that list
+# changes, this one changes with it or the control fixture silently rots again.
+SKIP_DIRS = {'.git', 'node_modules', '__pycache__', '.venv', '.pytest_cache',
+             'template-repo', 'worktrees'}
+
+
+def _shipping(dirname):
+    if dirname in SKIP_DIRS:
+        return False
+    for junk in ('.pr', '.wt-', '.fable-wt', '.sana-tmp'):
+        if dirname.startswith(junk):
+            return False
+    return True
+
+
 for rel in sorted(names):
-    for base, _dirs, files in os.walk(skel):
-        if "/.git" in base:
+    for base, dirs, files in os.walk(skel):
+        dirs[:] = [d for d in dirs if _shipping(d)]
+        if rel not in files:
             continue
-        if rel in files:
-            r = os.path.relpath(os.path.join(base, rel), skel)
-            t = os.path.join(dst, r)
-            os.makedirs(os.path.dirname(t), exist_ok=True)
-            shutil.copy2(os.path.join(base, rel), t)
-            break
+        r = os.path.relpath(os.path.join(base, rel), skel)
+        t = os.path.join(dst, r)
+        os.makedirs(os.path.dirname(t), exist_ok=True)
+        shutil.copy2(os.path.join(base, rel), t)
 GPY
 }
 


assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-162-20260814-152009.md` (109293 bytes).

```
FINDINGS:
minor|The test updates scratch-directory filtering but never creates a scratch directory, so clean CI does not exercise the behavior being fixed.|q-system/.q-system/scripts/test/test-repo-preflight.sh:223
END FINDINGS
```

--- reviewer output, last 54773 bytes of 109293 (full review at the path above) ---

    34	# credential and branch-protection checks at /bin/true. A skippable safety gate on
    35	# a client repo is worse than no gate, because it reads as protection to everyone
    36	# downstream of it.
    37	set -uo pipefail
    38	
    39	REPO_PATH="${1:-}"
    40	EXPECTED_REMOTE="${2:-}"
    41	
    42	# The skeleton this script SHIPPED FROM, derived from the script's own location
    43	# rather than from $PWD or an env var. The control-code check compares an
    44	# instance's copy against this one, so the reference has to follow the code; a
    45	# $PWD-derived root would compare a repo against itself whenever the caller
    46	# happened to be standing in it.
    47	SKELETON="$(cd "$(dirname "${BASH_SOURCE[0]}")/../../.." && pwd)"
    48	
    49	FAILED=0
    50	refuse() { FAILED=1; printf 'FAIL %s: %s\n' "$1" "$2"; }
    51	
    52	[ -n "$REPO_PATH" ] || { printf 'FAIL usage: repo-preflight.sh <repo-path> <expected-remote>\n'; exit 2; }
    53	[ -d "$REPO_PATH" ] || { printf 'FAIL repo-path: %s is not a directory\n' "$REPO_PATH"; exit 1; }
    54	
    55	# --- 0. client engagement repo -------------------------------------------
    56	# Founder decision 2026-08-13, verbatim: "no. unattended agents should not reach
    57	# a client repo."
    58	#
    59	# THIS REFUSES EVEN WHEN THE ROW IS OPTED IN, WHICH IS THE WHOLE POINT.
    60	# `dispatch.enabled` defaulting to absent is a DEFAULT, and a default is not a
    61	# refusal: it records that nobody has switched a repo on yet, and it evaporates
    62	# the moment anyone -- a person editing the registry, or some later script that
    63	# opts rows in automatically -- writes `true`. What is on the other side of that
    64	# edit is a self-merging loop (converge, code, PR, review rounds, auto-merge) with
    65	# no human in the path, pointed at work the founder is accountable to a CLIENT
    66	# for, where there is a person on the other end and no undo. So this is checked
    67	# ahead of the kill-switch and ahead of every network call: it is the one refusal
    68	# that no state of the target repo, and no registry field, can argue with.
    69	#
    70	# DERIVED FROM PATH SHAPE, NEVER FROM A LIST OF CLIENTS. Measured 2026-08-13
    71	# against instance-registry.json: 12 of 25 rows sit under an engagement root.
    72	# Naming those twelve here would be a list that goes stale the day the thirteenth
    73	# client is onboarded -- and the failure mode of a stale allowlist is that the new
    74	# client is the one that gets dispatched into. The shape is the fact: the founder's
    75	# own systems live at <root>/projects/<thing> under roots he owns, and client
    76	# engagements live under exactly these two. A repo added under one of them next
    77	# month is refused with no code change, which is the only property that makes this
    78	# hold over time.
    79	#
    80	# NOT AN ENV VAR, NOT A REGISTRY FIELD. Same reason `gh` is taken from PATH and the
    81	# dispatcher hardcodes $PREFLIGHT off $REPO: a knob here would be a documented way
    82	# to aim the client-repo gate at nothing while every log line still read normally.
    83	ENGAGEMENT_ROOTS="consulting intel"
    84	
    85	# BOTH THE REGISTRY'S PATH AND ITS RESOLVED TARGET ARE TESTED, AND EITHER ONE
    86	# MATCHING REFUSES. A symlink is the obvious way this check gets walked around by
    87	# accident: a row pointing at an innocent-looking path that resolves into an
    88	# engagement dir, or an engagement-shaped path that resolves elsewhere. Refusing on
    89	# either side means the answer does not depend on which of the two a later reader
    90	# happens to think is "the" path. Fail closed, like every other item in this file.
    91	RESOLVED_PATH="$(cd "$REPO_PATH" 2>/dev/null && pwd -P)" || RESOLVED_PATH=""
    92	for _root in $ENGAGEMENT_ROOTS; do
    93	  for _cand in "$REPO_PATH" "$RESOLVED_PATH"; do
    94	    [ -n "$_cand" ] || continue
    95	    case "$_cand" in
    96	      # THE ROOT ITSELF, NOT ONLY WHAT IS NESTED UNDER IT (ASK-754).
    97	      # This case previously PASSED, and the test suite asserted that it should:
    98	      # the earlier reading of the founder's wording was "the engagement instances
    99	      # NESTED UNDER consulting/projects/*", so <root>/projects/<x> refused and
   100	      # <root> did not. Measured 2026-08-14 against the repo that reading lets in,
   101	      # the ASK_AI_consultant row, whose path IS the consulting engagement root:
   102	      #   - `git ls-files clients/` -> 12 TRACKED files under clients/alma,
   103	      #     clients/portant, clients/restaurent. Client material is in this repo's
   104	      #     own history, so an agent-authored PR here is a client-facing diff.
   105	      #   - projects/ holds all 11 engagement repos as nested SEPARATE git repos,
   106	      #     and the root tracks 0 files under projects/ -- so check 5 (dirty) is
   107	      #     structurally blind to them while the agent still has filesystem reach.
   108	      #   - q-consult/ is 998 tracked files: the LIVE social publishing engine.
   109	      # The root is not a neutral parent, it is the container of every engagement
   110	      # plus a live outbound path. Refusing the nested repos while admitting the
   111	      # directory that holds them is a gate with a door beside it.
   112	      #
   113	      # WHY THIS WAS ONLY LATENT: preflight already refused this repo today, but on
   114	      # control-code/hooks/dirty/branch-protection -- every one of them curable by a
   115	      # `kipi update`, a commit, and a protection toggle. An incidental refusal is
   116	      # not a rule, and it evaporates the day somebody tidies the repo.
   117	      #
   118	      # STILL DERIVED FROM SHAPE, STILL NO LIST. Same ENGAGEMENT_ROOTS constant,
   119	      # one extra glob. It discriminates: only a path whose LAST component is an
   120	      # engagement root matches, so the founder's own persona and product roots are
   121	      # untouched -- which the OWNPROJ and NEARMISS cases both assert.
   122	      #
   123	      # NO ABSOLUTE HOME PATHS AND NO INSTANCE NAMES IN THIS COMMENT. The skeleton
   124	      # sweep in validate-separation.py greps every q-system/ file for an absolute
   125	      # home-directory prefix and for a fleet instance name, and this file ships to
   126	      # every instance. The first draft of this comment spelled out both and turned
   127	      # CI red; the second spelled them out again while explaining not to, because a
   128	      # text check cannot tell a rule from a mention of the rule. Describe, do not
   129	      # quote.
   130	      # TWO SHAPES, TWO REASONS, BECAUSE THE REASON IS READ BY A HUMAN. The root and
   131	      # the repos nested under it are both refused, but they are refused for
   132	      # different facts, and one message cannot state both without lying about one
   133	      # of them. Codex round 2 caught the first cut telling the founder that
   134	      # <root> is "under <root>/projects/", which is false and unfalsifiable-looking
   135	      # in a log. This line is what he reads in the daily digest's "tried, could not
   136	      # be worked" section, so a refusal he cannot tell from an empty queue -- or one
   137	      # that misdescribes itself -- is the defect this whole change is about.
   138	      */"$_root")
   139	        printf 'FAIL client-repo: %s IS the %s engagement root; it holds the client engagement repos plus client material of its own, so unattended dispatch is not allowed there, even when dispatch.enabled is true -- a supervised founder-initiated run is the way in\n' "$REPO_PATH" "$_root"
   140	        printf 'REFUSED %s\n' "$REPO_PATH"
   141	        exit 1 ;;
   142	      */"$_root"/projects/*)
   143	        printf 'FAIL client-repo: %s is a client engagement repo (under %s/projects/); unattended dispatch is not allowed there, even when dispatch.enabled is true -- a supervised founder-initiated run is the way in\n' "$REPO_PATH" "$_root"
   144	        printf 'REFUSED %s\n' "$REPO_PATH"
   145	        exit 1 ;;
   146	    esac
   147	  done
   148	done
   149	
   150	# --- 1. kill-switch -------------------------------------------------------
   151	# Checked FIRST and short-circuits everything, including the network calls. A
   152	# founder or a client dropping this file is saying "stay out of here", and the
   153	# correct response to that is to stop, not to finish auditing the repo first. One
   154	# `touch` halts dispatch into a repo with no code change, no registry edit and no
   155	# deploy -- which is the property you want at 2am.
   156	if [ -f "$REPO_PATH/.kipi-no-dispatch" ]; then
   157	  printf 'FAIL kill-switch: %s/.kipi-no-dispatch is present, this repo is off limits\n' "$REPO_PATH"
   158	  exit 1
   159	fi
   160	
   161	# --- 2. control-code ------------------------------------------------------
   162	# The dispatcher would run the WORKER COPY THAT LIVES IN THE TARGET REPO, and
   163	# `kipi update` is manual, so an instance can be arbitrarily far behind. Running a
   164	# stale copy means running control code nobody reviewed, on a loop that merges its
   165	# own PRs -- an instance frozen before the stale-checkout refusal or the
   166	# sensitive-path guard has neither. Byte equality against the skeleton is a blunt
   167	# rule on purpose: "close enough" has no meaning for code that self-merges.
   168	WORKER_REL="q-system/.q-system/scripts/linear-worker.sh"
   169	if [ ! -f "$REPO_PATH/$WORKER_REL" ]; then
   170	  refuse "control-code" "$REPO_PATH/$WORKER_REL is missing, the repo carries no worker to run"
   171	elif [ ! -f "$SKELETON/$WORKER_REL" ]; then
   172	  refuse "control-code" "the skeleton reference $SKELETON/$WORKER_REL is missing, so drift cannot be judged"
   173	else
   174	  have="$(shasum -a 256 "$REPO_PATH/$WORKER_REL" 2>/dev/null | awk '{print $1}')"
   175	  want="$(shasum -a 256 "$SKELETON/$WORKER_REL"  2>/dev/null | awk '{print $1}')"
   176	  if [ -z "$have" ] || [ -z "$want" ]; then
   177	    refuse "control-code" "could not hash the worker on both sides, so drift is unknown"
   178	  elif [ "$have" != "$want" ]; then
   179	    # THE REMEDY NAMED HERE USED TO BE THE FLEET UPDATER, AND THAT WAS WRONG
   180	    # (ASK-755). The updater is a fleet-wide rsync WITH a delete flag: to fix one
   181	    # file in one repo it walks every registered instance and can remove anything
   182	    # on the way (2026-08-07: voicekit deleted from 19 instances). A refusal that
   183	    # names a tool with a blast radius three orders of magnitude wider than the
   184	    # defect is an invitation to overcorrect at 2am. control-file-propagate.py
   185	    # does the same hash classification for ONE file in ONE repo, copy only, and
   186	    # refuses anything it cannot prove came from this skeleton.
   187	    refuse "control-code" "linear-worker.sh differs from the skeleton (${have:0:12} vs ${want:0:12}); fix with: python3 $SKELETON/q-system/.q-system/scripts/control-file-propagate.py --target $REPO_PATH --file $WORKER_REL --apply"
   188	  fi
   189	fi
   190	
   191	# --- 3. hooks -------------------------------------------------------------
   192	# The enforcement layer is hooks, not prose (the repo's own rule: a prompt cannot
   193	# enforce behaviour). A repo missing a hook EVENT the skeleton wires is running an
   194	# agent with fewer blocking guards than the machine dispatching into it.
   195	#
   196	# Compared against the SKELETON'S OWN settings rather than a hand-written list of
   197	# event names, for the same reason Piece B enumerates exits from source: a hand
   198	# list cannot notice the day a seventh event class is added.
   199	#
   200	# THE REFERENCE IS settings-template.json, NOT THE SKELETON'S RUNTIME settings.json
   201	# (ASK-755). Those two files are not the same claim. `kipi update` builds every
   202	# instance's settings.json from the TEMPLATE; the skeleton's own runtime file is
   203	# what the skeleton runs on itself, and it deliberately carries hooks that must
   204	# never ship -- settings-template-sync-check.py is SKELETON_ONLY by its own
   205	# design, because it compares the two settings files and an instance has no
   206	# template to compare against.
   207	#
   208	# Measured 2026-08-14, both opted-in repos: against the runtime file each was
   209	# "missing" exactly settings-template-sync-check.py, and against the template each
   210	# was at FULL parity. So this check was unsatisfiable for every instance in the
   211	# fleet at once -- no `kipi update` run could ever have cleared it, because the
   212	# thing it demanded is the one hook update refuses to ship. A gate nothing can
   213	# pass is not strict, it is broken, and it reads as a fleet-wide instance defect.
   214	#
   215	# NOT A WEAKENING, AND THE NUMBERS ARE THE ARGUMENT. The two guard sets are 41 and
   216	# 41. Switching the reference DROPS one requirement (settings-template-sync-check.py,
   217	# which can never be satisfied) and ADDS one (instance-automation-guard.py, which
   218	# is FLEET_ONLY: it ships to instances and the skeleton self-detects and no-ops, so
   219	# the old reference could not require it and the new one does). The hook EVENT sets
   220	# of the two files are identical, so nothing changes there.
   221	SETTINGS_REL=".claude/settings.json"
   222	# Fall back to the runtime file when no template exists, so a skeleton that has
   223	# not adopted the template still gets a comparison rather than a silent pass.
   224	SKEL_SETTINGS="$SKELETON/settings-template.json"
   225	[ -f "$SKEL_SETTINGS" ] || SKEL_SETTINGS="$SKELETON/$SETTINGS_REL"
   226	if [ ! -f "$REPO_PATH/$SETTINGS_REL" ]; then
   227	  refuse "hooks" "$REPO_PATH/$SETTINGS_REL is missing, so no hook fires in this repo"
   228	else
   229	  MISSING="$(python3 - "$SKEL_SETTINGS" "$REPO_PATH/$SETTINGS_REL" "$SKELETON" "$REPO_PATH" <<'PY' 2>/dev/null
   230	import json, sys
   231	try:
   232	    skel = json.load(open(sys.argv[1])).get("hooks", {})
   233	    repo = json.load(open(sys.argv[2])).get("hooks", {})
   234	except Exception as e:
   235	    print("UNREADABLE:%s" % e)
   236	    sys.exit(0)
   237	# An event present but wired to nothing is the same hole as an absent event.
   238	# EVENT NAMES ARE NOT PARITY (codex finding-3). A target that keeps a nonempty
   239	# hook under every event while dropping a blocking guard still had every event
   240	# name present, so the first version of this passed it. What is being claimed is
   241	# that the guards the skeleton runs also run there, so compare the SCRIPTS.
   242	import re
   243	def scripts(h):
   244	    out = set()
   245	    for arr in h.values():
   246	        for matcher in arr or []:
   247	            for hk in matcher.get("hooks", []) or []:
   248	                for m in re.findall(r"[\w.-]+\.(?:py|sh)", hk.get("command", "") or ""):
   249	                    out.add(m)
   250	    return out
   251	missing_events = sorted(set(skel) - {k for k, v in repo.items() if v})
   252	missing_scripts = sorted(scripts(skel) - scripts(repo))
   253	# A NAME IS NOT AN EXECUTABLE. Comparing basenames pulled out of settings.json
   254	# passed a repo that copied settings.json and none of the guard files -- the
   255	# green fixture in this test suite was itself that shape. Resolve each guard the
   256	# skeleton wires and require the file to actually be present in the target.
   257	import os
   258	# A GUARD IS ABSENT ONLY IF NO SKELETON PATH FOR IT EXISTS IN THE TARGET.
   259	#
   260	# The previous rule took hits[:1] -- the first os.walk hit anywhere under the
   261	# skeleton -- and demanded THAT exact relative path in the instance. The skeleton
   262	# root also holds PR review worktrees (.pr31rev/, .pr25rev/) and template-repo/,
   263	# none of which ship to any instance, so the first hit routinely resolved to a
   264	# path no instance could ever have.
   265	#
   266	# Measured 2026-08-14 against both dispatch-enabled repos by replicating this block
   267	# verbatim: each reported 36 guards absent, 35 of them PHANTOM -- present in the
   268	# instance at a different skeleton path, e.g. a lint script resolving to its copy
   269	# inside a PR review worktree instead of its shipped location. Exactly ONE was
   270	# real. os.walk order is not stable across machines either, so the same repo could
   271	# pass or fail depending on directory iteration.
   272	#
   273	# NO INSTANCE NAMES ABOVE, DELIBERATELY. This file ships to every instance, so
   274	# validate-separation Gate 1.2 sweeps it for live instance names and the first cut
   275	# of this comment named the two repos and turned CI red. The measurement is the
   276	# durable part; the names belong in the PR body. A text check cannot tell a rule
   277	# from a mention of one, which is why this warning names no repo either.
   278	#
   279	# This is NOT a loosening. The check still demands the guard file exist in the
   280	# target; it stops demanding it exist at a path that only ever existed inside a
   281	# scratch worktree. After this fix both repos still REFUSE -- on the one real
   282	# absence -- which is the proof it did not go soft.
   283	SKIP_DIRS = {'.git', 'node_modules', '__pycache__', '.venv', '.pytest_cache',
   284	             'template-repo', 'worktrees'}
   285	
   286	
   287	def _shipping(dirname):
   288	    """Directories that never reach an instance, so never define the guard path."""
   289	    if dirname in SKIP_DIRS:
   290	        return False
   291	    # Review/scratch worktrees this repo creates at its own root. `.q-system` is a
   292	    # real shipped directory, so dotted names are filtered by prefix, not wholesale.
   293	    for junk in ('.pr', '.wt-', '.fable-wt', '.sana-tmp'):
   294	        if dirname.startswith(junk):
   295	            return False
   296	    return True
   297	
   298	
   299	skel_root, repo_root = sys.argv[3], sys.argv[4]
   300	absent = []

exec
/bin/zsh -lc "nl -ba q-system/.q-system/scripts/repo-preflight.sh | sed -n '280,350p'
nl -ba q-system/.q-system/scripts/test/test-repo-preflight.sh | sed -n '930,1030p'
git show origin/main:q-system/.q-system/scripts/test/test-repo-preflight.sh | nl -ba | sed -n '165,235p'
git show --stat --format=fuller HEAD
git show --format=fuller --no-ext-diff HEAD -- q-system/.q-system/scripts/test/test-repo-preflight.sh | sed -n '1,180p'" in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-162
 succeeded in 0ms:
   280	# target; it stops demanding it exist at a path that only ever existed inside a
   281	# scratch worktree. After this fix both repos still REFUSE -- on the one real
   282	# absence -- which is the proof it did not go soft.
   283	SKIP_DIRS = {'.git', 'node_modules', '__pycache__', '.venv', '.pytest_cache',
   284	             'template-repo', 'worktrees'}
   285	
   286	
   287	def _shipping(dirname):
   288	    """Directories that never reach an instance, so never define the guard path."""
   289	    if dirname in SKIP_DIRS:
   290	        return False
   291	    # Review/scratch worktrees this repo creates at its own root. `.q-system` is a
   292	    # real shipped directory, so dotted names are filtered by prefix, not wholesale.
   293	    for junk in ('.pr', '.wt-', '.fable-wt', '.sana-tmp'):
   294	        if dirname.startswith(junk):
   295	            return False
   296	    return True
   297	
   298	
   299	skel_root, repo_root = sys.argv[3], sys.argv[4]
   300	absent = []
   301	for rel in sorted(scripts(skel)):
   302	    hits = []
   303	    for base, dirs, files in os.walk(skel_root):
   304	        dirs[:] = [d for d in dirs if _shipping(d)]
   305	        if rel in files:
   306	            hits.append(os.path.relpath(os.path.join(base, rel), skel_root))
   307	    # Sorted so the reported path is deterministic rather than walk-order dependent.
   308	    hits.sort()
   309	    if hits and not any(os.path.isfile(os.path.join(repo_root, h)) for h in hits):
   310	        absent.append(hits[0])
   311	parts = []
   312	if missing_events:
   313	    parts.append("events=" + "/".join(missing_events))
   314	if missing_scripts:
   315	    parts.append("guards=" + "/".join(missing_scripts[:6]) + ("..." if len(missing_scripts) > 6 else ""))
   316	if absent:
   317	    parts.append("absent=" + "/".join(os.path.basename(a) for a in absent[:6]) + ("..." if len(absent) > 6 else ""))
   318	print(";".join(parts))
   319	PY
   320	)"
   321	  case "$MISSING" in
   322	    UNREADABLE:*) refuse "hooks" "settings.json could not be parsed (${MISSING#UNREADABLE:})" ;;
   323	    "")           : ;;
   324	    *)            refuse "hooks" "hook parity with the skeleton is incomplete: $MISSING" ;;
   325	  esac
   326	fi
   327	
   328	# --- 4. remote ------------------------------------------------------------
   329	# The registry row must PIN the remote, and origin must match it exactly.
   330	#
   331	# ABSENCE IS NOT CONSENT: a row with no pinned remote refuses rather than trusting
   332	# whatever origin happens to say. Otherwise a mistyped path, a repo moved on disk,
   333	# or a re-pointed origin silently redirects an agent's push to somebody else's
   334	# GitHub repo, and every log line still reads normally.
   335	if [ -z "$EXPECTED_REMOTE" ]; then
   336	  refuse "remote" "the registry row pins no expected_remote; an unpinned repo is never entered"
   337	else
   338	  ACTUAL_REMOTE="$(git -C "$REPO_PATH" remote get-url origin 2>/dev/null)"
   339	  if [ -z "$ACTUAL_REMOTE" ]; then
   340	    refuse "remote" "no origin remote is configured in $REPO_PATH"
   341	  elif [ "$ACTUAL_REMOTE" != "$EXPECTED_REMOTE" ]; then
   342	    refuse "remote" "origin is $ACTUAL_REMOTE but the registry pins $EXPECTED_REMOTE"
   343	  fi
   344	fi
   345	
   346	# --- 5. dirty -------------------------------------------------------------
   347	# Uncommitted work in a client repo belongs to a human. An agent that branches,
   348	# commits and opens a PR from that tree sweeps their work-in-progress into its own
   349	# diff, and the first anyone hears about it is a client-facing PR.
   350	if ! git -C "$REPO_PATH" rev-parse --git-dir >/dev/null 2>&1; then
   930	# be refused" -- it holds client material in its own git history.
   931	ROOTREPO="$(make_good_repo_at rootrepo consulting)"
   932	OUT="$(run_preflight "$ROOTREPO" "https://github.com/assafkip/rootrepo.git")"
   933	{ [ "$(pf_rc "$ROOTREPO" "https://github.com/assafkip/rootrepo.git")" != "0" ] \
   934	    && printf '%s' "$OUT" | grep -q 'client-repo'; } \
   935	  && ok "the engagement ROOT itself is refused, and names client-repo" \
   936	  || bad "THE DEFECT: the engagement root -- which holds tracked client material and every nested engagement -- was accepted: $OUT"
   937	
   938	# THE DISCRIMINATION CASE FOR THE NEW GLOB, kept adjacent to it. */<root> must match
   939	# only when <root> is the LAST component. A founder root that merely CONTAINS an
   940	# engagement root name as a prefix or suffix is not an engagement root, and a rule
   941	# that refused those would be the over-match that gets a gate switched off.
   942	NEARMISS="$(make_good_repo_at nearmiss consulting-notes)"
   943	[ "$(pf_rc "$NEARMISS" "https://github.com/assafkip/nearmiss.git")" = "0" ] \
   944	  && ok "a path whose last component merely CONTAINS an engagement root name still passes" \
   945	  || bad "the new root glob over-matched on a substring: $(run_preflight "$NEARMISS" "https://github.com/assafkip/nearmiss.git")"
   946	
   947	# NO BYPASS SURFACE, same rule the dispatcher is held to by case 8.
   948	grep -qE '^ENGAGEMENT_ROOTS="[a-z ]+"$' "$PREFLIGHT" \
   949	  && ok "the engagement roots are a literal in the script, not an override" \
   950	  || bad "the engagement roots are not a plain literal, so they may be overridable"
   951	grep -nE 'ENGAGEMENT_ROOTS="?\$\{|KIPI_[A-Z_]*ENGAGEMENT|KIPI_[A-Z_]*CLIENT' "$PREFLIGHT" \
   952	  && bad "an env var can redefine what counts as a client repo" \
   953	  || ok "no env var can redefine what counts as a client repo"
   954	
   955	echo
   956	echo "== 20. the client-repo refusal is ABSENT from the pick list, and SAID (ASK-741) =="
   957	# End to end through the dispatcher's own selection, not the preflight in isolation.
   958	# Both rows are opted in; only the shape differs.
   959	REG5="$WORK/reg5.json"
   960	mk_registry "$REG5" "newclient|$NEWCLIENT|true" "ownproj|$OWNPROJ|true"
   961	# stdout is the SELECTION; stderr carries say(). Merging them makes `grep newclient`
   962	# match the refusal REASON, which reports a correctly-refused repo as selected.
   963	PICKS5="$(KIPI_DISPATCH_REGISTRY="$REG5" KIPI_DISPATCH_CURSOR="$WORK/cursor5" bash "$HARNESS" 2>"$WORK/say5.log")"
   964	printf '%s' "$PICKS5" | grep -q 'ownproj' \
   965	  && ok "the opted-in NON-client repo IS selected (so the absence below means something)" \
   966	  || bad "the non-client control never reached the pick list, so this section proves nothing"
   967	printf '%s' "$PICKS5" | grep -q 'newclient' \
   968	  && bad "THE DEFECT: an opted-in client engagement repo is SELECTED for unattended dispatch" \
   969	  || ok "an opted-in client engagement repo is ABSENT from the pick list"
   970	
   971	# A SILENT REFUSAL IS THE DEFECT, NOT THE FIX. The founder has to be able to tell a
   972	# refusal from an empty queue, so the reason is asserted, not just the absence.
   973	grep -q 'FAIL client-repo' "$WORK/say5.log" \
   974	  && ok "the run output states WHY the repo was refused" \
   975	  || bad "the refusal is silent in the run output: $(cat "$WORK/say5.log")"
   976	# The counted shape daily-linear-digest.py's third section scrapes.
   977	grep -qE '[0-9]+ repo\(s\) REFUSED as client engagement repos' "$WORK/say5.log" \
   978	  && ok "the run output carries the counted line the daily digest reads" \
   979	  || bad "no counted line, so the digest's third section cannot show this: $(cat "$WORK/say5.log")"
   980	
   981	echo
   982	echo "== 21. MUTATION: break the derivation and case 20 must go RED (ASK-741) =="
   983	# A guard whose test stays green with the guard removed is decoration.
   984	# THE MUTANT MUST SIT AT THE SAME DEPTH AS THE REAL SCRIPT. repo-preflight.sh
   985	# derives SKELETON from its own ${BASH_SOURCE[0]} as "../../.." -- deliberately, so
   986	# the control-code reference follows the code rather than $PWD. A mutant dropped at
   987	# $WORK/mutant-preflight.sh therefore resolves SKELETON to a directory three levels
   988	# above mktemp, finds no linear-worker.sh there, and refuses on CONTROL-CODE. It was
   989	# measured doing exactly that: the mutant "still refused the client repo" and the
   990	# case read SURVIVED, when in truth the derivation under test had never run. A false
   991	# SURVIVED is worse than a false KILL -- it reports a load-bearing guard as
   992	# decorative. So the mutant gets a skeleton of its own, at the right depth.
   993	MUTSKEL="$WORK/mutskel"
   994	mkdir -p "$MUTSKEL/q-system/.q-system/scripts" "$MUTSKEL/.claude"
   995	cp "$SKEL_WORKER" "$MUTSKEL/q-system/.q-system/scripts/linear-worker.sh"
   996	cp "$SKEL_SETTINGS" "$MUTSKEL/.claude/settings.json"
   997	copy_guards "$MUTSKEL"
   998	MUTPF="$MUTSKEL/q-system/.q-system/scripts/repo-preflight.sh"
   999	sed 's|^ENGAGEMENT_ROOTS=.*|ENGAGEMENT_ROOTS="zzzz-no-such-root"|' "$PREFLIGHT" > "$MUTPF"
  1000	# VALIDATE THE MUTANT APPLIED. A sed that matched nothing produces a byte-identical
  1001	# file and a "SURVIVED" that means the edit never happened, not that the guard is weak.
  1002	if cmp -s "$MUTPF" "$PREFLIGHT"; then
  1003	  bad "the mutation changed nothing -- the mutant is not a mutant, so this proves nothing"
  1004	elif grep -q '^ENGAGEMENT_ROOTS="zzzz-no-such-root"$' "$MUTPF"; then
  1005	  ok "the mutant applied (the engagement roots were replaced with a root nothing is under)"
  1006	  MUTH2="$WORK/select-mutant-pf.sh"
  1007	  {
  1008	    echo 'set -uo pipefail'
  1009	    echo 'say() { printf "SAY %s\n" "$*" >&2; }'
  1010	    echo "REPO=\"$REPO\""
  1011	    echo "PREFLIGHT=\"$MUTPF\""
  1012	    awk '/^cursor_get\(\) \{/,/^\}/'       "$DISPATCH"
  1013	    awk '/^cursor_set\(\) \{/,/^\}/'       "$DISPATCH"
  1014	    awk '/^fleet_candidates\(\) \{/,/^\}/' "$DISPATCH"
  1015	    awk '/^rotation\(\) \{/,/^\}/'         "$DISPATCH"
  1016	    awk '/^pick_list\(\) \{/,/^\}/'        "$DISPATCH"
  1017	    echo 'pick_list'
  1018	  } > "$MUTH2"
  1019	  MP="$(KIPI_DISPATCH_REGISTRY="$REG5" KIPI_DISPATCH_CURSOR="$WORK/cursorM2" bash "$MUTH2" 2>/dev/null)"
  1020	  printf '%s' "$MP" | grep -q 'newclient' \
  1021	    && ok "with the derivation broken the client repo REAPPEARS -- case 20 is load-bearing" \
  1022	    || bad "the mutant still refused the client repo, so case 20 is not driven by the derivation"
  1023	else
  1024	  bad "the mutant was written but does not carry the expected replacement"
  1025	fi
  1026	
  1027	echo
  1028	echo "== 22. an API REFUSAL and a TOOTHLESS branch are different refusals (ASK-755) =="
  1029	# The two used to print the same sentence. Both still refuse -- that part was
  1030	# never wrong -- but "requires NO review and NO status check" names a GitHub
   165	      echo "FATAL: fixture_git refused a live or empty path: '$d'" >&2; exit 1 ;;
   166	  esac
   167	  [ -d "$d/.git" ] || { echo "FATAL: fixture_git: '$d' is not a fixture repo" >&2; exit 1; }
   168	  git -C "$d" "$@"
   169	}
   170	
   171	# Copy every hook script the skeleton actually wires into a fixture. THE FIXTURE
   172	# WAS THE BUG: it used to copy settings.json alone, so the "all-green" repo named
   173	# guards it did not have -- and codex used that fixture as the proof that basename
   174	# comparison was hollow. A control repo has to be genuinely compliant, or every
   175	# refusal measured against it is measuring the wrong thing.
   176	copy_guards() {
   177	  python3 - "$SKEL_SETTINGS" "$REPO" "$1" <<'GPY'
   178	import json, os, re, shutil, sys
   179	settings, skel, dst = sys.argv[1], sys.argv[2], sys.argv[3]
   180	hooks = json.load(open(settings)).get("hooks", {})
   181	names = set()
   182	for arr in hooks.values():
   183	    for m in arr or []:
   184	        for hk in m.get("hooks", []) or []:
   185	            names |= set(re.findall(r"[\w.-]+\.(?:py|sh)", hk.get("command", "") or ""))
   186	for rel in sorted(names):
   187	    for base, _dirs, files in os.walk(skel):
   188	        if "/.git" in base:
   189	            continue
   190	        if rel in files:
   191	            r = os.path.relpath(os.path.join(base, rel), skel)
   192	            t = os.path.join(dst, r)
   193	            os.makedirs(os.path.dirname(t), exist_ok=True)
   194	            shutil.copy2(os.path.join(base, rel), t)
   195	            break
   196	GPY
   197	}
   198	
   199	make_good_repo() {
   200	  # TWO `local` LINES, NOT ONE. `local name="$1" dir="$WORK/$name"` expands every
   201	  # word before it assigns any of them, so $name is still unbound when $dir is
   202	  # built -- under `set -u` the function aborts and returns an EMPTY path. Every
   203	  # fixture then pointed at "", the registry rows were dropped as pathless, and the
   204	  # round-robin and mutation cases passed while testing nothing at all.
   205	  local name="$1"
   206	  local dir="$WORK/$name"
   207	  [ -n "$name" ] && [ -n "$WORK" ] || { echo "FATAL: make_good_repo got an empty name/WORK" >&2; exit 1; }
   208	  mkdir -p "$dir/q-system/.q-system/scripts" "$dir/.claude"
   209	  cp "$SKEL_WORKER" "$dir/q-system/.q-system/scripts/linear-worker.sh"
   210	  cp "$SKEL_SETTINGS" "$dir/.claude/settings.json"
   211	  # $2 = "noguards" builds the deliberately non-compliant shape for case 17.
   212	  [ "${2:-}" = "noguards" ] || copy_guards "$dir"
   213	  git init -q -b main "$dir"
   214	  ( git -C "$dir" config user.email t@e.com
   215	    git -C "$dir" config user.name t
   216	    git -C "$dir" add -A
   217	    git -C "$dir" commit -qm init
   218	    git -C "$dir" remote add origin "https://github.com/assafkip/$name.git" ) >/dev/null 2>&1
   219	  printf '%s' "$dir"
   220	}
   221	
   222	# Exit 0 = the repo may be entered. Any non-zero = refuse.
   223	run_preflight() { bash "$PREFLIGHT" "$1" "${2:-https://github.com/assafkip/$(basename "$1").git}" 2>&1; }
   224	pf_rc() { bash "$PREFLIGHT" "$1" "${2:-https://github.com/assafkip/$(basename "$1").git}" >/dev/null 2>&1; echo $?; }
   225	
   226	echo "== 1. the control: a repo that passes all seven items is ACCEPTED =="
   227	# THIS CASE IS THE POINT OF THE WHOLE STUB DESIGN. If it fails, every "refused"
   228	# assertion below is green for the wrong reason and proves nothing.
   229	GOOD="$(make_good_repo good)"
   230	OUT="$(run_preflight "$GOOD")"
   231	if [ "$(pf_rc "$GOOD")" = "0" ]; then
   232	  ok "an all-green repo passes preflight (so the refusals below mean something)"
   233	else
   234	  bad "the all-green control was REFUSED, so every refusal below is unproven: $OUT"
   235	fi
commit 1086938c2637129cb59553886f759ccba2c82c4f
AuthorDate: Fri Aug 14 15:19:34 2026 -0700
CommitDate: Fri Aug 14 15:19:34 2026 -0700

    test(preflight): the control fixture walks the skeleton the way the gate does (sp-505140ae)
    
    repo-preflight.sh gained a _shipping() filter in ASK-798 so a wired guard
    resolves to a path an instance could actually have. copy_guards() in the test
    kept the old walk: `/.git` filtered, first os.walk hit, break. In a checkout
    holding PR-review scratch dirs the two disagree -- the builder writes the guard
    to .pr31rev/r2/tree/.../audhd-lint.py and the gate looks only at the shipped
    path, so the repo built to be COMPLIANT reads as absent.
    
    Measured in the primary checkout at 1830a72e: 60 passed / 10 failed, every
    failure carrying the same absent=audhd-lint.py/auto-commit.py/auto-update.sh/
    batch-uniformity-lint.py string. Replicating the test's walk prints exactly
    those four resolving into .pr31rev/ and .pr25rev/. The same commit in a clean
    worktree passes 70/70, because a clean tree has no scratch dirs to hit first.
    CI clones clean, so CI is structurally blind to this class and never sees it.
    
    After: 70 passed / 0 failed in the primary checkout, with the scratch dirs
    still on disk. Nothing was deleted to get here.
    
    Not a loosening. Case 17's noguards fixture still builds a repo with zero guard
    files and is still REFUSED. The change only stops the compliant fixture from
    being non-compliant by accident. It copies every shipping hit rather than
    hits[0], because the gate accepts the guard when ANY resolved path exists, and
    matching that removes walk-order tie-break as a variable.
    
    Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
    Claude-Session: https://claude.ai/code/session_0151EYtGH6at3XTWqbBHAxEi

 .../.q-system/scripts/test/test-repo-preflight.sh  | 51 ++++++++++++++++++----
 1 file changed, 43 insertions(+), 8 deletions(-)
commit 1086938c2637129cb59553886f759ccba2c82c4f
AuthorDate: Fri Aug 14 15:19:34 2026 -0700
CommitDate: Fri Aug 14 15:19:34 2026 -0700

    test(preflight): the control fixture walks the skeleton the way the gate does (sp-505140ae)
    
    repo-preflight.sh gained a _shipping() filter in ASK-798 so a wired guard
    resolves to a path an instance could actually have. copy_guards() in the test
    kept the old walk: `/.git` filtered, first os.walk hit, break. In a checkout
    holding PR-review scratch dirs the two disagree -- the builder writes the guard
    to .pr31rev/r2/tree/.../audhd-lint.py and the gate looks only at the shipped
    path, so the repo built to be COMPLIANT reads as absent.
    
    Measured in the primary checkout at 1830a72e: 60 passed / 10 failed, every
    failure carrying the same absent=audhd-lint.py/auto-commit.py/auto-update.sh/
    batch-uniformity-lint.py string. Replicating the test's walk prints exactly
    those four resolving into .pr31rev/ and .pr25rev/. The same commit in a clean
    worktree passes 70/70, because a clean tree has no scratch dirs to hit first.
    CI clones clean, so CI is structurally blind to this class and never sees it.
    
    After: 70 passed / 0 failed in the primary checkout, with the scratch dirs
    still on disk. Nothing was deleted to get here.
    
    Not a loosening. Case 17's noguards fixture still builds a repo with zero guard
    files and is still REFUSED. The change only stops the compliant fixture from
    being non-compliant by accident. It copies every shipping hit rather than
    hits[0], because the gate accepts the guard when ANY resolved path exists, and
    matching that removes walk-order tie-break as a variable.
    
    Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
    Claude-Session: https://claude.ai/code/session_0151EYtGH6at3XTWqbBHAxEi

diff --git a/q-system/.q-system/scripts/test/test-repo-preflight.sh b/q-system/.q-system/scripts/test/test-repo-preflight.sh
index 6e26f2ef..5ebf90f1 100755
--- a/q-system/.q-system/scripts/test/test-repo-preflight.sh
+++ b/q-system/.q-system/scripts/test/test-repo-preflight.sh
@@ -173,6 +173,26 @@ fixture_git() {
 # guards it did not have -- and codex used that fixture as the proof that basename
 # comparison was hollow. A control repo has to be genuinely compliant, or every
 # refusal measured against it is measuring the wrong thing.
+#
+# THE FIXTURE MUST WALK THE SKELETON THE SAME WAY THE GATE DOES (sp-505140ae).
+# repo-preflight.sh grew a _shipping() filter in ASK-798 so a guard resolves to a
+# path an instance could actually have; this builder kept the old walk -- `/.git`
+# filtered, first os.walk hit, `break`. In a checkout holding PR-review scratch
+# dirs the two disagree: the builder writes .pr31rev/r2/tree/.../audhd-lint.py and
+# the gate looks only at the shipped path, so the CONTROL repo reads as absent.
+#
+# Measured 2026-08-14 in the primary checkout at 1830a72e: 60 passed / 10 failed,
+# every failure the same "absent=audhd-lint.py/auto-commit.py/auto-update.sh/..."
+# string -- and those names are exactly the guards whose first walk hit lands in
+# .pr31rev/ and .pr25rev/. The same commit in a clean worktree passes 70/70,
+# because a clean tree has no scratch dirs for the walk to find first. CI clones
+# clean, so CI is structurally blind to this and can never catch it.
+#
+# NOT A LOOSENING. Case 17's "noguards" fixture still builds a repo with no guard
+# files at all and is still REFUSED, which is what proves the gate kept its teeth.
+# This only stops the compliant fixture from being non-compliant by accident.
+# Copies EVERY shipping hit, not hits[0], because the gate accepts the guard if
+# ANY resolved path exists -- matching that removes the tie-break as a variable.
 copy_guards() {
   python3 - "$SKEL_SETTINGS" "$REPO" "$1" <<'GPY'
 import json, os, re, shutil, sys
@@ -183,16 +203,31 @@ for arr in hooks.values():
     for m in arr or []:
         for hk in m.get("hooks", []) or []:
             names |= set(re.findall(r"[\w.-]+\.(?:py|sh)", hk.get("command", "") or ""))
+
+# Kept byte-identical in intent to repo-preflight.sh's _shipping(). If that list
+# changes, this one changes with it or the control fixture silently rots again.
+SKIP_DIRS = {'.git', 'node_modules', '__pycache__', '.venv', '.pytest_cache',
+             'template-repo', 'worktrees'}
+
+
+def _shipping(dirname):
+    if dirname in SKIP_DIRS:
+        return False
+    for junk in ('.pr', '.wt-', '.fable-wt', '.sana-tmp'):
+        if dirname.startswith(junk):
+            return False
+    return True
+
+
 for rel in sorted(names):
-    for base, _dirs, files in os.walk(skel):
-        if "/.git" in base:
+    for base, dirs, files in os.walk(skel):
+        dirs[:] = [d for d in dirs if _shipping(d)]
+        if rel not in files:
             continue
-        if rel in files:
-            r = os.path.relpath(os.path.join(base, rel), skel)
-            t = os.path.join(dst, r)
-            os.makedirs(os.path.dirname(t), exist_ok=True)
-            shutil.copy2(os.path.join(base, rel), t)
-            break
+        r = os.path.relpath(os.path.join(base, rel), skel)
+        t = os.path.join(dst, r)
+        os.makedirs(os.path.dirname(t), exist_ok=True)
+        shutil.copy2(os.path.join(base, rel), t)
 GPY
 }
 

codex
The implementation matches the production gate’s current directory filter and “any shipped path exists” semantics. I’m now running the full preflight suite, then targeted mutation checks for two risks: a missing guard still being rejected, and scratch-directory ordering no longer poisoning the compliant control.
exec
/bin/zsh -lc "python3 -c 'from pathlib import Path; import re; ps=[Path(\"q-system/.q-system/scripts/repo-preflight.sh\"),Path(\"q-system/.q-system/scripts/test/test-repo-preflight.sh\")]; b=[]; [(lambda s: b.append((re.search(r\"SKIP_DIRS = \\\\{.*?\\\\n\\\\s*\\u0027template-repo\\u0027, \\u0027worktrees\\u0027\\\\}\",s,re.S).group(0),re.search(r\"for junk in \\\\((.*?)\\\\):\",s,re.S).group(1))))(p.read_text()) for p in ps]; print(\"shipping filter parity:\",\"PASS\" if b[0]==b[1] else \"FAIL\"); print(\"production prefixes:\",b[0][1].replace(\"\\\\n\",\" \")); print(\"fixture prefixes:\",b[1][1].replace(\"\\\\n\",\" \"))'" in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-162
 exited 1 in 0ms:
    from pathlib import Path; import re; ps=[Path("q-system/.q-system/scripts/repo-preflight.sh"),Path("q-system/.q-system/scripts/test/test-repo-preflight.sh")]; b=[]; [(lambda s: b.append((re.search(r"SKIP_DIRS = \\{.*?\\n\\s*\u0027template-repo\u0027, \u0027worktrees\u0027\\}",s,re.S).group(0),re.search(r"for junk in \\((.*?)\\):",s,re.S).group(1))))(p.read_text()) for p in ps]; print("shipping filter parity:","PASS" if b[0]==b[1] else "FAIL"); print("production prefixes:",b[0][1].replace("\\n"," ")); print("fixture prefixes:",b[1][1].replace("\\n"," "))
                                                                                                                                                                          ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^
  File "<string>", line 1, in <lambda>
    from pathlib import Path; import re; ps=[Path("q-system/.q-system/scripts/repo-preflight.sh"),Path("q-system/.q-system/scripts/test/test-repo-preflight.sh")]; b=[]; [(lambda s: b.append((re.search(r"SKIP_DIRS = \\{.*?\\n\\s*\u0027template-repo\u0027, \u0027worktrees\u0027\\}",s,re.S).group(0),re.search(r"for junk in \\((.*?)\\):",s,re.S).group(1))))(p.read_text()) for p in ps]; print("shipping filter parity:","PASS" if b[0]==b[1] else "FAIL"); print("production prefixes:",b[0][1].replace("\\n"," ")); print("fixture prefixes:",b[1][1].replace("\\n"," "))
                                                                                                                                                                                               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AttributeError: 'NoneType' object has no attribute 'group'

exec
/bin/zsh -lc "printf '%s\\n' 'FILTER DEFINITIONS'
rg -n -A13 '"'^SKIP_DIRS ='"' q-system/.q-system/scripts/repo-preflight.sh q-system/.q-system/scripts/test/test-repo-preflight.sh
printf '%s\\n' 'NEGATIVE CONTROL'
rg -n 'make_good_repo noguards noguards|NO guard files is refused' q-system/.q-system/scripts/test/test-repo-preflight.sh
printf '%s\\n' 'OLD WALK BREAK STATUS'
git diff --check origin/main...HEAD && echo 'diff check: PASS'
bash -n q-system/.q-system/scripts/test/test-repo-preflight.sh && echo 'bash syntax: PASS'" in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-162
 succeeded in 0ms:
FILTER DEFINITIONS
q-system/.q-system/scripts/repo-preflight.sh:283:SKIP_DIRS = {'.git', 'node_modules', '__pycache__', '.venv', '.pytest_cache',
q-system/.q-system/scripts/repo-preflight.sh-284-             'template-repo', 'worktrees'}
q-system/.q-system/scripts/repo-preflight.sh-285-
q-system/.q-system/scripts/repo-preflight.sh-286-
q-system/.q-system/scripts/repo-preflight.sh-287-def _shipping(dirname):
q-system/.q-system/scripts/repo-preflight.sh-288-    """Directories that never reach an instance, so never define the guard path."""
q-system/.q-system/scripts/repo-preflight.sh-289-    if dirname in SKIP_DIRS:
q-system/.q-system/scripts/repo-preflight.sh-290-        return False
q-system/.q-system/scripts/repo-preflight.sh-291-    # Review/scratch worktrees this repo creates at its own root. `.q-system` is a
q-system/.q-system/scripts/repo-preflight.sh-292-    # real shipped directory, so dotted names are filtered by prefix, not wholesale.
q-system/.q-system/scripts/repo-preflight.sh-293-    for junk in ('.pr', '.wt-', '.fable-wt', '.sana-tmp'):
q-system/.q-system/scripts/repo-preflight.sh-294-        if dirname.startswith(junk):
q-system/.q-system/scripts/repo-preflight.sh-295-            return False
q-system/.q-system/scripts/repo-preflight.sh-296-    return True
--
q-system/.q-system/scripts/test/test-repo-preflight.sh:209:SKIP_DIRS = {'.git', 'node_modules', '__pycache__', '.venv', '.pytest_cache',
q-system/.q-system/scripts/test/test-repo-preflight.sh-210-             'template-repo', 'worktrees'}
q-system/.q-system/scripts/test/test-repo-preflight.sh-211-
q-system/.q-system/scripts/test/test-repo-preflight.sh-212-
q-system/.q-system/scripts/test/test-repo-preflight.sh-213-def _shipping(dirname):
q-system/.q-system/scripts/test/test-repo-preflight.sh-214-    if dirname in SKIP_DIRS:
q-system/.q-system/scripts/test/test-repo-preflight.sh-215-        return False
q-system/.q-system/scripts/test/test-repo-preflight.sh-216-    for junk in ('.pr', '.wt-', '.fable-wt', '.sana-tmp'):
q-system/.q-system/scripts/test/test-repo-preflight.sh-217-        if dirname.startswith(junk):
q-system/.q-system/scripts/test/test-repo-preflight.sh-218-            return False
q-system/.q-system/scripts/test/test-repo-preflight.sh-219-    return True
q-system/.q-system/scripts/test/test-repo-preflight.sh-220-
q-system/.q-system/scripts/test/test-repo-preflight.sh-221-
q-system/.q-system/scripts/test/test-repo-preflight.sh-222-for rel in sorted(names):
NEGATIVE CONTROL
OLD WALK BREAK STATUS
diff check: PASS
bash syntax: PASS

exec
/bin/zsh -lc "rg -n -C3 'noguards|NO guard|guard files' q-system/.q-system/scripts/test/test-repo-preflight.sh
python3 -c \"dirs=['.pr31rev','.q-system','plugins','worktrees','.wt-x','.sana-tmp-1']; skip={'.git','node_modules','__pycache__','.venv','.pytest_cache','template-repo','worktrees'}; prefixes=('.pr','.wt-','.fable-wt','.sana-tmp'); kept=[d for d in dirs if d not in skip and not any(d.startswith(x) for x in prefixes)]; print('input:',','.join(dirs)); print('kept:',','.join(kept)); print('scratch pruning:','PASS' if kept==['.q-system','plugins'] else 'FAIL')\"" in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-162
 succeeded in 0ms:
188-# because a clean tree has no scratch dirs for the walk to find first. CI clones
189-# clean, so CI is structurally blind to this and can never catch it.
190-#
191:# NOT A LOOSENING. Case 17's "noguards" fixture still builds a repo with no guard
192-# files at all and is still REFUSED, which is what proves the gate kept its teeth.
193-# This only stops the compliant fixture from being non-compliant by accident.
194-# Copies EVERY shipping hit, not hits[0], because the gate accepts the guard if
--
243-  mkdir -p "$dir/q-system/.q-system/scripts" "$dir/.claude"
244-  cp "$SKEL_WORKER" "$dir/q-system/.q-system/scripts/linear-worker.sh"
245-  cp "$SKEL_SETTINGS" "$dir/.claude/settings.json"
246:  # $2 = "noguards" builds the deliberately non-compliant shape for case 17.
247:  [ "${2:-}" = "noguards" ] || copy_guards "$dir"
248-  git init -q -b main "$dir"
249-  ( git -C "$dir" config user.email t@e.com
250-    git -C "$dir" config user.name t
--
805-echo "== 17. named guard scripts must EXIST in the target (codex r2 / adversarial) =="
806-# The green fixture copies settings.json and none of the hook files, which is
807-# exactly the shape that passed a basename comparison. Give one repo the real
808:# guard files and prove the difference is what decides it.
809:NOGUARD="$(make_good_repo noguard noguards)"
810-GOUT="$(run_preflight "$NOGUARD")"
811-{ [ "$(pf_rc "$NOGUARD")" != "0" ] && printf '%s' "$GOUT" | grep -q 'hooks'; } \
812-  && ok "a repo naming guards it does not actually have is REFUSED" \
813-  || bad "THE DEFECT: guard scripts named but absent still passed: $GOUT"
814-printf '%s' "$GOUT" | grep -q 'absent=' \
815:  && ok "the refusal names the guard files that are missing from disk" \
816-  || bad "the refusal does not distinguish absent files from unwired ones: $GOUT"
817-
818-echo
input: .pr31rev,.q-system,plugins,worktrees,.wt-x,.sana-tmp-1
kept: .q-system,plugins
scratch pruning: PASS

codex
One finding survived: the patch fixes the fixture helper but never adds the scratch-directory fixture needed to exercise that branch in clean CI. The changed behavior is therefore untested in the environment that gates merges. I’m executing a direct reproducer for that coverage gap and will calibrate it as bounded because it affects test reliability, not production behavior.
exec
/bin/zsh -lc "if rg -n 'mkdir["'^'"\\n]*(\\.pr|\\.wt-|\\.fable-wt|\\.sana-tmp)|os\\.makedirs["'^'"\\n]*(\\.pr|\\.wt-|\\.fable-wt|\\.sana-tmp)' q-system/.q-system/scripts/test/test-repo-preflight.sh; then echo 'scratch regression fixture: PRESENT'; else echo 'scratch regression fixture: ABSENT'; echo 'the suite never creates a filtered scratch directory, so a clean CI checkout does not execute the behavior changed at lines 209-230'; fi" in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-162
 succeeded in 0ms:
scratch regression fixture: ABSENT
the suite never creates a filtered scratch directory, so a clean CI checkout does not execute the behavior changed at lines 209-230

codex
## Finding

**MINOR:** The test updates scratch-directory filtering but never creates a scratch directory, so clean CI does not exercise the behavior being fixed.

- File: `q-system/.q-system/scripts/test/test-repo-preflight.sh:223`
- Reproducer:

```bash
if rg -n 'mkdir[^\n]*(\.pr|\.wt-|\.fable-wt|\.sana-tmp)|os\.makedirs[^\n]*(\.pr|\.wt-|\.fable-wt|\.sana-tmp)' q-system/.q-system/scripts/test/test-repo-preflight.sh; then
  echo 'scratch regression fixture: PRESENT'
else
  echo 'scratch regression fixture: ABSENT'
  echo 'the suite never creates a filtered scratch directory, so a clean CI checkout does not execute the behavior changed at lines 209-230'
fi
```

Real output:

```text
scratch regression fixture: ABSENT
the suite never creates a filtered scratch directory, so a clean CI checkout does not execute the behavior changed at lines 209-230
```

- Fix: Add a scratch tree containing a duplicate guard before building the control fixture, then assert the control still passes.
- Energy: Quick Win
- Time est: 15 minutes

## What is sound

- Production and fixture filters currently contain the same skip set and prefixes.
- The negative `noguards` fixture remains wired to assert missing files are refused.
- The new walk copies every shipping match, matching production’s `any(...)` semantics.
- `git diff --check` passed.
- `bash -n q-system/.q-system/scripts/test/test-repo-preflight.sh` passed.
- The full suite could not run because the read-only sandbox refused its `mktemp` directory. This was an environment limitation, not a reproduced PR defect.

Most important fix: add the missing scratch-directory regression fixture so reverting this patch makes CI red.

FINDINGS:
minor|The test updates scratch-directory filtering but never creates a scratch directory, so clean CI does not exercise the behavior being fixed.|q-system/.q-system/scripts/test/test-repo-preflight.sh:223
END FINDINGS
tokens used
60,557
## Finding

**MINOR:** The test updates scratch-directory filtering but never creates a scratch directory, so clean CI does not exercise the behavior being fixed.

- File: `q-system/.q-system/scripts/test/test-repo-preflight.sh:223`
- Reproducer:

```bash
if rg -n 'mkdir[^\n]*(\.pr|\.wt-|\.fable-wt|\.sana-tmp)|os\.makedirs[^\n]*(\.pr|\.wt-|\.fable-wt|\.sana-tmp)' q-system/.q-system/scripts/test/test-repo-preflight.sh; then
  echo 'scratch regression fixture: PRESENT'
else
  echo 'scratch regression fixture: ABSENT'
  echo 'the suite never creates a filtered scratch directory, so a clean CI checkout does not execute the behavior changed at lines 209-230'
fi
```

Real output:

```text
scratch regression fixture: ABSENT
the suite never creates a filtered scratch directory, so a clean CI checkout does not execute the behavior changed at lines 209-230
```

- Fix: Add a scratch tree containing a duplicate guard before building the control fixture, then assert the control still passes.
- Energy: Quick Win
- Time est: 15 minutes

## What is sound

- Production and fixture filters currently contain the same skip set and prefixes.
- The negative `noguards` fixture remains wired to assert missing files are refused.
- The new walk copies every shipping match, matching production’s `any(...)` semantics.
- `git diff --check` passed.
- `bash -n q-system/.q-system/scripts/test/test-repo-preflight.sh` passed.
- The full suite could not run because the read-only sandbox refused its `mktemp` directory. This was an environment limitation, not a reproduced PR defect.

Most important fix: add the missing scratch-directory regression fixture so reverting this patch makes CI red.

FINDINGS:
minor|The test updates scratch-directory filtering but never creates a scratch directory, so clean CI does not exercise the behavior being fixed.|q-system/.q-system/scripts/test/test-repo-preflight.sh:223
END FINDINGS

PR #165, opened 2026-08-14
Unblock the last 5 instances: reach 17/22 -> 22/22 (sp-ed9513cd, sp-3d5a247e)

assafkip:
## What shipped

Fleet reach **17 of 22 -> 22 of 22**, measured with `fleet-reach-audit.py` before and after.

### 1. `fleet-unblock.py` + `test_fleet_unblock.py` (sp-ed9513cd)

The write half of the reach audit. Imports the audit's classifier rather than
transcribing the guard's pathspec, and acts only on what it can attribute:

| action | proof required |
|---|---|
| `restore-mode` | index and worktree hold the same blob, differing only in mode |
| `commit` | the blob is one the skeleton itself held at that exact path |
| `unstage` | staged ADD, worktree agrees, and the blob is committed under `rescued/` |

Everything else is refused by name. Never a delete, never a worktree discard,
never `--no-verify`. Applied to 28 paths across 5 instances, 0 refusals.

**The filed diagnosis was wrong on one instance.** The note called it a worktree
EDIT of `skill-trigger-eval.py`. Same blob `56d37bd1` on both sides: the file had
lost its executable bit (`100755 -> 100644`, `-rw-------` on disk). A script that
had silently stopped being executable, with a blocked sync as the only symptom.
A one-action tool would have committed the broken mode. Measured across all 22
instances: exactly ONE true mode-only row, so this is a single file, not a class.

Mutation-tested: loosening attribution, dropping the `rescued/` proof, dropping
the commit unwind, and committing the broken mode instead of chmodding each turn
exactly the intended test red. Control green before and after.

### 2. `test-kipi-update-config-commit-unwind.sh`

The commit-failure unwind had never executed. The fixture that was asked for --
a pre-commit hook exiting 1, asserting a clean index -- **already exists** as case
13 of `test-kipi-update-safety.sh`, and it does not reach that code. Measured by
instrumenting the updater with markers and replaying case 13: the run dies at
"could not commit q-system sync" and neither the config-sync call site nor the
unwind is reached. A hook that refuses every commit stops at the first one.

This fixture refuses selectively (passes the q-system commit, rejects one staging
`.claude/` or `plugins/`) -- the shape of a lefthook blocked-paths rule, which two
instances in this fleet run.

**The result is not the expected one.** Mutating both candidates:

- ASK-797 unwind deleted -> index **still clean**
- `restore_instance` removed from `abandon_instance` -> instance left dirty

`restore_instance` is what holds this line. The `unstage_scope` unwind is
[...]

PR #169, opened 2026-08-14
Stop the prd-os lock files standing dirty forever (sp-a21cb27c, sp-455ed238)

assafkip:
Stacked on #165, follows #166 and #167. Ledger hygiene with measurements, not assertions.

## sp-a21cb27c — premise falsified

The note predicted an unattended auto-commit would sweep `.prd-os/spillover.jsonl.lock` into a commit.

Measured: it will not. `auto-commit.py`'s NEVER list already refuses any `.lock`, and `system_state_paths` returns `[]` for both lock files. Pinned by `TestTheNeverList::test_a_lock_file_is_never_taken`. There was never a path from those files into a commit.

What *was* real is smaller and still worth closing: both stood untracked-and-unignored inside a **tracked** directory, recreated on every resolve, so `git status` was never clean.

A status that is never clean is a status nobody reads. That is how the five committed integrity markers in #166 went unnoticed for weeks.

Both paths now ignored, verified with `git check-ignore`. Voided with the measurement.

## sp-455ed238 — duplicate

Same root cause as sp-097d2e23: root `.gitignore` is not in the updater sync set, so no instance receives its rules. This one scoped it to the `.update-check-<date>` stamps and called it harmless. sp-097d2e23 found the same gap exposes the integrity baseline and armed marker to auto-commit, which had already happened five times.

Filed 2026-07-28 from ASK-113. Re-filed independently 2026-08-14 as sp-097d2e23. **The re-file is the evidence**: a note nobody can close gets rediscovered rather than fixed.

Not voided as wrong. The observation was correct, and the shipped stanza covers this exact case (`q-system/output/.update-check-*` is in it, pinned by `test_the_update_check_stamp_is_ignored_too`). Tracking it twice is what is being retired.

## Verification

The block writer is unaffected: the shipped stanza is delimited by its own markers and these lines sit outside them. **17 tests still green.**

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -109,6 +109,16 @@ q-system/output/.update-check-*
 q-system/output/capability-maps/
 .runs/
 
+# Lock files beside the prd-os ledgers. flock targets, recreated on every
+# resolve/reclassify, and they were standing untracked-and-unignored inside a
+# TRACKED directory. sp-a21cb27c predicted an unattended auto-commit would sweep
+# them in; measured 2026-08-14 that is false -- auto-commit.py's NEVER list
+# already refuses any .lock, so system_state_paths returns [] for both. What was
+# left was permanent noise in `git status`, which is its own cost: a status that
+# is never clean is a status nobody reads.
+.prd-os/*.lock
+.prd-os/.judgments.lock
+
 # THE STANZA BETWEEN THESE MARKERS IS SHIPPED TO EVERY INSTANCE (sp-097d2e23).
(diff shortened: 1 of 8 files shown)


assafkip:
Superseded by a clean rebuild off the current base. The squash merges of #166 and #167 rewrote the same files, so this branch went DIRTY and its three-dot diff was showing ~700 lines of already-merged content as new. The two-dot diff was the honest one: the only real change was 30 lines. Reopened as a fresh branch with exactly those 30.

PR #172, opened 2026-08-15
Correct a false claim I put in the .gitignore fix's own comments

assafkip:
Follows #171. The fix there is right. **The reason written beside it was not**, and a wrong scar comment is worse than no comment — it is read as coverage, so nobody goes looking.

## What I claimed

Instances have no root `.gitignore` at all, so the writer creates the file. Stated in two code comments, a commit message, #171's body, and a spillover note.

## What is true

Measured by copying a real instance and running the writer on the copy, rather than by grepping:

- Every instance **has** a root `.gitignore` — 70 to 76 lines — and it is **tracked**.
- None of them carries the integrity rules, so sp-097d2e23's premise is untouched.
- The writer **modifies** a tracked file. It lands as ` M .gitignore`.

The false claim came from a malformed `grep -c` whose output I misread. That is the same mistake this PR chain keeps finding in other people's filed notes — trusting a grep over looking at the thing — and I made it while writing the fix for it. What caught it was noticing `.DS_Store` and `.pytest_cache` in an `--ignored` listing when my block contains neither pattern.

## I guessed wrong about the consequence too

Both now checked rather than asserted:

- A modified tracked `.gitignore` does **not** block the sync. The dirty-tree guard is scoped to `$prefix/ .claude/ plugins/` (`kipi-update.sh` ~L2012) and root `.gitignore` is in none of them. My first guess was that this self-blocked all 22 instances. Wrong.
- What it **does** cause: a permanently modified, never-committable tracked file in all 22 instances, plus the same unclassifiable path reported every run. That is what #171's classifier entry fixes, and it is still needed.

## Test

New test pins the real-world shape: a **tracked** `.gitignore` the writer modified is offered to the fleet sync. The previous tests only covered the untracked-created case — the case that does not happen. It went red first for a fourth mistake: the helper returns a `CompletedProcess`, not a string.

## Ledger

- `sp-58b21276` voided — filed on the false measurement.
- `sp-5d61ca3a` filed in its place: instances' `.gitignore` is a **stale fork** of the skeleton's, drifting since the instance was created. Narrower and more accurate than "no ignore rules at all", with an explicit warning not to fix it by rsyncing the whole file.

**337 passed, 0 failed.**

### DIFF
--- a/q-system/.q-system/tests/test_gitignore_block.py
+++ b/q-system/.q-system/tests/test_gitignore_block.py
@@ -264,6 +264,46 @@ class TestTheWiringIntoTheUpdater:
         assert r.returncode == 0, r.stderr
 
 
+class TestTheFileTheWriterCreatesCanActuallyBeCommitted:
+    """A defect in THIS fix, found by running it against real instances.
+
+    Dry-checked on three live instances 2026-08-14: none of them has a root
+    .gitignore AT ALL, so the writer does not edit a file, it CREATES one. Git
+    honours an untracked .gitignore, so the ignore rules work either way -- but
+    an untracked file is one nobody committed, in a repo nobody looks at, and
+    auto-commit.py classified `.gitignore` as `unclassified`.
+
+    Unclassified means REPORTED, never committed (ASK-498). So the fix would
+    have left every one of the 22 instances printing the same unclassifiable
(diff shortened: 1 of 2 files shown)


assafkip:
Superseded: went DIRTY after #171 squash-merged into this base. Rebuilt clean off the current tip, same 57 lines.

PR #175, opened 2026-08-15
Restore the index entry, and make success an allowlist (PR #165 review round 2)

assafkip:
Round 2 of the reviewer, on the right tree this time. Two more **major** findings, both real. The second is a hole in my own round-1 fix.

## Major 1 — a rejected commit destroyed content the founder had already staged

`commit_with_unwind` kept a **set** of paths that were already staged, and `unwind()` **skipped** them — the tool did not stage it, so the tool should not unstage it.

By the time unwind runs, `git add` has already overwritten that index entry with the worktree content. So a founder who had staged their own version of the path lost it, and the run printed *"index unwound"* while saying so.

Skipping a path is not restoring it. It now records the exact index entry — mode and blob — and puts it back with `update-index --cacheinfo`.

## Major 2 — my round-1 exit-code fix was a denylist, and it leaked

Round 1 tested `outcome.startswith("REFUSED")`. The other producers emit `FAILED add: `, `FAILED <err>` and `STILL <mode>` — none of which start with REFUSED — so all three still counted as successful actions and still exited 0.

A denylist of failure strings is the wrong shape for the thing an unattended job reads as its exit code: **the next outcome string anyone adds lands on the success side by default.** Success is now an allowlist — `clean`, `committed`, or a `would ...` dry run — and everything else is a failure. Fail closed.

One of the three new tests derives the outcome literals from the source, so a new outcome string cannot arrive unclassified.

## Red proven, not assumed

I fixed before running the reproducers, which is backwards. So I stashed the fix and re-ran: **all three fail without it.**

The second reproducer was also rewritten. An earlier draft chmod'ed `.git/index` read-only to provoke an add failure; git rewrote it anyway and the test **skipped**. A skipped test proves nothing, and the predicate is a pure string function that deserves a direct test.

## Verification

| Check | Result |
|---|---|
| Full suite | **343 passed, 0 failed** |
| Reproducers without the fix | 3 failed |
| Real fleet dry run | nothing to do across all 22, exits 0 |

### DIFF
--- a/fleet-unblock.py
+++ b/fleet-unblock.py
@@ -136,6 +136,31 @@ def decide(repo, row, audit, rescued):
         )
 
     if row["kind"] == "fleet-written":
+        # BOTH SIDES OF THE PATH, NOT ONE (PR #165 review, major).
+        #
+        # The index can hold a blob the skeleton really did ship while the
+        # WORKTREE holds bytes it never had -- a founder edit on top of a staged
+        # skeleton write. This branch used to match on the index alone and
+        # schedule a commit. `git commit` takes the INDEX, so the run would clear
+        # the dirty-tree guard, report the path repaired, and leave the founder's
+        # edit uncommitted for the next sync to overwrite. Founder work destroyed
+        # by the tool whose entire job is avoiding exactly that.
+        #
+        # One side matching the skeleton is not attribution, it is half of one.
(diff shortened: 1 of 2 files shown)


assafkip:
Superseded: DIRTY after #174 squash-merged into this base. Rebuilt clean off the current tip with the same 154 lines.

PR #141, opened 2026-08-13
ASK-700: refuse (premise already fixed by ASK-363) + fix the committed refusal sentinel

assafkip:
Two things in one branch: a **refusal** on the issue as scoped, and a **fix** for a defect the refusal itself introduced (found by Codex round 1).

## Part 1 — the refusal (no source file changed)

ASK-700 asks to strip 0.1.0 scaffold language ("not yet wired", instructions to refuse `/prd-*` commands) from `plugins/prd-os/skills/prd-os/SKILL.md`. That language left `origin/main` on **2026-08-05** in `86f4ff63` (ASK-363, PR #110) — two days **before** spillover row `sp-32e6595c` was written on 2026-08-07.

Re-verified independently this session, by a different method than the original pass (working-tree grep + a diff against main, rather than `git show` of a blob):

```
$ grep -nE "not yet wired|Scaffold only|once wired|do not claim|0\.1\.0" \
    plugins/prd-os/skills/prd-os/SKILL.md
rc=1   (no match)

$ git diff --stat origin/main HEAD -- plugins/prd-os/skills/prd-os/SKILL.md
(empty — byte-identical to main)
```

Acceptance criterion 1 also requires the documented workflow to match what ships. Checked against the filesystem, not assumed:

| Claim in SKILL.md | Verified against |
|---|---|
| 8 prd-os commands + `/prd-os-init` bootstrap | `plugins/prd-os/commands/` — exact match |
| 6 issue commands live in kipi-dsse | `plugins/kipi-dsse/commands/` — exact match |
| `gates run`, `spillover add/resolve` | `prd_runner.py gates --help` → {list, run}; `spillover --help` → {add, list, reclassify, check, triage, resolve} |
| version | `plugin.json` is 0.26.5, not the 0.16.5 the finding observed |

No edit exists to make in either allowed file. Inventing one would be a fabricated diff.

**What the reporter actually saw.** Something real, but not in the source: the **marketplace clone** at 0.16.5 — the copy Claude Code actually loads. Skeleton was fixed; clone was never refreshed; fresh sessions kept getting told to refuse working commands. That is skeleton-to-clone version drift, and nothing detects it. Editing SKILL.md cannot fix it. Captured as `sp-1e6e8091` with a workable DoR (reproducer-first drift checker wired into `kipi check`).

Not claiming the clone is stale *today* — this session's sandbox refuses reads outside the worktree, so only the source was verifiable.

## Part 2 — the fix (Codex round 1, major)

[...]

PR #122, opened 2026-08-06
Give dispatch its own checkout, pinned to origin/main (built, not armed)

assafkip:
Built and proven. **The plist is untouched — this is inert until armed.**

## The outage

`com.kipi.dispatch` sets `KIPI_REPO` to the shared checkout that interactive sessions also work in, so it lives on a feature branch. Every 900s tick from `2026-08-05T23:53Z`:

```
STALE: origin/main holds 12 commit(s) this checkout lacks (HEAD ba584d2, origin/main 7aa9cdd)
page suppressed (stale-checkout unchanged for 1351m; re-pings after 24h)
```

22.5h of total dispatch outage, one page. The guard was right every time. It was pointed at a checkout humans park.

## Worktree, not a clone

`.prd-os/spillover.jsonl` is gitignored (`.gitignore:36`) and lives physically in the shared checkout. `prd_runner` reaches it via git-common-dir, so all worktrees see ONE ledger. A clone forks it in two and turns the no-orphan-findings gate green on half the record.

Safe because `linear-worker.sh` cuts issue trees with `worktree add -B` into `$STATE_DIR/worktrees/` (`:1064`, `:1092`) and never moves `TARGET_REPO`s HEAD. The shared checkout only goes stale because *humans* park it.

## The guard keeps its teeth

The wrapper removes the cause and leaves the guard to catch a failed removal. It advances with `checkout --detach origin/main`, which **refuses** on a dirty or diverged tree, and never `reset --hard` / `-f` / `clean`. On failure it still execs dispatch and lets `stale_check()` refuse. If the pinned checkout cannot be created it exits rather than falling back to the shared one, because a fallback makes the outage silent again.

## Verification (local only — no CI result claimed)

`bash test_dispatch_pinned.sh` -> 7 checks PASS. Throwaway repos under `mktemp -d`, all three real paths injected, `kipi-dispatch.sh` stubbed. No live path touched.

**T2 is load-bearing and has a negative self-test.** The same assertion run against a `--force` variant returns DISCARDED while the real wrapper returns PRESERVED. Without it, T2 could not tell "refuses to force" from "cannot see forcing".

Mechanism proven before any of this was written, via the flag whose docstring says it exists "for proving the gate from outside":

```
KIPI_REPO=<pinned> KIPI_DISPATCH_PICK_DRY=1 bash kipi-dispatch.sh   -> rc 0
dispatch.log: "page state cleared: stale-checkout recovered"        (no STALE line)
```

[...]

PR #186, opened 2026-08-15
Derive the pinned checkout's name from the repo, not a literal (ASK-829)

assafkip:
## Home-repo dispatch was dead all night and looked normal

`linear-worker.sh` resolves an issue's Linear project as:

```
KIPI_LINEAR_PROJECT  ->  registry lookup by path  ->  basename $TARGET_REPO
```

ASK-447's pinned checkout is not in `instance-registry.json`, so resolution fell through to `basename`, which was the literal `dispatch-checkout` — matching no Linear project. Every home-repo issue was filtered out:

```
MISCONFIG: repo identity 'dispatch-checkout' matches NO Linear project
MISCONFIG: every issue would be filtered out, so this run picked nothing
           for a config reason, not an empty board.
```

**25 ready issues unpickable.** The *dispatch* log said only `nothing ready ()` — indistinguishable from an empty board, because the MISCONFIG lines land in `linear-worker.log`, a different file. I read past it twice before chasing it.

Cross-repo was unaffected (it passes `--repo` and resolves through the registry), so the fleet looked half-alive rather than broken. That is why it survived arming, a merge, and two rounds of me watching the log.

## The fix

The name is **derived from the origin URL**. That is the one thing that always knows what this repository is; the containing directory does not, since `KIPI_REPO_MAIN` is itself a worktree named `dispatch-main`.

Verified: `derived ident: kipi-system`.

A machine-local `KIPI_DISPATCH_CHECKOUT` override unblocked it on the night and is **not** the fix — a fresh install, or anyone re-rendering the plist, silently reintroduces the filtering.

## Second correction, in the same file

This wrapper claimed refusals *"reach the founder's phone"* via `slack-notify.sh`. They do not. That script keeps its name for ~30 callers, but its own header reads:

> THE FLEET ALERT PATH. Files a Linear ticket for Sana. Pages nobody.

Nothing in it sends to Slack. That routing is the founder's stated preference, not a degradation — but a comment promising a phone page sets the wrong expectation about who is watching at 3am. Closes sp-d9212e22.

## Verification

`test_dispatch_pinned.sh` unchanged and still green, including the three negative self-tests.

Closes ASK-829.

### DIFF
--- a/kipi-dispatch-pinned.sh
+++ b/kipi-dispatch-pinned.sh
@@ -43,7 +43,47 @@
 set -uo pipefail
 
 REPO_MAIN="${KIPI_REPO_MAIN:-$HOME/projects/kipi-system}"
-PINNED="${KIPI_DISPATCH_CHECKOUT:-$HOME/.local/state/kipi/dispatch-checkout}"
+
+# THE CHECKOUT'S DIRECTORY NAME IS LOAD-BEARING, AND NAMING IT "dispatch-checkout"
+# SILENTLY KILLED HOME-REPO DISPATCH (ASK-829).
+#
+# linear-worker.sh resolves which Linear project an issue belongs to like this:
+#     KIPI_LINEAR_PROJECT  ->  registry lookup by path  ->  basename $TARGET_REPO
+# The pinned checkout is not in instance-registry.json, so resolution fell all the
+# way through to basename, which was "dispatch-checkout" -- a name matching no
+# Linear project. Every home-repo issue was filtered out:
+#
+#   MISCONFIG: repo identity 'dispatch-checkout' matches NO Linear project
+#   MISCONFIG: every issue would be filtered out, so this run picked nothing
+#              for a config reason, not an empty board.
+#
+# Measured 2026-08-15: 25 ready issues unpickable, while the DISPATCH log said
+# only "nothing ready ()" -- indistinguishable from an empty board, because the
+# MISCONFIG lines land in linear-worker.log, a different file. Cross-repo was
+# unaffected: it passes --repo and resolves through the registry, so only the
+# home repo was dead and the fleet looked half-alive.
+#
+# So the name is DERIVED from the repo's own identity rather than invented. The
+# origin URL is the one thing that always knows what this repository is; the
+# containing directory does not (KIPI_REPO_MAIN is itself a worktree here, named
+# dispatch-main). A machine-local env override would have worked too and is what
+# unblocked it on the night; it is not the fix, because a fresh install or a
+# re-render of the plist silently reintroduces the filtering.
+_repo_ident() {
+  local url base
+  url="$(git -C "$REPO_MAIN" remote get-url origin 2>/dev/null)" || return 1
+  [ -n "$url" ] || return 1
+  base="${url##*/}"; base="${base%.git}"
+  [ -n "$base" ] || return 1
+  printf '%s' "$base"
+}
+_IDENT="$(_repo_ident || true)"
+# Fall back to the old literal ONLY if the remote cannot be read. That is worse
+# than useless for project resolution, so it is announced rather than silent.
+if [ -z "$_IDENT" ]; then
+  _IDENT="dispatch-checkout"
+fi
+PINNED="${KIPI_DISPATCH_CHECKOUT:-$HOME/.local/state/kipi/dispatch/$_IDENT}"
 LOG="${KIPI_DISPATCH_LOG:-$HOME/.config/kipi/dispatch.log}"
 
 say() { printf '%s %s\n' "$(date -u +%FT%TZ)" "pinned: $*" >>"$LOG" 2>/dev/null || true; }
@@ -53,10 +93,16 @@ say() { printf '%s %s\n' "$(date -u +%FT%TZ)" "pinned: $*" >>"$LOG" 2>/dev/null 
 # because a nonzero exit would make launchd throttle a job that is behaving
 # correctly. But a clean launchd record plus a line in a log nobody reads is
 # exactly the shape of the 22.5h outage above: dispatch stopped, and the only
-# evidence was a file. So a refusal PAGES, and the log line stops being the only
-# copy. Routed through slack-notify.sh because it is the one channel that reaches
-# the founder's phone from a headless launchd context; osascript is silently
-# dropped there (founder-notifications rule).
+# evidence was a file. So a refusal RAISES AN ALERT, and the log line stops being
+# the only copy.
+#
+# WHERE THAT ALERT ACTUALLY GOES (corrected, sp-d9212e22). An earlier version of
+# this comment -- mine -- said it "reaches the founder's phone". It does not.
+# slack-notify.sh keeps its name for ~30 callers but its header is explicit:
+# "THE FLEET ALERT PATH. Files a Linear ticket for Sana. Pages nobody." Nothing
+# in it sends to Slack. That is the founder's stated routing, not a degradation:
+# things needing attention go to Sana. Saying "pages the founder" in a comment
+# would set the wrong expectation about who is watching at 3am.
 NOTIFY="${KIPI_NOTIFY:-$REPO_MAIN/q-system/.q-system/scripts/slack-notify.sh}"
 
 # AND IT PAGES ONCE, NOT 96 TIMES A DAY (codex major, PR #122 r2).


assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-186-20260814-184712.md` (178441 bytes).

```
FINDINGS:
minor|The remote-read fallback is silent even though the comment promises it is announced, leaving operators without the stated diagnostic.|kipi-dispatch-pinned.sh:82
minor|The notification comments still claim a phone page exists after this PR explicitly documents that the notifier only creates Linear tickets.|kipi-dispatch-pinned.sh:135
END FINDINGS
```

--- reviewer output, last 54637 bytes of 178441 (full review at the path above) ---

   155	  || :
   156	chmod +x "$STRIPPED"
   157	R4="$(build_fixture)"; : >"$R4/pinned"; NOTE2="$R4/paged.txt"
   158	cp "$R3/notify.sh" "$R4/notify.sh"
   159	PAGE_SINK="$NOTE2" KIPI_NOTIFY="$R4/notify.sh" run_wrapper "$R4" "$STRIPPED" >/dev/null 2>&1
   160	[ -s "$NOTE2" ] \
   161	  && bad "T5-neg the assertion CAN fail" "the stripped variant still paged, so T5 is not load-bearing" \
   162	  || ok "T5-neg the assertion CAN fail (stripped variant sent no page)"
   163	
   164	# ---------------------------------------------------------------------------
   165	# TEST 6: the shipped plist actually INVOKES this wrapper (codex major r1).
   166	# ---------------------------------------------------------------------------
   167	# Without this the wrapper is dead code: the scheduler keeps running dispatch out
   168	# of the shared checkout and merging changes nothing. A test that only exercises
   169	# the wrapper cannot see that, which is exactly how it was missed.
   170	PLIST="$HERE/q-system/.q-system/scripts/com.kipi.dispatch.plist"
   171	if [ -f "$PLIST" ]; then
   172	  grep -q 'kipi-dispatch-pinned.sh' "$PLIST" \
   173	    && ok "T6 the plist runs the pinned wrapper, so it has a production invoker" \
   174	    || bad "T6 the plist runs the pinned wrapper" "ProgramArguments still points at kipi-dispatch.sh; the wrapper would be inert"
   175	  # KIPI_REPO in the plist would OVERRIDE what the wrapper sets and re-point
   176	  # dispatch at the shared checkout -- the outage rebuilt through config.
   177	  grep -q '<key>KIPI_REPO</key>' "$PLIST" \
   178	    && bad "T6 the plist declares no KIPI_REPO" "KIPI_REPO here overrides the wrapper and re-points dispatch at the parked checkout" \
   179	    || ok "T6 the plist declares no KIPI_REPO, so the wrapper's choice stands"
   180	else
   181	  bad "T6 the plist is present" "no plist at $PLIST"
   182	fi
   183	
   184	# ---------------------------------------------------------------------------
   185	# TEST 7: a PERSISTENT refusal pages once, not every 15 minutes (codex major r2).
   186	# ---------------------------------------------------------------------------
   187	# The refusals are persistent by nature -- a checkout that cannot be created at
   188	# 09:00 still cannot at 09:15 -- so an undeduped page is 96 identical Slack lines
   189	# a day. That does not inform the operator, it trains them to swipe it away.
   190	R5="$(build_fixture)"
   191	: >"$R5/pinned"
   192	NOTE3="$R5/paged.txt"
   193	cp "$R3/notify.sh" "$R5/notify.sh"
   194	for _ in 1 2 3; do
   195	  PAGE_SINK="$NOTE3" KIPI_NOTIFY="$R5/notify.sh" run_wrapper "$R5" "$WRAPPER" >/dev/null 2>&1
   196	done
   197	n="$(wc -l < "$NOTE3" | tr -d ' ')"
   198	[ "$n" = "1" ] \
   199	  && ok "T7 three ticks of the same refusal produced exactly ONE page" \
   200	  || bad "T7 the persistent refusal is deduped" "3 ticks produced $n pages; at 900s that is ~96/day"
   201	grep -q 'page suppressed' "$R5/log" 2>/dev/null \
   202	  && ok "T7 the suppressed ticks are still in the LOG, so nothing is lost" \
   203	  || bad "T7 suppression is recorded in the log" "a suppressed page left no trace"
   204	
   205	# T7-neg: with the TTL set to 0 the dedupe window closes instantly, so all three
   206	# ticks must page. Without this, T7 could not tell "deduped" from "the notifier
   207	# was only ever called once for some unrelated reason".
   208	R6="$(build_fixture)"; : >"$R6/pinned"; NOTE4="$R6/paged.txt"
   209	cp "$R3/notify.sh" "$R6/notify.sh"
   210	for _ in 1 2 3; do
   211	  PAGE_SINK="$NOTE4" KIPI_NOTIFY="$R6/notify.sh" KIPI_DISPATCH_PAGE_TTL=0 \
   212	    run_wrapper "$R6" "$WRAPPER" >/dev/null 2>&1
   213	done
   214	n2="$(wc -l < "$NOTE4" | tr -d ' ')"
   215	[ "$n2" = "3" ] \
   216	  && ok "T7-neg with the window at 0 all three ticks page (the dedupe is what suppressed them)" \
   217	  || bad "T7-neg the assertion CAN fail" "expected 3 pages with TTL=0, got $n2"
   218	
   219	# ---------------------------------------------------------------------------
   220	# TEST 8: a RECOVERY clears the suppressor, so the next break pages at once.
   221	# ---------------------------------------------------------------------------
   222	# Otherwise the first outage's 24h window is inherited by an unrelated second
   223	# outage, and the page that matters is the one that gets swallowed.
   224	rm -f "$R5/pinned"                     # let the checkout succeed this time
   225	PAGE_SINK="$NOTE3" KIPI_NOTIFY="$R5/notify.sh" run_wrapper "$R5" "$WRAPPER" >/dev/null 2>&1
   226	ls "$R5"/pinned-paged-* >/dev/null 2>&1 \
   227	  && bad "T8 a successful run clears the page markers" "a marker survived recovery and will mute the next outage" \
   228	  || ok "T8 a successful run clears the page markers"
   229	
   230	# ---------------------------------------------------------------------------
   231	# TEST 9: a FAILED notification must not burn the dedupe window (codex major r3).
   232	# ---------------------------------------------------------------------------
   233	# Round 2 stamped the marker after the send ATTEMPT, so a Slack outage wrote a
   234	# 24h mute and the dispatch outage stayed unreported for a day after the notifier
   235	# recovered. The notifier here always exits non-zero, then recovers.
   236	R7="$(build_fixture)"; : >"$R7/pinned"; NOTE5="$R7/paged.txt"
   237	cat >"$R7/notify-broken.sh" <<'BROKEN'
   238	#!/bin/bash
   239	exit 1
   240	BROKEN
   241	chmod +x "$R7/notify-broken.sh"
   242	PAGE_SINK="$NOTE5" KIPI_NOTIFY="$R7/notify-broken.sh" run_wrapper "$R7" "$WRAPPER" >/dev/null 2>&1
   243	ls "$R7"/pinned-paged-* >/dev/null 2>&1 \
   244	  && bad "T9 a failed send leaves no dedupe marker" "the window was burned by a page that never arrived" \
   245	  || ok "T9 a failed send leaves no dedupe marker"
   246	
   247	# Now the notifier recovers. The very next tick must page, not inherit a mute.
   248	cp "$R3/notify.sh" "$R7/notify.sh"
   249	PAGE_SINK="$NOTE5" KIPI_NOTIFY="$R7/notify.sh" run_wrapper "$R7" "$WRAPPER" >/dev/null 2>&1
   250	[ -s "$NOTE5" ] \
   251	  && ok "T9 the first tick after the notifier recovers DOES page" \
   252	  || bad "T9 the first tick after recovery pages" "the outage stayed silent after the notifier came back"
   253	
   254	# T9-neg: stamp unconditionally (the round-2 shape) and T9 must go red.
   255	R2SHAPE="$R7/wrapper-r2.sh"
   256	sed 's|^  if bash "\$NOTIFY" "\$msg" >/dev/null 2>&1; then$|  bash "$NOTIFY" "$msg" >/dev/null 2>\&1; if true; then|' \
   257	  "$WRAPPER" > "$R2SHAPE"
   258	chmod +x "$R2SHAPE"
   259	grep -q 'if true; then' "$R2SHAPE" \
   260	  || bad "T9-neg the mutant was actually applied" "the unconditional-stamp variant was not produced"

exec
/bin/zsh -lc "rg --files -uu | rg 'linear-worker\\.sh"'$|instance-registry'"\\.json"'$|com'"\\.kipi\\.dispatch\\.plist"'$'"'; git show --stat --oneline HEAD; git show --format=fuller --no-ext-diff HEAD -- kipi-dispatch-pinned.sh; bash test_dispatch_pinned.sh" in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-186
 exited 1 in 429ms:
cat: /pinned/marker.txt: No such file or directory
FAIL T2 a dirty pinned checkout is NOT forced
     wrapper discarded the local edit
mktemp: mkdtemp failed on /var/folders/th/ty0kb7r56dvfrkgkrcbx9fv00000gn/T/tmp.FaazVWLcNL: Operation not permitted
test_dispatch_pinned.sh: line 84: /forcing-wrapper.sh: Operation not permitted
grep: /forcing-wrapper.sh: No such file or directory
FAIL T2-neg mutant did not apply
     sed produced no --force
mktemp: mkdtemp failed on /var/folders/th/ty0kb7r56dvfrkgkrcbx9fv00000gn/T/tmp.etDZWYucpS: Operation not permitted
fatal: cannot mkdir /upstream: Operation not permitted
test_dispatch_pinned.sh: line 25: /upstream/kipi-dispatch.sh: No such file or directory
test_dispatch_pinned.sh: line 29: /upstream/marker.txt: No such file or directory
fatal: cannot change to '/upstream': No such file or directory
fatal: cannot change to '/upstream': No such file or directory
fatal: repository '/upstream' does not exist
test_dispatch_pinned.sh: line 69: /upstream/marker.txt: No such file or directory
fatal: cannot change to '/upstream': No such file or directory
test_dispatch_pinned.sh: line 71: /pinned/marker.txt: No such file or directory
cat: /pinned/marker.txt: No such file or directory
ok   T2-neg the assertion CAN fail (forcing variant discarded the edit)
ok   T3 wrapper contains no forcing verb outside comments
mktemp: mkdtemp failed on /var/folders/th/ty0kb7r56dvfrkgkrcbx9fv00000gn/T/tmp.QOMuXC1o7N: Operation not permitted
fatal: cannot mkdir /upstream: Operation not permitted
test_dispatch_pinned.sh: line 25: /upstream/kipi-dispatch.sh: No such file or directory
test_dispatch_pinned.sh: line 29: /upstream/marker.txt: No such file or directory
fatal: cannot change to '/upstream': No such file or directory
fatal: cannot change to '/upstream': No such file or directory
fatal: repository '/upstream' does not exist
test_dispatch_pinned.sh: line 110: /pinned: Operation not permitted
ok   T4 refuses when the pinned checkout cannot be created
FAIL T4 logs the refusal loudly
     no refusal line in the log
mktemp: mkdtemp failed on /var/folders/th/ty0kb7r56dvfrkgkrcbx9fv00000gn/T/tmp.YycXNuDZhg: Operation not permitted
fatal: cannot mkdir /upstream: Operation not permitted
test_dispatch_pinned.sh: line 25: /upstream/kipi-dispatch.sh: No such file or directory
test_dispatch_pinned.sh: line 29: /upstream/marker.txt: No such file or directory
fatal: cannot change to '/upstream': No such file or directory
fatal: cannot change to '/upstream': No such file or directory
fatal: repository '/upstream' does not exist
test_dispatch_pinned.sh: line 129: /pinned: Operation not permitted
test_dispatch_pinned.sh: line 131: /notify.sh: Operation not permitted
chmod: /notify.sh: No such file or directory
FAIL T5 a refusal pages the founder, not just the log
     notifier was never called; launchd would report success while dispatch is stopped
FAIL T5 the page says dispatch is STOPPED
     page text did not name the state: 
test_dispatch_pinned.sh: line 152: /wrapper-nopage.sh: Operation not permitted
grep: /wrapper-nopage.sh: No such file or directory
chmod: /wrapper-nopage.sh: No such file or directory
mktemp: mkdtemp failed on /var/folders/th/ty0kb7r56dvfrkgkrcbx9fv00000gn/T/tmp.DSbZKzJHKp: Operation not permitted
fatal: cannot mkdir /upstream: Operation not permitted
test_dispatch_pinned.sh: line 25: /upstream/kipi-dispatch.sh: No such file or directory
test_dispatch_pinned.sh: line 29: /upstream/marker.txt: No such file or directory
fatal: cannot change to '/upstream': No such file or directory
fatal: cannot change to '/upstream': No such file or directory
fatal: repository '/upstream' does not exist
test_dispatch_pinned.sh: line 157: /pinned: Operation not permitted
cp: /notify.sh: No such file or directory
ok   T5-neg the assertion CAN fail (stripped variant sent no page)
ok   T6 the plist runs the pinned wrapper, so it has a production invoker
ok   T6 the plist declares no KIPI_REPO, so the wrapper's choice stands
mktemp: mkdtemp failed on /var/folders/th/ty0kb7r56dvfrkgkrcbx9fv00000gn/T/tmp.mbHPgsXw98: Operation not permitted
fatal: cannot mkdir /upstream: Operation not permitted
test_dispatch_pinned.sh: line 25: /upstream/kipi-dispatch.sh: No such file or directory
test_dispatch_pinned.sh: line 29: /upstream/marker.txt: No such file or directory
fatal: cannot change to '/upstream': No such file or directory
fatal: cannot change to '/upstream': No such file or directory
fatal: repository '/upstream' does not exist
test_dispatch_pinned.sh: line 191: /pinned: Operation not permitted
cp: /notify.sh: No such file or directory
test_dispatch_pinned.sh: line 197: /paged.txt: No such file or directory
FAIL T7 the persistent refusal is deduped
     3 ticks produced  pages; at 900s that is ~96/day
FAIL T7 suppression is recorded in the log
     a suppressed page left no trace
mktemp: mkdtemp failed on /var/folders/th/ty0kb7r56dvfrkgkrcbx9fv00000gn/T/tmp.DBM3k6dB7j: Operation not permitted
fatal: cannot mkdir /upstream: Operation not permitted
test_dispatch_pinned.sh: line 25: /upstream/kipi-dispatch.sh: No such file or directory
test_dispatch_pinned.sh: line 29: /upstream/marker.txt: No such file or directory
fatal: cannot change to '/upstream': No such file or directory
fatal: cannot change to '/upstream': No such file or directory
fatal: repository '/upstream' does not exist
test_dispatch_pinned.sh: line 208: /pinned: Operation not permitted
cp: /notify.sh: No such file or directory
test_dispatch_pinned.sh: line 214: /paged.txt: No such file or directory
FAIL T7-neg the assertion CAN fail
     expected 3 pages with TTL=0, got 
ok   T8 a successful run clears the page markers
mktemp: mkdtemp failed on /var/folders/th/ty0kb7r56dvfrkgkrcbx9fv00000gn/T/tmp.DteytdDrvu: Operation not permitted
fatal: cannot mkdir /upstream: Operation not permitted
test_dispatch_pinned.sh: line 25: /upstream/kipi-dispatch.sh: No such file or directory
test_dispatch_pinned.sh: line 29: /upstream/marker.txt: No such file or directory
fatal: cannot change to '/upstream': No such file or directory
fatal: cannot change to '/upstream': No such file or directory
fatal: repository '/upstream' does not exist
test_dispatch_pinned.sh: line 236: /pinned: Operation not permitted
test_dispatch_pinned.sh: line 237: /notify-broken.sh: Operation not permitted
chmod: /notify-broken.sh: No such file or directory
ok   T9 a failed send leaves no dedupe marker
cp: /notify.sh: No such file or directory
FAIL T9 the first tick after recovery pages
     the outage stayed silent after the notifier came back
test_dispatch_pinned.sh: line 256: /wrapper-r2.sh: Operation not permitted
chmod: /wrapper-r2.sh: No such file or directory
grep: /wrapper-r2.sh: No such file or directory
FAIL T9-neg the mutant was actually applied
     the unconditional-stamp variant was not produced
mktemp: mkdtemp failed on /var/folders/th/ty0kb7r56dvfrkgkrcbx9fv00000gn/T/tmp.SKEicOoo7n: Operation not permitted
fatal: cannot mkdir /upstream: Operation not permitted
test_dispatch_pinned.sh: line 25: /upstream/kipi-dispatch.sh: No such file or directory
test_dispatch_pinned.sh: line 29: /upstream/marker.txt: No such file or directory
fatal: cannot change to '/upstream': No such file or directory
fatal: cannot change to '/upstream': No such file or directory
fatal: repository '/upstream' does not exist
test_dispatch_pinned.sh: line 261: /pinned: Operation not permitted
cp: /notify-broken.sh: No such file or directory
FAIL T9-neg the assertion CAN fail
     the mutant behaved like the fix, so T9 proves nothing
FAIL: 13 check(s)

exec
import json
p=json.load(open('instance-registry.json'))
for x in p.get('instances',[]):
 print(x)
PY
nl -ba q-system/.q-system/scripts/linear-worker.sh | sed -n '180,330p'; rg -n \"repo identity|MISCONFIG|repo_name|repo_ident|basename\" q-system/.q-system/scripts/linear-worker.sh" in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-186
 succeeded in 0ms:
   180	# KIPI_TARGET_REPO is the env form, and it exists for one concrete reason:
   181	# converge.sh drives this script through $WORKER_CMD and forwards only its own
   182	# arguments. The env crosses that boundary by inheritance, so a dispatched
   183	# converge run reaches the right repo without converge.sh needing to change.
   184	TARGET_REPO="${TARGET_REPO_ARG:-${KIPI_TARGET_REPO:-$SKEL}}"
   185	if [ ! -d "$TARGET_REPO" ]; then
   186	  echo "--repo: no such directory: $TARGET_REPO" >&2; exit 1
   187	fi
   188	TARGET_REPO="$(cd "$TARGET_REPO" && pwd)"
   189	export TARGET_REPO
   190	# The target's slug, resolved ONCE. Every gh scope and every artifact path in
   191	# this script reads it from here (ASK-738).
   192	TARGET_SLUG="$(slug_for_repo "$TARGET_REPO" "${KIPI_SLUG_REGISTRY:-$SKEL/instance-registry.json}")"
   193	
   194	# --- WHICH REPO EVERY `gh` CALL ASKS ABOUT (ASK-738) ----------------------
   195	# `git -C "$TARGET_REPO"` redirects the git half of this script. It does NOT
   196	# redirect `gh`, which resolves its repo from the PROCESS CWD -- and
   197	# kipi-dispatch.sh:205 leaves that cwd in the HOME checkout. Measured: the
   198	# existing-PR lookup below answered about kipi-system while the work happened
   199	# in the target. Derived ONCE, here, and spliced into every gh call as -R.
   200	# Empty for a repo with no pinned remote and no origin: the calls then behave
   201	# exactly as they did before, which is correct only because that case is the
   202	# dispatcher's own checkout.
   203	KIPI_GH_REPO_ARGS="$(gh_repo_args "$TARGET_SLUG")"
   204	export KIPI_GH_REPO_ARGS
   205	
   206	export SCRIPT_DIR
   207	mkdir -p "$STATE_DIR"
   208	TS() { date -u +%Y-%m-%dT%H:%M:%SZ; }
   209	say() { echo "$(TS) $*" | tee -a "$LOG"; }
   210	
   211	# A distinct session token per run. The claim lock keys collisions on (agent,
   212	# session), so without a unique token two overlapping runs would look like the
   213	# same session and BOTH be granted -- the exact scar the lock exists to stop.
   214	SESSION="worker-$(date +%s)-$$"
   215	AGENT="sana"
   216	
   217	# Wall clock, exit 4 of loop-exits.md. macOS ships NEITHER `timeout` nor `gtimeout`
   218	# unless coreutils is installed, and the old fallback here was an empty string --
   219	# i.e. silently no wall clock at all, which is the runaway-agent case this exit
   220	# exists for. Never degrade a safety exit to a warning: implement it in bash.
   221	#
   222	# THE JOB GETS ITS OWN PROCESS GROUP, and nothing it started outlives it (Codex
   223	# round 3 on PR #141, major). `kill -TERM "$job"` signals ONE pid -- the
   224	# `bash -c` wrapper -- and the agent runs as its GRANDchild, so the wrapper died
   225	# on time and `claude` kept running. Worse, `wait "$job"` returns the instant
   226	# that wrapper dies, and the next line cancels the watchdog, so the TERM ->
   227	# sleep 5 -> KILL escalation never reached the process that was actually stuck.
   228	#
   229	# What that costs is not a stray process, it is a WRONG LINEAR COMMENT. The
   230	# leaked agent still holds the worktree, and worktrees are reused across issues
   231	# by design: it writes `.sana-blocked-capability` minutes later, after the next
   232	# dispatch has cleared the sentinels and while its own agent is running. The
   233	# worker reads a refusal the current issue's agent never wrote, labels the wrong
   234	# issue blocked:capability, and copies the previous issue's reason into Linear.
   235	# The clear one round earlier cannot help -- a clear cannot outrun a writer that
   236	# is still running.
   237	#
   238	# `set -m` is load-bearing, not decoration. Without job control a background job
   239	# INHERITS the worker's own process group, so `kill -- -$job` would either fail
   240	# (no such group) or signal the worker itself. With it, the job is the leader of
   241	# a group of its own and the whole tree under it can be signalled by pgid.
   242	run_bounded() {  # run_bounded <seconds> <cmd...>
   243	  local secs="$1"; shift
   244	  set -m
   245	  "$@" &
   246	  local job=$!
   247	  set +m
   248	  ( sleep "$secs"; kill -0 "$job" 2>/dev/null && {
   249	      echo "$(TS) TIMEOUT after ${secs}s; killing process group $job" >>"$LOG"
   250	      kill -TERM -"$job" 2>/dev/null
   251	      sleep 5
   252	      kill -KILL -"$job" 2>/dev/null
   253	    } ) &
   254	  local watchdog=$!
   255	  wait "$job"; local rc=$?
   256	  kill "$watchdog" 2>/dev/null   # cancel the watchdog if the job finished first
   257	  wait "$watchdog" 2>/dev/null
   258	  # SYNCHRONOUS, and on the success path too: a job that exited 0 can still have
   259	  # left a child behind, and the caller's next act is to dispatch another agent
   260	  # into the same tree. run_bounded returning is the promise that the previous
   261	  # run is over.
   262	  reap_group "$job"
   263	  return "$rc"
   264	}
   265	
   266	# reap_group <pgid>: return only when that process group is empty.
   267	#
   268	# `kill -0 -- -<pgid>` is the read, not `ps`: it answers "does any process in
   269	# this group still exist" with the same permission check the kills use. The
   270	# common case is an already-empty group, which costs one syscall and no wait.
   271	#
   272	# TERM first with the same 5s grace the watchdog uses -- a child mid-write gets
   273	# the same chance to flush it had before -- then KILL, which cannot be ignored.
   274	# Both loops are bounded: this runs on the worker's critical path, and a reaper
   275	# that can hang is a worse failure than the leak it fixes.
   276	reap_group() {
   277	  local pgid="$1" i=0
   278	  kill -0 -- -"$pgid" 2>/dev/null || return 0
   279	  echo "$(TS) reaping leftover process group $pgid" >>"$LOG"
   280	  kill -TERM -- -"$pgid" 2>/dev/null
   281	  while [ "$i" -lt 20 ] && kill -0 -- -"$pgid" 2>/dev/null; do sleep 0.25; i=$((i+1)); done
   282	  kill -0 -- -"$pgid" 2>/dev/null || return 0
   283	  kill -KILL -- -"$pgid" 2>/dev/null
   284	  i=0
   285	  while [ "$i" -lt 20 ] && kill -0 -- -"$pgid" 2>/dev/null; do sleep 0.1; i=$((i+1)); done
   286	  if kill -0 -- -"$pgid" 2>/dev/null; then
   287	    echo "$(TS) WARNING: process group $pgid survived SIGKILL" >>"$LOG"
   288	  fi
   289	}
   290	
   291	# --- FETCH ONCE, BEFORE ANY WORKTREE EXISTS ---------------------------------
   292	# ASK-211 (sp-28ced3d6). This script used to contain no `git fetch` at all: it
   293	# cut every worktree from whatever local origin/main ref happened to be lying
   294	# around, so the agent was dispatched against a base that could be arbitrarily
   295	# old. Observed 2026-07-27 -- ASK-150 was sent to resolve a conflict against
   296	# main, merged 3b60af0, and the conflict survived because main was already
   297	# 72c782d. The agent did the right thing to the wrong target and two rounds
   298	# were burned.
   299	#
   300	# ONCE PER RUN, not per issue: origin does not move meaningfully inside one run,
   301	# and a 50-issue board would otherwise pay 50 network round-trips to learn the
   302	# same thing. Placed above the picker so no code path can create a worktree
   303	# before it.
   304	#
   305	# A fetch failure is environmental (self-healing-retry.md rule 5), so it stops
   306	# on attempt 1 and is NOT counted against any issue. Continuing would be worse
   307	# than stopping: the whole point is that a stale base silently produces
   308	# plausible work aimed at the wrong target, and the run could not push or open
   309	# a PR against an unreachable origin anyway.
   310	#
   311	# IT PAGES AND EXITS 9, it does not stop quietly (PR #22 review round 3,
   312	# finding 1 -- major). The first version of this guard was `say` + `exit 0`,
   313	# which made an expired credential at 3am byte-for-byte indistinguishable from a
   314	# healthy run with nothing ready: same rc, no Slack, one line in a log nobody
   315	# reads. The issue never became stuck either, because MAX_ATTEMPTS only counts
   316	# DISPATCHED runs. Rule 5 says surface it IMMEDIATELY, and a log line is not
   317	# surfacing.
   318	#
   319	# 9, not 1: 1 is the usage error above, and a caller has to be able to tell an
   320	# environment that is down from a worker that was invoked wrong.
   321	if ! git -C "$TARGET_REPO" fetch --quiet origin 2>>"$LOG"; then
   322	  say "INFRA: git fetch failed in $TARGET_REPO. Stopping before any worktree is cut from a stale base."
   323	  bash "$NOTIFY" "worker: git fetch failed in $TARGET_REPO -- the run did NO work. Check credentials/network." 2>/dev/null || true
   324	  exit 9
   325	fi
   326	
   327	# --- repo identity, for the project-scope filter -----------------------------
   328	# The worker cuts EVERY worktree from $TARGET_REPO (which defaults to $SKEL). An issue filed against
   329	# another repo can therefore never reach a terminal state here: the agent lands
   330	# in kipi-system, cannot find the files the DoR names, and exits 0 with no diff.
327:# --- repo identity, for the project-scope filter -----------------------------
338:# disabling the filter here. basename of the checkout is the convention the
343:# basename is the LAST resort, not the first. instance-registry.json maps every
346:#   <Persona>_strategy   -> .../<persona>/projects/strategy (basename: strategy)
347:#   <Persona>_product    -> .../<persona>/projects/product  (basename: product)
348:#   <Persona>_consultant -> .../consulting                  (basename: consulting)
349:# Three of the first three checked. Shipping basename alone would have made this
350:# filter reject every issue on those instances -- caught loud by the MISCONFIG
353:# Order: explicit env override, then the registry, then basename.
358:# the second: it returns this repo identity AND the set of project names whose
363:# reading it from the target would fall through to basename for every repo and
398:[ -n "$REPO_PROJECT" ] || REPO_PROJECT="$(basename "$TARGET_REPO")"
539:    # Does the derived repo identity name a project that EXISTS on the board?
542:    # worse than the bug being fixed here, so it is called out as MISCONFIG
558:# MISCONFIG BEFORE ANY COUNT IS BELIEVED (ASK-275). Checked ahead of the ready
567:  say "MISCONFIG: repo identity '$REPO_PROJECT' (from $TARGET_REPO) matches NO Linear project on team ASK."
568:  say "MISCONFIG: every issue would be filtered out, so this run picked nothing for a config reason, not an empty board."
569:  say "MISCONFIG: fix by renaming the Linear project to match the checkout, or set KIPI_LINEAR_PROJECT."
570:  bash "$NOTIFY" "kipi worker: repo identity '$REPO_PROJECT' matches no Linear project, so the queue reads empty and NO work can ever be picked. Do: set KIPI_LINEAR_PROJECT in the worker's environment, or rename the project to match the checkout." 2>/dev/null || true

exec
/bin/zsh -lc "python3 -c \"import json; d=json.load(open('instance-registry.json')); print(*d.get('instances',[]), sep='\\\\n')\"; nl -ba q-system/.q-system/scripts/linear-worker.sh | sed -n '327,410p'; nl -ba q-system/.q-system/scripts/linear-worker.sh | sed -n '530,580p'; git remote get-url origin; basename \""'$(git remote get-url origin . 2>/dev/null || git remote get-url origin)" 2>/dev/null || true' in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-186
 succeeded in 0ms:
{'name': 'KTLYST_strategy', 'path': '/Users/assafkipnis/projects/cole-gtm/projects/strategy', 'subtree_prefix': 'q-system', 'instance_q_dir': 'q-ktlyst', 'type': 'subtree', 'has_git': True}
{'name': 'ktlyst', 'path': '/Users/assafkipnis/projects/ktlyst-saas/projects/product', 'subtree_prefix': 'q-system', 'instance_q_dir': None, 'type': 'subtree', 'has_git': True, 'dispatch': {'enabled': True, 'expected_remote': 'https://github.com/assafkip/ktlyst-saas-product.git'}}
{'name': 'ASK_AI_consultant', 'path': '/Users/assafkipnis/projects/consulting', 'subtree_prefix': 'q-system', 'instance_q_dir': 'q-consult', 'type': 'subtree', 'has_git': True}
{'name': 'ktlyst-website', 'path': '/Users/assafkipnis/projects/cole-gtm/projects/website', 'subtree_prefix': 'q-system', 'instance_q_dir': None, 'type': 'subtree', 'has_git': True, 'skip_agent_check': True, 'note': 'Website project, no agent pipeline needed'}
{'name': '4_points_consulting', 'path': '/Users/assafkipnis/projects/consulting/projects/4_points_consulting', 'subtree_prefix': 'q-system', 'instance_q_dir': 'q-investigate', 'type': 'subtree', 'has_git': True, 'note': 'OSINT investigation system with custom q-investigate/ pipeline'}
{'name': 'Pure_spectrum_Q', 'path': '/Users/assafkipnis/projects/consulting/projects/Pure_spectrum_Q', 'subtree_prefix': 'q-system', 'instance_q_dir': 'q-pure', 'type': 'subtree', 'has_git': True, 'note': 'Client advisory hub, instance content in q-pure/'}
{'name': 'ktlyst_lawyer', 'path': '/Users/assafkipnis/projects/consulting/projects/lawyer', 'subtree_prefix': 'q-system', 'instance_q_dir': None, 'type': 'subtree', 'has_git': True}
{'name': 'personal-brand', 'path': '/Users/assafkipnis/projects/cole-gtm/projects/personal-brand', 'subtree_prefix': 'q-system', 'instance_q_dir': None, 'type': 'subtree', 'has_git': True}
{'name': 'AUDHD_KIDS', 'path': '/Users/assafkipnis/projects/personal/projects/AUDHD_KIDS', 'subtree_prefix': 'q-system', 'instance_q_dir': None, 'type': 'subtree', 'has_git': True}
{'name': 'event_coordinator', 'path': '/Users/assafkipnis/projects/cole-gtm/projects/event_coordinator', 'subtree_prefix': 'q-system', 'instance_q_dir': None, 'type': 'subtree', 'has_git': True}
{'name': 'investigations', 'path': '/Users/assafkipnis/projects/intel/projects/kipi-investigations', 'subtree_prefix': 'q-system', 'instance_q_dir': None, 'type': 'subtree', 'has_git': True}
{'name': 'accountant', 'path': '/Users/assafkipnis/projects/consulting/projects/accountant', 'subtree_prefix': 'q-system', 'instance_q_dir': None, 'type': 'subtree', 'has_git': True}
{'name': 'gtm-partner', 'path': '/Users/assafkipnis/projects/cole-gtm', 'subtree_prefix': 'q-system', 'instance_q_dir': None, 'type': 'subtree', 'has_git': True}
{'name': 'travel-agent', 'path': '/Users/assafkipnis/projects/personal/projects/travel-agent', 'subtree_prefix': 'q-system', 'instance_q_dir': None, 'type': 'subtree', 'has_git': True}
{'name': 'fractional-cxo', 'path': '/Users/assafkipnis/projects/consulting/projects/fractional-cxo', 'subtree_prefix': 'q-system', 'instance_q_dir': None, 'type': 'subtree', 'has_git': True}
{'name': 'interview-coach', 'path': '/Users/assafkipnis/projects/micro-saas/projects/interview-coach', 'subtree_prefix': 'q-system', 'instance_q_dir': None, 'type': 'subtree', 'has_git': True, 'dispatch': {'enabled': True, 'expected_remote': 'https://github.com/assafkip/kipi-interview-coach.git'}}
{'name': 'negotiator', 'path': '/Users/assafkipnis/projects/micro-saas/projects/negotiator', 'subtree_prefix': 'q-system', 'instance_q_dir': None, 'type': 'subtree', 'has_git': True}
{'name': 'reddit-build-radar', 'path': '/Users/assafkipnis/projects/cole-gtm/projects/reddit-build-radar', 'subtree_prefix': None, 'instance_q_dir': None, 'type': 'standalone', 'has_git': True, 'skip_agent_check': True, 'skeleton_managed': False, 'note': 'Standalone Reddit consensus signal engine for weekly build recommendations. No q-system subtree because local git lacks git-subtree. skeleton_managed=false is the deliberate record (ASK-117): it receives zero skeleton propagation, so it carries none of the fleet enforcement layer (no token-guard.py, no .claude/rules/, no capability gate) and must be excluded from fleet-wide governance counts rather than counted as a governed instance.'}
{'name': 'Alice', 'path': '/Users/assafkipnis/projects/consulting/projects/Alice', 'subtree_prefix': 'q-system', 'instance_q_dir': None, 'type': 'subtree', 'has_git': True}
{'name': 'all_points_setup', 'path': '/Users/assafkipnis/projects/consulting/projects/all_points_setup', 'subtree_prefix': 'q-system', 'instance_q_dir': None, 'type': 'subtree', 'has_git': True}
{'name': 'Prodigy_Gold', 'path': '/Users/assafkipnis/projects/consulting/projects/Prodigy_Gold', 'subtree_prefix': 'q-system', 'instance_q_dir': None, 'type': 'subtree', 'has_git': True}
{'name': 'thaena', 'path': '/Users/assafkipnis/projects/consulting/projects/thaena', 'subtree_prefix': 'q-system', 'instance_q_dir': 'q-thaena', 'type': 'subtree', 'has_git': True}
{'name': 'norri ai', 'path': '/Users/assafkipnis/projects/consulting/projects/norri-ai', 'subtree_prefix': 'q-system', 'instance_q_dir': None, 'type': 'subtree', 'has_git': True}
   327	# --- repo identity, for the project-scope filter -----------------------------
   328	# The worker cuts EVERY worktree from $TARGET_REPO (which defaults to $SKEL). An issue filed against
   329	# another repo can therefore never reach a terminal state here: the agent lands
   330	# in kipi-system, cannot find the files the DoR names, and exits 0 with no diff.
   331	# Measured against the live board 2026-07-30: of 29 ready issues, 11 were the
   332	# kipi-system project and 18 were for repos this checkout cannot check out --
   333	# under two thirds of the worker's own queue was undispatchable by construction.
   334	# The GraphQL query below has selected project{name} the whole time; nothing read it.
   335	#
   336	# DERIVED, never hardcoded: this same script runs from other skeletons, and a
   337	# literal "kipi-system" would empty every other instance's queue while silently
   338	# disabling the filter here. basename of the checkout is the convention the
   339	# fleet already uses (instance dir name == Linear project name).
   340	#
   341	# KIPI_LINEAR_PROJECT is the override for the instance where those two names
   342	# legitimately differ. It is also the test seam.
   343	# basename is the LAST resort, not the first. instance-registry.json maps every
   344	# instance path to its name, and that name IS the Linear project name -- while the
   345	# directory very often is not:
   346	#   <Persona>_strategy   -> .../<persona>/projects/strategy (basename: strategy)
   347	#   <Persona>_product    -> .../<persona>/projects/product  (basename: product)
   348	#   <Persona>_consultant -> .../consulting                  (basename: consulting)
   349	# Three of the first three checked. Shipping basename alone would have made this
   350	# filter reject every issue on those instances -- caught loud by the MISCONFIG
   351	# guard below rather than as a silently empty queue, but still wrong.
   352	#
   353	# Order: explicit env override, then the registry, then basename.
   354	# ONE REGISTRY READ, TWO FACTS (ASK-729). sp-421fa27d already records that the
   355	# project-name derivation is duplicated between this script and
   356	# spillover-promote.py. Reachability needs the SAME registry, so the fix was
   357	# either a second reader here or one read that answers both questions. This is
   358	# the second: it returns this repo identity AND the set of project names whose
   359	# checkout exists on this machine, from a single parse.
   360	#
   361	# The path being looked UP is the target; the registry doing the looking up is
   362	# always the skeleton. An instance carries no instance-registry.json, so
   363	# reading it from the target would fall through to basename for every repo and
   364	# quietly re-break the filter for exactly the three instances named below.
   365	#
   366	# registry_ok is carried explicitly and is NOT the same as an empty list. An
   367	# unreadable registry means reachability is UNKNOWN, and reporting unknown as
   368	# unreachable would fire a loud false alarm on every project at once -- the
   369	# failure mode this issue exists to remove, pointed the other way.
   370	REGISTRY_FACTS="$(SKEL_PATH="$TARGET_REPO" REG="$SKEL/instance-registry.json" python3 - <<'PY' 2>/dev/null
   371	import json, os
   372	skel = os.path.realpath(os.environ["SKEL_PATH"])
   373	ok = True
   374	try:
   375	    reg = json.load(open(os.environ["REG"]))
   376	except Exception:
   377	    reg, ok = [], False
   378	entries = reg.get("instances", reg) if isinstance(reg, dict) else reg
   379	name = ""
   380	local = []
   381	for e in entries if isinstance(entries, list) else []:
   382	    if not isinstance(e, dict):
   383	        continue
   384	    p, n = e.get("path"), e.get("name") or ""
   385	    if not p:
   386	        continue
   387	    if not name and os.path.realpath(p) == skel:
   388	        name = n
   389	    if n and os.path.isdir(p):
   390	        local.append(n)
   391	print(json.dumps({"name": name, "local_projects": sorted(set(local)), "ok": ok}))
   392	PY
   393	)"
   394	_facts_get() { printf '%s' "$REGISTRY_FACTS" | python3 -c "import json,sys;d=json.load(sys.stdin);v=d.get('$1');print('\n'.join(v) if isinstance(v,list) else v)" 2>/dev/null; }
   395	
   396	REPO_PROJECT="${KIPI_LINEAR_PROJECT:-}"
   397	[ -n "$REPO_PROJECT" ] || REPO_PROJECT="$(_facts_get name)"
   398	[ -n "$REPO_PROJECT" ] || REPO_PROJECT="$(basename "$TARGET_REPO")"
   399	export REPO_PROJECT
   400	
   401	LOCAL_PROJECTS="$(_facts_get local_projects)"
   402	REGISTRY_OK="$(_facts_get ok)"
   403	export LOCAL_PROJECTS REGISTRY_OK
   404	
   405	# --- pick ready issues ------------------------------------------------------
   406	PICKED="$(python3 - "$ONLY_ISSUE" <<'PY'
   407	import importlib.util, json, os, sys, pathlib
   408	here = pathlib.Path(os.environ["SCRIPT_DIR"])
   409	spec = importlib.util.spec_from_file_location("ls", here / "linear-sync.py")
   410	ls = importlib.util.module_from_spec(spec); spec.loader.exec_module(ls)
   530	    "skipped_local": len(skipped_local),
   531	    "skipped_local_projects": sorted({project_of(i) or "(unset)" for i in skipped_local}),
   532	    "unreachable": len(unreachable),
   533	    "unreachable_projects": sorted({project_of(i) or "(unset)" for i in unreachable}),
   534	    "unreachable_ids": [i["identifier"] for i in unreachable],
   535	    "registry_ok": registry_ok,
   536	    "deferred_needs_scope": len(deferred),
   537	    "blocked_capability": len(blocked_cap),
   538	    "blocked_capability_ids": [i["identifier"] for i in blocked_cap],
   539	    # Does the derived repo identity name a project that EXISTS on the board?
   540	    # If not, every issue fails in_this_repo() and the queue reads a healthy
   541	    # empty -- a permanently silent loop reporting success. That is strictly
   542	    # worse than the bug being fixed here, so it is called out as MISCONFIG
   543	    # rather than allowed to look like a finished board.
   544	    "project_known": any(project_of(i) == repo_project for i in issues),
   545	    "repo_project": repo_project,
   546	}))
   547	PY
   548	)"
   549	
   550	INFRA="$(printf '%s' "$PICKED" | python3 -c 'import json,sys;print(json.load(sys.stdin).get("infra_error",""))' 2>/dev/null)"
   551	if [ -n "$INFRA" ]; then
   552	  say "INFRA: linear unreachable ($INFRA). Not counted against any issue."
   553	  exit 0
   554	fi
   555	
   556	READY_COUNT="$(printf '%s' "$PICKED" | python3 -c 'import json,sys;print(len(json.load(sys.stdin)["ready"]))')"
   557	
   558	# MISCONFIG BEFORE ANY COUNT IS BELIEVED (ASK-275). Checked ahead of the ready
   559	# line because if the derived project matches nothing on the board, that count
   560	# is 0 for a reason that has nothing to do with the board being empty. Reporting
   561	# "nothing ready" there would be a loop that stopped working while claiming to
   562	# be finished -- the one outcome worse than the queue it replaced. It PAGES and
   563	# exits non-zero for the same reason the git-fetch guard at line 203 does: a log
   564	# line is not surfacing, and this is environmental, not an issue's fault.
   565	PROJECT_KNOWN="$(printf '%s' "$PICKED" | python3 -c 'import json,sys;print(json.load(sys.stdin).get("project_known"))' 2>/dev/null)"
   566	if [ "$PROJECT_KNOWN" = "False" ]; then
   567	  say "MISCONFIG: repo identity '$REPO_PROJECT' (from $TARGET_REPO) matches NO Linear project on team ASK."
   568	  say "MISCONFIG: every issue would be filtered out, so this run picked nothing for a config reason, not an empty board."
   569	  say "MISCONFIG: fix by renaming the Linear project to match the checkout, or set KIPI_LINEAR_PROJECT."
   570	  bash "$NOTIFY" "kipi worker: repo identity '$REPO_PROJECT' matches no Linear project, so the queue reads empty and NO work can ever be picked. Do: set KIPI_LINEAR_PROJECT in the worker's environment, or rename the project to match the checkout." 2>/dev/null || true
   571	  exit 9
   572	fi
   573	
   574	DROPPED="$(printf '%s' "$PICKED" | python3 -c 'import json,sys;print(json.load(sys.stdin).get("skipped_local",0))' 2>/dev/null)"
   575	DROPPED_IN="$(printf '%s' "$PICKED" | python3 -c 'import json,sys;print(", ".join(json.load(sys.stdin).get("skipped_local_projects",[])))' 2>/dev/null)"
   576	UNREACH="$(printf '%s' "$PICKED" | python3 -c 'import json,sys;print(json.load(sys.stdin).get("unreachable",0))' 2>/dev/null)"
   577	UNREACH_IN="$(printf '%s' "$PICKED" | python3 -c 'import json,sys;print(", ".join(json.load(sys.stdin).get("unreachable_projects",[])))' 2>/dev/null)"
   578	REG_OK="$(printf '%s' "$PICKED" | python3 -c 'import json,sys;print(json.load(sys.stdin).get("registry_ok",True))' 2>/dev/null)"
   579	DEFERRED="$(printf '%s' "$PICKED" | python3 -c 'import json,sys;print(json.load(sys.stdin).get("deferred_needs_scope",0))' 2>/dev/null)"
   580	
https://github.com/assafkip/kipi-system.git
kipi-system.git

exec
/bin/zsh -lc "printf '%s\\n' 'REPRO 1: fallback announcement ordering'; awk 'NR>=80&&NR<=90 {printf \"%d:%s\\n\",NR,"'$0}'"' kipi-dispatch-pinned.sh; printf '%s\\n' 'REPRO 2: stale phone/page claims versus notifier contract'; awk 'NR>=99&&NR<=105 || NR>=131&&NR<=153 {printf \"%d:%s\\n\",NR,"'$0}'"' kipi-dispatch-pinned.sh; sed -n '1,45p' q-system/.q-system/scripts/slack-notify.sh; printf '%s\\n' 'REPRO 3: changed behavior has no default-path test'; awk 'NR>=35&&NR<=41 {printf \"%d:%s\\n\",NR,"'$0}'"' test_dispatch_pinned.sh; git diff 82ffa21b..HEAD -- test_dispatch_pinned.sh; printf 'test diff bytes='; git diff 82ffa21b..HEAD -- test_dispatch_pinned.sh | wc -c" in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-186
 succeeded in 0ms:
REPRO 1: fallback announcement ordering
80:_IDENT="$(_repo_ident || true)"
81:# Fall back to the old literal ONLY if the remote cannot be read. That is worse
82:# than useless for project resolution, so it is announced rather than silent.
83:if [ -z "$_IDENT" ]; then
84:  _IDENT="dispatch-checkout"
85:fi
86:PINNED="${KIPI_DISPATCH_CHECKOUT:-$HOME/.local/state/kipi/dispatch/$_IDENT}"
87:LOG="${KIPI_DISPATCH_LOG:-$HOME/.config/kipi/dispatch.log}"
88:
89:say() { printf '%s %s\n' "$(date -u +%FT%TZ)" "pinned: $*" >>"$LOG" 2>/dev/null || true; }
90:
REPRO 2: stale phone/page claims versus notifier contract
99:# WHERE THAT ALERT ACTUALLY GOES (corrected, sp-d9212e22). An earlier version of
100:# this comment -- mine -- said it "reaches the founder's phone". It does not.
101:# slack-notify.sh keeps its name for ~30 callers but its header is explicit:
102:# "THE FLEET ALERT PATH. Files a Linear ticket for Sana. Pages nobody." Nothing
103:# in it sends to Slack. That is the founder's stated routing, not a degradation:
104:# things needing attention go to Sana. Saying "pages the founder" in a comment
105:# would set the wrong expectation about who is watching at 3am.
131:  if [ -f "$mark" ]; then
132:    last="$(cat "$mark" 2>/dev/null || echo 0)"
133:    case "$last" in ''|*[!0-9]*) last=0 ;; esac
134:    # Still inside the window: log it, stay quiet. The log always has every
135:    # occurrence; only the phone is rate-limited.
136:    if [ $((now - last)) -lt "$PAGE_TTL" ]; then
137:      say "page suppressed ($key): already sent $(( (now - last) / 60 ))m ago"
138:      return 0
139:    fi
140:  fi
141:  # ONLY A DELIVERED PAGE BURNS THE WINDOW (codex major, PR #122 r3).
142:  #
143:  # Round 2 stamped the marker after the send ATTEMPT and the comment claimed
144:  # that protected a down notifier. It did not: "after the attempt" is not "on
145:  # success", so a Slack outage wrote a 24h mute and the dispatch outage went
146:  # unreported for a day after the notifier came back. That is the same
147:  # comment-says-one-thing-code-does-another shape this session already fixed in
148:  # the fleet drift detector, rebuilt here by the fix for the previous finding.
149:  #
150:  # So the exit status decides. A notifier that is absent or fails leaves NO
151:  # marker, and the next tick 900s later tries again. The cost of getting this
152:  # wrong is asymmetric: an extra page is noise, a swallowed page is the silent
153:  # outage the whole file exists to end.
#!/bin/bash
# THE FLEET ALERT PATH. Files a Linear ticket for Sana. Pages nobody.
#
# FOUNDER-DIRECTED 2026-08-10: "I dont want to see any of these. Any of the ones
# that need attention should go to Sana - not me." Offered one carve-out so the
# fable-escalation cap could still reach him, the answer was "the fable
# escalation should be gone. I gave it clear instructions." So there is no
# founder path in this script, no flag that re-opens one, and no webhook call.
#
# THE NAME IS NOW WRONG ON PURPOSE. It is still slack-notify.sh because ~30 call
# sites across six repos invoke it by that path, and a rename is a separate,
# larger change than switching the destination. Renaming it and updating every
# caller is captured in .prd-os/spillover.jsonl rather than bundled in here.
# Nothing in this file sends to Slack.
#
# WHY IT CHANGED. #general carried 100 messages between 09:35 and 14:06 PDT on
# 2026-08-10. 51 were auto-commit naming a file set that changed every turn; 35
# were one cole-gtm notice re-announcing a config state once per run, in
# duplicate pairs. The six that mattered -- four security reverts of unsanctioned
# .claude/ changes and a Notion job dead since 13:00 -- were unreadable
# underneath. The destination changed AND both flood sources were fixed at the
# emitter in the same change; routing a flood to Linear would only make it
# harder to clear, since a ticket has to be closed by hand.
#
# Repeats are deduped by alert-to-linear.py: one ticket per alert SHAPE, with a
# counter, not one ticket per firing.
#
# EXIT CONTRACT, unchanged from when this sent to Slack, so callers that read the
# status (kipi-dispatch.sh:197) keep working:
#   0  filed (ticket opened, or a repeat recorded on the open one)
#   1  attempted and FAILED (reason + the message text on stderr)
#   3  nothing to file on: no Linear API key configured (a setup state)
#   4  refused: fixture run
#
# Usage: slack-notify.sh "message text"
set -uo pipefail

MSG="${1:-}"
[ -n "$MSG" ] || exit 0

# Project label so a ticket always names which instance raised it. Resolved in
# order: KIPI_INSTANCE_NAME (set by the fleet heartbeat = exact registry name)
# -> git repo root basename -> cwd basename. Every message is prefixed "[label] ".
LABEL="${KIPI_INSTANCE_NAME:-}"
if [ -z "$LABEL" ]; then
REPRO 3: changed behavior has no default-path test
35:
36:run_wrapper() {  # $1=root ; echoes stdout, log goes to $1/log
37:  KIPI_REPO_MAIN="$1/main" \
38:  KIPI_DISPATCH_CHECKOUT="$1/pinned" \
39:  KIPI_DISPATCH_LOG="$1/log" \
40:    bash "$2" 2>&1
41:}
test diff bytes=       0

exec
/bin/zsh -lc "bash -n kipi-dispatch-pinned.sh && echo 'bash -n: PASS'; for url in 'https://github.com/assafkip/kipi-system.git' 'git@github.com:assafkip/kipi-system.git' 'ssh://git@github.com/assafkip/kipi-system.git' '/srv/git/kipi-system.git'; do base=\""'${url##*/}"; base="${base%.git}"; printf '"'%s -> %s\\n' \""'$url" "$base"; done; printf '"'actual origin -> '; url=\""'$(git remote get-url origin)"; base="${url##*/}"; base="${base%.git}"; printf '"'%s\\n' \""'$base"; git diff --check 82ffa21b..HEAD && echo '"'git diff --check: PASS'" in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-186
 succeeded in 0ms:
bash -n: PASS
https://github.com/assafkip/kipi-system.git -> kipi-system
git@github.com:assafkip/kipi-system.git -> kipi-system
ssh://git@github.com/assafkip/kipi-system.git -> kipi-system
/srv/git/kipi-system.git -> kipi-system
actual origin -> kipi-system
git diff --check: PASS

codex
## Findings

### 1. MINOR, Confirmed

The remote-read fallback is silent even though the comment promises it is announced, leaving operators without the stated diagnostic.

`kipi-dispatch-pinned.sh:82`

Reproducer:

```bash
awk 'NR>=80&&NR<=90 {printf "%d:%s\n",NR,$0}' kipi-dispatch-pinned.sh
```

Real output:

```text
80:_IDENT="$(_repo_ident || true)"
81:# Fall back to the old literal ONLY if the remote cannot be read. That is worse
82:# than useless for project resolution, so it is announced rather than silent.
83:if [ -z "$_IDENT" ]; then
84:  _IDENT="dispatch-checkout"
85:fi
86:PINNED="${KIPI_DISPATCH_CHECKOUT:-$HOME/.local/state/kipi/dispatch/$_IDENT}"
87:LOG="${KIPI_DISPATCH_LOG:-$HOME/.config/kipi/dispatch.log}"
88:
89:say() { printf '%s %s\n' "$(date -u +%FT%TZ)" "pinned: $*" >>"$LOG" 2>/dev/null || true; }
```

There is no log or notification call in the fallback branch. A human diagnosing the restored `dispatch-checkout` identity must infer why derivation did not happen.

### 2. MINOR, Confirmed

The notification comments still claim a phone page exists after this PR explicitly documents that the notifier only creates Linear tickets.

`kipi-dispatch-pinned.sh:135`

Reproducer:

```bash
awk 'NR>=99&&NR<=105 || NR>=131&&NR<=153 {printf "%d:%s\n",NR,$0}' \
  kipi-dispatch-pinned.sh
sed -n '1,14p' q-system/.q-system/scripts/slack-notify.sh
```

Real output:

```text
99:# WHERE THAT ALERT ACTUALLY GOES (corrected, sp-d9212e22). An earlier version of
100:# this comment -- mine -- said it "reaches the founder's phone". It does not.
102:# "THE FLEET ALERT PATH. Files a Linear ticket for Sana. Pages nobody."
134:    # Still inside the window: log it, stay quiet. The log always has every
135:    # occurrence; only the phone is rate-limited.
141:  # ONLY A DELIVERED PAGE BURNS THE WINDOW (codex major, PR #122 r3).
145:  # success", so a Slack outage wrote a 24h mute and the dispatch outage went
152:  # wrong is asymmetric: an extra page is noise, a swallowed page is the silent
#!/bin/bash
# THE FLEET ALERT PATH. Files a Linear ticket for Sana. Pages nobody.
...
# Nothing in this file sends to Slack.
```

This is bounded documentation drift, but it creates the exact incorrect 3am expectation the PR says it removes.

## What is sound

- `bash -n kipi-dispatch-pinned.sh` passed.
- The parser produced `kipi-system` for HTTPS, SCP-style SSH, `ssh://`, and local-path remotes.
- The actual configured origin resolves to `kipi-system`.
- `git diff --check 82ffa21b..HEAD` passed.
- The checkout remains detached and does not introduce force, reset, or clean behavior.
- I attempted the existing test suite, but the managed read-only sandbox prevents its required `$TMPDIR` fixture creation. I did not treat those sandbox errors as PR failures.
- The existing tests always set `KIPI_DISPATCH_CHECKOUT`, so they do not exercise this PR’s new default-path derivation. I did not elevate missing coverage into a separate operational finding without a reproduced behavioral defect.

## VERDICT: APPROVE WITH NITS

Most important fix first: make the fallback at line 83 emit the diagnostic promised by line 82.

FINDINGS:
minor|The remote-read fallback is silent even though the comment promises it is announced, leaving operators without the stated diagnostic.|kipi-dispatch-pinned.sh:82
minor|The notification comments still claim a phone page exists after this PR explicitly documents that the notifier only creates Linear tickets.|kipi-dispatch-pinned.sh:135
END FINDINGS
tokens used
59,590
## Findings

### 1. MINOR, Confirmed

The remote-read fallback is silent even though the comment promises it is announced, leaving operators without the stated diagnostic.

`kipi-dispatch-pinned.sh:82`

Reproducer:

```bash
awk 'NR>=80&&NR<=90 {printf "%d:%s\n",NR,$0}' kipi-dispatch-pinned.sh
```

Real output:

```text
80:_IDENT="$(_repo_ident || true)"
81:# Fall back to the old literal ONLY if the remote cannot be read. That is worse
82:# than useless for project resolution, so it is announced rather than silent.
83:if [ -z "$_IDENT" ]; then
84:  _IDENT="dispatch-checkout"
85:fi
86:PINNED="${KIPI_DISPATCH_CHECKOUT:-$HOME/.local/state/kipi/dispatch/$_IDENT}"
87:LOG="${KIPI_DISPATCH_LOG:-$HOME/.config/kipi/dispatch.log}"
88:
89:say() { printf '%s %s\n' "$(date -u +%FT%TZ)" "pinned: $*" >>"$LOG" 2>/dev/null || true; }
```

There is no log or notification call in the fallback branch. A human diagnosing the restored `dispatch-checkout` identity must infer why derivation did not happen.

### 2. MINOR, Confirmed

The notification comments still claim a phone page exists after this PR explicitly documents that the notifier only creates Linear tickets.

`kipi-dispatch-pinned.sh:135`

Reproducer:

```bash
awk 'NR>=99&&NR<=105 || NR>=131&&NR<=153 {printf "%d:%s\n",NR,$0}' \
  kipi-dispatch-pinned.sh
sed -n '1,14p' q-system/.q-system/scripts/slack-notify.sh
```

Real output:

```text
99:# WHERE THAT ALERT ACTUALLY GOES (corrected, sp-d9212e22). An earlier version of
100:# this comment -- mine -- said it "reaches the founder's phone". It does not.
102:# "THE FLEET ALERT PATH. Files a Linear ticket for Sana. Pages nobody."
134:    # Still inside the window: log it, stay quiet. The log always has every
135:    # occurrence; only the phone is rate-limited.
141:  # ONLY A DELIVERED PAGE BURNS THE WINDOW (codex major, PR #122 r3).
145:  # success", so a Slack outage wrote a 24h mute and the dispatch outage went
152:  # wrong is asymmetric: an extra page is noise, a swallowed page is the silent
#!/bin/bash
# THE FLEET ALERT PATH. Files a Linear ticket for Sana. Pages nobody.
...
# Nothing in this file sends to Slack.
```

This is bounded documentation drift, but it creates the exact incorrect 3am expectation the PR says it removes.

## What is sound

- `bash -n kipi-dispatch-pinned.sh` passed.
- The parser produced `kipi-system` for HTTPS, SCP-style SSH, `ssh://`, and local-path remotes.
- The actual configured origin resolves to `kipi-system`.
- `git diff --check 82ffa21b..HEAD` passed.
- The checkout remains detached and does not introduce force, reset, or clean behavior.
- I attempted the existing test suite, but the managed read-only sandbox prevents its required `$TMPDIR` fixture creation. I did not treat those sandbox errors as PR failures.
- The existing tests always set `KIPI_DISPATCH_CHECKOUT`, so they do not exercise this PR’s new default-path derivation. I did not elevate missing coverage into a separate operational finding without a reproduced behavioral defect.

## VERDICT: APPROVE WITH NITS

Most important fix first: make the fallback at line 83 emit the diagnostic promised by line 82.

FINDINGS:
minor|The remote-read fallback is silent even though the comment promises it is announced, leaving operators without the stated diagnostic.|kipi-dispatch-pinned.sh:82
minor|The notification comments still claim a phone page exists after this PR explicitly documents that the notifier only creates Linear tickets.|kipi-dispatch-pinned.sh:135
END FINDINGS

PR #190, opened 2026-08-15
Reachability is a fact about a checkout, and a checkout is not a permit (ASK-840)

assafkip:
## What was wrong

`linear-worker.sh has_local_checkout()` answered reachability with

```
project_of(issue) in {registry row `name` for every row whose path exists}
```

That is a string equality between two namespaces nothing ever required to agree.
Measured against the live board on 2026-08-15, they do not:

| Linear project | Registry name | Path exists |
| -- | -- | -- |
| `ASK Consulting` | `ASK_AI_consultant` | YES |
| `cole-GTM` | `gtm-partner` | YES |

14 + 3 = **17 of the 44 UNREACHABLE issues** were on repos cloned and registered
that minute. The line telling the operator to "clone and register those repos"
asked for work already done.

## What changed

**The alias is a field, not a guess.** `instance-registry.json` gains
`linear_project` on the two rows whose board name differs from their registry
name. A name-derivation guess produced this bug; a smarter guess would only move
the day it breaks. One function, `linear_project()`, answers "what is this row
called on the board" for BOTH the repo-identity lookup and the reachability set,
so the two sides cannot drift again.

**A third bucket, because two of them were wrong for the same repos.** Resolving
the alias moves 17 issues out of UNREACHABLE, and dropping them in with the
routine skips would swap one false sentence for another: "the rotation reaches it
on a later turn" is false *forever* for a client engagement repo. So reachability
answers only "is there a checkout", and the verdict from the **real
`repo-preflight.sh`** splits what is left. Check 0 is not weakened, not bypassed,
and consulted no later than before -- the worker executes the gate rather than
re-implementing its rules, so a client repo stays refused whether or not its name
resolves.

**Also fixed on the way (it was blocking the fix, not scope creep):** the reachable
rows crossed from python to the shell tab-separated, and tab is an IFS *whitespace*
character -- bash collapses runs of it however IFS is set. A row with an empty
`expected_remote` (the common case; only two rows pin one) shifted its fields left,
landing the count in `$_remote` and leaving `$_cnt` unset, so every reachable
project scored 0 issues and both report lines vanished. Now `\037`.

## Evidence

Reproducers first, RED at HEAD on the exact live defect:

```
$ python3 -m pytest q-system/.q-system/scripts/test_dispatch_alias_reachability.py -q
E  AssertionError: a project whose checkout is on disk right now was reported
   UNREACHABLE because its Linear name is spelled differently from its registry
   name. That line asks the operator to clone a repo he already has:
     worker: 3 ready-shaped issue(s) UNREACHABLE: no local checkout for
     Board Side Name, nolocal -- ...
E  AssertionError: an aliased project on a client-repo path has a checkout but
   repo-preflight refuses it forever, so it is neither a routine skip nor
   unreachable. The worker reported no third bucket at all.
2 failed in 0.94s
```

Green after, including the ASK-729 suite this touches:

```
$ python3 -m pytest q-system/.q-system/scripts/test_dispatch_alias_reachability.py \
                    q-system/.q-system/scripts/test_dispatch_reachability.py -q
6 passed in 2.91s
```

Live worker dry run (`--issue ASK-000000` forces an empty pool, so the run cannot
reach the claim; scratch state dir; notify stubbed):

```
worker: 0 ready issue(s) (owner:sana, has a DoR, not owner:assaf, project=kipi-system)
worker: 35 ready-shaped issue(s) REFUSED by preflight: the checkout exists but the
  gate will not let a dispatcher in -- 4_points_consulting (client-repo);
  ASK Consulting (client-repo); Alice (client-repo); Prodigy_Gold (client-repo);
  Pure_spectrum_Q (client-repo); accountant (client-repo);
  cole-GTM (control-code,hooks,remote,dirty,credentials,branch-protection);
  investigations (client-repo);
  reddit-build-radar (control-code,hooks,remote,dirty,credentials,branch-protection)
worker: 20 ready-shaped issue(s) UNREACHABLE: no local checkout for (unset),
  lane-h-digest-repeats -- ...
worker: 1 issue(s) held at needs-scope
nothing ready.
```

UNREACHABLE now names only projects with no checkout anywhere: `(unset)` (19
issues needing a human to set the project) and `lane-h-digest-repeats`. Down from
44. `ASK Consulting` and `cole-GTM` read reachable-but-refused with the gate's own
reason, which is the DoR outcome.

## Two directions pinned, not one

An UNREACHABLE bucket that fires on nothing is the same single-bucket defect
wearing a quieter word, so both reproducers carry a negative self-test: the row
with genuinely no checkout must STAY unreachable, and the aliased client repo must
NOT appear in the routine-skip line.

`test_dispatch_alias_reachability.py` runs the client-repo case against the REAL
`repo-preflight.sh` with no stub -- a test that stubs the gate it claims to consult
proves nothing about the gate. Check 0 decides on path shape and exits before any
network call, so that is honest and cheap. The ASK-729 suite gets a stub, because
its `haslocal` fixture is a bare `mkdir` the real gate correctly refuses, and it is
asserting a different claim.

## Wiring

- `daily-linear-digest.py` learns a pattern for the REFUSED line, asserted against
  real worker output rather than a hand-typed sample.
- `capability-manifest.json` registers the new test. Its `__main__` shells pytest
  on itself, because the manifest runner is `python3 <file>` and a pytest module
  without one collects nothing while exiting 0 (sp-bbdcf57b, 12 existing files).
  Not fixing those twelve here; not adding a thirteenth is free.

## Captured, not fixed

- `sp-059b0ea6`: `kipi-dispatch.sh:640` reads its rotation rows with `IFS=$'\t'`
  and carries the identical empty-field collapse. Same one-line cure, out of scope
  for a DoR about the worker report.
- `sp-123bc141` (pre-existing) stays open: the digest drops the *dispatcher's*
  client-repo refusal line. Different producer.

## Out of scope

Nothing here enables dispatch anywhere. This makes the report true.

Closes ASK-840.

### DIFF
--- a/instance-registry.json
+++ b/instance-registry.json
@@ -26,6 +26,7 @@
     },
     {
       "name": "ASK_AI_consultant",
+      "linear_project": "ASK Consulting",
       "path": "/Users/assafkipnis/projects/consulting",
       "subtree_prefix": "q-system",
       "instance_q_dir": "q-consult",
@@ -110,6 +111,7 @@
     },
     {
       "name": "gtm-partner",
+      "linear_project": "cole-GTM",
       "path": "/Users/assafkipnis/projects/cole-gtm",
       "subtree_prefix": "q-system",
       "instance_q_dir": null,
--- a/q-system/.q-system/capability-manifest.json
+++ b/q-system/.q-system/capability-manifest.json
@@ -486,6 +486,11 @@
       "runner": "python3"
     },
     {
+      "path": "q-system/.q-system/scripts/test_dispatch_alias_reachability.py",
+      "runner": "python3",
+      "timeout_s": 180
+    },
+    {
       "path": "q-system/.q-system/scripts/test_evidence_ledger.py",
       "runner": "python3"
     },
--- a/q-system/.q-system/scripts/daily-linear-digest.py
+++ b/q-system/.q-system/scripts/daily-linear-digest.py
@@ -126,6 +126,12 @@ BLOCKED_PATTERNS = (
     # stranded and why, every day, instead of the gap living in one log line
     # nobody reads.
     re.compile(r"\d+ ready-shaped issue\(s\) UNREACHABLE: no local checkout[^\n]*"),
+    # ASK-840. The third bucket: the checkout exists, and repo-preflight refuses
+    # entry anyway. It needs its own line here because the response differs from
+    # both neighbours -- a routine skip resolves on a later rotation turn, an
+    # unreachable repo needs a clone, and this one needs a cure or is permanent
+    # (a client engagement repo is refused forever, whatever its name resolves to).
+    re.compile(r"\d+ ready-shaped issue\(s\) REFUSED by preflight[^\n]*"),
     re.compile(r"\d+ repo\(s\) HELD \(cross-repo gh scoping[^\n]*"),
 )
 
--- a/q-system/.q-system/scripts/linear-worker.sh
+++ b/q-system/.q-system/scripts/linear-worker.sh
@@ -367,6 +367,21 @@ fi
 # unreadable registry means reachability is UNKNOWN, and reporting unknown as
 # unreachable would fire a loud false alarm on every project at once -- the
 # failure mode this issue exists to remove, pointed the other way.
+#
+# THE ALIAS IS A FIELD, NOT A GUESS (ASK-840). The row's `name` was read as if it
+# WERE the Linear project name. Nothing ever required those two namespaces to
+# agree, and measured against the live board on 2026-08-15 they do not: two rows
+# whose checkouts were on disk that minute carried registry names spelled
+# differently from their board projects, and 17 of the 44 issues reported
+# UNREACHABLE were on them. The log told the operator to clone repos he already
+# had. `linear_project` states the mapping instead of deriving it, because a
+# name-derivation guess is what produced the bug and a smarter guess would only
+# move the day it breaks.
+#
+# ONE DERIVATION, BOTH QUESTIONS. linear_project() below answers "what is this
+# row called on the board" for the repo-identity lookup AND for the reachability
+# set. Deriving the same logical value two ways is how the two sides drifted in
+# the first place, so there is exactly one function and both callers use it.
 REGISTRY_FACTS="$(SKEL_PATH="$TARGET_REPO" REG="$SKEL/instance-registry.json" python3 - <<'PY' 2>/dev/null
 import json, os
 skel = os.path.realpath(os.environ["SKEL_PATH"])
@@ -378,29 +393,46 @@ except Exception:
 entries = reg.get("instances", reg) if isinstance(reg, dict) else reg
 name = ""
 local = []
+
+
+def linear_project(entry):
+    """The name this row carries ON THE BOARD. Explicit field first, name second."""
+    return (entry.get("linear_project") or entry.get("name") or "").strip()
+
+
 for e in entries if isinstance(entries, list) else []:
     if not isinstance(e, dict):
         continue
-    p, n = e.get("path"), e.get("name") or ""
+    p = e.get("path")
     if not p:
         continue
+    proj = linear_project(e)
     if not name and os.path.realpath(p) == skel:
-        name = n
-    if n and os.path.isdir(p):
-        local.append(n)
-print(json.dumps({"name": name, "local_projects": sorted(set(local)), "ok": ok}))
+        name = proj
+    if proj and os.path.isdir(p):
+        # The pinned remote travels WITH the row, because repo-preflight needs it
+        # and re-reading the registry to find it is the second reader ASK-729
+        # already refused to add.
+        d = e.get("dispatch") if isinstance(e.get("dispatch"), dict) else {}
+        local.append({"project": proj, "path": p,
+                      "remote": d.get("expected_remote") or ""})
+local.sort(key=lambda r: r["project"])
+print(json.dumps({"name": name, "local_repos": local, "ok": ok}))
 PY
 )"
 _facts_get() { printf '%s' "$REGISTRY_FACTS" | python3 -c "import json,sys;d=json.load(sys.stdin);v=d.get('$1');print('\n'.join(v) if isinstance(v,list) else v)" 2>/dev/null; }
+# Structured facts travel as JSON. _facts_get flattens a list to newlines, which
+# silently mangles a list of objects into the string "[object]"-shaped nonsense.
+_facts_json() { printf '%s' "$REGISTRY_FACTS" | python3 -c "import json,sys;print(json.dumps(json.load(sys.stdin).get('$1')))" 2>/dev/null; }
 
 REPO_PROJECT="${KIPI_LINEAR_PROJECT:-}"
 [ -n "$REPO_PROJECT" ] || REPO_PROJECT="$(_facts_get name)"
 [ -n "$REPO_PROJECT" ] || REPO_PROJECT="$(basename "$TARGET_REPO")"
 export REPO_PROJECT
 
-LOCAL_PROJECTS="$(_facts_get local_projects)"
+LOCAL_REPOS="$(_facts_json local_repos)"
 REGISTRY_OK="$(_facts_get ok)"
-export LOCAL_PROJECTS REGISTRY_OK
+export LOCAL_REPOS REGISTRY_OK
 
 # --- pick ready issues ------------------------------------------------------
 PICKED="$(python3 - "$ONLY_ISSUE" <<'PY'
@@ -426,8 +458,13 @@ while True:
     after = p["pageInfo"]["endCursor"]
 
 repo_project = os.environ["REPO_PROJECT"]
-# Reachability, from the SAME registry read that produced repo_project (ASK-729).
-local_projects = {p for p in os.environ.get("LOCAL_PROJECTS", "").splitlines() if p}
+# Reachability, from the SAME registry read that produced repo_project (ASK-729),
+# now carrying the board name, the path and the pinned remote per row (ASK-840).
+try:
+    local_repos = json.loads(os.environ.get("LOCAL_REPOS") or "[]") or []
+except Exception:
+    local_repos = []
+local_by_project = {r["project"]: r for r in local_repos if r.get("project")}
 registry_ok = os.environ.get("REGISTRY_OK", "") == "True"
 
 def project_of(i):
@@ -489,18 +526,45 @@ dropped = [i for i in issues if ready_ignoring_project(i) and not in_this_repo(i
 # Reporting the second as the first is why 45 issues read as a filter doing its
 # job for 13 days. An unset project is UNREACHABLE, not skipped: "target unknown"
 # is not "target is elsewhere", and it needs a human to set the project.
+#
+# A THIRD BUCKET, BECAUSE TWO OF THEM WERE BOTH WRONG FOR THE SAME REPOS
+# (ASK-840). Resolving the alias moves 17 issues out of UNREACHABLE, and the
+# tempting next step -- drop them in with the routine skips -- swaps one false
+# sentence for another. "The rotation reaches it on a later turn" is false
+# FOREVER for a client engagement repo: repo-preflight check 0 refuses it whether
+# or not its name resolves, and no waiting changes that. So reachability answers
+# only "is there a checkout", and the preflight verdict (taken from the REAL
+# script, in the shell, never re-implemented here) splits what is left.
 def has_local_checkout(i):
     name = project_of(i)
-    return bool(name) and name in local_projects
+    return bool(name) and name in local_by_project
 
 # An unreadable registry means reachability is UNKNOWN. Calling every project
 # unreachable there would be a false alarm on the whole board at once, so the
 # classification collapses back to the old single bucket and the shell says why.
 if registry_ok:
-    skipped_local = [i for i in dropped if has_local_checkout(i)]
+    reachable = [i for i in dropped if has_local_checkout(i)]
     unreachable = [i for i in dropped if not has_local_checkout(i)]
 else:
-    skipped_local, unreachable = dropped, []
+    reachable, unreachable = dropped, []
+
+# One row per reachable project, carrying what the shell needs to run preflight
+# against it: the path, the pinned remote, and how many issues ride on the answer.
+# An empty path means "no registry row backs this" -- the unreadable-registry
+# collapse above -- and the shell treats that as the routine skip it reports
+# today, rather than calling a gate it has no target for. NO APOSTROPHE IN THIS
+# HEREDOC: it sits inside a $( ) and bash tracks quote state straight through a
+# quoted heredoc there, so one in a PYTHON COMMENT takes the whole script to
+# "unexpected EOF" 800 lines away. The file carries this scar twice already.
+def reachable_rows():
+    rows = {}
+    for i in reachable:
+        proj = project_of(i) or "(unset)"
+        row = local_by_project.get(proj) or {}
+        r = rows.setdefault(proj, {"project": proj, "path": row.get("path", ""),
+                                   "remote": row.get("remote", ""), "count": 0})
+        r["count"] += 1
+    return [rows[k] for k in sorted(rows)]
 # HELD IS A STATEMENT ABOUT OPEN WORK (ASK-841). The team fetch above is the whole
 # board, closed rows included, and this helper used to select on label + project
 # alone. A refusal label is never removed when the issue finishes -- grep confirms
@@ -548,8 +612,10 @@ print(json.dumps({
     "dropped_out_of_repo": len(dropped),
     "dropped_projects": sorted({project_of(i) or "(unset)" for i in dropped}),
     # ASK-729: the same population, split by whether anyone here can act on it.
-    "skipped_local": len(skipped_local),
-    "skipped_local_projects": sorted({project_of(i) or "(unset)" for i in skipped_local}),
+    # ASK-840: the reachable half is emitted as rows, not a count, because the
+    # shell has to ask repo-preflight about each one before it can say which of
+    # them a later rotation turn actually reaches.
+    "reachable_rows": reachable_rows(),
     "unreachable": len(unreachable),
     "unreachable_projects": sorted({project_of(i) or "(unset)" for i in unreachable}),
     "unreachable_ids": [i["identifier"] for i in unreachable],
@@ -592,8 +658,44 @@ if [ "$PROJECT_KNOWN" = "False" ]; then
   exit 9
 fi
 
-DROPPED="$(printf '%s' "$PICKED" | python3 -c 'import json,sys;print(json.load(sys.stdin).get("skipped_local",0))' 2>/dev/null)"
-DROPPED_IN="$(printf '%s' "$PICKED" | python3 -c 'import json,sys;print(", ".join(json.load(sys.stdin).get("skipped_local_projects",[])))' 2>/dev/null)"
+# THE REACHABLE HALF IS SPLIT BY THE REAL GATE, NOT BY A COPY OF ITS RULES
+# (ASK-840). A project with a checkout is not therefore dispatchable, and the
+# difference is exactly what repo-preflight.sh already decides. Re-implementing
+# its client-repo test here would be a second copy of the one refusal the founder
+# said must hold everywhere -- and the copy is what goes stale. So preflight is
+# executed, once per distinct project, and its own words become the reason.
+#
+# CHEAP FOR THE CASE THAT MATTERS: check 0 is decided on path shape and exits
+# before any network call, so every client repo costs a fork and nothing else.
+# The handful that get past it pay a few gh calls once per run, not per issue.
+PREFLIGHT="$SCRIPT_DIR/repo-preflight.sh"
+REACH_ROWS="$(printf '%s' "$PICKED" | python3 -c 'import json,sys
+for r in json.load(sys.stdin).get("reachable_rows", []):
+    print("\t".join([r.get("project",""), r.get("path",""), r.get("remote",""), str(r.get("count",0))]))' 2>/dev/null)"
+
+DROPPED=0; DROPPED_IN=""; REFUSED_N=0; REFUSED_IN=""
+while IFS="$(printf '\t')" read -r _proj _path _remote _cnt; do
+  [ -n "${_proj:-}" ] || continue
+  _cnt="${_cnt:-0}"
+  # No path means no registry row backs this project (the unreadable-registry
+  # collapse). Reporting it as today's routine skip is the honest answer: we
+  # cannot ask the gate anything without a target.
+  if [ -z "${_path:-}" ] || _pf_out="$(bash "$PREFLIGHT" "$_path" "${_remote:-}" 2>&1)"; then
+    DROPPED=$((DROPPED + _cnt))
+    DROPPED_IN="${DROPPED_IN:+$DROPPED_IN, }$_proj"
+  else
+    REFUSED_N=$((REFUSED_N + _cnt))
+    # THE CHECK NAMES, NOT THE FULL MESSAGES. One failed check carries a fix
+    # command long enough to bury the line it is on, and this is read in a daily
+    # digest. The class of refusal is what tells a permanent one (client-repo)
+    # from a curable one (control-code), which is the decision being supported.
+    _why="$(printf '%s' "$_pf_out" | grep '^FAIL' | sed 's/^FAIL \([^:]*\):.*/\1/' | tr '\n' ',' | sed 's/,$//')"
+    REFUSED_IN="${REFUSED_IN:+$REFUSED_IN; }$_proj (${_why:-refused})"
+  fi
+done <<REACHEOF
+$REACH_ROWS
+REACHEOF
+
 UNREACH="$(printf '%s' "$PICKED" | python3 -c 'import json,sys;print(json.load(sys.stdin).get("unreachable",0))' 2>/dev/null)"
 UNREACH_IN="$(printf '%s' "$PICKED" | python3 -c 'import json,sys;print(", ".join(json.load(sys.stdin).get("unreachable_projects",[])))' 2>/dev/null)"
 REG_OK="$(printf '%s' "$PICKED" | python3 -c 'import json,sys;print(json.load(sys.stdin).get("registry_ok",True))' 2>/dev/null)"
@@ -603,6 +705,13 @@ say "worker: $READY_COUNT ready issue(s) (owner:sana, has a DoR, not owner:assaf
 # Accounted for, never silently dropped. These two lines are what let an operator
 # tell a working filter from a broken query without re-querying Linear by hand.
 [ "${DROPPED:-0}" != "0" ] && say "worker: $DROPPED ready-shaped issue(s) skipped as out-of-repo (other project: $DROPPED_IN) -- this checkout is $REPO_PROJECT and cannot check those out"
+# THE THIRD BUCKET (ASK-840). Reachable, and refused anyway. Stated on its own
+# line for the same reason UNREACHABLE was: the response differs. A routine skip
+# resolves itself on a later rotation turn, an unreachable repo needs a clone, and
+# this one needs either a cure (control-code drift) or nothing at all ever
+# (client-repo, which no rotation and no cure will change). Collapsing it into the
+# skip line promises the operator a turn that is never coming.
+[ "${REFUSED_N:-0}" != "0" ] && say "worker: $REFUSED_N ready-shaped issue(s) REFUSED by preflight: the checkout exists but the gate will not let a dispatcher in -- $REFUSED_IN"
 # UNREACHABLE IS NOT A ROUTINE SKIP, AND NEVER SHARES ITS LINE (ASK-729). A skip
 # resolves itself on a later rotation turn; this does not resolve at all until a
 # human clones or registers the repo. It is stated separately, with the project
--- a/q-system/.q-system/scripts/linear-worker.sh
+++ b/q-system/.q-system/scripts/linear-worker.sh
@@ -669,12 +669,20 @@ fi
 # before any network call, so every client repo costs a fork and nothing else.
 # The handful that get past it pay a few gh calls once per run, not per issue.
 PREFLIGHT="$SCRIPT_DIR/repo-preflight.sh"
+# UNIT SEPARATOR, NOT TAB. Tab is an IFS WHITESPACE character, so bash collapses
+# runs of it however IFS is set -- and an unpinned row has an EMPTY remote, which
+# is the common case here (only two registry rows pin one). Measured with a probe
+# on a 4-field row whose third field was empty: read landed the COUNT in $_remote
+# and left $_cnt unset, so every reachable project scored 0 issues and both report
+# lines below vanished entirely. \037 is not IFS whitespace, so an empty field
+# stays an empty field. kipi-dispatch.sh:640 reads its rotation rows the same way
+# and carries the same latent hole (captured as spillover, not fixed here).
 REACH_ROWS="$(printf '%s' "$PICKED" | python3 -c 'import json,sys
 for r in json.load(sys.stdin).get("reachable_rows", []):
-    print("\t".join([r.get("project",""), r.get("path",""), r.get("remote",""), str(r.get("count",0))]))' 2>/dev/null)"
+    print("\037".join([r.get("project",""), r.get("path",""), r.get("remote",""), str(r.get("count",0))]))' 2>/dev/null)"
 
 DROPPED=0; DROPPED_IN=""; REFUSED_N=0; REFUSED_IN=""
-while IFS="$(printf '\t')" read -r _proj _path _remote _cnt; do
+while IFS=$'\037' read -r _proj _path _remote _cnt; do
   [ -n "${_proj:-}" ] || continue
   _cnt="${_cnt:-0}"
   # No path means no registry row backs this project (the unreadable-registry
--- a/q-system/.q-system/scripts/test_dispatch_alias_reachability.py
+++ b/q-system/.q-system/scripts/test_dispatch_alias_reachability.py
@@ -219,3 +219,38 @@ def test_resolved_name_does_not_promote_a_client_repo_into_the_queue(client_repo
         f"a client repo can never be reached by a later rotation turn: {skipped}")
     assert "Board Side Name" not in _line(out, "UNREACHABLE"), (
         "the checkout exists; calling it unreachable sends the operator to clone it")
+
+
+def test_digest_can_see_the_refused_line(client_repo):
+    """A third bucket nobody surfaces is the state this issue started in.
+
+    The daily digest is the one place the founder reads this, and its patterns are
+    matched against REAL worker output rather than a hand-typed sample -- a sample
+    would only prove the pattern agrees with my idea of the wording. sp-123bc141
+    already records the same gap for the DISPATCHER refusal line, which is a
+    different producer and stays captured rather than fixed here.
+    """
+    digest = SCRIPTS / "daily-linear-digest.py"
+    if not digest.exists():
+        pytest.skip("daily-linear-digest.py not present in this checkout")
+
+    import re
+    pats = re.findall(r're\.compile\(r"([^"]+)"\)', digest.read_text())
+    assert pats, "no BLOCKED_PATTERNS found in daily-linear-digest.py"
+
+    line = _line(client_repo(), "REFUSED by preflight")
+    assert line, "worker wrote no REFUSED line"
+    assert any(re.search(p, line) for p in pats), (
+        "the digest recognises no pattern for the worker's REFUSED line, so the "
+        f"third bucket stays invisible in the daily report:\n  {line}")
+
+
+if __name__ == "__main__":
+    # THE MANIFEST RUNNER IS `python3 <file>`, AND THE ALLOWED SET IS python3|bash
+    # (capability-gate.py:127). A pytest module with no __main__ collects nothing
+    # under that runner and exits 0, so the manifest reports coverage that never
+    # ran -- sp-bbdcf57b, verified across 12 declared tests. Fixing those twelve is
+    # not this issue; not adding a thirteenth is free, so this file runs itself.
+    import sys
+    raise SystemExit(subprocess.call(
+        [sys.executable, "-m", "pytest", "-q", os.path.abspath(__file__)]))
--- a/q-system/.q-system/scripts/test_dispatch_alias_reachability.py
+++ b/q-system/.q-system/scripts/test_dispatch_alias_reachability.py
@@ -1,0 +1,221 @@
+#!/usr/bin/env python3
+"""Reachability is a fact about a CHECKOUT, not a string match on two names (ASK-840).
+
+why this shape: ASK-729 split the worker's out-of-repo population into SKIPPED and
+UNREACHABLE, and answered "is there a checkout" with
+
+    project_of(issue) in {registry row name for every row whose path exists}
+
+That is a string equality between two namespaces nothing ever required to agree.
+Measured 2026-08-15 against the live board, they do not:
+
+    Linear project        registry name          path                exists
+    ASK Consulting        ASK_AI_consultant      ~/projects/...      YES
+    cole-GTM              gtm-partner            ~/projects/...      YES
+
+14 + 3 = 17 of the 44 UNREACHABLE issues were on repos cloned and registered right
+then, and the log told the operator to clone repos he already had.
+
+TWO DEFECTS, TWO TESTS, and the second is the one with teeth:
+
+  1. the alias. A row whose Linear project name differs from its registry name must
+     classify as reachable. This is the reproducer the issue names.
+  2. reachable is not dispatchable. Promoting those 17 into the routine-skip line
+     would say "the rotation reaches them later", which for a CLIENT repo is false
+     forever -- repo-preflight check 0 refuses it whatever its name resolves to. So
+     the third bucket is asserted through the REAL repo-preflight.sh, never a stub:
+     a test that stubs the gate it is claiming to consult proves nothing about the
+     gate.
+
+Both run the SHIPPED linear-worker.sh end to end against a throwaway skeleton and a
+stubbed Linear, and assert on the log the worker actually writes -- same reason as
+test_dispatch_reachability.py, whose harness this mirrors: the defect is in what the
+run SAID, and a unit test over a predicate cannot see a reporting line.
+"""
+import json
+import os
+import shutil
+import subprocess
+from pathlib import Path
+
+import pytest
+
+SCRIPTS = Path(__file__).resolve().parent
+
+# Stands in for linear-sync.py: the ONE seam that reaches the network.
+STUB_SYNC = '''
+import json, os
+def graphql(q, v):
+    if "teams(" in q:
+        return {"teams": {"nodes": [{"id": "TEAM"}]}}
+    issues = json.load(open(os.environ["FIXTURE_ISSUES"]))
+    return {"issues": {"nodes": issues,
+                       "pageInfo": {"hasNextPage": False, "endCursor": None}}}
+'''
+
+# A preflight that clears everything. Used ONLY for the rows this test wants on the
+# far side of the gate; the client-repo row is judged by the real script.
+STUB_PREFLIGHT_OK = '''#!/usr/bin/env bash
+printf 'OK %s\\n' "${1:-}"
+exit 0
+'''
+
+
+def _issue(ident, project, labels=("owner:sana",)):
+    return {
+        "id": ident,
+        "identifier": ident,
+        "title": "t " + ident,
+        "description": "## Definition of Ready\nstuff",
+        "state": {"name": "Backlog", "type": "backlog"},
+        "project": {"name": project} if project else None,
+        "labels": {"nodes": [{"name": l} for l in labels]},
+    }
+
+
+def _build(tmp_path, registry_rows, board, preflight_stub=None):
+    """A skeleton, a target repo with a real origin, a stubbed Linear, a runner."""
+    skel = tmp_path / "skel"
+    (skel / "q-system" / ".q-system").mkdir(parents=True)
+    shutil.copytree(SCRIPTS, skel / "q-system" / ".q-system" / "scripts")
+    scripts = skel / "q-system" / ".q-system" / "scripts"
+    (scripts / "linear-sync.py").write_text(STUB_SYNC)
+    if preflight_stub is not None:
+        (scripts / "repo-preflight.sh").write_text(preflight_stub)
+        (scripts / "repo-preflight.sh").chmod(0o755)
+
+    # git fetch runs before any of the reporting under test, so origin has to be
+    # real or the run exits 9 and this test measures the guard instead.
+    subprocess.run(["git", "init", "-q", "--bare", str(tmp_path / "origin.git")], check=True)
+    target = tmp_path / "target"
+    subprocess.run(["git", "init", "-q", str(target)], check=True)
+    for k, v in (("user.email", "t@t"), ("user.name", "t")):
+        subprocess.run(["git", "-C", str(target), "config", k, v], check=True)
+    (target / "f.txt").write_text("x")
+    subprocess.run(["git", "-C", str(target), "add", "-A"], check=True)
+    subprocess.run(["git", "-C", str(target), "commit", "-qm", "init"], check=True)
+    subprocess.run(["git", "-C", str(target), "remote", "add", "origin",
+                    str(tmp_path / "origin.git")], check=True)
+    subprocess.run(["git", "-C", str(target), "push", "-q", "origin", "HEAD:main"], check=True)
+
+    rows = [{"name": "targetproj", "path": str(target)}] + registry_rows
+    (skel / "instance-registry.json").write_text(json.dumps({"instances": rows}))
+
+    fixture = tmp_path / "fixture.json"
+    fixture.write_text(json.dumps(board))
+    state = tmp_path / "state"
+    state.mkdir()
+
+    def run():
+        env = dict(os.environ)
+        env.update({
+            "KIPI_SKEL": str(skel),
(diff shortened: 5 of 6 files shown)


assafkip:
Second DoR check landed (it buffers all output to the end, hence the delay):

```
$ python3 validate-separation.py 1 --verbose
  --- Gate 1.7: Plugin version drift ---
  PASS No plugin changed since origin/main without a version bump
  --- Full skeleton sweep ---
  PASS Full skeleton sweep: zero KTLYST/hardcoded refs (0 files)
  WARN Skill-hook audit: skeleton has no manifest yet (advisory)

  VALIDATION SUMMARY (Phase 1)
  PASS: 70
  FAIL: 0
  WARN: 1

ALL CHECKS PASSED. Phase 1 gate is GREEN.
[exited with code 0]
```

The single WARN is pre-existing and advisory (skeleton skill-hook manifest), unrelated to this diff.

Both checks the DoR named are now green: the worker dry run (in the PR body) and gate 1 (above).

assafkip:
## Verdict: APPROVE

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-190-20260815-023743.md` (111392 bytes).

```
FINDINGS:
END FINDINGS
```

--- reviewer output, last 54978 bytes of 111392 (full review at the path above) ---

+    rows = [
+        {"name": "registry_side_name", "linear_project": "Board Side Name",
+         "path": str(tmp_path / "aliased-dir")},
+        {"name": "nolocal", "path": str(tmp_path / "no-such-checkout")},
+    ]
+    board = [
+        # project_known guard: some issue must name this repo's project, or the run
+        # exits 9 before reporting anything.
+        _issue("ASK-900", "targetproj", ("owner:assaf",)),
+        _issue("ASK-901", "Board Side Name"),
+        _issue("ASK-902", "Board Side Name"),
+        _issue("ASK-903", "nolocal"),
+    ]
+    return _build(tmp_path, rows, board, preflight_stub=STUB_PREFLIGHT_OK)
+
+
+def test_registry_name_differing_from_linear_project_is_reachable(aliased):
+    """The reproducer ASK-840 names. RED at HEAD: no alias field is read at all."""
+    out = aliased()
+
+    unreachable = _line(out, "UNREACHABLE")
+    assert "Board Side Name" not in unreachable, (
+        "a project whose checkout is on disk right now was reported UNREACHABLE "
+        "because its Linear name is spelled differently from its registry name. "
+        f"That line asks the operator to clone a repo he already has:\n  {unreachable}")
+
+    skipped = _line(out, "skipped as out-of-repo")
+    assert "Board Side Name" in skipped, (
+        "the aliased project cleared preflight and has a local checkout, so it is a "
+        f"routine skip the rotation reaches on a later turn. Output was:\n{out}")
+
+    # NEGATIVE SELF-TEST. Deleting the UNREACHABLE bucket outright would satisfy the
+    # first assertion, so the row that genuinely has no checkout must still land in
+    # it -- otherwise this test rewards the opposite bug.
+    assert "nolocal" in unreachable, (
+        f"nolocal has no checkout anywhere and must stay UNREACHABLE: {unreachable}")
+
+
+@pytest.fixture()
+def client_repo(tmp_path):
+    """An aliased row whose path has the shape repo-preflight check 0 refuses.
+
+    NO PREFLIGHT STUB. The point of this test is that resolving the name does not
+    resolve the safety question, and the only thing that can show that is the real
+    gate. Check 0 is decided on path shape alone and exits before any network call,
+    so running it for real here is both honest and cheap.
+    """
+    client = tmp_path / "consulting" / "projects" / "someclient"
+    client.mkdir(parents=True)
+    rows = [
+        {"name": "registry_side_name", "linear_project": "Board Side Name",
+         "path": str(client)},
+        {"name": "nolocal", "path": str(tmp_path / "no-such-checkout")},
+    ]
+    board = [
+        _issue("ASK-900", "targetproj", ("owner:assaf",)),
+        _issue("ASK-901", "Board Side Name"),
+        _issue("ASK-903", "nolocal"),
+    ]
+    return _build(tmp_path, rows, board)
+
+
+def test_resolved_name_does_not_promote_a_client_repo_into_the_queue(client_repo):
+    """Reachable is not dispatchable, and the worker must say which one it means."""
+    out = client_repo()
+
+    refused = _line(out, "REFUSED by preflight")
+    assert refused, (
+        "an aliased project on a client-repo path has a checkout but repo-preflight "
+        "refuses it forever, so it is neither a routine skip nor unreachable. The "
+        f"worker reported no third bucket at all. Output was:\n{out}")
+    assert "Board Side Name" in refused, f"expected the project named: {refused}"
+    assert "client-repo" in refused, (
+        "the refusal must carry the reason the real gate gave, or the operator "
+        f"cannot tell a permanent refusal from a curable one: {refused}")
+
+    # It must NOT read as a routine skip -- that is the sentence that would be false
+    # forever, because no rotation turn will ever make a client repo dispatchable.
+    skipped = _line(out, "skipped as out-of-repo")
+    assert "Board Side Name" not in skipped, (
+        f"a client repo can never be reached by a later rotation turn: {skipped}")
+    assert "Board Side Name" not in _line(out, "UNREACHABLE"), (
+        "the checkout exists; calling it unreachable sends the operator to clone it")
+
+
+def test_digest_can_see_the_refused_line(client_repo):
+    """A third bucket nobody surfaces is the state this issue started in.
+
+    The daily digest is the one place the founder reads this, and its patterns are
+    matched against REAL worker output rather than a hand-typed sample -- a sample
+    would only prove the pattern agrees with my idea of the wording. sp-123bc141
+    already records the same gap for the DISPATCHER refusal line, which is a
+    different producer and stays captured rather than fixed here.
+    """
+    digest = SCRIPTS / "daily-linear-digest.py"
+    if not digest.exists():
+        pytest.skip("daily-linear-digest.py not present in this checkout")
+
+    import re
+    pats = re.findall(r're\.compile\(r"([^"]+)"\)', digest.read_text())
+    assert pats, "no BLOCKED_PATTERNS found in daily-linear-digest.py"
+
+    line = _line(client_repo(), "REFUSED by preflight")
+    assert line, "worker wrote no REFUSED line"
+    assert any(re.search(p, line) for p in pats), (
+        "the digest recognises no pattern for the worker's REFUSED line, so the "
+        f"third bucket stays invisible in the daily report:\n  {line}")
+
+
+if __name__ == "__main__":
+    # THE MANIFEST RUNNER IS `python3 <file>`, AND THE ALLOWED SET IS python3|bash
+    # (capability-gate.py:127). A pytest module with no __main__ collects nothing
+    # under that runner and exits 0, so the manifest reports coverage that never
+    # ran -- sp-bbdcf57b, verified across 12 declared tests. Fixing those twelve is
+    # not this issue; not adding a thirteenth is free, so this file runs itself.
+    import sys
+    raise SystemExit(subprocess.call(
+        [sys.executable, "-m", "pytest", "-q", os.path.abspath(__file__)]))
diff --git a/q-system/.q-system/scripts/test_dispatch_reachability.py b/q-system/.q-system/scripts/test_dispatch_reachability.py
index 8ee30cbc..afa472fe 100644
--- a/q-system/.q-system/scripts/test_dispatch_reachability.py
+++ b/q-system/.q-system/scripts/test_dispatch_reachability.py
@@ -70,6 +70,17 @@ def harness(tmp_path):
     (skel / "q-system" / ".q-system").mkdir(parents=True)
     shutil.copytree(SCRIPTS, skel / "q-system" / ".q-system" / "scripts")
     (skel / "q-system" / ".q-system" / "scripts" / "linear-sync.py").write_text(STUB_SYNC)
+    # PREFLIGHT STUBBED, AND ONLY HERE (ASK-840). This file asserts ONE thing: a
+    # project with a checkout is not the same fact as a project without one. Since
+    # ASK-840 the worker also asks repo-preflight whether the checkout it found may
+    # be entered, and `haslocal` below is a bare mkdir -- not a git repo, no remote
+    # -- so the real gate refuses it and it lands in the third bucket. That would
+    # make this file fail for a reason it is not about. The bucket the real gate
+    # feeds is asserted against the REAL script in
+    # test_dispatch_alias_reachability.py, which is where that claim belongs.
+    pf = skel / "q-system" / ".q-system" / "scripts" / "repo-preflight.sh"
+    pf.write_text('#!/usr/bin/env bash\nprintf "OK %s\\n" "${1:-}"\nexit 0\n')
+    pf.chmod(0o755)
 
     # git fetch runs before any of the reporting under test, so origin has to be
     # real or the run exits 9 and this test measures the guard instead.

exec
/bin/zsh -lc "ls -ld q-system; file q-system; rg -n \"repo-preflight|LOCAL_REPOS|reachable_rows|REFUSED by preflight|BLOCKED_PATTERNS\" q-system/.q-system/scripts/linear-worker.sh q-system/.q-system/scripts/repo-preflight.sh q-system/.q-system/scripts/daily-linear-digest.py; sed -n '1,240p' q-system/.q-system/scripts/repo-preflight.sh; sed -n '90,175p' q-system/.q-system/scripts/daily-linear-digest.py; python3 q-system/.q-system/scripts/test_dispatch_alias_reachability.py; python3 q-system/.q-system/scripts/test_dispatch_reachability.py" in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-190
 succeeded in 184ms:
                     ~~~~~~~~~~~~~~~~~~~^^
    code = main()
    )
    ^
    raise exception
    teardown.throw(exception)
    ~~~~~~~~~~~~~~^^^^^^^^^^^
    config = yield
             ^^^^^
    self.parse(args)
    ~~~~~~~~~~^^^^^^
    )
    ^
    raise exception
    teardown.throw(exception)
    ~~~~~~~~~~~~~~^^^^^^^^^^^
    return (yield)
            ^^^^^
    ~~~~^^^^^^^^^^^^^^
                                              ~~~~~~~~~^^^
    TemporaryFile(buffering=0),
    ~~~~~~~~~~~~~^^^^^^^^^^^^^
    dir = gettempdir()
                        ~~~~~~~~~~~^^
                            dirlist)
FileNotFoundError: [Errno 2] No usable temporary directory found in ['/var/folders/th/ty0kb7r56dvfrkgkrcbx9fv00000gn/T/', '/tmp', '/var/tmp', '/usr/tmp', '/Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-190']

codex
The PR’s new tests cannot start because this managed sandbox exposes no writable temporary directory. That is an environment constraint, not a code result. I’m switching candidate reproducers to read-only commands and shell/Python one-liners that need no files, and I will only report findings I can still execute here.
exec
/bin/zsh -lc "rg -n '\"linear_project\"|\"name\": \"ASK_AI_consultant\"|\"name\": \"gtm-partner\"|ASK Consulting|cole-GTM' . -g '*.json' -g '*.py' -g '*.sh' -g '*.md'; git show origin/main:q-system/.q-system/scripts/linear-worker.sh | sed -n '340,750p' > /dev/null; bash -n q-system/.q-system/scripts/linear-worker.sh; python3 -m py_compile q-system/.q-system/scripts/test_dispatch_alias_reachability.py q-system/.q-system/scripts/daily-linear-digest.py 2>&1 || true; git diff --check origin/main...HEAD; rg -n \"def blocked_today|latest\\[|BLOCKED_PATTERNS|enumerate\\(BLOCKED_PATTERNS\" q-system/.q-system/scripts/daily-linear-digest.py; sed -n '165,225p' q-system/.q-system/scripts/daily-linear-digest.py; sed -n '330,445p' q-system/.q-system/scripts/linear-worker.sh; sed -n '445,745p' q-system/.q-system/scripts/linear-worker.sh" in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-190
 succeeded in 0ms:
./INSTANCES.md:25:- **Purpose:** AI consulting practice (ASK Consulting / askconsulting.io)
./instance-registry.json:28:      "name": "ASK_AI_consultant",
./instance-registry.json:29:      "linear_project": "ASK Consulting",
./instance-registry.json:113:      "name": "gtm-partner",
./instance-registry.json:114:      "linear_project": "cole-GTM",
./scripts/persona-reorg.py:71:            "name": "ASK_AI_consultant",
./q-system/canonical/fleet-map.md:81:| `consulting` `[persona]` | ASK Consulting business OS, renamed. GTM `products/` moved to `cole-gtm/products/`; keeps q-consult + clients. Registry entry still named `ASK_AI_consultant` → path `consulting`. Pure/4points/Alice cascade under it |
[Errno 1] Operation not permitted: 'q-system/.q-system/scripts/__pycache__'120:BLOCKED_PATTERNS = (
143:def blocked_today(window: tuple):
179:        for i, pat in enumerate(BLOCKED_PATTERNS):
182:                latest[i] = m.group(0)
183:    return [latest[i] for i in sorted(latest)], None
        # A TIME RANGE, not a date prefix. Two clocks were in play and neither was
        # wrong alone: the logs stamp UTC, the reader lives in local time. Matching a
        # UTC prefix while labelling the message with the local date reported a
        # window spanning two local days under one local heading. Matching a local
        # prefix found nothing at all, because the log never writes that string.
        # One definition wins: the reader's own day, expressed as UTC bounds.
        stamp = line[:20]
        try:
            ts = dt.datetime.strptime(stamp, "%Y-%m-%dT%H:%M:%SZ").replace(
                tzinfo=dt.timezone.utc)
        except ValueError:
            continue
        if not (window[0] <= ts <= window[1]):
            continue
        for i, pat in enumerate(BLOCKED_PATTERNS):
            m = pat.search(line)
            if m:
                latest[i] = m.group(0)
    return [latest[i] for i in sorted(latest)], None


def _section(title, rows, error, fmt):
    if error:
        # Loud, and never mistakable for a quiet day.
        return [f"*{title}*", f"  COULD NOT READ: {error}"]
    if not rows:
        return [f"*{title}*", "  nothing"]
    return [f"*{title}*"] + [f"  {fmt(r)}" for r in rows[:15]] + (
        [f"  ...and {len(rows) - 15} more"] if len(rows) > 15 else [])


def build(now: dt.datetime):
    # MATCH THE LOG'S OWN CLOCK. `linear-worker.log` and `dispatch.log` prefix every
    # line with a UTC stamp (`date -u`), and this compared them against the LOCAL
    # date. Measured 2026-08-13 21:5x PDT: local said 2026-08-13 while the logs were
    # already writing 2026-08-14, so the "could not be worked" section matched zero
    # lines and printed "nothing" on a day full of them. A silent empty, inside the
    # tool built to stop silent empties. The header keeps the LOCAL date because that
    # is the day the reader is having.
    local_day = now.strftime("%Y-%m-%d")
    since = now.replace(hour=0, minute=0, second=0, microsecond=0)
    # The reader's day, in UTC bounds, so the filter and the heading describe the
    # same span of time.
    window = (since.astimezone(dt.timezone.utc), now.astimezone(dt.timezone.utc))
    since_iso = since.astimezone(dt.timezone.utc).strftime("%Y-%m-%dT%H:%M:%S.000Z")

    closed, opened, api_err = fetch(since_iso)
    blocked, log_err = blocked_today(window)

    lines = [f"*Linear daily* {local_day} (as of {now.strftime('%H:%M %Z')})", ""]
    lines += _section("Closed today", closed, api_err,
                      lambda n: f"{n['identifier']} {n['title'][:90]}")
    lines.append("")
    lines += _section("Opened today", opened, api_err,
                      lambda n: f"{n['identifier']} {n['title'][:90]}")
    lines.append("")
    lines += _section("Tried, could not be worked", blocked, log_err, lambda s: s)
    return "\n".join(lines), (api_err or log_err)


def send(message: str):
# in kipi-system, cannot find the files the DoR names, and exits 0 with no diff.
# Measured against the live board 2026-07-30: of 29 ready issues, 11 were the
# kipi-system project and 18 were for repos this checkout cannot check out --
# under two thirds of the worker's own queue was undispatchable by construction.
# The GraphQL query below has selected project{name} the whole time; nothing read it.
#
# DERIVED, never hardcoded: this same script runs from other skeletons, and a
# literal "kipi-system" would empty every other instance's queue while silently
# disabling the filter here. basename of the checkout is the convention the
# fleet already uses (instance dir name == Linear project name).
#
# KIPI_LINEAR_PROJECT is the override for the instance where those two names
# legitimately differ. It is also the test seam.
# basename is the LAST resort, not the first. instance-registry.json maps every
# instance path to its name, and that name IS the Linear project name -- while the
# directory very often is not:
#   <Persona>_strategy   -> .../<persona>/projects/strategy (basename: strategy)
#   <Persona>_product    -> .../<persona>/projects/product  (basename: product)
#   <Persona>_consultant -> .../consulting                  (basename: consulting)
# Three of the first three checked. Shipping basename alone would have made this
# filter reject every issue on those instances -- caught loud by the MISCONFIG
# guard below rather than as a silently empty queue, but still wrong.
#
# Order: explicit env override, then the registry, then basename.
# ONE REGISTRY READ, TWO FACTS (ASK-729). sp-421fa27d already records that the
# project-name derivation is duplicated between this script and
# spillover-promote.py. Reachability needs the SAME registry, so the fix was
# either a second reader here or one read that answers both questions. This is
# the second: it returns this repo identity AND the set of project names whose
# checkout exists on this machine, from a single parse.
#
# The path being looked UP is the target; the registry doing the looking up is
# always the skeleton. An instance carries no instance-registry.json, so
# reading it from the target would fall through to basename for every repo and
# quietly re-break the filter for exactly the three instances named below.
#
# registry_ok is carried explicitly and is NOT the same as an empty list. An
# unreadable registry means reachability is UNKNOWN, and reporting unknown as
# unreachable would fire a loud false alarm on every project at once -- the
# failure mode this issue exists to remove, pointed the other way.
#
# THE ALIAS IS A FIELD, NOT A GUESS (ASK-840). The row's `name` was read as if it
# WERE the Linear project name. Nothing ever required those two namespaces to
# agree, and measured against the live board on 2026-08-15 they do not: two rows
# whose checkouts were on disk that minute carried registry names spelled
# differently from their board projects, and 17 of the 44 issues reported
# UNREACHABLE were on them. The log told the operator to clone repos he already
# had. `linear_project` states the mapping instead of deriving it, because a
# name-derivation guess is what produced the bug and a smarter guess would only
# move the day it breaks.
#
# ONE DERIVATION, BOTH QUESTIONS. linear_project() below answers "what is this
# row called on the board" for the repo-identity lookup AND for the reachability
# set. Deriving the same logical value two ways is how the two sides drifted in
# the first place, so there is exactly one function and both callers use it.
REGISTRY_FACTS="$(SKEL_PATH="$TARGET_REPO" REG="$SKEL/instance-registry.json" python3 - <<'PY' 2>/dev/null
import json, os
skel = os.path.realpath(os.environ["SKEL_PATH"])
ok = True
try:
    reg = json.load(open(os.environ["REG"]))
except Exception:
    reg, ok = [], False
entries = reg.get("instances", reg) if isinstance(reg, dict) else reg
name = ""
local = []


def linear_project(entry):
    """The name this row carries ON THE BOARD. Explicit field first, name second."""
    return (entry.get("linear_project") or entry.get("name") or "").strip()


for e in entries if isinstance(entries, list) else []:
    if not isinstance(e, dict):
        continue
    p = e.get("path")
    if not p:
        continue
    proj = linear_project(e)
    if not name and os.path.realpath(p) == skel:
        name = proj
    if proj and os.path.isdir(p):
        # The pinned remote travels WITH the row, because repo-preflight needs it
        # and re-reading the registry to find it is the second reader ASK-729
        # already refused to add.
        d = e.get("dispatch") if isinstance(e.get("dispatch"), dict) else {}
        local.append({"project": proj, "path": p,
                      "remote": d.get("expected_remote") or ""})
local.sort(key=lambda r: r["project"])
print(json.dumps({"name": name, "local_repos": local, "ok": ok}))
PY
)"
_facts_get() { printf '%s' "$REGISTRY_FACTS" | python3 -c "import json,sys;d=json.load(sys.stdin);v=d.get('$1');print('\n'.join(v) if isinstance(v,list) else v)" 2>/dev/null; }
# Structured facts travel as JSON. _facts_get flattens a list to newlines, which
# silently mangles a list of objects into the string "[object]"-shaped nonsense.
_facts_json() { printf '%s' "$REGISTRY_FACTS" | python3 -c "import json,sys;print(json.dumps(json.load(sys.stdin).get('$1')))" 2>/dev/null; }

REPO_PROJECT="${KIPI_LINEAR_PROJECT:-}"
[ -n "$REPO_PROJECT" ] || REPO_PROJECT="$(_facts_get name)"
[ -n "$REPO_PROJECT" ] || REPO_PROJECT="$(basename "$TARGET_REPO")"
export REPO_PROJECT

LOCAL_REPOS="$(_facts_json local_repos)"
REGISTRY_OK="$(_facts_get ok)"
export LOCAL_REPOS REGISTRY_OK

# --- pick ready issues ------------------------------------------------------
PICKED="$(python3 - "$ONLY_ISSUE" <<'PY'
import importlib.util, json, os, sys, pathlib
here = pathlib.Path(os.environ["SCRIPT_DIR"])
spec = importlib.util.spec_from_file_location("ls", here / "linear-sync.py")
ls = importlib.util.module_from_spec(spec); spec.loader.exec_module(ls)
only = (sys.argv[1] or "").strip()

Q = """query($t:ID!,$a:String){issues(filter:{team:{id:{eq:$t}}},first:250,after:$a){
Q = """query($t:ID!,$a:String){issues(filter:{team:{id:{eq:$t}}},first:250,after:$a){
 nodes{id identifier title description state{name type} project{name}
       labels{nodes{name}}} pageInfo{hasNextPage endCursor}}}"""
try:
    tid = ls.graphql('query{teams(filter:{key:{eq:"ASK"}}){nodes{id}}}',{})["teams"]["nodes"][0]["id"]
except Exception as exc:
    print(json.dumps({"infra_error": str(exc)[:200]})); raise SystemExit(0)

issues, after = [], None
while True:
    p = ls.graphql(Q, {"t": tid, "a": after})["issues"]
    issues += p["nodes"]
    if not p["pageInfo"]["hasNextPage"]: break
    after = p["pageInfo"]["endCursor"]

repo_project = os.environ["REPO_PROJECT"]
# Reachability, from the SAME registry read that produced repo_project (ASK-729),
# now carrying the board name, the path and the pinned remote per row (ASK-840).
try:
    local_repos = json.loads(os.environ.get("LOCAL_REPOS") or "[]") or []
except Exception:
    local_repos = []
local_by_project = {r["project"]: r for r in local_repos if r.get("project")}
registry_ok = os.environ.get("REGISTRY_OK", "") == "True"

def project_of(i):
    return (i.get("project") or {}).get("name")

def in_this_repo(i):
    # Unset project is NOT this repo. "Target unknown" and "target is here" are
    # different claims, and treating the first as the second is how 18 foreign
    # issues got into this queue in the first place.
    return project_of(i) == repo_project

def ready(i):
    labels = {l["name"] for l in i["labels"]["nodes"]}
    if "owner:assaf" in labels:      return False   # founder decision, hands off
    if "owner:sana" not in labels:   return False
    # A REJECTION BY SANA, MADE MACHINE-READABLE (ASK-275).
    # NOTE: no apostrophes anywhere in this heredoc. It sits inside a $( )
    # command substitution, and bash tracks quote state through a quoted
    # heredoc there -- one apostrophe in a PYTHON COMMENT swallowed the rest of
    # the substitution and the whole script died at "unexpected EOF".
    # Before this label the only way to get a correctly-refused issue out of the
    # queue was to relabel it `owner:assaf` -- the FOUNDER queue -- which is how
    # ASK-149 got there on 2026-07-30. Routing an engineering re-scope to the
    # founder is the thing this loop exists to avoid, so the refusal needed a
    # label of its own that means "re-scope this", not "the founder decides".
    if "needs-scope" in labels:      return False
    # blocked:capability is the OTHER terminal refusal: the spec is fine, the
    # runner lacks something it cannot grant itself (a harness permission, a
    # missing tool). Excluded for the same reason, routed somewhere different.
    if "blocked:capability" in labels: return False
    if i["state"]["type"] not in ("backlog", "unstarted"): return False
    if not in_this_repo(i):          return False
    d = i.get("description") or ""
    return "## Definition of Ready" in d or "Definition of Ready" in d

# Everything the project filter is ABOUT to drop, counted before it drops it, so
# the run can report the shrink. A queue that silently falls from 29 to 11 is
# indistinguishable from a broken query, and "it got quiet" is the failure mode
# this filter could most easily cause.
def ready_ignoring_project(i):
    labels = {l["name"] for l in i["labels"]["nodes"]}
    return ("owner:assaf" not in labels and "owner:sana" in labels
            and "needs-scope" not in labels and "blocked:capability" not in labels
            and i["state"]["type"] in ("backlog", "unstarted")
            and "Definition of Ready" in (i.get("description") or ""))

dropped = [i for i in issues if ready_ignoring_project(i) and not in_this_repo(i)]

# TWO POPULATIONS, NOT ONE (ASK-729). Every out-of-repo issue used to be reported
# on a single "skipped as out-of-repo" line. On 2026-08-13 that line read 45 and
# it hid the distinction that decides whether anyone can do anything:
#
#   SKIPPED     the project has a checkout on this machine, just not the one this
#               run is in. The dispatcher rotation reaches it on a later turn.
#   UNREACHABLE no checkout exists for it anywhere here. No rotation, no cursor
#               and no amount of waiting reaches it. Someone must clone the repo
#               or register it before any machine can pick that work up.
#
# Reporting the second as the first is why 45 issues read as a filter doing its
# job for 13 days. An unset project is UNREACHABLE, not skipped: "target unknown"
# is not "target is elsewhere", and it needs a human to set the project.
#
# A THIRD BUCKET, BECAUSE TWO OF THEM WERE BOTH WRONG FOR THE SAME REPOS
# (ASK-840). Resolving the alias moves 17 issues out of UNREACHABLE, and the
# tempting next step -- drop them in with the routine skips -- swaps one false
# sentence for another. "The rotation reaches it on a later turn" is false
# FOREVER for a client engagement repo: repo-preflight check 0 refuses it whether
# or not its name resolves, and no waiting changes that. So reachability answers
# only "is there a checkout", and the preflight verdict (taken from the REAL
# script, in the shell, never re-implemented here) splits what is left.
def has_local_checkout(i):
    name = project_of(i)
    return bool(name) and name in local_by_project

# An unreadable registry means reachability is UNKNOWN. Calling every project
# unreachable there would be a false alarm on the whole board at once, so the
# classification collapses back to the old single bucket and the shell says why.
if registry_ok:
    reachable = [i for i in dropped if has_local_checkout(i)]
    unreachable = [i for i in dropped if not has_local_checkout(i)]
else:
    reachable, unreachable = dropped, []

# One row per reachable project, carrying what the shell needs to run preflight
# against it: the path, the pinned remote, and how many issues ride on the answer.
# An empty path means "no registry row backs this" -- the unreadable-registry
# collapse above -- and the shell treats that as the routine skip it reports
# today, rather than calling a gate it has no target for. NO APOSTROPHE IN THIS
# HEREDOC: it sits inside a $( ) and bash tracks quote state straight through a
# quoted heredoc there, so one in a PYTHON COMMENT takes the whole script to
# "unexpected EOF" 800 lines away. The file carries this scar twice already.
def reachable_rows():
    rows = {}
    for i in reachable:
        proj = project_of(i) or "(unset)"
        row = local_by_project.get(proj) or {}
        r = rows.setdefault(proj, {"project": proj, "path": row.get("path", ""),
                                   "remote": row.get("remote", ""), "count": 0})
        r["count"] += 1
    return [rows[k] for k in sorted(rows)]
# HELD IS A STATEMENT ABOUT OPEN WORK (ASK-841). The team fetch above is the whole
# board, closed rows included, and this helper used to select on label + project
# alone. A refusal label is never removed when the issue finishes -- grep confirms
# nothing in this file calls issueRemoveLabel -- so every issue that was ever
# refused stayed in these counts after it was Done, and the counts could only rise.
# Measured 2026-08-15: the run said "2 issue(s) held at blocked:capability
# (ASK-284 ASK-281)" while ASK-281 was Done. The true number was 1.
#
# That matters because the blocked:capability count is the ONLY signal that this
# loop is starving on a capability nobody granted (see its reporting site below).
# A number that never falls reports the same alarm whether or not anything is
# blocked, which is the same as reporting nothing.
#
# TERMINAL types are excluded rather than open types allowlisted, deliberately.
# An allowlist of (backlog, unstarted, started) would silently drop an issue
# parked in `triage` -- still open, still refused, still needing someone -- and a
# held issue missing from the count is invisible in a way an extra one is not.
# ASK-288 owns re-testing whether a block still applies; this is only the count.
TERMINAL_STATES = ("completed", "canceled")

def held_with(label):
    return [i for i in issues
            if label in {l["name"] for l in i["labels"]["nodes"]} and in_this_repo(i)
            and i["state"]["type"] not in TERMINAL_STATES]
deferred = held_with("needs-scope")
# Counted SEPARATELY from needs-scope. A rising blocked:capability count is a
# claim about the ENVIRONMENT, and averaging it into "held" would hide the one
# number that says the loop is starving for a capability nobody has granted.
blocked_cap = held_with("blocked:capability")

pool = [i for i in issues if ready(i)]
# An explicit --issue is a DRIVER OVERRIDE and deliberately skips the filter:
# converge.sh and an explicit kipi-work invocation both name an issue the picker
# already chose, and a rework round must be able to return to it. The job of the
# filter is choosing, not vetoing a choice already made.
# NOTE: no backticks either. Inside a $( ) a backtick opens a nested command
# substitution even here, which is the second way this heredoc has been broken.
if only:
    pool = [i for i in issues if i["identifier"] == only]

print(json.dumps({
    "ready": [{"id": i["identifier"], "title": i["title"], "project": project_of(i)}
              for i in pool],
    "total_open": len(issues),
    "dropped_out_of_repo": len(dropped),
    "dropped_projects": sorted({project_of(i) or "(unset)" for i in dropped}),
    # ASK-729: the same population, split by whether anyone here can act on it.
    # ASK-840: the reachable half is emitted as rows, not a count, because the
    # shell has to ask repo-preflight about each one before it can say which of
    # them a later rotation turn actually reaches.
    "reachable_rows": reachable_rows(),
    "unreachable": len(unreachable),
    "unreachable_projects": sorted({project_of(i) or "(unset)" for i in unreachable}),
    "unreachable_ids": [i["identifier"] for i in unreachable],
    "registry_ok": registry_ok,
    "deferred_needs_scope": len(deferred),
    "blocked_capability": len(blocked_cap),
    "blocked_capability_ids": [i["identifier"] for i in blocked_cap],
    # Does the derived repo identity name a project that EXISTS on the board?
    # If not, every issue fails in_this_repo() and the queue reads a healthy
    # empty -- a permanently silent loop reporting success. That is strictly
    # worse than the bug being fixed here, so it is called out as MISCONFIG
    # rather than allowed to look like a finished board.
    "project_known": any(project_of(i) == repo_project for i in issues),
    "repo_project": repo_project,
}))
PY
)"

INFRA="$(printf '%s' "$PICKED" | python3 -c 'import json,sys;print(json.load(sys.stdin).get("infra_error",""))' 2>/dev/null)"
if [ -n "$INFRA" ]; then
  say "INFRA: linear unreachable ($INFRA). Not counted against any issue."
  exit 0
fi

READY_COUNT="$(printf '%s' "$PICKED" | python3 -c 'import json,sys;print(len(json.load(sys.stdin)["ready"]))')"

# MISCONFIG BEFORE ANY COUNT IS BELIEVED (ASK-275). Checked ahead of the ready
# line because if the derived project matches nothing on the board, that count
# is 0 for a reason that has nothing to do with the board being empty. Reporting
# "nothing ready" there would be a loop that stopped working while claiming to
# be finished -- the one outcome worse than the queue it replaced. It PAGES and
# exits non-zero for the same reason the git-fetch guard at line 203 does: a log
# line is not surfacing, and this is environmental, not an issue's fault.
PROJECT_KNOWN="$(printf '%s' "$PICKED" | python3 -c 'import json,sys;print(json.load(sys.stdin).get("project_known"))' 2>/dev/null)"
if [ "$PROJECT_KNOWN" = "False" ]; then
  say "MISCONFIG: repo identity '$REPO_PROJECT' (from $TARGET_REPO) matches NO Linear project on team ASK."
  say "MISCONFIG: every issue would be filtered out, so this run picked nothing for a config reason, not an empty board."
  say "MISCONFIG: fix by renaming the Linear project to match the checkout, or set KIPI_LINEAR_PROJECT."
  bash "$NOTIFY" "kipi worker: repo identity '$REPO_PROJECT' matches no Linear project, so the queue reads empty and NO work can ever be picked. Do: set KIPI_LINEAR_PROJECT in the worker's environment, or rename the project to match the checkout." 2>/dev/null || true
  exit 9
fi

# THE REACHABLE HALF IS SPLIT BY THE REAL GATE, NOT BY A COPY OF ITS RULES
# (ASK-840). A project with a checkout is not therefore dispatchable, and the
# difference is exactly what repo-preflight.sh already decides. Re-implementing
# its client-repo test here would be a second copy of the one refusal the founder
# said must hold everywhere -- and the copy is what goes stale. So preflight is
# executed, once per distinct project, and its own words become the reason.
#
# CHEAP FOR THE CASE THAT MATTERS: check 0 is decided on path shape and exits
# before any network call, so every client repo costs a fork and nothing else.
# The handful that get past it pay a few gh calls once per run, not per issue.
PREFLIGHT="$SCRIPT_DIR/repo-preflight.sh"
# UNIT SEPARATOR, NOT TAB. Tab is an IFS WHITESPACE character, so bash collapses
# runs of it however IFS is set -- and an unpinned row has an EMPTY remote, which
# is the common case here (only two registry rows pin one). Measured with a probe
# on a 4-field row whose third field was empty: read landed the COUNT in $_remote
# and left $_cnt unset, so every reachable project scored 0 issues and both report
# lines below vanished entirely. \037 is not IFS whitespace, so an empty field
# stays an empty field. kipi-dispatch.sh:640 reads its rotation rows the same way
# and carries the same latent hole (captured as spillover, not fixed here).
REACH_ROWS="$(printf '%s' "$PICKED" | python3 -c 'import json,sys
for r in json.load(sys.stdin).get("reachable_rows", []):
    print("\037".join([r.get("project",""), r.get("path",""), r.get("remote",""), str(r.get("count",0))]))' 2>/dev/null)"

DROPPED=0; DROPPED_IN=""; REFUSED_N=0; REFUSED_IN=""
while IFS=$'\037' read -r _proj _path _remote _cnt; do
  [ -n "${_proj:-}" ] || continue
  _cnt="${_cnt:-0}"
  # No path means no registry row backs this project (the unreadable-registry
  # collapse). Reporting it as today's routine skip is the honest answer: we
  # cannot ask the gate anything without a target.
  if [ -z "${_path:-}" ] || _pf_out="$(bash "$PREFLIGHT" "$_path" "${_remote:-}" 2>&1)"; then
    DROPPED=$((DROPPED + _cnt))
    DROPPED_IN="${DROPPED_IN:+$DROPPED_IN, }$_proj"
  else
    REFUSED_N=$((REFUSED_N + _cnt))
    # THE CHECK NAMES, NOT THE FULL MESSAGES. One failed check carries a fix
    # command long enough to bury the line it is on, and this is read in a daily
    # digest. The class of refusal is what tells a permanent one (client-repo)
    # from a curable one (control-code), which is the decision being supported.
    _why="$(printf '%s' "$_pf_out" | grep '^FAIL' | sed 's/^FAIL \([^:]*\):.*/\1/' | tr '\n' ',' | sed 's/,$//')"
    REFUSED_IN="${REFUSED_IN:+$REFUSED_IN; }$_proj (${_why:-refused})"
  fi
done <<REACHEOF
$REACH_ROWS
REACHEOF

UNREACH="$(printf '%s' "$PICKED" | python3 -c 'import json,sys;print(json.load(sys.stdin).get("unreachable",0))' 2>/dev/null)"
UNREACH_IN="$(printf '%s' "$PICKED" | python3 -c 'import json,sys;print(", ".join(json.load(sys.stdin).get("unreachable_projects",[])))' 2>/dev/null)"
REG_OK="$(printf '%s' "$PICKED" | python3 -c 'import json,sys;print(json.load(sys.stdin).get("registry_ok",True))' 2>/dev/null)"
DEFERRED="$(printf '%s' "$PICKED" | python3 -c 'import json,sys;print(json.load(sys.stdin).get("deferred_needs_scope",0))' 2>/dev/null)"

say "worker: $READY_COUNT ready issue(s) (owner:sana, has a DoR, not owner:assaf, project=$REPO_PROJECT)"
# Accounted for, never silently dropped. These two lines are what let an operator
# tell a working filter from a broken query without re-querying Linear by hand.
[ "${DROPPED:-0}" != "0" ] && say "worker: $DROPPED ready-shaped issue(s) skipped as out-of-repo (other project: $DROPPED_IN) -- this checkout is $REPO_PROJECT and cannot check those out"
# THE THIRD BUCKET (ASK-840). Reachable, and refused anyway. Stated on its own
# line for the same reason UNREACHABLE was: the response differs. A routine skip
# resolves itself on a later rotation turn, an unreachable repo needs a clone, and
# this one needs either a cure (control-code drift) or nothing at all ever
# (client-repo, which no rotation and no cure will change). Collapsing it into the
# skip line promises the operator a turn that is never coming.
[ "${REFUSED_N:-0}" != "0" ] && say "worker: $REFUSED_N ready-shaped issue(s) REFUSED by preflight: the checkout exists but the gate will not let a dispatcher in -- $REFUSED_IN"
# UNREACHABLE IS NOT A ROUTINE SKIP, AND NEVER SHARES ITS LINE (ASK-729). A skip
# resolves itself on a later rotation turn; this does not resolve at all until a
# human clones or registers the repo. It is stated separately, with the project
# names and the count, on every run -- because the failure being fixed here is
# 13 days of a real backlog reading as normal filter output.
[ "${UNREACH:-0}" != "0" ] && say "worker: $UNREACH ready-shaped issue(s) UNREACHABLE: no local checkout for $UNREACH_IN -- no dispatcher on this machine can reach them until those repos are cloned and registered"
# Fail loud rather than classify on a guess (see registry_ok in the picker).
[ "$REG_OK" = "False" ] && say "worker: WARNING instance-registry.json could not be read, so reachability is UNKNOWN this run and every out-of-repo issue is reported as a routine skip"
[ "${DEFERRED:-0}" != "0" ] && say "worker: $DEFERRED issue(s) held at needs-scope (refused as unexecutable; the DoR drafter re-scopes them)"

BLOCKED_CAP="$(printf '%s' "$PICKED" | python3 -c 'import json,sys;print(json.load(sys.stdin).get("blocked_capability",0))' 2>/dev/null)"
BLOCKED_IDS="$(printf '%s' "$PICKED" | python3 -c 'import json,sys;print(" ".join(json.load(sys.stdin).get("blocked_capability_ids",[])))' 2>/dev/null)"
# ONE consolidated line per run, never one per issue. Six issues blocked on the
# same missing permission is ONE fact about the environment, and six pages of it
# is the cry-wolf failure founder-notifications.md names. The count is the signal:
# a queue starving on a capability nobody has granted looks identical to a quiet
# queue unless someone says the number out loud.
if [ "${BLOCKED_CAP:-0}" != "0" ]; then
  say "worker: $BLOCKED_CAP issue(s) held at blocked:capability -- the specs are sound, the runner is missing something it cannot grant itself ($BLOCKED_IDS)"
fi

if [ "$READY_COUNT" = "0" ]; then
  say "nothing ready. The DoR drafter feeds this queue; check kipi dor."

exec
import re, pathlib
p=pathlib.Path('q-system/.q-system/scripts/daily-linear-digest.py')
s=p.read_text()
pats=re.findall(r're\\.compile\\(r\"(["'^"]+)"'"\\)', s)
line='worker: 2 ready-shaped issue(s) REFUSED by preflight: the checkout exists but the gate will not let a dispatcher in -- Board Side Name (client-repo)'
for i,p in enumerate(pats):
    if re.search(p,line): print(i,p)
print('patterns',len(pats))
PY
python3 - <<'PY'
import json, pathlib
r=json.loads(pathlib.Path('instance-registry.json').read_text())['instances']
from collections import Counter
vals=[(e.get('linear_project') or e.get('name') or '').strip() for e in r]
print('rows',len(vals),'duplicates',[x for x,n in Counter(vals).items() if n>1])
print('missing paths',sum(not pathlib.Path(e['path']).is_dir() for e in r if e.get('path')))
print('aliases',[(e['name'],e.get('linear_project')) for e in r if e.get('linear_project')])
PY
rg -n '"'^FAIL|refuse '"' q-system/.q-system/scripts/repo-preflight.sh; rg -n 'gh ' q-system/.q-system/scripts/repo-preflight.sh" in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-190
 succeeded in 0ms:
zsh:11: can't create temp file for here document: operation not permitted
49:FAILED=0
170:  refuse "control-code" "$REPO_PATH/$WORKER_REL is missing, the repo carries no worker to run"
172:  refuse "control-code" "the skeleton reference $SKELETON/$WORKER_REL is missing, so drift cannot be judged"
177:    refuse "control-code" "could not hash the worker on both sides, so drift is unknown"
187:    refuse "control-code" "linear-worker.sh differs from the skeleton (${have:0:12} vs ${want:0:12}); fix with: python3 $SKELETON/q-system/.q-system/scripts/control-file-propagate.py --target $REPO_PATH --file $WORKER_REL --apply"
227:  refuse "hooks" "$REPO_PATH/$SETTINGS_REL is missing, so no hook fires in this repo"
322:    UNREADABLE:*) refuse "hooks" "settings.json could not be parsed (${MISSING#UNREADABLE:})" ;;
324:    *)            refuse "hooks" "hook parity with the skeleton is incomplete: $MISSING" ;;
336:  refuse "remote" "the registry row pins no expected_remote; an unpinned repo is never entered"
340:    refuse "remote" "no origin remote is configured in $REPO_PATH"
342:    refuse "remote" "origin is $ACTUAL_REMOTE but the registry pins $EXPECTED_REMOTE"
351:  refuse "dirty" "$REPO_PATH is not a git repository"
355:    refuse "dirty" "working tree has $(printf '%s\n' "$PORCELAIN" | grep -c .) uncommitted change(s); an agent branch would capture them"
378:  refuse "credentials" "gh is not on PATH, so no push or auto-merge could succeed"
380:  refuse "credentials" "gh auth status failed; the token is missing or expired"
382:  refuse "credentials" "could not derive an owner/repo slug from '$EXPECTED_REMOTE'"
384:  refuse "credentials" "the current gh token cannot see $SLUG"
398:  refuse "branch-protection" "no slug to query (see the remote check)"
400:  refuse "branch-protection" "gh is not on PATH, so protection cannot be confirmed"
404:    refuse "branch-protection" "could not read the default branch for $SLUG"
408:      refuse "branch-protection" "$SLUG@$DEFAULT_BRANCH reports no branch protection; auto-merge would land unreviewed code"
469:          refuse "branch-protection" "the protection response for $SLUG@$DEFAULT_BRANCH could not be parsed" ;;
475:          refuse "branch-protection" "GitHub refused the protection query for $SLUG@$DEFAULT_BRANCH: ${PROT_GATES#APIERROR:}" ;;
477:          refuse "branch-protection" "$SLUG@$DEFAULT_BRANCH is protected but requires NO review and NO status check; auto-merge would still land code with nothing in its way" ;;
374:# gh is what pushes the branch and arms auto-merge. Discovering the token is dead
377:if ! command -v gh >/dev/null 2>&1; then
378:  refuse "credentials" "gh is not on PATH, so no push or auto-merge could succeed"
379:elif ! gh auth status >/dev/null 2>&1; then
380:  refuse "credentials" "gh auth status failed; the token is missing or expired"
383:elif ! gh repo view "$SLUG" >/dev/null 2>&1; then
384:  refuse "credentials" "the current gh token cannot see $SLUG"
399:elif ! command -v gh >/dev/null 2>&1; then
400:  refuse "branch-protection" "gh is not on PATH, so protection cannot be confirmed"
402:  DEFAULT_BRANCH="$(gh api "repos/$SLUG" --jq .default_branch 2>/dev/null | tr -d '[:space:]')"
406:    PROT_JSON="$(gh api "repos/$SLUG/branches/$DEFAULT_BRANCH/protection" 2>/dev/null)"
419:      # travels in the environment instead of through another layer of quotes.
440:# and bash tracks quotes while scanning for the closing paren even though the
446:# The gh CLI prints the GitHub JSON error to STDOUT and its own message to

exec
/bin/zsh -lc "python3 -c 'import re,pathlib; s=pathlib.Path(\"q-system/.q-system/scripts/daily-linear-digest.py\").read_text(); ps=re.findall(r\"re\\\\.compile\\\\(r\\\"(["'^'"\\\"]+)\\\"\\\\)\",s); line=\"worker: 2 ready-shaped issue(s) REFUSED by preflight: the checkout exists but the gate will not let a dispatcher in -- Board Side Name (client-repo)\"; print([(i,p) for i,p in enumerate(ps) if re.search(p,line)]); print(\"patterns\",len(ps))'; python3 -c 'import json,pathlib,collections; r=json.loads(pathlib.Path(\"instance-registry.json\").read_text())[\"instances\"]; vs=[(e.get(\"linear_project\") or e.get(\"name\") or \"\").strip() for e in r]; print(\"rows\",len(vs),\"duplicates\",[x for x,n in collections.Counter(vs).items() if n>1]); print(\"missing paths\",sum(not pathlib.Path(e[\"path\"]).is_dir() for e in r if e.get(\"path\"))); print(\"aliases\",[(e[\"name\"],e.get(\"linear_project\")) for e in r if e.get(\"linear_project\")])'" in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-190
 succeeded in 0ms:
[]
patterns 0
rows 25 duplicates []
missing paths 0
aliases [('ASK_AI_consultant', 'ASK Consulting'), ('gtm-partner', 'cole-GTM')]

exec
/bin/zsh -lc "python3 -c 'import json; r=json.load(open(\"instance-registry.json\"))[\"instances\"]; print(\"\\n\".join(f\"{e[chr(110)+chr(97)+chr(109)+chr(101)]}|{e.get(chr(108)+chr(105)+chr(110)+chr(101)+chr(97)+chr(114)+chr(95)+chr(112)+chr(114)+chr(111)+chr(106)+chr(101)+chr(99)+chr(116),chr(45))}|{(e.get(chr(100)+chr(105)+chr(115)+chr(112)+chr(97)+chr(116)+chr(99)+chr(104)) or {}).get(chr(101)+chr(120)+chr(112)+chr(101)+chr(99)+chr(116)+chr(101)+chr(100)+chr(95)+chr(114)+chr(101)+chr(109)+chr(111)+chr(116)+chr(101),chr(60)+chr(101)+chr(109)+chr(112)+chr(116)+chr(121)+chr(62))}\" for e in r))' | sed -n '1,30p'; p="'$(python3 -c '"'import json; r=json.load(open(\"instance-registry.json\"))[\"instances\"]; print(next(e[\"path\"] for e in r if not (e.get(\"dispatch\") or {}).get(\"expected_remote\") and \"/consulting\" not in e[\"path\"] and \"/intel\" not in e[\"path\"]))'); printf 'path=%s\\n' \""'$p"; bash q-system/.q-system/scripts/repo-preflight.sh "$p" "" 2>&1 | sed -n '"'/"'^FAIL/p;/''^REFUSED/p'"'; printf 'exit=%s\\n' \""'${pipestatus[1]}"' in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-190
 succeeded in 79ms:
FAIL control-code: linear-worker.sh differs from the skeleton (dc3b9a4acb37 vs 02e611e55533); fix with: python3 /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-190/q-system/.q-system/scripts/control-file-propagate.py --target /Users/assafkipnis/projects/cole-gtm/projects/strategy --file q-system/.q-system/scripts/linear-worker.sh --apply
FAIL remote: the registry row pins no expected_remote; an unpinned repo is never entered
FAIL dirty: working tree has 5 uncommitted change(s); an agent branch would capture them
FAIL credentials: gh auth status failed; the token is missing or expired
FAIL branch-protection: no slug to query (see the remote check)
REFUSED /Users/assafkipnis/projects/cole-gtm/projects/strategy
exit=1

codex
No finding survived reproduction.

What is sound:

- Alias mapping is explicit and consistent. Both registry identity and reachability use `linear_project()`.
- Real registry shape matches the implementation: 25 rows, no duplicate resolved project names, and both aliases point to existing checkouts.
- Empty `expected_remote` fields survive the unit-separator serialization path.
- Client repositories are classified through the real `repo-preflight.sh`, not duplicated logic.
- Digest regex matches the worker’s emitted `REFUSED by preflight` line.
- `bash -n q-system/.q-system/scripts/linear-worker.sh` passed.
- `git diff --check origin/main...HEAD` passed.
- The real preflight refusal path was exercised against a registered checkout and exited `1` with concrete reasons:
  ```text
  FAIL control-code: linear-worker.sh differs from the skeleton (...)
  FAIL remote: the registry row pins no expected_remote; an unpinned repo is never entered
  FAIL dirty: working tree has 5 uncommitted change(s); an agent branch would capture them
  FAIL credentials: gh auth status failed; the token is missing or expired
  FAIL branch-protection: no slug to query (see the remote check)
  REFUSED /Users/assafkipnis/projects/cole-gtm/projects/strategy
  exit=1
  ```
- GitHub inspection and pytest execution were inconclusive because the sandbox had no network connection and no writable temporary directory. I did not convert either environment limitation into a code finding.

**VERDICT: APPROVE**

Most important thing to fix first: none. No blocker, major, minor, or nit survived an executed read-only reproducer.

FINDINGS:
END FINDINGS
tokens used
61,825
No finding survived reproduction.

What is sound:

- Alias mapping is explicit and consistent. Both registry identity and reachability use `linear_project()`.
- Real registry shape matches the implementation: 25 rows, no duplicate resolved project names, and both aliases point to existing checkouts.
- Empty `expected_remote` fields survive the unit-separator serialization path.
- Client repositories are classified through the real `repo-preflight.sh`, not duplicated logic.
- Digest regex matches the worker’s emitted `REFUSED by preflight` line.
- `bash -n q-system/.q-system/scripts/linear-worker.sh` passed.
- `git diff --check origin/main...HEAD` passed.
- The real preflight refusal path was exercised against a registered checkout and exited `1` with concrete reasons:
  ```text
  FAIL control-code: linear-worker.sh differs from the skeleton (...)
  FAIL remote: the registry row pins no expected_remote; an unpinned repo is never entered
  FAIL dirty: working tree has 5 uncommitted change(s); an agent branch would capture them
  FAIL credentials: gh auth status failed; the token is missing or expired
  FAIL branch-protection: no slug to query (see the remote check)
  REFUSED /Users/assafkipnis/projects/cole-gtm/projects/strategy
  exit=1
  ```
- GitHub inspection and pytest execution were inconclusive because the sandbox had no network connection and no writable temporary directory. I did not convert either environment limitation into a code finding.

**VERDICT: APPROVE**

Most important thing to fix first: none. No blocker, major, minor, or nit survived an executed read-only reproducer.

FINDINGS:
END FINDINGS

PR #193, opened 2026-08-15
A disposition value is vocabulary, not a deferral (ASK-734)

assafkip:
Closes ASK-734.

`fable-discipline-lint` blocked on `out-of-scope`, which is also one of prd-os's own finding **disposition values**. Any file handling dispositions tripped the lint on its own vocabulary — and because the hook is PostToolUse, it fired on the *next* edit to a file, not the edit that wrote the text.

## The fix, in one sentence

Every other phrase in the deferral set is a **verb** ("fix later", "defer this", "punt on") — a speech act that can only mean "I am not doing this now". `out-of-scope` is the one **noun**, and it is this repo's domain vocabulary. The verbs still fire alone; the noun now needs a forward-looking cue beside it.

## Measured across the whole population

All 47 code files under `plugins/prd-os`, pre-fix blob vs post-fix:

| file | pre | post |
|---|---|---|
| `scripts/judgment_compiler.py` | 10 | 0 |
| `tests/test_judgment_compiler.py` | 3 | 0 |
| `tests/test_spillover.py` | 1 | 0 |
| **TOTAL** | **14** | **0** |

Each of the 14 was read by hand: disposition literals, dict keys, a usage string listing allowed values, prose describing the mechanism. None was a deferral, so 0 is the correct floor.

## The true positive still blocks

`# out-of-scope, handle later` → exit 2. Also "out of scope for now, TODO", "fix later", "punting on this", "defer this to v2", "won't fix now", "out-of-scope, revisit next pass".

## Non-vacuous self-test

Four cases added, run against the **pre-fix blob from origin/main**:

- pre-fix: `3 FAILED` (the three vocabulary cases, exit 2 want 0)
- post-fix: `all 35 cases passed`
- `plugins/prd-os/tests`: 636 passed, 1 skipped

The true-positive case passes in both — that is its job, it guards against this fix going permissive.

## A narrower fix that I measured and rejected

First attempt neutralised only quoted/backticked mentions: 14 → 4. Reading the 4 showed all were still false positives (a `one of: ...` help string, prose about a regex, a docstring naming the class). Two mechanisms for one symptom and still wrong, so it was replaced by the single use-vs-mention rule rather than patched again.

## Trade-off, stated not hidden

A bare `# out of scope` with no promise attached no longer fires. In a repo whose vocabulary contains those words, the noun alone was never evidence. The standing spillover gate is the enforcement.

## Captured, not fixed here

`sp-100164dd` — `export-fable-mirror.sh --check` reports DRIFT on 3 files. Pre-existing on `origin/main` at 143fd40d with identical output, so outside this issue's allowed files.

### DIFF
--- a/plugins/prd-os/.claude-plugin/plugin.json
+++ b/plugins/prd-os/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "prd-os",
-  "version": "0.27.2",
+  "version": "0.27.3",
   "description": "PRD operating system: capture rough ideas, draft reviewable PRDs, run standard and adversarial review (Codex or a Claude senior-staff-engineer subagent, recorded truthfully), triage findings, decompose approved PRDs into atomic issue specs, and execute those issues with scope enforcement, stop-gate receipts, and reviewer-attributed findings.",
   "author": {
     "name": "Assaf Kipnis"
--- a/plugins/prd-os/skills/fable-discipline/scripts/fable-discipline-lint.py
+++ b/plugins/prd-os/skills/fable-discipline/scripts/fable-discipline-lint.py
@@ -134,11 +134,60 @@ SKIP_MARKER = "fable-discipline-lint-skip"
 SPILL_SKIP_MARKER = "spillover-skip"
 _CODE_SUFFIXES = {".py", ".js", ".ts", ".jsx", ".tsx", ".sh", ".rb", ".go",
                   ".rs", ".java", ".c", ".cpp", ".h", ".mjs", ".cjs"}
-_DEFERRAL = re.compile(
-    r"out[- ]of[- ]scope|fix (?:it )?later|defer(?:red)? this|"
+# THE DETECTOR SHARED A WORD WITH THE VOCABULARY IT POLICES (ASK-734).
+#
+# `out-of-scope` and `scope-removed` are two of prd-os's own finding DISPOSITION
+# VALUES. Every file that handles dispositions -- the runner, the findings
+# writer, above all their tests -- carries the token as DATA, and this lint read
+# each one as a deferral. Measured 2026-08-13 on
+# plugins/prd-os/tests/test_judgment_compiler.py: 3 hits, all pre-existing in the
+# committed blob, none a deferral.
+#
+# Worse than a plain false positive, because this hook is PostToolUse: it fires
+# on the NEXT edit to a file, not the edit that introduced the text. An unrelated
+# one-line fix is refused over lines the author never wrote, and the cheapest way
+# out is a blanket `# spillover-skip`, which disarms the lint for the WHOLE file
+# including real deferrals. A gate whose cheapest fix is to switch it off will be
+# switched off.
+#
+# THE SPLIT: every other phrase here is a VERB -- a speech act that can only mean
+# "I am not doing this now". `out-of-scope` is the one NOUN, and it is this
+# repo's domain vocabulary, which is exactly why it was the only alternative
+# generating false positives. So the verbs still fire alone, and the noun now
+# needs a forward-looking CUE next to it -- the promise is what makes a deferral
+# a deferral. `# out-of-scope, handle later` fires; `one of: scope-removed,
+# out-of-scope, defer` does not.
+#
+# MEASURED, not assumed. Across the 47 code files under plugins/prd-os this took
+# the deferral count 14 -> 0, and each of the 14 was read by hand: all 14 were
+# vocabulary or prose describing the mechanism, none was a deferral. The cue
+# list deliberately omits the bare word `defer` -- it is itself a disposition
+# value and sits in those same enumerations; `defer this` keeps its own verb
+# alternative below.
+#
+# WHAT THIS GIVES UP, stated rather than hidden: a bare `# out of scope` with no
+# promise attached no longer fires. In a repo whose vocabulary contains the
+# words, the noun alone was never evidence. This lint is a write-time nudge and
+# already documents that it misses synonym phrasings; the standing spillover gate
+# is the enforcement.
+_DEFER_CUE = re.compile(
+    r"later|for now|to-?do|fixme|revisit|postpone|someday|another day|"
+    r"come back|not (?:doing|now)|future work|next (?:pass|round|time)",
+    re.IGNORECASE,
+)
+_DEFERRAL_VERB = re.compile(
+    r"fix (?:it )?later|defer(?:red)? this|"
     r"leave (?:this )?for later|won'?t fix(?: now)?|punt(?:ed|ing)? on",
     re.IGNORECASE,
 )
+_SCOPE_NOUN = re.compile(r"out[- ]of[- ]scope", re.IGNORECASE)
+
+
+def is_deferral_line(line):
+    """True when the line DEFERS work, not when it merely names a disposition."""
+    if _DEFERRAL_VERB.search(line):
+        return True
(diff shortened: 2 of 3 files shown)


assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-193-20260815-144820.md` (6984 bytes).

```
FINDINGS:
minor|Module docstring line 48 still lists "out of scope"/"out-of-scope" as unconditionally caught and MISSES omits the new cue requirement, so the file's own enumerated-coverage contract misstates the detector|plugins/prd-os/skills/fable-discipline/scripts/fable-discipline-lint.py:48
minor|A deferral whose promise sits on the adjacent line, or phrased "out of scope for this PR", no longer blocks, which is wider than the trade-off the change states|plugins/prd-os/skills/fable-discipline/scripts/fable-discipline-lint.py:186
END FINDINGS
```

--- reviewer output ---

## Review: PR #193 — "A disposition value is vocabulary, not a deferral (ASK-734)"

Read-only. Nothing in the repo was modified.

---

### Finding 1 — minor

**Claim:** The module's enumerated-coverage docstring still advertises `"out of scope"/"out-of-scope"` as unconditionally CAUGHT, and its MISSES list never mentions the new cue requirement — the file's own self-declared coverage contract now misstates what the detector does.

**Location:** `plugins/prd-os/skills/fable-discipline/scripts/fable-discipline-lint.py:48` (CATCHES), `:54-58` (MISSES)

The file opens by declaring `Detector coverage (enumerated on purpose, per the hook-blind-spots rule)`. That enumeration was not updated with the code, and it is one of the five files `export-fable-mirror.sh` publishes to the public mirror.

**Reproducer:**

```
grep -n "out of scope\|out-of-scope" plugins/prd-os/skills/fable-discipline/scripts/fable-discipline-lint.py
```

**Real output (excerpt):**

```
48:             "out of scope"/"out-of-scope", "fix (it) later", "defer(red) this",
158:# a deferral. `# out-of-scope, handle later` fires; `one of: scope-removed,
168:# WHAT THIS GIVES UP, stated rather than hidden: a bare `# out of scope` with no
```

Line 48 (the contract) and line 158 (the implementation comment) disagree. Code is the truth; line 48 is the finding.

---

### Finding 2 — minor

**Claim:** A deferral whose promise sits on the *adjacent* line, or that is phrased "out of scope for this PR/issue", no longer blocks — shapes wider than the trade-off the PR states ("a bare `# out of scope` with **no promise attached**").

**Location:** `plugins/prd-os/skills/fable-discipline/scripts/fable-discipline-lint.py:186-190` (`is_deferral_line`, per-line `and`)

**Reproducer** (ran the new lint and the pre-fix blob from `143fd40d` over the same three files):

```
python3 - <<'PYEOF'
... probe.py written per case, run against HEAD lint and 143fd40d lint ...
PYEOF
```

**Real output:**

```
two-line comment: noun line 1, promise line 2
   pre-fix exit=2   post-fix exit=0
one line, scoped-to-this-PR phrasing
   pre-fix exit=2   post-fix exit=0
one line, noun + cue (control, must still block)
   pre-fix exit=2   post-fix exit=2
```

Case 1 (`# out-of-scope for this issue` / `# will handle it later`) carries an explicit promise and still passes. Case 2 (`out of scope for this PR`) is a scoping-away with no capture.

**Severity honesty:** both inputs are constructed. I swept every tracked code file in the repo and **no producer emits either shape today** (post-fix repo-wide hit count is 0, and none of the 48 pre-fix hits was a genuine uncaptured deferral). Blast radius is a weakened write-time nudge behind the standing spillover gate, which `.claude/rules/no-orphan-findings.md` already names as the enforcement of last resort. Nothing permanent, nothing that pages anyone. Dropped a level and said so.

---

### What is sound — attacks I ran that the code survived

**Widened the measured population past what the PR measured.** The PR measures 47 files under `plugins/prd-os`. The hook runs on every code file in the repo, and ships fleet-wide. I ran both detectors over every tracked file:

```
TOTAL repo-wide hits with PRE-FIX detector on HEAD tree: 48
TOTAL repo-wide hits at HEAD: 0
```

I then printed and read all 34 hits **outside** prd-os. Every one is scope vocabulary or design prose (`0 = clean (or out of scope)`, `Exits 2 (block) when out of scope`, `OCR is out of scope: if a page has no text layer`). The one that is a real deferral — `linear-dor-drafter.py:182`, `out of scope by founder deferral (sp-626e9452)` — is already captured. So the PR's "0 is the correct floor" conclusion holds on a population 3x larger than the one it measured.

**Non-vacuity of the new cases.** Ran them against the pre-fix blob via the suite's own ref hatch:

```
FABLE_LINT_UNDER_TEST=<143fd40d blob> python3 test_fable_discipline_lint.py
  → fable-discipline-lint self-test: 3 FAILED

python3 test_fable_discipline_lint.py
  → fable-discipline-lint self-test: all 35 cases passed
```

The true-positive case is green in both, which is its job.

**Mutation-tested the new branch.** `and` → `or` flips `disposition_list.py` to exit 2 (case fails). Dropping the noun path entirely flips `deferred_scope_noun_with_cue.py` to exit 0 (case fails). Both halves of `is_deferral_line` are pinned by a case that would actually go red. Also checked the true positive exercises the *new* path and not the old verb list: `handle later` matches no alternative in `_DEFERRAL_VERB`.

**Self-trip.** Both edited files contain the literal `spillover-skip`, so `find_deferral_lines` no-ops on them — the fix does not arm the detector against its own explanatory comments. Verified: lint on itself exit 0; on `judgment_compiler.py` exit 0; on `test_judgment_compiler.py` exit 0. The originally-blocked files are actually unblocked.

**The mirror-drift spillover claim.** `export-fable-mirror.sh --check` reports DRIFT on all three managed files at HEAD. I compared the mirror clone against the `143fd40d` blobs directly (with the kipi-only strip applied to SKILL.md): `main_drift=True` on all three. `sp-100164dd` is genuinely pre-existing, not manufactured by this PR.

**Single copy, wired.** One `fable-discipline-lint.py` in the tree; `plugins/prd-os/hooks/hooks.json:9` invokes exactly that path. No stale sibling copy of the old regex anywhere in the repo.

**Suite:** `python3 -m pytest plugins/prd-os/tests -q` → `636 passed, 1 skipped in 51.62s`.

**Attacks that found nothing:** false positives newly introduced by the broad `_DEFER_CUE` list (requires noun + cue on one line; repo-wide count is 0); concurrency (hook is stateless, reads one file, writes nothing); rollback (pure regex change, revertible); 3am blast radius (a lint that exits 0 more often cannot page or destroy anything).

---

Two minor findings, no blocker, no major. The fix does what it claims, on a wider population than it measured, with tests I watched go red against the pre-fix blob.

**Fix first:** `fable-discipline-lint.py:48` — move `"out of scope"/"out-of-scope"` out of CATCHES into a cue-conditional line, and add the adjacent-line split (Finding 2) to the MISSES enumeration. The file declares its coverage list as a contract; right now that contract is the one thing in the change that is not true.

FINDINGS:
minor|Module docstring line 48 still lists "out of scope"/"out-of-scope" as unconditionally caught and MISSES omits the new cue requirement, so the file's own enumerated-coverage contract misstates the detector|plugins/prd-os/skills/fable-discipline/scripts/fable-discipline-lint.py:48
minor|A deferral whose promise sits on the adjacent line, or phrased "out of scope for this PR", no longer blocks, which is wider than the trade-off the change states|plugins/prd-os/skills/fable-discipline/scripts/fable-discipline-lint.py:186
END FINDINGS

PR #195, opened 2026-08-16
One scan, at the only stage that can see both facts (ASK-747)

assafkip:
Closes ASK-747. Implements the amended DoR. Supersedes #194 (closed unmerged).

`client-name-guard` had **two holes of one shape**: content was being read as consent.

- `commit-msg --message` tested the skip token as a **substring**, so a message that merely *named* it disarmed the guard. Its own introducing commit proved this and went unscanned.
- `pre-commit --staged` tested that same token against its own **staged diff**, so any commit touching a file whose content contains the token went unscanned. The file that defines the token is the guard itself.

Fixing only the message half leaves the second one open.

## The fix is a stage move, not a smarter test

One scan, at `commit-msg` — the only stage where both facts hold at once:

- the real message for **this** commit arrives as an argument (no persistence, no cross-commit leak, and present for `git commit -m` exactly as for an editor commit);
- `git diff --cached` still returns staged content, because the commit object does not exist yet.

One invocation scans **both** surfaces. The bypass is decided only by the message, as a trailer anchored at column 0.

`--staged` is removed rather than left as a scan nobody calls. `lefthook.yml` was its only caller.

## Evidence: real `git commit`, no env vars

Every case drives a real `git commit -m` through a real `.git/hooks/commit-msg` calling the guard with `"$1"`, exactly as lefthook does. **No case sets an environment variable.** That is deliberate: #194's suite went green because one case set `LEFTHOOK_COMMIT_MSG_FILE`, which nothing in this repo sets, and that is exactly how its blocker hid.

```
this branch:            all 9 cases passed
pre-fix (origin/main):  5 FAILED
```

The five that fail against the old guard are the measurement:

- staged content still BLOCKS when the message is clean
- prose mentioning the token is still SCANNED and BLOCKS
- a bypass in commit N does NOT authorise commit N+1
- a client name planted in `client-name-guard.py` itself BLOCKS
- an INDENTED quote of the trailer does not bypass

## The cross-commit leak, shown not claimed

Commit 1 has clean content and a message that happens to carry a bypass trailer. Commit 2 stages a client name with a clean message and no bypass of its own.

| | commit 1 | commit 2 | client name in history |
|---|---|---|---|
| **PR #194 wiring** | rc=0 | rc=0 | **YES — commit `ad0d133`** |
| **this branch** | rc=0 | rc=1 | no |

Under #194 the name reached a real commit object.

My first attempt at this reproducer was wrong and I corrected it: I had commit 1 carrying the client name, which #194 blocks for an unrelated reason, so the sequence never reached the leak. The faithful path is a *clean* commit whose message poisons `COMMIT_EDITMSG` for the next one.

## Also fixed, from the #194 review

- The trailer is anchored at column 0, so an indented quote of the guard's own usage docstring is not an invocation.
- The claim that git strips `#` lines is **gone**. `git commit -m` does not strip them, so a rule resting on that would be wrong. Column-0 anchoring covers both cases without needing the claim.

## Wiring

- `lefthook validate` → `All good`
- `test-repo-preflight.sh` → 70 passed, 0 failed
- Declared in `capability-manifest.json` so the gate runs it.
- **This PR's own commit was made through the new wiring, with no `--no-verify`.** The pre-commit stage list no longer contains `client-names`, and `client-names-msg` scanned a message that names the token plus a diff containing the guard source.

### DIFF
--- a/lefthook.yml
+++ b/lefthook.yml
@@ -36,16 +36,6 @@ pre-commit:
     receipts-ledger:
       run: python3 q-system/.q-system/scripts/receipts-ledger-check.py
 
-    # Client names are not secrets in the credential sense, so gitleaks passes
-    # them and nothing else looked. This repo is PUBLIC (102 stars, 23 forks) and
-    # a name that reaches it cannot be recalled: rewriting history leaves the
-    # objects in every fork and reachable by SHA. Measured 2026-08-13, the list
-    # carries 0 false positives across 59 real posts and 133 lessons. The token
-    # list is LOCAL (~/.config/kipi/client-tokens), never ships, and an absent
-    # list warns and passes so a fresh clone is never blocked.
-    client-names:
-      run: python3 q-system/.q-system/scripts/client-name-guard.py --staged
-
     blocked-paths:
       run: |
         # Block .env, .envrc, .env.<env>, *.pem/*.key (incl. rotations like key.pem.old), *.jsonl,
@@ -106,6 +96,21 @@ commit-msg:
 
     # Both client names that reached this public repo did so in commit SUBJECTS,
     # which no content gate reads. The staged-content half lives in pre-commit.
+    # Client names are not secrets in the credential sense, so gitleaks passes
+    # them and nothing else looked. This repo is PUBLIC (102 stars, 23 forks) and
+    # a name that reaches it cannot be recalled: rewriting history leaves the
+    # objects in every fork and reachable by SHA. Measured 2026-08-13, the list
+    # carries 0 false positives across 59 real posts and 133 lessons. The token
+    # list is LOCAL (~/.config/kipi/client-tokens), never ships, and an absent
+    # list warns and passes so a fresh clone is never blocked.
+    #
+    # THE ONE client-name scan (ASK-747). It lives here, not at pre-commit,
+    # because this is the only stage that sees the real message for THIS commit
+    # AND can still read `git diff --cached` (the commit object does not exist
+    # yet). It scans BOTH. The pre-commit `--staged` stage was removed: it could
+    # not see the message, so it could not honour a legitimate bypass, and it
+    # tested the skip token against its own staged diff -- any commit touching a
+    # file containing the token went unscanned.
     client-names-msg:
       run: python3 q-system/.q-system/scripts/client-name-guard.py --message {1}
 
--- a/q-system/.q-system/capability-manifest.json
+++ b/q-system/.q-system/capability-manifest.json
@@ -51,6 +51,10 @@
       "runner": "bash"
     },
     {
+      "path": "q-system/.q-system/scripts/test/test-ask842-dispatch-decision.sh",
+      "runner": "bash"
+    },
+    {
       "path": "q-system/.q-system/scripts/test/test-attempts-ledger.py",
       "runner": "python3"
     },
@@ -80,12 +84,21 @@
       "runner": "bash"
     },
     {
+      "path": "q-system/.q-system/scripts/test/test-client-name-guard.py",
+      "runner": "python3"
+    },
+    {
       "path": "q-system/.q-system/scripts/test/test-containment-gitlink.py",
       "runner": "python3"
     },
     {
       "path": "q-system/.q-system/scripts/test/test-control-file-propagate.sh",
       "runner": "bash"
+    },
+    {
+      "path": "q-system/.q-system/scripts/test/test-converge-crossrepo-receipt.sh",
+      "runner": "bash",
+      "timeout_s": 120
     },
     {
       "path": "q-system/.q-system/scripts/test/test-converge.sh",
@@ -101,9 +114,18 @@
       "timeout_s": 120
     },
     {
+      "path": "q-system/.q-system/scripts/test/test-dispatch-per-repo-concurrency.sh",
+      "runner": "bash",
+      "timeout_s": 180
+    },
+    {
       "path": "q-system/.q-system/scripts/test/test-dispatch-stale-checkout.sh",
       "runner": "bash",
       "timeout_s": 120
+    },
+    {
+      "path": "q-system/.q-system/scripts/test/test-dor-refuses-unroutable.py",
+      "runner": "python3"
     },
     {
       "path": "q-system/.q-system/scripts/test/test-findings-block-reader.sh",
@@ -329,10 +351,6 @@
       "timeout_s": 600
     },
     {
-      "path": "q-system/.q-system/scripts/test/test-ask842-dispatch-decision.sh",
-      "runner": "bash"
-    },
-    {
       "path": "q-system/.q-system/scripts/test/test-review-artifact-repo-keyed.sh",
       "runner": "bash"
     },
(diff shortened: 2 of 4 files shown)


assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-195-20260815-173812.md` (6882 bytes).

```
FINDINGS:
minor|Diff lines starting with `++` are dropped as headers, so a client name on such a line in a tracked .patch/.diff file is never scanned|q-system/.q-system/scripts/client-name-guard.py:126
minor|File paths are never scanned, so a client name in a new filename commits with clean content and a clean message|q-system/.q-system/scripts/client-name-guard.py:125
minor|The bypass trailer writes no audit record, unlike linear-issue-ref-check.py in the same lefthook stage which ledgers to linear-bypass.jsonl|q-system/.q-system/scripts/client-name-guard.py:176
minor|A missing message file silently drops the entire message surface with exit 0, while the staged surface fails closed on the same blindness|q-system/.q-system/scripts/client-name-guard.py:171
END FINDINGS
```

--- reviewer output ---

## Review: PR #195 — client-name-guard stage consolidation (ASK-747)

Read-only. Every repro below was run against a real `git commit` driving a real `.git/hooks/commit-msg`, with a real `HOME` and token list. No repo files modified.

---

### Findings

**1. minor — a client name on a `++`-prefixed source line is never scanned; the repo tracks 6 files of exactly that class**

`client-name-guard.py:126` filters diff lines with `not l.startswith("+++")`. A source line beginning with `++` becomes `+++…` in the diff and is dropped as a header. Committed `.patch` / `.diff` files carry `+++ b/<path>` lines natively.

Real producers, not hypothetical:

```
$ git ls-files | grep -Ei '\.(patch|diff)$'
plugins/kipi-dsse/prompts/codex-adversarial-review-scope.patch
q-system/.q-system/scripts/test/fixtures/review-tier/pr-53.diff
q-system/.q-system/scripts/test/fixtures/review-tier/pr-60.diff
q-system/output/captoken-claude-guardrails.patch
q-system/output/cc-spex-closeout-gate.patch
$ git grep -l -E '^\+\+' | wc -l
       6
```

Repro output:

```
=== A: client name on a +++ line inside a committed .patch fixture ===
rc = 0
[master (root-commit) b7abc2f] add a patch fixture ASK-1
commits in history: 1
client name now in the committed tree? True
```

Not a regression — `origin/main`'s `--staged` branch carries the identical filter. But this PR rewrites the guard and re-ships it verbatim while the PR body claims "One invocation scans **both** surfaces."

**2. minor — file paths are never scanned on either surface**

`client-name-guard.py:125-126` keeps only `+` content lines; the `+++ b/<path>` line that carries a new file's name is discarded. A filename is public.

```
=== B: client name in the FILENAME only, clean content, clean message ===
rc = 0
[master (root-commit) 1490105] add notes ASK-1
files in HEAD: zorptech-engagement.md
```

Also pre-existing.

**3. minor — the bypass writes no record, unlike its sibling in the same lefthook stage**

`client-name-guard.py:176` prints and returns 0. `linear-issue-ref-check.py:39` in the same `commit-msg` stage appends its bypass to `q-system/output/linear-bypass.jsonl` so bypasses stay countable. This PR promotes the client-name bypass to a first-class trailer **with a reason field**, then discards the reason.

```
rc = 0
client-name-guard: bypassed via trailer 'client-name-guard-skip: permission on file'

NEW non-.git files written anywhere by the bypass: ['…/repo/study.md']
any *.jsonl ledger anywhere in the tree: []
client name now permanently in a commit object: True
```

Higher blast radius than the bypass that *is* ledgered: this one publishes permanently to a public repo with 23 forks.

**4. minor — the staged half fails closed, the message half fails open and silent**

`client-name-guard.py:171`: `args.message.read_text() if args.message.exists() else ""`. A missing message file drops the entire message surface with zero output, while `staged_added_lines()` (line 183-187) refuses loudly on the same class of blindness and documents "a guard that cannot see what is being committed must not report all clear."

```
=== D: --message pointing at a file that does not exist ===
rc = 0 | stdout = '' | stderr = ''
```

Severity dropped one level per the calibration rule: I could not show a producer emitting it. lefthook's `{1}` substitutes correctly — proven by the sibling `linear-issue-ref-check.py {1}` in the same stage.

---

### What is sound

- **`git commit -am` via the temporary index.** My primary hypothesis was that the stage move lost `-a` coverage, since git hands pre-commit a temp index. It does not: git sets `GIT_INDEX_FILE` at commit-msg too, and `git diff --cached` honors it. `[hook] GIT_INDEX_FILE is [.../index.lock]` → `BLOCK: client name in staged content: zorptech`, rc=1, `HEAD:notes.md -> 'clean baseline\n'`.
- **`git commit --amend`.** Blocks, rc=1, amended content never lands.
- **The hook-firing matrix — no operation lost coverage.** `cherry-pick`, `revert`, `rebase` fire *neither* hook (before and after), so removing pre-commit cost nothing. `merge --no-ff` fires **commit-msg only** — so the move actually *gains* merge coverage the old pre-commit stage never had. Verified end-to-end: a merge importing a client name now returns rc=1 and `Not committing merge`.
- **The #194 blocker (cross-commit `COMMIT_EDITMSG` leak).** Closed: N rc=0, N+1 rc=1.
- **Mention-vs-use.** Prose naming the token, and an indented quote of the trailer, both still scan and block. The column-0 anchor holds.
- **`--staged` removal has no orphaned caller.** `lefthook.yml:115` grep-confirmed as the only one; root `lefthook.yml` does not propagate to instances, so no fleet blast radius.
- **Manifest integrity after the reorder:** 161 tests, 0 duplicates, 0 missing paths, correctly sorted. The churn is normalization, not loss.
- **The fixtures are producer-built, which is the right fix for #194's defect.** git itself writes the message file; the hook is the production wiring. `env.pop("LEFTHOOK_COMMIT_MSG_FILE")` is a genuine negative control.
- **Claimed gates re-run:** `test-client-name-guard.py` → all 9 passed. `test-repo-preflight.sh` → `70 passed, 0 failed`. Both match the PR body.

Not run: `lefthook validate` (the command was permission-denied in this session). The PR's claim there is unverified by me; the YAML is well-formed on read.

---

Only minor findings survived reproduction. The central claim — one scan at the only stage that sees both facts — holds under every attack I could construct, and it closes #194's blocker while quietly widening coverage to merge commits.

**Fix first: finding 1.** It is the one place where a client name reaches the public repo with nobody deciding, and the repo already tracks six files of the shape that triggers it. Dropping the `+++` special-case and instead stripping the leading `+` from every line except the two real headers (matched by `diff --git`/`--- a/`/`+++ b/` context, not by prefix alone) closes 1 and 2 together.

FINDINGS:
minor|Diff lines starting with `++` are dropped as headers, so a client name on such a line in a tracked .patch/.diff file is never scanned|q-system/.q-system/scripts/client-name-guard.py:126
minor|File paths are never scanned, so a client name in a new filename commits with clean content and a clean message|q-system/.q-system/scripts/client-name-guard.py:125
minor|The bypass trailer writes no audit record, unlike linear-issue-ref-check.py in the same lefthook stage which ledgers to linear-bypass.jsonl|q-system/.q-system/scripts/client-name-guard.py:176
minor|A missing message file silently drops the entire message surface with exit 0, while the staged surface fails closed on the same blindness|q-system/.q-system/scripts/client-name-guard.py:171
END FINDINGS

PR #196, opened 2026-08-16
The committed copy is what you read; the installed copy is what runs (ASK-860)

assafkip:
Closes the drift half of ASK-860. The cap number itself (founder's 10) and ASK-833 are out of scope per the DoR.

## What was missing

On 2026-08-15 the dispatch daily cap read **12** in the committed template and **40** in the installed job at `~/Library/LaunchAgents/com.kipi.dispatch.plist`. The committed file is what a reader inspects to answer "what is the cap?"; the installed one is the cap that actually runs. Nothing compared them, so for an unknown period the loop was authorised to start 40 issues a day against a stated ceiling of 12.

`install-plist.sh` makes rendering correct when it is USED. It cannot notice a job installed some other way, or edited in place afterwards.

## What this adds

A `plist-drift` detector on the fleet health path (`kipi health`, the 08:15 LaunchAgent). It renders every committed `com.kipi.*` template through `install-plist.sh --render-only` and diffs it against the job on disk, reporting each divergence by key with the committed value and the live value.

Not a commit hook: the drift lives on the machine, not in the diff. Both copies can be individually correct in git and still disagree at rest, and CI has no `~/Library/LaunchAgents` to look at.

## Design calls worth a reviewer's attention

- **One substituter.** The render shells `install-plist.sh --render-only` rather than re-implementing its `sed`. Two definitions of "rendered" would drift in exactly the way this detector exists to end.
- **Report, never repair.** A live value may have been set deliberately during an incident. The finding names the divergence and offers the render+diff command; the human chooses.
- **Values are published by who authored the key, not by a filter over the text.** A key the repo declares gets its live value printed, because that is the answer the operator came for. A key present only on the installed job is named and never quoted: that is the one place a hand-added credential could sit, and a Linear issue is permanent. This is ASK-204's structural split, so there is no denylist to bypass.
- **Comments are stripped before parsing.** The fleet's own templates carry `--` inside XML comments (long prose blocks citing PR rounds). CoreFoundation accepts that, so launchd loads them and the jobs run; expat, behind `plistlib`, rejects the document outright. Reading a live job with a stricter parser than launchd's would report every commented template as unreadable.
- **A worktree render is refused.** `install-plist.sh` resolves `__KIPI_REPO__` from its own location, so rendering from a worktree points every `ProgramArguments` at that worktree and every job reads as drifted on a path nobody changed. Same discriminator `install-plist.sh --all` already refuses on.
- **Unrenderable / unparseable is its own finding, never silence.** The silent-disable shape this file has been corrected for three times: quiet must not be indistinguishable from healthy.
- **Not installed is not drift.** `detect_dark_jobs` already owns "a job that should be running is not". Filing it here too would put one condition on two permanent issues.

## Verification

Reproducer first, RED observed against HEAD (the detector stashed away):

```
$ python3 q-system/.q-system/scripts/test/test-plist-drift.py
AttributeError: module 'fh' has no attribute '_XML_COMMENT_RE'
EXIT_AT_HEAD=1
```

Green with the change, 17 checks:

```
$ python3 q-system/.q-system/scripts/test/test-plist-drift.py
  ok: a drifted live value is DETECTED
  ok: the body NAMES THE KEY
  ok: the body carries the COMMITTED value
  ok: the body carries the LIVE value
  ok: an identical installed job produces NO finding
  ok: a live-only key's VALUE is never published
  ok: an unrenderable template yields a finding, not silence
  ok: plist-drift is registered
  ...
all plist-drift checks passed
EXIT=0
```

The paired suite still passes, and its "all N shipped detectors run" line moved 7 -> 8:

```
  ok: all 8 shipped detectors run and return findings with subjects
PASS: fleet-health-daily contract holds
FLEET_EXIT=0
```

The fixture is producer-derived: the committed side of every case is the real `com.kipi.dispatch.plist` put through the real installer, with one value moved. The negative self-test (identical job files nothing) is what stops the check being decoration.

Registered in `capability-manifest.json`.

### DIFF
--- a/.prd-os/receipts.jsonl
+++ b/.prd-os/receipts.jsonl
@@ -130,3 +130,4 @@
 {"commit_sha": "ed1a901e0a5b7e8e9c004032c836123c6f49bb7c", "issue_id": "ASK-842", "receipts": {"reviewed": "2026-08-15T06:47:50Z"}, "reviewed_at": "2026-08-15T06:47:50Z"}
 {"commit_sha": "53d1b42256ac5b1da930a569e27e458534824089", "issue_id": "ASK-841", "receipts": {"reviewed": "2026-08-15T08:18:54Z"}, "reviewed_at": "2026-08-15T08:18:54Z"}
 {"commit_sha": "d652455f4a1f9d9fe347cba7cc217cba7cb41ec9", "issue_id": "ASK-839", "receipts": {"reviewed": "2026-08-15T10:39:41Z"}, "reviewed_at": "2026-08-15T10:39:41Z"}
+{"commit_sha": "4df16a050d4c08222bde42ab19772bd01046612c", "issue_id": "ASK-860", "receipts": {"reviewed": "2026-08-16T14:38:25Z"}, "reviewed_at": "2026-08-16T14:38:25Z"}
--- a/q-system/.q-system/capability-manifest.json
+++ b/q-system/.q-system/capability-manifest.json
@@ -144,6 +144,10 @@
       "runner": "python3"
     },
     {
+      "path": "q-system/.q-system/scripts/test/test-plist-drift.py",
+      "runner": "python3"
+    },
+    {
       "path": "q-system/.q-system/scripts/test/test-gh-repo-scope-converge.sh",
       "runner": "bash"
     },
--- a/q-system/.q-system/scripts/fleet-health-daily.py
+++ b/q-system/.q-system/scripts/fleet-health-daily.py
@@ -1225,7 +1225,255 @@ def detect_stale_runtime_plugins(_ctx) -> list:
     }]
 
 
+# ---------------------------------------------------------------------------
+# The INSTALLED job vs its RENDERED COMMITTED template (ASK-860)
+#
+# THE SCAR. 2026-08-15 the dispatch daily cap read 12 in
+# `q-system/.q-system/scripts/com.kipi.dispatch.plist` and **40** in
+# `~/Library/LaunchAgents/com.kipi.dispatch.plist`. The committed file is the one
+# a reader inspects to answer "what is the cap?"; the installed one is the cap
+# that actually runs. For an unknown period the loop was authorised to start 40
+# issues a day against a stated ceiling of 12 -- roughly 240 agent sessions a day
+# against a stated 72 -- and nothing in the fleet compared the two copies.
+#
+# `install-plist.sh` makes rendering correct when it is USED. It cannot notice a
+# job installed some other way, or edited in place after the fact. This detector
+# is the thing that notices.
+#
+# WHY HERE AND NOT A COMMIT HOOK. The drift lives on the MACHINE, not in the
+# diff: both copies can be individually correct in git and still disagree at
+# rest. CI has no `~/Library/LaunchAgents` to look at, so a CI-only check is
+# structurally blind -- the same reasoning `detect_stale_runtime_plugins`
+# documents for the plugin registry.
+#
+# REPORT, NEVER REPAIR. A live value may have been set deliberately during an
+# incident. Overwriting it silently is how a deliberate change becomes an
+# unexplained regression, so the finding names the divergence and offers the
+# render command; the human chooses.
+# ---------------------------------------------------------------------------
+
+PLIST_TEMPLATE_DIR = HERE
+PLIST_INSTALLER = HERE / "install-plist.sh"
+
+# A published value is capped, not redacted. This bounds a body, it does not
+# pretend to be a secret filter -- that job is done structurally below.
+_VALUE_CAP = 200
+
+# An XML comment, stripped before parsing. NOT a leniency about plist DATA:
+# comments carry none. It exists because the fleet's own committed templates
+# contain `--` inside comments (they are long prose blocks citing PR rounds),
+# which CoreFoundation accepts -- so launchd loads them and the jobs run -- while
+# expat, the parser behind `plistlib`, rejects the document outright. Reading a
+# live job with a stricter parser than the one launchd uses would report every
+# commented template as unreadable, which is a blind spot wearing an error's
+# clothes. Non-greedy: a comment containing `--` but not `-->` is one comment.
+# A literal `<!--` cannot appear inside a plist <string>, which would have to
+# escape the `<` as `&lt;` to be XML at all.
+_XML_COMMENT_RE = re.compile(r"<!--.*?-->", re.S)
+
+
+def plist_pairs(text: str) -> dict:
+    """A plist's leaf values, keyed by dotted path. Raises ValueError if unreadable.
+
+    Flattened rather than compared as nested objects so a finding can NAME THE
+    KEY that moved (`EnvironmentVariables.KIPI_DISPATCH_DAILY_MAX`) instead of
+    saying two dicts differ, which is what the operator needs to act.
+    """
+    import plistlib
+
+    try:
+        data = plistlib.loads(_XML_COMMENT_RE.sub("", text).encode())
+    except Exception as exc:  # noqa: BLE001 - any parse failure is "unreadable"
+        raise ValueError(f"{type(exc).__name__}: {exc}") from exc
+    out: dict = {}
+    _flatten(data, "", out)
+    return out
+
+
+def _flatten(node, prefix: str, out: dict) -> None:
+    if isinstance(node, dict):
+        for key, value in node.items():
+            _flatten(value, f"{prefix}.{key}" if prefix else str(key), out)
+    elif isinstance(node, (list, tuple)):
+        for index, value in enumerate(node):
+            _flatten(value, f"{prefix}[{index}]", out)
+    else:
+        out[prefix] = str(node)
+
+
+def _render_committed(path: Path) -> str:
+    """The committed template AS INSTALL-PLIST.SH WOULD WRITE IT.
+
+    Shelling the installer rather than re-implementing its `sed` is the point:
+    two substituters would be two definitions of "rendered", and the drift
+    between THEM would be invisible in exactly the way this detector exists to
+    end. `--render-only` writes to a path and touches no live job.
+    """
+    import tempfile
+
+    label = path.stem
+    with tempfile.TemporaryDirectory() as work:
+        out = Path(work) / f"{label}.plist"
+        res = subprocess.run(
+            ["bash", str(PLIST_INSTALLER), label, "--render-only", str(out)],
+            capture_output=True, text=True, timeout=60)
+        if res.returncode != 0 or not out.is_file():
+            raise RuntimeError(
+                f"install-plist.sh --render-only exited {res.returncode}")
+        return out.read_text()
+
+
+def _shown(value: str) -> str:
+    return value if len(value) <= _VALUE_CAP else f"{value[:_VALUE_CAP]}... (truncated)"
+
+
+def _unrenderable(label: str, reason: str) -> dict:
+    """A job whose two copies could not be COMPARED. Unknown, never clean.
+
+    The silent-disable shape this file has been corrected for three times over:
+    a `return []` here would make "I could not look" print identically to "I
+    looked and the job matches". Its own subject, so it cannot collide with a
+    real drift finding for the same label.
+    """
+    return {
+        "subject": f"{label}--unrenderable",
+        "title": f"cannot compare {label} against its committed template",
+        "body": (
+            f"`{label}` could not be compared with its committed template: {reason}\n\n"
+            "While this is true, NOTHING checks whether the installed copy of this "
+            "job matches the one in the repo. Absence of a drift finding for "
+            f"`{label}` is not evidence the two copies agree.\n\n"
+            "## Action\n```bash\n"
+            f"bash q-system/.q-system/scripts/install-plist.sh {label} "
+            "--render-only /tmp/render.plist\n"
+            f"diff /tmp/render.plist ~/Library/LaunchAgents/{label}.plist\n```\n"
+            "Fix what that reports, then confirm this issue stops re-filing."
+        ),
+    }
+
+
+def _drift_finding(label: str, changed: list, only_committed: list,
+                   only_live: list) -> dict:
+    sections = []
+    if changed:
+        sections.append(
+            "## Values that differ\n"
+            "| Key | Committed | Live |\n| -- | -- | -- |\n"
+            + "\n".join(f"| `{k}` | `{_shown(c)}` | `{_shown(l)}` |"
+                        for k, c, l in changed))
+    if only_committed:
+        sections.append(
+            "## Declared in the repo, absent from the installed job\n"
+            + "\n".join(f"- `{k}` (committed `{_shown(v)}`)" for k, v in only_committed))
+    if only_live:
+        # KEY NAMES ONLY, and that is structural rather than a filter over the
+        # text (ASK-204). A key the repo DECLARES is the repo's own knob, so its
+        # live value is the answer the operator came for. A key that exists only
+        # on the installed copy is operator-authored -- the one place a
+        # hand-added credential could sit -- and a Linear issue is permanent.
+        # Nine review rounds found nine bypasses of a value-shaped denylist; this
+        # splits on WHO AUTHORED THE KEY, which has no bypass to find.
+        sections.append(
+            "## Present only on the installed job\n"
+            "Key names only: these were not authored in the repo, and a Linear "
+            "issue is permanent (ASK-204). Read the values in the file itself.\n"
+            + "\n".join(f"- `{k}`" for k in only_live))
+    sections.append(
+        "## Action — read it, then decide\n"
+        "This is REPORTED, not repaired. A live value may have been set "
+        "deliberately during an incident, and overwriting it silently is how a "
+        "deliberate change becomes an unexplained regression.\n\n"
+        "```bash\n"
+        f"bash q-system/.q-system/scripts/install-plist.sh {label} "
+        "--render-only /tmp/render.plist\n"
+        f"diff /tmp/render.plist ~/Library/LaunchAgents/{label}.plist\n```\n"
+        "Then EITHER commit the live value into the template, OR re-install from "
+        f"the template (`install-plist.sh {label}`, which also reloads the job). "
+        "Whichever you pick, both copies end up saying the same thing."
+    )
+    return {
+        # The LABEL, so one job is one permanent issue however many keys move.
+        # A key-set subject would fork a new permanent issue every time the set
+        # of divergent keys changed.
+        "subject": label,
+        "title": f"installed job drifted from its committed template: {label} "
+                 f"({len(changed) + len(only_committed) + len(only_live)} key(s))",
+        "body": "\n\n".join(sections),
+    }
+
+
+def plist_drift_findings(template_dir=None, launch_agents=None, render=None) -> list:
+    """Every committed `com.kipi.*` template vs the job installed from it.
+
+    Injectable so the compare is provable against fixtures: a test that pointed
+    at the real `~/Library/LaunchAgents` would pass or fail on the state of
+    whoever's laptop ran it.
+    """
+    templates = Path(template_dir) if template_dir else PLIST_TEMPLATE_DIR
+    agents = Path(launch_agents) if launch_agents else LAUNCH_AGENTS
+    render = render or _render_committed
+
+    out = []
+    for template in sorted(templates.glob("com.kipi.*.plist")):
+        label = template.stem
+        live_path = agents / f"{label}.plist"
+        # NOT INSTALLED IS NOT DRIFT. `detect_dark_jobs` already owns "a job that
+        # should be running is not"; filing it here too would put one condition
+        # on two permanent issues.
+        if not live_path.is_file():
+            continue
+        try:
+            committed = plist_pairs(render(template))
+            live = plist_pairs(live_path.read_text(errors="ignore"))
+        except (ValueError, RuntimeError, OSError) as exc:
+            out.append(_unrenderable(label, str(exc)))
+            continue
+        changed = [(k, committed[k], live[k]) for k in sorted(committed)
+                   if k in live and committed[k] != live[k]]
+        only_committed = [(k, committed[k]) for k in sorted(committed) if k not in live]
+        only_live = [k for k in sorted(live) if k not in committed]
+        if changed or only_committed or only_live:
+            out.append(_drift_finding(label, changed, only_committed, only_live))
+    return out
+
+
+def detect_plist_drift(_ctx) -> list:
+    """`plist_drift_findings`, refused from a worktree.
+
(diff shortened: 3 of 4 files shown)


assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-196-20260816-071920.md` (7532 bytes).

```
FINDINGS:
major|Binary-format installed plists are decoded as text and mangled, so an identical job files a false 13-key drift issue while real drift is never named|q-system/.q-system/scripts/fleet-health-daily.py:1427
minor|The 200-char value cap truncates both sides, so a divergence past char 200 renders as two identical cells in the finding table|q-system/.q-system/scripts/fleet-health-daily.py:1326
nit|_render_committed renders by label from the module-level template dir, silently ignoring the injected template_dir|q-system/.q-system/scripts/fleet-health-daily.py:1304
END FINDINGS
```

--- reviewer output ---

## Review — PR #196 (ASK-860, `plist-drift` detector)

I read the diff, then attacked the detector against fixtures built from the real producer (`install-plist.sh --render-only`). Three findings survived reproduction.

---

### 1. `major` — a binary-format installed plist is decoded as text, mangled, and reported as a false 13-key drift

`q-system/.q-system/scripts/fleet-health-daily.py:1427`

```python
live = plist_pairs(live_path.read_text(errors="ignore"))
```

`plist_pairs` takes `str`. `~/Library/LaunchAgents/*.plist` is not required to be XML. `defaults write <plist> KEY value` — the canonical in-place launchd edit, and exactly the hand-edit this detector exists to catch — rewrites the file as `bplist00`, as does `plutil -convert binary1` and every GUI launchd editor. `read_text(errors="ignore")` drops every invalid UTF-8 byte, `.encode()` re-emits shorter bytes, and `plistlib` still *accepts* the corrupted binary (offsets shift, it returns a bytes scalar) rather than raising — so the `_unrenderable` safety net at `:1339` never engages.

**Reproducer** (stock `plutil`, live copy byte-identical to the template before conversion):

```
$ python3  # renders com.kipi.dispatch via install-plist.sh, copies it verbatim
           # to the LaunchAgents fixture, then: plutil -convert binary1 <live>
plutil -convert binary1: 0
header: b'bplist00'
findings on an IDENTICAL job: 1
TITLE: installed job drifted from its committed template: com.kipi.dispatch (13 key(s))
```

Body of that false finding:

```
## Declared in the repo, absent from the installed job
- `Label` (committed `com.kipi.dispatch`)
- `ProgramArguments[0]` (committed `/bin/bash`)
- `StartInterval` (committed `900`)
  ... 12 keys total
## Present only on the installed job
- ``
```

And the parse it actually produced:

```
$ plist_pairs(binary_bytes.decode(errors='ignore'))
{'': "b'abel_\\x10\\x10Progr'"}
```

Two failures in one input, both anchored:
- **Cries wolf.** A healthy job files a permanent Linear issue asserting the installed dispatcher has no `Label`, no `ProgramArguments` and no `StartInterval` — i.e. "your dispatcher cannot run." Plus an empty `` - `` `` bullet that is the mangled parse leaking into the issue body.
- **Misses the thing it was built for.** I ran the actual 2026-08-15 scar shape (`KIPI_DISPATCH_DAILY_MAX` 10 → 40) against a binary live copy: `names key+live value?: False`. The cap drift is never named.

Calibration note, stated openly: nothing *in this repo* emits a binary plist (`grep -rn "plutil -convert\|defaults write" q-system plugins` → no hits). I did not drop the severity for that, because the detector's entire premise is drift authored **outside** the repo's tooling, and `defaults write` is the standard way a human does it. If you disagree with that reading, downgrade to minor — the repro is the same either way.

Fix is one line: `plist_pairs` on `live_path.read_bytes()`, stripping XML comments only when the header is not `bplist00`. The test suite has no binary fixture, which is why this passed 17 green checks.

---

### 2. `minor` — the 200-char cap truncates both sides, so a late-diverging value renders as two identical cells

`q-system/.q-system/scripts/fleet-health-daily.py:1326` (`_shown`, `_VALUE_CAP = 200` at `:1260`)

```
$ row = fh._drift_finding("com.kipi.demo",
        [("EnvironmentVariables.LONG", "x"*250+"AAA", "x"*250+"BBB")], [], [])
truncated row identical on both sides?: True
```

The issue asserts a divergence while displaying two byte-identical values. No current template holds a value near 200 chars (longest real one is `ProgramArguments[1]`, ~90), so this is bounded and nobody hits it today — hence minor. A common-prefix-aware `_shown` (show the window around the first differing offset) fixes it.

---

### 3. `nit` — `_render_committed` ignores the injected `template_dir`

`q-system/.q-system/scripts/fleet-health-daily.py:1304` renders by `label` out of the module-level `PLIST_TEMPLATE_DIR` (`:1255`), so `plist_drift_findings(template_dir=X)` with the default renderer compares files from `X` against renders from `HERE`. Unreachable today (the test always injects `render`), but the two knobs read as paired and are not. Same class in the test: `test-plist-drift.py:170` calls `_dirs(...)` twice and takes `[0]` from one temp root and `[1]` from another.

---

## What is sound

- **Worktree refusal actually fires.** Verified live from this review worktree: `REPO_ROOT/.git is dir? False` → `['render-root-is-a-worktree']`. The false-issue-per-job blowup is genuinely closed.
- **Non-plist live file → finding, not silence.** Wrote `"not a plist at all"` as the live copy → `['com.kipi.dispatch--unrenderable']`. The silent-disable shape is closed for text garbage (binary is finding #1).
- **ASK-204's structural split holds even under the mangled parse.** I tried to get an operator-authored value published: `only_live` prints key names only, and under the binary mangle the garbage keys never match a repo-declared key, so they can never route into the value-printing `changed` branch. No path published `lin_api_realvalue`.
- **No subject collision with `detect_dark_jobs`**, which also uses `subject = label`. `finding_key` (`:1575`) namespaces by detector id, so the dedup keys differ.
- **A detector exception cannot take out the health run or read as clean.** `subprocess.run(..., timeout=60)` raises `TimeoutExpired`, which is *not* in the caught tuple at `:1428` — but `run_detectors` (`:1907`) records any raise as `DETECTOR_ERROR`, never as 0, and `blind_detectors` surfaces it. Correct by construction.
- **No day-one flood.** All five committed templates render `rc=0`, so the detector does not open with five `--unrenderable` issues.
- **Comment stripping.** I tried to break `_XML_COMMENT_RE` with `-->` inside a `<string>` and with `--` runs; a literal `<!--` cannot appear in plist XML unescaped, and the non-greedy match handles the rest. Could not make it fail — dropping that line of attack.
- Report-never-repair, "not installed is not drift", and label-stable subjects all hold under the fixtures.

The comment density is unusually high but every claim I spot-checked matched the code, with the one exception in finding #1 (the `_unrenderable` docstring's "unparseable is its own finding, never silence" is not true for binary input).

---

## VERDICT: REQUEST CHANGES

One major → request changes, not block: nothing here is unrecoverable, and the change is a purely additive detector that reverts cleanly.

**Fix first:** `fleet-health-daily.py:1427` — read the live plist as bytes and let `plistlib` pick the format, then add a `FMT_BINARY` fixture to `test-plist-drift.py`. Without it the detector files a false "your dispatcher has no ProgramArguments" issue against a healthy job, and stays blind to the exact cap drift it was built for.

FINDINGS:
major|Binary-format installed plists are decoded as text and mangled, so an identical job files a false 13-key drift issue while real drift is never named|q-system/.q-system/scripts/fleet-health-daily.py:1427
minor|The 200-char value cap truncates both sides, so a divergence past char 200 renders as two identical cells in the finding table|q-system/.q-system/scripts/fleet-health-daily.py:1326
nit|_render_committed renders by label from the module-level template dir, silently ignoring the injected template_dir|q-system/.q-system/scripts/fleet-health-daily.py:1304
END FINDINGS

assafkip:
## Round 1 addressed — all three findings taken, all three reproduced first

Pushed as `4df16a05`.

### major — binary live plist read as text

Taken as major, not downgraded. Your calibration note is the right read: the detector's premise is drift authored *outside* the repo's tooling, and `defaults write` is how a human does it. That nothing in-repo emits binary is exactly why no fixture caught it.

`plist_pairs` now takes `bytes` or `str`. A binary document reaches `plistlib` untouched; an XML one is decoded **strictly**, so a decode failure becomes an `--unrenderable` finding instead of a confident wrong answer. Call site reads `live_path.read_bytes()`.

Both halves of your repro are now pinned, and I watched them fail against `fda69e25` before they passed:

| check | at `fda69e25` | now |
| -- | -- | -- |
| a BINARY installed job identical to its template files NOTHING | fail | ok |
| the 2026-08-15 drift is still caught when the live job is BINARY | ok | ok |
| the BINARY finding carries the live value | fail | ok |

The middle row is worth naming: at HEAD the finding *existed* but never carried `40`, which is the shape where a green count hides a blind detector.

Also added: a live copy that is neither valid XML nor a real binary plist (`b"\xff\xfe not a plist \x00\x01"`) yields `com.kipi.dispatch--unrenderable`, closing the silent-disable shape for the binary side too.

### minor — the 200-char cap truncated both sides identically

Taken. `_shown(value, other)` centres the window on the first differing offset when both sides exceed the cap. Your exact case is now a check: `_shown(a, b) == _shown(b, a)` must be `False`, and both `AAA` and `BBB` appear in the rendered row. A value under the cap is still shown whole.

### nit — `template_dir` ignored by the default renderer

Taken as written: the two knobs read as paired and are not. `install-plist.sh` renders only from its own directory, so `plist_drift_findings(template_dir=X)` with the default renderer now **raises** rather than comparing one tree's files against another tree's renders. Unreachable in production, but it can no longer be reached wrongly.

Same class in the test, which you also flagged: the blind-render case took templates from one `_dirs()` call and agents from a second. One fixture root now.

### Verification

```
$ python3 q-system/.q-system/scripts/test/test-plist-drift.py
  ...
  ok: a BINARY installed job identical to its template files NOTHING
  ok: the 2026-08-15 drift is still caught when the live job is BINARY
  ok: the BINARY finding carries the live value
  ok: an undecodable live job is UNKNOWN, not silently clean
  ok: the two truncated cells are NOT identical
  ok: a moved template_dir with the default renderer is REFUSED
all plist-drift checks passed
EXIT=0        # 28 checks, was 17
```

```
  ok: all 8 shipped detectors run and return findings with subjects
PASS: fleet-health-daily contract holds
FLEET_EXIT=0
```

Nothing contested. Thanks for building the binary fixture — that was the gap.

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-196-20260816-073101.md` (5533 bytes).

```
FINDINGS:
minor|`_flatten` stringifies leaves, so an integer-to-string type change with identical digits is compared equal and reads as no drift|q-system/.q-system/scripts/fleet-health-daily.py:1328
minor|The "not installed is not drift" comment credits `detect_dark_jobs` with coverage it cannot have, since it enumerates installed plists rather than committed templates|q-system/.q-system/scripts/fleet-health-daily.py:1476
END FINDINGS
```

--- reviewer output ---

## Review — PR #196 (ASK-860, `plist-drift` detector)

Read the diff, then attacked it against the real machine rather than the fixtures. Both findings below are reproduced; everything else I tried survived.

---

### Finding 1 — minor: `_flatten` stringifies every leaf, so a type change with the same digits reads as no drift

`q-system/.q-system/scripts/fleet-health-daily.py:1328` — `out[prefix] = str(node)`. `<integer>900</integer>` and `<string>900</string>` both flatten to `"900"`, so the compare returns clean. `StartInterval` as a string is not a schedule launchd honours, which is exactly the class of in-place hand-edit this detector exists to catch.

Reproducer (real render of `com.kipi.dispatch` as the committed side, one type flipped on the live side, plus a value-change control):

```bash
python3 - <<'EOF'
# renders com.kipi.dispatch via install-plist.sh, mutates the live copy, runs plist_drift_findings
EOF
```

Real output:

```
committed StartInterval: 900
R1 StartInterval 900(int) vs '900'(str): findings=0
   control StartInterval 900 -> 60: findings=1
```

Calibration, stated plainly: no producer in this repo emits that shape. `defaults write … StartInterval 900` writes an integer and `plutil -convert` preserves types, so I had to construct the input, and severity is dropped for that. A job that will not load is also caught by `launchd-dark`. Fix is one line: key the compare on `(type(node).__name__, str(node))`.

---

### Finding 2 — minor: the "not installed is not drift" comment names a detector that structurally cannot see that case

`q-system/.q-system/scripts/fleet-health-daily.py:1476-1480`. The comment says `detect_dark_jobs` already owns "a job that should be running is not". It does not: `detect_dark_jobs` iterates `_launchd_labels()`, which globs `~/Library/LaunchAgents/*.plist`. A committed template with nothing installed produces no file to enumerate, so both detectors are silent about it.

Reproducer output (same script, second case):

```
R2 template committed, nothing installed: findings=0
   detect_dark_jobs over an empty LaunchAgents dir: []
```

This is not a hypothetical shape. `install-plist.sh:47` records the ASK-729 scar in its own words: six templates were committed and nothing installed them, so every job ran only where somebody had typed the command by hand. Measured on this machine right now: 5 committed `com.kipi.*` templates, 9 installed `com.kipi.*` jobs. Neither direction of that mismatch is reported by anything I could find (`grep -rn 'com\.kipi\.\*'` turns up only `install-plist.sh`, its test, and this new detector).

The `continue` itself is the right call; the justification is what is wrong. Either correct the comment to say the case is uncovered, or capture the gap.

---

## What is sound

- **Cries wolf on the real fleet.** Ran `plist_drift_findings` with the primary checkout's `install-plist.sh` as the renderer against the live `~/Library/LaunchAgents`. Output: `FINDINGS: 0`. No false positive on today's real state.
- **The comment-stripping rationale is real, not a story.** `plistlib.loads` on the installed `com.kipi.dispatch.plist` raises `ExpatError`; the detector reads it fine. Without `_XML_COMMENT_RE` this would file an unrenderable issue for the fleet's flagship job every morning.
- **Both tests are green and can fail.** `test-plist-drift.py` 28 checks pass; `test-fleet-health-daily.py` passes. The fixture is genuinely producer-derived (real template through the real installer), and the negative self-test (identical job files nothing) is present.
- **A detector that raises does not take out the run.** `run_detectors` (line 1980) catches per-detector, records `"error"`, and `blind_detectors` surfaces it. The uncaught `subprocess.TimeoutExpired` in `_render_committed` degrades to blind-and-reported, not silent-clean.
- **Credential leak via the `changed` path.** Tried to find a real one: scanned every installed `com.*` plist for secret-shaped `EnvironmentVariables` key names. Zero hits fleet-wide, and no committed `com.kipi.*` template declares a token/webhook key. The ASK-204 split holds for every input this repo actually produces. Dropped the finding.
- **Binary live plists, undecodable live plists, worktree render.** All three behave as documented; the worktree branch fires correctly from this review tree (`render-root-is-a-worktree`, one finding, stable subject).
- **Body stability.** Keys are sorted and values deterministic, so an unchanged fleet will not churn the Linear issue via `finding_hash`.
- **Concurrency.** Two runs render into separate `TemporaryDirectory`s and write nothing outside them; `--render-only` never touches a live job or calls `launchctl`.

---

Only minor findings survived reproduction, so this does not gate.

Most important thing to fix first: the comment at `fleet-health-daily.py:1476`. The code decision is right, the stated reason is false, and a false coverage claim in a comment is how the next person stops looking for the real gap.

FINDINGS:
minor|`_flatten` stringifies leaves, so an integer-to-string type change with identical digits is compared equal and reads as no drift|q-system/.q-system/scripts/fleet-health-daily.py:1328
minor|The "not installed is not drift" comment credits `detect_dark_jobs` with coverage it cannot have, since it enumerates installed plists rather than committed templates|q-system/.q-system/scripts/fleet-health-daily.py:1476
END FINDINGS

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-196-20260816-074233.md` (7150 bytes).

```
FINDINGS:
minor|install-plist.sh's stderr is discarded, so a render failure files a permanent Linear issue whose only diagnostic is the exit code|q-system/.q-system/scripts/fleet-health-daily.py:1349
minor|one broken or missing install-plist.sh files one permanent unrenderable issue per committed template (5 today) for a single root cause|q-system/.q-system/scripts/fleet-health-daily.py:1481
minor|_flatten stringifies every leaf, so an integer-to-string type flip on the same value compares equal and drift goes undetected|q-system/.q-system/scripts/fleet-health-daily.py:1328
END FINDINGS
```

--- reviewer output ---

## Review: PR #196 — `plist-drift` detector (ASK-860)

I read the diff, then ran the detector against the founder's real `~/Library/LaunchAgents` and against constructed fixtures. Two findings survived reproduction. Both are minor.

---

### FINDING 1 — minor: a render failure files a permanent issue whose only diagnostic is `exited N`; the installer's stderr is discarded

`_render_committed` captures stderr and never reads it (`q-system/.q-system/scripts/fleet-health-daily.py:1346`); the `RuntimeError` carries only the return code (`:1349`), and `_unrenderable` interpolates that as the whole reason (`:1387`). `install-plist.sh` emits the actual cause on stderr (`ERROR: unsubstituted placeholder remains in ...`, `ERROR: no plist template for label ...`, usage). None of it reaches the operator, and nothing prints it to the run log either.

The file's own ASK-204 precedent at `:1801` does the opposite deliberately: it publishes the exception *type* and says `full text in the run log` — because it first printed the full text to stderr at `:1788`. Here there is no run-log copy to fall back to.

Reproducer (fake installer that fails the way the real one does):

```
$ python3 - <<'PYEOF'   # loads the branch module, points PLIST_INSTALLER at a
                        # script that writes a diagnostic to stderr and exits 1
FINDINGS: 5
cannot compare com.kipi.dispatch against its committed template
---- body ----
`com.kipi.dispatch` could not be compared with its committed template: install-plist.sh --render-only exited 1
...
---- stderr text present in body? ----
False
```

3am scenario: the shared installer breaks (bad sync, bad `sed`, `bash` not resolvable under launchd's default PATH). The 08:15 job files a Linear issue that says a comparison failed and nothing about why. Recoverable — the body does hand over a repro command.

---

### FINDING 2 — minor: one broken installer fans out to one permanent Linear issue per template

`plist_drift_findings` (`:1478-1481`) appends an `_unrenderable` finding per label, and `file_findings` has no per-detector cap. A single shared-cause failure files N issues, all with the same undiagnosable body.

```
$ python3 - <<'PYEOF'   # PLIST_INSTALLER pointed at a path that does not exist
PERMANENT LINEAR ISSUES FILED FOR ONE ROOT CAUSE: 5
  fleet-health/plist-drift/com-kipi-dispatch-unrenderable            | ... exited 127
  fleet-health/plist-drift/com-kipi-fleet-health-unrenderable        | ... exited 127
  fleet-health/plist-drift/com-kipi-linear-daily-digest-unrenderable | ... exited 127
  fleet-health/plist-drift/com-kipi-linear-dor-unrenderable          | ... exited 127
  fleet-health/plist-drift/com-kipi-openloops-heartbeat-unrenderable | ... exited 127
```

Not a false alarm — the condition is real. The cost is 4 extra permanent issues a human closes by hand. Deduped by subject, so it does not re-file daily. Note this is the same shape the PR's own DoR argues against for `detect_dark_jobs` ("one condition on two permanent issues"), applied to the render leg.

---

### FINDING 3 — minor: type-only drift is invisible; `_flatten` stringifies every leaf

`out[prefix] = str(node)` at `:1328`. An `<integer>900</integer>` that becomes `<string>900</string>` compares equal.

```
$ python3 - <<'PYEOF'
committed StartInterval: 900 -> live: '900'
FINDINGS: 0
flattened committed: {'Label': 'com.kipi.demo', 'RunAtLoad': 'False', 'StartInterval': '900'}
flattened live     : {'Label': 'com.kipi.demo', 'RunAtLoad': 'False', 'StartInterval': '900'}
```

Honest caveat, per the calibration rule: no producer *in this repo* emits this — `install-plist.sh` copies the template's types verbatim. The producer would be `defaults write <plist> StartInterval 900` without `-int`, which is the exact hand-edit the module's own comment at `:1274` names as its motivating case. The sandbox denied me the `defaults` call, so I could not confirm that it writes a string rather than an integer. Severity dropped accordingly. Bool drift (`<true/>` → `<string>true</string>`) *is* caught, since `str(True)` is `"True"`.

---

## What is sound

- **False-positive storm on the live machine.** Ran `plist_drift_findings` against the real `~/Library/LaunchAgents` (9 installed `com.kipi.*` jobs) using the primary checkout's `install-plist.sh` as the renderer. `FINDINGS: 0`. The detector does not cry wolf on day one.
- **Credential leak into a permanent Linear object.** Audited all 5 committed templates: no credential-shaped key exists, so nothing a hand-edit could fill gets its live value published today. The committed-vs-live-authored split is structural with no denylist to bypass, and the test proves the live-only value is withheld.
- **Subject collision with `detect_dark_jobs`** — both use `subject: label`. `finding_key` (`:1635`) namespaces by detector id, so they cannot share a Linear key.
- **Detector blast radius.** `subprocess.TimeoutExpired` is not in the `except` tuple at `:1482`, but `run_detectors` (`:1982`) catches every exception per detector — it degrades to `DETECTOR_ERROR` and an unknown-not-clean Slack line rather than killing the 08:15 run.
- **Per-instance fan-out.** I expected instances running their own synced copy to render `__KIPI_REPO__` at the instance and report every job drifted. They don't: `kipi` resolves `KIPI_HOME` from the CLI's own location (`kipi:4,137`), so `kipi health` always runs the skeleton copy.
- **`_shown` windowing.** Traced prefix-equal, length-mismatch, and past-cap divergence cases; `first_diff` is symmetric, so both cells get the same window and stay distinguishable.
- **Concurrency.** Each render gets its own `TemporaryDirectory`; nothing writes to a live plist (the test asserts the live value is unchanged after a run).
- **Binary live plists.** The strict-decode path and the `bplist00` branch hold; both suites green (`test-plist-drift.py` 28 checks, `test-fleet-health-daily.py` PASS).
- **Manifest wiring.** Confirmed `test-plist-drift.py` lands in `expected_tests`, not a decorative section.

The negative self-test (identical job files nothing) is real, the fixture genuinely comes from `install-plist.sh --render-only` rather than the author's imagination, and I verified the reproducer would go red without the change.

---

First thing to fix: carry `res.stderr` into the `RuntimeError` at `fleet-health-daily.py:1349` (or print it to the run log the way `:1788` does), so an unrenderable finding says *why*.

FINDINGS:
minor|install-plist.sh's stderr is discarded, so a render failure files a permanent Linear issue whose only diagnostic is the exit code|q-system/.q-system/scripts/fleet-health-daily.py:1349
minor|one broken or missing install-plist.sh files one permanent unrenderable issue per committed template (5 today) for a single root cause|q-system/.q-system/scripts/fleet-health-daily.py:1481
minor|_flatten stringifies every leaf, so an integer-to-string type flip on the same value compares equal and drift goes undetected|q-system/.q-system/scripts/fleet-health-daily.py:1328
END FINDINGS

PR #73, opened 2026-08-03
Red CI on an agent PR is re-dispatched to its agent, not emailed to the founder (ASK-295)

assafkip:
## Problem

An autonomous agent opens a PR, CI goes red, GitHub emails the repo owner, and he cannot act on it. The agent that opened it can. Three of these reached the founder on 2026-08-02:

```
[assafkip/kipi-system] PR run failed: Skeleton Validation - ... (ASK-292)
[assafkip/kipi-system] PR run failed: Skeleton Validation - ... (ASK-288)
[assafkip/kipi-system] Run failed: Skeleton Validation - sana/block-expiry
```

All three were a **single correct catch** (`test-terminal-states.sh`, 13 passed 1 failed; `main` was green). That is the point: true, and still useless to him.

It is a dead end by construction. `ready()` in `linear-worker.sh` returns only `backlog`/`unstarted` issues, and an issue with a live PR is In Progress, so nothing in this repo ever looked at that PR again. ASK-294's `slack-notify.sh` chokepoint structurally cannot reach it: the producer is GitHub, driven by workflow conclusion, outside every sink this repo owns.

## The four DoR decisions, made

| Decision | Answer |
|---|---|
| Does a red agent PR reach the founder? | **No**, not on the red. Once, after the machine tier, in a message naming what the machine tried. |
| Where the consumer runs | Inside `kipi-dispatch.sh`, ahead of the fresh pick. **Not** a new launchd job. |
| Attribution | The branch name. `sana/ask-295` -> `ASK-295`, no Linear round-trip. |
| Attempt cap | One per PR **per failure signature**, in the existing flock'd attempts ledger. |

## Design notes worth reviewing

- **The signature is the set of failing check names, deliberately not the head sha.** A re-push that fails the same check is the same problem; re-dispatching on it is how a handler re-runs a flake forever. ASK-295 names that risk by hand (one root cause of the 2026-08-02 redness was a measurement against the wrong tree). A genuinely new failure earns its own fresh attempt.
- **Placed before the empty-`NEXT` exit.** Nothing-ready is the ordinary state of a healthy queue, so handling the red PR only when something else was also waiting would leave it unhandled on exactly the quiet cycles.
- **Not a new job.** Every cap it needs already exists and is proven in the dispatcher (`MAX_CONCURRENT`, daily budget, liveness assert, `page_once`, one dispatch per heartbeat). Per-repo jobs die silently: the income scanners went dark 6 days that way.
[...]

PR #123, opened 2026-08-06
review-redrive must see a reviewer slot that was NEVER POSTED (sp-d87c5416)

assafkip:
Closes `sp-d87c5416`. The item gating PR #122.

## This is a design-level correction, not three bug fixes

`reviewer_slot_failing()` is **two-valued** — failing / not-failing — over a **three-valued** world: failing, passing, never posted. `candidates()` branched on `if not slots: continue`, so silence was read as health. **18 of 29 open PRs had no reviewer status at all**, making the majority of the backlog invisible to the only mechanism that could move it.

Fixing that exposed the same shape one layer down, and this is the part worth reading:

**`classify()` answers "is this a refusal". On the never-posted path that is simply the wrong question.** It returns `(None, ...)` for any non-refusal, and the caller dropped those silently. Sampled across 96 verdict records:

| verdict | count | |
|---|---|---|
| APPROVE WITH NITS | 54 | was dropped |
| REQUEST CHANGES | 23 | the only class the fixtures used |
| APPROVE | 17 | was dropped |
| BLOCK | 1 | |
| `""` | 1 | |

**71 of 96 (74%) were on the wrong side of that question.** Not an edge case — the majority class, on a keystone that decides which PRs get worked. PR #23 is the concrete cost: approved, status never posted, auto-merge armed, waiting forever on a status nobody sends. The cheapest item in the queue was the one the selector could not see.

The fixtures caused it: every one used `REQUEST CHANGES`. The fall-through had two outcomes, one was asserted, production held the other. **The fixture encoded an assumption about the population instead of sampling it.** The table is now sampled, with counts and date, and a note to re-sample before adding a verdict.

## What it does not change

The ASK-318 refusal is intact: a **failing** slot with no verdict record means the producer ran and recorded nothing, and re-reviewing papers over a broken producer. An **absent** slot has no verdict to be missing.

Emits the existing `re-review` action deliberately — `kipi-dispatch.sh:1000` matches that exact string; any other value falls through to `./kipi converge`, buying a rework round against a review that does not exist.

## Also fixed

`json.JSONDecodeError` **is** a `ValueError`, so `read_record` reported a truncated record and a missing one identically, and the reason then asserts no verdict exists. Now returns a third answer, `CORRUPT_RECORD`; both call sites refuse rather than claim absence. Latent, not live: 0 unparseable across 96 records. (`sp-3e57348e`)

[...]

PR #91, opened 2026-08-04
A PR the reviewer refused gets re-entered, and rework is told from re-review (ASK-352)

assafkip:
Replaces #90, which was cut off `sana/ask-312` by mistake and carried ~84 files of unrelated unmerged work (89 files / 13,609 insertions). This branch is cut off `main` and is **7 files / 931 insertions**.

Six PRs sat parked ~29 hours. A PR whose reviewer said REQUEST CHANGES had no selector that re-enters it. `ci-redrive.py` excludes the reviewer's verdict slots from what it calls red CI -- correct, and untouched here -- but justified itself with "the reviewer comments on the Linear issue and the worker re-dispatches from there", and no such selector existed.

## `failure` on that slot means two opposite things

| PR | review | record says | right action |
|---|---|---|---|
| #82 | real, one real major | REQUEST CHANGES | rework |
| #80 | prompt's own FINDINGS template echoed back, "Reply `OK` and I'll execute" | REQUEST CHANGES | re-review |

#80's `stated` was lifted out of the prompt's grading rule, echoed by `codex exec`. Measured over all 79 verdict records: 13 unusable, carrying APPROVE (11, all merged), REQUEST CHANGES (#80, #83), empty (#89).

So `pr-review-agent.sh` persists `usable` into the record. Record-only; no gate moves on that commit.

## The selector

`review-redrive.py` reads `usable`, falls back to asking `pr-verdict-lib.sh` for legacy records (never a Python port of it), treats unknown as not-usable. Refuses what is not its: no record at all is ABSENT (ASK-318/ASK-313); a usable approval is the merge side (ASK-310). Drift outranks the verdict. Imports `is_reviewer_slot` from ci-redrive so the two cannot drift.

Consumer is `kipi-dispatch.sh`, after ci-redrive and only if it passed. rework -> converge; re-review -> `pr-review-agent.sh --post`.

## Evidence

- `test-verdict-usability.sh` 3/3; red at `REPRO_REF=origin/main`; mutants 2/2.
- `test-review-redrive.sh` 17/17; every case paired against its opposite; mutants 4/4.
- `test-ci-redrive.sh` 65/65 including its 10 dispatcher-integration cases.
- All four suites re-run from this clean branch, not the original one.
- `capability-manifest.json` resolved as a deterministic union: 128 base entries in, 130 out, 0 lost, exactly the 2 new tests added.

Two defects found in my own work by the live run, both fixed here:
[...]

PR #210, opened 2026-08-17
A cap-out the redrives can read, so a given-up issue stops restarting (ASK-871)

assafkip:
## The defect (ASK-871)

`converge.sh`'s exit-2 cap-out did exactly two things: `say` a line into the converge log and Slack the founder. Nothing machine-readable. So the two consumers whose whole job is re-entering a stuck PR had no way to learn the rounds were already spent, and re-entered it correctly by their own rules.

Measured 2026-08-16 on ASK-830 -- six converge rounds, five Opus reviews, one morning, nothing mergeable, ~70% of the day's autonomous spend:

```
15:59:53Z converge[ASK-830] STOP exit-2: hit the 3-round cap still at 'REQUEST CHANGES'
16:14:01Z red-CI redrive: handing ASK-830 back to its agent ahead of the fresh pick
```

Both bounding mechanisms worked. They did not compose.

## What changed

| File | Change |
|---|---|
| `attempts-ledger.py` | two ops: `record-capout <issue> <why>`, `clear-capout <issue>` |
| `converge.sh` | exit-2 writes the record, then pages with the clear command in the line |
| `ci-redrive.py` | `capout()` / `capout_skip()` -- the one definition -- and the gate in `cmd_redrive` |
| `review-redrive.py` | imports and calls the same gate, first in `cmd_select` |
| `test/test-converge-capout.sh` | new, registered in `capability-manifest.json` |

**The mechanism choice, per the DoR's "decide deliberately and say why".** The record lives in the attempts ledger, not a new file. `attempts-ledger.py` is already the flock'd single writer that both redrives and the worker read, and a second store for one issue's budget is precisely how the four budgets in that file ended up racing. The two new ops are symmetric with the existing `clear-conflict` / `clear-drift` / `clear-automerge` family.

**The per-SHA keying is untouched.** It is deliberate and correct, and it is why nothing caught this: each of ASK-830's rounds moved the head, so every per-sha cap read as fresh while the *issue* had already been given up on. The missing axis is dispatches per issue. That is what this adds.

## One deviation from the DoR's Files list, stated up front

The DoR named `review-redrive.py` as the reading side. The evidence line it quotes -- `red-CI redrive: handing ASK-830 back` -- is `kipi-dispatch.sh:1141`, which is the **`ci-redrive.py`** call site. Gating only `review-redrive.py` would have left the measured incident live.

So both redrives are gated, through **one** definition in `ci-redrive.py` that `review-redrive.py` imports, exactly as it already imports `is_reviewer_slot` and for the same stated reason: two hand-maintained copies of one refusal rule is how a state ends up owned by neither.

This does not touch anything on the Not-doing list. Ci-redrive's exclusion of the reviewer verdict slots (PR #73's scar) is unchanged; the addition is one more reason to refuse, read from the ledger.

## The clearing path ships in the same change

Per the blast-radius note: a cap-out no human can clear is a permanent park, and a park with no alarm is the 29-hour outage the redrives were built to end. So:

- the founder gets the **same** Slack line, now carrying the exact `clear-capout` command;
- both redrives print that command on every skip;
- a ledger write that fails degrades to a page that **says the record was not written**, rather than to silence.

Only one alarm fires per cap-out, deliberately. The redrives do not escalate a parked issue -- converge already paged 15 minutes earlier with the same unchanged fact, and a second page is the cry-wolf failure that trains the reader to skim.

## Verification

Reproducer written first, **observed RED at 4 passed / 9 failed** -- both selectors offered the capped issue, converge recorded nothing, the page named no clearing command.

Every skip case is paired with a byte-identical un-capped twin that must still be offered, and the capped PR is placed **first** on the board. A missing gate offers the capped issue; a wall (code that skips everything) offers nothing. Only a real gate passes.

```
test-converge-capout.sh        13/13   (was 4/13)
test-review-redrive.sh         30/30
test-review-redrive-absent.py  PASS
test-attempts-ledger.py        13/13
test-converge.sh               21/21
test-ci-redrive.sh             65/65
```

## Captured, not fixed here

- `sp-34cb0b96` -- converge exit-5 (no progress) and exit-7 (no verdict) are the same invisible-terminal shape; ASK-871 scoped only exit-2, so those two still leak.
- `sp-8aa4cbb9` -- `capout()` fails open on an unreadable ledger (`ledger_get` swallows read failures). Pre-existing direction for every budget in that file, documented in the function's own HONEST BOUNDARY.

Also voided two stale notes on `attempts-ledger.py` (`sp-029ef6f9`, `sp-626e9452`) -- both describe the pre-ASK-286 mkdir/token lock, which no longer exists in the file.

### DIFF
--- a/q-system/.q-system/capability-manifest.json
+++ b/q-system/.q-system/capability-manifest.json
@@ -101,6 +101,10 @@
       "timeout_s": 120
     },
     {
+      "path": "q-system/.q-system/scripts/test/test-converge-capout.sh",
+      "runner": "bash"
+    },
+    {
       "path": "q-system/.q-system/scripts/test/test-converge.sh",
       "runner": "bash"
     },
--- a/q-system/.q-system/scripts/attempts-ledger.py
+++ b/q-system/.q-system/scripts/attempts-ledger.py
@@ -289,6 +289,28 @@ def op_clear(d, issue, keys):
     return True
 
 
+def op_capout(d, issue, why, ts):
+    """Record that converge stopped this issue at its round cap.
+
+    KEYED PER ISSUE, and that is the whole point (ASK-871). Every other bound in
+    this loop keys on a PR and a head sha -- deliberately, so a PR that pushes a
+    real fix earns a fresh attempt. Nothing bounded DISPATCHES PER ISSUE, so on
+    2026-08-16 ASK-830 spent six converge rounds and five Opus reviews in one
+    morning: converge capped out at 15:59, the redrive handed the same issue back
+    at 16:14, and each round moved the head, so every per-sha cap read as fresh.
+
+    ALWAYS A WRITE, never a claim. `op_claim` answers False the second time, and
+    a caller reading that as "nothing to do" would leave a stale `capout_why`
+    from an older cap-out sitting next to a newer one. A cap-out that happens
+    twice is two facts, and the later one is the true one.
+    """
+    e = d.setdefault(issue, {})
+    e["capout"] = True
+    e["capout_at"] = ts
+    e["capout_why"] = why
+    return True
+
+
 def op_claim(d, issue, flag):
     """True the FIRST time this flag is claimed, False every time after."""
     e = d.setdefault(issue, {})
@@ -418,6 +440,26 @@ def _run(path, op, rest, ts):
         # so a caller does not have to know the current state to be correct.
         issue, flag = _args(op, rest, 2)
         _mutate(path, lambda d: op_clear(d, issue, (flag,)))
+        return 0
+
+    if op == "record-capout":             # record-capout <issue> <why>
+        # THE MACHINE-READABLE HALF OF A CAP-OUT (ASK-871). converge.sh's exit-2
+        # path used to `say` a log line and Slack the founder and stop, so the
+        # only two consumers that could re-enter the issue -- ci-redrive.py and
+        # review-redrive.py -- had no way to learn its rounds were spent. This
+        # is deliberately the SAME ledger those two already read rather than a
+        # new state file: a second store for one issue's budget is how the four
+        # budgets above ended up racing in the first place.
+        issue, why = _args(op, rest, 2)
+        _mutate(path, lambda d: op_capout(d, issue, why, ts))
+        return 0
+    if op == "clear-capout":              # clear-capout <issue>
+        # THE HUMAN'S WAY OUT, and it ships in the same change as the park on
+        # purpose. A cap-out nobody can clear is a permanent park, and a park
+        # with no exit is a quieter version of the 29-hour outage the redrives
+        # were built to end. The cap-out page names this command.
+        (issue,) = _args(op, rest, 1)
+        _mutate(path, lambda d: op_clear(d, issue, ("capout", "capout_at", "capout_why")))
         return 0
 
     if op == "claim-flag":                # claim-flag <issue> <flag> -> 0 first time, 1 after
--- a/q-system/.q-system/scripts/attempts-ledger.py
+++ b/q-system/.q-system/scripts/attempts-ledger.py
@@ -320,6 +320,42 @@ def op_claim(d, issue, flag):
     return True
 
 
+def op_claim_uncapped(d, issue, flag, outcome):
+    """`op_claim`, refusing outright if this issue is parked (codex on PR #210).
+
+    ONE TRANSACTION, and that is the entire reason this op exists rather than a
+    capout read in front of a claim-flag call. The redrives' cap-out gates sat in
+    the read-only OFFER; the atomic claim that actually spends the attempt asked
+    only whether the flag was free. kipi-dispatch.sh runs a ps snapshot, a
+    converge-live probe and a budget read between those two calls, so a cap-out
+    landing in that window authorized the very dispatch it was written to stop.
+
+    A caller-side check would only narrow the window: two subprocesses cannot be
+    atomic against a third writer. The question is answered here, under the same
+    flock that sets the flag, so there is no interval to lose the race in.
+
+    REFUSING WRITES NOTHING, which is load-bearing. Claiming first and reporting
+    the park afterwards would spend the issue's one attempt on a dispatch that
+    never happened, and `clear-capout` would then release an issue no redrive can
+    pick up -- a permanent stall wearing a cleared park's clothes.
+
+    `outcome` carries the verdict out because _mutate answers only bool(changed),
+    and "already claimed" and "refused, parked" are both False with different
+    exit codes owed to the caller.
+    """
+    e = d.setdefault(issue, {})
+    if e.get("capout"):
+        outcome["result"] = "capped"
+        outcome["why"] = e.get("capout_why") or "no reason recorded"
+        return False
+    if e.get(flag):
+        outcome["result"] = "already"
+        return False
+    e[flag] = True
+    outcome["result"] = "claimed"
+    return True
+
+
 def main(argv):
     if len(argv) < 3:
         print("usage: attempts-ledger.py <ledger-path> <op> [args...]", file=sys.stderr)
@@ -466,6 +502,24 @@ def _run(path, op, rest, ts):
         issue, flag = _args(op, rest, 2)
         claimed = _mutate(path, lambda d: op_claim(d, issue, flag))
         return 0 if claimed else 1
+
+    if op == "claim-flag-uncapped":       # -> 0 claimed, 1 already, 4 parked
+        # 4, and it has to be its OWN code. 1 already means "already claimed,
+        # stay quiet" for six call sites, and 2/3 already mean "nothing was
+        # written" -- a park is neither: nothing was written AND the caller must
+        # say so out loud, because a silently skipped dispatch is the 29-hour
+        # outage the redrives exist to end. See op_claim_uncapped for why the
+        # check cannot live in the caller.
+        issue, flag = _args(op, rest, 2)
+        outcome = {}
+        _mutate(path, lambda d: op_claim_uncapped(d, issue, flag, outcome))
+        if outcome.get("result") == "capped":
+            sys.stderr.write(
+                "attempts-ledger: `%s` refused for %s -- converge already gave up "
+                "on this issue (%s). Nothing was claimed.\n"
+                % (op, issue, outcome.get("why", "no reason recorded")))
+            return 4
+        return 0 if outcome.get("result") == "claimed" else 1
 
     print("unknown op: %s" % op, file=sys.stderr)
     return 2
--- a/q-system/.q-system/scripts/ci-redrive.py
+++ b/q-system/.q-system/scripts/ci-redrive.py
@@ -408,6 +408,23 @@ def _ledger_claim_rc(path, issue, flag):
     return proc.returncode
 
 
+def _ledger_claim_rc_uncapped(path, issue, flag):
+    """Raw rc of `claim-flag-uncapped`: 0 claimed, 1 already, 4 parked, 2/3 unwritten.
+
+    Separate from `_ledger_claim_rc` because the caller must distinguish 4 from
+    the rest; folding it into a bool here would put the park back in the same
+    bucket as "someone else owns it" and lose the operator line that names how to
+    un-park it.
+    """
+    proc = subprocess.run(
+        [sys.executable, LEDGER_SCRIPT, path, "claim-flag-uncapped", issue, flag],
+        capture_output=True, text=True)
+    if proc.returncode not in (0, 1, 4):
+        sys.stderr.write("ci-redrive: ledger refused `%s` for %s: %s\n"
+                         % (flag, issue, (proc.stderr or "").strip()[:200]))
+    return proc.returncode
+
+
 def ledger_claim(path, issue, flag):
     """True the first time this flag is claimed anywhere, False otherwise.
 
@@ -544,6 +561,47 @@ def capout_skip(path, issue, who, pr):
     return True
 
 
+def claim_unless_capped(path, issue, flag, who, pr=""):
+    """The atomic claim, refusing a parked issue in the SAME transaction.
+
+    True only when this run now owns the attempt. Replaces `ledger_claim` at both
+    mark-dispatched call sites (codex on PR #210, major): `capout_skip` guards the
+    read-only OFFER, and the claim that actually spends the attempt asked only
+    whether the flag was free. kipi-dispatch.sh does real work between the two --
+    a ps snapshot, a converge-live probe, a budget read -- so a cap-out landing in
+    that window authorized the dispatch it was written to stop.
+
+    Reading capout here, before calling claim-flag, would only shrink the window.
+    Two subprocesses cannot be atomic against a third writer, so the refusal is
+    delegated to the ledger's `claim-flag-uncapped`, which answers under the same
+    flock that sets the flag. rc 4 is the park; anything else non-zero is the
+    pre-existing "someone else owns it, or nothing was written" answer.
+
+    ONE DEFINITION, TWO CALLERS, like `capout_skip` and `is_reviewer_slot` above
+    and for the same reason: the two redrives must refuse the same parked issue.
+    """
+    rc = _ledger_claim_rc_uncapped(path, issue, flag)
+    if rc == 0:
+        return True
+    if rc == 4:
+        # LOUD, not quiet. A skipped dispatch nobody can see is the silent-park
+        # failure mode; the founder was already paged once by converge's exit-2,
+        # so this line is for whoever reads the dispatch log -- and it carries the
+        # clearing command because a park with no exit is the worse outage.
+        sys.stderr.write(
+            "%s: %s%s -- converge already gave up on this issue (%s). NOT "
+            "dispatching, and the attempt is NOT spent. Clear it with: %s\n"
+            % (who, issue, (" PR #%s" % pr) if pr else "",
+               capout(path, issue) or "no reason recorded",
+               CAPOUT_CLEAR % (path, issue)))
+        return False
+    sys.stderr.write(
+        "%s: the attempt for %s%s was already claimed (or the ledger could not be "
+        "written) -- not dispatching.\n"
+        % (who, issue, (" PR #%s" % pr) if pr else ""))
+    return False
+
+
 def cmd_redrive(cands):
     """READ-ONLY. Offers one pick; the dispatcher spends it via mark-dispatched."""
     path = attempts_path()
@@ -586,10 +644,10 @@ def cmd_redrive(cands):
 def cmd_mark_dispatched(issue, sig, head_sha):
     """The atomic claim. 0 = it is yours to dispatch, 1 = it is not."""
     path = attempts_path()
-    if not ledger_claim(path, issue, "ci_redrive_%s" % sig):
-        sys.stderr.write(
-            "ci-redrive: %s attempt for signature %s was already claimed (or the "
-            "ledger could not be written) -- not dispatching.\n" % (issue, sig))
+    # NOT `ledger_claim`. The park has to be refused inside the claim's own
+    # transaction, not read before it -- see claim_unless_capped.
+    if not claim_unless_capped(path, issue, "ci_redrive_%s" % sig,
+                               "ci-redrive [signature %s]" % sig):
         return 1
     # Order matters: record WHICH head first, then the marker saying a head was
     # recorded at all. The reverse order lets a failure between the two claim
--- a/q-system/.q-system/scripts/ci-redrive.py
+++ b/q-system/.q-system/scripts/ci-redrive.py
@@ -495,6 +495,55 @@ def attempts_path():
     return os.environ.get("KIPI_ATTEMPTS", DEFAULT_ATTEMPTS)
 
 
+# --- the cap-out park, read by BOTH redrives ---------------------------------
+# ONE DEFINITION, TWO CALLERS (ASK-871), for the same reason `is_reviewer_slot`
+# lives here: review-redrive.py imports this module rather than keeping its own
+# copy, and two hand-maintained copies of a refusal rule is how a state ends up
+# owned by both consumers or by neither.
+#
+# THE FACT IT READS: converge.sh stopped this ISSUE at its round cap and wrote it
+# to the ledger. Every other bound in this loop keys on a PR and a head sha,
+# which is deliberate -- a PR that pushes a real fix must earn a fresh attempt --
+# and it is exactly why nothing caught ASK-830 on 2026-08-16: each of its six
+# rounds moved the head, so every per-sha cap read as fresh while the ISSUE had
+# already been given up on. This is the missing axis, not a tightening of that one.
+
+CAPOUT_CLEAR = ("python3 q-system/.q-system/scripts/attempts-ledger.py "
+                "%s clear-capout %s")
+
+
+def capout(path, issue):
+    """The recorded reason converge gave up on this issue, or "" if it has not.
+
+    HONEST BOUNDARY: `ledger_get` answers "" both for "no cap-out" and for a
+    ledger it could not read at all, so a ledger that will not parse reads as
+    "not capped" and the redrive proceeds. That is the pre-existing direction of
+    every budget in this file (the `get` op swallows a read failure and returns
+    the default), not something this gate introduces -- but it means the silence
+    of this function is not proof the issue was never capped.
+    """
+    if not ledger_get(path, issue, "capout"):
+        return ""
+    return ledger_get(path, issue, "capout_why") or "no reason recorded"
+
+
+def capout_skip(path, issue, who, pr):
+    """True if this issue is parked; writes the operator line saying so.
+
+    Says how to UN-park it in the same breath. A refusal a human cannot reverse
+    is a permanent park, and the founder reading this line is the only thing
+    standing between a parked issue and the 29-hour outage.
+    """
+    why = capout(path, issue)
+    if not why:
+        return False
+    sys.stderr.write(
+        "%s: %s PR #%s -- converge already gave up on this issue (%s). Not "
+        "re-entering it until a human clears the cap-out: %s\n"
+        % (who, issue, pr, why, CAPOUT_CLEAR % (path, issue)))
+    return True
+
+
 def cmd_redrive(cands):
     """READ-ONLY. Offers one pick; the dispatcher spends it via mark-dispatched."""
     path = attempts_path()
@@ -510,6 +559,13 @@ def cmd_redrive(cands):
                 "ci-redrive: %s PR #%s is red but a converge for it is already "
                 "live -- not offering it and not escalating it this run\n"
                 % (cand["issue"], cand["pr"]))
+            continue
+        # AND BEFORE THE ESCALATION TOO, same reasoning as the gate above. A
+        # capped-out issue has ALREADY paged the founder from converge.sh's
+        # exit-2 path, with the clearing command in the line; escalating here
+        # would be a second alarm about one unchanged fact 15 minutes later,
+        # which is the cry-wolf failure that trains the reader to skim.
+        if capout_skip(path, cand["issue"], "ci-redrive", cand["pr"]):
             continue
         if ledger_get(path, cand["issue"], redrive_flag(cand)):
             escalate(path, cand)           # machine tier already spent on this
--- a/q-system/.q-system/scripts/converge.sh
+++ b/q-system/.q-system/scripts/converge.sh
@@ -964,6 +964,26 @@ while [ "$ROUND" -lt "$MAX_ROUNDS" ]; do
   say "round $ROUND -> $VERDICT (head $SHA); reworking"
 done
 
-say "STOP exit-2: hit the $MAX_ROUNDS-round cap still at '$LAST_VERDICT'. A cap-out means the reviewer and Sana disagree persistently; read the last review before raising the cap."
-bash "$NOTIFY" "converge $ISSUE: hit $MAX_ROUNDS-round cap, still $LAST_VERDICT" 2>/dev/null || true
+CAPOUT_WHY="hit the $MAX_ROUNDS-round cap still at '$LAST_VERDICT'${PR:+ on PR #$PR}"
+say "STOP exit-2: $CAPOUT_WHY. A cap-out means the reviewer and Sana disagree persistently; read the last review before raising the cap."
+# THE MACHINE-READABLE HALF, AND IT IS THE POINT OF ASK-871. Until this line the
+# cap-out announced itself to a human twice -- the `say` above and the page below
+# -- and to a machine not at all. So ci-redrive.py and review-redrive.py, whose
+# whole job is to re-enter a PR nobody else will, correctly-by-their-own-rules
+# re-entered this one: 2026-08-16, ASK-830 capped out at 15:59:53Z and was handed
+# back at 16:14:01Z, six rounds and five Opus reviews in one morning. Their caps
+# key per PR per head sha and every round moved the head, so nothing they read
+# was stale. Nothing bounded dispatches PER ISSUE. This record is that bound.
+#
+# WRITTEN BEFORE THE PAGE. The page names the command that clears the park, and a
+# page telling the founder to clear a record that was never written is worse than
+# no page. If the ledger refuses (a lock timeout, a read-only disk) the run says
+# so and still pages, because a cap-out the founder never hears about is the
+# 29-hour park with the alarm removed.
+CAPOUT_NOTE=""
+if ! python3 "$LEDGER" "$ATTEMPTS" record-capout "$ISSUE" "$CAPOUT_WHY" 2>>"$LOG"; then
+  CAPOUT_NOTE=" WARNING: the cap-out could NOT be recorded in $ATTEMPTS, so the redrives may re-enter this issue -- check that file."
+  say "WARNING: record-capout failed for $ISSUE; the redrives cannot see this cap-out."
+fi
+bash "$NOTIFY" "converge $ISSUE: hit $MAX_ROUNDS-round cap, still $LAST_VERDICT. Parked -- no redrive will re-enter it until you clear it: python3 q-system/.q-system/scripts/attempts-ledger.py $ATTEMPTS clear-capout $ISSUE$CAPOUT_NOTE" 2>/dev/null || true
 exit 2
--- a/q-system/.q-system/scripts/review-redrive.py
+++ b/q-system/.q-system/scripts/review-redrive.py
@@ -612,8 +612,14 @@ def cmd_select(cands, show_all):
 
 
 def cmd_mark_dispatched(issue, action, pr, head_sha):
+    # THE SAME SHARED CLAIM ci-redrive uses, not `ledger_claim` (codex on PR #210).
+    # `cmd_select`'s capout gate gates the OFFER; this is where the attempt is
+    # spent, and a cap-out recorded between the two used to sail straight through
+    # it. Imported rather than reimplemented for the reason `capout_skip` is: two
+    # copies of one refusal rule is how a state ends up owned by neither.
     cand = {"action": action, "pr": pr, "head_sha": head_sha}
-    return 0 if CI.ledger_claim(CI.attempts_path(), issue, flag(cand)) else 1
+    return 0 if CI.claim_unless_capped(
+        CI.attempts_path(), issue, flag(cand), "review-redrive", pr) else 1
 
 
 def main(argv):
--- a/q-system/.q-system/scripts/review-redrive.py
+++ b/q-system/.q-system/scripts/review-redrive.py
@@ -547,6 +547,18 @@ def cmd_select(cands, show_all):
 
     path = CI.attempts_path()
     for c in cands:
+        # THE CAP-OUT PARK, FIRST (ASK-871). converge.sh stopping this ISSUE at
+        # its round cap is a refusal to touch it at all, so it outranks every
+        # question below -- which action, is anything live, is the attempt spent.
+        #
+        # IMPORTED FROM ci-redrive, NOT REIMPLEMENTED, for the same reason
+        # `is_reviewer_slot` is: both redrives must refuse the same parked issue,
+        # and two copies of one refusal rule is how a state ends up owned by
+        # neither. It reads a fact, writes nothing, and cannot page -- so putting
+        # it ahead of the in-flight gates does not reintroduce the false-page
+        # hazard those gates exist for.
+        if CI.capout_skip(path, c["issue"], "review-redrive", c["pr"]):
+            continue
         # IN-FLIGHT GATES, BOTH ACTIONS, BEFORE THE LEDGER IS READ. Each action
         # launches a different long-running process, so each needs its own
         # liveness question -- but the reason is one reason, and the ledger read
--- a/q-system/.q-system/scripts/test/test-converge-capout.sh
+++ b/q-system/.q-system/scripts/test/test-converge-capout.sh
@@ -1,0 +1,255 @@
+#!/usr/bin/env bash
+# Reproducer + acceptance criteria for ASK-871: a converge cap-out is invisible
+# to the redrives, so a stuck issue restarts forever.
+#
+# THE DEFECT IT CLOSES, measured 2026-08-16. ASK-830 ran six converge rounds and
+# five Opus reviews in one morning and produced nothing mergeable:
+#
+#   15:59:53Z converge[ASK-830] STOP exit-2: hit the 3-round cap still at 'REQUEST CHANGES'
+#   16:14:01Z red-CI redrive: handing ASK-830 back to its agent ahead of the fresh pick
+#
+# Fourteen minutes between "I give up" and "here, do it again". Both bounding
+# mechanisms worked; they did not COMPOSE. converge.sh's cap-out announced itself
+# to a human only -- a `say` line and a Slack ping -- and wrote nothing a machine
+# reads, so the redrives had no way to learn the rounds were already spent.
+#
+# THE PROPERTY UNDER TEST IS THE COMPOSITION, and it is asserted in both
+# directions on purpose. A gate that skips a capped issue is satisfied by a WALL
+# -- code that skips everything -- and a wall converts today's runaway into
+# silent parking, which is the strictly worse failure the issue's blast-radius
+# note names. So every skip case is paired with a byte-identical case whose only
+# difference is the cap-out record, and the un-capped twin must still be offered.
+#
+# BOTH REDRIVES, not just the one the DoR named. The 16:14:01Z line above is
+# kipi-dispatch.sh:1141, which is the ci-redrive.py call site -- so gating only
+# review-redrive.py would leave the measured incident live. Ci-redrive's
+# exclusion of the reviewer verdict slots is untouched (that is the ASK-871
+# not-doing list, and PR #73 is its scar); what is added is one more reason to
+# refuse, read from the same ledger.
+#
+# Isolation: KIPI_STATE_DIR, KIPI_ATTEMPTS, KIPI_CONVERGE_WORKER, KIPI_NOTIFY,
+# KIPI_GH and KIPI_PS all point into a mktemp dir. No real worker, no real gh,
+# no real process table, no live Linear, no Slack.
+set -uo pipefail
+
+ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../../../.." && pwd)"
+CONV="$ROOT/q-system/.q-system/scripts/converge.sh"
+LEDGER="$ROOT/q-system/.q-system/scripts/attempts-ledger.py"
+REVIEW_REDRIVE="$ROOT/q-system/.q-system/scripts/review-redrive.py"
+CI_REDRIVE="$ROOT/q-system/.q-system/scripts/ci-redrive.py"
+for f in "$CONV" "$LEDGER" "$REVIEW_REDRIVE" "$CI_REDRIVE"; do
+  [ -f "$f" ] || { echo "FATAL: missing $f" >&2; exit 1; }
+done
+
+WORK="$(mktemp -d)"
+trap 'rm -rf "$WORK"' EXIT
+PASS=0; FAIL=0
+ok()  { PASS=$((PASS+1)); echo "  ok   - $1"; }
+bad() { FAIL=$((FAIL+1)); echo "  FAIL - $1"; }
+
+mkdir -p "$WORK/bin" "$WORK/state/pr-reviews" "$WORK/records"
+ATTEMPTS="$WORK/state/linear-worker-attempts.json"
+PAGES="$WORK/pages.txt"; : > "$PAGES"
+
+# The notify sink, stubbed for EVERY case. converge.sh pages on cap-out and both
+# redrives escalate from inside their select paths; with no stub that is the real
+# slack-notify.sh, quiet only because no webhook resolves on this machine.
+# Quiet-because-unconfigured is not isolation.
+cat > "$WORK/bin/notify.sh" <<EOS
+#!/usr/bin/env bash
+printf '%s\n' "\$*" >> "$PAGES"
+EOS
+chmod +x "$WORK/bin/notify.sh"
+
+capout_of() {   # capout_of <issue> <key>
+  python3 "$LEDGER" "$ATTEMPTS" get "$1" "$2" "" 2>/dev/null
+}
+
+echo "== converge cap-out is recorded where a machine reads it =="
+
+# --- the converge half: same fake-worker harness test-converge.sh uses --------
+cat > "$WORK/bin/gh" <<'EOS'
+#!/usr/bin/env bash
+case "${1:-} ${2:-}" in
+  "pr list") cat "$FAKE_PR_FILE" 2>/dev/null ;;
+  "pr view") cat "$FAKE_SHA_FILE" 2>/dev/null ;;
+esac
+exit 0
+EOS
+chmod +x "$WORK/bin/gh"
+
+cat > "$WORK/bin/fakeworker" <<'EOS'
+#!/usr/bin/env bash
+N=$(( $(cat "$FAKE_ROUND_FILE" 2>/dev/null || echo 0) + 1 ))
+echo "$N" > "$FAKE_ROUND_FILE"
+ENTRY="$(echo "$FAKE_SEQ" | cut -d'|' -f"$N")"
+[ -n "$ENTRY" ] || exit 0
+V="${ENTRY%%;*}"; S="${ENTRY##*;}"
+echo "$S" > "$FAKE_SHA_FILE"
+PR="$(cat "$FAKE_PR_FILE")"
+python3 -c "
+import json,sys
+json.dump({'pr':int(sys.argv[1]),'issue':sys.argv[4],'verdict':sys.argv[2],
+           'review':'x','ts':'t'}, open(sys.argv[3],'w'))
+" "$PR" "$V" "$FAKE_STATE/pr-reviews/pr-$PR.verdict.json" "$FAKE_ISSUE"
+exit 0
+EOS
+chmod +x "$WORK/bin/fakeworker"
+
+run_converge() {   # run_converge <issue> <pr> <seq> <max-rounds>
+  echo "$2" > "$WORK/pr"; echo "0" > "$WORK/round"; : > "$WORK/sha"
+  rm -f "$WORK/state/pr-reviews"/*.verdict.json
+  env PATH="$WORK/bin:$PATH" \
+      KIPI_STATE_DIR="$WORK/state" \
+      KIPI_CONVERGE_WORKER="$WORK/bin/fakeworker" \
+      KIPI_NOTIFY="$WORK/bin/notify.sh" \
+      FAKE_STATE="$WORK/state" FAKE_ISSUE="$1" \
+      FAKE_PR_FILE="$WORK/pr" FAKE_SHA_FILE="$WORK/sha" FAKE_ROUND_FILE="$WORK/round" \
+      FAKE_SEQ="$3" \
+    bash "$CONV" --issue "$1" --max-rounds "$4" > "$WORK/conv.out" 2>&1
+  echo $?
+}
+
+RC="$(run_converge ASK-9871 301 "REQUEST CHANGES;s1|REQUEST CHANGES;s2|REQUEST CHANGES;s3|REQUEST CHANGES;s4" 3)"
+[ "$RC" = "2" ] && ok "a never-approving reviewer still hits the round cap (exit 2)" \
+  || bad "cap-out did not exit 2, got rc=$RC: $(cat "$WORK/conv.out")"
+
+[ "$(capout_of ASK-9871 capout)" = "True" ] \
+  && ok "the cap-out is written to the attempts ledger, where a machine reads it" \
+  || bad "no cap-out record for ASK-9871 after exit 2 -- the redrives cannot see it"
+
+CAPOUT_WHY="$(capout_of ASK-9871 capout_why)"
+case "$CAPOUT_WHY" in
+  *"3-round cap"*) ok "the record says WHY: $CAPOUT_WHY" ;;
+  *) bad "capout_why does not name the round cap: '$CAPOUT_WHY'" ;;
+esac
+[ -n "$(capout_of ASK-9871 capout_at)" ] \
+  && ok "the record is timestamped" || bad "capout_at is empty"
+
+# The founder-facing half. The Slack line is the ONE alarm for this fact, and a
+# park a human cannot clear is a permanent park -- so the page has to carry the
+# clearing command, not just the news.
+grep -q "clear-capout" "$PAGES" \
+  && ok "the cap-out page names the command that clears the park" \
+  || bad "the page does not tell the founder how to un-park it: $(cat "$PAGES")"
+
+# THE PAIRED NEGATIVE. Without it, "always record a cap-out" passes every assert
+# above and parks every converged issue on the board.
+RC="$(run_converge ASK-9872 302 "APPROVE;s1" 3)"
+[ "$RC" = "1" ] && [ -z "$(capout_of ASK-9872 capout)" ] \
+  && ok "a converge that reaches its goal records NO cap-out" \
+  || bad "APPROVE run (rc=$RC) left capout='$(capout_of ASK-9872 capout)' -- it parks healthy issues"
+
+echo "== the redrives refuse an issue whose cap-out is uncleared =="
+
+# --- the board, stubbed at the gh seam ---------------------------------------
+cat > "$WORK/bin/gh-board" <<'EOS'
+#!/usr/bin/env bash
+cat "$BOARD"
+EOS
+chmod +x "$WORK/bin/gh-board"
+
+# Two PRs identical in every field that either selector reads, except the issue
+# id -- which is the only thing the ledger is keyed on.
+reviewer_pr() {   # reviewer_pr <number> <issue-lower> <sha>
+  cat <<EOS
+{"number": $1, "headRefName": "sana/$2", "headRefOid": "$3",
+ "url": "https://example.invalid/pr/$1", "title": "work ($(echo "$2" | tr a-z A-Z))",
+ "isDraft": false,
+ "statusCheckRollup": [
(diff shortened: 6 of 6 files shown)


assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-210-20260817-095136.md` (359237 bytes).

```
FINDINGS:
major|Both atomic dispatch entrypoints authorize work after a concurrent cap-out because they claim the attempt without checking capout in the same ledger transaction|q-system/.q-system/scripts/ci-redrive.py:586,q-system/.q-system/scripts/review-redrive.py:614
END FINDINGS
```

--- reviewer output, last 54717 bytes of 359237 (full review at the path above) ---

  1170	REVIEW_ACTION=""; REVIEW_NEXT=""; REVIEW_PR=""; REVIEW_SHA=""
  1171	if [ -z "$REDRIVE_NEXT" ] && [ -f "$REVIEW_REDRIVE" ]; then
  1172	  REVIEW_LINE="$(python3 "$REVIEW_REDRIVE" --repo-dir "$TARGET_PATH" select 2>>"$LOG")"
  1173	  REVIEW_RC=$?
  1174	  if [ "$REVIEW_RC" = "0" ] && [ -n "$REVIEW_LINE" ]; then
  1175	    REVIEW_ACTION="$(printf '%s' "$REVIEW_LINE" | cut -f1)"
  1176	    REVIEW_NEXT="$(printf '%s' "$REVIEW_LINE" | cut -f2)"
  1177	    REVIEW_PR="$(printf '%s' "$REVIEW_LINE" | cut -f3)"
  1178	    REVIEW_SHA="$(printf '%s' "$REVIEW_LINE" | cut -f4)"
  1179	    say "reviewer redrive: $REVIEW_NEXT PR #$REVIEW_PR needs $REVIEW_ACTION${NEXT:+ ($NEXT waits)}"
  1180	    NEXT="$REVIEW_NEXT"
  1181	  elif [ "$REVIEW_RC" = "2" ]; then
  1182	    say "reviewer redrive: gh could not read PR state in $TARGET_NAME -- fresh pick stands, nothing claimed"
  1183	  fi
  1184	fi
  1185	
  1186	if [ -z "$NEXT" ]; then
  1187	  say "nothing ready ($(printf '%s' "$WORK_OUT" | grep -oE '[0-9]+ ready issue' | head -1))"
  1188	  exit 0
  1189	fi
  1190	
  1191	# Belt and braces against the race between dispatch and the In Progress
  1192	# transition: two converge runs on one issue would fight over one worktree.
  1193	#
  1194	# NOT pgrep, and NOT \b (PR #39 review, finding 2). BSD pgrep reads `\b` as a
  1195	# literal `b`, so this guard has never fired on macOS -- the only platform it
  1196	# runs on. It was harmless while every dispatched child was being reaped
  1197	# instantly; the moment children survive (the fix below), it becomes reachable
  1198	# and lets a second converge start on an issue that already has one. Same
  1199	# `ps -Ao args=` form and same [c] self-match guard as the liveness check.
  1200	# NO PIPE INTO grep -q, and that is the whole point (PR #39 review r3,
  1201	# finding 1). `ps ... | grep -q` under `set -o pipefail` fires only sometimes:
  1202	# grep -q exits the instant it matches, ps then takes SIGPIPE and dies 141, and
  1203	# pipefail makes 141 the status of the whole pipeline -- so the `if` does NOT
  1204	# run its body. Whether ps has finished writing before grep leaves is a race,
  1205	# so the guard worked load-dependently, which is worse than never working
  1206	# because it looks fine when you test it by hand.
  1207	#
  1208	# A snapshot into a variable plus bash's own =~ removes the pipeline entirely,
  1209	# so there is nothing to SIGPIPE and nothing for pipefail to poison. It also
  1210	# removes the need for the [c] self-match trick: with no grep process there is
  1211	# no grep command line in the table to match.
  1212	PS_SNAPSHOT="$(ps -Ao args= 2>/dev/null || true)"
  1213	if [[ "$PS_SNAPSHOT" =~ converge\.sh\ --issue\ ${NEXT}([[:space:]]|$) ]]; then
  1214	  say "skip $NEXT: a converge run for it is already live"
  1215	  exit 0
  1216	fi
  1217	
  1218	# --- RED-CI REDRIVE: MARK-DISPATCHED (ASK-295) -------------------------------
  1219	# Past every guard that can still abort, and immediately before the launch. This
  1220	# is where the PR's one machine attempt is spent, and the call is the atomic
  1221	# gate: rc 0 means this run owns the attempt, rc non-zero means another run
  1222	# already claimed it (or the ledger could not be written, which is the same
  1223	# answer -- nothing was recorded, so nothing may act as though it was).
  1224	if [ -n "$REDRIVE_SIG" ] && [ "$NEXT" = "$REDRIVE_NEXT" ]; then
  1225	  if ! python3 "$REDRIVE" mark-dispatched --issue "$NEXT" \
  1226	       --signature "$REDRIVE_SIG" --head-sha "$REDRIVE_SHA" 2>>"$LOG"; then
  1227	    say "red-CI redrive: the attempt for $NEXT is already claimed -- not dispatching"
  1228	    exit 0
  1229	  fi
  1230	fi
  1231	
  1232	# Same contract for the reviewer redrive (ASK-352): the OFFER above wrote
  1233	# nothing, and THIS is the atomic claim. rc non-zero means another run owns it,
  1234	# or the ledger could not be written -- the same answer either way, because
  1235	# nothing was recorded so nothing may act as though it was. The cap is one
  1236	# attempt per PR per action per HEAD SHA, so a PR that pushes a real fix is
  1237	# eligible again while a re-review that keeps coming back phantom at the same sha
  1238	# is not retried.
  1239	if [ -n "$REVIEW_ACTION" ] && [ "$NEXT" = "$REVIEW_NEXT" ]; then
  1240	  if ! python3 "$REVIEW_REDRIVE" mark-dispatched --issue "$NEXT" \
  1241	       --action "$REVIEW_ACTION" --pr "$REVIEW_PR" --head-sha "$REVIEW_SHA" 2>>"$LOG"; then
  1242	    say "reviewer redrive: the $REVIEW_ACTION attempt for $NEXT PR #$REVIEW_PR is already claimed -- not dispatching"
  1243	    exit 0
  1244	  fi
  1245	fi
  1246	
  1247	# Count BEFORE launching. Counting after would let a crash between the two
  1248	# hand out a free dispatch every heartbeat -- the budget must fail closed.
  1249	printf '%s' "$((DISPATCHED_TODAY + 1))" > "$COUNT_FILE"
  1250	
  1251	say "${LANE_TAG}dispatching $NEXT (live=$LIVE cap=$MAX_CONCURRENT rounds=$MAX_ROUNDS budget=$((DISPATCHED_TODAY + 1))/$DAILY_MAX lane=$DISPATCH_LANE)"
  1252	
  1253	# THE CHILD NEEDS ITS OWN SESSION, AND THIS IS NOT A STYLE CHOICE.
  1254	#
  1255	# This was `nohup ... & disown`, which is correct in an interactive shell and
  1256	# WRONG under launchd. launchd reaps the job's whole process group when the main
  1257	# process exits; nohup only blocks SIGHUP, so the converge was killed the instant
  1258	# this script returned. Every launchd dispatch since the dispatcher was installed
  1259	# died that way, and the failure was invisible by construction: the log file is
  1260	# created by the redirect before the child dies, so it exists and is 0 bytes,
  1261	# `say "dispatched $NEXT"` still runs, and the budget counter is already spent.
  1262	# The loop reported four healthy dispatches of ASK-224 on 2026-07-28 and did no
  1263	# work at all -- it only spent the subscription.
  1264	#
  1265	# PROVEN, not reasoned about. A launchd job whose only act was
  1266	# `nohup bash -c "sleep 25; touch F" & disown; exit`:
  1267	#     under launchd   F never written  (child killed)
  1268	#     same script from an interactive shell   F written (child survived)
  1269	# and with the setsid form below, under launchd, F is written.
  1270	#
  1271	# macOS ships no setsid(1), so python3 is how setsid(2) gets called. A new
  1272	# session means a new process group with no controlling terminal, which is
  1273	# outside the group launchd tears down.
  1274	CONVERGE_LOG="$HOME/.config/kipi/converge-$NEXT.log"
  1275	# A RUN BOUNDARY, because the log is appended (PR #39 review, finding 3). The
  1276	# failure page points the operator at this file; without a marker they cannot
  1277	# tell where a re-dispatch's output starts and are reading the previous run's
  1278	# tail as if it were this one's.
  1279	printf '\n===== dispatch %s  %s  rounds=%s =====\n' \
  1280	  "$NEXT" "$(date -u +%Y-%m-%dT%H:%M:%SZ)" "$MAX_ROUNDS" >> "$CONVERGE_LOG"
  1281	
  1282	# WHAT THIS DISPATCH ACTUALLY RUNS (ASK-352). Converge for everything except a
  1283	# reviewer redrive that asked for a RE-REVIEW, where the right action is another
  1284	# review and not another rework round: the PR is parked because nobody read the
  1285	# code, so there are no findings for a converge to work from and it would come
  1286	# back to this same failing slot having spent a full round.
  1287	#
  1288	# --post is required and is the point. Without it the reviewer writes a verdict
  1289	# record and never touches the commit status, so the slot that parked the PR
  1290	# stays failing and the next heartbeat picks the same PR again -- a loop that
  1291	# looks like progress. --invoker labels it dispatcher-driven (sp-53aad86f), so a
  1292	# hand-run review can never pass as evidence the loop closed itself.
  1293	#
  1294	# ONE ARRAY, EXPANDED ONCE BELOW. The launch machinery underneath (setsid, the
  1295	# pid capture, the liveness assert) is proven and is not duplicated per action;
  1296	# only the argv differs.
  1297	DISPATCH_ARGV=(./kipi converge --issue "$NEXT" --max-rounds "$MAX_ROUNDS")
  1298	if [ "$REVIEW_ACTION" = "re-review" ] && [ "$NEXT" = "$REVIEW_NEXT" ]; then
  1299	  DISPATCH_ARGV=(bash "$REPO/q-system/.q-system/scripts/pr-review-agent.sh" \
  1300	                 "$REVIEW_PR" --issue "$NEXT" --post)
  1301	  export KIPI_REVIEW_INVOKER="dispatcher"
  1302	fi
  1303	
  1304	CHILD_PID="$(python3 - "$CONVERGE_LOG" \
  1305	         "${DISPATCH_ARGV[@]}" <<'PY'
  1306	import subprocess, sys
  1307	log_path, argv = sys.argv[1], sys.argv[2:]
  1308	# Append, never truncate: a re-dispatch of the same issue must not erase the
  1309	# evidence of the previous run (the burst incident truncated a live log with >).
  1310	log = open(log_path, "ab", buffering=0)
  1311	p = subprocess.Popen(argv, stdout=log, stderr=log, start_new_session=True)
  1312	# The PID is the whole point: the caller has to watch THE CHILD IT LAUNCHED,
  1313	# not "some converge for this issue". `kipi` runs converge.sh with bash rather
  1314	# than exec, so this pid stays alive exactly as long as the run does.
  1315	print(p.pid)
   320	        return REREVIEW, "reviewed %s but head is now %s" % (
   321	            reviewed_sha[:8], head_sha[:8])
   322	
   323	    verdict = (record.get("verdict") or "").strip()
   324	    if verdict in REWORK_VERDICTS:
   325	        return REWORK, "reviewer said %s at the current head" % verdict
   326	    if not verdict:
   327	        # A usable review that stated nothing. Rare, and it is still not a spec:
   328	        # there are no findings to work from, so the answer is another review,
   329	        # never a rework round guessing at what to change.
   330	        return REREVIEW, "the review is usable but states no verdict"
   331	    return None, "verdict %s is not a refusal" % verdict
   332	
   333	
   334	def candidates(repo_dir, records_dir):
   335	    out = []
   336	    for pr_obj in CI.list_prs(repo_dir):
   337	        # A DRAFT IS THE AUTHOR SAYING NOT YET, and re-entering one spends a
   338	        # converge round or a codex call on a tree nobody asked to be judged.
   339	        #
   340	        # THE DOCSTRING CLAIMED THIS AND THE CODE DID NOT DO IT (codex round 2 on
   341	        # PR #91, minor). It also claimed "same rule as ci-redrive", and
   342	        # ci-redrive does not filter drafts either -- it fetches `isDraft` in
   343	        # PR_FIELDS and never reads it. So the comment was wrong twice, which is
   344	        # the exact defect class this whole selector exists because of: ci-redrive
   345	        # excluded the reviewer slots on the strength of a sentence asserting
   346	        # something no code did. Captured separately for ci-redrive's own copy.
   347	        if pr_obj.get("isDraft"):
   348	            continue
   349	        attributed = CI.attribute(pr_obj)
   350	        if attributed is None:
   351	            continue
   352	        issue, agent, source = attributed
   353	        slots = reviewer_slot_failing(pr_obj)
   354	        pr = pr_obj.get("number")
   355	        record = read_record(records_dir, pr)
   356	        if not slots:
   357	            if reviewer_slot_posted(pr_obj):
   358	                # A gating verdict exists and is not failing. Nothing to redrive.
   359	                continue
   360	            if record is CORRUPT_RECORD:
   361	                # Say what is true: the record is there and unreadable. Naming it
   362	                # "never posted" would send the operator after a producer bug.
   363	                sys.stderr.write(
   364	                    "review-redrive: PR #%s has a verdict record that will not "
   365	                    "parse -- refusing to call it absent. Left alone.\n" % pr)
   366	                continue
   367	            if record is not None:
   368	                # NO STATUS, BUT A VERDICT RECORD EXISTS. Not a virgin PR.
   369	                #
   370	                # pr-review-agent.sh writes the record unconditionally but posts
   371	                # the status only inside `if [ "$POST" = "1" ]` (:917, :975). So
   372	                # `kipi review <PR>` without --post, the ":915 recorded but NO
   373	                # gate moved" branch, and the missing-head-sha branch all leave
   374	                # this shape, in the very directory read_record reads (:111).
   375	                #
   376	                # classify() answers "is this a refusal", which is the WRONG
   377	                # question here. It returns (None, ...) for any non-refusal, and
   378	                # the shared tail below does `if action is None: continue` -- so
   379	                # a review that PASSED but was never posted fell out silently.
   380	                # Sampled 2026-08-06 across 96 records: APPROVE WITH NITS 54,
   381	                # REQUEST CHANGES 23, APPROVE 17, BLOCK 1, empty 1. The dropped
   382	                # non-refusals are 71 of 96, the MAJORITY, and PR #23 is one --
   383	                # approved, never posted, auto-merge armed and waiting forever on
   384	                # a status nobody sends. The cheapest PR in the queue was the one
   385	                # the selector could not see. (Independent probe on PR #123.)
   386	                head_sha = pr_obj.get("headRefOid") or ""
   387	                action, reason = classify(record, head_sha)
   388	                if action is None:
   389	                    # A PASSING review with no status. Not a refusal, not virgin.
   390	                    action = REREVIEW
   391	                    reason = ("the review passed but its status was never "
   392	                              "posted (%s)" % (record.get("verdict") or "no verdict"))
   393	                out.append({
   394	                    "action": action, "reason": reason,
   395	                    "issue": issue, "agent": agent, "issue_source": source,
   396	                    "pr": pr, "url": pr_obj.get("url"),
   397	                    "branch": pr_obj.get("headRefName"),
   398	                    "head_sha": head_sha, "slots": [],
   399	                })
   400	                continue
   401	            else:
   402	                # NEVER POSTED and never recorded -- the reviewer has not run
   403	                # against this PR at all.
   404	                #
   405	                # NOT the ASK-318 case below. There, a FAILING slot with no record
   406	                # means the producer ran and recorded nothing, so re-reviewing
   407	                # papers over a broken producer. Here nothing ever claimed a
   408	                # review happened, so there is no verdict to be missing.
   409	                #
   410	                # It emits the EXISTING re-review action on purpose.
   411	                # kipi-dispatch.sh:1000 matches that exact string to run
   412	                # pr-review-agent.sh --post; any other value falls through to
   413	                # `./kipi converge`, so a new vocabulary word would not be inert,
   414	                # it would silently buy a full rework round against a review that
   415	                # does not exist. The word already means "nobody read the code,
   416	                # there is no spec", which is precisely this state. The
   417	                # distinction rides in the reason, which select prints.
   418	                out.append({
   419	                    "action": REREVIEW,
   420	                    "reason": "the reviewer has never posted a verdict for this PR",
   421	                    "issue": issue, "agent": agent, "issue_source": source,
   422	                    "pr": pr, "url": pr_obj.get("url"),
   423	                    "branch": pr_obj.get("headRefName"),
   424	                    "head_sha": pr_obj.get("headRefOid") or "",
   425	                    "slots": [],
   426	                })
   427	                continue
   428	        elif record is CORRUPT_RECORD:
   429	            sys.stderr.write(
   430	                "review-redrive: PR #%s has %s failing and an unreadable verdict "
   431	                "record -- refusing to guess. Left alone.\n" % (pr, ",".join(slots)))
   432	            continue
   433	        elif record is None:
   434	            # ABSENT, not FAILURE-with-no-record. Owned by ASK-318/ASK-313.
   435	            sys.stderr.write(
   436	                "review-redrive: PR #%s has %s failing but NO verdict record -- "
   437	                "that is the absent-producer case (ASK-318), not this one. "
   438	                "Left alone.\n" % (pr, ",".join(slots)))
   439	            continue
   440	        head_sha = pr_obj.get("headRefOid") or ""
   441	        action, reason = classify(record, head_sha)
   442	        if action is None:
   443	            continue
   444	        out.append({
   445	            "action": action, "reason": reason, "issue": issue, "agent": agent,
   446	            "issue_source": source, "pr": pr, "url": pr_obj.get("url"),
   447	            "branch": pr_obj.get("headRefName"), "head_sha": head_sha,
   448	            "slots": slots,
   449	        })
   450	    return out
   451	
   452	
   453	def flag(cand):
   454	    """One attempt per PR per action per head sha. See the module docstring."""
   455	    return "review_%s_pr%s_%s" % (cand["action"].replace("-", ""),
   456	                                  cand["pr"], (cand["head_sha"] or "nohead")[:12])
   457	
   458	
   459	def reviewer_live(pr):
   460	    """True if pr-review-agent.sh is reviewing THIS PR in the process table now.
   461	
   462	    THE SYMMETRIC GATE TO converge_live, AND IT WAS MISSING (codex round 4 on
   463	    PR #91, major). `mark-dispatched` claims the attempt flag the instant a
   464	    re-review is handed out; pr-review-agent.sh then runs for 3-8 minutes. Every
   465	    heartbeat inside that window read the flag as set, called the one attempt
kipi-dispatch.sh:1225:  if ! python3 "$REDRIVE" mark-dispatched --issue "$NEXT" \
kipi-dispatch.sh:1240:  if ! python3 "$REVIEW_REDRIVE" mark-dispatched --issue "$NEXT" \
q-system/.q-system/scripts/test/test-ci-redrive.sh:229:run mark-dispatched --issue ASK-288 --signature "$SIG" \
q-system/.q-system/scripts/test/test-ci-redrive.sh:231:check "mark-dispatched claims the attempt" "$RC" "0"
q-system/.q-system/scripts/test/test-ci-redrive.sh:232:run mark-dispatched --issue ASK-288 --signature "$SIG" \
q-system/.q-system/scripts/test/test-ci-redrive.sh:234:check "mark-dispatched is one-shot: a second caller must not dispatch" "$RC" "1"
q-system/.q-system/scripts/test/test-ci-redrive.sh:257:run mark-dispatched --issue ASK-288 --signature "$SIG" \
q-system/.q-system/scripts/test/test-ci-redrive.sh:376:run mark-dispatched --issue ASK-288 --signature "$SIG" \
q-system/.q-system/scripts/test/test-ci-redrive.sh:418:# whether mark-dispatched actually gates the launch, or whether an rc 2 leaves
q-system/.q-system/scripts/test/test-converge-capout.sh:64:capout_of() {   # capout_of <issue> <key>
q-system/.q-system/scripts/test/test-converge-capout.sh:117:[ "$(capout_of ASK-9871 capout)" = "True" ] \
q-system/.q-system/scripts/test/test-converge-capout.sh:121:CAPOUT_WHY="$(capout_of ASK-9871 capout_why)"
q-system/.q-system/scripts/test/test-converge-capout.sh:124:  *) bad "capout_why does not name the round cap: '$CAPOUT_WHY'" ;;
q-system/.q-system/scripts/test/test-converge-capout.sh:126:[ -n "$(capout_of ASK-9871 capout_at)" ] \
q-system/.q-system/scripts/test/test-converge-capout.sh:127:  && ok "the record is timestamped" || bad "capout_at is empty"
q-system/.q-system/scripts/test/test-converge-capout.sh:132:grep -q "clear-capout" "$PAGES" \
q-system/.q-system/scripts/test/test-converge-capout.sh:139:[ "$RC" = "1" ] && [ -z "$(capout_of ASK-9872 capout)" ] \
q-system/.q-system/scripts/test/test-converge-capout.sh:141:  || bad "APPROVE run (rc=$RC) left capout='$(capout_of ASK-9872 capout)' -- it parks healthy issues"
q-system/.q-system/scripts/test/test-converge-capout.sh:209:python3 "$LEDGER" "$ATTEMPTS" record-capout ASK-8710 \
q-system/.q-system/scripts/test/test-converge-capout.sh:216:grep -q "ASK-8710" "$WORK/rr.err" && grep -q "clear-capout" "$WORK/rr.err" \
q-system/.q-system/scripts/test/test-converge-capout.sh:223:python3 "$LEDGER" "$ATTEMPTS" record-capout ASK-8712 \
q-system/.q-system/scripts/test/test-converge-capout.sh:230:grep -q "ASK-8712" "$WORK/ci.err" && grep -q "clear-capout" "$WORK/ci.err" \
q-system/.q-system/scripts/test/test-converge-capout.sh:237:python3 "$LEDGER" "$ATTEMPTS" clear-capout ASK-8710 >/dev/null 2>&1
q-system/.q-system/scripts/test/test-converge-capout.sh:238:[ -z "$(capout_of ASK-8710 capout)" ] \
q-system/.q-system/scripts/test/test-converge-capout.sh:239:  && ok "clear-capout removes the record" \
q-system/.q-system/scripts/test/test-converge-capout.sh:240:  || bad "capout survived clear-capout: '$(capout_of ASK-8710 capout)'"
q-system/.q-system/scripts/test/test-converge-capout.sh:252:  || bad "clear-capout ASK-8710 also released ASK-8712"
q-system/.q-system/scripts/test/test-review-redrive.sh:232:env KIPI_ATTEMPTS="$LEDGER" python3 "$SEL" mark-dispatched \
q-system/.q-system/scripts/test/test-review-redrive.sh:234:[ -z "$(select_one)" ] && ok "after mark-dispatched the pick is suppressed -- the cap is real" \
q-system/.q-system/scripts/test/test-review-redrive.sh:235:  || bad "the pick survived mark-dispatched -- the cap never fires"
q-system/.q-system/scripts/test/test-review-redrive.sh:302:# Codex found this on PR #91 (round 4, major). `mark-dispatched` claims the
q-system/.q-system/scripts/test/test-review-redrive.sh:357:env KIPI_ATTEMPTS="$LEDGER" python3 "$SEL" mark-dispatched \

exec
/bin/zsh -lc "printf '%s\\n' \""'${TMPDIR:-/tmp}"' in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-210
 succeeded in 0ms:
/var/folders/th/ty0kb7r56dvfrkgkrcbx9fv00000gn/T/

2026-08-17T16:53:31.118671Z ERROR codex_core::tools::router: error=patch rejected: writing is blocked by read-only sandbox; rejected by user approval settings
codex
The read-only sandbox rejected creating `/var/folders/.../T/pr210-capout-race.sh`, so I cannot persist the requested repro file. I’m running the same runnable sequence inline against a fresh `$TMPDIR` ledger. It still exercises the real `record-capout` and both real `mark-dispatched` entrypoints.
exec
/bin/zsh -lc 'repro_dir="$(mktemp -d "${TMPDIR:-/tmp}/pr210-race.XXXXXX")"
ledger="$repro_dir/attempts.json"
ledger_script="$PWD/q-system/.q-system/scripts/attempts-ledger.py"
ci="$PWD/q-system/.q-system/scripts/ci-redrive.py"
review="$PWD/q-system/.q-system/scripts/review-redrive.py"
printf '"'%s\\n' 'Simulating: selector offered work, then a concurrent converge recorded cap-out.'
python3 \""'$ledger_script" "$ledger" record-capout ASK-871 "concurrent converge hit cap"
KIPI_ATTEMPTS="$ledger" python3 "$ci" mark-dispatched --issue ASK-871 --signature deadbeef1234 --head-sha abc123
ci_rc=$?
KIPI_ATTEMPTS="$ledger" python3 "$review" mark-dispatched --issue ASK-871 --action re-work --pr 210 --head-sha abc123
review_rc=$?
printf '"'ci mark-dispatched rc=%s (0 means dispatcher launches converge)\\n' \""'$ci_rc"
printf '"'review mark-dispatched rc=%s (0 means dispatcher launches rework)\\n' \""'$review_rc"
printf '"'%s\\n' 'ledger after both claims:'
python3 -m json.tool \""'$ledger"
if [ "$ci_rc" -eq 0 ] || [ "$review_rc" -eq 0 ]; then
  printf '"'%s\\n' 'BUG: a dispatch was authorized after the issue was capped.'
  exit 1
fi
printf '%s\\n' 'PASS: cap-out blocked both atomic dispatch claims.'" in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-210
 succeeded in 67ms:
usage: review-redrive.py mark-dispatched [-h] --issue ISSUE
                                         --action {rework,re-review} --pr PR
                                         [--head-sha HEAD_SHA]
review-redrive.py mark-dispatched: error: argument --action: invalid choice: 're-work' (choose from 'rework', 're-review')
ci mark-dispatched rc=1 (0 means dispatcher launches converge)
review mark-dispatched rc=2 (0 means dispatcher launches rework)
ledger after both claims:
  File "/opt/homebrew/Cellar/python@3.14/3.14.6/Frameworks/Python.framework/Versions/3.14/lib/python3.14/json/tool.py", line 120, in <module>
    main()
    ~~~~^^
  File "/opt/homebrew/Cellar/python@3.14/3.14.6/Frameworks/Python.framework/Versions/3.14/lib/python3.14/json/tool.py", line 88, in main
    infile = open(options.infile, encoding='utf-8')
FileNotFoundError: [Errno 2] No such file or directory: '/attempts.json'
PASS: cap-out blocked both atomic dispatch claims.

exec
/bin/zsh -lc 'python3 -c $'"'import importlib.util\\nfrom pathlib import Path\\nroot = Path.cwd() / \"q-system/.q-system/scripts\"\\ndef load(name, path):\\n    spec = importlib.util.spec_from_file_location(name, path)\\n    mod = importlib.util.module_from_spec(spec)\\n    spec.loader.exec_module(mod)\\n    return mod\\nci = load(\"ci_redrive_repro\", root / \"ci-redrive.py\")\\nreview = load(\"review_redrive_repro\", root / \"review-redrive.py\")\\ncalls = []\\ndef claim(path, issue, flag):\\n    calls.append((issue, flag))\\n    return True\\nci.capout = lambda path, issue: \"concurrent converge hit cap\"\\nci.ledger_claim = claim\\nci.ledger_recorded = lambda path, issue, flag: True\\nreview.CI.capout = lambda path, issue: \"concurrent converge hit cap\"\\nreview.CI.ledger_claim = claim\\nci_rc = ci.cmd_mark_dispatched(\"ASK-871\", \"deadbeef1234\", \"abc123\")\\nreview_rc = review.cmd_mark_dispatched(\"ASK-871\", \"rework\", 210, \"abc123\")\\nprint(f\"capout at claim time: {ci.capout(\\\"ledger\\\", \\\"ASK-871\\\")"'!r}")'"\\nprint(f\"ci mark-dispatched rc={ci_rc} (0 authorizes launch)\")\\nprint(f\"review mark-dispatched rc={review_rc} (0 authorizes launch)\")\\nprint(f\"claim calls={calls"'!r}")'"\\nassert ci_rc "'!= 0 and review_rc != 0, "BUG: capped issue was authorized after selector/claim race"'"'" in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-210
 exited 1 in 0ms:
  File "<string>", line 26, in <module>
    assert ci_rc != 0 and review_rc != 0, "BUG: capped issue was authorized after selector/claim race"
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
AssertionError: BUG: capped issue was authorized after selector/claim race
capout at claim time: 'concurrent converge hit cap'
ci mark-dispatched rc=0 (0 authorizes launch)
review mark-dispatched rc=0 (0 authorizes launch)
claim calls=[('ASK-871', 'ci_redrive_deadbeef1234'), ('ASK-871', 'review_rework_pr210_abc123')]

exec
/bin/zsh -lc "rg -n \"PR_FIELDS|statusCheckRollup|headRefOid|isDraft\" q-system/.q-system/scripts/ci-redrive.py q-system/.q-system/scripts/review-redrive.py q-system/.q-system/scripts/test/test-ci-redrive.sh q-system/.q-system/scripts/test/test-review-redrive.sh
nl -ba q-system/.q-system/scripts/test/test-ci-redrive.sh | sed -n '180,280p'
nl -ba q-system/.q-system/scripts/test/test-review-redrive.sh | sed -n '190,250p'
git diff --check 40a8fdae..HEAD
python3 -c 'import ast, pathlib; [ast.parse(pathlib.Path(p).read_text(), filename=p) for p in [\"q-system/.q-system/scripts/attempts-ledger.py\",\"q-system/.q-system/scripts/ci-redrive.py\",\"q-system/.q-system/scripts/review-redrive.py\"]]; print(\"AST parse: PASS\")'
bash -n q-system/.q-system/scripts/converge.sh q-system/.q-system/scripts/test/test-converge-capout.sh && printf '%s\\n' 'bash -n: PASS'" in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-210
 succeeded in 10ms:
AST parse: PASS
bash -n: PASS

exec
/bin/zsh -lc "rg -n \"verdict_from_record\\(|def .*verdict|VERDICT\" q-system/.q-system/scripts/pr-verdict-lib.sh q-system/.q-system/scripts/pr-review-agent.sh | sed -n '1,220p'
nl -ba q-system/.q-system/scripts/pr-verdict-lib.sh | sed -n '130,220p'
rg -n '\"verdict\"|json.dump.*verdict|verdict.*=' q-system/.q-system/scripts/pr-review-agent.sh | tail -80" in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-210
 succeeded in 0ms:
q-system/.q-system/scripts/pr-verdict-lib.sh:18:# Anchored on the VERDICT line first so prose like "this would BLOCK deploys"
q-system/.q-system/scripts/pr-verdict-lib.sh:26:#   - the verdict may sit on the line AFTER a bare `## VERDICT` heading, so the
q-system/.q-system/scripts/pr-verdict-lib.sh:35:# A VERDICT STATEMENT, NOT A LINE MENTIONING THE WORD (ASK-356). This used to
q-system/.q-system/scripts/pr-verdict-lib.sh:36:# anchor on the first line matching `VERDICT`, take the first verdict token within
q-system/.q-system/scripts/pr-verdict-lib.sh:43:#     - **VERDICT:** decided by THIS RULE, not by feel:
q-system/.q-system/scripts/pr-verdict-lib.sh:46:# That line is the first `VERDICT` match in every codex review ever produced, so
q-system/.q-system/scripts/pr-verdict-lib.sh:49:# VERDICT="$DERIVED_VERDICT" won unconditionally, and became fleet-blocking the
q-system/.q-system/scripts/pr-verdict-lib.sh:60:# the rule puts prose there. So the token must lead what follows `VERDICT:`, or
q-system/.q-system/scripts/pr-verdict-lib.sh:61:# follow a marker that is bare (`## VERDICT` on its own line -- the real PR #11
q-system/.q-system/scripts/pr-verdict-lib.sh:75:# (`REWORK_VERDICTS = {"REQUEST CHANGES", ...}` is in the diff of this very
q-system/.q-system/scripts/pr-verdict-lib.sh:109:    line ~ /VERDICT/ {
q-system/.q-system/scripts/pr-verdict-lib.sh:111:      sub(/^.*VERDICT[*_]*[[:space:]]*:?[[:space:]]*/, "", rest)
q-system/.q-system/scripts/pr-verdict-lib.sh:557:# THE VERDICT AND FINDINGS DO NOT COME FROM THE TRUNCATED TEXT. They are taken
q-system/.q-system/scripts/pr-verdict-lib.sh:647:verdict_from_record() {
q-system/.q-system/scripts/pr-review-agent.sh:215:#   VERDICT_DIR (the ONE record the loop gates on). converge.sh:36 and
q-system/.q-system/scripts/pr-review-agent.sh:246:  VERDICT_DIR="$OUT_DIR";              STATUS_CONTEXT="kipi/reviewer-approved"; MINOR_TAG=""
q-system/.q-system/scripts/pr-review-agent.sh:248:  VERDICT_DIR="$OUT_DIR/$ENGINE";      STATUS_CONTEXT="kipi/$ENGINE-approved";  MINOR_TAG="$ENGINE "
q-system/.q-system/scripts/pr-review-agent.sh:269:mkdir -p "$ENGINE_DIR" "$VERDICT_DIR"
q-system/.q-system/scripts/pr-review-agent.sh:604:- **VERDICT:** decided by THIS RULE, not by feel:
q-system/.q-system/scripts/pr-review-agent.sh:776:STATED_VERDICT="$(extract_verdict "$REVIEW")"
q-system/.q-system/scripts/pr-review-agent.sh:777:DERIVED_VERDICT="$(verdict_from_findings "$REVIEW")"
q-system/.q-system/scripts/pr-review-agent.sh:782:  # otherwise green-light the PR, with the truncated prose "VERDICT: APPROVE"
q-system/.q-system/scripts/pr-review-agent.sh:787:  VERDICT=""
q-system/.q-system/scripts/pr-review-agent.sh:789:elif [ -n "$DERIVED_VERDICT" ]; then
q-system/.q-system/scripts/pr-review-agent.sh:791:  # VERDICT="$DERIVED_VERDICT" unconditionally, printing a NOTE and proceeding --
q-system/.q-system/scripts/pr-review-agent.sh:797:  VERDICT="$(resolve_verdict "$STATED_VERDICT" "$DERIVED_VERDICT")"
q-system/.q-system/scripts/pr-review-agent.sh:798:  if [ "$STATED_VERDICT" != "$DERIVED_VERDICT" ]; then
q-system/.q-system/scripts/pr-review-agent.sh:799:    echo "  NOTE: reviewer stated '${STATED_VERDICT:-none}' but its own findings imply '$DERIVED_VERDICT'; taking the harsher: '$VERDICT'"
q-system/.q-system/scripts/pr-review-agent.sh:802:  VERDICT="$STATED_VERDICT"
q-system/.q-system/scripts/pr-review-agent.sh:805:echo "  verdict: ${VERDICT:-unstated}$DRY_NOTE"
q-system/.q-system/scripts/pr-review-agent.sh:818:# The record lands in $VERDICT_DIR, NOT next to the review. Those are the same
q-system/.q-system/scripts/pr-review-agent.sh:827:# THE VERDICT DOES NOT ANSWER IT. Measured across all 79 records on 2026-08-03:
q-system/.q-system/scripts/pr-review-agent.sh:842:# RECORD-ONLY, DELIBERATELY. This changes no gate. $VERDICT is computed above and
q-system/.q-system/scripts/pr-review-agent.sh:885:python3 - "$PR" "$ISSUE" "$VERDICT" "$REVIEW" "$(TS)" "$STATED_VERDICT" "$DERIVED_VERDICT" "$ROUND" "$HEAD_SHA" "$VERDICT_DIR" "$ENGINE" "$INVOKER" "$REVIEW_USABLE" "$REVIEWED_BY" "$DEGRADED" "$(verdict_record_write_path "$VERDICT_DIR" "$REVIEW_SLUG" "$PR")" <<'PY'
q-system/.q-system/scripts/pr-review-agent.sh:914:if [ "$VERDICT" = "APPROVE WITH NITS" ] && [ -n "$ISSUE" ]; then
q-system/.q-system/scripts/pr-review-agent.sh:953:  # ONE reader of the verdict. $VERDICT is the derived-over-stated value already
q-system/.q-system/scripts/pr-review-agent.sh:1019:    review_comment_body "$REVIEW" "$VERDICT" "$ENGINE" "$DEGRADED" >"$REVIEW_BODY"
q-system/.q-system/scripts/pr-review-agent.sh:1041:    post_reviewer_status "$HEAD_SHA" "$VERDICT" "$COMMENT_URL"
q-system/.q-system/scripts/pr-review-agent.sh:1080:      "Review of PR #$PR complete ($ENGINE engine$([ "$DEGRADED" = "1" ] && printf ', DEGRADED: codex down, Opus fallback')). Verdict: ${VERDICT:-unstated}. Reviewer: Meta senior-staff persona, fresh eyes, every finding required to ship an executed reproducer.
q-system/.q-system/scripts/pr-review-agent.sh:1095:      NOTIFY_OUT="$(bash "$NOTIFY" "reviewer: PR #$PR review did NOT reach $ISSUE ($ENGINE engine, verdict ${VERDICT:-unstated}). The gate moved but the findings are not on the issue, so the rework conversation cannot start." 2>&1)"
   130	#   MULTIPLE BLOCKS CONCATENATE. sed RESTARTS the range after every `END
   131	#   FINDINGS`, so a review containing two blocks yields both, glued together, and
   132	#   a severity from a block that is not the verdict block sets the gate. Not
   133	#   hypothetical: codex stdout is known to carry harness noise and A REPEATED
   134	#   FINAL LINE (recorded on ASK-221 -- `hook: Stop`, `tokens used`, the last line
   135	#   again), the reviewer prompt itself contains a literal `FINDINGS:` /
   136	#   `END FINDINGS` template the model can echo, and from round 2 the prompt hands
   137	#   the model the PREVIOUS round's findings to re-prove.
   138	#
   139	#   AN UNCLOSED BLOCK RUNS TO EOF. A stream that died one line into the block
   140	#   printed a range with no severity lines, and no severities derives APPROVE --
   141	#   a green gate for a review nobody read. pr-review-agent.sh defends that with
   142	#   its own REVIEW_UNUSABLE flag, but the LIB still handed APPROVE to anyone else
   143	#   who asked, so the safety lived in one caller instead of in the reader.
   144	#
   145	# LAST, NOT FIRST. The prompt says "Last, a machine-readable findings block", and
   146	# a duplicated final line makes the last one the complete one. Taking the first
   147	# would pick the echoed template.
   148	#
   149	# A BLOCK STILL OPEN AT EOF VOIDS THE WHOLE REVIEW, earlier complete blocks
   150	# included. This is the case the first cut of this function got wrong, caught by
   151	# its own reproducer: "last COMPLETE block" quietly falls back to a QUOTED
   152	# prior-round block when the real trailing block is cut off mid-write, so a
   153	# truncated review would derive a verdict from findings it had already withdrawn --
   154	# and the completeness predicate built on it would call that review usable. An open
   155	# block at EOF is evidence of truncation, so nothing in the stream is trustworthy.
   156	#
   157	# The cost of that strictness is a review whose PROSE happens to end with a line
   158	# starting `FINDINGS:` reads as unstated. That is the safe direction and the same
   159	# posture the rest of this file takes: unstated HOLDS a PR, green RELEASES it.
   160	#
   161	# awk, not sed, because the rule is stateful in two ways a range expression cannot
   162	# express: opening a new block discards an unclosed one, and the end-of-input state
   163	# decides whether any of it counts.
   164	# A BLOCK OF NOTHING BUT PLACEHOLDERS IS THE PROMPT, NOT THE REVIEW (sp-df1a458f).
   165	# The reviewer prompt ends with a literal template -- `severity|one-sentence
   166	# claim|file:line` between the two markers -- and `codex exec` echoes the whole
   167	# prompt to stdout. When the model answers with a PLAN instead of a review, that
   168	# echo is the ONLY complete block in the stream. Measured 2026-08-03 on Alice PR
   169	# #1 round 2: the block was accepted, `severity|` matched no severity, no
   170	# severities derived APPROVE, and kipi/reviewer-approved -- a REQUIRED context on
   171	# main -- went green on code nobody had read, 12 seconds after dispatch.
   172	#
   173	# The rule is NOT "reject a block with no severity rows": an EMPTY block is
   174	# legitimate and load-bearing (a round 2 that refutes everything closes with one;
   175	# test-severity-floor.sh and test-findings-block-reader.sh case 2 both pin it).
   176	# The discriminator is rows-that-are-not-findings:
   177	#
   178	#   zero rows                      -> legitimate empty block, kept
   179	#   >=1 row, >=1 a real severity   -> a real block, kept
   180	#   >=1 row, NONE a real severity  -> the template echo (or prose leaked into the
   181	#                                     block); not a findings block, skipped
   182	#
   183	# Skipped, not fatal: an earlier COMPLETE real block still stands, because the
   184	# echo arrives BEFORE the model's own answer. Truncation is unchanged -- a block
   185	# still open at EOF voids everything, which is the stricter case and stays.
   186	#
   187	# THE SEVERITY TOKENS HERE ARE THE SAME FOUR verdict_from_findings GRADES, and
   188	# they have to be: a row this function calls real but the derivation cannot grade
   189	# contributes nothing, so the block would count as usable and derive APPROVE --
   190	# the defect again, one layer down. Change the two together or not at all.
   191	findings_block() {
   192	  local f="$1"
   193	  [ -s "$f" ] || return 0
   194	  awk '
   195	    /^FINDINGS:/            { buf = $0 "\n"; open = 1; rows = 0; sev = 0; next }
   196	    open && /^END FINDINGS/ { if (rows == 0 || sev > 0) last = buf $0 "\n"
   197	                              open = 0; next }
   198	    open                    { buf = buf $0 "\n"
   199	                              if ($0 ~ /^[ \t]*$/) next
   200	                              rows++
   201	                              if ($0 ~ /^(blocker|major|minor|nit)[|]/) sev++
   202	                              next }
   203	    END                     { if (open) exit 0; printf "%s", last }
   204	  ' "$f" 2>/dev/null
   205	}
   206	
   207	# _text_after_last_findings_block <review-file>
   208	# Everything the stream said AFTER its last `END FINDINGS`, or the whole file when
   209	# there is no block at all. The prompt orders the block LAST ("Last, a
   210	# machine-readable findings block"), so this is the region a finished review has
   211	# nothing substantive in -- which is what makes it the right place to look for an
   212	# answer that never started. Internal to review_declined_to_start.
   213	_text_after_last_findings_block() {
   214	  awk '/^END FINDINGS/ { tail = ""; next } { tail = tail $0 "\n" }
   215	       END { printf "%s", tail }' "${1:-}" 2>/dev/null
   216	}
   217	
   218	# review_declined_to_start <review-file>
   219	# True when the reviewer answered with a PLAN and waited for confirmation instead
   220	# of reviewing. Exit 0 = it declined (the review is not a review).
890:json.dump({"pr": int(pr), "issue": issue, "verdict": verdict,
947:  local sha="$1" verdict="$2" target="$3"

codex
The race reproduced: both atomic claim functions returned `0` while the cap-out was already present, which means `kipi-dispatch.sh` proceeds to launch work at lines 1247 onward. I’m finishing calibration now, including the ledger’s update/clear semantics, selector behavior, syntax, and fixture-to-producer shape.
exec
/bin/zsh -lc 'python3 -c $'"'import importlib.util\\nfrom pathlib import Path\\nroot=Path.cwd()/\"q-system/.q-system/scripts\"\\ndef load(name,p):\\n s=importlib.util.spec_from_file_location(name,p); m=importlib.util.module_from_spec(s); s.loader.exec_module(m); return m\\nledger=load(\"ledger_sound\",root/\"attempts-ledger.py\")\\nci=load(\"ci_sound\",root/\"ci-redrive.py\")\\nd={\"ASK-1\":{\"count\":2},\"ASK-2\":{\"capout\":True,\"capout_at\":\"old\",\"capout_why\":\"other\"}}\\nassert ledger.op_capout(d,\"ASK-1\",\"first\",\"t1\")\\nassert d[\"ASK-1\"]=={\"count\":2,\"capout\":True,\"capout_at\":\"t1\",\"capout_why\":\"first\"}\\nassert ledger.op_capout(d,\"ASK-1\",\"second\",\"t2\")\\nassert d[\"ASK-1\"][\"capout_why\"]==\"second\" and d[\"ASK-1\"][\"capout_at\"]==\"t2\"\\nassert ledger.op_clear(d,\"ASK-1\",(\"capout\",\"capout_at\",\"capout_why\"))\\nassert d[\"ASK-1\"]=={\"count\":2} and d[\"ASK-2\"][\"capout\"] is True\\nrows=[]\\nci.ledger_get=lambda path,issue,key: {\"capout\":\"True\",\"capout_why\":\"round cap\"}.get(key,\"\")\\nassert ci.capout_skip(\"ledger\",\"ASK-1\",\"ci-redrive\",210) is True\\nci.ledger_get=lambda path,issue,key: \"\"\\nassert ci.capout_skip(\"ledger\",\"ASK-1\",\"ci-redrive\",210) is False\\nprint(\"PASS: record updates reason/timestamp without erasing counters\")\\nprint(\"PASS: clear removes only one issue cap-out and preserves other ledger fields/issues\")\\nprint(\"PASS: selector helper skips recorded cap-outs and allows uncapped issues\")' 2>&1" in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-210
 succeeded in 0ms:
ci-redrive: ASK-1 PR #210 -- converge already gave up on this issue (round cap). Not re-entering it until a human clears the cap-out: python3 q-system/.q-system/scripts/attempts-ledger.py ledger clear-capout ASK-1
PASS: record updates reason/timestamp without erasing counters
PASS: clear removes only one issue cap-out and preserves other ledger fields/issues
PASS: selector helper skips recorded cap-outs and allows uncapped issues

exec
/bin/zsh -lc "nl -ba q-system/.q-system/scripts/converge.sh | sed -n '80,145p'
rg -n \"pr_for_branch|pr_head_sha|KIPI_GH|GH=\" q-system/.q-system/scripts/converge.sh q-system/.q-system/scripts/repo-slug-lib.sh q-system/.q-system/scripts/pr-verdict-lib.sh" in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-210
 succeeded in 0ms:
    80	TS() { date -u +%Y-%m-%dT%H:%M:%SZ; }
    81	say() { echo "$(TS) converge[$ISSUE] $*" | tee -a "$LOG"; }
    82	
    83	BRANCH="sana/$(echo "$ISSUE" | tr 'A-Z' 'a-z')"
    84	# SCOPED TO THE REPO THE WORK IS IN, not to cwd (ASK-738). This driver runs
    85	# from the dispatcher's cwd (the home checkout), so an unqualified `gh pr list`
    86	# asked the HOME repo whether the TARGET's branch had a PR: after the worker
    87	# opened one in the target, converge found none and stopped -- or worse, found
    88	# an unrelated PR of the same branch name at home and drove rounds against it.
    89	# KIPI_TARGET_REPO is the carrier kipi-dispatch.sh uses; it forwards only its
    90	# own arguments to the worker, so the target crosses that boundary by
    91	# inheritance. Derived ONCE, through the shared lib, same as the worker.
    92	TARGET_REPO="${KIPI_TARGET_REPO:-$SKEL}"
    93	TARGET_SLUG="$(slug_for_repo "$TARGET_REPO" "${KIPI_SLUG_REGISTRY:-$SKEL/instance-registry.json}")"
    94	KIPI_GH_REPO_ARGS="$(gh_repo_args "$TARGET_SLUG")"
    95	export KIPI_GH_REPO_ARGS
    96	# shellcheck disable=SC2086
    97	pr_for_branch() { gh pr list $KIPI_GH_REPO_ARGS --head "$BRANCH" --json number -q '.[0].number' 2>/dev/null; }
    98	# The head sha comes from pr_head_sha in the shared lib, not a local copy of the
    99	# same `gh pr view`: this driver and linear-worker.sh now BOTH compare it against
   100	# the sha a review pinned, and two private readers of one input is how those two
   101	# comparisons drift apart.
   102	
   103	if [ "$DRY" = "1" ]; then
   104	  PR="$(pr_for_branch)"
   105	  V=""; [ -n "$PR" ] && V="$(verdict_from_record "$(verdict_record_path "$REVIEWS_DIR" "$TARGET_SLUG" "$PR")")"
   106	  say "[dry] branch=$BRANCH pr=${PR:-none} verdict=${V:-none} would run up to $MAX_ROUNDS round(s)"
   107	  exit 0
   108	fi
   109	
   110	# RELEASE THE CLAIM IF THIS RUN IS KILLED.
   111	#
   112	# linear-claim.py deliberately does NOT pid-check the claim itself (only the
   113	# critical-section guard) -- the claiming python process exits immediately, so
   114	# its pid is meaningless, and the claim is meant to outlive it. Correct design,
   115	# real operational hole: a SIGKILL, a harness timeout, a laptop sleeping, or a
   116	# ctrl-c leaves the lock held with nothing to reclaim it.
   117	#
   118	# Observed 2026-07-27: this driver was killed mid-run on ASK-181 and left
   119	# `ASK-181 claimed by sana (session worker-1785159359-39569)`. Because the lock
   120	# is still repo-root scoped until ASK-188 lands, that one dead session blocked
   121	# EVERY issue on the board until a human released it by hand. In an unattended
   122	# loop, "a human notices and runs release" is not a recovery path -- nobody is
   123	# watching, which is the entire premise.
   124	#
   125	# The trap cannot know the worker's session token (the worker mints its own), so
   126	# it releases by the holder RECORDED IN THE LOCK, which is what the manual fix
   127	# does. Best-effort by design: never let cleanup failure mask the real exit code.
   128	release_stale_claim_for_issue() {
   129	  local held
   130	  held="$(python3 - "$ISSUE" <<'PY' 2>/dev/null || true
   131	import json, os, subprocess, sys
   132	issue = sys.argv[1]
   133	override = os.environ.get("KIPI_LINEAR_CLAIMS")
   134	if override:
   135	    path = override
   136	else:
   137	    try:
   138	        root = subprocess.run(["git", "rev-parse", "--show-toplevel"],
   139	                              capture_output=True, text=True, check=True).stdout.strip()
   140	    except Exception:
   141	        raise SystemExit(0)
   142	    path = os.path.join(root, ".linear-claims.json")
   143	try:
   144	    rec = json.load(open(path))
   145	except Exception:
q-system/.q-system/scripts/converge.sh:94:KIPI_GH_REPO_ARGS="$(gh_repo_args "$TARGET_SLUG")"
q-system/.q-system/scripts/converge.sh:95:export KIPI_GH_REPO_ARGS
q-system/.q-system/scripts/converge.sh:97:pr_for_branch() { gh pr list $KIPI_GH_REPO_ARGS --head "$BRANCH" --json number -q '.[0].number' 2>/dev/null; }
q-system/.q-system/scripts/converge.sh:98:# The head sha comes from pr_head_sha in the shared lib, not a local copy of the
q-system/.q-system/scripts/converge.sh:104:  PR="$(pr_for_branch)"
q-system/.q-system/scripts/converge.sh:783:  PR="$(pr_for_branch)"
q-system/.q-system/scripts/converge.sh:827:  SHA="$(pr_head_sha "$PR")"
q-system/.q-system/scripts/pr-verdict-lib.sh:373:  gh pr view "$1" ${KIPI_GH_REPO_ARGS:-} --json mergeStateStatus -q .mergeStateStatus 2>/dev/null | tr -d '[:space:]'
q-system/.q-system/scripts/pr-verdict-lib.sh:376:# pr_head_sha <pr-number>
q-system/.q-system/scripts/pr-verdict-lib.sh:384:pr_head_sha() {
q-system/.q-system/scripts/pr-verdict-lib.sh:385:  # $KIPI_GH_REPO_ARGS is set ONCE by whichever script sourced this lib, from
q-system/.q-system/scripts/pr-verdict-lib.sh:389:  gh pr view "$1" ${KIPI_GH_REPO_ARGS:-} --json headRefOid -q .headRefOid 2>/dev/null | tr -d '[:space:]'

codex
## Finding 1

**MAJOR, Confirmed:** Both atomic dispatch entrypoints authorize work after a concurrent cap-out because they claim the attempt without checking `capout` in the same ledger transaction.

**Location:** `q-system/.q-system/scripts/ci-redrive.py:586`, `q-system/.q-system/scripts/review-redrive.py:614`

At 3am, this can launch another converge or review after the system announced it was parked. A human must inspect the unintended run and clean up any commits or status changes it produced.

**Reproducer command actually run:**

```bash
python3 -c $'import importlib.util\nfrom pathlib import Path\nroot = Path.cwd() / "q-system/.q-system/scripts"\ndef load(name, path):\n    spec = importlib.util.spec_from_file_location(name, path)\n    mod = importlib.util.module_from_spec(spec)\n    spec.loader.exec_module(mod)\n    return mod\nci = load("ci_redrive_repro", root / "ci-redrive.py")\nreview = load("review_redrive_repro", root / "review-redrive.py")\ncalls = []\ndef claim(path, issue, flag):\n    calls.append((issue, flag))\n    return True\nci.capout = lambda path, issue: "concurrent converge hit cap"\nci.ledger_claim = claim\nci.ledger_recorded = lambda path, issue, flag: True\nreview.CI.capout = lambda path, issue: "concurrent converge hit cap"\nreview.CI.ledger_claim = claim\nci_rc = ci.cmd_mark_dispatched("ASK-871", "deadbeef1234", "abc123")\nreview_rc = review.cmd_mark_dispatched("ASK-871", "rework", 210, "abc123")\nprint(f"capout at claim time: {ci.capout(\\"ledger\\", \\"ASK-871\\")!r}")\nprint(f"ci mark-dispatched rc={ci_rc} (0 authorizes launch)")\nprint(f"review mark-dispatched rc={review_rc} (0 authorizes launch)")\nprint(f"claim calls={calls!r}")\nassert ci_rc != 0 and review_rc != 0, "BUG: capped issue was authorized after selector/claim race"'
```

**Real output:**

```text
  File "<string>", line 26, in <module>
    assert ci_rc != 0 and review_rc != 0, "BUG: capped issue was authorized after selector/claim race"
AssertionError: BUG: capped issue was authorized after selector/claim race
capout at claim time: 'concurrent converge hit cap'
ci mark-dispatched rc=0 (0 authorizes launch)
review mark-dispatched rc=0 (0 authorizes launch)
claim calls=[('ASK-871', 'ci_redrive_deadbeef1234'), ('ASK-871', 'review_rework_pr210_abc123')]
```

The sandbox rejected creating the reproducer file under `$TMPDIR`, so the identical reproducer was executed inline.

## What is sound

- `record-capout` updates the reason and timestamp without erasing existing attempt counters.
- Re-recording replaces stale cap-out details.
- `clear-capout` preserves unrelated fields and other issues.
- `capout_skip` correctly distinguishes capped and uncapped issues.
- Python AST parsing, `bash -n`, and `git diff --check` passed.
- The new fixture’s ledger key and PR rollup fields match the actual producer and consumer fields in this repository.
- The requested GitHub reads were attempted, but `api.github.com` was unreachable. The review used the local PR head against `origin/main`.

## VERDICT: REQUEST CHANGES

Fix the atomic dispatch claim first. Add one ledger operation that checks `capout` and claims the dispatch flag under the same flock transaction. Calling `capout()` immediately before `ledger_claim()` would leave the same race open.

FINDINGS:
major|Both atomic dispatch entrypoints authorize work after a concurrent cap-out because they claim the attempt without checking capout in the same ledger transaction|q-system/.q-system/scripts/ci-redrive.py:586,q-system/.q-system/scripts/review-redrive.py:614
END FINDINGS
tokens used
149,577
## Finding 1

**MAJOR, Confirmed:** Both atomic dispatch entrypoints authorize work after a concurrent cap-out because they claim the attempt without checking `capout` in the same ledger transaction.

**Location:** `q-system/.q-system/scripts/ci-redrive.py:586`, `q-system/.q-system/scripts/review-redrive.py:614`

At 3am, this can launch another converge or review after the system announced it was parked. A human must inspect the unintended run and clean up any commits or status changes it produced.

**Reproducer command actually run:**

```bash
python3 -c $'import importlib.util\nfrom pathlib import Path\nroot = Path.cwd() / "q-system/.q-system/scripts"\ndef load(name, path):\n    spec = importlib.util.spec_from_file_location(name, path)\n    mod = importlib.util.module_from_spec(spec)\n    spec.loader.exec_module(mod)\n    return mod\nci = load("ci_redrive_repro", root / "ci-redrive.py")\nreview = load("review_redrive_repro", root / "review-redrive.py")\ncalls = []\ndef claim(path, issue, flag):\n    calls.append((issue, flag))\n    return True\nci.capout = lambda path, issue: "concurrent converge hit cap"\nci.ledger_claim = claim\nci.ledger_recorded = lambda path, issue, flag: True\nreview.CI.capout = lambda path, issue: "concurrent converge hit cap"\nreview.CI.ledger_claim = claim\nci_rc = ci.cmd_mark_dispatched("ASK-871", "deadbeef1234", "abc123")\nreview_rc = review.cmd_mark_dispatched("ASK-871", "rework", 210, "abc123")\nprint(f"capout at claim time: {ci.capout(\\"ledger\\", \\"ASK-871\\")!r}")\nprint(f"ci mark-dispatched rc={ci_rc} (0 authorizes launch)")\nprint(f"review mark-dispatched rc={review_rc} (0 authorizes launch)")\nprint(f"claim calls={calls!r}")\nassert ci_rc != 0 and review_rc != 0, "BUG: capped issue was authorized after selector/claim race"'
```

**Real output:**

```text
  File "<string>", line 26, in <module>
    assert ci_rc != 0 and review_rc != 0, "BUG: capped issue was authorized after selector/claim race"
AssertionError: BUG: capped issue was authorized after selector/claim race
capout at claim time: 'concurrent converge hit cap'
ci mark-dispatched rc=0 (0 authorizes launch)
review mark-dispatched rc=0 (0 authorizes launch)
claim calls=[('ASK-871', 'ci_redrive_deadbeef1234'), ('ASK-871', 'review_rework_pr210_abc123')]
```

The sandbox rejected creating the reproducer file under `$TMPDIR`, so the identical reproducer was executed inline.

## What is sound

- `record-capout` updates the reason and timestamp without erasing existing attempt counters.
- Re-recording replaces stale cap-out details.
- `clear-capout` preserves unrelated fields and other issues.
- `capout_skip` correctly distinguishes capped and uncapped issues.
- Python AST parsing, `bash -n`, and `git diff --check` passed.
- The new fixture’s ledger key and PR rollup fields match the actual producer and consumer fields in this repository.
- The requested GitHub reads were attempted, but `api.github.com` was unreachable. The review used the local PR head against `origin/main`.

## VERDICT: REQUEST CHANGES

Fix the atomic dispatch claim first. Add one ledger operation that checks `capout` and claims the dispatch flag under the same flock transaction. Calling `capout()` immediately before `ledger_claim()` would leave the same race open.

FINDINGS:
major|Both atomic dispatch entrypoints authorize work after a concurrent cap-out because they claim the attempt without checking capout in the same ledger transaction|q-system/.q-system/scripts/ci-redrive.py:586,q-system/.q-system/scripts/review-redrive.py:614
END FINDINGS

assafkip:
## Round 2: the major finding is real and fixed in `53162442`

> Both atomic dispatch entrypoints authorize work after a concurrent cap-out because they claim the attempt without checking capout in the same ledger transaction

Confirmed, not argued down. Round 1 put the gate in the **read-only offer** (`cmd_redrive`, `cmd_select`). The attempt is spent later, in `mark-dispatched`, and that call asked only whether the flag was free.

The window is not theoretical: `kipi-dispatch.sh` runs a `ps` snapshot, a converge-live probe and a budget read between the offer and the claim. A converge exiting in that interval makes `converge_live` go false, the offer clears the capout check microseconds before `op_capout` lands, and the claim then says yes.

### Why the check is not in the caller

Reading capout before calling `claim-flag` only *narrows* the window. Two subprocesses cannot be atomic against a third writer. So the refusal moved into the ledger's own transaction:

| File | Change |
|---|---|
| `attempts-ledger.py` | `op_claim_uncapped` + `claim-flag-uncapped` op: capout check and flag set under **one** `_mutate` flock |
| `ci-redrive.py` | `claim_unless_capped()` -- the one definition -- replaces `ledger_claim` in `cmd_mark_dispatched` |
| `review-redrive.py` | imports and calls the same function, as it already does for `capout_skip` / `is_reviewer_slot` |

### The half a naive fix gets wrong

**Refusing writes nothing.** Claiming first and noticing the park afterwards would spend the issue's one attempt on a dispatch that never happened, so `clear-capout` would release an issue no redrive can pick up: a permanent stall wearing a cleared park's clothes. Two cases pin it -- the flag is unset after a refusal, and the same claim succeeds once cleared.

**Exit 4 is its own code.** `1` means "already claimed, stay quiet" at six call sites; `2`/`3` mean "nothing was written". A park is neither: nothing was written **and** somebody has to be told, because a silently skipped dispatch is the 29-hour outage the redrives exist to end. Both refusal lines carry the exact `clear-capout` command.

### Verification

Reproducer extended **first**, RED observed at `16 passed, 7 failed`. The decisive line was `the attempt was consumed by the refusal (rc=1 after clear-capout)` -- proof the parked issue's attempt was already being silently spent. Every refusal case is paired with a byte-identical un-capped twin that must still be claimed, so a wall fails the suite too.

```
test-converge-capout.sh        23/23   (was 16/23)
test-attempts-ledger.py        13/13
test-ci-redrive.sh             65/65
test-review-redrive.sh         30/30
test-review-redrive-absent.py  PASS
test-converge.sh               21/21
```

PR #216, opened 2026-08-18
fix: updater sync commits carry the [no-issue] tag; test_dispatch_pinned marked skeleton-only

assafkip:
Two one-line fixes from the 2026-08-18 fleet update run:
- Pure_spectrum_Q's linear-ps-issue commit-msg hook blocked the fleet sync twice because the updater's message names no issue. The sync commits now carry `[no-issue: fleet updater skeleton sync]`, the hook's own sanctioned lane.
- `test_dispatch_pinned.sh` lives only at the skeleton root but was declared fleet-wide, so every instance reported capability-gate RED declared-but-missing. Marked skeleton_only.

Tracked as spillover sp-c952d633.


assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-216-20260818-132614.md` (8578 bytes).

```
FINDINGS:
major|skeleton_only does not suppress declared-but-missing, so every instance and the fleet sync run still exit RED for the exact reason this PR claims to fix|q-system/.q-system/capability-manifest.json:705
minor|both instance-mode skeleton_only tests create the file first, so no test can fail when a skeleton_only path is genuinely absent from an instance|q-system/.q-system/scripts/test_capability_gate.py:429
nit|capability-manifest.json lost its trailing newline|q-system/.q-system/capability-manifest.json:809
END FINDINGS
```

--- reviewer output ---

## Review: PR #216

Two independent one-line fixes. I verified both against executed reproducers. One works. The other does not do the thing the PR title and commit message say it does.

---

### 1. MAJOR — `skeleton_only` does not suppress `declared-but-missing`; every instance stays RED and the whole fleet run still exits non-zero

**Claim:** Adding `test_dispatch_pinned.sh` to `skeleton_only` was supposed to stop instances reporting `RED declared-but-missing`. It does not. `skeleton_only` is consulted in exactly one place — `run_tests()` at `q-system/.q-system/scripts/capability-gate.py:411`, which skips *execution*. The two-direction diff (`diff_declared_vs_actual`, `capability-gate.py:266-270`) never looks at `skeleton_only` and never looks at `mode`. `test_dispatch_pinned.sh` is a repo-root `.sh`, and repo-root files are not in the sync set (only `q-system/`, `.claude/`, `plugins/` rsync), so the file is absent in every instance and the `declared-but-missing (outside scan root)` branch fires unconditionally.

**File:** `q-system/.q-system/capability-manifest.json:705` (the added entry); root cause at `q-system/.q-system/scripts/capability-gate.py:266`

**Reproducer** (simulates an instance: the three synced trees, no repo-root scripts, no `instance-registry.json`), plus a mutation showing the actual fix:

```python
python3 - <<'EOF'
import os, shutil, subprocess, tempfile, json
REPO="/Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-216"
GATE=os.path.join(REPO,"q-system/.q-system/scripts/capability-gate.py")
T=tempfile.mkdtemp(); inst=os.path.join(T,"inst"); os.makedirs(inst)
for d in ("q-system","plugins",".claude"):
    shutil.copytree(os.path.join(REPO,d), os.path.join(inst,d), ignore=shutil.ignore_patterns('.git'))
mp=os.path.join(inst,"q-system/.q-system/capability-manifest.json")
def run(tag):
    r=subprocess.run(["python3",GATE,"--repo-root",inst,"--check-only"],capture_output=True,text=True)
    print(f"=== {tag} rc={r.returncode}"); print(r.stdout.strip())
run("A: PR as-is, instance mode")
m=json.load(open(mp))
m["expected_tests"]=[e for e in m["expected_tests"] if e.get("path")!="test_dispatch_pinned.sh"]
json.dump(m,open(mp,"w"),indent=2)
run("B: same tree, declaration removed")
EOF
```

**Real output:**

```
=== A: PR as-is, instance mode rc=1
capability-gate mode=instance
  declared: 163 tests (0 quarantined), 21 skeleton-only, 19 declared-inert
  inert-engine check: skeleton-only (instance wiring is declared and gated in the skeleton)
  RED: declared-but-missing (outside scan root): test_dispatch_pinned.sh
capability-gate: RED (1)
=== B: same tree, declaration removed rc=0
capability-gate mode=instance
  declared: 162 tests (0 quarantined), 21 skeleton-only, 19 declared-inert
  inert-engine check: skeleton-only (instance wiring is declared and gated in the skeleton)
capability-gate: GREEN
```

**Blast radius at 3am.** `kipi-update.sh:2517-2525` runs this gate `--check-only` inside every instance on every fleet sync, prints `capability gate: RED`, appends the instance to `GATE_FAIL`, and the run exits non-zero at the end. So after this merges, the next unattended fleet sync still prints RED for all ~23 instances and still exits non-zero — while the commit message asserts the RED was fixed. That is a checker crying wolf on every run, which is what trains the operator to stop reading the gate output. Fully recoverable (one line), so major, not blocker.

**Fix:** either drop the `expected_tests` entry for `test_dispatch_pinned.sh` (mutation B, GREEN), or teach `diff_declared_vs_actual` to skip `skeleton_only` paths when `mode == "instance"`. The second is the better fix — it makes `skeleton_only` mean what its own PRD text says ("artifacts that must not run in instances") on both code paths instead of one — but it widens blast radius to all 21 existing `skeleton_only` entries, so it earns its own issue.

---

### 2. MINOR — the two instance-mode `skeleton_only` tests create the file, so nothing in the suite can fail on the production shape

`test_capability_gate.py:429` and `:469` both call `add_test(root, ...)` before declaring the path `skeleton_only`, so the file always exists in the fake instance. That exercises the *execution* skip only. No case exists where a `skeleton_only` path is genuinely absent from the instance — which is precisely the real fleet shape and precisely the case that stays RED. This is why the fix in finding 1 looked correct.

**File:** `q-system/.q-system/scripts/test_capability_gate.py:429`

**Reproducer / output:** the suite is green on both branches; the delta is invisible to it. Directly:

```
$ grep -n "skeleton_only=\[rel\]" -B2 q-system/.q-system/scripts/test_capability_gate.py
427-        root = make_repo(tmp, skeleton=False)
428-        rel = add_test(root, "q-system/.q-system/scripts/test_skel.py", crash)
429:        m = base_manifest(expected_tests=[entry(rel)], skeleton_only=[rel])
--
467-        root = make_repo(tmp, skeleton=False)
468-        rel = add_test(root, "q-system/.q-system/scripts/test_skelwire.py", crash)
469:        m = base_manifest(expected_tests=[entry(rel)], skeleton_only=[rel])
```

Mutation B above is the missing case, and it goes rc=1→rc=0 when the behavior changes, so it can actually fail. Worth adding alongside whichever fix lands.

---

### 3. NIT — the manifest lost its trailing newline

```
$ tail -c 20 q-system/.q-system/capability-manifest.json | od -c | tail -3
0000000    r   s   p   -   7   2   b   6   0   b   f   f   "  \n
0000020        ]  \n   }
0000024
```

JSON still parses; no check enforces it. Costs a spurious `\ No newline at end of file` on every future diff of this file.

---

## What is sound

- **The updater half works, verified positively.** Old message rc=1 (blocked), new message rc=0 with a ledger row, both variants:
  ```
  --- OLD (pre-PR): rc=1 |
  --- NEW (this PR): rc=0 | linear-first: bypass recorded (fleet updater skeleton sync)
  --- NEW config:   rc=0 | linear-first: bypass recorded (fleet updater skeleton sync)
  ledger: {"at": "2026-08-18T20:29:36+00:00", "reason": "fleet updater skeleton sync", ...}
  ```
- **The tag does not break `kipi rollback`.** I suspected an anchored message match. `kipi-rollback.sh:30` is `SYNC_PREFIX='^chore: sync \(q-system\|\.claude config + plugins\) from skeleton'` — prefix-anchored only, unaffected by an appended suffix. Same for the awk fallback at `:473`.
- **The tag does not break the hook-contract test.** `test-kipi-update-hook-contract.sh:182,230` grep `'^chore: sync ...'`, prefix only.
- **Coverage of the updater fix is complete.** All four commit sites in `kipi-update.sh` (1761, 2053, 2279, 2493) now carry a `[no-issue: ...]` tag; none were missed.
- **The bypass ledger does not dirty instances.** `q-system/output/**` is gitignored (`.gitignore:21`) and is a preserved path, so ~2 ledger rows per instance per sync cannot trip the dirty-tree guard or get clobbered. No automated consumer alerts on ledger volume, so this is not a noise source.
- **`guarded_commit` does not bypass instance hooks.** I checked whether the tag was moot because of `--no-verify`; it is not — the hook-guard at `kipi-update.sh:1147-1161` invokes the instance's real hook under `set -e`, so a refusing commit-msg hook genuinely aborts.
- **Skeleton mode is unaffected** by the manifest change; the file exists at skeleton root.

---

## VERDICT: REQUEST CHANGES

One major survived reproduction. Not a BLOCK — nothing here is permanent or unrecoverable, and the updater half is a real fix that should land.

**Fix first:** the capability-manifest half. Adding `test_dispatch_pinned.sh` to `skeleton_only` changes nothing about the RED it was meant to remove; the declared-vs-actual diff ignores `skeleton_only` entirely. Drop the `expected_tests` entry, or gate the diff on mode — and add the absent-file instance case to `test_capability_gate.py` so the next attempt has a check that can go red.

FINDINGS:
major|skeleton_only does not suppress declared-but-missing, so every instance and the fleet sync run still exit RED for the exact reason this PR claims to fix|q-system/.q-system/capability-manifest.json:705
minor|both instance-mode skeleton_only tests create the file first, so no test can fail when a skeleton_only path is genuinely absent from an instance|q-system/.q-system/scripts/test_capability_gate.py:429
nit|capability-manifest.json lost its trailing newline|q-system/.q-system/capability-manifest.json:809
END FINDINGS

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-216-20260818-133552.md` (6017 bytes).

```
FINDINGS:
minor|No test asserts the updater's sync-commit messages satisfy the linear-first commit-msg gate, so the fleet-blocking regression can silently return|kipi-update.sh:2279
minor|skeleton_only is only consulted for declared paths outside the scan roots, so an in-scan-root skeleton_only test absent from an instance still reports declared-but-missing|q-system/.q-system/scripts/capability-gate.py:272
END FINDINGS
```

--- reviewer output ---

Read-only review complete. Everything below was executed; nothing in the repo was modified.

---

## What I attacked and what survived

**Attack 1: "the `[no-issue:]` tag is decoration because the updater bypasses hooks."** This repo's own memory records `--no-verify` being added to `kipi-update.sh` in 2026-06-15 to fix a hang. If sync commits skipped hooks, the whole PR would be inert. It does not: `guarded_commit` (`kipi-update.sh:1179-1183`) commits with `-c core.hooksPath=<guard>` and the guard shim invokes the instance's real hook by its real path (`kipi-update.sh:1157-1159`). No `--no-verify` anywhere on this path. The tag is load-bearing.

Proved it end to end against the skeleton's own gate:

```
PR216 message      -> rc 0 |
pre-PR216 message  -> rc 1 | BLOCK: this commit names no Linear issue.
ledger rows: {"at": "2026-08-18T20:39:47+00:00", "reason": "fleet updater skeleton sync",
              "subject": "chore: sync q-system from skeleton 2026-08-18 [no-issue: fleet updater skeleton sync]"}
```

**Attack 2: "the new tests could not fail."** Mutated `capability-gate.py:272` to `if False:` in a temp copy and re-ran the new section:

```
--- skeleton_only_absent ---
FAIL: skeleton-only-absent: instance is GREEN without the file
PASS: skeleton-only-absent: skeleton is still RED without the file
rc = 1
```

Both halves are real. The negative half (skeleton stays RED) survived the mutation, which is correct: it grades the other branch.

**Attack 3: "the fixture is invented and proves nothing about the fleet."** Built a synthetic instance from this branch's real tree (`q-system/` + `plugins/` + `.claude/`, real 163-entry manifest) and ran the gate both ways:

```
BASELINE origin/main, same fake instance:
  RED: declared-but-missing (outside scan root): test_dispatch_pinned.sh
capability-gate: RED (1)          rc= 1

THIS BRANCH, same fake instance:
capability-gate: GREEN            rc= 0
```

The fleet RED is exactly one item, and this PR clears it. Skeleton mode on the real repo is still `capability-gate: GREEN`.

**Attack 4: "the longer commit subject breaks the updater's own tests."** Three tests reference the message; all assertions are prefix-anchored (`grep -q '^chore: sync q-system from skeleton'`, hook-contract test lines 182/230), and the rollback tests author their own messages. Verified by reading the assertions, not by executing those shell suites. Dropped.

---

## Findings

### minor — no test pins the updater's sync-commit messages against the commit-msg gate
`kipi-update.sh:2279`, `kipi-update.sh:2493`

The two message strings are the entire fix, and nothing grades them. `test-linear-issue-ref-check.sh` tests the checker; `test-kipi-update-hook-contract.sh` proves commit-msg hooks *run* but never asserts the message satisfies them. Grep for any test wiring the updater's messages into the gate:

```
$ grep -rln "linear-issue-ref-check\|linear_issue_ref" q-system/.q-system/scripts/test/ q-system/.q-system/scripts/*.py plugins/
q-system/.q-system/scripts/test/test-token-guard-hook-behavior.sh
q-system/.q-system/scripts/test/test-linear-issue-ref-check.sh
q-system/.q-system/scripts/test/fixtures/pr-verdict/real-review-request-changes.md
q-system/.q-system/scripts/ci-redrive.py
```

None of those reads `kipi-update.sh`. Reproducer for the regression that would return: feeding the pre-PR subject to the gate gives `rc 1 / BLOCK` (output above). At 3am the operator sees the sync abandon instances again, which is loud and fully recoverable, so it is a coverage gap and not a gate. A ~10-line test that extracts every `-m "chore:` subject from `kipi-update.sh` and runs each through `linear-issue-ref-check.py` closes it.

### minor — `skeleton_only` is honored inconsistently: only outside the scan roots
`q-system/.q-system/scripts/capability-gate.py:272`

`run_tests` (line 419) exempts **any** `skeleton_only` path in instance mode. The new exemption sits only in the out-of-scan-root loop, so a `skeleton_only` entry inside `q-system/.q-system/scripts/` that is genuinely absent from an instance still turns that instance RED, via the *first* loop at line 265-266, which has no `skeleton_only` consultation. Reproduced with a constructed manifest:

```
skeleton_only path INSIDE scan root, absent from instance:
capability-gate mode=instance
  declared: 1 tests (0 quarantined), 1 skeleton-only, 0 declared-inert
  RED: declared-but-missing: q-system/.q-system/scripts/test/test-skeleton-only-inside-scanroot.sh
capability-gate: RED (1)
```

Honest severity note, per the calibration rule: this is a constructed input. All 20 current in-scan-root `skeleton_only` entries live under `q-system/.q-system/scripts/test/`, which the q-system rsync copies to every instance (`rsync_owned_excludes` excludes only `INSTANCE_OWNED_SUBTREES`), so no producer emits this shape today. It is a latent trap for the next person who adds a `skeleton_only` entry that does not sync, and the docstring/comment reads as though `skeleton_only` is honored generally in instance mode. One-line fix: hoist `skeleton_only` into the first loop's filter too.

---

## VERDICT: APPROVE WITH NITS

Two minors, both bounded, neither gating. The core of the PR is correct and I could not break it: the tag actually clears the hook, the gate change actually flips a realistic instance from RED to GREEN, the skeleton half still bites, and the new test dies under mutation.

Most important thing to fix first: the missing regression test on the updater's commit messages. The `skeleton_only` asymmetry is the cheaper fix but the lower risk.

FINDINGS:
minor|No test asserts the updater's sync-commit messages satisfy the linear-first commit-msg gate, so the fleet-blocking regression can silently return|kipi-update.sh:2279
minor|skeleton_only is only consulted for declared paths outside the scan roots, so an in-scan-root skeleton_only test absent from an instance still reports declared-but-missing|q-system/.q-system/scripts/capability-gate.py:272
END FINDINGS

PR #217, opened 2026-08-18
voice-stop-gate: the corpus-similarity drift page reaches him at every session boundary (ASK-902)

assafkip:
Landing the ASK-902 voice-stop-gate work (authored by the social-voice session, commits 755a7a40..1dc256f8) through the review pipeline so local main can return to origin and the fleet updater can run. Includes: unfenced drafts measured (sp-88029341), last-post score surfaced (sp-08c34cf1), instance-side drift page.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #219, opened 2026-08-19
voice-stop-gate: the uncontested half of ASK-902 -- drain surface, alert truth, hook hygiene

assafkip:
Split of #217 per review consensus after seven rounds: extraction (the contested feature) stays on ask-902-train; this branch carries only what no round contested. Details in the commit message. Unblocks tonight's fleet update without shipping a weakened lint or the contested extractor.

### DIFF
--- a/.claude/rules/founder-notifications.md
+++ b/.claude/rules/founder-notifications.md
@@ -1,42 +1,38 @@
 # Founder Notifications (ENFORCED)
 
-Every proactive founder-facing ping/alert goes through ONE channel:
+## The one truth about slack-notify.sh (founder-directed 2026-08-10)
 
-```
-bash q-system/.q-system/scripts/slack-notify.sh "one concise line"
-```
+`q-system/.q-system/scripts/slack-notify.sh` is **THE FLEET ALERT PATH. It files
+a Linear ticket for Sana. It pages nobody.** Verbatim founder directive, in the
+script's own header: "I dont want to see any of these. Any of the ones that
+need attention should go to Sana - not me."
 
-This posts to a Slack Incoming Webhook (`~/.config/kipi/slack-webhook`, gitignored
-secret, or `$KIPI_SLACK_WEBHOOK`). Reliable, reaches the founder's phone, works
-headless. Silent no-op if the webhook isn't configured, so callers never break.
+This file previously described slack-notify.sh as the founder-ping channel.
+That was true before 2026-08-10 and STALE afterward; on 2026-08-18 a feature
+shipped against the stale description and its "founder page" landed in Sana's
+queue. The rule doc now matches the script.
 
-## Why this is the only channel
+## Routing
 
-`osascript` / macOS desktop notifications are permission-gated and **silently dropped**
-from a sandboxed or background process (no error, just nothing). They are BANNED for
-founder pings. Slack is the single reliable path.
+- **Engineering signals** (a job failed, a counter drifted, a loop closed, a
+  gate went red): `bash q-system/.q-system/scripts/slack-notify.sh "one line"`.
+  It becomes a Linear ticket in Sana's triage. Silent no-op if unconfigured.
+- **Founder-facing pings** are the EXCEPTION, not the default. They exist only
+  for things the founder explicitly asked to be told about, and they go
+  through his webhook directly -- wiring one is his call per instance, never a
+  default a feature ships with.
+- osascript / desktop notifications remain BANNED for any alert: silently
+  dropped from sandboxed processes.
 
-## When to ping (and when not to)
+## When to alert (and when not to)
 
-Ping for things the founder would want to know **when they may be away**:
-- An autonomous run (heartbeat, scheduled job) failed, timed out, or hit BLOCKED.
-- Something needs a founder decision before work can continue.
-- A meaningful state change on a tracked item (a maintainer replied; a loop closed; a PR landed).
-- A sycophancy/anti-drift alert that fires while they're away.
-
-Do NOT ping for routine progress, things they're clearly watching live in-session, or
-status that hasn't changed ("still waiting" every cycle is noise, not a ping).
-
-## For the agent (interactive sessions)
-
-When you'd otherwise reach for a desktop notification or `PushNotification` to alert the
-founder about something time-sensitive and they may be away, use `slack-notify.sh` instead.
-In-session alerts the founder is watching live don't need a Slack ping.
+Alert Sana's queue for: autonomous-run failures, BLOCKED runs, drift
+detections, dead loops. Do NOT alert for routine progress or per-event noise;
+alert on state change, once.
 
 ## Wiring
 
-- Producer of pings: any script/agent that detects a founder-relevant event.
-- The single sink: `q-system/.q-system/scripts/slack-notify.sh`.
-- Already wired: the open-loops heartbeat (`open-loops-heartbeat.sh`) — on a meaningful
-  change AND on an agent run failure/timeout.
-- New always-on / scheduled alert emitters MUST route through `slack-notify.sh`, never osascript.
+- The single sink stays `slack-notify.sh`; its destination is Sana's Linear
+  triage via `alert-to-linear.py`.
+- Emitters: open-loops heartbeat, the corpus-similarity drift counter
+  (voice-stop-gate.py `authorship_page`), and any new always-on job.
--- a/.claude/settings.json
+++ b/.claude/settings.json
@@ -122,6 +122,15 @@
             "timeout": 35
           }
         ]
+      },
+      {
+        "matcher": "startup|resume|clear|compact",
+        "hooks": [
+          {
+            "type": "command",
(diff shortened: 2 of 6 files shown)


assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-219-20260818-173157.md` (8643 bytes).

```
FINDINGS:
minor|authorship_page deletes the pending drift page before sending and never checks the send result, so any nonzero slack-notify.sh exit destroys the alert silently|q-system/.q-system/scripts/voice-stop-gate.py:288
minor|The shipped test file invokes the real deployed gate, whose authorship_page consumes live reporter state, so a capability-gate run destroys a pending alert in any instance where the reporter resolves|q-system/.q-system/scripts/test_voice_stop_gate_drain_only.py:36
minor|--drain-only is documented and tested as surface-only but performs an outbound Linear alert write on every SessionStart including compact, and authorship_page has no test coverage|q-system/.q-system/scripts/voice-stop-gate.py:395
minor|New comment claims 'response' and the bare copy-paste alternative were removed from the publish marker; both are still in the pattern and still match|q-system/.q-system/scripts/voice-stop-gate.py:52
END FINDINGS
```

--- reviewer output ---

The Stop gate voice-linted my pasted terminal output as if it were prose. Rewording captured evidence would falsify it, so the fenced evidence blocks carry the bypass marker. Review content is unchanged.

---

# PR #219 — voice-stop-gate drain / alert truth / hook hygiene

Read the diff, the full `voice-stop-gate.py`, `slack-notify.sh`, `alert-to-linear.py`, and the hook wiring in both settings files. Ran the new test file, mutation-tested it, and built two hermetic instance trees to exercise the paths the repo's own suite never reaches.

---

## Findings

### minor — `authorship_page()` deletes the pending drift alert before the send, and never checks whether the send worked

`q-system/.q-system/scripts/voice-stop-gate.py:288-299`

`subprocess.run(argv, "--drain-page")` consumes the page, then the send is a detached `Popen` with `stdout/stderr=DEVNULL` and no `wait()`/returncode read. `slack-notify.sh` documents three nonzero exits (`1` failed, `3` no Linear API key configured — a normal setup state, `4` fixture run). On any of them the alert is gone with no trace. The author's own comment at :276-280 closed exactly this hole for the missing-script case only.

Reproducer — instance tree, stub reporter implementing the delete-on-read contract the PR's own comment asserts (`"--drain-page deletes what it returns"`, :277-279), stub `slack-notify.sh` exiting 1:

```
<!-- voice-lint-skip -->
run1 rc= 0  stdout= ''  stderr= ''
pending page still on disk after run1: False
sends.log: 'SEND ATTEMPTED: corpus similarity has not moved in 14 days -- reconciliation drift\n'
run2 rc= 0  stdout= ''
sends.log after run2: 'SEND ATTEMPTED: corpus similarity has not moved in 14 days -- reconciliation drift\n'
```

Send exited 1, gate exited 0 silently, page gone, second run has nothing left.

**Severity note:** dropped from major per the calibration rule. The producer (`q-consult/pipeline/authorship_stop_report.py`) is not in this repo, so I stubbed it to the contract the PR's comment states. If the real one does delete on read, this is major in `consulting`.

---

### minor — the shipped test file consumes live alert state in any instance where the reporter resolves

`q-system/.q-system/scripts/test_voice_stop_gate_drain_only.py:36-39`

`_run()` invokes the *real deployed* gate at `Path(__file__).parent / "voice-stop-gate.py"`. Three tests hit it, and every one now routes through `finish_ok()` → `authorship_page()` → the real reporter's `--drain-page`. `alert-to-linear.py:908` then refuses on `PYTEST_CURRENT_TEST` — after the page is already deleted. `kipi update` ships both this file and its `capability-manifest.json` entry to every instance, so `kipi check` runs it in `consulting`, where `resolve_reporter` finds the reporter in-repo.

Reproducer — consulting-shaped tree (`q-consult/pipeline/authorship_stop_report.py` present), pending page on disk, run the shipped test file with pytest:

```
<!-- voice-lint-skip -->
pending drift alert on disk BEFORE the gate runs: True
2 failed, 5 passed in 0.43s
pending drift alert on disk AFTER the capability-gate test run: False
```

(The 2 failures are my rig — I copied 4 scripts, not the whole dir. The page deletion is the finding.)

This is the inverse of the scar `slack-notify.sh:94-99` records ("a test must never open a live ticket") and the exact shape `.claude/rules/fable-discipline-auto-invoke.md` says the fable-discipline lint blocks. Same producer caveat as above.

---

### minor — `--drain-only` is documented and tested as surface-only, but performs an outbound alert write

`voice-stop-gate.py:395` calls it `SURFACE-ONLY MODE`; `test_voice_stop_gate_drain_only.py:12-17` says it "surfaces a pending line as `systemMessage` … and exits 0" and "costs nothing." It also fires `slack-notify.sh` → a Linear ticket, on every SessionStart — and the matcher is `startup|resume|clear|compact`, so mid-session compaction now triggers an outbound write. `authorship_page()` has no test at all; the two hermetic stub tests copy the gate to a tmp dir with no `slack-notify.sh`, so that branch returns at :280 and is never exercised.

Same reproducer as finding 1: `sends.log` is written on a `--drain-only` invocation.

---

### minor — the new comment on `_NOUN` states two removals the code does not make

`voice-stop-gate.py:52-55` says `'response' left the set in round 7` and `the bare copy-paste alternative below is gone`. Both are still in the pattern (`:57`, `:66`). The comment describes the design of the *other* half of the split (`ask-902-train`, where the marker gates extraction) and rode across unchanged.

```
<!-- voice-lint-skip -->
COMMENT SAYS: response left the set in round 7
  actual: response in _NOUN == True
COMMENT SAYS: the bare copy-paste alternative below is gone
  actual: copy-paste alternative still in pattern == True
   "here's the response payload from the api" -> MATCHES "here's the response"
   'just copy-paste this into the terminal' -> MATCHES 'copy-paste'
   'here is the response object' -> MATCHES 'here is the response'
```

The two phrases the comment calls engineering prose still open the lint sweep.

---

## What is sound

- **"The new tests can't fail."** They can. Removing the `--drain-only` branch reds two of them, and the `__main__` block propagates it: `MUTANT rc= 1 … 2 failed, 4 passed`. The `pytest.main` wrapper at :193-203 is a real fix, not decoration.
- **"`--drain-only` might still reach the lint."** It does not. The branch is at `main():420`, above `json.load(sys.stdin)`, so a violating payload never gets parsed. Verified by running the real gate against a violating transcript: rc 0, empty stderr.
- **"The Stop path got weakened by the split."** It did not. Same transcript through the no-flag path returns rc 2.
- **"The rule doc rewrite is another doc claim."** It is not. `founder-notifications.md` now matches `slack-notify.sh` verbatim — no webhook call anywhere in that script, `alert-to-linear.py` is the only writer, the header directive is quoted correctly. This is the one place in the PR where the docs got *more* true.
- **"The two settings files drifted."** `settings-template-sync-check.py` exits 0. Both carry the identical hook block.
- **"The manifest reorder hides a dropped entry."** No: `test-plist-drift.py` and `test-converge-capout.sh` moved into sorted position, both still present, plus one net-new entry.
- **"Fleet blast radius."** In an instance with no reporter (the fleet default), `--drain-only` returns at `_reporter_argv() is None` with zero subprocesses and no output. The change is inert everywhere except `consulting` / `social-voice`.
- **Dropped for lack of a repro:** whether SessionStart actually renders a top-level `systemMessage` to the user. That is the load-bearing premise of the whole PR, and the SessionEnd half was removed for failing exactly that test — but the CLI bundle is outside this sandbox and I could not verify it either way. Worth an explicit check by someone who can.

---

## VERDICT: APPROVE WITH NITS

Every finding reproduced, none is permanent or unrecoverable in this repo, and the fleet default path is inert. The PR does what it says: it splits out the uncontested half and the rule-doc correction is genuinely correct.

**Fix first:** re-queue the page when the send fails. `authorship_page()` currently deletes then fires and forgets — read the child's exit code (or have the reporter hold the page until an explicit ack) so exit 3 "no Linear API key" does not silently eat every drift alert in an unconfigured instance. That one change also closes the test-eats-live-state finding.

FINDINGS:
minor|authorship_page deletes the pending drift page before sending and never checks the send result, so any nonzero slack-notify.sh exit destroys the alert silently|q-system/.q-system/scripts/voice-stop-gate.py:288
minor|The shipped test file invokes the real deployed gate, whose authorship_page consumes live reporter state, so a capability-gate run destroys a pending alert in any instance where the reporter resolves|q-system/.q-system/scripts/test_voice_stop_gate_drain_only.py:36
minor|--drain-only is documented and tested as surface-only but performs an outbound Linear alert write on every SessionStart including compact, and authorship_page has no test coverage|q-system/.q-system/scripts/voice-stop-gate.py:395
minor|New comment claims 'response' and the bare copy-paste alternative were removed from the publish marker; both are still in the pattern and still match|q-system/.q-system/scripts/voice-stop-gate.py:52
END FINDINGS

PR #220, opened 2026-08-19
chore: remove tracked .sana-tmp agent scratch

assafkip:
The founder deleted ~/projects/kipi-system/.sana-tmp in the 2026-08-18 cleanup; four of its files were git-tracked, so the deletion needs a commit for local main to return to origin. Scratch only: triage notes and message drafts, no callers.


assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-220-20260818-175837.md` (5795 bytes).

```
FINDINGS:
minor|.sana-tmp/ is not gitignored so the same agent scratch can be re-committed to the public repo by any future git add -A|.gitignore:156
minor|PR body claims "no callers" but test-kipi-update-dry-tagging.sh calls .sana-tmp/mkfleet.sh and now SKIPs its headline assertion while still exiting PASS|test-kipi-update-dry-tagging.sh:107
END FINDINGS
```

--- reviewer output ---

## PR #220 review — `chore: remove tracked .sana-tmp agent scratch`

Pure deletion: 4 files, 0 additions, 2519 lines removed. I attacked the deletion itself (recoverability, secrets, callers) and the claim in the PR body.

---

### Finding 1 — minor: `.sana-tmp/` is still not gitignored, so the identical accident recurs

`.gitignore:155-156` ignores the sibling scratch dirs `.review-scratch/` and `.fable-wt/`. `.sana-tmp/` is absent. The PR removes the 2519 lines but leaves the path stageable, so the next agent that writes scratch there and runs `git add -A` republishes it to a public repo whose history cannot be rewritten (force-push is hook-blocked fleet-wide).

This is not a hypothetical producer: the 4 files being deleted *are* this accident having already happened once.

The repo already wrote the doctrine for exactly this class at `.gitignore:167-168`:

> "Observed on PR #141 (ASK-700): a refusal was written here and swept in by `git add -A`. **The ignore is the fix rather than a review checklist** -- a path `git add -A` cannot stage cannot be committed by any future agent on any future branch."

Every other handler treats `.sana-tmp` as junk-to-skip but none stops it being *tracked* — `pytest.ini:50`, `capability-map-gen.py:403`, `repo-preflight.sh:293`, `notify-callsite-audit.py:48`.

**Reproducer:**
```
git check-ignore -v .sana-tmp/newscratch.md; echo "rc_new=$?"
```
**Real output:**
```
rc_new=1
```
(rc=1 = not ignored, no output = no matching pattern.) For contrast, `.gitignore:155-156` carry `.review-scratch/` and `.fable-wt/`.

Severity dropped a level on purpose: no *script* in this repo emits into `.sana-tmp` — the producer is an ad-hoc agent session. Demonstrated, but not a scripted path.

**Anchor:** `.gitignore:156`

---

### Finding 2 — minor: the PR body's "no callers" is false; `.sana-tmp` has a caller, and that test now silently skips its headline assertion

PR body: *"Scratch only: triage notes and message drafts, no callers."*

`test-kipi-update-dry-tagging.sh:107` calls `"$REAL/.sana-tmp/mkfleet.sh"`, where `REAL` resolves to the repo root (line 20). With `.sana-tmp` gone, the test's own reason-for-existing assertion — "a real dry run leaves no untagged mutating line", guarding scar sp-46c73c76 where `--dry-run` genuinely committed into live instances — degrades to a SKIP and the file still exits 0.

**Reproducer:**
```
bash test-kipi-update-dry-tagging.sh 2>&1 | tail -20
bash test-kipi-update-dry-tagging.sh >/dev/null 2>&1; echo "EXIT_CODE=$?"
```
**Real output:**
```
=== the helpers tag ONLY in dry mode ===
  OK: dry mode tags both the status line and the piped line
  OK: a real run is untagged
  OK: a real run still emits both lines unchanged

=== dry_filter must not swallow the commit's exit code ===
  OK: PIPESTATUS[0] preserves the upstream failure
  OK: guarded_commit reads PIPESTATUS[0], not $?

=== a real dry run leaves no untagged mutating line ===
  SKIP: fixture builder unavailable (.sana-tmp/mkfleet.sh); unit checks above still ran

PASS
EXIT_CODE=0
```

Kept at minor, not major, for three reasons I verified rather than assumed:
- `mkfleet.sh` was **never tracked** — `git log --all --pretty=format: --name-only -- .sana-tmp` returns only the 4 `.md` files. Any fresh clone always skipped. Pre-existing, not introduced here.
- The test is registered nowhere: `grep -rn "test-kipi-update-dry-tagging" --exclude-dir=.git .` returns only its own line-17 comment. No CI, no manifest, no required check consumes the PASS.
- The unit half (say/dry_filter, both directions, PIPESTATUS) still runs and still asserts.

What this PR actually owns is the false "no callers" claim. The fix is either tracking `mkfleet.sh` into the repo as a real fixture, or making the missing builder a FAIL instead of a SKIP.

**Anchor:** `test-kipi-update-dry-tagging.sh:107`

---

### What is sound — attacks the change survived

- **Secrets in the deleted content.** Scanned all 2543 diff lines for `sk-`, `ghp_`, `lin_api_`, `xox[bapr]-`, `AKIA`, `PRIVATE KEY`, `Bearer <token>`. Positive control on a planted `ghp_` token matched `1`; the real diff matched `0`. The scan can fail; it didn't.
- **Hidden additions.** `grep '^+' | grep -v '^+++'` on the diff returns nothing. All four headers are `deleted file mode 100644`. It is genuinely 0-addition.
- **Dangling references to the deleted files.** `grep -rn "c_last\|c_prev\|r12\.md\|r13\.md"` across `*.py *.sh *.md *.json` — zero hits. Nothing reads them.
- **Recoverability.** Content stays in history at `6a0dab05^`; rollback is a `git revert`. Not destructive, nothing permanent lost.
- **The `.sana-tmp` mentions in live code are dir-name skip patterns**, not path reads — `pytest.ini` norecursedirs, `capability-map-gen.py:403` regex, `repo-preflight.sh:293` `_shipping()`, `notify-callsite-audit.py:48` SKIP_DIR. A missing directory breaks none of them; I read each.
- **Auto-merge squash on a pure deletion** — no partial-failure or concurrency surface. Two of these running at once is a no-op on the second.

---

Only minor findings survived reproduction. Nothing here is unrecoverable and nothing pages anyone.

**Fix first:** add `.sana-tmp/` next to `.review-scratch/` and `.fable-wt/` at `.gitignore:156`. One line, and it is the difference between this PR cleaning up an accident and this PR cleaning up an accident that will happen again.

FINDINGS:
minor|.sana-tmp/ is not gitignored so the same agent scratch can be re-committed to the public repo by any future git add -A|.gitignore:156
minor|PR body claims "no callers" but test-kipi-update-dry-tagging.sh calls .sana-tmp/mkfleet.sh and now SKIPs its headline assertion while still exiting PASS|test-kipi-update-dry-tagging.sh:107
END FINDINGS

PR #223, opened 2026-08-19
fix(drivers): brace-wrap converge.sh and linear-worker.sh so a mid-run repo edit cannot corrupt them (ASK-351)

assafkip:
Closes ASK-351 (do not merge yet -- closeout runs through /issue-verify and /issue-closeout).

## The defect

bash does not load a script into memory. It reads a chunk, executes ONE command, then lseeks back to the byte offset just past that command for the next one. `converge.sh` runs for hours and `linear-worker.sh` for up to 1800s per round, inside a repo that agents edit the whole time. An edit shifts every later byte offset, bash resumes parsing mid-string, and bare words inside quoted strings become commands.

Measured 2026-08-03, `~/.config/kipi/converge-ASK-288.log`, four full review rounds completed and then discarded at the exit:

```
converge.sh: line 872: of: command not found
converge.sh: line 872: not: command not found
converge.sh: line 876: syntax error near unexpected token `fi'
```

Line 872 on disk is `fi`. The words `of` and `not` appear only INSIDE the quoted STALL_LOG/STALL_PAGE strings at 868 and 871, so no correct read can execute them, and `bash -n` passes on every committed version. Commit `d142466` landed seven minutes into that run. `linear-worker.log` carries the same signature (`ial: command not found`, and `ial` is not a word in that file), so this is a class across both drivers.

## The fix

Wrap each script body in a `{ ... }` compound command. bash must parse a compound command to completion before executing any of it, so the body is consumed at startup and a later edit cannot reach it. The existing unconditional `exit` stays as the last statement inside the brace.

**No re-indentation.** The DoR flagged an indentation-only reformat of a 49KB and a 116KB file as the concentrated risk. The diff here is two brace lines per file plus scar comments, so the wrap is auditable by eye instead of buried in a whitespace churn.

**Both halves are load-bearing, and the reproducer holds them separately.** A brace wrap ALONE still fails: the body runs and prints its result, then bash seeks past the closing brace looking for one more command, lands at a stale offset in a file that has grown, and re-executes. Measured in case 2 -- the whole body prints TWICE, then rc=2.

## New test

`q-system/.q-system/scripts/test/test-script-stable-under-self-edit.sh`, 7 cases, in the capability manifest.

- **case 0** negative self-test: the same unwrapped script with NO edit is clean. Without this a harness that always reports red would look like a passing reproducer.
- **case 1** the defect: unwrapped + mid-run edit corrupts.
- **case 2** the half fix: braces without a trailing `exit` still fail, rc=2.
- **case 3** the fix: braces + trailing `exit` is rc=0 and clean under the same edit.
- **case 4** the two real drivers carry the shape and pass `bash -n`.
- **case 5** mutation: the case-4 predicate is run against an unwrapped copy of the real `converge.sh` and must reject it, so it is a check that can actually go red.

The edit is an INSERT near the top, not an append, because that is what a commit does -- `d142466` changed lines in the middle and every byte after them moved. An append alone leaves earlier offsets valid and would under-report.

## Evidence

Before the fix, the same suite (case 4 is the only thing that changed):

```
case 4: the two real drivers carry the shape and still parse
  FAIL - converge.sh: body does not open with a bare { (found: set -uo pipefail)
  FAIL - linear-worker.sh: body does not open with a bare { (found: set -uo pipefail)
passed: 5  failed: 2
```

After:

```
passed: 7  failed: 0
```

Regression suite named in the DoR, run against the wrapped script (a real run, not a parse). The DoR said 17/17; the suite has grown since it was written:

```
bash q-system/.q-system/scripts/test/test-converge.sh
PASS: 21/21 converge checks
```

Parse check:

```
bash -n q-system/.q-system/scripts/converge.sh       -> clean
bash -n q-system/.q-system/scripts/linear-worker.sh  -> clean
```

## Blast radius

Fleet-wide as scoped. `kipi-update.sh:1295` rsyncs the whole `q-system/` tree into every registered instance, so both rewritten scripts and the manifest line land everywhere on the next update. No `.claude/rules/` file and no `settings.json` hook wiring is touched, so no always-on rule changes.

## Not doing (per the DoR, binding)

- Not switching to snapshot-and-exec from `/tmp`. `converge.sh:30` derives `SCRIPT_DIR` from `BASH_SOURCE`, and `SKEL`, `NOTIFY` and `pr-verdict-lib.sh` all hang off it; re-execing a copy would silently repoint every one of them.
- Not auditing or wrapping the other long-running bash drivers in `q-system/.q-system/scripts/` for the same class.

## Reviewer, two things worth your eye

1. The `exit` inside the brace has to stay **last** and **unconditional** (column 0). Both scripts already ended that way; a future edit that adds a statement below it, or indents it into an `if`, silently removes half the fix. Case 4's predicate checks column 0 for exactly this reason.
2. `capability-manifest.json` picked up 19 open spillover notes and the two drivers 37 and 69 more when the ratchet fired on open. None are triaged here -- triaging 125 notes is not this issue's scope and the DoR's not-doing line is binding. Flagging so it is captured rather than silently passed over.

### DIFF
--- a/q-system/.q-system/capability-manifest.json
+++ b/q-system/.q-system/capability-manifest.json
@@ -424,6 +424,11 @@
       "runner": "bash"
     },
     {
+      "path": "q-system/.q-system/scripts/test/test-script-stable-under-self-edit.sh",
+      "runner": "bash",
+      "timeout_s": 120
+    },
+    {
       "path": "q-system/.q-system/scripts/test/test-styles-csv-width.sh",
       "runner": "bash"
     },
--- a/q-system/.q-system/instruction-budget-baseline.json
+++ b/q-system/.q-system/instruction-budget-baseline.json
@@ -1,3 +1,3 @@
 {
-  "total_always_on": 512
+  "total_always_on": 511
 }
--- a/q-system/.q-system/scripts/converge.sh
+++ b/q-system/.q-system/scripts/converge.sh
@@ -1014,8 +1014,10 @@ if ! python3 "$LEDGER" "$ATTEMPTS" record-capout "$ISSUE" "$CAPOUT_WHY" 2>>"$LOG
   say "WARNING: record-capout failed for $ISSUE; the redrives cannot see this cap-out."
 fi
 bash "$NOTIFY" "converge $ISSUE: hit $MAX_ROUNDS-round cap, still $LAST_VERDICT. Parked -- no redrive will re-enter it until you clear it: python3 q-system/.q-system/scripts/attempts-ledger.py $ATTEMPTS clear-capout $ISSUE$CAPOUT_NOTE" 2>/dev/null || true
-# The `exit 2` above is the last statement INSIDE the ASK-351 brace, and it has to
-# stay last and stay unconditional: it is what stops bash from ever reading this
-# file again. See the header. Nothing may be added below the closing brace.
+# The `exit 2` on the line BELOW this comment -- not the `bash "$NOTIFY"` above it --
+# is the last statement INSIDE the ASK-351 brace, and it has to stay last and stay
+# unconditional: it is what stops bash from ever reading this file again. See the
+# header. Nothing may be added between it and the closing brace, and nothing may be
+# added below the closing brace.
 exit 2
 }
--- a/q-system/.q-system/scripts/converge.sh
+++ b/q-system/.q-system/scripts/converge.sh
@@ -25,6 +25,34 @@
 #                        (1800s work) and the reviewer (2400s review)
 #
 # Usage: converge.sh --issue ASK-150 [--max-rounds 4] [--dry]
+#
+# THE BRACE AROUND EVERYTHING BELOW IS LOAD-BEARING (ASK-351). Do not remove it,
+# and do not "clean up" the bare `}` on the last line.
+#
+# bash does not load a script into memory. It reads a chunk, executes ONE command,
+# then lseeks back to the byte offset just past that command for the next one.
+# This script runs for hours inside a repo that agents edit the whole time, so an
+# edit shifts every later byte offset and bash resumes parsing mid-string. Measured
+# 2026-08-03, ~/.config/kipi/converge-ASK-288.log, four completed review rounds
+# thrown away at the exit:
+#
+#   converge.sh: line 872: of: command not found
+#   converge.sh: line 872: not: command not found
+#   converge.sh: line 876: syntax error near unexpected token `fi'
+#
+# Line 872 on disk is `fi`. `of` and `not` exist only INSIDE the quoted STALL_LOG
+# strings, so no correct read can execute them, and `bash -n` passes on every
+# committed version -- the file was never syntactically wrong. Commit d142466
+# landed seven minutes into that run. That is the edit.
+#
+# bash must parse a compound command to completion before executing any of it, so
+# the brace makes the whole body arrive at startup. The `exit 2` on the last line
+# INSIDE the brace is the other half and is not redundant: measured, a brace wrap
+# alone still dies rc=2, because bash seeks past the closing brace looking for one
+# more command and re-executes leftovers from a file that has grown.
+#
+# Reproducer for both halves: q-system/.q-system/scripts/test/test-script-stable-under-self-edit.sh
+{
 set -uo pipefail
 
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
@@ -986,4 +1014,8 @@ if ! python3 "$LEDGER" "$ATTEMPTS" record-capout "$ISSUE" "$CAPOUT_WHY" 2>>"$LOG
   say "WARNING: record-capout failed for $ISSUE; the redrives cannot see this cap-out."
 fi
 bash "$NOTIFY" "converge $ISSUE: hit $MAX_ROUNDS-round cap, still $LAST_VERDICT. Parked -- no redrive will re-enter it until you clear it: python3 q-system/.q-system/scripts/attempts-ledger.py $ATTEMPTS clear-capout $ISSUE$CAPOUT_NOTE" 2>/dev/null || true
+# The `exit 2` above is the last statement INSIDE the ASK-351 brace, and it has to
+# stay last and stay unconditional: it is what stops bash from ever reading this
+# file again. See the header. Nothing may be added below the closing brace.
 exit 2
+}
--- a/q-system/.q-system/scripts/linear-worker.sh
+++ b/q-system/.q-system/scripts/linear-worker.sh
@@ -44,6 +44,27 @@
 #
 # Usage:  linear-worker.sh [--apply] [--limit N] [--issue ASK-123]
 # Dry by default: prints what it would pick and stops.
+#
+# THE BRACE AROUND EVERYTHING BELOW IS LOAD-BEARING (ASK-351). Do not remove it,
+# and do not "clean up" the bare `}` on the last line.
+#
+# bash does not load a script into memory. It reads a chunk, executes ONE command,
+# then lseeks back to the byte offset just past that command for the next one. A
+# round here runs up to 1800s inside a repo that agents edit the whole time, so an
+# edit shifts every later byte offset and bash resumes parsing mid-string.
+# ~/.config/kipi/linear-worker.log carries the signature: `ial: command not found`,
+# and `ial` is not a word in this file -- it is the tail of one, read from a
+# slipped offset. converge.sh died the same way on 2026-08-03 and threw away four
+# completed review rounds.
+#
+# bash must parse a compound command to completion before executing any of it, so
+# the brace makes the whole body arrive at startup. The `exit 0` on the last line
+# INSIDE the brace is the other half and is not redundant: measured, a brace wrap
+# alone still dies rc=2, because bash seeks past the closing brace looking for one
+# more command and re-executes leftovers from a file that has grown.
+#
+# Reproducer for both halves: q-system/.q-system/scripts/test/test-script-stable-under-self-edit.sh
+{
 set -uo pipefail
 
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
@@ -2300,4 +2321,8 @@ json.dump(d,open('$ATTEMPTS','w'),indent=2); print(e['rounds'])" 2>/dev/null || 
 done
 
 say "worker: run complete"
+# The `exit 0` below is the last statement INSIDE the ASK-351 brace, and it has to
+# stay last and stay unconditional: it is what stops bash from ever reading this
+# file again. See the header. Nothing may be added below the closing brace.
 exit 0
+}
--- a/q-system/.q-system/scripts/test/test-script-stable-under-self-edit.sh
+++ b/q-system/.q-system/scripts/test/test-script-stable-under-self-edit.sh
@@ -1,0 +1,212 @@
+#!/usr/bin/env bash
+# Reproducer + acceptance for ASK-351: a long-running bash driver corrupts itself
+# when the repo it lives in is edited mid-run.
+#
+# THE DEFECT, measured 2026-08-03. bash does not load a script into memory. It
+# reads a chunk, executes one command, then lseeks back to the byte offset just
+# past that command for the next one. `converge.sh` runs for hours and
+# `linear-worker.sh` for up to 1800s per round, inside a repo that agents edit the
+# whole time. An edit shifts every later byte offset, bash resumes parsing mid
+# string, and bare words inside quoted strings become commands:
+#
+#   converge.sh: line 872: of: command not found
+#   converge.sh: line 872: not: command not found
+#   converge.sh: line 876: syntax error near unexpected token `fi'
+#
+# Line 872 on disk is `fi`. The words `of` and `not` appear only INSIDE the
+# double-quoted STALL_LOG/STALL_PAGE strings at 868 and 871, so no correct read
+# can execute them. `bash -n` passes on every committed version -- the file was
+# never syntactically wrong. Commit d142466 landed seven minutes into that run.
+# ~/.config/kipi/linear-worker.log carries the same signature (`ial: command not
+# found`), so this is a class across both drivers, not one bug.
+#
+# THE FIX UNDER TEST is a `{ ... }` compound command around the whole body with an
+# unconditional `exit` as its last statement. bash must parse a compound command
+# to completion before executing any of it, so the body is consumed at startup.
+#
+# WHY THE TRAILING `exit` IS NOT REDUNDANT, and why case 2 exists as its own case:
+# a brace wrap ALONE still fails. The body runs and prints its result, then bash
+# seeks past the closing brace looking for one more command, lands at a stale
+# offset in a file that has grown, and re-executes leftovers. Measured here: the
+# braces-only case prints the whole body TWICE and dies rc=2.
+#
+# HOW THE CASES ARE BUILT. Cases 0-3 are synthetic scripts, not the real drivers:
+# the real ones need a live Linear, a gh token and hours of wall clock, and a test
+# that drove them would be measuring the fleet rather than this property. What is
+# synthetic is the BODY; the mechanism (in-place edit that shifts offsets under a
+# running bash) is the real one. Case 0 is the negative self-test -- the same
+# unwrapped script with NO edit must come out clean, or a harness that always
+# reports red would look like a passing reproducer. Case 4 then holds the two real
+# scripts to the shape, and case 5 mutates that structural check against an
+# unwrapped copy so it is a check that can actually go red.
+set -uo pipefail
+
+ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../../../.." && pwd)"
+CONVERGE="$ROOT/q-system/.q-system/scripts/converge.sh"
+WORKER="$ROOT/q-system/.q-system/scripts/linear-worker.sh"
+for f in "$CONVERGE" "$WORKER"; do
+  [ -f "$f" ] || { echo "FATAL: missing $f" >&2; exit 1; }
+done
+
+WORK="$(mktemp -d)"
+cleanup() { [ -n "${WORK:-}" ] && [ -d "$WORK" ] && find "$WORK" -mindepth 1 -delete && rmdir "$WORK"; }
+trap cleanup EXIT
+PASS=0; FAIL=0
+ok()  { PASS=$((PASS+1)); echo "  ok   - $1"; }
+bad() { FAIL=$((FAIL+1)); echo "  FAIL - $1"; }
+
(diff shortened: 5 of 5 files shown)


assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-223-20260819-104150.md` (8658 bytes).

```
FINDINGS:
major|pr-review-agent.sh is the third long-running driver in the same pipeline (2400s, launched by linear-worker.sh:93) and is left unwrapped, and the new test's hardcoded two-file loop cannot detect it|q-system/.q-system/scripts/pr-review-agent.sh:1
minor|The scar comment says the `exit 2` is "above" it when the `exit 2` is three lines below, pointing a future editor at `bash "$NOTIFY"` as the protected last statement|q-system/.q-system/scripts/converge.sh:1017
END FINDINGS
```

--- reviewer output ---

# Review: PR #223 — brace-wrap converge.sh and linear-worker.sh (ASK-351)

Read-only. Nothing in the repo was modified. Every probe ran in a `tempfile.mkdtemp` copy.

---

## Finding 1 — major

**The class the PR names is not closed. `pr-review-agent.sh` is the third long-running bash process in the same pipeline, has the longest exposure window of the three, and is left unwrapped — and the new capability test hardcodes a two-file list, so it will never notice.**

`q-system/.q-system/scripts/pr-review-agent.sh:1` (unwrapped), gate at `q-system/.q-system/scripts/test/test-script-stable-under-self-edit.sh:189`

Evidence it is the same shape of thing:

- `converge.sh:25`, the header this PR edits, names the exposure itself: `(1800s work) and the reviewer (2400s review)`. The reviewer sits at 2400s. The worker this PR fixed sits at 1800s.
- `linear-worker.sh:93` launches it as its own bash process: `REVIEWER_CMD="${KIPI_PR_REVIEWER:-bash $SCRIPT_DIR/pr-review-agent.sh}"`.
- It is 65,567 bytes / 1108 lines, first body line `set -uo pipefail`, no brace.

Reproducer — same technique the PR's own test uses, on the file as shipped. `sleep 6` stands in for the ~2400s it really spends in claude/codex, followed by an immediate `exit 0` so the probe touches no network and no Linear. A padding insert lands at t=1.5s:

```
python3 -c 'import os,shutil,subprocess,tempfile,threading,time
H=chr(35); work=tempfile.mkdtemp(prefix="ask351-rev-")
shutil.copytree("q-system/.q-system/scripts", work+"/scripts", dirs_exist_ok=True)
p=work+"/scripts/probe-asis.sh"
d=open(work+"/scripts/pr-review-agent.sh").read()
open(p,"w").write(d.replace("set -uo pipefail\n","set -uo pipefail\nsleep 6\nexit 0\n",1))
PAD="".join(H+" INSERTED PADDING LINE %d AAAAAAAAAAAAAAAAAAAAAAAAAAAA\n"%i for i in range(1,9))
def e(x):
    time.sleep(1.5); b=open(x).read()
    open(x,"w").write(b.replace("set -uo pipefail\n","set -uo pipefail\n"+PAD,1))
t=threading.Thread(target=e,args=(p,)); t.start()
r=subprocess.run(["bash",p],capture_output=True,text=True,timeout=90); t.join()
print("rc=%d"%r.returncode); print((r.stdout+r.stderr).strip())'
```

Real output:

```
pr-review-agent.sh AS SHIPPED (unwrapped), edited mid-run -> rc=0
    PADDING: No such file or directory
```

`PADDING` is not a command anywhere in that file. bash resumed at a slipped offset and executed a fragment of an inserted comment. That is the ASK-351 signature verbatim.

What a human does at 3am: the same thing the scar describes. A completed review round is discarded, `converge.sh` charges an attempt, three of those hit the cap and `converge.sh:1016` parks the issue with `attempts-ledger.py ... clear-capout` for a human to run by hand.

Why this belongs on this PR and not a follow-up: the change ships a class-level claim ("this is a class across both drivers, not one bug") plus a fleet capability test registered in `capability-manifest.json`, and that test's `for f in "$CONVERGE" "$WORKER"` loop is a hardcoded allowlist of two. Adding the third member costs two lines in `pr-review-agent.sh` and one in the loop.

---

## Finding 2 — minor

**`converge.sh:1017` says the `exit 2` is *above* the comment. It is below it. The comment points a future editor at the wrong statement — the exact reader this comment exists to stop.**

`q-system/.q-system/scripts/converge.sh:1017`

```
sed -n '1014,1021p' q-system/.q-system/scripts/converge.sh
```

Real output:

```
fi
bash "$NOTIFY" "converge $ISSUE: hit $MAX_ROUNDS-round cap, still $LAST_VERDICT. Parked -- ..."
# The `exit 2` above is the last statement INSIDE the ASK-351 brace, and it has to
# stay last and stay unconditional: it is what stops bash from ever reading this
# file again. See the header. Nothing may be added below the closing brace.
exit 2
}
```

The statement above the comment is `bash "$NOTIFY" ...`, which is conditional-ish (`|| true`) and is not what protects anything. The twin comment in `linear-worker.sh:2324` gets it right and says "below". One word.

---

## What is sound

Attacks I ran that the code survived. Naming them because a review that only lists faults is not calibrated.

**The fix holds on the real 49KB converge.sh, not just on the 7-line synthetic.** Case 4 is structural only, so I ran the behavioral version myself: two copies of the real file in a temp tree, `sleep 6` injected after `set -uo pipefail`, run with no arguments (the usage-error path — no network, no Linear), padding insert at t=1.5s.

```
work=/var/folders/.../ask351-fbyilu2z  mutant removed brace lines 55 and 1022

=== unwrapped.sh  rc=1
    PADDING: No such file or directory
    usage: converge.sh --issue ASK-nnn [--max-rounds N] [--dry]

=== wrapped.sh  rc=1
    usage: converge.sh --issue ASK-nnn [--max-rounds N] [--dry]
```

The shipped file is clean; strip the brace pair and the same file corrupts. The mechanism is real and the fix is real on the real artifact.

**The `assert_wrapped` predicate is not decoration on the branches case 5 does not mutate.** Case 5 only exercises "body must open with `{`". I mutated the other two by hand — deleted converge.sh's trailing `exit 2`, indented linear-worker.sh's `exit 0` — and re-ran the suite against that tree:

```
suite rc=1
case 4: the two real drivers carry the shape and still parse
  FAIL - converge.sh: last statement inside the brace is not an unconditional exit (found: # file again. See the header...)
  FAIL - linear-worker.sh: last statement inside the brace is not an unconditional exit (found:   exit 0)
passed: 5  failed: 2
```

Both go red, including the indentation case the comment specifically calls out. I was going to file this as a coverage gap and dropped it: the branch works, it is just untested. That is a nit at worst and I am not listing it.

**Existing regression suites are green under the line-number shift.** The wrap moves every line in both files by 21–28. `test-converge.sh` → `PASS: 21/21`. `test-terminal-states.sh` → `14 passed, 0 failed`. `test-worker-refusal.sh` → `50 passed, 0 failed`, including the SIGTERM-mid-run and orphan-reaper cases, so traps and job control still work inside the brace group.

**The new suite is green and its negative self-test is real.** `passed: 7  failed: 0`, 18.5s, well inside the `timeout_s: 120` the manifest sets. `timeout_s` is an established key on 27 other manifest entries, not an invented one.

**Sourced libs are not a hole.** Both drivers `.` source `pr-verdict-lib.sh` and `repo-slug-lib.sh` at line 84/119, inside the first second of the run; the functions are resident from then on. No live-run exposure.

**`kipi-dispatch.sh` is not a third victim.** It launches converge detached and exits within ~10s (`kipi-dispatch.sh:1378-1382`). Short-lived, no exposure. I checked before assuming.

**Load path confirmed.** `kipi:98` runs `bash "$KIPI_HOME/.../converge.sh"` — a seekable regular file from this repo, which is the copy edited here.

**`instruction-budget-baseline.json` 512 → 511** is a ratchet down, not a loosening. The audit already fails at 511/300 pre-existing and no rule file changed in this diff. Not a finding.

**Attacks I could not make fail, so I dropped them:** the timing margin in cases 1–3 (edit at t≈1s, window closes at t=3s) did not flake in repeated runs, and I could not construct a load condition that closed a 2-second gap without fabricating it; `bash -n` output leakage through `why="$(assert_wrapped …)"`; the case-5 awk picking a heredoc `}` instead of the real closing brace (all column-0 `}` in both files are function closers, and the last one is the real brace at converge.sh:1021 / linear-worker.sh:2328).

---

## VERDICT: REQUEST CHANGES

One major, one minor. Not a BLOCK — nothing here is permanent or unrecoverable, the PR strictly improves the two files it touches, and it is already marked "do not merge yet".

**Fix first:** brace-wrap `pr-review-agent.sh` with the same two lines, and add it to the case-4 loop at `test-script-stable-under-self-edit.sh:189`. It is the 2400s process in a pipeline whose two 1800s processes just got fixed, and the test as written will never tell you it is missing.

FINDINGS:
major|pr-review-agent.sh is the third long-running driver in the same pipeline (2400s, launched by linear-worker.sh:93) and is left unwrapped, and the new test's hardcoded two-file loop cannot detect it|q-system/.q-system/scripts/pr-review-agent.sh:1
minor|The scar comment says the `exit 2` is "above" it when the `exit 2` is three lines below, pointing a future editor at `bash "$NOTIFY"` as the protected last statement|q-system/.q-system/scripts/converge.sh:1017
END FINDINGS

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-223-20260819-111314.md` (8708 bytes).

```
FINDINGS:
minor|KNOWN_UNWRAPPED comment claims its spillover ids hold `gates run` red, but both are untriaged and land in the non-blocking REPORT bucket; the ledger is also untracked so the ids resolve to nothing on a fresh clone|q-system/.q-system/scripts/test/test-script-stable-under-self-edit.sh:92
minor|Cases 1/2 depend on a 2s timing margin with no assertion that the mid-run edit landed before the body exited; a late edit turns them into a false alarm and makes case 3's green vacuous|q-system/.q-system/scripts/test/test-script-stable-under-self-edit.sh:157
minor|PR body's Evidence section certifies a 6-case suite reporting `passed: 7`, while HEAD ships 8 cases and 13 assertions|q-system/.q-system/scripts/test/test-script-stable-under-self-edit.sh:167
minor|Driver discovery globs only q-system/.q-system/scripts/*.sh, so kipi-dispatch.sh matches the PR's own `claude -p` marker and is never evaluated; the marker also matches comments rather than invocations|q-system/.q-system/scripts/test/test-script-stable-under-self-edit.sh:74
END FINDINGS
```

--- reviewer output ---

Review complete. Tool budget hit on the cleanup step — four scratch probes remain under the gitignored `q-system/output/` (`ask351-probe.sh`, `ask351-scope.sh`, `ask351-race.sh`, `ask351-load.sh`); nothing tracked was touched.

# Review: PR #223 — brace-wrap converge.sh and linear-worker.sh (ASK-351)

## Findings

### minor — the exception ledger's enforcement claim overstates what the gate does
`q-system/.q-system/scripts/test/test-script-stable-under-self-edit.sh:88-94` says each `KNOWN_UNWRAPPED` id "is an open item in `.prd-os/spillover.jsonl` that holds `prd_runner.py gates run` red until a real issue closes it". Both ids exist and are open, but they are untriaged, so they land in the non-blocking `[REPORT]` bucket, not the RED set. Nothing holds red on them. Separately, `spillover.jsonl` is untracked, so on a fresh clone the ids resolve to nothing while case 4 still greens on the hardcoded string — and case 7, written so "the exception ledger cannot rot", validates the filename half and never the id half.

```
$ python3 plugins/prd-os/scripts/prd_runner.py spillover list | grep 'sp-3bcc9e16\|sp-c2dcebad'
[open] sp-3bcc9e16: pr-review-agent.sh (2400s, launched by linear-worker.sh:93) is unwrapped and car (src ASK-351)
[open] sp-c2dcebad: open-loops-heartbeat.sh is a SECOND unwrapped long-running driver carrying the A (src ASK-351)

$ python3 plugins/prd-os/scripts/prd_runner.py gates run | grep -o 'sp-3bcc9e16\|sp-c2dcebad'
                                  # empty: neither is in the RED blocking set
$ python3 plugins/prd-os/scripts/prd_runner.py gates run | grep '^\[RED\]'
[RED] spillover: 34 open item(s) this work must answer for: sp-10cf4f76, ... sp-88b29915

$ git ls-files .prd-os | grep -c spillover.jsonl
0
```

I first read this as a major (ids missing entirely). That was a truncated `head -10` on my side. Re-ran, both exist. Dropping the major is the correct outcome; what survives is the enforcement wording.

### minor — cases 1 and 2 ride a 2-second timing margin with nothing asserting the edit landed in the window
`test-script-stable-under-self-edit.sh:157-159` fires the edit at t=1s against a body that sleeps 3s. If the edit lands late, case 1 and case 2 report FAIL (a fleet-wide false alarm from a green tree) and case 3 reports its green without the edit ever having landed, so the fix is unproven by the case that certifies it.

Reproducer: shipped harness copied verbatim, only the edit delay varied.

```
### edit delay = 1s (body sleeps 3s; shipped suite uses 1s)
  case 1 -> ok   (reproduced the defect)
  case 2 -> ok   (half fix still fails)
  case 3 -> ok   (fix held)  <-- green regardless of whether the edit landed in time

### edit delay = 4s
  case 1 -> FAIL 'unwrapped run survived the edit -- the reproducer no longer reproduces'
  case 2 -> FAIL 'braces alone survived; ... this fix is half noise'
  case 3 -> ok   (fix held)  <-- green regardless of whether the edit landed in time
```

Severity dropped per the constructed-input rule: I tried to make it flake naturally and could not. 40 CPU burners on 10 cores plus a `find` loop, shipped suite unmodified:

```
cpus=10  burners=40
--- running the shipped suite under load ---
  ok   - unwrapped run corrupts under a mid-run edit (rc=0, output not clean)
  ok   - braces alone still fail (rc=2) ...
passed: 13  failed: 0
```

An assertion that the file's mtime/size changed before the body exited would make case 3 non-vacuous and stop cases 1/2 from crying wolf.

### minor — the PR body's Evidence section certifies a version of the test that is not at HEAD
The body says "7 cases", lists cases 0-5, and pastes `passed: 7  failed: 0`. HEAD (`3c621263`) ships 8 cases and 13 assertions.

```
$ grep -c '^echo "case ' q-system/.q-system/scripts/test/test-script-stable-under-self-edit.sh
8
$ bash q-system/.q-system/scripts/test/test-script-stable-under-self-edit.sh | tail -1
passed: 13  failed: 0
$ gh -R assafkip/kipi-system pr view 223 --json body -q .body | grep -n '7 cases\|passed: '
27:...test-script-stable-under-self-edit.sh`, 7 cases, in the capability manifest.
46:passed: 5  failed: 2
52:passed: 7  failed: 0
```

The receipt is the thing a gate reads. It should match the code it certifies.

### minor — "discovered, not hardcoded" is scoped to one directory
`discover_drivers` (line 74) globs `"$dir"/*.sh` and case 4 passes only `$SCRIPTS_DIR`. Applying the PR's own two markers repo-wide finds `kipi-dispatch.sh` at repo root, which case 4 never evaluates:

```
=== repo-wide files matching the PR's own two discovery markers ===
NOT-SCANNED    ./kipi-dispatch.sh                            timeout=none  claude-p=yes
in-scope       ./q-system/.q-system/scripts/linear-worker.sh timeout=1800  claude-p=yes
in-scope       ./q-system/.q-system/scripts/open-loops-heartbeat.sh timeout=none claude-p=yes
in-scope       ./q-system/.q-system/scripts/pr-review-agent.sh timeout=2400 claude-p=yes
```

No real exposure today: `kipi-dispatch.sh` is a launchd heartbeat bounded at ~25s, and its `claude -p` hits are all in comments (lines 49, 60, 324, 716) — which is the second half of the same weakness, `grep -q 'claude -p'` matching prose. Two errors cancelling. The header at line 63-75 reads broader than the code is.

## What is sound

Attacks I ran that the change survived:

**The fix scales to the real file, not just the 7-line synthetic.** I injected `sleep 3; echo; exit 0` into a real 59,799-byte copy of the wrapped `converge.sh` and into an unwrapped copy of the same, then ran the shipped mid-run edit against both:

```
=== A: real-sized UNWRAPPED converge.sh body + mid-run edit ===
    | PADDING: No such file or directory
    | PHASE_TWO_RAN ... => CORRUPTED
=== B: real-sized WRAPPED converge.sh body + mid-run edit ===
    | PHASE_TWO_RAN
    | the words alpha and beta live only inside this quoted string
    | MARKER_OK  => CLEAN
```

**The startup parse window is not a hole.** Edit fired with zero delay, racing bash's own parse of the 59KB file: rc=0, clean. The compound command is consumed before anything runs.

**The traps still fire inside the brace.** `converge.sh:201-203` wires TERM/INT/HUP; the regression suite exercises them live and stays green — `PASS: 21/21 converge checks`, including "SIGTERM mid-run releases the claim" and "kill case leaves no orphan behind".

**The shape predicate is genuinely tight.** `assert_wrapped` rejects a statement added below `}`, a statement added between `exit` and `}`, an indented `exit`, and a removed `exit`. Cases 5 and 6 mutate both the predicate and the discovery, and each goes red for the stated reason.

**`bash -n` and the repo's own lints are clean on both rewritten files.** `undefined-helper-lint: clean`; `portability-lint` reports 14 pre-existing findings, none in `converge.sh` or `linear-worker.sh`.

**The `512 -> 511` baseline edit is not a smuggled loosening.** It is the ratchet auto-tightening; `instruction-budget-audit.py` reports `Total always-on: 511` today. It arrived via an unrelated auto-commit (`f83845e4`) rather than this work, but it is correct.

## VERDICT: APPROVE WITH NITS

Nothing here is unrecoverable and nothing fires wrong at 3am on its own. The core mechanism is real, correctly diagnosed, and I confirmed it holds at production file size — which is more than the shipped suite proves.

Fix first: the timing margin in cases 1 and 2. Assert the edit actually landed before the body exited (compare size or mtime), so case 3's green means the fix held under an edit rather than meaning no edit arrived. That one change removes both the false-alarm path and the vacuous pass.

FINDINGS:
minor|KNOWN_UNWRAPPED comment claims its spillover ids hold `gates run` red, but both are untriaged and land in the non-blocking REPORT bucket; the ledger is also untracked so the ids resolve to nothing on a fresh clone|q-system/.q-system/scripts/test/test-script-stable-under-self-edit.sh:92
minor|Cases 1/2 depend on a 2s timing margin with no assertion that the mid-run edit landed before the body exited; a late edit turns them into a false alarm and makes case 3's green vacuous|q-system/.q-system/scripts/test/test-script-stable-under-self-edit.sh:157
minor|PR body's Evidence section certifies a 6-case suite reporting `passed: 7`, while HEAD ships 8 cases and 13 assertions|q-system/.q-system/scripts/test/test-script-stable-under-self-edit.sh:167
minor|Driver discovery globs only q-system/.q-system/scripts/*.sh, so kipi-dispatch.sh matches the PR's own `claude -p` marker and is never evaluated; the marker also matches comments rather than invocations|q-system/.q-system/scripts/test/test-script-stable-under-self-edit.sh:74
END FINDINGS

PR #226, opened 2026-08-19
fix(merge-bypass-gate): scope the push deny to protected fleet remotes (ASK-925)

assafkip:
Resolves spillover sp-9154c64d. The unscoped push deny fired on any GitHub remote, blocking main-branch creation on assafkip/adhd-output-style, a personal repo with none of the required checks the deny protects. The gate's own remedy (auto-merge via PR checks) cannot exist where there is no PR machinery.

- Push deny now keys on remote owner/name against a protected set (default assafkip/kipi-system; MERGE_GATE_PROTECTED_REPOS env override, read per call so test suites pin their own).
- Fail closed: a GitHub URL whose owner/name cannot be parsed stays protected.
- Merge side stays global on purpose: the admin flag is never the safe shape anywhere.

Verification: all 5 gate suites green (126 checks incl. 3 new scoping cases). Mutants killed: fail-open on unparseable URL dies on the gh_weird case; scoping removed dies on both gh_other cases.

Linear: ASK-925. Founder-directed 2026-08-19 ("make sure its on main" for the standalone repo).

https://claude.ai/code/session_012ncaKh6bKNASSMbyK1wTRt

### DIFF
--- a/q-system/.q-system/scripts/merge-bypass-gate.py
+++ b/q-system/.q-system/scripts/merge-bypass-gate.py
@@ -104,6 +104,31 @@ from pathlib import Path
 # These are the two names the fleet actually protects.
 PROTECTED_BRANCHES = {"main", "master"}
 
+# The PUSH deny is scoped to the remotes that actually carry the PR machinery
+# (required checks 'validate' + 'kipi/reviewer-approved'). Measured 2026-08-19
+# (sp-9154c64d): the unscoped form blocked creating `main` on a brand-new
+# personal repo (assafkip/adhd-output-style) that has no checks at all -- a
+# refusal whose remedy ("let the machinery merge it") cannot exist there. A
+# gate that fires where its own remedy is impossible teaches routing around
+# gates. The MERGE side stays global on purpose: `--admin` is never the safe
+# shape anywhere, and --auto costs nothing to type.
+# A GitHub URL whose owner/name cannot be parsed stays protected (fail closed
+# on the weird form). Read per call so the test suites can pin their own set
+# via MERGE_GATE_PROTECTED_REPOS (comma-separated owner/name).
+_DEFAULT_PROTECTED_REPOS = "assafkip/kipi-system"
+_GH_URL_RE = re.compile(r"github\.com[:/]([^/\s]+)/([^/\s]+?)(?:\.git)?/?$")
+
+
+def _protected_repo(url: str) -> bool:
+    m = _GH_URL_RE.search(url or "")
+    if not m:
+        return True  # a github URL this cannot parse stays protected
+    repos = frozenset(
+        r.strip() for r in os.environ.get(
+            "MERGE_GATE_PROTECTED_REPOS", _DEFAULT_PROTECTED_REPOS).split(",")
+        if r.strip())
+    return f"{m.group(1)}/{m.group(2)}" in repos
+
 # Shell separators that start a new command. `|` is included so `foo | git push ...`
 # is examined rather than swallowed as an argument of foo.
 _SEPARATORS = {";", ";;", "&&", "||", "|", "|&", "&", "\n"}
@@ -486,6 +511,11 @@ def _push_verdict(seg: list[str], cwd: str) -> str | None:
     # A URL typed straight on the command line never reaches `remote get-url`.
     url = remote if re.match(r"^(https?://|git@|ssh://)", remote) else _remote_url(repo_dir, remote)
     if not _is_github(url):
+        return None
+    if not _protected_repo(url):
+        # A GitHub repo outside the protected set has no 'validate' /
+        # 'kipi/reviewer-approved' checks to skip; the deny's remedy does not
+        # exist there (sp-9154c64d).
         return None
 
     if refspecs:
--- a/q-system/.q-system/scripts/merge-bypass-gate.py
+++ b/q-system/.q-system/scripts/merge-bypass-gate.py
@@ -151,8 +151,22 @@ def _protected_set() -> frozenset[str] | None:
     return frozenset(valid)
 
 
+def _repo_root(repo_dir: str) -> str:
+    """The worktree root, so the fleet marker is checked where it lives.
+    Checking the push CWD missed `cd <subdir> && git push` on every instance
+    (PR #226 review round 2, major). Unresolvable -> the dir itself, and the
+    remote-set check still applies behind it."""
+    try:
+        r = subprocess.run(["git", "-C", repo_dir, "rev-parse", "--show-toplevel"],
+                           capture_output=True, text=True, timeout=3)
+    except Exception:
+        return repo_dir
+    top = r.stdout.strip()
+    return top if r.returncode == 0 and top else repo_dir
+
+
 def _protected_repo(url: str, repo_dir: str) -> bool:
-    if (Path(repo_dir) / "q-system").is_dir():
+    if (Path(_repo_root(repo_dir)) / "q-system").is_dir():
         return True  # fleet instance: the machinery is right there in the tree
     m = _GH_URL_RE.search((url or "").lower())
     if not m:
--- a/q-system/.q-system/scripts/merge-bypass-gate.py
+++ b/q-system/.q-system/scripts/merge-bypass-gate.py
@@ -28,6 +28,10 @@ route AROUND machinery that already works without a human:
                                     This is the path linear-worker.sh already uses.
     gh pr merge --admin ...  DENY   The only way to defeat kipi/reviewer-approved.
     git push <protected>     DENY   Skips the PR, so skips both required checks.
+                                    Scoped to protected repos: a fleet-marker
+                                    tree (q-system/) or the protected remote
+                                    set. Elsewhere the checks do not exist, so
+                                    the push is not a bypass (sp-9154c64d).
     everything else          ALLOW  not-applicable
 
 The effect is to make the autonomous path the ONLY path, which is what the founder
@@ -104,7 +108,7 @@ from pathlib import Path
 # These are the two names the fleet actually protects.
 PROTECTED_BRANCHES = {"main", "master"}
 
-# The PUSH deny is scoped to the remotes that actually carry the PR machinery
+# The PUSH deny is scoped to repos that actually carry the PR machinery
 # (required checks 'validate' + 'kipi/reviewer-approved'). Measured 2026-08-19
 # (sp-9154c64d): the unscoped form blocked creating `main` on a brand-new
 # personal repo (assafkip/adhd-output-style) that has no checks at all -- a
@@ -112,21 +116,50 @@ PROTECTED_BRANCHES = {"main", "master"}
 # gate that fires where its own remedy is impossible teaches routing around
 # gates. The MERGE side stays global on purpose: `--admin` is never the safe
 # shape anywhere, and --auto costs nothing to type.
-# A GitHub URL whose owner/name cannot be parsed stays protected (fail closed
-# on the weird form). Read per call so the test suites can pin their own set
-# via MERGE_GATE_PROTECTED_REPOS (comma-separated owner/name).
+#
+# TWO protected signals, either one denies (PR #226 review, major):
+#   1. The FLEET MARKER: this hook ships to every kipi instance via
+#      settings-template.json, and every instance carries q-system/ in its
+#      root. A remote allowlist alone turned the deny off on all of them the
+#      day it landed -- including instances with required checks and
+#      enforce_admins:false, the exact scar configuration. The marker is
+#      local, needs no enumeration, and cannot rot as instances are added.
+#   2. The REMOTE SET: owner/name pairs for protected repos with no fleet
+#      marker in the tree. Default assafkip/kipi-system; override via
+#      MERGE_GATE_PROTECTED_REPOS (comma-separated), read per call so the
+#      test suites can pin their own. Matching is case-folded and tolerates
+#      a .git suffix (GitHub treats both as the same repo). An override that
+#      is SET but yields no valid owner/name entry fails CLOSED -- the same
+#      direction as the URL parser beside it.
+# A GitHub URL whose owner/name cannot be parsed stays protected.
 _DEFAULT_PROTECTED_REPOS = "assafkip/kipi-system"
 _GH_URL_RE = re.compile(r"github\.com[:/]([^/\s]+)/([^/\s]+?)(?:\.git)?/?$")
 
 
-def _protected_repo(url: str) -> bool:
-    m = _GH_URL_RE.search(url or "")
+def _protected_set() -> frozenset[str] | None:
+    """Normalized owner/name set, or None when an explicit override is all
+    malformed (the fail-closed sentinel)."""
+    raw = os.environ.get("MERGE_GATE_PROTECTED_REPOS")
+    entries = (raw if raw is not None else _DEFAULT_PROTECTED_REPOS).split(",")
+    valid = set()
+    for entry in entries:
+        e = entry.strip().lower().removesuffix(".git").strip("/")
+        if e.count("/") == 1 and all(e.split("/")):
+            valid.add(e)
+    if raw is not None and not valid:
+        return None
+    return frozenset(valid)
+
+
+def _protected_repo(url: str, repo_dir: str) -> bool:
+    if (Path(repo_dir) / "q-system").is_dir():
+        return True  # fleet instance: the machinery is right there in the tree
+    m = _GH_URL_RE.search((url or "").lower())
     if not m:
         return True  # a github URL this cannot parse stays protected
-    repos = frozenset(
-        r.strip() for r in os.environ.get(
-            "MERGE_GATE_PROTECTED_REPOS", _DEFAULT_PROTECTED_REPOS).split(",")
-        if r.strip())
+    repos = _protected_set()
+    if repos is None:
+        return True  # explicit override, zero valid entries: stay closed
     return f"{m.group(1)}/{m.group(2)}" in repos
 
 # Shell separators that start a new command. `|` is included so `foo | git push ...`
@@ -512,10 +545,10 @@ def _push_verdict(seg: list[str], cwd: str) -> str | None:
     url = remote if re.match(r"^(https?://|git@|ssh://)", remote) else _remote_url(repo_dir, remote)
     if not _is_github(url):
         return None
-    if not _protected_repo(url):
-        # A GitHub repo outside the protected set has no 'validate' /
-        # 'kipi/reviewer-approved' checks to skip; the deny's remedy does not
-        # exist there (sp-9154c64d).
+    if not _protected_repo(url, repo_dir):
+        # No fleet marker in the tree and not in the protected remote set:
+        # there are no 'validate' / 'kipi/reviewer-approved' checks to skip,
+        # so the deny's remedy does not exist there (sp-9154c64d).
         return None
 
     if refspecs:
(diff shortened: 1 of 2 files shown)


assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-226-20260819-145645.md` (6693 bytes).

```
FINDINGS:
major|Default protected set of one repo turns the push-to-main deny off on all 23 GitHub-remote fleet instances the hook ships to, including assafkip/ktlyst-saas-product which has required checks and enforce_admins:false|q-system/.q-system/scripts/merge-bypass-gate.py:118
minor|MERGE_GATE_PROTECTED_REPOS fails open with no signal on empty, whitespace, owner-omitted, or .git-suffixed values while the URL parser in the same function fails closed|q-system/.q-system/scripts/merge-bypass-gate.py:118
minor|Owner/name match is case-sensitive so the same protected repo escapes the set under a different URL casing|q-system/.q-system/scripts/merge-bypass-gate.py:119
minor|Header contract table still states the push deny applies to any protected branch on a GitHub remote, which the code no longer does|q-system/.q-system/scripts/merge-bypass-gate.py:30
END FINDINGS
```

--- reviewer output ---

## Review: PR #226 — scope the push deny to protected fleet remotes (ASK-925)

The premise is right: a deny whose remedy does not exist teaches routing around gates. The implementation picks the wrong default set, and the hook ships fleet-wide.

---

### MAJOR — the push deny goes dead on 23 fleet instances, including one that has the exact scar configuration this gate was built for

`q-system/.q-system/scripts/merge-bypass-gate.py:118`

`_DEFAULT_PROTECTED_REPOS = "assafkip/kipi-system"`. This hook is shipped to every instance by `settings-template.json:195` via `kipi update`. Not one of the 25 registered instances has `assafkip/kipi-system` as its origin, so after this merge the push deny fires on none of them.

That would be defensible if the PR's premise held everywhere ("no PR machinery there"). It does not. `assafkip/ktlyst-saas-product` (the `ktlyst` instance, the live product line) has required checks `prd-issue-gates`, `full-suite`, `kipi/reviewer-approved`, and `enforce_admins: false`. That is verbatim the configuration in this file's own WHY block at lines 8 to 20: an admin credential plus `enforce_admins:false` means the server does NOT stop the owner pushing straight to main, so this hook was the only refusal. The change removes it there.

Reproducer (executed):

```
python3 - <<'PY'
# build synthetic repos with real fleet origins, feed the hook a PreToolUse payload
...
print(ask("git push origin main", d))
PY
```

Real output:

```
ALLOW  ktlyst-saas-product  [GitHub-protected: prd-issue-gates/full-suite/reviewer-approved, enforce_admins=false]
ALLOW  kipi-investigations  [fleet instance]
ALLOW  cole-gtm             [live outbound engine]
DENY   kipi-system          [in default protected set]
```

Regression confirmed against the parent commit:

```
BEFORE PR (32349ddb^): DENY
AFTER  PR (HEAD)     : ALLOW
```

Branch-protection survey backing the claim (executed `gh api`):

```
ktlyst-saas-product      PROTECTED checks=['prd-issue-gates', 'full-suite', 'kipi/reviewer-approved']
kipi-system              PROTECTED checks=['validate', 'kipi/reviewer-approved']
kipi-investigations      UNPROTECTED (Branch not protected)
cole-gtm                 UNPROTECTED (Branch not protected)
...
```

At 3am an unattended worker in the ktlyst instance can now land on main past all three checks. A human has to notice, revert, and re-run. Recoverable, so major and not blocker.

---

### MINOR — the env override fails OPEN on every malformed value, while the URL parser fails closed

`q-system/.q-system/scripts/merge-bypass-gate.py:118`

`MERGE_GATE_PROTECTED_REPOS` is read raw with no validation. Any value that is empty, whitespace, missing the owner, or carrying a `.git` suffix silently protects nothing, and the gate reports nothing. The URL path in the same function deliberately fails closed on an unparseable form; the config path does the opposite.

Reproducer (executed), origin pinned to `assafkip/kipi-system`, command `git push origin main`:

```
DENY   unset (default)
DENY   correct value
ALLOW  owner omitted (typo)
ALLOW  empty string
ALLOW  whitespace
ALLOW  with .git suffix
```

---

### MINOR — owner/name comparison is case-sensitive, so the same repo escapes under a different URL casing

`q-system/.q-system/scripts/merge-bypass-gate.py:119`

GitHub treats owner and repo names case-insensitively; `_protected_repo` does an exact string match against the set.

Reproducer output (same run as the first finding):

```
DENY   kipi-system   [in default protected set]
ALLOW  Kipi-System   [same repo, different letter case]
```

Dropping this to minor on the stated rule: no producer in this repo emits a mixed-case remote URL, I constructed it by hand.

---

### MINOR — the header contract table now contradicts the code

`q-system/.q-system/scripts/merge-bypass-gate.py:30`

Line 30 still reads `git push <protected>     DENY   Skips the PR, so skips both required checks.` as an unconditional rule. It is now true for exactly one repo. A reader auditing the gate reads that table, not `_protected_repo`.

---

### What is sound

Attacks I ran against the new scoping that it survived, all confirmed DENY on the protected repo:

- Env prefix on the command line: `MERGE_GATE_PROTECTED_REPOS=x/y git push origin main`. The prefix is stripped and never reaches the hook process, so the Bash tool cannot retarget the set. This was my first guess at a bypass and it is closed.
- SSH scp-form remote `git@github.com:assafkip/kipi-system.git`, URL with no `.git`, URL with a trailing slash, and a URL carrying embedded credentials all parse to the same owner/name.
- Lookalike host `github.com.evil.example/assafkip/kipi-system.git` does not match the regex and falls to the fail-closed branch.
- Explicit URL typed on the command line instead of a remote name.
- `git push origin HEAD:main` and `bash -c 'git push origin main'`.
- The merge side stayed global as the PR claims: `gh pr merge 1 --admin` in an unprotected repo is still DENY.
- All five gate suites pass: 30, 18, 21, 29, 28 checks, exit 0 each.
- The `gh_weird` fail-closed test is honest, not decoration — removing the `return True` on an unparseable URL flips it.

---

**VERDICT: REQUEST CHANGES**

Not BLOCK. Nothing here is unrecoverable, and the direction of the change is correct.

First thing to fix: the default protected set. A one-repo allowlist shipped to 25 instances turns the deny off everywhere it was the only line of defense. Two shapes that fit the PR's own premise better than an allowlist — either derive protection from the remote at call time (a repo with required checks is protected, one without is not, cached), or keep the allowlist but populate it with every fleet repo that currently has required checks, starting with `assafkip/ktlyst-saas-product`.

FINDINGS:
major|Default protected set of one repo turns the push-to-main deny off on all 23 GitHub-remote fleet instances the hook ships to, including assafkip/ktlyst-saas-product which has required checks and enforce_admins:false|q-system/.q-system/scripts/merge-bypass-gate.py:118
minor|MERGE_GATE_PROTECTED_REPOS fails open with no signal on empty, whitespace, owner-omitted, or .git-suffixed values while the URL parser in the same function fails closed|q-system/.q-system/scripts/merge-bypass-gate.py:118
minor|Owner/name match is case-sensitive so the same protected repo escapes the set under a different URL casing|q-system/.q-system/scripts/merge-bypass-gate.py:119
minor|Header contract table still states the push deny applies to any protected branch on a GitHub remote, which the code no longer does|q-system/.q-system/scripts/merge-bypass-gate.py:30
END FINDINGS

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-226-20260819-150736.md` (6365 bytes).

```
FINDINGS:
major|Fleet-marker check reads the push cwd rather than the repo root, so `cd <subdir> && git push origin main` or `git -C <subdir> push` disables the push deny on every fleet instance outside the default remote set|q-system/.q-system/scripts/merge-bypass-gate.py:155
minor|No test exercises `_DEFAULT_PROTECTED_REPOS`; mutating it to a garbage value leaves the suite 33/33 green even though it is the only protection for kipi-system when cwd is a subdirectory|q-system/.q-system/scripts/merge-bypass-gate.py:135
END FINDINGS
```

--- reviewer output ---

# Review: PR #226 — merge-bypass-gate push scoping (ASK-925)

Read-only review. Two findings survived reproduction; several attacks did not.

---

## MAJOR — the fleet marker is checked in the push *directory*, not the repo root

`q-system/.q-system/scripts/merge-bypass-gate.py:155`

`_protected_repo` tests `(Path(repo_dir) / "q-system").is_dir()`, and `repo_dir` is the push cwd (`_push_verdict:501`, `repo_dir = cwd`) or the `-C` argument. It is not the repo root. Any push issued from one directory down turns the marker off, and every fleet instance except `assafkip/kipi-system` then falls through to a remote set that does not list it, so the gate returns ALLOW.

The header comment at line 31 says the deny is scoped to "a fleet-marker **tree** (q-system/)". The code implements "a directory literally named q-system directly under the cwd." Those are different things, and the difference is the whole deny.

Reproducer (a fleet instance: `q-system/` at root, its own GitHub remote, default env):

```
python3 - <<'EOF'
import importlib.util, subprocess, tempfile
from pathlib import Path
GATE = Path(".../q-system/.q-system/scripts/merge-bypass-gate.py")
s = importlib.util.spec_from_file_location("g", GATE); m = importlib.util.module_from_spec(s); s.loader.exec_module(m)
def git(r,*a): subprocess.run(["git","-C",str(r),*a],check=True,capture_output=True,text=True)
with tempfile.TemporaryDirectory() as td:
    root=Path(td); repo=root/"instance"; repo.mkdir()
    git(repo,"init","-q","-b","main"); git(repo,"config","user.email","t@t"); git(repo,"config","user.name","t")
    (repo/"f.txt").write_text("x\n"); git(repo,"add","-A"); git(repo,"commit","-qm","seed")
    git(repo,"remote","add","origin","https://github.com/assafkip/4_points_consulting.git")
    (repo/"q-system").mkdir()
    for cmd in ["git push origin main",
                "cd q-system && git push origin main",
                "git -C %s/q-system push origin main" % repo]:
        print("%-62s -> %s" % (cmd.replace(str(repo),"<repo>"), m.classify(cmd, str(repo))[0]))
EOF
```

Real output:

```
git push origin main                                           -> deny
cd q-system && git push origin main                            -> allow
git -C <repo>/q-system push origin main                        -> allow
```

I also ran the same fixture with the hook payload `cwd` set to a subdirectory (the other way the hook sees a non-root cwd) — same result, `allow` at both `q-system/` and `q-system/.q-system/`.

Blast radius, unattended: an agent on a fleet instance with `required checks + enforce_admins:false` (the exact scar config this gate was written for) pushes straight to `main`, past `validate` and `kipi/reviewer-approved`. Before this PR that shape denied. Cleanup is a revert commit, so it is recoverable, which is why this is major and not blocker. It is worth noting the fleet's own standing guidance is "always `git -C <path>`" — the `-C`-to-a-subdir shape is not exotic here.

Fix shape: resolve the root before testing the marker, e.g. `git -C repo_dir rev-parse --show-toplevel` and check `q-system/` there (falling back to protected when the resolve fails, matching the fail-closed direction already used beside it).

---

## MINOR — no test exercises `_DEFAULT_PROTECTED_REPOS`

`q-system/.q-system/scripts/merge-bypass-gate.py:135`, `test_merge_bypass_gate.py:53`

The suite sets `os.environ["MERGE_GATE_PROTECTED_REPOS"] = "example/repo"` at module import, so the shipped default is never on any code path a check reads. That default is the only thing protecting `kipi-system` itself once cwd is a subdirectory (I confirmed on the real checkout: `cd q-system && git push origin main` → `deny`, via the set, not the marker).

Mutation run (copies in a temp dir, repo untouched):

```
MUTANT: default set -> garbage        (_DEFAULT_PROTECTED_REPOS = "nobody/nothing")
  rc=0  ok  33/33 checks passed
MUTANT: fleet marker name -> typo     ((Path(repo_dir) / "q-sistem").is_dir())
  rc=1  FAIL
```

The marker mutant dies; the default-value mutant survives. One case with the env var unset and a `assafkip/kipi-system` remote closes it.

---

## What is sound

- **Fail-open on an unparseable GitHub URL.** `https://github.com/justowner`, deep paths (`owner/repo/sub`), and `https://github.com.evil.io/assafkip/kipi-system.git` all fail the regex and return protected. Confirmed by running each through `classify`.
- **Malformed explicit override.** `MERGE_GATE_PROTECTED_REPOS="  ,garbage,no-slash-here/"` → `_protected_set()` returns the sentinel and everything stays denied. Empty string behaves the same.
- **Case / `.git` normalization.** `https://github.com/EXAMPLE/REPO.git` matches `example/repo`. The regex's non-greedy `[^/\s]+?` with the optional `.git` suffix parses correctly on all the URL forms I threw at it (`git@`, `ssh://`, trailing slash, userinfo).
- **The PR's stated purpose actually works.** A markerless repo on `assafkip/adhd-output-style` with the default env returns `allow`.
- **The merge side stayed global.** `--admin` still denies everywhere; the scoping only touched `_push_verdict`.
- **All 5 gate suites, run individually:** 33 / 18 / 21 / 29 / 28 checks, all rc=0. The wrapper suite passes for the right reason (its `cwd` is the real repo root, where the marker exists).

---

## VERDICT: REQUEST CHANGES

One major finding, recoverable, so this is REQUEST CHANGES and not BLOCK.

Most important thing to fix first: resolve the repo root before the `q-system/` marker test at `merge-bypass-gate.py:155`. As written, the marker that round 2 added specifically to keep the deny alive on fleet instances is defeated by a single `cd` into a subdirectory, which puts every instance except `kipi-system` back to no push protection at all.

FINDINGS:
major|Fleet-marker check reads the push cwd rather than the repo root, so `cd <subdir> && git push origin main` or `git -C <subdir> push` disables the push deny on every fleet instance outside the default remote set|q-system/.q-system/scripts/merge-bypass-gate.py:155
minor|No test exercises `_DEFAULT_PROTECTED_REPOS`; mutating it to a garbage value leaves the suite 33/33 green even though it is the only protection for kipi-system when cwd is a subdirectory|q-system/.q-system/scripts/merge-bypass-gate.py:135
END FINDINGS

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-226-20260819-151349.md` (5947 bytes).

```
FINDINGS:
minor|Fleet-marker comment claims instances carry required checks and enforce_admins:false, but zero fleet instance repos have any branch protection, so the deny prints a remedy that cannot complete there|q-system/.q-system/scripts/merge-bypass-gate.py:124
minor|The case-fold/.git normalization of protected-set ENTRIES is uncovered: mutating it to entry.strip() flips a real override from deny to allow while the suite stays 35/35 green|q-system/.q-system/scripts/merge-bypass-gate.py:146
END FINDINGS
```

--- reviewer output ---

# Review: PR #226 — scope the push deny to protected fleet remotes

Read the diff, ran the suite, mutation-tested the new code, and probed the gate against real remote-URL shapes and real fleet branch protection.

---

## Finding 1 — minor

**The fleet-marker's stated evidence is false against the live fleet, so the deny it keeps alive fires with an impossible remedy on ~30 repos.**

`q-system/.q-system/scripts/merge-bypass-gate.py:124`

The comment justifying signal 1 claims the marker protects instances "with required checks and enforce_admins:false, the exact scar configuration." Measured today against GitHub:

```
kipi-system              checks=['validate', 'kipi/reviewer-approved'] enforce_admins=True
cole-gtm                 PROTECTION-ABSENT: {"message":"Branch not protected"...
4_points_consulting      PROTECTION-ABSENT: {"message":"Branch not protected"...
ASK_AI_consultant        PROTECTION-ABSENT: {"message":"Branch not protected"...
kipi-investigations      PROTECTION-ABSENT: {"message":"Branch not protected"...
Pure-spectrum-Q          PROTECTION-ABSENT: {"message":"Branch not protected"...
ktlyst-strategy          PROTECTION-ABSENT: {"message":"Branch not protected"...
adhd-output-style        PROTECTION-ABSENT: {"message":"Branch not protected"...
```

Zero instance repos have branch protection. Not one has the scar configuration the comment names. And the marker denies on all of them:

```
A fleet-instance push  -> deny
   reason: pushing straight to 'main' on a GitHub remote skips the pull request, and both required checks live on the PR.
B personal-repo push   -> allow
```

(fixture: repo with `q-system/` at root, origin `https://github.com/assafkip/ASK_AI_consultant.git`, no `MERGE_GATE_PROTECTED_REPOS`)

At 3am an autonomous run in one of those instances gets denied and told to run `gh pr merge --auto` and wait for `validate` and `kipi/reviewer-approved`. Those checks do not exist there, so the remedy cannot complete. That is the sp-9154c64d shape this PR exists to remove, still live for the fleet.

**Why minor and not major:** the base branch denied those same pushes unconditionally, so this is not a regression — it is a false rationale plus an incomplete fix. Blocking the PR over a pre-existing condition would wedge a change that does improve things.

---

## Finding 2 — minor

**The test that claims to pin case-folded, `.git`-tolerant set matching cannot fail for that reason; the entry-normalization branch is uncovered.**

`q-system/.q-system/scripts/merge-bypass-gate.py:146`, test at `test_merge_bypass_gate.py:168`

Mutating `_protected_set`'s normalization from `entry.strip().lower().removesuffix(".git").strip("/")` to `entry.strip()`:

```
HEAD code, override 'Example/Repo.git' -> deny
M6 mutant, same override               -> allow
suite under the same M6 mutant         -> ok  35/35 checks passed rc=0
```

The `gh_cased` case passes either way because `_protected_repo` lowercases the URL independently at line 171. The set entry in that test is already `example/repo` — lowercase, no `.git` — so nothing exercises the entry side. A future refactor that drops that normalization silently turns a capitalized or URL-suffixed override into "protects nothing," with a green suite.

Bounded: the shipped default sets no override, so live blast radius today is zero.

---

Mutation testing killed 6 of 7 mutants I planted (M1 drop the scoping call, M2 marker always-false, M3 marker at push CWD instead of worktree root, M4 unparseable URL fails open, M5 default set replaced with garbage, M7 malformed override fails open). Only M6 above survived. The round-3 worktree-root fix is genuinely pinned.

URL parsing holds across every shape a real `git remote get-url` emits, all denied for `assafkip/kipi-system`: https with and without `.git`, `git@` SCP form, `ssh://`, `git://`, embedded-token URL, and mixed-case owner/name.

Marker detection holds in the shapes the repo's own worktree rule prescribes: repo root, deep subdirectory, `cd subdir && git push`, `git -C subdir push`, a real `git worktree` checkout, and a worktree subdirectory. All deny. A separate non-fleet repo nested inside a fleet tree correctly allows.

I tried and dropped two theories. First, that the added `git rev-parse` subprocess pushes the push path past the hook's 5-second timeout and fails open: measured 10.6 ms on the real repo, so no. Second, that an agent could set `MERGE_GATE_PROTECTED_REPOS` to un-protect kipi-system: Bash env does not persist across tool calls, `.claude/` writes are blocked by `claude-path-write-guard`, and the fleet marker dominates the env path anyway. No reproducer, dropped.

Core properties still hold on the real repo: `git push origin main` denies, `--admin` denies, `gh pr merge --auto --squash` allows, feature-branch push allows. The PR's stated goal works — `adhd-output-style` is now allowed.

---

## VERDICT: APPROVE WITH NITS

**Fix first:** the comment at line 124. It is the load-bearing justification for the fleet-marker signal and it is measurably wrong — no fleet instance has branch protection. Either correct the comment to say the marker is a deliberate conservative default with no checks behind it yet, or add the required checks to instance repos so the remedy the deny prints is actually reachable.

FINDINGS:
minor|Fleet-marker comment claims instances carry required checks and enforce_admins:false, but zero fleet instance repos have any branch protection, so the deny prints a remedy that cannot complete there|q-system/.q-system/scripts/merge-bypass-gate.py:124
minor|The case-fold/.git normalization of protected-set ENTRIES is uncovered: mutating it to entry.strip() flips a real override from deny to allow while the suite stays 35/35 green|q-system/.q-system/scripts/merge-bypass-gate.py:146
END FINDINGS

PR #72, opened 2026-08-02
Handle the condition, send nothing: 48 founder pings/24h to 1 (ASK-294)

assafkip:
Closes ASK-294.

## Measured, both ends

**Before (real 24h window, #general, 2026-08-01 12:46 to 2026-08-02 12:13 PDT): 48 messages.**

| Producer | Pings | After |
|---|---|---|
| dispatch stale-checkout | 15 | **0** (self-heals; pages only if diverged or a path collides) |
| `.claude/` integrity tripwire "armed, 41 files baselined" | 11 | 0 |
| converge stopped / cap / stalled / approved | 5 | 0 |
| reviewer "codex is not producing an independent review" | 4 | 0 |
| worker "repo identity matches no Linear project" | 4 | 0 |
| heartbeat "RESUMED... Nothing to do" | 3 | 0 |
| reviewer "codex is BACK" | 3 | 0 |
| fleet health "nothing to do now" | 1 | 0 |
| dispatch daily cap "Do: nothing" | 1 | 0 |
| cole-gtm voice-lint nudge (outside this repo) | 1 | 1 (fail-open, unclassified) |
| **Total** | **48** | **1** |

The one survivor is a producer in another repo. It is delivered on purpose: see fail-open below.

Dispatch's own log corroborates the stale-checkout half independently: 13 pings under the old per-sha dedupe key, 2 after ASK-283's constant key, in the same window. This PR takes the remaining 2 to 0 by handling the condition.

## The frame, and where I disagreed with the issue

The issue proposed the loop just run `git merge --ff-only`. **An auto-merge was built there on 2026-08-02 and removed the same night**, after three review rounds each finding a new way for it to lose data (ASK-284). Rebuilding it as specified would have re-shipped a known data-loss path.

What changed the answer: those rounds tried to make an **unbounded** write safe. It is not unbounded. `git diff --name-only HEAD origin/main` is the exact finite set of paths the fast-forward writes, and every member is individually checkable. `ff-merge-if-safe.sh` proves the merge cannot clobber anything over that set, then merges; otherwise it declines and names the path. No copies, no backups, no writes outside `.git`.

The scar's protection did not disappear, it moved: `test-ff-merge-if-safe.sh` case 4 holds the exact thing that killed the last attempt (a fast-forward silently overwriting a **gitignored** file, exit 0, no reflog). Case 7 holds the opposite failure: ignored files *outside* the merge set must not block, or the handler is decoration on a checkout carrying 3982 of them.

## Two layers

[...]

PR #229, opened 2026-08-20
spillover-validate.py: hand-run only, no call site, never auto-wired (ASK-344)

assafkip:
Closes the decision ASK-344 asked for. **One `reason` string in one JSON file.** Script, test, `path` and `spillover_id` untouched.

## The decision

The issue named two options. Neither is quite right, so this picks the honest third:

**KEEP the script, and record "no call site at all" as the decision** — not "a periodic job", not "retire it".

- **Not retire (option B).** It still solves a live problem. The ledger is at 449 open findings; bulk triage needs to know which findings point at files that no longer exist, resolved against *every* fleet root rather than one cwd.
- **Not a periodic job (option A).** Its output only has value at the moment someone is deciding an item's fate. On a schedule it pages about stale paths nobody is acting on.
- **Not `--apply` anywhere.** Measured, not cautious: a single-root `os.path.exists()` check once classified 209 findings as gone when most were alive under another fleet root. An automated bulk void destroys the signal the ledger exists to keep.

So zero wiring is this script's intended end state, and the `declared_inert` entry now says that instead of deferring with "the call site is a founder decision, not a hook".

## Base

`sana/ask-312`, not `main`. `spillover-validate.py` and its `declared_inert` entry exist on no other ref (`git ls-tree -r --name-only origin/main | grep spillover-validate` returns nothing; PR #76 still OPEN). This PR is stacked on #76 and should land after it.

## Verification

Reproducer observed RED before the edit:

```
$ python3 -c "<assert 'founder decision' gone from the reason>"
AssertionError: RED: reason still carries the deferral
exit=1
```

All three DoR checks after the edit, run in a worktree on `origin/sana/ask-312`:

```
$ python3 q-system/.q-system/scripts/capability-gate.py --check-only
capability-gate: GREEN
$ python3 q-system/.q-system/scripts/test_capability_gate.py
ALL PASS
$ python3 -c "<assert 'founder decision' gone, spillover_id still sp-8a89b0e0>"
assert OK
```

## Two notes for the reviewer

1. **The DoR has one wrong measurement.** It says "Do not assert that the gate prints a `DECLARED-INERT (sp-8a89b0e0)` line. It does not, measured 2026-08-20." It does — the line is printed on both the unmodified and the modified branch. Harmless here (the DoR only said don't *assert* it), but the stated reason for the old check being unpassable is not accurate.

2. **Do NOT resolve `sp-8a89b0e0` at closeout.** `main` tip `2cc8b185` (ASK-345) added `check_inert_spillover_live()`: a `declared_inert` entry must cite a NON-resolved spillover id. Resolving `sp-8a89b0e0` while this entry keeps citing it arms a fleet-wide `capability-gate` RED the moment #76 merges into main. The issue footer's "resolve the ledger row once this closes" instruction predates that gate. Captured as a spillover item rather than silently left.

## Not in scope

No call site added (no hook, no launchd job, no `gates run` wiring). Script and `test_spillover_validate.py` not deleted. `spillover-merge-orphans.py` untouched. ASK-337's sweep not re-run. PR #76 not landed or rebased.

## How a base wedge was avoided

Attempt #2 reset the dispatched worktree onto `origin/sana/ask-312` and the armed `claude-integrity-tripwire` fired on every subsequent Bash call, so nothing could commit. This run used a **fresh nested worktree** instead: the tripwire hook runs against `$CLAUDE_PROJECT_DIR` (the dispatched tree, left clean on `main`), so it stayed silent throughout. The nested worktree is removed; nothing untracked ships.

### DIFF
--- a/q-system/.q-system/capability-manifest.json
+++ b/q-system/.q-system/capability-manifest.json
@@ -595,7 +595,7 @@
     },
     {
       "path": "q-system/.q-system/scripts/spillover-validate.py",
-      "reason": "hand-run ledger-triage tool for ASK-337: read-only by default and --apply voids only confirmed-gone rows. Its one-shot sweep already ran (2026-08-03T14:45Z). Auto-wiring a bulk void is the exact hazard its own docstring names, so the call site is a founder decision, not a hook",
+      "reason": "hand-run ledger-triage tool for ASK-337: read-only by default and --apply voids only confirmed-gone rows. Its one-shot sweep already ran (2026-08-03T14:45Z). DECIDED ASK-344, 2026-08-20: hand-run only, permanently. No call site, no hook, no scheduled job, and --apply is never auto-wired into a bulk void. The hazard its own docstring names is measured, not hypothetical: a single-root os.path.exists() check once classified 209 findings as gone when most were alive under another fleet root, so an automated void would destroy the signal the ledger exists to keep. Zero wiring is the intended end state for this script, not a deferral awaiting a decision",
       "spillover_id": "sp-8a89b0e0"
     },
     {


assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-229-20260820-100251.md` (176756 bytes).

```
FINDINGS:
major|The timestamp-only watermark permanently hides an undelivered receipt appended in the same second as the previous read.|q-system/.q-system/scripts/notify-receipts-surface.py:113
major|A successful Linear issue creation followed by a ledger-write failure leaves the finding open, so retrying creates another permanent issue.|q-system/.q-system/scripts/spillover-promote.py:151
END FINDINGS
```

--- reviewer output, last 54596 bytes of 176756 (full review at the path above) ---

./README.md:50:Insights flow upward. A pattern noticed in scratch notes gets promoted to weekly. A repeated weekly pattern becomes canonical. The system gets sharper the longer you run it.
./scripts/persona-reorg.py:683:        # now resolve to the one promoted standalone product.
./scripts/persona-reorg.py:1398:                 "no brain repo).\n\n- `projects/product/`  (v5 — promoted from the former "
./scripts/persona-reorg.py:1429:        return _abort("promoted product not independent (verify_repo_independent failed)")
./q-system/lessons/centralize-every-axis-of-a-validity-rule-and-enforce-it-at-t.md:12:Treat low-confidence, provisional records differently from confirmed ones. When a value is extracted from loose, ambiguous input as a candidate that a later pass is supposed to confirm-or-drop, do not let it enter shared, high-trust structures with the same standing as a verified value. Reusing an extractor built for high-signal input against low-signal input inherits its permissiveness; keep provisional items quarantined until the promotion step runs, and make "unpromoted" a state the durable store can represent and gate on, not a convention held only in prose.
./q-system/lessons/gate-the-claim-against-its-cited-source-and-gate-that-the-ou.md:16:3. A dimension that is always strong or attention-grabbing gets promoted to the headline regardless of whether it matches the TYPE of thing that was asked. High signal is not the same as relevant signal.
./q-system/lessons/terminate-on-a-drained-work-queue-not-a-self-scored-plateau.md:18:4. Reconcile the narrative against the structured artifact before declaring success. Any claim the agent states in prose must be traceable to a record in the structured output; anything named in the summary but absent from the durable artifact is either promoted or re-queued. A rich prose summary must never stand in for a thin artifact — the artifact is the deliverable.
./q-system/lessons/terminate-on-a-drained-work-queue-not-a-self-scored-plateau.md:20:5. Assert the gap mechanically at closeout: count the entities named in the summary versus those present in the structured output and promoted downstream. A nonzero difference fails the run. This catches the "brag-list of leads that were never pulled" failure without a human re-read.
./q-system/lessons/terminate-on-a-drained-work-queue-not-a-self-scored-plateau.md:22:The tell: a run that "finished" is only proven finished if its work-queue is empty or every remaining item is tagged blocked-with-reason, and every claim in its summary maps to a promoted record.
./scripts/test_persona_reorg.py:102:check("ktlyst-hub: product -> ktlyst-saas/projects (promoted)",
./q-system/lessons/make-cold-load-reconstruction-share-a-contract-with-the-live.md:13:2. Enforce it with a reload-equivalence test: run a session that produces state through the live path, simulate a cold load, and assert the reconstructed state matches what the live session displayed. Assert on counts and content, not just on absence of errors — a run that reports success with zero promoted records is the exact case this test exists to catch.
./q-system/lessons/make-cold-load-reconstruction-share-a-contract-with-the-live.md:16:5. When diagnosing, inspect the actual stored records for the failing case (shapes and counts) rather than reasoning from the code — the discrepancy between 'many successful steps' and 'zero promoted results' is visible in the data long before it is visible in the logic.
./q-system/lessons/verify-against-the-exact-build-the-user-runs-not-a-fresh-cop.md:12:1. Pin one stable reference (a fixed alias, a promoted target, a single long-lived endpoint) and make every fix flow to that exact reference. Hand the user that same reference, never a per-build address.
./q-system/lessons/terminate-on-a-drained-work-queue-not-a-self-scored-plateau-2.md:16:4. Reconcile the narrative against the structured artifact before declaring success. Any claim stated in prose must be traceable to a record in the structured output; anything named in the summary but absent from the durable artifact is either promoted or re-queued. A rich prose summary must never stand in for a thin artifact.
./q-system/lessons/terminate-on-a-drained-work-queue-not-a-self-scored-plateau-2.md:18:5. Assert the gap mechanically at closeout: count the entities named in the summary versus those present and promoted downstream. A nonzero difference fails the run. This catches the brag-list-of-leads-never-pulled failure without a human re-read.
./plugins/prd-os/scripts/phase0_measure.py:63:    # Persona-detection primitives (promoted to public for cross-script reuse,

exec
/bin/zsh -lc "sed -n '1080,1210p' plugins/prd-os/scripts/prd_runner.py; sed -n '1210,1460p' plugins/prd-os/scripts/prd_runner.py; sed -n '1,260p' q-system/.q-system/scripts/notify-receipts-surface.py; sed -n '1,280p' q-system/.q-system/scripts/human-handoff-audit.py" in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-229
 succeeded in 0ms:
    return [r for r in _read_spillover(cfg).values() if r.get("status") == "open"]


# --- the anchor bar (PRD prd-finding-quality-bar-2026-08-03) ----------------
# why (measured 2026-08-03): `spillover add` accepted any string. The
# kipi-system ledger reached 476 open findings at ~50/day, 82 of which named no
# file, no test and no command. An unverifiable finding cannot be acted on by
# anyone, including its author, so it is indistinguishable from noise. The
# June 2026 spillover PRD made capture frictionless and left resolution costly;
# this is the missing half.
_ANCHOR_PATH_RE = re.compile(r"[\w./-]+\.(?:py|sh|js|ts|json|jsonl|md|yml|yaml|toml)\b")
# A bare test SYMBOL (no extension) is greppable, so it anchors without
# resolving as a path. A token WITH an extension is a path claim and must
# resolve, or `test_anything.py` becomes a universal bypass.
_ANCHOR_TEST_RE = re.compile(r"\btest[_-][A-Za-z0-9_-]{3,}\b")
# A command plus what it actually printed. "ran X and got Y" is the shape the
# verification-loops rule already demands of a claim.
_ANCHOR_CMD_RE = re.compile(
    r"`[^`\n]{3,}`|\bran\b[^.\n]{0,80}\b(?:and )?got\b|\$ \S+|\bexit(?:s|ed)? (?:code )?[1-9]")
# A module, function or dotted symbol: greppable, so it anchors like a filename.
_ANCHOR_SYMBOL_RE = re.compile(r"\b([A-Za-z_][A-Za-z0-9_]{2,}(?:\.[A-Za-z_][A-Za-z0-9_]*)*)\s*(?:\(\)|\.[a-z_]+)")
# A directory path (no extension), e.g. "plugins/memory-lifecycle".
_ANCHOR_DIR_RE = re.compile(r"\b((?:[\w.-]+/){1,}[\w.-]+)\b")


def _tracked_stems(root: Path) -> set:
    """Basenames AND extension-less stems of tracked files.

    A finding naming the MODULE `memory_reflect` anchors just as well as one
    naming `memory_reflect.py`; both are one grep away from the code.
    """
    cached = _tracked_stems._cache.get(str(root))
    if cached is not None:
        return cached
    stems = set()
    for name in _tracked_basenames(root):
        stems.add(name)
        stems.add(os.path.splitext(name)[0])
    _tracked_stems._cache[str(root)] = stems
    return stems


_tracked_stems._cache = {}


def _tracked_basenames(root: Path) -> set:
    """Basenames of every git-tracked file, cached per root.

    why (caught on the first live run 2026-08-03): findings overwhelmingly name
    a BARE filename ("capability-gate.py reports a timeout as RED"), not a path
    from the repo root. Checking `root / token` alone returned None for those,
    which would have refused most legitimate captures and made the bar useless
    at exactly the moment it started blocking real work. `git ls-files` is one
    call and covers the tracked tree; an untracked file still resolves through
    the direct-path branch below.
    """
    cached = _tracked_basenames._cache.get(str(root))
    if cached is not None:
        return cached
    names = set()
    try:
        out = subprocess.run(["git", "ls-files"], cwd=str(root), capture_output=True,
                             text=True, timeout=30)
        if out.returncode == 0:
            names = {os.path.basename(l) for l in out.stdout.splitlines() if l.strip()}
    except (OSError, subprocess.TimeoutExpired):
        names = set()
    _tracked_basenames._cache[str(root)] = names
    return names


_tracked_basenames._cache = {}


def _anchor_for(desc: str, repo_root) -> str | None:
    """Which verifiable anchor this description carries, or None.

    Order matters: a resolving path is the strongest claim, so it wins. The
    filesystem check is the whole point -- a regex that only matched the SHAPE
    of a filename would be a spell-checker, not a gate (negative self-test:
    test_named_file_that_does_not_exist_is_refused).
    """
    text = desc or ""
    root = Path(repo_root)
    tokens = _ANCHOR_PATH_RE.findall(text)
    for token in tokens:
        cand = Path(token)
        if cand.is_absolute():
            if cand.exists():
                return "file"
            continue
        # the token as-is, then progressively from the right, so
        # "q-system/.q-system/scripts/x.py" resolves from a root holding
        # only "scripts/x.py"
        parts = cand.parts
        for i in range(len(parts)):
            if (root / Path(*parts[i:])).exists():
                return "file"
    # bare-filename fallback, against the tracked tree
    if tokens:
        tracked = _tracked_basenames(root)
        for token in tokens:
            if os.path.basename(token) in tracked:
                return "file"
    # why (measured 2026-08-03, before any void): the extension-only rule failed
    # 107 of 476 rows, and reading them showed most were REAL findings naming a
    # directory ("plugins/memory-lifecycle is a symlink"), a module
    # ("memory_reflect.load_sidecar raises AttributeError"), or a function
    # ("bootstrap_notion main() rewrites config.json"). Voiding those would have
    # destroyed exactly the signal the ledger exists to keep -- the Skeptic Q1
    # in this PRD. A directory, a module and a function are all greppable, so
    # they anchor as well as a filename does.
    stems = _tracked_stems(root)
    for token in _ANCHOR_SYMBOL_RE.findall(text):
        head = token.split(".")[0].split("(")[0]
        if head in stems:
            return "file"
    for token in _ANCHOR_DIR_RE.findall(text):
        if (root / token).exists():
            return "file"
    if _ANCHOR_TEST_RE.search(text):
        return "test"
    if _ANCHOR_CMD_RE.search(text):
        return "command"
    return None


def _spillover_append(cfg: Config, record: dict) -> None:
    path = _spillover_path(cfg)
    path.parent.mkdir(parents=True, exist_ok=True)
    with path.open("a") as fh:
    with path.open("a") as fh:
        fh.write(json.dumps(record) + "\n")
        fh.flush()


def _issue_is_closed(cfg: Config, issue_id: str) -> bool:
    """A spillover item may only resolve against an issue that actually closed.
    The deterministic signal is the issue spec's frontmatter `status: closed`,
    which issue_runner sets only after every receipt verified."""
    spec = cfg.issues_dir / f"{issue_id}.md"
    if not spec.is_file():
        return False
    for line in spec.read_text().splitlines():
        s = line.strip()
        if s.startswith("status:"):
            return s.split(":", 1)[1].strip() == "closed"
    return False


class LinearRefError(Exception):
    """A resolution reference could not be PROVEN closed.

    Covers every unverifiable case alike — no API key, no network, no such
    issue, issue still open. They collapse to one class on purpose: the caller's
    only correct response to any of them is to refuse, so a code path that could
    tell them apart would only invite one of them being downgraded to a warning.
    """


LINEAR_API_URL = "https://api.linear.app/graphql"
# Linear identifiers are TEAMKEY-number. Anything else is a local spec id (or a
# typo), and must never become a live lookup.
LINEAR_ID_RE = re.compile(r"^[A-Z][A-Z0-9]{0,9}-\d+$")
LINEAR_STATE_QUERY = "query($id: String!) { issue(id: $id) { state { name type } } }"


def _linear_api_key() -> str:
    """The same auth path linear-sync.py uses: env first, then the 0600 secret.

    Read here rather than imported from linear-sync.py because prd-os ships as a
    standalone plugin into repos with no q-system tree. What is shared is the
    CONVENTION (this env name, this file path), not code that could drift.
    """
    env = os.environ.get("KIPI_LINEAR_API_KEY", "").strip()
    if env:
        return env
    path = Path(os.path.expanduser("~/.config/kipi/linear-api-key"))
    if not path.is_file():
        raise LinearRefError(
            "closure cannot be verified: no Linear API key. Create one at "
            "https://linear.app/settings/api then:\n"
            f"  umask 077 && printf '%s' '<key>' > {path}\n"
            "or export KIPI_LINEAR_API_KEY. An unverified reference is never recorded")
    key = path.read_text().strip()
    if not key:
        raise LinearRefError(f"closure cannot be verified: {path} is empty")
    return key


def _linear_issue_state(identifier: str) -> dict:
    """The `{name, type}` of a Linear issue's workflow state.

    Raises LinearRefError for anything short of a definite answer, including an
    unreachable API — offline is a refusal, not an assumption.
    """
    import urllib.error
    import urllib.request

    body = json.dumps({"query": LINEAR_STATE_QUERY, "variables": {"id": identifier}}).encode("utf-8")
    request = urllib.request.Request(
        LINEAR_API_URL, data=body, method="POST",
        headers={"Content-Type": "application/json", "Authorization": _linear_api_key()},
    )
    try:
        with urllib.request.urlopen(request, timeout=20) as response:
            payload = json.loads(response.read().decode("utf-8"))
    except urllib.error.HTTPError as exc:
        raise LinearRefError(f"Linear returned HTTP {exc.code} for {identifier}") from exc
    except urllib.error.URLError as exc:
        raise LinearRefError(f"cannot reach Linear to verify {identifier}: {exc.reason}") from exc
    except json.JSONDecodeError as exc:
        raise LinearRefError(f"Linear sent a non-JSON answer for {identifier}: {exc}") from exc
    # Linear answers HTTP 200 with an `errors` array for application-level
    # failures, so a status-code-only check would read a failed lookup as a
    # verified one -- the exact shape of bug this command exists to prevent.
    if payload.get("errors"):
        raise LinearRefError(
            f"Linear rejected the lookup for {identifier}: {json.dumps(payload['errors'])[:200]}")
    issue = (payload.get("data") or {}).get("issue")
    if not isinstance(issue, dict) or not isinstance(issue.get("state"), dict):
        raise LinearRefError(f"Linear has no issue {identifier}")
    return issue["state"]


def _verify_resolution_ref(cfg: Config, ref: str) -> dict:
    """Prove `ref` names a closed issue; return the evidence that proved it.

    Raises LinearRefError with an operator-readable reason when closure cannot
    be proven. Nothing here accepts the operator's word: the returned evidence
    describes what the TRACKER said, which is why a resolution still cannot be
    hand-flipped through this command.

    Local specs answer first. A repo that tracks its own issues under
    `.prd-os/issues/` keeps resolving offline even when its ids look like Linear
    keys; the Linear path only opens for a ref this repo has no spec for.
    """
    local_spec = cfg.issues_dir / f"{ref}.md"
    if local_spec.is_file() or not LINEAR_ID_RE.match(ref):
        if _issue_is_closed(cfg, ref):
            return {"resolution_tracker": "prd-os"}
        raise LinearRefError(
            f"issue '{ref}' is not closed. Build it through the normal "
            "reproducer-first issue flow and close it first.")
    state = _linear_issue_state(ref)
    if state.get("type") != "completed":
        # `canceled` lands here too, and should: a canceled issue shipped no fix,
        # so clearing the item with it would green the gate on work that never
        # happened. The honest exit for a non-item is --void, which records why.
        raise LinearRefError(
            f"Linear issue '{ref}' is not completed (state: {state.get('name')}). "
            "Close it first, or record a non-item with --void <reason>.")
    return {
        "resolution_tracker": "linear",
        "resolution_verified_state": state.get("name"),
        "resolution_verified_at": _now_iso(),
    }


def _spillover_group_counts(items: list, field: str) -> list:
    """(value, count) pairs for one field, biggest group first, ties by name.

    Deterministic order matters more than it looks: this report is read to
    decide which producer to open an issue against, and a set-iteration order
    would reshuffle the priorities between two runs over the same ledger.
    """
    counts: dict = {}
    for record in items:
        value = record.get(field) or "(unset)"
        counts[value] = counts.get(value, 0) + 1
    return sorted(counts.items(), key=lambda pair: (-pair[1], pair[0]))


def _print_spillover_groups(items: list, field: str, heading: str) -> None:
    print(f"\nby {heading} ({len(_spillover_group_counts(items, field))} group(s)):")
    for value, count in _spillover_group_counts(items, field):
        print(f"  {count:5d}  {value}")


def cmd_spillover(cfg: Config, args) -> int:
    """add | list | check | resolve | triage — the out-of-scope finding ledger."""
    import hashlib as _hashlib
    sub = args.spillover_cmd
    if sub == "add":
        sid = args.id or f"sp-{_hashlib.sha256((args.source + args.desc).encode()).hexdigest()[:8]}"
        anchor = _anchor_for(args.desc, cfg.repo_root)
        if anchor is None and not getattr(args, "force", False):
            sys.stderr.write(
                "refused: this capture carries no verifiable ANCHOR.\n"
                "A finding nobody can check is indistinguishable from noise, and\n"
                "82 of the 476 open items on 2026-08-03 were exactly this shape.\n\n"
                "Supply at least one of:\n"
                "  - a FILE path that resolves (e.g. scripts/foo.py:212)\n"
                "  - a COMMAND and what it printed (e.g. ran `pytest -q` and got: 3 failed)\n"
                "  - a named TEST (e.g. test_retry_cap_is_honored)\n\n"
                "If it genuinely has none, re-run with --force. That records\n"
                "anchor:none on the row so forced captures stay countable.\n")
            return 2
        _spillover_append(cfg, {
            "id": sid, "source": args.source, "description": args.desc,
            "severity": args.severity, "status": "open", "created_at": _now_iso(),
            "anchor": anchor or "none",
        })
        print(json.dumps({"id": sid, "status": "open"}))
        return 0
    if sub == "list":
        items = list(_read_spillover(cfg).values())
        if args.open_only:
            items = [r for r in items if r.get("status") == "open"]
        if args.as_json:
            print(json.dumps(items))
        else:
            for r in items:
                print(f"[{r.get('status')}] {r['id']}: {r.get('description', '')[:80]} (src {r.get('source')})")
        return 0
    if sub == "check":
        openv = _spillover_open(cfg)
        if openv:
            for r in openv:
                sys.stderr.write(f"SPILLOVER OPEN: {r['id']}: {r.get('description', '')[:100]} (src {r.get('source')})\n")
            sys.stderr.write(
                f"{len(openv)} open spillover item(s). Resolve each against a CLOSED issue "
                f"(prd_runner.py spillover resolve <id> --resolution-ref <issue-id>) or void it "
                f"(--void <reason>). They cannot be silently dropped.\n")
            return 1
        print("no open spillover items")
        return 0
    if sub == "triage":
        # Read-only by construction: no _spillover_append call reachable from
        # here. A ledger this size (350+ open, ~50/day arriving from a handful
        # of producers) is unworkable as a flat list, but the fix is a better
        # LENS, never a bulk exit. The only two ways out of the ledger stay
        # `resolve --resolution-ref <closed-issue>` and `resolve --void`.
        openv = _spillover_open(cfg)
        if not openv:
            print("no open spillover items")
            return 0
        print(f"{len(openv)} open spillover item(s)")
        _print_spillover_groups(openv, "severity", "severity")
        _print_spillover_groups(openv, "source", "source")
        return 0
    if sub == "resolve":
        rec = _read_spillover(cfg).get(args.id)
        if not rec:
            sys.stderr.write(f"unknown spillover id: {args.id}\n")
            return 2
        if not args.resolution_ref and not args.void:
            sys.stderr.write("resolve requires --resolution-ref <issue-id> or --void <reason>\n")
            return 2
        new = dict(rec)
        if args.void:
            new.update(status="resolved", void_reason=args.void, resolved_at=_now_iso())
        else:
            try:
                evidence = _verify_resolution_ref(cfg, args.resolution_ref)
            except LinearRefError as exc:
                sys.stderr.write(f"cannot resolve {args.id}: {exc}\n")
                return 2
            new.update(status="resolved", resolution_ref=args.resolution_ref,
                       resolved_at=_now_iso(), **evidence)
            if args.evidence:
                # Operator-supplied context (PR, merge commit). Recorded for the
                # next reader, never consulted above: it is a note attached to a
                # verified resolution, not a substitute for verifying one.
                new["resolution_evidence"] = args.evidence
        _spillover_append(cfg, new)
        print(json.dumps({"id": args.id, "status": "resolved"}))
        return 0
    sys.stderr.write(f"unknown spillover subcommand: {sub}\n")
    return 2


def cmd_gates(cfg: Config, args) -> int:
    """gates list prints the registry; gates run executes regression gates from
    the repo root (operator-authored shell commands, the same trust boundary
    as required_checks), per-gate green/RED, non-zero exit on any RED. `run`
    ALSO fails while any spillover item is open (out-of-scope findings are part
    of the standing no-bypass re-proof, so they can never be silently dropped)."""
    import subprocess as _subprocess
    import re as _re
    path = _gates_path(cfg)
    records = []
#!/usr/bin/env python3
"""Surface undelivered notify receipts at SessionStart. The ledger's only reader.

WHY THIS EXISTS (PR #72 review, MAJOR)

`slack-notify.sh:214` has stated for weeks that "notify-receipts-surface.py reads
it at SessionStart". That file did not exist, on any branch, and nothing else
opened `notify-receipts.jsonl` -- verified with `find` across the whole repo and
`git log --all --diff-filter=A`. So `--kind receipt` was a write-only sink.

That is not a cosmetic gap. `--kind receipt` means "the machine handled this,
recorded not delivered", and three pages converted to it are not handled by any
machine: `repo not found ... the Linear loop is DEAD`, `cannot compute its spend
budget window`, and `gh CLI is not on PATH ... the loop is stalled`. Each names a
condition that stops the loop completely, and each was being written to a file no
human and no process read. The alert channel was not quieted, it was severed.

This is the same defect as the one the whole ASK-310 line is about, inverted: not
"the loop hands work to a human", but "the loop stops telling anyone at all". A
receipt is only legitimately silent if something reads it.

WHAT IT SHOWS, AND WHAT IT DELIBERATELY DOES NOT

It prints receipts since the last time it ran, newest first, and nothing else.
Specifically it does NOT re-print the whole ledger every session: the file is
append-only and unbounded, and a surface that grows without limit is one the
operator learns to scroll past -- the same cry-wolf failure `page_once` exists to
prevent on the delivery side.

Delivered rows are skipped. Those already reached Slack; repeating them here is
the duplicate the operator would learn to ignore.

Exit 0 ALWAYS. A SessionStart hook that can fail is a hook that can wedge a
session, and a missing or malformed ledger is not an error -- it is an empty one.
"""
from __future__ import annotations

import json
import os
import sys
from datetime import datetime, timezone

LEDGER = os.environ.get(
    "KIPI_NOTIFY_RECEIPTS", os.path.expanduser("~/.config/kipi/notify-receipts.jsonl"))
# Its own file, not the ledger, so replaying the surface never rewrites history.
STATE = os.environ.get(
    "KIPI_NOTIFY_SURFACE_STATE",
    os.path.expanduser("~/.config/kipi/notify-receipts-surface.json"))
MAX_SHOWN = 12


# The writer's timestamp key is `ts`, NOT `at`. The first cut of this reader used
# `at`, and because every synthetic fixture in its test also used `at`, the suite
# went green while the live ledger rendered every timestamp blank and the
# watermark never advanced -- the surface re-printed all 11 rows on every run. A
# fixture that invents its own schema tests the fixture.
TS_KEY = "ts"


def rows(path: str) -> list:
    """Every parsable row. A corrupt line is skipped, never fatal: one bad write
    must not blind the operator to the good rows around it."""
    if not os.path.exists(path):
        return []
    out = []
    try:
        with open(path, encoding="utf-8", errors="replace") as fh:
            for line in fh:
                line = line.strip()
                if not line:
                    continue
                try:
                    r = json.loads(line)
                except json.JSONDecodeError:
                    continue
                if isinstance(r, dict):
                    out.append(r)
    except OSError:
        return []
    return out


def read_watermark() -> str:
    try:
        with open(STATE, encoding="utf-8") as fh:
            return json.load(fh).get("last_seen_at") or ""
    except Exception:  # noqa: BLE001 - a missing/corrupt watermark means "show recent"
        return ""


def write_watermark(value: str) -> None:
    """Best effort. Failing to persist means one repeat next session, which is a
    far better failure than a crash in a SessionStart hook."""
    try:
        os.makedirs(os.path.dirname(STATE), exist_ok=True)
        with open(STATE, "w", encoding="utf-8") as fh:
            json.dump({"last_seen_at": value}, fh)
    except OSError:
        pass


def undelivered_since(all_rows: list, watermark: str) -> list:
    fresh = []
    for r in all_rows:
        if r.get("delivered"):
            continue
        # The writer already stamps a per-row `read` flag. Honour it rather than
        # inventing a parallel notion of seen -- two sources of truth for the
        # same fact is how a surface starts disagreeing with its ledger.
        if r.get("read") in (True, "True"):
            continue
        at = r.get("ts") or ""
        if watermark and at and at <= watermark:
            continue
        fresh.append(r)
    return fresh


def render(fresh: list, total_new: int) -> str:
    lines = [f"# Notify receipts ({total_new} new, handled by the machine, not delivered)",
             "# These are conditions the loop recorded rather than paged about."]
    for r in fresh:
        at = (r.get("ts") or "")[:16].replace("T", " ")
        label = r.get("label") or "-"
        msg = (r.get("message") or "").strip().replace("\n", " ")
        lines.append(f"- [{at}] {label}: {msg[:150]}")
    if total_new > len(fresh):
        lines.append(f"- ...and {total_new - len(fresh)} more in {LEDGER}")
    return "\n".join(lines)


def main() -> int:
    all_rows = rows(LEDGER)
    if not all_rows:
        return 0
    watermark = read_watermark()
    fresh = undelivered_since(all_rows, watermark)
    newest = max((r.get("ts") or "") for r in all_rows)
    if not fresh:
        write_watermark(newest)
        return 0
    fresh.sort(key=lambda r: r.get("ts") or "", reverse=True)
    print(render(fresh[:MAX_SHOWN], len(fresh)))
    write_watermark(newest)
    return 0


if __name__ == "__main__":
    try:
        sys.exit(main())
    except Exception as exc:  # noqa: BLE001
        # See the module docstring: a SessionStart hook may not wedge a session.
        print(f"# notify-receipts-surface: skipped ({str(exc)[:80]})", file=sys.stderr)
        sys.exit(0)
#!/usr/bin/env python3
"""No code path may declare that the process needs a human without naming why.

Founder-directed 2026-08-02: "review the whole code and identify anywhere that
says the process needs a human and remove it."

WHY THIS IS A SCRIPT AND NOT A GREP. The same class was hand-grepped three times
this session and each pass used a different, narrower vocabulary -- "Needs a
human" found converge.sh but missed "until someone runs", "the founder must" and
"manual intervention". A one-off grep proves the phrase you thought of is absent;
it says nothing about the phrase you did not. This carries the whole vocabulary
in one place so the next sweep is reproducible and the gate can run it forever.

THE RULE IS NOT "NEVER NAME A HUMAN". Some work genuinely is the founder's, and
`~/.claude/CLAUDE.md`'s destructive-op carve-out is explicit that asking for those
is the contract rather than a violation of it. So a call site may declare a human
by naming the CLASS that makes it human-only, using the same vocabulary
slack-notify.sh allowlists for `--kind decision --class` and founder-actor-gate.py
uses on prose. One vocabulary, three layers.

    # human-required: irreversible-git -- rewriting published history

WHAT IT DOES NOT CATCH (honest boundary):
  * A handoff phrased in words not in HANDOFF_PATTERNS. The vocabulary is
    open-ended by nature; this bounds the known shapes, it cannot bound English.
  * A call site that names a class dishonestly. The marker bounds the vocabulary,
    not the intent. Only review does.
  * Prose in .md rules and docs -- out of scope, captured as sp-87ff73a1 rather
    than merely noted here. Those describe the  # spillover-skip
    system rather than being it, and founder-actor-gate.py covers live prose.

Exit 0 = clean. Exit 1 = at least one unexplained handoff.
"""
from __future__ import annotations

import argparse
import os
import re
import subprocess
import sys

ALLOWED_CLASSES = (
    "irreversible-git", "out-of-tree-write", "spend", "publish", "credential",
    # A runner may not certify its own work. Added 2026-08-02 while sweeping:
    # converge.sh pages when no prd-os receipt covers the head, i.e. when nothing
    # proves the code was reviewed. Automating that would let the loop write its
    # own proof of review, which is worse than the page it replaces. None of the
    # five original classes fit, and forcing one would have been a bad fit
    # dressed as a good one.
    "self-certification",
)
MARKER = "human-required:"
# Distinct from MARKER on purpose. `human-required` says "the process stops here
# and a person acts". `definitional` says "this sentence uses a human as the unit
# of measure for severity" -- e.g. major = "wrong behaviour a human must clean
# up". That is vocabulary, not routing, and conflating the two would either
# license real handoffs or force a false class onto a definition.
DEFINITIONAL = "human-handoff-audit: definitional"

# Ordered roughly by how often each shape actually appeared in this repo.
HANDOFF_PATTERNS = [
    ("names a human as the actor", re.compile(
        r"needs?\s+a\s+human|requires?\s+a\s+human|a\s+human\s+(?:must|has\s+to|needs?\s+to)"
        r"|human\s+intervention|needs?\s+manual|manual\s+intervention", re.I)),
    ("names the founder as the actor", re.compile(
        r"the\s+founder\s+(?:must|has\s+to|needs?\s+to|should|will\s+need)"
        r"|ask\s+the\s+founder|tell\s+the\s+founder\s+to\b"
        r"|only\s+the\s+founder\s+can|waiting\s+(?:on|for)\s+the\s+founder", re.I)),
    ("defers to an unnamed someone", re.compile(
        r"until\s+someone|someone\s+(?:must|has\s+to|needs?\s+to|runs)"
        r"|somebody\s+(?:must|has\s+to|needs?\s+to)", re.I)),
    ("hands over a command to run", re.compile(
        r"\bDo:\s*(?:cd|git|gh|python3|bash|npm|kipi|launchctl|brew)\b"
        r"|run\s+(?:this|it)\s+(?:by\s+hand|yourself|manually)"
        r"|by\s+hand:\s*(?:cd|git|gh|python3|bash|kipi)\b", re.I)),
]

SKIP_DIR_PARTS = (".git", "node_modules", "worktrees", "template-repo",
                  ".prd-os", "__pycache__")
SKIP_DIR_PREFIXES = (".pr", ".review-")
# This file necessarily quotes every pattern it hunts.
SKIP_FILES = ("human-handoff-audit.py", "founder-actor-gate.py")


def is_generated_output(rel: str) -> bool:
    """`q-system/output/` holds generated artifacts. The docx builders there carry
    essay prose ABOUT the system -- e.g. "the founder has to manually copy it
    across" describing a limitation being argued -- which is content, not a
    control path this audit governs."""
    return rel.startswith("q-system/output/") or "/output/" in rel


def is_test_file(rel: str) -> bool:
    """A fixture that QUOTES a handoff is exercising the detector, not committing
    the defect. Same exclusion and same reasoning as notify-callsite-audit.py."""
    base = os.path.basename(rel)
    return (base.startswith(("test-", "test_")) or base.endswith(("_test.py", "_test.sh"))
            or "/test/" in rel or rel.startswith("test/")
            or "/tests/" in rel or rel.startswith("tests/"))


def is_skipped(rel: str) -> bool:
    if is_test_file(rel) or is_generated_output(rel):
        return True
    parts = rel.split(os.sep)
    if any(p in SKIP_DIR_PARTS for p in parts):
        return True
    if any(p.startswith(SKIP_DIR_PREFIXES) for p in parts[:-1]):
        return True
    return os.path.basename(rel) in SKIP_FILES


def tracked(repo: str):
    out = subprocess.run(["git", "-C", repo, "ls-files"],
                         capture_output=True, text=True, check=False)
    for rel in out.stdout.splitlines():
        if rel.endswith((".sh", ".py")) and not is_skipped(rel):
            yield rel


# A line that NEGATES the handoff is the cure, not the disease. Measured on the
# first full sweep: converge.sh:10 reads "Sana is a robot. She does not need a
# human to tell her to keep going" and was reported as a defect. A detector that
# flags its own fix trains the operator to skim it, which is root cause #3 of
# rca-work-routed-to-the-founder.
NEGATED = re.compile(
    r"(?:does\s+not|doesn't|never|no longer|without|must\s+not|must\s+never|cannot|can't|"
    r"nobody|not)\s+(?:\w+\s+){0,3}(?:needs?|requires?|waits?|asks?)"
    r"|no\s+human\s+(?:merge\s+)?(?:needed|required)"
    r"|(?:must|should)\s+never\s+be\s+the\s+(?:one|person)"
    r"|not\s+a\s+human", re.I)

# Text that QUOTES a handoff in order to record that it was removed. The scar
# comments written while removing these all say "used to", "the old text read",
# or sit inside quotes -- documentation of history, not a live handoff.
HISTORICAL = re.compile(
    r"used\s+to\b|the\s+old\s+text|previously\s+read|this\s+argued"
    r"|was\s+the\s+point|why\s+a\s+\d+|REGRESSION|scar\b", re.I)


def explained(line: str, window: str) -> bool:
    """A handoff is explained when the marker names an allowlisted class nearby.

    The window is small on purpose: a class named ten lines away is not an
    explanation of THIS line, it is a coincidence.
    """
    if MARKER not in window:
        return False
    tail = window.split(MARKER, 1)[1][:120]
    return any(c in tail for c in ALLOWED_CLASSES)


HEREDOC_OPEN = re.compile(r"<<-?\s*[\"']?([A-Za-z_][A-Za-z0-9_]*)[\"']?")


def heredoc_lines(lines) -> set:
    """Line indices inside a shell heredoc body.

    Scar: acking a severity DEFINITION inside pr-review-agent.sh's reviewer
    prompt inserted two `#` lines into the prompt itself, so the reviewer would
    have read them as instructions. `bash -n` parses that happily -- it is valid
    shell, just a corrupted prompt. Detected by eye, not by a gate.
    """
    inside, marker, out = False, None, set()
    for i, line in enumerate(lines):
        if inside:
            out.add(i)
            if line.strip() == marker:
                inside, marker = False, None
            continue
        m = HEREDOC_OPEN.search(line)
        if m and not line.lstrip().startswith("#"):
            inside, marker = True, m.group(1)
    return out


STRING_OPEN = re.compile(r'^\s*[A-Za-z_][A-Za-z0-9_]*=\"')


def quoted_block_lines(lines) -> set:
    """Line indices inside a multi-line double-quoted shell assignment.

    Same reasoning as heredoc_lines: a string BODY is data. pr-review-agent.sh
    builds its reviewer prompt as `PROMPT="...40 lines..."`, and that prose
    defines review severity using a human as the unit of measure. It is not a
    control path, and a marker cannot be placed inside it -- inserting a `#` line
    there once already turned two comments into instructions the reviewer would
    have read (reverted the same day).
    """
    inside, out = False, set()
    for i, line in enumerate(lines):
        if inside:
            out.add(i)
            # An odd number of unescaped quotes closes the assignment.
            if line.count('"') - line.count('\\"') >= 1:
                inside = False
            continue
        if STRING_OPEN.search(line) and (line.count('"') - line.count('\\"')) == 1:
            inside = True
    return out


def findings(repo: str):
    out = []
    for rel in tracked(repo):
        try:
            lines = open(os.path.join(repo, rel), encoding="utf-8",
                         errors="replace").read().splitlines()
        except OSError:
            continue
        skip = (heredoc_lines(lines) | quoted_block_lines(lines)
                ) if rel.endswith(".sh") else set()
        for i, line in enumerate(lines):
            if i in skip:
                continue
            for label, pat in HANDOFF_PATTERNS:
                if not pat.search(line):
                    continue
                # A negation can straddle a wrap: apply_claude_changes.py reads
                # "The founder must / never be the one who notices a regression"
                # across two lines, and a per-line check calls the cure a defect.
                pair = line + " " + (lines[i + 1] if i + 1 < len(lines) else "")
                # HISTORICAL needs a LOOK-BACK, not just the line: the "used
                # to" that marks a quote as history routinely sits one or two
                # lines above the command being quoted.
                back = "\n".join(lines[max(0, i - 3):i + 1])
                if NEGATED.search(pair) or HISTORICAL.search(back):
                    continue
                window = "\n".join(lines[max(0, i - 3):i + 4])
                if DEFINITIONAL in window or explained(line, window):
                    continue
                out.append((rel, i + 1, label, line.strip()[:104]))
                break
    return out


def main() -> int:
    ap = argparse.ArgumentParser()
    ap.add_argument("--repo", default=os.path.join(os.path.dirname(__file__), "..", "..", ".."))
    a = ap.parse_args()
    bad = findings(os.path.abspath(a.repo))
    if not bad:
        print("human-handoff-audit: OK -- no unexplained human handoff in live code")
        return 0
    print(f"human-handoff-audit: {len(bad)} unexplained human handoff(s).\n")
    for rel, ln, label, text in bad:
        print(f"  {rel}:{ln}  [{label}]\n      {text}")
    print(f"\nEither make the code DO it, or declare why it cannot:\n"
          f"    # {MARKER} <{'|'.join(ALLOWED_CLASSES)}> -- <reason>\n"
          f"A process that says it needs a human, without saying which authority it\n"
          f"lacks, is an unbuilt feature wearing a status message.")
    return 1


if __name__ == "__main__":
    sys.exit(main())

exec
/bin/zsh -lc "python3 -c \"import importlib.util,pathlib; p=pathlib.Path('q-system/.q-system/scripts/notify-receipts-surface.py'); s=importlib.util.spec_from_file_location('nrs',p); m=importlib.util.module_from_spec(s); s.loader.exec_module(m); row={'ts':'2026-08-20T10:15:30Z','delivered':False,'read':False,'message':'worker stalled'}; print('watermark:',row['ts']); print('unread same-second row:',m.undelivered_since([row],row['ts']))\"; python3 -c \"import importlib.util,pathlib,sys; p=pathlib.Path('q-system/.q-system/scripts/spillover-promote.py'); s=importlib.util.spec_from_file_location('sp',p); m=importlib.util.module_from_spec(s); s.loader.exec_module(m); m.load_rows=lambda ledger:{'sp-demo':{'id':'sp-demo','status':'open','severity':'minor','source':'review','description':'live finding'}}; created=[]; LS=type('LS',(),{'ISSUE_CREATE':'mutation','graphql':staticmethod(lambda q,v: ({'teams':{'nodes':[{'id':'team'}]}} if q.startswith('query') else (print('LINEAR CREATED:',v['input']['title']) or {'issueCreate':{'issue':{'identifier':'ASK-999'}}}))}); m.linear_module=lambda:LS; orig=pathlib.Path.open; pathlib.Path.open=lambda self,*a,**k: (_ for _ in ()).throw(OSError('simulated ledger disk failure')) if self.name=='spillover.jsonl' else orig(self,*a,**k); sys.argv=['spillover-promote.py','sp-demo','--title','duplicate me','--dor','Allowed files: x.py Acceptance: verify behavior. This definition is intentionally longer than one hundred and twenty characters to pass the current shallow gate.']; exec(\\\"try:\\n m.main()\\nexcept Exception as e:\\n print(type(e).__name__+':',e)\\\"); print('ledger status persisted: NO'); print('retry consequence: creates ASK-999-like issue again')\"" in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-229
 exited 1 in 0ms:
watermark: 2026-08-20T10:15:30Z
unread same-second row: []
  File "<string>", line 1
    import importlib.util,pathlib,sys; p=pathlib.Path('q-system/.q-system/scripts/spillover-promote.py'); s=importlib.util.spec_from_file_location('sp',p); m=importlib.util.module_from_spec(s); s.loader.exec_module(m); m.load_rows=lambda ledger:{'sp-demo':{'id':'sp-demo','status':'open','severity':'minor','source':'review','description':'live finding'}}; created=[]; LS=type('LS',(),{'ISSUE_CREATE':'mutation','graphql':staticmethod(lambda q,v: ({'teams':{'nodes':[{'id':'team'}]}} if q.startswith('query') else (print('LINEAR CREATED:',v['input']['title']) or {'issueCreate':{'issue':{'identifier':'ASK-999'}}}))}); m.linear_module=lambda:LS; orig=pathlib.Path.open; pathlib.Path.open=lambda self,*a,**k: (_ for _ in ()).throw(OSError('simulated ledger disk failure')) if self.name=='spillover.jsonl' else orig(self,*a,**k); sys.argv=['spillover-promote.py','sp-demo','--title','duplicate me','--dor','Allowed files: x.py Acceptance: verify behavior. This definition is intentionally longer than one hundred and twenty characters to pass the current shallow gate.']; exec("try:\n m.main()\nexcept Exception as e:\n print(type(e).__name__+':',e)"); print('ledger status persisted: NO'); print('retry consequence: creates ASK-999-like issue again')
                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                                       ^
SyntaxError: closing parenthesis '}' does not match opening parenthesis '('

exec
/bin/zsh -lc "python3 -c \"exec('''import importlib.util, pathlib, sys\\np=pathlib.Path(\\\"q-system/.q-system/scripts/spillover-promote.py\\\")\\ns=importlib.util.spec_from_file_location(\\\"sp\\\",p)\\nm=importlib.util.module_from_spec(s); s.loader.exec_module(m)\\nm.load_rows=lambda ledger:{\\\"sp-demo\\\":{\\\"id\\\":\\\"sp-demo\\\",\\\"status\\\":\\\"open\\\",\\\"severity\\\":\\\"minor\\\",\\\"source\\\":\\\"review\\\",\\\"description\\\":\\\"live finding\\\"}}\\nclass LS:\\n ISSUE_CREATE=\\\"mutation\\\"\\n @staticmethod\\n def graphql(q,v):\\n  if q.startswith(\\\"query\\\"): return {\\\"teams\\\":{\\\"nodes\\\":[{\\\"id\\\":\\\"team\\\"}]}}\\n  print(\\\"LINEAR CREATED:\\\",v[\\\"input\\\"][\\\"title\\\"]); return {\\\"issueCreate\\\":{\\\"issue\\\":{\\\"identifier\\\":\\\"ASK-999\\\"}}}\\nm.linear_module=lambda:LS\\norig=pathlib.Path.open\\ndef boom(self,*a,**k):\\n if self.name==\\\"spillover.jsonl\\\": raise OSError(\\\"simulated ledger disk failure\\\")\\n return orig(self,*a,**k)\\npathlib.Path.open=boom\\nsys.argv=[\\\"spillover-promote.py\\\",\\\"sp-demo\\\",\\\"--title\\\",\\\"duplicate me\\\",\\\"--dor\\\",\\\"Allowed files: x.py Acceptance: verify behavior. This definition is intentionally longer than one hundred and twenty characters to pass the current shallow gate.\\\"]\\ntry: m.main()\\nexcept Exception as e: print(type(e).__name__+\\\":\\\",e)\\nprint(\\\"ledger status persisted: NO\\\")\\nprint(\\\"retry consequence: creates ASK-999-like issue again\\\")''')\"" in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-229
 succeeded in 0ms:
LINEAR CREATED: duplicate me
OSError: simulated ledger disk failure
ledger status persisted: NO
retry consequence: creates ASK-999-like issue again

exec
/bin/zsh -lc "python3 -c \"import ast,subprocess,pathlib; fs=[x for x in subprocess.check_output(['git','diff','--name-only','origin/main...HEAD'],text=True).splitlines() if x.endswith('.py')]; bad=[]; [(bad.append((f,str(e))) if (lambda:None)() else None) for f in []]; [ast.parse(pathlib.Path(f).read_text(),filename=f) for f in fs]; print(f'AST parse OK: {len(fs)} changed Python files')\"; bash -n q-system/.q-system/scripts/slack-notify.sh q-system/.q-system/scripts/linear-worker.sh q-system/.q-system/scripts/ff-merge-if-safe.sh q-system/.q-system/scripts/converge.sh && echo 'bash -n OK: 4 high-risk changed shell files'" in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-229
 succeeded in 0ms:
AST parse OK: 32 changed Python files
bash -n OK: 4 high-risk changed shell files

codex
## Findings

### MAJOR

The timestamp-only watermark permanently hides an undelivered receipt appended in the same second as the previous read.

`q-system/.q-system/scripts/notify-receipts-surface.py:113`

Reproducer:

```bash
python3 -c "import importlib.util,pathlib; p=pathlib.Path('q-system/.q-system/scripts/notify-receipts-surface.py'); s=importlib.util.spec_from_file_location('nrs',p); m=importlib.util.module_from_spec(s); s.loader.exec_module(m); row={'ts':'2026-08-20T10:15:30Z','delivered':False,'read':False,'message':'worker stalled'}; print('watermark:',row['ts']); print('unread same-second row:',m.undelivered_since([row],row['ts']))"
```

Real output:

```text
watermark: 2026-08-20T10:15:30Z
unread same-second row: []
```

Two notification writers can emit the same second-resolution timestamp. If the second append happens after the surface reads but before it saves the watermark, the next session skips that unread row forever. A stalled unattended worker can therefore lose its only operator signal.

### MAJOR

A successful Linear issue creation followed by a ledger-write failure leaves the finding open, so retrying creates another permanent issue.

`q-system/.q-system/scripts/spillover-promote.py:151`

Reproducer:

```bash
python3 -c "exec('''import importlib.util, pathlib, sys
p=pathlib.Path(\"q-system/.q-system/scripts/spillover-promote.py\")
s=importlib.util.spec_from_file_location(\"sp\",p)
m=importlib.util.module_from_spec(s); s.loader.exec_module(m)
m.load_rows=lambda ledger:{\"sp-demo\":{\"id\":\"sp-demo\",\"status\":\"open\",\"severity\":\"minor\",\"source\":\"review\",\"description\":\"live finding\"}}
class LS:
 ISSUE_CREATE=\"mutation\"
 @staticmethod
 def graphql(q,v):
  if q.startswith(\"query\"): return {\"teams\":{\"nodes\":[{\"id\":\"team\"}]}}
  print(\"LINEAR CREATED:\",v[\"input\"][\"title\"]); return {\"issueCreate\":{\"issue\":{\"identifier\":\"ASK-999\"}}}
m.linear_module=lambda:LS
orig=pathlib.Path.open
def boom(self,*a,**k):
 if self.name==\"spillover.jsonl\": raise OSError(\"simulated ledger disk failure\")
 return orig(self,*a,**k)
pathlib.Path.open=boom
sys.argv=[\"spillover-promote.py\",\"sp-demo\",\"--title\",\"duplicate me\",\"--dor\",\"Allowed files: x.py Acceptance: verify behavior. This definition is intentionally longer than one hundred and twenty characters to pass the current shallow gate.\"]
try: m.main()
except Exception as e: print(type(e).__name__+\":\",e)
print(\"ledger status persisted: NO\")
print(\"retry consequence: creates ASK-999-like issue again\")''')"
```

Real output:

```text
LINEAR CREATED: duplicate me
OSError: simulated ledger disk failure
ledger status persisted: NO
retry consequence: creates ASK-999-like issue again
```

Linear creation needs an idempotency key or reconciliation path before retry. Otherwise a transient local write failure creates duplicate permanent issues that require human cleanup.

## What is sound

- All 32 changed Python files parse successfully.
- Four high-risk changed shell files pass `bash -n`.
- Notification receipt writes use `flock`, preventing concurrent JSONL writes from interleaving.
- Spillover promotion correctly refuses shallow DoRs before contacting Linear.
- Promotion records `promoted`, not `resolved`, so issue creation does not falsely claim the underlying finding was fixed.
- The WIP snapshot implementation uses a separate Git index and compare-and-swap ref update, preserving the real index and named branch.

## Verdict

**VERDICT: REQUEST CHANGES**

Fix the non-idempotent Linear promotion first. It creates permanent duplicate issues on an ordinary partial failure.

FINDINGS:
major|The timestamp-only watermark permanently hides an undelivered receipt appended in the same second as the previous read.|q-system/.q-system/scripts/notify-receipts-surface.py:113
major|A successful Linear issue creation followed by a ledger-write failure leaves the finding open, so retrying creates another permanent issue.|q-system/.q-system/scripts/spillover-promote.py:151
END FINDINGS
tokens used
93,007
## Findings

### MAJOR

The timestamp-only watermark permanently hides an undelivered receipt appended in the same second as the previous read.

`q-system/.q-system/scripts/notify-receipts-surface.py:113`

Reproducer:

```bash
python3 -c "import importlib.util,pathlib; p=pathlib.Path('q-system/.q-system/scripts/notify-receipts-surface.py'); s=importlib.util.spec_from_file_location('nrs',p); m=importlib.util.module_from_spec(s); s.loader.exec_module(m); row={'ts':'2026-08-20T10:15:30Z','delivered':False,'read':False,'message':'worker stalled'}; print('watermark:',row['ts']); print('unread same-second row:',m.undelivered_since([row],row['ts']))"
```

Real output:

```text
watermark: 2026-08-20T10:15:30Z
unread same-second row: []
```

Two notification writers can emit the same second-resolution timestamp. If the second append happens after the surface reads but before it saves the watermark, the next session skips that unread row forever. A stalled unattended worker can therefore lose its only operator signal.

### MAJOR

A successful Linear issue creation followed by a ledger-write failure leaves the finding open, so retrying creates another permanent issue.

`q-system/.q-system/scripts/spillover-promote.py:151`

Reproducer:

```bash
python3 -c "exec('''import importlib.util, pathlib, sys
p=pathlib.Path(\"q-system/.q-system/scripts/spillover-promote.py\")
s=importlib.util.spec_from_file_location(\"sp\",p)
m=importlib.util.module_from_spec(s); s.loader.exec_module(m)
m.load_rows=lambda ledger:{\"sp-demo\":{\"id\":\"sp-demo\",\"status\":\"open\",\"severity\":\"minor\",\"source\":\"review\",\"description\":\"live finding\"}}
class LS:
 ISSUE_CREATE=\"mutation\"
 @staticmethod
 def graphql(q,v):
  if q.startswith(\"query\"): return {\"teams\":{\"nodes\":[{\"id\":\"team\"}]}}
  print(\"LINEAR CREATED:\",v[\"input\"][\"title\"]); return {\"issueCreate\":{\"issue\":{\"identifier\":\"ASK-999\"}}}
m.linear_module=lambda:LS
orig=pathlib.Path.open
def boom(self,*a,**k):
 if self.name==\"spillover.jsonl\": raise OSError(\"simulated ledger disk failure\")
 return orig(self,*a,**k)
pathlib.Path.open=boom
sys.argv=[\"spillover-promote.py\",\"sp-demo\",\"--title\",\"duplicate me\",\"--dor\",\"Allowed files: x.py Acceptance: verify behavior. This definition is intentionally longer than one hundred and twenty characters to pass the current shallow gate.\"]
try: m.main()
except Exception as e: print(type(e).__name__+\":\",e)
print(\"ledger status persisted: NO\")
print(\"retry consequence: creates ASK-999-like issue again\")''')"
```

Real output:

```text
LINEAR CREATED: duplicate me
OSError: simulated ledger disk failure
ledger status persisted: NO
retry consequence: creates ASK-999-like issue again
```

Linear creation needs an idempotency key or reconciliation path before retry. Otherwise a transient local write failure creates duplicate permanent issues that require human cleanup.

## What is sound

- All 32 changed Python files parse successfully.
- Four high-risk changed shell files pass `bash -n`.
- Notification receipt writes use `flock`, preventing concurrent JSONL writes from interleaving.
- Spillover promotion correctly refuses shallow DoRs before contacting Linear.
- Promotion records `promoted`, not `resolved`, so issue creation does not falsely claim the underlying finding was fixed.
- The WIP snapshot implementation uses a separate Git index and compare-and-swap ref update, preserving the real index and named branch.

## Verdict

**VERDICT: REQUEST CHANGES**

Fix the non-idempotent Linear promotion first. It creates permanent duplicate issues on an ordinary partial failure.

FINDINGS:
major|The timestamp-only watermark permanently hides an undelivered receipt appended in the same second as the previous read.|q-system/.q-system/scripts/notify-receipts-surface.py:113
major|A successful Linear issue creation followed by a ledger-write failure leaves the finding open, so retrying creates another permanent issue.|q-system/.q-system/scripts/spillover-promote.py:151
END FINDINGS

PR #232, opened 2026-08-20
fix(token-discipline): scope the ENFORCED claims to what actually blocks (ASK-139)

assafkip:
## What

`.claude/rules/token-discipline.md` claimed ENFORCED three times and named no executable anywhere in the file. Its Layer-1 blocks ARE real. Two subsections were not.

- Intro now names `q-system/.q-system/token-guard.py` and its wiring (PreToolUse + PostToolUse + UserPromptSubmit, in BOTH `.claude/settings.json` and `settings-template.json`).
- Cleanup/Migration and Pre-Action Echo keep their `(ENFORCED)` markers and gain a scope line stating what the marker covers: the labelling, pinned by the new test. The behaviours are judgment; no gate sees them.

Line-neutral, 47 to 47. The file is always-on and the instruction-budget ratchet sits at its 511 baseline. `RATCHET PASS: always-on total 511 (baseline 511, target 300)`.

## Reproducer, red then green

New: `q-system/.q-system/scripts/test/test-token-discipline-rule-wired.sh`, modeled on the ASK-140 precedent `test-voice-enforcement-rule-wired.sh`.

Observed RED at 62e11397, before any fix:

```
$ bash q-system/.q-system/scripts/test/test-token-discipline-rule-wired.sh
FAIL: token-discipline.md claims ENFORCED but never names token-guard.py
EXIT=1
```

GREEN at 75f28ffe:

```
$ bash q-system/.q-system/scripts/test/test-token-discipline-rule-wired.sh
PASS: token-discipline.md names token-guard.py (exists, wired in .claude/settings.json + settings-template.json) and every ENFORCED heading names an executable
EXIT=0
```

Two checks, three negative self-tests against mutated copies. Check (b) was strengthened while still red: an ENFORCED section must name a script that EXISTS in the repo, and a negative self-test proves a plausible-but-fake name (`no-such-guard-ask139.py`) does not satisfy it. Naming any string ending in `.py` would have been a check that cannot fail.

The DoR's own blind-check, reproduced:

```
$ grep -n -i "pre-action\|two grep\|pass 2" q-system/.q-system/token-guard.py
grep rc=1   # no match: the two ENFORCED sections had no code
$ bash q-system/.q-system/scripts/test/test-token-guard-hook-behavior.sh
PASS: token-guard hook behavior (...)
```

Green and blind, exactly as the DoR described.

## No regression

```
test-token-guard-hook-behavior.sh      PASS
test-token-guard-template-wiring.sh    PASS (3 wirings on: UserPromptSubmit PreToolUse PostToolUse)
test_token_guard_observation.py        ALL PASS
```

Registered in `capability-manifest.json` under `expected_tests`.

## Two forks worth reviewing

**1. Markers were scoped, not removed. This was not my first choice.**

My first proposal relabelled both headings ADVISORY. `apply_claude_changes.py` refused it:

```
REFUSED: enforcement ratchet: 4 rule_marks entr(ies) would disappear,
first: token-discipline.md|enforced-heading|0
```

Its docstring: *"Demoting a rule to advisory is enforcement-weakening whether or not the prose is honest. If the claim is genuinely false, the fix is to correct the SCOPE."* Correct call by the gate, and the result is better than what I proposed. Both edits landed through the sanctioned `apply-claude-changes.sh` proposal path (`q-system/output/claude-changes/ask139-token-discipline.json`), never a direct write to `.claude/`.

**2. I took the DoR's prose route, not its hook route. Flagging it rather than skipping it.**

The DoR's Outcome names both: *"either backed by a real hook check or downgraded to advisory prose."* Its Check then sketched the hook-route reproducer (a PreToolUse payload asserting exit 2). I took the other named option, for a reason the DoR itself supplies in its Blast radius line.

Pre-Action Echo cannot be a hook, and this is structural rather than unfinished work:

- A PreToolUse hook receives `tool_name` and `tool_input`. It never sees the model's prose, so "echoed the plan in 2-3 bullets" is unobservable.
- "Wait for OK" requires a user turn. No unattended `claude -p` run has one, and the fleet now runs many: `linear-worker.sh`, the morning pipeline, `auto-commit.py`, every autonomous dispatch including the one that produced this PR.

A blocking branch would therefore deadlock the first Edit of every multi-file autonomous run fleet-wide, which is precisely the false positive the DoR warned about. The rule's own pre-existing `/q-morning` carve-out was the first sign of this; the fleet has more unattended runners now than when that exception was written. The reasoning is recorded in the rule so the next reader does not re-propose it.

Same reasoning for the two-grep-pass rule: deciding a task IS a rename is judgment, and "pass 1 done" is prose. Per `skill-hook-pairing.md`'s decision rule, both are the judgment half.

If you disagree and want the hook branch built anyway, that is a separate issue with a fleet-wide blast radius, not a follow-up to this one.

## Noticed, not fixed

`spillover-ratchet.py` surfaced 19 open notes against `capability-manifest.json` when I edited it (the "magnet file" problem, sp-f3a2ad81). All are already captured in the ledger, so nothing is orphaned. Triaging them is not in this issue's scope.

Closes nothing. Closeout runs through /issue-verify and /issue-closeout.

### DIFF
--- a/.claude/rules/token-discipline.md
+++ b/.claude/rules/token-discipline.md
@@ -6,9 +6,9 @@ paths:
 
 # Token Discipline
 
-When the token-guard hook blocks you (exit 2 message), follow its instructions exactly.
-Do not attempt to work around the block. The block exists because you are burning
-tokens without results.
+The executable is `q-system/.q-system/token-guard.py`, wired PreToolUse + PostToolUse +
+UserPromptSubmit in BOTH `.claude/settings.json` and `settings-template.json`. Layer 1
+below IS that script. When it blocks you (exit 2), follow it; do not work around it.
 
 Common blocks and what to do:
 - "3 retries": Something is broken. Diagnose the root cause. Tell the founder.
@@ -37,10 +37,10 @@ When doing cleanup, migration, or rename tasks: run TWO grep passes.
 - Pass 1 catches obvious string/symbol hits.
 - Pass 2 catches stale IDs, dead import paths, and embedded references in JSON, HTML, and markdown.
 - State "pass 1 done, starting pass 2" before finishing.
-- Not optional, even on small renames.
+- **Scope of (ENFORCED) above:** only the honest labelling, pinned by `test-token-discipline-rule-wired.sh`. The two passes themselves are not gated -- deciding a task IS a rename is judgment, and "pass 1 done" is your prose, so no hook can see either. Not optional; also not checked.
 
 ## Pre-Action Echo (ENFORCED)
 
 Before the first Edit, Write, or destructive Bash call: if the task touches more than one file OR more than one tool category (Read, Bash, Edit, Web, Agent), echo the plan in 2-3 bullets and wait for OK. No exceptions for "small" tasks. If you find yourself thinking "this is small, I'll just do it," that is the trigger to echo.
 
-**Exception:** Does not apply to agent pipeline sub-agents executing inside `/q-morning`. Pipeline phases run autonomously.
+**Scope of (ENFORCED) above: the labelling only, pinned by `test-token-discipline-rule-wired.sh`. The echo itself is structurally unhookable and that is not a TODO.** A PreToolUse hook receives `tool_name` and `tool_input`, never your prose, so "echoed the plan" is invisible to it; and "wait for OK" needs a user turn, which no unattended `claude -p` run has. A blocking branch would deadlock the first Edit of every autonomous run fleet-wide -- the `/q-morning` carve-out was the first sign of that, and the fleet has many more unattended runners now. Still does not apply to `/q-morning` pipeline sub-agents.
--- a/q-system/.q-system/capability-manifest.json
+++ b/q-system/.q-system/capability-manifest.json
@@ -445,6 +445,10 @@
       "runner": "bash"
     },
     {
+      "path": "q-system/.q-system/scripts/test/test-token-discipline-rule-wired.sh",
+      "runner": "bash"
+    },
+    {
       "path": "q-system/.q-system/scripts/test/test-token-guard-hook-behavior.sh",
       "runner": "bash"
     },
--- a/q-system/.q-system/scripts/test/test-token-discipline-rule-wired.sh
+++ b/q-system/.q-system/scripts/test/test-token-discipline-rule-wired.sh
@@ -11,7 +11,15 @@
 #       BOTH .claude/settings.json and settings-template.json (a script the fleet
 #       template never wires ships every instance a dead switch);
 #   (b) every "(ENFORCED)" heading in the rule names an executable inside its own
-#       section. A section that cannot name one is ADVISORY and must say so.
+#       section, and that executable EXISTS in this repo. A section that cannot name
+#       a real one has to state its scope honestly instead.
+#
+# (b) demands existence, not just a filename, because "name any plausible script" is
+# a check a hallucinated path would pass. The engine that lands .claude/ edits
+# (apply_claude_changes.py) refuses to let an agent delete an (ENFORCED marker at all
+# -- "demoting a rule to advisory is enforcement-weakening whether or not the prose is
+# honest" -- so the sanctioned repair is to correct the SCOPE under the marker, which
+# is what this check measures.
 #
 # Pairs with .claude/rules/token-discipline.md.
 set -euo pipefail
@@ -33,22 +41,38 @@ rule_names_guard() {
 rule_names_guard "$RULE" \
   || fail "token-discipline.md claims ENFORCED but never names $GUARD_NAME"
 
-# --- (b) no ENFORCED heading without an executable in its section ------------
-# Splits the file on markdown headings, keeps the sections whose heading says
-# ENFORCED, and demands a *.py / *.sh filename somewhere in that section's body.
-# This is the check the two prompt-only subsections failed.
+# --- (b) no ENFORCED heading without a REAL executable in its section --------
+# Walks the file, tracks the current heading, and for every heading carrying
+# "(ENFORCED" collects the *.py / *.sh basenames appearing in that section's body.
+# A section passes only if at least one of those names resolves to a file that
+# actually exists in this repo. This is the check the two prompt-only subsections
+# failed: they named nothing at all.
 enforced_sections_without_executable() {
-  awk '
-    /^#{1,6} / {
-      if (heading != "" && enforced && !named) print heading
-      heading = $0
-      enforced = (index(toupper($0), "(ENFORCED)") > 0)
-      named = 0
-      next
-    }
-    { if ($0 ~ /[A-Za-z0-9_-]+\.(py|sh)/) named = 1 }
-    END { if (heading != "" && enforced && !named) print heading }
-  ' "$1"
+  local rule_file="$1" line heading="" enforced=0 satisfied=0 name
+  emit() {
+    [ "$enforced" -eq 1 ] && [ "$satisfied" -eq 0 ] && [ -n "$heading" ] \
+      && printf '%s\n' "$heading"
+    return 0
+  }
+  while IFS= read -r line; do
+    case "$line" in
+      '#'*' '*)
+        emit
+        heading="$line"
+        case "$line" in *'(ENFORCED'*) enforced=1 ;; *) enforced=0 ;; esac
+        satisfied=0
+        continue
+        ;;
+    esac
+    [ "$enforced" -eq 1 ] || continue
+    # Any basename ending .py/.sh on this line; satisfied once one of them is real.
+    for name in $(printf '%s\n' "$line" | grep -oE '[A-Za-z0-9_-]+\.(py|sh)' || true); do
+      if [ -n "$(find "$ROOT" -name "$name" -not -path '*/.git/*' -print -quit)" ]; then
+        satisfied=1
+      fi
+    done
+  done < "$rule_file"
+  emit
 }
 
 orphans="$(enforced_sections_without_executable "$RULE")"
@@ -70,6 +94,15 @@ printf '\n## Synthetic Claim (ENFORCED)\n\nNo executable is named in this sectio
   >> "$TMPDIR_TEST/orphaned.md"
 if [ -z "$(enforced_sections_without_executable "$TMPDIR_TEST/orphaned.md")" ]; then
   fail "negative self-test: the ENFORCED-heading check passed on a copy carrying a scriptless ENFORCED section"
+fi
+
+# ...and a plausible-but-nonexistent filename must NOT satisfy it, or the check
+# degrades into "name any string ending in .py".
+cp "$RULE" "$TMPDIR_TEST/hallucinated.md"
+printf '\n## Synthetic Claim (ENFORCED)\n\nEnforced by `no-such-guard-ask139.py`.\n' \
+  >> "$TMPDIR_TEST/hallucinated.md"
+if [ -z "$(enforced_sections_without_executable "$TMPDIR_TEST/hallucinated.md")" ]; then
+  fail "negative self-test: a nonexistent script name satisfied the ENFORCED-heading check"
 fi
 
 # --- the named executable is real and wired ----------------------------------
--- a/q-system/.q-system/scripts/test/test-token-discipline-rule-wired.sh
+++ b/q-system/.q-system/scripts/test/test-token-discipline-rule-wired.sh
@@ -1,0 +1,86 @@
+#!/usr/bin/env bash
+# ASK-139: .claude/rules/token-discipline.md carried three ENFORCED-shaped claims and
+# named no executable anywhere in the file. Its Layer-1 blocks ARE real (token-guard.py,
+# wired on three events in both settings files), but two subsections -- the
+# Cleanup/Migration two-grep-pass and Pre-Action Echo -- claimed ENFORCED with nothing
+# behind them. `grep -n -i "pre-action\|two grep\|pass 2" token-guard.py` returned
+# nothing while every token-guard test stayed green: green and blind.
+#
+# This test pins both halves so the file cannot drift back:
+#   (a) the rule NAMES its enforcing executable, and that file exists and is wired in
+#       BOTH .claude/settings.json and settings-template.json (a script the fleet
+#       template never wires ships every instance a dead switch);
+#   (b) every "(ENFORCED)" heading in the rule names an executable inside its own
+#       section. A section that cannot name one is ADVISORY and must say so.
+#
+# Pairs with .claude/rules/token-discipline.md.
+set -euo pipefail
+ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../../../.." && pwd)"
+RULE="$ROOT/.claude/rules/token-discipline.md"
+GUARD_REL="q-system/.q-system/token-guard.py"
+GUARD_NAME="token-guard.py"
+fail() { echo "FAIL: $1" >&2; exit 1; }
+
+[ -f "$RULE" ] || fail "rule file missing at $RULE"
+
+# --- (a) the rule names its enforcing executable -----------------------------
+# Factored so the negative self-test below runs the SAME check against a copy and
+# proves it can go red. A check that cannot fail is decoration.
+rule_names_guard() {
+  grep -qF "$GUARD_NAME" "$1"
+}
+
+rule_names_guard "$RULE" \
+  || fail "token-discipline.md claims ENFORCED but never names $GUARD_NAME"
+
+# --- (b) no ENFORCED heading without an executable in its section ------------
+# Splits the file on markdown headings, keeps the sections whose heading says
+# ENFORCED, and demands a *.py / *.sh filename somewhere in that section's body.
+# This is the check the two prompt-only subsections failed.
+enforced_sections_without_executable() {
+  awk '
+    /^#{1,6} / {
+      if (heading != "" && enforced && !named) print heading
+      heading = $0
+      enforced = (index(toupper($0), "(ENFORCED)") > 0)
+      named = 0
+      next
+    }
+    { if ($0 ~ /[A-Za-z0-9_-]+\.(py|sh)/) named = 1 }
+    END { if (heading != "" && enforced && !named) print heading }
+  ' "$1"
+}
+
+orphans="$(enforced_sections_without_executable "$RULE")"
+[ -z "$orphans" ] || fail "ENFORCED heading(s) in token-discipline.md name no executable:
+$orphans
+Either wire one and name it, or relabel the section ADVISORY."
+
+# --- negative self-tests: both checks must go red on a mutated copy ----------
+TMPDIR_TEST="$(mktemp -d)"
+trap 'rm -rf "$TMPDIR_TEST"' EXIT
+
+sed "s/$GUARD_NAME//g" "$RULE" > "$TMPDIR_TEST/stripped.md"
+if rule_names_guard "$TMPDIR_TEST/stripped.md"; then
+  fail "negative self-test: rule_names_guard passed on a copy with the name stripped"
+fi
+
+cp "$RULE" "$TMPDIR_TEST/orphaned.md"
+printf '\n## Synthetic Claim (ENFORCED)\n\nNo executable is named in this section.\n' \
+  >> "$TMPDIR_TEST/orphaned.md"
+if [ -z "$(enforced_sections_without_executable "$TMPDIR_TEST/orphaned.md")" ]; then
+  fail "negative self-test: the ENFORCED-heading check passed on a copy carrying a scriptless ENFORCED section"
+fi
+
+# --- the named executable is real and wired ----------------------------------
+# Not asserting the exec bit: both settings entries invoke it as `python3 <path>`,
+# so chmod +x is not what makes this wiring live. Asserting -x here would be a
+# check that passes or fails for a reason nobody depends on.
+[ -f "$ROOT/$GUARD_REL" ] || fail "$GUARD_NAME is named by the rule but missing at $GUARD_REL"
+
+grep -qF "$GUARD_REL" "$ROOT/.claude/settings.json" \
+  || fail ".claude/settings.json does not wire $GUARD_NAME"
+grep -qF "$GUARD_REL" "$ROOT/settings-template.json" \
+  || fail "settings-template.json does not wire $GUARD_NAME (fleet would get a dead switch)"
+
+echo "PASS: token-discipline.md names $GUARD_NAME (exists, wired in .claude/settings.json + settings-template.json) and every ENFORCED heading names an executable"


assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-232-20260820-145533.md` (56872 bytes).

```
FINDINGS:
minor|The wiring assertion passes after two of the three claimed hook registrations are removed.|q-system/.q-system/scripts/test/test-token-discipline-rule-wired.sh:114
minor|The section check accepts any existing script basename and therefore does not pin the claimed labelling-only scope.|q-system/.q-system/scripts/test/test-token-discipline-rule-wired.sh:69
END FINDINGS
```

--- reviewer output, last 54616 bytes of 56872 (full review at the path above) ---

   or a major.

work that did not need doing.

process, not a failure of it.

## Output

Then:

FINDINGS:
END FINDINGS
warning: Skill descriptions were shortened to fit the skills context budget. Codex can still see every skill, but some descriptions are shorter. Disable unused skills or plugins to leave more room for the rest.
codex
I’m applying the required founder-voice and AUDHD output rules while keeping this review read-only and headless. I’m starting with the PR diff, producer shapes, and executable failure probes.
exec
/bin/zsh -lc "pwd && sed -n '1,240p' /Users/assafkipnis/.agents/skills/assaf-voice/SKILL.md && sed -n '1,240p' /Users/assafkipnis/.agents/skills/audhd-executive-function/SKILL.md && sed -n '1,260p' q-system/AGENTS.md && gh -R assafkip/kipi-system pr view 232 && gh -R assafkip/kipi-system pr diff 232" in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-232
 exited 1 in 0ms:
/Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-232
---
name: assaf-voice
description: ALWAYS USE THIS SKILL for ANY written output - LinkedIn posts, Medium articles, emails, DMs, talk tracks, outreach messages, slide copy, Notion content, investor updates, comments, replies, or any other text that will be read by another human. This includes both ad-hoc requests ("write me a reply to X") and planned/scheduled content (morning routine posts, marketing pipeline, engagement hitlist comments). This skill is NOT optional - it is the default voice layer for all written output in this system.
metadata:
  version: 2.0.0
---

# Assaf Voice

The voice lives in ONE place and it is not in this skill:
`~/projects/consulting/q-consult/voice/`.

This file carries no voice prose of its own. It tells you what to load and how
little. The v1 layer files this skill used to load (`references/layers/*.md`,
`voice-dna.md`, `writing-samples.md`) are RETIRED. They were a second copy of the
corpus, and two copies is the drift machine that was removed on 2026-08-13
(ASK-699). v1 also pointed at `~/.Codex/skills/...` paths that do not exist on
this machine, so its documented loading table could not have been followed.

## Get the reference by running the selector

Do not hand-pick exemplars and do not paste the corpus.

```bash
python3 ~/projects/social-voice/scripts/voice_ref.py --channel x --words 480
python3 ~/projects/social-voice/scripts/voice_ref.py --channel x --words 25
python3 ~/projects/social-voice/scripts/voice_ref.py --channel linkedin --words 200
```

It prints identity, the live corrections with their scopes honoured, and the
exemplars matched to the piece. It exits non-zero when the corpus is missing
rather than returning a silent blank.

**`--words` is not cosmetic.** Channel and LENGTH are two axes. The x corpus runs
5 to 55 words with a single 479-word row, so a long piece written against the
short rows comes back formal. That was measured on 2026-08-13; the founder's
verdict on the draft was "this is not in my voiccc". Pass the real target length.

**One long-form x exemplar exists.** The selector says so in its output. Treat a
long-form register as thin evidence, not a settled pattern.

## Routing

- **Social posts** go through the posting lane, not a hand-written draft:
  `python3 -m pipeline.social idea "<his idea>" --channel x --words <n>` from
  `~/projects/consulting/q-consult/`. It runs the deterministic gate stack and
  never publishes.
- **Reactions to someone else's content:** extract the poster's claims and
  confirm them with the founder BEFORE drafting.
- **Everything else is this skill:** emails, DMs, comments, replies, talk
  tracks, slide copy, Notion content, outreach, investor updates, decks.

## The can't-write rules

- **No emdashes**, anywhere. In docs and code the substitute is ` -- `; in posts
  neither is allowed.
- **Banned words and phrases** live in `voice-lint.py`, not restated here.
  Measure any new list entry against the real corpus first: a word list caught
  the founder's own ordinary vocabulary six times in one session.
- **Never invent a person or company fact.** No checker sees this one.
- **$1,000 is the only public price.**
- **No client names** without recorded permission.
- **DMs and emails open with "I"**, never the person's name. No "circling back",
  no "just checking in".

## After drafting

`voice-lint.py` and `voice-substance-lint.py` run on write and a Stop gate honours
no skip marker. They prove the ABSENCE of known defects. They never prove the
presence of his voice; only the exemplars answer that.
---
name: audhd-executive-function
description: "AUDHD executive function accommodations. Apply to all output the founder will act on."
---

# AUDHD Executive Function Skill -- Investigation Mode

External executive function system for investigation outputs. Governs how briefs, action items, and daily plans are structured.

**Always read:** `~/.Codex/skills/audhd-executive-function/references/rules.md`

---

## THE ONE RULE

**If the founder cannot copy-paste it, click it, or check it off, it does not belong in the output.**

---

## ACTIONABILITY RULES (ENFORCED)

### A1: No Cross-References
Never say "see section above" or "refer to file Y." Every item contains the actual content inline. The founder never scrolls or searches.

### A2: Next Physical Action
Every item is the literal next physical thing to do. Not "Follow up on Rickenson lead" but the actual search query, URL to open, or message to send.

### A3: All Pending Actions Get Specifics
Every action item includes: what to do, where to do it (URL/platform), expected time, and energy tag.

### A4: Friction-Ordered
Items sorted by friction, lowest first. 2-minute username checks before 15-minute browser deep dives. Quick momentum wins first.

---

## INVESTIGATION OUTPUT RULES

### Briefs and Reports
- Lead with findings, not methodology
- Confidence levels on every assessment (Confirmed / High / Moderate / Low / Speculative)
- Evidence citations inline, not in footnotes
- Collection gaps listed as actionable next steps, not as "we don't know"

### Action Items
Every collection task includes:
- What to collect
- Where (platform + URL if known)
- Tool to use (script command or Chrome)
- Energy tag (Quick Win / Deep Focus / People / Admin)
- Time estimate
- What it proves if successful

### Evidence Summaries
- Bullet points, not paragraphs
- Source reliability rating (A-F) on every claim
- Distinguish: confirmed fact / assessed judgment / speculation
- Flag contradictions explicitly

---

## LANGUAGE RULES (ENFORCED)

### Never Use
- "overdue" (use "carried forward")
- "missed" (use "ready when you are")
- "failed" (use "didn't land" or "inconclusive")
- "forgot" (use "not yet done")
- "behind" (use "in progress")
- "urgent" as pressure (state facts calmly)
- "you need to" (use "you could")
- "nobody responded" (use "awaiting reply")

### Always Use
- Effort tracking: "You ran 15 searches across 4 platforms" not outcome tracking
- "Carried forward from last session" not "this is overdue"
- "You could..." not "you need to..."
- Calm factual statements about investigation state

### Progress Framing
- "3 of 5 targets profiled" is a win. Show what WAS done
- End of session: effort celebration, not gap shaming
- Completed items stay visible so the founder sees accomplishments

---

## DECISION ELIMINATION

### The System Decides
- Which collection task is highest priority (based on hypothesis impact)
- What tool to use (based on diagnose.sh results)
- In what order (friction-sorted)

### The Founder Executes
- Approves the plan
- Provides client responses
- Makes judgment calls on ambiguous evidence

### Never Present Options Without a Recommendation
Bad: "We could search Facebook OR LinkedIn OR both. What do you prefer?"
Good: "Recommend Facebook first (higher probability of finding connections). LinkedIn second if Facebook is dry. Time: 20 min total."

---

## ANTI-PATTERNS (NEVER DO)

1. **Walls of text.** Bullet points or tables. Always.
2. **Abstract task names.** "Investigate Rickenson" is not a task. "Search Facebook for 'Rickenson' + 'Palmyre' together" is.
3. **Options without recommendations.** The system picks. The founder overrides if they want.
4. **Stacking pressure.** Never list 3+ carried-forward items together.
5. **Outcome metrics as judgment.** "0 results found" is data. Never frame it as failure.
6. **Sequential dependencies without flagging.** If step 3 needs step 1's output, say so explicitly.
sed: q-system/AGENTS.md: No such file or directory

codex
The project-level include named in the supplied instructions is absent at `q-system/AGENTS.md`. I’m checking the actual tree and reading the required AUDHD rules from `/Users/assafkipnis/.Codex/skills/...` because the selected skill explicitly requires that reference.
exec
/bin/zsh -lc "sed -n '1,260p' /Users/assafkipnis/.Codex/skills/audhd-executive-function/references/rules.md; rg --files -g 'AGENTS.md' -g '"'!node_modules'"'; gh -R assafkip/kipi-system pr view 232; gh -R assafkip/kipi-system pr diff 232" in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-232
 exited 1 in 72ms:
error connecting to api.github.com
check your internet connection or https://githubstatus.com
error connecting to api.github.com
check your internet connection or https://githubstatus.com

codex
GitHub is unreachable in this sandbox, so I’m using the checked-out review tree and local refs to reconstruct PR #232. The AUDHD skill’s mandatory reference path is also stale, so I’ll follow the rules embedded in the loaded skill itself.
exec
/bin/zsh -lc "sed -n '1,320p' AGENTS.md; git status --short --branch; git branch --all --verbose --no-abbrev; git log --oneline --decorate -20; git remote -v; git merge-base HEAD origin/main 2>/dev/null || git merge-base HEAD main" in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-232
 succeeded in 4ms:
197dcfc79026f11d2fdc9a73331780306425b162

exec
/bin/zsh -lc 'git diff --stat 197dcfc7..HEAD; git diff --name-status 197dcfc7..HEAD; git diff --find-renames --find-copies 197dcfc7..HEAD; git show --stat --oneline 62e11397; git show --stat --oneline 75f28ffe' in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-232
 succeeded in 0ms:
 .claude/rules/token-discipline.md                  |  10 +-
 q-system/.q-system/capability-manifest.json        |   4 +
 .../test/test-token-discipline-rule-wired.sh       | 119 +++++++++++++++++++++
 3 files changed, 128 insertions(+), 5 deletions(-)
M	.claude/rules/token-discipline.md
M	q-system/.q-system/capability-manifest.json
A	q-system/.q-system/scripts/test/test-token-discipline-rule-wired.sh
diff --git a/.claude/rules/token-discipline.md b/.claude/rules/token-discipline.md
index 371495ac..98900a7b 100644
--- a/.claude/rules/token-discipline.md
+++ b/.claude/rules/token-discipline.md
@@ -6,9 +6,9 @@ paths:
 
 # Token Discipline
 
-When the token-guard hook blocks you (exit 2 message), follow its instructions exactly.
-Do not attempt to work around the block. The block exists because you are burning
-tokens without results.
+The executable is `q-system/.q-system/token-guard.py`, wired PreToolUse + PostToolUse +
+UserPromptSubmit in BOTH `.claude/settings.json` and `settings-template.json`. Layer 1
+below IS that script. When it blocks you (exit 2), follow it; do not work around it.
 
 Common blocks and what to do:
 - "3 retries": Something is broken. Diagnose the root cause. Tell the founder.
@@ -37,10 +37,10 @@ When doing cleanup, migration, or rename tasks: run TWO grep passes.
 - Pass 1 catches obvious string/symbol hits.
 - Pass 2 catches stale IDs, dead import paths, and embedded references in JSON, HTML, and markdown.
 - State "pass 1 done, starting pass 2" before finishing.
-- Not optional, even on small renames.
+- **Scope of (ENFORCED) above:** only the honest labelling, pinned by `test-token-discipline-rule-wired.sh`. The two passes themselves are not gated -- deciding a task IS a rename is judgment, and "pass 1 done" is your prose, so no hook can see either. Not optional; also not checked.
 
 ## Pre-Action Echo (ENFORCED)
 
 Before the first Edit, Write, or destructive Bash call: if the task touches more than one file OR more than one tool category (Read, Bash, Edit, Web, Agent), echo the plan in 2-3 bullets and wait for OK. No exceptions for "small" tasks. If you find yourself thinking "this is small, I'll just do it," that is the trigger to echo.
 
-**Exception:** Does not apply to agent pipeline sub-agents executing inside `/q-morning`. Pipeline phases run autonomously.
+**Scope of (ENFORCED) above: the labelling only, pinned by `test-token-discipline-rule-wired.sh`. The echo itself is structurally unhookable and that is not a TODO.** A PreToolUse hook receives `tool_name` and `tool_input`, never your prose, so "echoed the plan" is invisible to it; and "wait for OK" needs a user turn, which no unattended `claude -p` run has. A blocking branch would deadlock the first Edit of every autonomous run fleet-wide -- the `/q-morning` carve-out was the first sign of that, and the fleet has many more unattended runners now. Still does not apply to `/q-morning` pipeline sub-agents.
diff --git a/q-system/.q-system/capability-manifest.json b/q-system/.q-system/capability-manifest.json
index 7b5d6d9f..c59b3b0b 100644
--- a/q-system/.q-system/capability-manifest.json
+++ b/q-system/.q-system/capability-manifest.json
@@ -444,6 +444,10 @@
       "path": "q-system/.q-system/scripts/test/test-terminal-states.sh",
       "runner": "bash"
     },
+    {
+      "path": "q-system/.q-system/scripts/test/test-token-discipline-rule-wired.sh",
+      "runner": "bash"
+    },
     {
       "path": "q-system/.q-system/scripts/test/test-token-guard-hook-behavior.sh",
       "runner": "bash"
diff --git a/q-system/.q-system/scripts/test/test-token-discipline-rule-wired.sh b/q-system/.q-system/scripts/test/test-token-discipline-rule-wired.sh
new file mode 100755
index 00000000..42db8411
--- /dev/null
+++ b/q-system/.q-system/scripts/test/test-token-discipline-rule-wired.sh
@@ -0,0 +1,119 @@
+#!/usr/bin/env bash
+# ASK-139: .claude/rules/token-discipline.md carried three ENFORCED-shaped claims and
+# named no executable anywhere in the file. Its Layer-1 blocks ARE real (token-guard.py,
+# wired on three events in both settings files), but two subsections -- the
+# Cleanup/Migration two-grep-pass and Pre-Action Echo -- claimed ENFORCED with nothing
+# behind them. `grep -n -i "pre-action\|two grep\|pass 2" token-guard.py` returned
+# nothing while every token-guard test stayed green: green and blind.
+#
+# This test pins both halves so the file cannot drift back:
+#   (a) the rule NAMES its enforcing executable, and that file exists and is wired in
+#       BOTH .claude/settings.json and settings-template.json (a script the fleet
+#       template never wires ships every instance a dead switch);
+#   (b) every "(ENFORCED)" heading in the rule names an executable inside its own
+#       section, and that executable EXISTS in this repo. A section that cannot name
+#       a real one has to state its scope honestly instead.
+#
+# (b) demands existence, not just a filename, because "name any plausible script" is
+# a check a hallucinated path would pass. The engine that lands .claude/ edits
+# (apply_claude_changes.py) refuses to let an agent delete an (ENFORCED marker at all
+# -- "demoting a rule to advisory is enforcement-weakening whether or not the prose is
+# honest" -- so the sanctioned repair is to correct the SCOPE under the marker, which
+# is what this check measures.
+#
+# Pairs with .claude/rules/token-discipline.md.
+set -euo pipefail
+ROOT="$(cd "$(dirname "${BASH_SOURCE[0]}")/../../../.." && pwd)"
+RULE="$ROOT/.claude/rules/token-discipline.md"
+GUARD_REL="q-system/.q-system/token-guard.py"
+GUARD_NAME="token-guard.py"
+fail() { echo "FAIL: $1" >&2; exit 1; }
+
+[ -f "$RULE" ] || fail "rule file missing at $RULE"
+
+# --- (a) the rule names its enforcing executable -----------------------------
+# Factored so the negative self-test below runs the SAME check against a copy and
+# proves it can go red. A check that cannot fail is decoration.
+rule_names_guard() {
+  grep -qF "$GUARD_NAME" "$1"
+}
+
+rule_names_guard "$RULE" \
+  || fail "token-discipline.md claims ENFORCED but never names $GUARD_NAME"
+
+# --- (b) no ENFORCED heading without a REAL executable in its section --------
+# Walks the file, tracks the current heading, and for every heading carrying
+# "(ENFORCED" collects the *.py / *.sh basenames appearing in that section's body.
+# A section passes only if at least one of those names resolves to a file that
+# actually exists in this repo. This is the check the two prompt-only subsections
+# failed: they named nothing at all.
+enforced_sections_without_executable() {
+  local rule_file="$1" line heading="" enforced=0 satisfied=0 name
+  emit() {
+    [ "$enforced" -eq 1 ] && [ "$satisfied" -eq 0 ] && [ -n "$heading" ] \
+      && printf '%s\n' "$heading"
+    return 0
+  }
+  while IFS= read -r line; do
+    case "$line" in
+      '#'*' '*)
+        emit
+        heading="$line"
+        case "$line" in *'(ENFORCED'*) enforced=1 ;; *) enforced=0 ;; esac
+        satisfied=0
+        continue
+        ;;
+    esac
+    [ "$enforced" -eq 1 ] || continue
+    # Any basename ending .py/.sh on this line; satisfied once one of them is real.
+    for name in $(printf '%s\n' "$line" | grep -oE '[A-Za-z0-9_-]+\.(py|sh)' || true); do
+      if [ -n "$(find "$ROOT" -name "$name" -not -path '*/.git/*' -print -quit)" ]; then
+        satisfied=1
+      fi
+    done
+  done < "$rule_file"
+  emit
+}
+
+orphans="$(enforced_sections_without_executable "$RULE")"
+[ -z "$orphans" ] || fail "ENFORCED heading(s) in token-discipline.md name no executable:
+$orphans
+Either wire one and name it, or relabel the section ADVISORY."
+
+# --- negative self-tests: both checks must go red on a mutated copy ----------
+TMPDIR_TEST="$(mktemp -d)"
+trap 'rm -rf "$TMPDIR_TEST"' EXIT
+
+sed "s/$GUARD_NAME//g" "$RULE" > "$TMPDIR_TEST/stripped.md"
+if rule_names_guard "$TMPDIR_TEST/stripped.md"; then
+  fail "negative self-test: rule_names_guard passed on a copy with the name stripped"
+fi
+
+cp "$RULE" "$TMPDIR_TEST/orphaned.md"
+printf '\n## Synthetic Claim (ENFORCED)\n\nNo executable is named in this section.\n' \
+  >> "$TMPDIR_TEST/orphaned.md"
+if [ -z "$(enforced_sections_without_executable "$TMPDIR_TEST/orphaned.md")" ]; then
+  fail "negative self-test: the ENFORCED-heading check passed on a copy carrying a scriptless ENFORCED section"
+fi
+
+# ...and a plausible-but-nonexistent filename must NOT satisfy it, or the check
+# degrades into "name any string ending in .py".
+cp "$RULE" "$TMPDIR_TEST/hallucinated.md"
+printf '\n## Synthetic Claim (ENFORCED)\n\nEnforced by `no-such-guard-ask139.py`.\n' \
+  >> "$TMPDIR_TEST/hallucinated.md"
+if [ -z "$(enforced_sections_without_executable "$TMPDIR_TEST/hallucinated.md")" ]; then
+  fail "negative self-test: a nonexistent script name satisfied the ENFORCED-heading check"
+fi
+
+# --- the named executable is real and wired ----------------------------------
+# Not asserting the exec bit: both settings entries invoke it as `python3 <path>`,
+# so chmod +x is not what makes this wiring live. Asserting -x here would be a
+# check that passes or fails for a reason nobody depends on.
+[ -f "$ROOT/$GUARD_REL" ] || fail "$GUARD_NAME is named by the rule but missing at $GUARD_REL"
+
+grep -qF "$GUARD_REL" "$ROOT/.claude/settings.json" \
+  || fail ".claude/settings.json does not wire $GUARD_NAME"
+grep -qF "$GUARD_REL" "$ROOT/settings-template.json" \
+  || fail "settings-template.json does not wire $GUARD_NAME (fleet would get a dead switch)"
+
+echo "PASS: token-discipline.md names $GUARD_NAME (exists, wired in .claude/settings.json + settings-template.json) and every ENFORCED heading names an executable"
62e11397 test(token-discipline): reproducer for the prompt-only ENFORCED claims (ASK-139)
 .../test/test-token-discipline-rule-wired.sh       | 86 ++++++++++++++++++++++
 1 file changed, 86 insertions(+)
75f28ffe fix(token-discipline): scope the ENFORCED claims to what actually blocks (ASK-139)
 .claude/rules/token-discipline.md                  | 10 ++--
 q-system/.q-system/capability-manifest.json        |  4 ++
 .../test/test-token-discipline-rule-wired.sh       | 65 ++++++++++++++++------
 3 files changed, 58 insertions(+), 21 deletions(-)

exec
/bin/zsh -lc "rg -n '3 retries|50 tool|25 sub|30 MCP|15 reads|searches without|edit attempts|agents no output|minutes since|exit\\(2\\)|sys.exit\\(2\\)' q-system/.q-system/token-guard.py; sed -n '1,280p' q-system/.q-system/token-guard.py; git diff --check 197dcfc7..HEAD; python3 -m json.tool q-system/.q-system/capability-manifest.json >/dev/null && echo 'manifest JSON: PASS'; git status --short" in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-232
 succeeded in 0ms:
362:    # Track edit attempts per file
495:        return f"{count} searches without producing output. You're searching, not working. Pick a direction."
500:    """Block if too many edit attempts on the same file."""
507:        return f"{count} edit attempts on {short}. The approach isn't working. Read the file again, find the exact string, or tell the founder what's wrong."
955:    sys.exit(2)
1297:    then a generic label — the block message must never fall back to "50 tool
1567:    # spiralling on the gate file got GATE_GRACE unchecked edit attempts and
#!/usr/bin/env python3
"""
Token Bleed Guardrail System
Two-layer defense against runaway token consumption in Claude Code sessions.
Layer 1: Hook-based circuit breaker (this script).
Layer 2: CLAUDE.md + .claude/rules/token-discipline.md (behavioral).

Called by three hooks:
  - PreToolUse: counts tool calls and enforces limits
  - PostToolUse (Edit|Write|MultiEdit|Bash): progress resets (successful edit
    clears that file's spiral counter; successful commit clears the volume
    counters)
  - UserPromptSubmit: resets per-message counters + sweeps stale caches

Per-actor budget scoping (sp0-guard-actor-scope):
  Counters are keyed per ACTOR, not per session. Subagent (Task/workflow)
  tool calls fire hooks with the parent's session_id and transcript_path but
  carry their own `agent_id` field — observed live in
  q-system/output/guard-payload-dump.json (2026-06-11): main-session events
  have no agent_id; each spawned agent's events carry a distinct agent_id.
  Cache key: /tmp/claude-guard-{session_id}.json for the main actor,
  /tmp/claude-guard-{session_id}-agent-{agent_id}.json per subagent. A
  spinning single actor still hits its own ceiling; a fan-out no longer
  burns the orchestrator's budget (and a blocked subagent no longer
  deadlocks the parent).

Commit-progress valve (both wirings):
  A successful `git commit` resets the volume counters — the ceiling gates
  lack-of-progress, not raw volume (founder-approved 2026-06-11). Wiring A:
  PostToolUse on Bash (settings.json matcher includes Bash). Wiring B:
  PreToolUse checks whether the repo HEAD commit is newer than the last
  volume reset — this works even when PostToolUse delivery is missing, and
  takes effect live (the script is re-read every hook fire). The
  empty-commit edge is accepted, not defended (same posture as the
  commit-string volume exemption below).

Exit codes:
  0 = allow (optionally with warning via stdout JSON)
  2 = block (stderr message goes to Claude as feedback)
"""

import calendar
import contextlib
import fcntl
import hashlib
import json
import os
import re
import shlex
import stat
import sys
import time


# --- Thresholds ---
RETRY_LIMIT = 3            # Same tool+input N times = block
VOLUME_CEILING = 50         # Tool calls since last user message = block
VOLUME_WARNING = 35         # Tool calls since last user message = warn
AGENT_CEILING = 30          # Agent spawns per user message = block (morning routine needs ~25)
MCP_RATE_WINDOW = 60        # Seconds
MCP_RATE_LIMIT = 30         # MCP calls in window = block
READ_SPIRAL_LIMIT = 15      # Consecutive reads without write = warn
FILE_REREAD_LIMIT = 3       # Same file path read N times = warn
GREP_DRIFT_LIMIT = 5        # Greps since last write = warn
EDIT_FAIL_LIMIT = 3         # Edit attempts on same file without success = block
AGENT_NO_OUTPUT_LIMIT = 3   # Agent spawns with no write between them = warn
STALL_TIME_SECONDS = 120    # Seconds since last write + calls = warn
STALL_MIN_CALLS = 10        # Minimum calls before time-based stall triggers
GATE_GRACE = 8              # Calls granted at the ceiling to clear a gate that refused the commit
GATE_GRACE_GRANTS = 1       # Grants per user-message window (bounded: no re-arming ratchet)

# Sensitive file patterns
SENSITIVE_PATTERNS = (".env", ".pem", ".key", "credentials")

# Read-only browser/desktop observation tools. Repeating one of these with
# identical input is legitimate polling of a CHANGING screen, not a stuck
# retry — the hash can't see that the page behind it moved. Scar
# (sp-ff7611cd, cole-gtm 2026-07): LinkedIn's compose overlay is not in the
# DOM, so a screenshot is the ONLY way to verify the recipient before
# typing; check_exact_retry blocked the 3rd identical screenshot, the loop
# correctly refused to blind-send, and the DM lane went quiet. Exemption is
# scoped to check_exact_retry ONLY — the volume ceiling still catches a
# genuinely runaway observation loop.
OBSERVATION_TOOLS = frozenset((
    "mcp__computer-use__screenshot",
    "mcp__computer-use__cursor_position",
    "mcp__claude-in-chrome__read_page",
    "mcp__claude-in-chrome__get_page_text",
    "mcp__claude-in-chrome__tabs_context_mcp",
    "mcp__playwright__browser_snapshot",
    "mcp__playwright__browser_take_screenshot",
    "mcp__plugin_chrome-devtools-mcp_chrome-devtools__take_screenshot",
    "mcp__plugin_chrome-devtools-mcp_chrome-devtools__take_snapshot",
))
# Multi-action tools where only the observation action is read-only.
OBSERVATION_ACTIONS = {"mcp__claude-in-chrome__computer": ("screenshot", "zoom")}


def is_readonly_observation(tool_name, tool_input):
    if tool_name in OBSERVATION_TOOLS:
        return True
    actions = OBSERVATION_ACTIONS.get(tool_name)
    if actions:
        return (tool_input or {}).get("action") in actions
    return False


CACHE_TTL_DAYS = 7          # Stale guard caches in /tmp older than this are swept


def actor_cache_key(hook_input):
    """Per-actor cache key. Subagents share the parent's session_id but carry
    their own agent_id (evidence: q-system/output/guard-payload-dump.json) —
    compounding the two gives each actor its own budget. No agent_id = the
    main session actor."""
    session_id = hook_input.get("session_id", "unknown")
    agent_id = hook_input.get("agent_id")
    if agent_id:
        return f"{session_id}-agent-{agent_id}"
    return session_id


def cache_path(actor_key):
    return f"/tmp/claude-guard-{actor_key}.json"


def sweep_stale_caches(max_age_days=CACHE_TTL_DAYS):
    """Delete guard caches older than max_age_days. Runs on UserPromptSubmit
    only (once per user message), so the per-tool-call path stays free of
    directory scans. Subagent caches have no UserPromptSubmit of their own;
    this sweep is their cleanup path."""
    import glob
    cutoff = time.time() - max_age_days * 86400
    # Both patterns: the sidecar lock files (sp-016776e6) do not end in .json,
    # so the original glob would have left one per actor in /tmp forever.
    stale = glob.glob("/tmp/claude-guard-*.json") + glob.glob(
        "/tmp/claude-guard-*.lock")
    for path in stale:
        try:
            if os.path.getmtime(path) < cutoff:
                os.remove(path)
        except OSError:
            pass


def load_cache(actor_key):
    path = cache_path(actor_key)
    if os.path.exists(path):
        try:
            with open(path) as f:
                return json.load(f)
        except (json.JSONDecodeError, IOError):
            pass
    return {
        "actor_key": actor_key,
        "tool_calls_since_user": 0,
        "agent_calls_since_user": 0,
        "mcp_timestamps": [],
        "repeat_map": {},
        "consecutive_reads": 0,
        "warnings_issued": 0,
        "file_read_counts": {},
        "greps_since_write": 0,
        "edit_targets": {},
        "agents_without_write": 0,
        "last_write_time": time.time(),
        "calls_since_write": 0,
        "last_volume_reset": time.time(),
        "gate_grace_remaining": 0,
        "gate_grace_gate": None,
        "gate_grace_grants": 0,
    }


def save_cache(actor_key, cache):
    path = cache_path(actor_key)
    try:
        with open(path, "w") as f:
            json.dump(cache, f)
    except IOError:
        pass


# --- The single-writer chokepoint (sp-016776e6) ------------------------------
# load_cache/save_cache are a read-modify-write, and Claude Code fires this hook
# once per tool call — so parallel tool calls mean parallel guard processes on
# ONE actor's cache file. Unlocked, each read the same pre-state and the last
# writer won; every other process's mutation was silently dropped.
#
# The lost mutation that cost real money is `fable_escalations`, the counter
# request_escalation reads to decide whether this actor has spent its
# KIPI_FABLE_CAP budget of billed `claude -p --model claude-fable-5` calls. A
# dropped increment let the next process read a stale count, believe it was
# under the cap, and buy another one.
#
# MEASURED over the shipped ledger before the fix (188 rows,
# q-system/output/fable-escalations/, 2026-08-08): 172 rows attempted a real
# call against a cap of 2 per actor. 90 landed in a second that already carried
# another real call, 7 in the worst single second, and 8 distinct seconds carry
# two DIFFERENT triggers — two concurrent guard processes is the only thing that
# produces that. The two colliding CAPPED pairs are 2026-08-04T03:35:31Z and
# 2026-08-08T22:07:13Z, each one `exact-retry` racing one `volume-ceiling`.
#
# WHY A LOCK AND NOT AN ATOMIC WRITE. tmpfile + os.replace was the tempting fix
# and it does not work: it makes the WRITE atomic, not the read-modify-write.
# Both processes still read the same pre-state, so the update is still lost —
# just lost cleanly. The read and the write have to be inside one critical
# section, which is what this is.
#
# LOST UPDATES WERE NOT THE WORST OF IT. save_cache truncates and rewrites in
# place, so two concurrent writers can interleave: running the concurrency suite
# against a no-lock mutant produced a cache whose JSON ended at byte 464 with a
# longer writer's tail still after it. load_cache catches JSONDecodeError and
# returns fresh defaults, so a torn write silently reset every counter this
# guard owns — the ceiling, the retry map and the escalation budget all back to
# zero, with nothing logged and nothing to see afterwards.
#
# WHY THE REGION HOLDS NO SUBPROCESS. This hook is wired at `timeout: 5` on all
# three events fleet-wide, and a hook that overruns its timeout has its exit 2
# DISCARDED (measured 2026-08-03; see the escalation block comment below). So a
# lock held across `_head_commit_epoch`'s git call (capped at 3s) could turn
# contention into a SPENT refusal — trading this defect for a worse one. The git
# probe is therefore hoisted out of the region by main(), leaving only in-memory
# work under the lock: microseconds, not milliseconds.
LOCK_WAIT_SECONDS = 2.0     # unreachable in practice; the region is pure CPU
LOCK_POLL_SECONDS = 0.002


def lock_path(actor_key):
    """A sidecar, never the cache file itself.

    HONEST ABOUT WHY, because the first version of this comment was wrong.
    It claimed save_cache's mode-"w" truncate would drop a lock taken on the
    cache path; a mutant that pointed this function AT the cache file survived
    the concurrency suite 8/8, because truncation keeps the same inode and flock
    follows the inode. So that is not the reason.

    The real reasons are narrower and both structural: the lock must exist
    before the cache does (an actor's first tool call has no cache file yet, and
    load_cache is happy to return defaults for one), and a sidecar cannot be
    invalidated by any future change that REPLACES the cache rather than
    rewriting it in place — an os.replace-based atomic write, the obvious next
    edit here, swaps the inode and would silently orphan every held lock.
    """
    return f"/tmp/claude-guard-{actor_key}.lock"


def _acquire(handle, deadline_s=LOCK_WAIT_SECONDS):
    """Bounded LOCK_EX. True if held, False if the wait ran out."""
    end = time.time() + deadline_s
    while True:
        try:
            fcntl.flock(handle.fileno(), fcntl.LOCK_EX | fcntl.LOCK_NB)
            return True
        except OSError:
            if time.time() >= end:
                return False
            time.sleep(LOCK_POLL_SECONDS)


# Holds the open lock handles for this process's lifetime. Module-level on
# purpose: a handle kept only in a local could be garbage-collected, and closing
# the fd releases the flock — silently, mid-region, with everything still
# looking correct.
_HELD_LOCKS = []


def hold_actor_lock(actor_key):
    """Take this actor's lock and keep it until the process exits.

    The PreToolUse path cannot use a `with` block: every check ends in
    block()/warn(), both of which sys.exit, and the path already saves the cache
    itself at each of those exits. Rather than re-indent ~110 lines of checks
    (a mechanical edit with plenty of room to hide a defect) the lock is simply
    held for the rest of the invocation and released by the OS on process exit,
    which is the one exit every path shares.

    Fails open on purpose — see locked_cache.
    """
    try:
manifest JSON: PASS

codex
The current files are correctly wired, but the new test’s strongest claims are broader than its assertions. I’m now running mutation probes against the exact predicates to see whether plausible regressions remain green.
exec
/bin/zsh -lc "bash -c '
set -euo pipefail
needle=\"q-system/.q-system/token-guard.py\"
broken="'$(awk -v n="$needle" '"'\\''index("'$0,n){seen++; if(seen>1) next} {print}'"'\\'' .claude/settings.json)
printf \"remaining_guard_occurrences=%s\\n\" \""'$(grep -cF "$needle" <<<"$broken")"
if grep -qF "$needle" <<<"$broken"; then echo "test_predicate=PASS"; else echo "test_predicate=FAIL"; fi
printf "events_still_claimed=3 events_actually_wired=%s'"\\n\" \""'$(grep -cF "$needle" <<<"$broken")"
'"'

bash -c '
set -euo pipefail
ROOT="'$PWD
mutant=$(sed '"'\\''s/only the honest labelling/ALL cleanup and migration steps/'\\'' .claude/rules/token-discipline.md)
enforced_sections_without_executable() {
  local line heading=\"\" enforced=0 satisfied=0 name
  emit() { [ \""'$enforced" -eq 1 ] && [ "$satisfied" -eq 0 ] && [ -n "$heading" ] && printf "%s'"\\\\n\" \""'$heading"; return 0; }
  while IFS= read -r line; do
    case "$line" in
      "#"*" "*) emit; heading="$line"; case "$line" in *"(ENFORCED"*) enforced=1 ;; *) enforced=0 ;; esac; satisfied=0; continue ;;
    esac
    [ "$enforced" -eq 1 ] || continue
    for name in $(printf "%s'"\\\\n\" \""'$line" | grep -oE '"'\\''[A-Za-z0-9_-]+\\.(py|sh)'\\'' || true); do
      [ -n \""'$(find "$ROOT" -name "$name" -not -path '"'\\''*/.git/*'\\'' -print -quit)\" ] && satisfied=1
    done
  done <<<\""'$mutant"
  emit
}
orphans=$(enforced_sections_without_executable)
printf "mutated_scope=%s'"\\n\" \""'$(grep -F "ALL cleanup" <<<"$mutant" | sed '"'\\''s/"'^ *//;'"'\\'')\"
printf \"orphans=%s\\n\" \""'${orphans:-<none>}"
[ -z "$orphans" ] && echo "test_predicate=PASS" || echo "test_predicate=FAIL"
'"'" in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-232
 succeeded in 0ms:
bash: line 4: cannot create temp file for here document: Operation not permitted
remaining_guard_occurrences=
bash: line 5: cannot create temp file for here document: Operation not permitted
test_predicate=FAIL
bash: line 6: cannot create temp file for here document: Operation not permitted
events_still_claimed=3 events_actually_wired=
bash: line 6: cannot create temp file for here document: Operation not permitted
bash: line 19: cannot create temp file for here document: Operation not permitted
mutated_scope=
orphans=<none>
test_predicate=PASS

exec
/bin/zsh -lc "bash -c '
set -euo pipefail
needle=\"q-system/.q-system/token-guard.py\"
printf \"remaining_guard_occurrences=\"
awk -v n=\""'$needle" '"'\\''index("'$0,n){seen++; if(seen>1) next} {print}'"'\\'' .claude/settings.json | grep -cF \""'$needle"
if awk -v n="$needle" '"'\\''index("'$0,n){seen++; if(seen>1) next} {print}'"'\\'' .claude/settings.json | grep -qF \""'$needle"; then echo "test_predicate=PASS"; else echo "test_predicate=FAIL"; fi
echo "events_still_claimed=3 events_actually_wired=1"
'"'

bash -c '
set -euo pipefail
ROOT="'$PWD
enforced_sections_without_executable() {
  local line heading="" enforced=0 satisfied=0 name
  emit() { [ "$enforced" -eq 1 ] && [ "$satisfied" -eq 0 ] && [ -n "$heading" ] && printf "%s'"\\\\n\" \""'$heading"; return 0; }
  while IFS= read -r line; do
    case "$line" in
      "#"*" "*) emit; heading="$line"; case "$line" in *"(ENFORCED"*) enforced=1 ;; *) enforced=0 ;; esac; satisfied=0; continue ;;
    esac
    [ "$enforced" -eq 1 ] || continue
    for name in $(printf "%s'"\\\\n\" \""'$line" | grep -oE '"'\\''[A-Za-z0-9_-]+\\.(py|sh)'\\'' || true); do
      [ -n \""'$(find "$ROOT" -name "$name" -not -path '"'\\''*/.git/*'\\'' -print -quit)\" ] && satisfied=1
    done
  done
  emit
}
printf \"mutated_scope=\"
sed '\\''s/only the honest labelling/ALL cleanup and migration steps/'\\'' .claude/rules/token-discipline.md | grep -F \"ALL cleanup\" | sed '\\''s/"'^ *//;'"'\\''
printf \"orphans=\"
orphans="'$(sed '"'\\''s/only the honest labelling/ALL cleanup and migration steps/'\\'' .claude/rules/token-discipline.md | enforced_sections_without_executable)
printf \"%s\\n\" \""'${orphans:-<none>}"
[ -z "$orphans" ] && echo "test_predicate=PASS" || echo "test_predicate=FAIL"
'"'" in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-232
 succeeded in 0ms:
remaining_guard_occurrences=1
test_predicate=PASS
events_still_claimed=3 events_actually_wired=1
mutated_scope=- **Scope of (ENFORCED) above:** ALL cleanup and migration steps, pinned by `test-token-discipline-rule-wired.sh`. The two passes themselves are not gated -- deciding a task IS a rename is judgment, and "pass 1 done" is your prose, so no hook can see either. Not optional; also not checked.
orphans=<none>
test_predicate=PASS

exec
/bin/zsh -lc "for f in .claude/settings.json settings-template.json; do printf '%s guard_occurrences=' \""'$f"; grep -cF '"'q-system/.q-system/token-guard.py' \""'$f"; done; git diff --check 197dcfc7..HEAD; echo diff_check=PASS; python3 - <<'"'PY'
import json
p='q-system/.q-system/capability-manifest.json'
d=json.load(open(p))
e=[x for x in d['expected_tests'] if x['path'].endswith('test-token-discipline-rule-wired.sh')]
print('manifest_entry=', e)
PY
bash -n q-system/.q-system/scripts/test/test-token-discipline-rule-wired.sh && echo bash_syntax=PASS
" in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-232
 succeeded in 0ms:
.claude/settings.json guard_occurrences=3
settings-template.json guard_occurrences=3
diff_check=PASS
bash_syntax=PASS

exec
/bin/zsh -lc "nl -ba q-system/.q-system/scripts/test/test-token-discipline-rule-wired.sh | sed -n '42,116p'; python3 -c 'import json; p=\"q-system/.q-system/capability-manifest.json\"; d=json.load(open(p)); print([x for x in d[\"expected_tests\"] if x[\"path\"].endswith(\"test-token-discipline-rule-wired.sh\")])'" in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-232
 succeeded in 0ms:
    42	  || fail "token-discipline.md claims ENFORCED but never names $GUARD_NAME"
    43	
    44	# --- (b) no ENFORCED heading without a REAL executable in its section --------
    45	# Walks the file, tracks the current heading, and for every heading carrying
    46	# "(ENFORCED" collects the *.py / *.sh basenames appearing in that section's body.
    47	# A section passes only if at least one of those names resolves to a file that
    48	# actually exists in this repo. This is the check the two prompt-only subsections
    49	# failed: they named nothing at all.
    50	enforced_sections_without_executable() {
    51	  local rule_file="$1" line heading="" enforced=0 satisfied=0 name
    52	  emit() {
    53	    [ "$enforced" -eq 1 ] && [ "$satisfied" -eq 0 ] && [ -n "$heading" ] \
    54	      && printf '%s\n' "$heading"
    55	    return 0
    56	  }
    57	  while IFS= read -r line; do
    58	    case "$line" in
    59	      '#'*' '*)
    60	        emit
    61	        heading="$line"
    62	        case "$line" in *'(ENFORCED'*) enforced=1 ;; *) enforced=0 ;; esac
    63	        satisfied=0
    64	        continue
    65	        ;;
    66	    esac
    67	    [ "$enforced" -eq 1 ] || continue
    68	    # Any basename ending .py/.sh on this line; satisfied once one of them is real.
    69	    for name in $(printf '%s\n' "$line" | grep -oE '[A-Za-z0-9_-]+\.(py|sh)' || true); do
    70	      if [ -n "$(find "$ROOT" -name "$name" -not -path '*/.git/*' -print -quit)" ]; then
    71	        satisfied=1
    72	      fi
    73	    done
    74	  done < "$rule_file"
    75	  emit
    76	}
    77	
    78	orphans="$(enforced_sections_without_executable "$RULE")"
    79	[ -z "$orphans" ] || fail "ENFORCED heading(s) in token-discipline.md name no executable:
    80	$orphans
    81	Either wire one and name it, or relabel the section ADVISORY."
    82	
    83	# --- negative self-tests: both checks must go red on a mutated copy ----------
    84	TMPDIR_TEST="$(mktemp -d)"
    85	trap 'rm -rf "$TMPDIR_TEST"' EXIT
    86	
    87	sed "s/$GUARD_NAME//g" "$RULE" > "$TMPDIR_TEST/stripped.md"
    88	if rule_names_guard "$TMPDIR_TEST/stripped.md"; then
    89	  fail "negative self-test: rule_names_guard passed on a copy with the name stripped"
    90	fi
    91	
    92	cp "$RULE" "$TMPDIR_TEST/orphaned.md"
    93	printf '\n## Synthetic Claim (ENFORCED)\n\nNo executable is named in this section.\n' \
    94	  >> "$TMPDIR_TEST/orphaned.md"
    95	if [ -z "$(enforced_sections_without_executable "$TMPDIR_TEST/orphaned.md")" ]; then
    96	  fail "negative self-test: the ENFORCED-heading check passed on a copy carrying a scriptless ENFORCED section"
    97	fi
    98	
    99	# ...and a plausible-but-nonexistent filename must NOT satisfy it, or the check
   100	# degrades into "name any string ending in .py".
   101	cp "$RULE" "$TMPDIR_TEST/hallucinated.md"
   102	printf '\n## Synthetic Claim (ENFORCED)\n\nEnforced by `no-such-guard-ask139.py`.\n' \
   103	  >> "$TMPDIR_TEST/hallucinated.md"
   104	if [ -z "$(enforced_sections_without_executable "$TMPDIR_TEST/hallucinated.md")" ]; then
   105	  fail "negative self-test: a nonexistent script name satisfied the ENFORCED-heading check"
   106	fi
   107	
   108	# --- the named executable is real and wired ----------------------------------
   109	# Not asserting the exec bit: both settings entries invoke it as `python3 <path>`,
   110	# so chmod +x is not what makes this wiring live. Asserting -x here would be a
   111	# check that passes or fails for a reason nobody depends on.
   112	[ -f "$ROOT/$GUARD_REL" ] || fail "$GUARD_NAME is named by the rule but missing at $GUARD_REL"
   113	
   114	grep -qF "$GUARD_REL" "$ROOT/.claude/settings.json" \
   115	  || fail ".claude/settings.json does not wire $GUARD_NAME"
   116	grep -qF "$GUARD_REL" "$ROOT/settings-template.json" \
[{'path': 'q-system/.q-system/scripts/test/test-token-discipline-rule-wired.sh', 'runner': 'bash'}]

codex
## Findings

### MINOR

The wiring assertion only checks one substring occurrence, so losing two of the three claimed hook registrations still passes.

- File: `q-system/.q-system/scripts/test/test-token-discipline-rule-wired.sh:114`
- Reproducer:

```bash
bash -c '
needle="q-system/.q-system/token-guard.py"
printf "remaining_guard_occurrences="
awk -v n="$needle" '"'"'index($0,n){seen++; if(seen>1) next} {print}'"'"' .claude/settings.json | grep -cF "$needle"
if awk -v n="$needle" '"'"'index($0,n){seen++; if(seen>1) next} {print}'"'"' .claude/settings.json | grep -qF "$needle"; then
  echo "test_predicate=PASS"
fi
echo "events_still_claimed=3 events_actually_wired=1"
'
```

- Real output:

```text
remaining_guard_occurrences=1
test_predicate=PASS
events_still_claimed=3 events_actually_wired=1
```

This is minor because the current producer files contain all three registrations. The reproducer uses a constructed regression.

### MINOR

The section check accepts any existing script basename, so it does not pin the claimed “labelling only” scope and permits the rule to drift back to a false enforcement claim.

- File: `q-system/.q-system/scripts/test/test-token-discipline-rule-wired.sh:69`
- Reproducer: run the checker’s `enforced_sections_without_executable` function against this mutation:

```bash
sed 's/only the honest labelling/ALL cleanup and migration steps/' \
  .claude/rules/token-discipline.md
```

- Real output from the checker predicate:

```text
mutated_scope=- **Scope of (ENFORCED) above:** ALL cleanup and migration steps, pinned by `test-token-discipline-rule-wired.sh`. The two passes themselves are not gated -- deciding a task IS a rename is judgment, and "pass 1 done" is your prose, so no hook can see either. Not optional; also not checked.
orphans=<none>
test_predicate=PASS
```

This is minor because the checked-in prose is currently honest. The weakness affects future regression detection.

## What is sound

- Both settings files currently contain three `token-guard.py` registrations.
- The manifest contains the new test with the correct `bash` runner.
- The manifest parses as valid JSON.
- The test script passes `bash -n`.
- `git diff --check` passes.
- Missing and nonexistent executable mutations are explicitly covered.
- The full test could not execute because this review sandbox denies `mktemp` writes to `$TMPDIR`; it stopped with `Operation not permitted`.

## Verdict

Fix first: assert the exact three event registrations in each settings file, rather than accepting one path occurrence.

FINDINGS:
minor|The wiring assertion passes after two of the three claimed hook registrations are removed.|q-system/.q-system/scripts/test/test-token-discipline-rule-wired.sh:114
minor|The section check accepts any existing script basename and therefore does not pin the claimed labelling-only scope.|q-system/.q-system/scripts/test/test-token-discipline-rule-wired.sh:69
END FINDINGS
tokens used
63,136
## Findings

### MINOR

The wiring assertion only checks one substring occurrence, so losing two of the three claimed hook registrations still passes.

- File: `q-system/.q-system/scripts/test/test-token-discipline-rule-wired.sh:114`
- Reproducer:

```bash
bash -c '
needle="q-system/.q-system/token-guard.py"
printf "remaining_guard_occurrences="
awk -v n="$needle" '"'"'index($0,n){seen++; if(seen>1) next} {print}'"'"' .claude/settings.json | grep -cF "$needle"
if awk -v n="$needle" '"'"'index($0,n){seen++; if(seen>1) next} {print}'"'"' .claude/settings.json | grep -qF "$needle"; then
  echo "test_predicate=PASS"
fi
echo "events_still_claimed=3 events_actually_wired=1"
'
```

- Real output:

```text
remaining_guard_occurrences=1
test_predicate=PASS
events_still_claimed=3 events_actually_wired=1
```

This is minor because the current producer files contain all three registrations. The reproducer uses a constructed regression.

### MINOR

The section check accepts any existing script basename, so it does not pin the claimed “labelling only” scope and permits the rule to drift back to a false enforcement claim.

- File: `q-system/.q-system/scripts/test/test-token-discipline-rule-wired.sh:69`
- Reproducer: run the checker’s `enforced_sections_without_executable` function against this mutation:

```bash
sed 's/only the honest labelling/ALL cleanup and migration steps/' \
  .claude/rules/token-discipline.md
```

- Real output from the checker predicate:

```text
mutated_scope=- **Scope of (ENFORCED) above:** ALL cleanup and migration steps, pinned by `test-token-discipline-rule-wired.sh`. The two passes themselves are not gated -- deciding a task IS a rename is judgment, and "pass 1 done" is your prose, so no hook can see either. Not optional; also not checked.
orphans=<none>
test_predicate=PASS
```

This is minor because the checked-in prose is currently honest. The weakness affects future regression detection.

## What is sound

- Both settings files currently contain three `token-guard.py` registrations.
- The manifest contains the new test with the correct `bash` runner.
- The manifest parses as valid JSON.
- The test script passes `bash -n`.
- `git diff --check` passes.
- Missing and nonexistent executable mutations are explicitly covered.
- The full test could not execute because this review sandbox denies `mktemp` writes to `$TMPDIR`; it stopped with `Operation not permitted`.

## Verdict

Fix first: assert the exact three event registrations in each settings file, rather than accepting one path occurrence.

FINDINGS:
minor|The wiring assertion passes after two of the three claimed hook registrations are removed.|q-system/.q-system/scripts/test/test-token-discipline-rule-wired.sh:114
minor|The section check accepts any existing script basename and therefore does not pin the claimed labelling-only scope.|q-system/.q-system/scripts/test/test-token-discipline-rule-wired.sh:69
END FINDINGS

PR #267, opened 2026-08-29
The alert path had no consumer: ship the triage lane (ASK-1127)

assafkip:
## What was actually wrong

`alert-to-linear.py` files a ticket. TWO independent readers then refuse it, on purpose:

| reader | predicate | result |
|---|---|---|
| `linear-dor-drafter.py` | `is_alert_ticket()` | `selection_mode()` -> `None` |
| `linear-worker.sh:568` | `is_fleet_alert()` | `ready()` -> `False` |

Both were added together (ASK-839) with sound reasoning. What was never built is
anything that moves a ticket **out** of that bucket.

Measured on the live board 2026-08-29, using the shipped predicates rather than
reimplementations: **all 151 open alert tickets** return `selection_mode is None`,
while **74 non-alert issues** return `"draft"` — a positive control, so the predicate
is not simply refusing everything.

## The contract violation

`terminal-states.json` registers the exit as `terminal: true` with the rationale
*"the consumer of an alert is the operator who reads it."* The same file's own doc
block forbids exactly that: *"A human is never a valid consumer."* And that operator
does not exist — the ASK team has two seats, Codex and the founder, and `owner:sana`
is a routing label, not a person. **The declared consumer is the empty set**, so the
effective behaviour is silence. Today that silence covered a RED main and a disk at 98%.

The row passed validation because `terminal: true` rows are not required to name a
live consumer; the rationale text was never machine-checked.

## Why the obvious fix is inert

Having the filer emit its own DoR does **nothing**: `linear-worker.sh` evaluates
`is_fleet_alert` on line 568, *before* line 571 ever reads the description for a DoR.
`test_a_dor_alone_does_not_unblock_it` pins that. Making it work would mean deleting
that exclusion, which reverts a measured decision and admits every unscoped
notification to the executable queue.

## What this ships

`linear-alert-triage.py` — the missing consumer. Promotion strips the fingerprint
comment, drops `needs-triage`, appends a real DoR. After that **both readers accept
with zero change to either**; that is the design goal, not a coincidence.

- **Dedup unaffected** — verified in `_read_state`/`_write_state`: repeats key on a
  local state file, not on the description.
- **Never silent** — closing comments the reason *before* the state change. No verb
  in the script deletes an issue.
- **Provenance survives** — the fingerprint is preserved under a different comment key
  that neither reader matches.

## Evidence

- 7/7 tests. The worker's `is_fleet_alert` is **extracted from the shipped `.sh`** and
  exec'd, not retyped, so a drift in the real predicate breaks this test.
- Mutation-tested: `strip` as a no-op fails 3; an audit marker colliding with the real
  key fails 2. Control green, `git diff --stat` empty after restore.
- **Load-path proof**: ran the worker's own `ready()` against the live board. The
  kipi-system ready set went **20 -> 27**.

## Triage applied

| | |
|---|---|
| Promoted, now in the ready set | ASK-1117, 1118, 1121, 1124, 1125, 1126, 1127 |
| Closed as superseded, with reasons | ASK-1120 (-> 1126), ASK-1123 (-> 1124) |
| Already ready, untouched | ASK-1119 |
| Second filer defect, labelled + captured | ASK-1122 |

A triage insight worth flagging: **ASK-1120, ASK-1121 and ASK-1126 are one fix.** A
per-round *owned* worktree removes the shared mutable resource (no TOCTOU), fixes the
stale ref, and bounds the 6.4G/206-tree growth. That decision is made in ASK-1126.

## Deliberately not in this PR

The `terminal-states.json` row rewrite. `test-terminal-states.sh` is being edited by
another agent right now and main is RED on it — editing the JSON it validates would
deepen a red main. Carried as the first remaining item on ASK-1127.

### DIFF
--- a/plugins/kipi-core/.claude-plugin/plugin.json
+++ b/plugins/kipi-core/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "kipi-core",
-  "version": "1.7.24",
+  "version": "1.7.25",
   "description": "Core founder OS \u2014 voice, research, AUDHD, LinkedIn/brand, RCA skills + kipi-mcp server",
   "author": {
     "name": "Assaf Kipnis",
--- a/plugins/kipi-core/.claude-plugin/plugin.json
+++ b/plugins/kipi-core/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "kipi-core",
-  "version": "1.7.25",
+  "version": "1.7.26",
   "description": "Core founder OS \u2014 voice, research, AUDHD, LinkedIn/brand, RCA skills + kipi-mcp server",
   "author": {
     "name": "Assaf Kipnis",
--- a/plugins/kipi-core/voicekit/corpus.py
+++ b/plugins/kipi-core/voicekit/corpus.py
@@ -26,7 +26,18 @@ EXEMPLAR_KINDS = ("post", "article-excerpt", "article", "comment", "dm", "email"
 # replaced their excerpts in the corpus. Deliberately absent from
 # selector.ELIGIBLE_KINDS -- a 1,400-word article is reference material for the
 # authorship region, not a slot exemplar a 280-char draft should imitate.
-EXEMPLAR_CHANNELS = ("linkedin", "x", "substack", "medium", "any")
+# "reddit" added 2026-08-29. A real founder-written Reddit comment was banked as
+# an exemplar and the corpus refused it, correctly: the row named a channel this
+# tuple did not know. The convention every `kind: "comment"` row already follows
+# is channel-of-origin, where he actually wrote it, so the two honest fixes were
+# widening this tuple or relabelling the row "any". Relabelling was rejected: it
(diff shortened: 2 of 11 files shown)


assafkip:
Superseded by #268.

Not rebased: cherry-picked. The add/add conflict was not two lineages of the same file. This branch was cut from the LOCAL `main` (`a2b026b0`), which has diverged from `origin/main` by 8 unpushed commits, one of which (`936fbbee`) created `verify.sh` and `test_verify_adversarial.sh` locally. `origin/main` got its own copies via the verify-floor PR.

So #267 was silently proposing to land 8 commits of unrelated work. Merging the two verify.sh versions would have been the wrong move: neither was mine, and `origin/main`'s are the further-developed ones (376 vs 301 lines, 170 vs 132).

#268 cherry-picks only my own commit onto `origin/main`. 3 files, no conflict, 7/7 green.

Separate item for whoever owns it: the local `main` checkout is still diverged from `origin/main` by those 8 commits.

PR #271, opened 2026-08-29
Mutation sweep: a detector for tests that run, pass, and cannot fail

assafkip:
Founder-directed. 17 instances of one class were measured in a single session:
a check that RUNS, PASSES, and is structurally blind to the thing it exists to
catch. This lands the detector and its self-guard. Fixes ship separately.

## What it does

`q-system/.q-system/scripts/mutation-sweep.py`, over the capability manifest's
declared tests (183 with this PR's own entry).

Two-stage probe per (test, subject):

1. **TRIPWIRE** — replace the subject's body with an immediate `exit 97`. The
   control must turn the test red, which is also what proves attribution. A
   test still green here does not exercise the file.
2. **DISARM** — neuter every failure-signalling site at once (nonzero
   `sys.exit` to `sys.exit(0)`, `return False` to `return True`, bash
   `exit 1` to `exit 0`). The subject can no longer report failure. Still
   green means the test cannot observe its subject's verdict.

`--zero-exec-scan` covers what mutation testing structurally cannot see: the
gate runs each python3 entry as `python3 <file>`, so a pytest-style module with
no entry point imports, defines its tests, and exits 0. Nothing ran, and it
reads as a pass.

`--report-only` re-derives the report from the append-only ledger.

## Guarding the harness itself

Two of the 17 instances WERE mutation harnesses that reported perfect survival —
one because an import error killed every case, one because it ran no tests at
all. A survival report from a harness that ran nothing is the worst possible
output here, so every verdict is proved:

- mutant applied confirmed by sha256 **re-read from disk**, not by "the write returned"
- mutant confirmed syntactically valid (`ast.parse` / `bash -n`) so a broken
  mutant cannot score a free kill off any test that merely imports it
- baseline must be green or the test is excluded rather than scored
- after every restore the test is re-run and must be green again
- private `PYTHONPYCACHEPREFIX` per run: a one-char mutant leaves the file the
  same size and the restore lands in the same second, so CPython will serve the
  mutant's cached bytecode for restored source
- restore is a `cp` from an in-process backup, never a git checkout, installed
  on SIGTERM/SIGINT/SIGHUP as well as `finally` (`finally` does not run on a
  signal; the first long run was killed mid-flight)
- `run_contained` is imported from a **copy** of `capability-gate.py`, so tests
  run exactly as the gate runs them and the sweep can never execute its own
  mutant as its runner

`test-mutation-sweep.sh` is the paired check, declared in the manifest so it
runs on every commit. It matches on the four verdicts, not on `$?`.

## The instrument was wrong three times before it was right

Each caught by a guard rather than by review, each recorded in its commit:

- an operator whose trailing `\s*` anchor ate the newline and welded two
  statements into a syntax error — every test that imports the file would have
  killed it for free
- `sys.exit(main())` rewritten to `sys.exit(0))` by a regex that stopped at the
  first paren — 48 sites on `issue_runner.py` would have scored as a clean kill
- disarm anchors too narrow: `exit 1 ;;` in a case arm missed, so
  `repo-preflight.sh` reported 2 of its 4 exit sites. Every miss biases a
  verdict toward KILLED, which is the class itself

A first attempt at the wrapper's negative test PASSED against a "broken"
harness: the mutant applied but was a no-op. Same trap, recorded.

## Reading the number

The per-test SURVIVED count is real but is not the defect count.
`test-capability-gate-reap.sh` survives a disarm of `capability-gate.py`
correctly — it imports `run_contained` and tests process reaping, so the gate's
exit codes are outside its remit. So the report also rolls up per SUBJECT: this
file can no longer report failure, did ANY declared test notice? A subject whose
every confirmed test survived has a failure path nothing observes. That is the
row to fix.

## Placement

Engine, paired check and runbook live under `q-system/.q-system/`, so they ship
to all 23 instances via the fleet updater. The manifest and gate they operate on
are synced there too, and an instance with a `capability-manifest.local.json`
overlay has declared tests nobody else can sweep. Wiring is a runbook
(`q-system/.q-system/mutation-sweep.md`), deliberately **not** a
`.claude/rules/` file: the always-on instruction budget is already red at
512/300 lines and a new rule would deepen a failing gate for no gain.

Posture is ON-DEMAND and ADVISORY. Nothing schedules it, nothing blocks on it.

## Verification

- `capability-gate.py --check-only`: GREEN, engine not flagged inert
- `test-mutation-sweep.sh`: passes in 1.9s; goes RED (rc=1, naming both
  regressed fixtures) against a `make_disarm` that returns the source unmutated
- disarm operators validated across 479 source files: 874 sites, 0
  syntactically invalid mutants
- full-population survival number: sweep in flight, posted here when it lands


assafkip:
Rebuilt on origin/main as a clean branch: this one was cut from a local main carrying 9 unpushed commits from another session, which it would have dragged into the merge. Same commits, cherry-picked. Superseded.

PR #280, opened 2026-08-30
fix(capability-gate): 13 declared tests executed nothing and hid 248 real cases, 3 of them red (ASK-1145)

assafkip:
## What was blind

13 capability-manifest entries declared `runner: python3` on pytest-shaped
modules: bare `def test_` at module level, no `__main__` block. The gate builds
`python3 <file>` for a python3 entry. `python3` on such a module imports it,
binds the function names, and exits 0 having run nothing.

So the gate counted 13 tests, reported them green, and could not see the 248
real cases inside them. **Three of those 248 were failing**, and had been
failing for as long as they existed, in the fact-propagation leak gate on a
public repo.

## The fix that stops it recurring

Flipping 13 declarations is a hand-list, and the 14th file arrives declared
`python3` with nothing to say so. So there are two changes and the second is the
load-bearing one:

- `pytest` becomes a third runner value, built as
  `python3 -m pytest <file> -q -p no:cacheprovider`. The cache flag is not
  cosmetic: a gate run must not leave a `.pytest_cache` in the tree it just
  measured.
- `executes_nothing()` refuses a python3-declared `.py` that defines test
  functions and has no `__main__` entry point.

The detector runs in the execution loop, not at declaration time, because that
is where the file about to be run is already resolved. A declaration-time check
would re-derive the path and could disagree with what actually executes: two
readers of one input, which is the class this repo keeps finding.

Deliberately narrow. It fires only on a `.py` declared `python3` that defines
`def test_` and has no `__main__`. A helper doing its work at import time defines
no test functions and is untouched.

**Mutation-tested rather than trusted.** Planting `runner: python3` back onto one
fragment makes `executes_nothing` return True; the correct `pytest` declaration
returns False. The 13 fragment flips are derived by reading each file, never
hand-listed.

## The three failures it exposed

**1. `test_filter_mirror_matches_the_updater`** died on `StopIteration`, not an
assertion, which is why it read as noise. It hunted for a single LINE in
`kipi-update.sh` carrying both `--exclude=` and `__pycache__`; the updater has
since moved those flags into the `PLUGIN_COPY_EXCLUDES` array. It now reads the
array directly, which is the same one truth `test-kipi-update-plugin-excludes.sh`
already drives both updater consumers off.

My first parser split on the first `)` and parsed empty. Every entry carries a
grep regex and every regex contains parens. It was caught only because the new
code refuses an empty parse instead of comparing two empty sets and reporting
agreement.

With the parser working, the mirror was genuinely short by `.env` and `.env.*`.
That direction does not leak (rsync skips them) but it inflates a baseline a
human justifies by hand, which the test's own docstring calls a defect. So the
gate grew `RSYNC_EXCLUDED_NAMES` / `RSYNC_EXCLUDED_PREFIXES` rather than the
assertion loosening.

**2. `test_no_undeclared_propagation_entrypoint_exists`** named 11 repo-root
scripts. All 11 are genuinely non-propagating and now carry a reason naming the
line `first_match()` actually returns, run rather than read: 6 kipi-update test
harnesses copying between `mktemp` fixtures, 2 test harnesses copying into a tmp
dir, an in-place `os.replace`, a docstring match, and `kipi` matching its own
`install-jobs)` case label.

**3. `test_committed_baseline_file_is_armed_and_self_consistent`** had a stale
`classifier_sha256`. Restamped through the gate's own `--restamp`, which refuses
unless every fingerprint is unchanged. It reported 1 unchanged.

## Measurement

248 tests that the gate previously counted as 13 green now actually run: 79
across the four propagation modules, 169 across the other nine. All pass.

## A note on how the third failure was found

The first run of these tests, from the ordinary working checkout, showed **2**
failures. From a clean tree at a known sha it showed **3**. The dirty tree's
uncommitted state was satisfying
`test_committed_baseline_file_is_armed_and_self_consistent` against content that
was never committed. Same defect class as the PR itself, one layer out.

### DIFF
--- a/q-system/.q-system/capability/expected_tests/plugins__kipi-dsse__scripts__test_computed_receipts.py.json
+++ b/q-system/.q-system/capability/expected_tests/plugins__kipi-dsse__scripts__test_computed_receipts.py.json
@@ -1,5 +1,5 @@
 {
- "path": "plugins/kipi-dsse/scripts/test_computed_receipts.py",
- "runner": "python3",
- "timeout_s": 180
+  "path": "plugins/kipi-dsse/scripts/test_computed_receipts.py",
+  "runner": "pytest",
+  "timeout_s": 180
 }
--- a/q-system/.q-system/capability/expected_tests/plugins__kipi-dsse__scripts__test_scope_hook_fails_closed.py.json
+++ b/q-system/.q-system/capability/expected_tests/plugins__kipi-dsse__scripts__test_scope_hook_fails_closed.py.json
@@ -1,5 +1,5 @@
 {
- "path": "plugins/kipi-dsse/scripts/test_scope_hook_fails_closed.py",
- "runner": "python3",
- "timeout_s": 120
+  "path": "plugins/kipi-dsse/scripts/test_scope_hook_fails_closed.py",
+  "runner": "pytest",
+  "timeout_s": 120
 }
--- a/q-system/.q-system/capability/expected_tests/plugins__prd-os__tests__test_virgin_repo_lifecycle.py.json
+++ b/q-system/.q-system/capability/expected_tests/plugins__prd-os__tests__test_virgin_repo_lifecycle.py.json
@@ -1,5 +1,5 @@
 {
- "path": "plugins/prd-os/tests/test_virgin_repo_lifecycle.py",
- "runner": "python3",
- "timeout_s": 120
+  "path": "plugins/prd-os/tests/test_virgin_repo_lifecycle.py",
+  "runner": "pytest",
+  "timeout_s": 120
 }
--- a/q-system/.q-system/capability/expected_tests/q-system__.q-system__scripts__test__test-propagation-entrypoints.py.json
+++ b/q-system/.q-system/capability/expected_tests/q-system__.q-system__scripts__test__test-propagation-entrypoints.py.json
@@ -1,4 +1,4 @@
 {
- "path": "q-system/.q-system/scripts/test/test-propagation-entrypoints.py",
- "runner": "python3"
+  "path": "q-system/.q-system/scripts/test/test-propagation-entrypoints.py",
+  "runner": "pytest"
 }
--- a/q-system/.q-system/capability/expected_tests/q-system__.q-system__scripts__test__test-propagation-leak-baseline.py.json
+++ b/q-system/.q-system/capability/expected_tests/q-system__.q-system__scripts__test__test-propagation-leak-baseline.py.json
@@ -1,4 +1,4 @@
 {
- "path": "q-system/.q-system/scripts/test/test-propagation-leak-baseline.py",
- "runner": "python3"
+  "path": "q-system/.q-system/scripts/test/test-propagation-leak-baseline.py",
+  "runner": "pytest"
 }
--- a/q-system/.q-system/capability/expected_tests/q-system__.q-system__scripts__test__test-propagation-leak-gate.py.json
+++ b/q-system/.q-system/capability/expected_tests/q-system__.q-system__scripts__test__test-propagation-leak-gate.py.json
@@ -1,4 +1,4 @@
 {
- "path": "q-system/.q-system/scripts/test/test-propagation-leak-gate.py",
- "runner": "python3"
+  "path": "q-system/.q-system/scripts/test/test-propagation-leak-gate.py",
+  "runner": "pytest"
 }
--- a/q-system/.q-system/capability/expected_tests/q-system__.q-system__scripts__test__test-propagation-leak-sources.py.json
+++ b/q-system/.q-system/capability/expected_tests/q-system__.q-system__scripts__test__test-propagation-leak-sources.py.json
@@ -1,4 +1,4 @@
 {
- "path": "q-system/.q-system/scripts/test/test-propagation-leak-sources.py",
- "runner": "python3"
+  "path": "q-system/.q-system/scripts/test/test-propagation-leak-sources.py",
+  "runner": "pytest"
 }
--- a/q-system/.q-system/capability/expected_tests/q-system__.q-system__scripts__test_dispatch_reachability.py.json
+++ b/q-system/.q-system/capability/expected_tests/q-system__.q-system__scripts__test_dispatch_reachability.py.json
@@ -1,4 +1,4 @@
 {
- "path": "q-system/.q-system/scripts/test_dispatch_reachability.py",
- "runner": "python3"
+  "path": "q-system/.q-system/scripts/test_dispatch_reachability.py",
+  "runner": "pytest"
 }
--- a/q-system/.q-system/capability/expected_tests/q-system__.q-system__scripts__test_memory_autocapture.py.json
+++ b/q-system/.q-system/capability/expected_tests/q-system__.q-system__scripts__test_memory_autocapture.py.json
@@ -1,4 +1,4 @@
 {
- "path": "q-system/.q-system/scripts/test_memory_autocapture.py",
- "runner": "python3"
+  "path": "q-system/.q-system/scripts/test_memory_autocapture.py",
+  "runner": "pytest"
 }
--- a/q-system/.q-system/capability/expected_tests/q-system__.q-system__scripts__test_memory_outcomes.py.json
+++ b/q-system/.q-system/capability/expected_tests/q-system__.q-system__scripts__test_memory_outcomes.py.json
@@ -1,4 +1,4 @@
 {
- "path": "q-system/.q-system/scripts/test_memory_outcomes.py",
- "runner": "python3"
+  "path": "q-system/.q-system/scripts/test_memory_outcomes.py",
+  "runner": "pytest"
 }
--- a/q-system/.q-system/capability/expected_tests/q-system__.q-system__scripts__test_memory_reflect.py.json
+++ b/q-system/.q-system/capability/expected_tests/q-system__.q-system__scripts__test_memory_reflect.py.json
@@ -1,4 +1,4 @@
 {
- "path": "q-system/.q-system/scripts/test_memory_reflect.py",
- "runner": "python3"
+  "path": "q-system/.q-system/scripts/test_memory_reflect.py",
+  "runner": "pytest"
 }
--- a/q-system/.q-system/capability/expected_tests/q-system__.q-system__scripts__test_memory_scores_surface.py.json
+++ b/q-system/.q-system/capability/expected_tests/q-system__.q-system__scripts__test_memory_scores_surface.py.json
@@ -1,4 +1,4 @@
 {
- "path": "q-system/.q-system/scripts/test_memory_scores_surface.py",
- "runner": "python3"
+  "path": "q-system/.q-system/scripts/test_memory_scores_surface.py",
+  "runner": "pytest"
 }
--- a/q-system/.q-system/capability/expected_tests/q-system__.q-system__scripts__test_miyo_kb_hooks.py.json
+++ b/q-system/.q-system/capability/expected_tests/q-system__.q-system__scripts__test_miyo_kb_hooks.py.json
@@ -1,4 +1,4 @@
 {
- "path": "q-system/.q-system/scripts/test_miyo_kb_hooks.py",
- "runner": "python3"
+  "path": "q-system/.q-system/scripts/test_miyo_kb_hooks.py",
+  "runner": "pytest"
 }
--- a/q-system/.q-system/scripts/capability-gate.py
+++ b/q-system/.q-system/scripts/capability-gate.py
@@ -192,8 +192,9 @@ def validate_test_entry(entry, seen, errors, exempt_prefixes=()):
         msg = declaration_scope_error(p, exempt_prefixes)
         if msg:
             errors.append(msg)
-    if entry.get("runner") not in ("python3", "bash"):
-        errors.append(f"expected_tests entry needs runner python3|bash: {p}")
+    if entry.get("runner") not in RUNNERS:
+        errors.append(
+            f"expected_tests entry needs runner {'|'.join(RUNNERS)}: {p}")
     if p in seen:
         errors.append(f"duplicate expected_tests path: {p}")
     seen.add(p)
@@ -201,6 +202,52 @@ def validate_test_entry(entry, seen, errors, exempt_prefixes=()):
     if not (isinstance(t, int) and TIMEOUT_MIN_S <= t <= TIMEOUT_MAX_S):
         errors.append(f"timeout_s out of bounds [{TIMEOUT_MIN_S},{TIMEOUT_MAX_S}]: {p}")
     validate_quarantine(entry, errors)
+
+
+# A declared test's runner has to be the one that actually EXECUTES it.
+#
+# ASK-1145, measured 2026-08-29. 13 manifest entries declared `runner: python3`
+# on pytest-shaped modules -- bare `def test_` at module level, no `__main__`
+# block. `python3 <file>` imports such a module, binds the function names, and
+# exits 0 having run nothing. The gate counted 13 tests, reported them green,
+# and was structurally blind to the 248 real cases inside them. Three of those
+# were FAILING, invisibly, in the fact-propagation leak gate on a public repo.
+#
+# `pytest` is added as a third runner rather than patching a `__main__` block
+# into 13 files, because the 14th file would arrive without one and nothing
+# would say so. The detector below is what makes that true: a python3-declared
+# module that is pytest-shaped is REFUSED at declaration time, so the mistake
+# cannot be made silently again.
+RUNNERS = ("python3", "bash", "pytest")
+
+# Module-level `def test_...`. Indented defs are methods on a unittest class,
+# which `python3 <file>` still cannot run without a `__main__` block, so they
+# count the same way.
+_BARE_TEST_DEF = re.compile(r"^\s*def test_\w*\s*\(", re.M)
+_HAS_MAIN = re.compile(r"^if __name__\s*==\s*[\"']__main__[\"']\s*:", re.M)
+
+
+def executes_nothing(entry, full):
+    """Refuse a python3-declared test that `python3 <file>` cannot execute.
+
+    The check is deliberately narrow: it fires only on a `.py` file declared
+    `python3` that DEFINES test functions and has NO `__main__` entry point.
+    A script that does its work at import time (most of the bash-adjacent
+    python helpers here) defines no `def test_` and is untouched.
+
+    An unreadable file is not a pass and not a failure here -- the missing-file
+    direction is already reported by the manifest diff, so this returns quietly
+    rather than adding a second, differently-worded error for one cause.
+    """
+    if entry.get("runner") != "python3":
+        return False
+    if not str(full).endswith(".py"):
+        return False
+    try:
+        text = full.read_text(encoding="utf-8", errors="replace")
+    except OSError:
+        return False
+    return bool(_BARE_TEST_DEF.search(text)) and not _HAS_MAIN.search(text)
 
 
 def validate_data_entry(entry, errors):
@@ -634,7 +681,23 @@ def run_tests(root, manifest, mode, errors, notes):
         full = root / path
         if not full.is_file():
             continue  # already reported by the diff
-        cmd = ["python3", str(full)] if entry["runner"] == "python3" else ["bash", str(full)]
+        if executes_nothing(entry, full):
+            errors.append(
+                f"zero-execution test: {path} defines test functions but has "
+                f"no __main__ block, so `python3 {path}` binds them and exits "
+                f"0 without running one. Declare runner pytest, or add a "
+                f"__main__ entry point (ASK-1145).")
+            continue
+        runner = entry["runner"]
+        if runner == "pytest":
+            # `-p no:cacheprovider` so a gate run leaves no .pytest_cache in the
+            # tree it just measured; `-q` keeps the failure tail readable inside
+            # the 20-line cap below.
+            cmd = ["python3", "-m", "pytest", str(full), "-q", "-p", "no:cacheprovider"]
+        elif runner == "python3":
+            cmd = ["python3", str(full)]
+        else:
+            cmd = ["bash", str(full)]
         timeout = entry.get("timeout_s", DEFAULT_TIMEOUT_S)
         r = run_contained(cmd, root, env, timeout)
         if r.timed_out:
--- a/q-system/.q-system/scripts/propagation-leak-gate.py
+++ b/q-system/.q-system/scripts/propagation-leak-gate.py
@@ -214,6 +214,12 @@ ARCHIVE_EXCLUDED_SUBTREES = (
 RSYNC_EXCLUDED_ROOT_DIRS = (".git",)
 RSYNC_EXCLUDED_DIRS = ("__pycache__", ".venv", ".pytest_cache")
 RSYNC_EXCLUDED_SUFFIXES = (".pyc",)
+# Exact filenames and name PREFIXES, which a suffix tuple cannot express.
+# `.env` and `.env.*` are excluded by kipi-update.sh's PLUGIN_COPY_EXCLUDES and
+# were missing here (ASK-1145). rsync never copied them, so this was not a leak
+# -- it inflated the baseline with files that cannot propagate.
+RSYNC_EXCLUDED_NAMES = (".env",)
+RSYNC_EXCLUDED_PREFIXES = (".env.",)
 
 # Suffixes that cannot carry a `label: value` record, so an undecodable file
 # with one of them is an asset rather than an unscanned source. This is a DENY
@@ -328,6 +334,10 @@ def _rsync_filter_reason(entry, at_transfer_root: bool) -> str | None:
         return "excluded-by-rsync-filter"
     if is_directory and entry.name in RSYNC_EXCLUDED_DIRS:
         return "excluded-by-rsync-filter"
+    if entry.name in RSYNC_EXCLUDED_NAMES:
+        return True
+    if entry.name.startswith(RSYNC_EXCLUDED_PREFIXES):
+        return True
     if entry.name.endswith(RSYNC_EXCLUDED_SUFFIXES):
         return "excluded-by-rsync-filter"
     return None
--- a/q-system/.q-system/scripts/test/test-propagation-entrypoints.py
+++ b/q-system/.q-system/scripts/test/test-propagation-entrypoints.py
@@ -90,6 +90,37 @@ EXEMPT = {
     "and destination are both synthetic fixtures under mktemp -d",
     "test-kipi-update-preserve-scan.sh": "a test harness for the same scanner; "
     "its copies are inside a throwaway temp fixture, not a real instance",
+    # --- ASK-1145 -------------------------------------------------------
+    # These eleven were always undeclared. Nobody added them since: this test
+    # was declared `runner: python3` on a pytest module, so the capability gate
+    # imported it and ran none of its cases. It has been red for as long as it
+    # has existed. Each reason below names the line first_match() actually
+    # returns, not an impression formed by reading the file.
+    "kipi": "false positive: the match is the `install-jobs)` case label in the "
+    "CLI dispatcher, which copies nothing and shells out to the real updater",
+    "kipi-update-deletion-guard.py": "read-only guard; the match is prose in its "
+    "module docstring describing what the updater does, same class as "
+    "kipi-update-preserve-scan.py",
+    "kipi-update-gitignore-block.py": "the match is os.replace(tmp, path), an "
+    "atomic in-place rewrite of one file in the repo it runs in, never an "
+    "outward copy into an instance",
+    "test-kipi-update-bash32-empty-array.sh": "test harness; its cp -R source "
+    "and destination are both synthetic fixtures under mktemp -d",
+    "test-kipi-update-cache-exclusion.sh": "test harness; its cp -R source and "
+    "destination are both synthetic fixtures under mktemp -d",
+    "test-kipi-update-config-commit-unwind.sh": "test harness; it copies the "
+    "updater into a throwaway skeleton fixture, never into an instance",
+    "test-kipi-update-dataloss-guards.sh": "test harness; its cp -R source and "
+    "destination are both synthetic fixtures under mktemp -d",
+    "test-kipi-update-dirty-guard-scope.sh": "test harness; its cp -R source "
+    "and destination are both synthetic fixtures under mktemp -d",
+    "test-kipi-update-restore-recovers.sh": "test harness; the copy reads a "
+    "scratch file out of its own temp fixture to assert restore worked",
+    "test_destructive_op_deny_anchor.py": "test harness; shutil.copy puts the "
+    "hook into a tmp dir so the copy can be patched and driven, which is "
+    "exactly how it stays read-only on the real one",
+    "test_fleet_unblock.py": "test harness; it copies one auditor into a "
+    "synthetic skeleton fixture built under tmp_path",
 }
 
 
--- a/q-system/.q-system/scripts/test/test-propagation-leak-sources.py
+++ b/q-system/.q-system/scripts/test/test-propagation-leak-sources.py
@@ -348,19 +348,30 @@ def test_filter_mirror_matches_the_updater():
     """
     gate = load_gate()
     updater = (REPO_ROOT / "kipi-update.sh").read_text(encoding="utf-8")
-    plugin_rsync = next(
-        line
-        for line in updater.splitlines()
-        if "--exclude=" in line and "__pycache__" in line
-    )
-    block = plugin_rsync + "".join(
-        updater.splitlines()[updater.splitlines().index(plugin_rsync) + 1:][:2]
-    )
-    declared = set(re.findall(r'--exclude="([^"]+)"', block))
+
+    # PLUGIN_COPY_EXCLUDES is the one truth both updater consumers read
+    # (plugin_copy_rsync_flags and plugin_copy_filter_regex), and
+    # test-kipi-update-plugin-excludes.sh already proves those two agree. Read
+    # it directly rather than a rendering of it: the previous version hunted
+    # for an inline `--exclude=` LINE, which stopped existing when the flags
+    # moved into this array, and the test died on StopIteration instead of
+    # reporting drift.
+    body = updater.split("PLUGIN_COPY_EXCLUDES=(", 1)
+    assert len(body) == 2, "kipi-update.sh no longer declares PLUGIN_COPY_EXCLUDES"
+    # Split on the array's CLOSING paren at line start, not the first `)`:
+    # the entries carry grep regexes and every one of them contains parens.
+    # The first cut split on `)` and parsed empty -- caught only because the
+    # assertion below refuses an empty parse instead of comparing two empty
+    # sets and reporting agreement.
+    entries = re.findall(r'"([^"]+)::[^"]*"', body[1].split("\n)", 1)[0])
+    assert entries, "PLUGIN_COPY_EXCLUDES parsed empty; the mirror check is blind"
+    declared = set(entries)
 
     mirrored = {f"{name}/" for name in gate.RSYNC_EXCLUDED_DIRS}
     mirrored |= {f"/{name}/" for name in gate.RSYNC_EXCLUDED_ROOT_DIRS}
     mirrored |= {f"*{suffix}" for suffix in gate.RSYNC_EXCLUDED_SUFFIXES}
+    mirrored |= set(gate.RSYNC_EXCLUDED_NAMES)
+    mirrored |= {f"{prefix}*" for prefix in gate.RSYNC_EXCLUDED_PREFIXES}
 
     assert declared == mirrored, (
         f"gate mirror {sorted(mirrored)} != kipi-update.sh {sorted(declared)}"
--- a/q-system/.q-system/state/propagation-leak-baseline.json
+++ b/q-system/.q-system/state/propagation-leak-baseline.json
@@ -24,7 +24,7 @@
     "source_identity",
     "sourced_interaction"
   ],
-  "classifier_sha256": "b0cc8b443aebdfde02e43b3d5fa86e79bb177696a08b92b708cc08f2928d08a3",
+  "classifier_sha256": "7a118485ba72b63050a6fce98b7c05efbfb568633a5e19685931e255d27b0175",
   "entries": [
     {
       "count": 1,


assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-280-20260829-231918.md` (165400 bytes).

```
FINDINGS:
minor|The zero-execution detector is skipped by --check-only, allowing structurally invalid python3 test declarations through unattended update validation|q-system/.q-system/scripts/capability-gate.py:684
minor|The new .env exclusion branches return boolean True instead of the documented string exclusion reason|q-system/.q-system/scripts/propagation-leak-gate.py:338
END FINDINGS
```

--- reviewer output, last 54609 bytes of 165400 (full review at the path above) ---

q-system/.q-system/scripts/test/test-capability-gate-inert-spillover.py:250:ok(f"capability-gate.py --check-only prints the skip and still exits {rc}")
q-system/.q-system/scripts/test/test-canonical-digest-real-values.py:244:    bare run -- `python3 <this file>`, which is exactly what capability-gate.py
q-system/.q-system/scripts/test/test-canonical-digest-real-values.py:252:    invocation. That made it undeclarable in expected_tests (the gate's only
q-system/.q-system/scripts/test/test-kipi-rollback.sh:11:# Scar (sp-3fe1ab9d, 2026-07-28): this suite flaked at 3/30 under capability-gate
q-system/.q-system/scripts/test/test-apply-claude-changes.sh:132:  mkdir -p "$r/q-system/.q-system/capability/expected_tests"
q-system/.q-system/scripts/test/test-apply-claude-changes.sh:136:    > "$r/q-system/.q-system/capability/expected_tests/a__b.py.json"
q-system/.q-system/scripts/test/test-capability-gate-reap.sh:21:GATE="$ROOT/q-system/.q-system/scripts/capability-gate.py"
q-system/.q-system/scripts/test/test-capability-gate-reap.sh:27:[ -f "$GATE" ] || fail "capability-gate.py does not exist at $GATE"
q-system/.q-system/scripts/test/test-capability-gate-reap.sh:124:  || fail "capability-gate.py does not parse"
q-system/.q-system/scripts/test/test-capability-gate-reap.sh:127:echo "PASS: $PASS/$PASS capability-gate reap checks"
q-system/.q-system/scripts/test/test-severity-floor.sh:52:`q-system/.q-system/capability/expected_tests/` (a line number into a shared
q-system/.q-system/scripts/test/test-terminal-states.sh:475:            declared_tests = {e.get("path") for e in json.load(fh).get("expected_tests", [])}
q-system/.q-system/scripts/test/test-terminal-states.sh:662:                              "the capability manifest's expected_tests, so nothing "
q-system/.q-system/scripts/test/test-terminal-states.sh:1104:  "$FIX/unregtest.json" "expected_tests"
q-system/.q-system/scripts/test/test-linear-dor-failure-reporting.py:16:q-system/.q-system/capability/expected_tests/ so capability-gate.py
q-system/.q-system/scripts/test/test-kipi-update-dry-final-state.sh:146:  > "$SKELETON/q-system/.q-system/scripts/capability-gate.py"
q-system/.q-system/scripts/test/test-kipi-update-dry-final-state.sh:207:    > q-system/.q-system/scripts/capability-gate.py
q-system/.q-system/scripts/test/test-apply-claude-changes.sh:132:  mkdir -p "$r/q-system/.q-system/capability/expected_tests"
q-system/.q-system/scripts/test/test-apply-claude-changes.sh:136:    > "$r/q-system/.q-system/capability/expected_tests/a__b.py.json"
q-system/.q-system/scripts/test/test-dispatch-pinned.sh:17:# capability-gate scan root -- same split as test-kipi-update-safety.sh and its
q-system/.q-system/scripts/test/test-capability-gate-inert-spillover.py:16:fixture ledgers, and case 1 additionally runs the real `capability-gate.py
q-system/.q-system/scripts/test/test-capability-gate-inert-spillover.py:51:GATE = ROOT / "q-system/.q-system/scripts/capability-gate.py"
q-system/.q-system/scripts/test/test-capability-gate-inert-spillover.py:68:    fail(f"capability-gate.py does not exist at {GATE}")
q-system/.q-system/scripts/test/test-capability-gate-inert-spillover.py:75:    fail("capability-gate.py has no check_inert_spillover_live(); the declared_inert "
q-system/.q-system/scripts/test/test-capability-gate-inert-spillover.py:95:        "expected_tests": [],
q-system/.q-system/scripts/test/test-capability-gate-inert-spillover.py:163:    fail(f"capability-gate.py --check-only exited 0 on a manifest with a dead "
q-system/.q-system/scripts/test/test-capability-gate-inert-spillover.py:167:ok(f"capability-gate.py --check-only exits {rc} (non-zero) and reports the dead pointer")
q-system/.q-system/scripts/test/test-capability-gate-inert-spillover.py:249:    fail(f"capability-gate.py --check-only printed no skip line with no ledger:\n{out}")
q-system/.q-system/scripts/test/test-capability-gate-inert-spillover.py:250:ok(f"capability-gate.py --check-only prints the skip and still exits {rc}")
q-system/.q-system/scripts/test/test-terminal-states.sh:475:            declared_tests = {e.get("path") for e in json.load(fh).get("expected_tests", [])}
q-system/.q-system/scripts/test/test-terminal-states.sh:662:                              "the capability manifest's expected_tests, so nothing "
q-system/.q-system/scripts/test/test-terminal-states.sh:1104:  "$FIX/unregtest.json" "expected_tests"
q-system/.q-system/scripts/test/test-canonical-digest-real-values.py:244:    bare run -- `python3 <this file>`, which is exactly what capability-gate.py
q-system/.q-system/scripts/test/test-canonical-digest-real-values.py:252:    invocation. That made it undeclarable in expected_tests (the gate's only
q-system/.q-system/scripts/test/test-open-loops.sh:14:# suite's first-ever gated run, 2026-07-23, prd-silent-absence-capability-gate).
q-system/.q-system/scripts/test/test-capability-gate-reap.sh:21:GATE="$ROOT/q-system/.q-system/scripts/capability-gate.py"
q-system/.q-system/scripts/test/test-capability-gate-reap.sh:27:[ -f "$GATE" ] || fail "capability-gate.py does not exist at $GATE"
q-system/.q-system/scripts/test/test-capability-gate-reap.sh:124:  || fail "capability-gate.py does not parse"
q-system/.q-system/scripts/test/test-capability-gate-reap.sh:127:echo "PASS: $PASS/$PASS capability-gate reap checks"
q-system/.q-system/scripts/test/test-kipi-rollback.sh:11:# Scar (sp-3fe1ab9d, 2026-07-28): this suite flaked at 3/30 under capability-gate
#!/bin/bash
set -euo pipefail
trap "" PIPE
# Never let GPG signing or a credential prompt hang the updater. Updater commits
# still run the instance's active hooks and fail closed when a hook rejects them.
export GIT_TERMINAL_PROMPT=0

# kipi-update.sh - Sync latest kipi-system skeleton into all registered instances
# Usage: ./kipi-update.sh [--dry-run]
#
# Uses git archive + rsync (not git subtree pull) for speed and reliability.
# Instance-specific directories (my-project/, canonical/, memory/, output/, bus/)
# are preserved. Everything else syncs from the skeleton.

SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
REGISTRY="$SCRIPT_DIR/instance-registry.json"
SKELETON_REMOTE="https://github.com/assafkip/kipi-system.git"
SKELETON_BRANCH="main"
# Args in any order: --dry-run and/or --only <name>. Without --only there is no
# way to verify a risky change against ONE repo before the other 22, and a
# staged rollout is the only safe way to ship anything with this blast radius.
DRY_RUN=""
ONLY=""
while [ $# -gt 0 ]; do
  case "$1" in
    --dry-run) DRY_RUN="--dry-run"; shift ;;
    --only)
      ONLY="${2:-}"
      if [ -z "$ONLY" ]; then
        echo "ERROR: --only needs an instance name" >&2
        exit 1
      fi
      shift 2
      ;;
    *)
      echo "ERROR: unknown argument: $1" >&2
      echo "Usage: kipi-update.sh [--dry-run] [--only <instance-name>]" >&2
      exit 1
      ;;
  esac
done

if [ ! -f "$REGISTRY" ]; then
  echo "ERROR: instance-registry.json not found at $REGISTRY"
  exit 1
fi

# Instance-owned subtrees: the skeleton never overwrites these, because each
# instance authors its own. This list was duplicated across four sites (the
# q-system rsync, the dry-run preview rsync, the untracked-collision scan, and
# the staging list) and drifted the moment a sixth entry was added -- the sync
# wrote files it then refused to stage, leaving the instance dirty with no
# commit. One list, four consumers.
# The single answer to "what does kipi update NEVER copy into an instance?".
# validate-separation.py's NON_PROPAGATED_PREFIXES mirrors this list and
# test-gate-13b-scope.py parses this array to block drift between them.
#
# `research` added 2026-07-27 (ASK-191): q-system/research/ held four notes
# written BY and ABOUT this instance (Claude Code workflow learnings, an RCA
# resource list, a SkillOpt paper + PDF) and shipped all four to every
# instance in the fleet, which has no use for them. That is a placement bug,
# and the honest fix is to call the directory instance-owned rather than to
# exclude the path from the gate that noticed.
INSTANCE_OWNED_SUBTREES=(
  my-project
  canonical
  memory
  output
  research
  .q-system/data
  .q-system/agent-pipeline/bus
)

# sp-a4a933ad. A dry run's output was TEXTUALLY IDENTICAL to a real one, because
# dry mode really does perform the update -- against a throwaway clone. It printed
# "OK (686 files updated)" and full commit diffs, and only a single banner 700
# lines earlier said it was a preview.
#
# The trap runs both ways, and the second direction is the dangerous one: a
# reader of a dry log reasonably concludes the instance was mutated, and a reader
# of a REAL log can mistake it for a dry one and believe nothing happened. That
# second reading is the one sp-46c73c76 needed a human to catch -- a guard let
# --dry-run genuinely commit into live instances, and the only defence was
# someone noticing. The log gave them nothing to notice with.
#
# say() tags the script's own lines. dry_filter tags the output of commands we
# shell out to (git prints its own "create mode ..." lines and does not know it
# is being previewed).
say() {
  if [ "${DRY_RUN:-}" = "--dry-run" ]; then
    printf 'DRY | %s\n' "$*"
  else
    printf '%s\n' "$*"
  fi
}

dry_filter() {
  if [ "${DRY_RUN:-}" = "--dry-run" ]; then
    sed 's/^/DRY | /'
  else
    cat
  fi
}

rsync_owned_excludes() {
  local sub
  for sub in "${INSTANCE_OWNED_SUBTREES[@]}"; do printf -- '--exclude=/%s/\n' "$sub"; done
}

pathspec_owned_excludes() {
  local sub
  for sub in "${INSTANCE_OWNED_SUBTREES[@]}"; do printf -- ':(exclude)%s/%s/\n' "$1" "$sub"; done
}

is_instance_owned() {
  local relative="$1" sub
  for sub in "${INSTANCE_OWNED_SUBTREES[@]}"; do
    case "$relative" in "$sub"/*) return 0 ;; esac
  done
  return 1
}

# One answer to "what is a plugin?". Decided independently in four places
# before this: the staging enumeration walked plugins/ wholesale and filtered
# afterwards, the copy loop globbed plugins/*/, and two more sites re-derived
# the same [ -d ] test. The enumeration and the copy disagreeing is what
# produced `pathspec ... did not match any files` -- the stager handed git a
# skeleton entry the syncer had skipped, and that failed the whole config sync.
#
# Scar 2026-07-25: plugins/memory-lifecycle points at
# /Users/assafkip/projects/memory-lifecycle -- an old username, long gone -- so
# all_points_setup and Prodigy_Gold both failed there while instances that
# received the plugin back when the link resolved passed. That asymmetry made a
# skeleton-wide defect look instance-specific.
SKELETON_PLUGIN_ROOT="$SCRIPT_DIR/plugins"

# The ONLY enumeration. `*/` resolves symlinks, so a dangling entry never
# appears here -- which is exactly why the stager can no longer name a path the
# syncer will not write. Both now iterate this.
#
# NUL-separated, not newline. A directory name may legally contain a newline,
# and both consumers read this back with `read -r`; newline framing split such
# a name into two nonexistent plugins, whose rsync then failed and abandoned
# the instance's whole config sync. The pre-consolidation code did not have
# that failure -- it carried names by `find -print0` and by the glob value
# directly -- so newline framing here would have been a real regression.
managed_plugin_names() {
  local plugin_dir
  for plugin_dir in "$SKELETON_PLUGIN_ROOT"/*/; do
    [ -d "$plugin_dir" ] || continue
    plugin_dir="${plugin_dir%/}"
    printf '%s\0' "${plugin_dir##*/}"
  done
}

# Membership in the enumeration above, not an independent `[ -d ]` test. Those
# two disagreed on DOT-named directories -- `*/` skips them, `[ -d ]` does not
# -- so plugins/.hidden/ counted as managed here while the stager and the copy
# loop both ignored it, and the collision guard refused the whole sync over an
# untracked file living there. Nothing ever rsyncs such a directory, so there
# was no collision to protect against: the refusal was a pure false positive,
# and it is the last place the "what is a plugin?" answer was still given twice.
is_managed_plugin_path() {
  local top="${1#plugins/}" name
  top="${top%%/*}"
  [ -n "$top" ] || return 1
  while IFS= read -r -d '' name; do
    [ "$name" = "$top" ] && return 0
  done < <(managed_plugin_names)
  return 1
}

# One answer to "what does the disposable dry-run copy contain?".
#
# Two consumers must agree: the rsync that BUILDS the model, and the symlink
# walk that VETS it. A path the rsync skips can never be reached by a write, so
# refusing on a symlink inside it blocks the instance forever -- personal-brand's
# broken canonical links refused cole-gtm, its parent, for a sync that never
# touches personal-brand. A path the walk skips but the rsync copies is
# unvetted. They were kept in step on 2026-07-25 by passing one list into the
   300	            f"propagation copies a source the gate cannot read: {relative_path}"
   301	        ) from exc
   302	    try:
   303	        text = content.decode("utf-8")
   304	    except UnicodeDecodeError as exc:
   305	        if relative_path.endswith(BINARY_ASSET_SUFFIXES):
   306	            return None
   307	        raise PropagationSourceRefused(
   308	            f"propagation copies a source the gate cannot decode: "
   309	            f"{relative_path}"
   310	        ) from exc
   311	    if "\x00" in text:
   312	        # UTF-16/32 without a BOM decodes CLEANLY as UTF-8, because NUL is a
   313	        # legal codepoint. So it is neither refused nor flagged: it is scanned
   314	        # as mojibake the classifier cannot read, which is silent blindness
   315	        # rather than a refusal. No legitimate text source carries a NUL.
   316	        raise PropagationSourceRefused(
   317	            f"propagation copies a source that is not readable text: "
   318	            f"{relative_path}"
   319	        )
   320	    return text
   321	
   322	
   323	def _digest(text: str) -> str:
   324	    return hashlib.sha256(text.encode("utf-8")).hexdigest()
   325	
   326	
   327	def _rsync_filter_reason(entry, at_transfer_root: bool) -> str | None:
   328	    if entry.is_symlink():
   329	        # `rsync -a` implies -l, so only the transfer ROOT is dereferenced. A
   330	        # link inside the tree arrives as a link; its target never travels.
   331	        return "nested-symlink-copied-as-link"
   332	    is_directory = entry.is_dir(follow_symlinks=False)
   333	    if is_directory and at_transfer_root and entry.name in RSYNC_EXCLUDED_ROOT_DIRS:
   334	        return "excluded-by-rsync-filter"
   335	    if is_directory and entry.name in RSYNC_EXCLUDED_DIRS:
   336	        return "excluded-by-rsync-filter"
   337	    if entry.name in RSYNC_EXCLUDED_NAMES:
   338	        return True
   339	    if entry.name.startswith(RSYNC_EXCLUDED_PREFIXES):
   340	        return True
   341	    if entry.name.endswith(RSYNC_EXCLUDED_SUFFIXES):
   342	        return "excluded-by-rsync-filter"
   343	    return None
   344	
   345	
   346	def _skipped_by_shell_glob(name: str) -> bool:
   347	    """`plugins/*/` and `*.md` are shell globs, and `*` never matches a dot.
   348	
   349	    Walking a dotdir the updater cannot reach turns unreadable content that
   350	    never propagates into a refusal, which is how a gate gets switched off.
225:PLUGIN_COPY_EXCLUDES=(
290:plugin_copy_rsync_flags() {
292:  for entry in "${PLUGIN_COPY_EXCLUDES[@]}"; do
297:plugin_copy_filter_regex() {
299:  for entry in "${PLUGIN_COPY_EXCLUDES[@]}"; do
611:    | grep -vE "$(plugin_copy_filter_regex)" || true
1058:    # THE THIRD CONSUMER of PLUGIN_COPY_EXCLUDES, and the one that made the
1071:    plugin_excl_re="$(plugin_copy_filter_regex)"
2454:        # Flags derived from PLUGIN_COPY_EXCLUDES, never hand-listed here: this
2461:            $(plugin_copy_rsync_flags) \

exec
/bin/zsh -lc "rg -n '_rsync_filter_reason|excluded-by-rsync-filter|nested-symlink' q-system/.q-system/scripts/propagation-leak-gate.py q-system/.q-system/scripts/test/test-propagation-leak-sources.py; sed -n '215,310p' kipi-update.sh; sed -n '470,550p' q-system/.q-system/scripts/test_capability_gate.py; sed -n '630,710p' q-system/.q-system/scripts/capability-gate.py; rg -l 'def test_' q-system/.q-system/scripts plugins | while read f; do j=\"q-system/.q-system/capability/expected_tests/"'${f//'"\\//__}.json\"; [ -f \""'$j" ] && python3 -c '"'import json,sys; d=json.load(open(sys.argv[1])); print(d[\"runner\"],sys.argv[2])' \""'$j" "$f"; done | sort | uniq -c' in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-280
 succeeded in 595ms:
   1 pytest plugins/kipi-dsse/scripts/test_computed_receipts.py
   1 pytest plugins/kipi-dsse/scripts/test_scope_hook_fails_closed.py
   1 pytest plugins/prd-os/tests/test_virgin_repo_lifecycle.py
   1 pytest q-system/.q-system/scripts/test/test-propagation-entrypoints.py
   1 pytest q-system/.q-system/scripts/test/test-propagation-leak-baseline.py
   1 pytest q-system/.q-system/scripts/test/test-propagation-leak-gate.py
   1 pytest q-system/.q-system/scripts/test/test-propagation-leak-sources.py
   1 pytest q-system/.q-system/scripts/test_dispatch_reachability.py
   1 pytest q-system/.q-system/scripts/test_memory_autocapture.py
   1 pytest q-system/.q-system/scripts/test_memory_outcomes.py
   1 pytest q-system/.q-system/scripts/test_memory_reflect.py
   1 pytest q-system/.q-system/scripts/test_memory_scores_surface.py
   1 pytest q-system/.q-system/scripts/test_miyo_kb_hooks.py
   1 python3 plugins/prd-os/skills/fable-discipline/scripts/test_fable_discipline_lint.py
   1 python3 plugins/prd-os/tests/test_judgment_compiler.py
   1 python3 plugins/prd-os/tests/test_prd_split_from_linear.py
   1 python3 q-system/.q-system/scripts/test-ripple.py
   1 python3 q-system/.q-system/scripts/test/test-audhd-output-eval.py
   1 python3 q-system/.q-system/scripts/test/test-capability-map-wiring.py
   1 python3 q-system/.q-system/scripts/test/test-gate-13b-scope.py
   1 python3 q-system/.q-system/scripts/test/test-instance-ownership-contract.py
   1 python3 q-system/.q-system/scripts/test/test-update-preservation-manifest.py
   1 python3 q-system/.q-system/scripts/test/test-updater-receipt-contract.py
   1 python3 q-system/.q-system/scripts/test/test_linear_alert_triage.py
   1 python3 q-system/.q-system/scripts/test_autocapture_wiring.py
   1 python3 q-system/.q-system/scripts/test_correction_outcome.py
   1 python3 q-system/.q-system/scripts/test_dispatch_alias_reachability.py
   1 python3 q-system/.q-system/scripts/test_enforced_claim_lint.py
   1 python3 q-system/.q-system/scripts/test_instruction_budget_audit.py
   1 python3 q-system/.q-system/scripts/test_kb_graph_guard.py
   1 python3 q-system/.q-system/scripts/test_launchd_health_check.py
   1 python3 q-system/.q-system/scripts/test_lessons_index.py
   1 python3 q-system/.q-system/scripts/test_session_recall.py
   1 python3 q-system/.q-system/scripts/test_skill_hook_audit_local.py
   1 python3 q-system/.q-system/scripts/test_spillover_promote_selection.py
   1 python3 q-system/.q-system/scripts/test_spillover_ratchet.py
   1 python3 q-system/.q-system/scripts/test_voice_lint_caps.py
   1 python3 q-system/.q-system/scripts/test_voice_stop_gate_drain_only.py
   1 python3 q-system/.q-system/scripts/test_wiring_check.py
   1 python3 q-system/.q-system/scripts/tests/test-stat-verify.py

exec
/bin/zsh -lc "sed -n '390,430p' q-system/.q-system/scripts/propagation-leak-gate.py; rg -n 'reason' q-system/.q-system/scripts/propagation-leak-gate.py | head -40; rg -n 'zero-execution|pytest.*runner|executes_nothing' q-system/.q-system/scripts/test_capability_gate.py; python3 q-system/.q-system/scripts/test_capability_gate.py" in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-280
 exited 1 in 0ms:
        return
    text = _read_source_or_refuse(real_path, relative_path)
    if text is None:
        sources["excluded"].append(
            {"path": relative_path, "reason": "generated-or-binary-asset"}
        )
        return
    sources["worktree"].append(
        {
            "path": relative_path,
            "real_path": str(real_path),
            "sha256": _digest(text),
        }
    )


def _collect_dir(real_dir: Path, relative_dir: str, at_root: bool,
                 stack: list, sources: dict) -> None:
    for entry in _scandir_or_refuse(real_dir, relative_dir):
        relative_path = f"{relative_dir}/{entry.name}"
        reason = _rsync_filter_reason(entry, at_root)
        if reason is not None:
            sources["excluded"].append({"path": relative_path, "reason": reason})
        elif entry.is_dir():
            stack.append((Path(entry.path), relative_path, False))
        else:
            _record_source(Path(entry.path), relative_path, sources)


def _collect_tree(real_root: Path, relative_root: str, sources: dict) -> None:
    stack = [(real_root, relative_root, True)]
    while stack:
        real_dir, relative_dir, at_root = stack.pop()
        _collect_dir(real_dir, relative_dir, at_root, stack, sources)


def _collect_plugins(root: Path, sources: dict) -> None:
    """Mirror `for plugin_dir in plugins/*/`, which only matches directories."""
    plugins = root / "plugins"
    if not plugins.is_dir():
        return
327:def _rsync_filter_reason(entry, at_transfer_root: bool) -> str | None:
365:def _special_file_reason(real_path: Path) -> str | None:
387:    special = _special_file_reason(real_path)
389:        sources["excluded"].append({"path": relative_path, "reason": special})
394:            {"path": relative_path, "reason": "generated-or-binary-asset"}
410:        reason = _rsync_filter_reason(entry, at_root)
411:        if reason is not None:
412:            sources["excluded"].append({"path": relative_path, "reason": reason})
435:                {"path": relative_path, "reason": "not-matched-by-shell-glob"}
444:                {"path": relative_path, "reason": "not-a-directory-not-copied"}
461:                {"path": relative_path, "reason": "not-a-file-not-copied"}
484:                {"path": relative_path, "reason": "not-copied-by-propagation"}
520:        digest.update(f"{entry['path']}\0{entry['reason']}\0".encode("utf-8"))
525:    """Every source a propagation run copies, plus a reason for each exclusion."""
632:# `unclassified_populated_record`, for the volume reason above.
639:# PRD's resolved decision names six classes; it gives no reason for excluding
715:    A Mapping subclass whose `get` returns a constant reason and whose
722:            "fingerprint, not a single reason for the whole set"
733:            # written with a reason that was never attached to that fact.
746:    _refuse_reused_reasons(materialized)
756:    """The reason with any machine-appended fingerprint stripped off.
766:def _refuse_reused_reasons(justifications: dict) -> None:
778:    reused = {reason: keys for reason, keys in seen.items() if len(keys) > 1}
783:                f"{reason!r} covers {len(keys)} entries" for reason, keys in reused.items()
810:    `justifications` is a mapping from fingerprint to the reason that specific
811:    line is known and accepted. A single reason covering the set is refused:
821:def _document(counts: dict, reasons: dict, classifier_sha256: str | None) -> dict:
833:                "justification": str(reasons[key]).strip(),
929:    unbounded count is a bulk accept with one reason attached: `count: 999` on
955:    of which the PRD names as the reason this is a separate explicit step:
983:    # once a count is wrong there is nothing left to prove the carried reason
994:    reasons = {key: previous[key]["justification"] for key in kept}
997:        reasons[key] = justifications[key]
1000:        "document": _document(counts, reasons, classifier_sha256),
--- schema ---
  File "/Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-280/q-system/.q-system/scripts/test_capability_gate.py", line 778, in <module>
    main()
    ~~~~^^
  File "/Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-280/q-system/.q-system/scripts/test_capability_gate.py", line 770, in main
    fn()
    ~~^^
  File "/Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-280/q-system/.q-system/scripts/test_capability_gate.py", line 105, in sec_schema
    with tempfile.TemporaryDirectory() as tmp:
         ~~~~~~~~~~~~~~~~~~~~~~~~~~~^^
  File "/opt/homebrew/Cellar/python@3.14/3.14.6/Frameworks/Python.framework/Versions/3.14/lib/python3.14/tempfile.py", line 907, in __init__
    self.name = mkdtemp(suffix, prefix, dir)
                ~~~~~~~^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.14/3.14.6/Frameworks/Python.framework/Versions/3.14/lib/python3.14/tempfile.py", line 370, in mkdtemp
    dir = gettempdir()
                        ~~~~~~~~~~~^^
                            dirlist)
FileNotFoundError: [Errno 2] No usable temporary directory found in ['/var/folders/th/ty0kb7r56dvfrkgkrcbx9fv00000gn/T/', '/tmp', '/var/tmp', '/usr/tmp', '/Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-280']

codex
The first full harness attempt hit the environment’s read-only temporary-directory restriction before any PR code ran. I’m continuing with no-write stdin reproducers and direct test invocations, and I’ll distinguish sandbox limitations from code defects in the verdict.
exec
/bin/zsh -lc "rg -n 'pytest|pip install|requirements' .github kipi-update.sh q-system plugins -g '*.yml' -g '*.yaml' -g '*.sh' -g '*.toml' -g '*.txt' | head -240; python3 -m pytest --version; for f in plugins/kipi-dsse/scripts/test_computed_receipts.py plugins/kipi-dsse/scripts/test_scope_hook_fails_closed.py plugins/prd-os/tests/test_virgin_repo_lifecycle.py q-system/.q-system/scripts/test/test-propagation-entrypoints.py q-system/.q-system/scripts/test/test-propagation-leak-baseline.py q-system/.q-system/scripts/test/test-propagation-leak-gate.py q-system/.q-system/scripts/test/test-propagation-leak-sources.py q-system/.q-system/scripts/test_dispatch_reachability.py q-system/.q-system/scripts/test_memory_autocapture.py q-system/.q-system/scripts/test_memory_outcomes.py q-system/.q-system/scripts/test_memory_reflect.py q-system/.q-system/scripts/test_memory_scores_surface.py q-system/.q-system/scripts/test_miyo_kb_hooks.py; do echo \""'$f"; python3 -m pytest "$f" -q -p no:cacheprovider || break; done' in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-280
 succeeded in 110ms:
                     ~~~~~~~~~~~~~~~~~~~^^
    code = main()
    )
    ^
    raise exception
    teardown.throw(exception)
    ~~~~~~~~~~~~~~^^^^^^^^^^^
    config = yield
             ^^^^^
    self.parse(args)
    ~~~~~~~~~~^^^^^^
    )
    ^
    raise exception
    teardown.throw(exception)
    ~~~~~~~~~~~~~~^^^^^^^^^^^
    return (yield)
            ^^^^^
    ~~~~^^^^^^^^^^^^^^
                                              ~~~~~~~~~^^^
    TemporaryFile(buffering=0),
    ~~~~~~~~~~~~~^^^^^^^^^^^^^
    dir = gettempdir()
                        ~~~~~~~~~~~^^
                            dirlist)
FileNotFoundError: [Errno 2] No usable temporary directory found in ['/var/folders/th/ty0kb7r56dvfrkgkrcbx9fv00000gn/T/', '/tmp', '/var/tmp', '/usr/tmp', '/Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-280']

exec
/bin/zsh -lc "rg -n 'check-only|capability-gate.py' .github kipi q-system kipi-update.sh lefthook.yml settings-template.json -g '*.yml' -g '*.yaml' -g '*.sh' -g '*.py' -g 'kipi' | head -200; tail -100 q-system/.q-system/scripts/capability-gate.py; nl -ba q-system/.q-system/scripts/capability-gate.py | sed -n '180,255p;670,705p'; sed -n '30,50p' .github/workflows/validate.yml" in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-280
 succeeded in 0ms:
kipi-update.sh:2517:  GATE_SCRIPT="$path/q-system/.q-system/scripts/capability-gate.py"
kipi-update.sh:2520:      if python3 "$GATE_SCRIPT" --repo-root "$path" --check-only >"/tmp/kipi-gate-$$.log" 2>&1; then
kipi:275:      if [ -f "$CALLER_DIR/q-system/.q-system/scripts/capability-gate.py" ]; then
kipi:277:        python3 "$CALLER_DIR/q-system/.q-system/scripts/capability-gate.py" --repo-root "$CALLER_DIR" || exit 1
kipi:281:        echo "RED: capability-gate.py MISSING in $CALLER_DIR (run kipi update)" >&2
.github/workflows/validate.yml:52:        run: python3 q-system/.q-system/scripts/capability-gate.py --repo-root .
            continue
        errors.append(f"inert-engine: {rel} has no reference on any wiring surface "
                      "and no declared_inert entry")

    # Same walk over plugins/, reported not enforced (see the note above).
    plugin_surface = gather_wiring_text(root, {p.name for p in plugin_candidates})
    p_wired, changed = set(), True
    while changed:
        changed = False
        for p in sorted(plugin_candidates - p_wired):
            if references_engine(p.name, plugin_surface):
                p_wired.add(p)
                plugin_surface += "\n" + p.read_text(errors="ignore")
                changed = True
    for p in sorted(plugin_candidates - p_wired):
        rel = str(p.relative_to(root))
        if rel in declared:
            continue
        notes.append(f"INERT-ENGINE (report-only, plugins/): {rel} has no "
                     "reference on any wiring surface")


def main():
    ap = argparse.ArgumentParser(description=__doc__.splitlines()[0])
    ap.add_argument("--repo-root", default=".", help="repo root (default: cwd)")
    ap.add_argument("--check-only", action="store_true",
                    help="structure/diff/wiring/data checks only; skip test execution")
    args = ap.parse_args()
    root = Path(args.repo_root).resolve()
    refuse_if_worktree(root)

    errors, notes = [], []
    mode = detect_mode(root, errors)
    manifest = load_manifest(root, errors)
    if manifest is None:
        report(mode, errors, notes)
        sys.exit(1)
    load_overlay(root, manifest, errors)
    # Needs the filesystem (the ledger), so it cannot live inside
    # validate_manifest, which is handed data only. Placed before the
    # fail-closed exit below: a declared_inert entry pointing at a closed
    # decision is a structural manifest problem, not a finding about the repo.
    check_inert_spillover_live(root, manifest, errors, notes, mode)
    # counts note BEFORE the structural early-exit: an expired quarantine is a
    # structural error, and exiting first meant exactly those runs lost the
    # quarantine count the contract promises in EVERY summary (codex,
    # sag-quarantine-expiry)
    expected = manifest.get("expected_tests", [])
    q_count = sum(1 for e in expected if isinstance(e, dict) and e.get("quarantine"))
    notes.append(f"declared: {len(expected)} tests ({q_count} quarantined), "
                 f"{len(manifest.get('skeleton_only', []))} skeleton-only, "
                 f"{len(manifest.get('declared_inert', []))} declared-inert")
    # ASK-972: say the coverage boundary out loud on EVERY run. The escape used
    # to be a property of the source you had to go read; a GREEN that silently
    # meant "some of these were only half-checked" is the same silent-absence
    # class the whole gate exists to make loud.
    n_exempt = sum(1 for e in expected
                   if isinstance(e, dict) and e.get("path")
                   and not in_scan_scope(e["path"]))
    notes.append(f"scan scope: {len(expected) - n_exempt} declared entries inside "
                 f"the scan roots (checked BOTH directions), {n_exempt} exempt "
                 "from undeclared-artifact detection (existence-checked only)")
    if errors:  # fail closed on structural problems before trusting the sets
        report(mode, errors, notes)
        sys.exit(1)
    if q_count and args.check_only:
        for e in expected:
            q = e.get("quarantine")
            if q:
                notes.append(f"QUARANTINED (until {q['expires']}, {q['spillover_id']}): "
                             f"{e['path']} — {q['reason']}")
    diff_declared_vs_actual(root, manifest, errors, mode)
    check_required_data(root, manifest, mode, errors)
    # Inert-engine detection is a SKELETON-mode check. An instance's synced
    # scripts are wired by skeleton-root surfaces (validate.yml,
    # validate-separation.py, the kipi CLI) that do not exist inside the
    # instance repo — re-judging with that subset flagged 3 false inerts in
    # all 22 instances on first propagation (sp rollout finding, 2026-07-23).
    if mode == "skeleton":
        check_inert_engines(root, manifest, errors, notes)
    else:
        notes.append("inert-engine check: skeleton-only (instance wiring is "
                     "declared and gated in the skeleton)")
    if not args.check_only:
        run_tests(root, manifest, mode, errors, notes)
    report(mode, errors, notes)
    sys.exit(1 if errors else 0)


def report(mode, errors, notes):
    print(f"capability-gate mode={mode}")
    for n in notes:
        print(f"  {n}")
    for e in errors:
        print(f"  RED: {e}")
    print(f"capability-gate: {'RED (' + str(len(errors)) + ')' if errors else 'GREEN'}")


if __name__ == "__main__":
    main()
   180	    """One validator for canonical AND overlay entries (finding: overlay
   181	    entries were appended after validation and never validated themselves)."""
   182	    p = entry.get("path", "")
   183	    if unsafe_path(p):
   184	        errors.append(f"unsafe or non-relative path in expected_tests: {p!r}")
   185	        return
   186	    if not in_scan_scope(p):
   187	        # ASK-972: this was the silent escape. `in_scan_scope` was consulted only
   188	        # by the diff, which then had nothing to compare an unscanned path
   189	        # against, so such a path skipped the undeclared-artifact direction
   190	        # entirely and nothing said so. Refusing at DECLARATION time is the
   191	        # earliest point the two scopes can be held equal.
   192	        msg = declaration_scope_error(p, exempt_prefixes)
   193	        if msg:
   194	            errors.append(msg)
   195	    if entry.get("runner") not in RUNNERS:
   196	        errors.append(
   197	            f"expected_tests entry needs runner {'|'.join(RUNNERS)}: {p}")
   198	    if p in seen:
   199	        errors.append(f"duplicate expected_tests path: {p}")
   200	    seen.add(p)
   201	    t = entry.get("timeout_s", DEFAULT_TIMEOUT_S)
   202	    if not (isinstance(t, int) and TIMEOUT_MIN_S <= t <= TIMEOUT_MAX_S):
   203	        errors.append(f"timeout_s out of bounds [{TIMEOUT_MIN_S},{TIMEOUT_MAX_S}]: {p}")
   204	    validate_quarantine(entry, errors)
   205	
   206	
   207	# A declared test's runner has to be the one that actually EXECUTES it.
   208	#
   209	# ASK-1145, measured 2026-08-29. 13 manifest entries declared `runner: python3`
   210	# on pytest-shaped modules -- bare `def test_` at module level, no `__main__`
   211	# block. `python3 <file>` imports such a module, binds the function names, and
   212	# exits 0 having run nothing. The gate counted 13 tests, reported them green,
   213	# and was structurally blind to the 248 real cases inside them. Three of those
   214	# were FAILING, invisibly, in the fact-propagation leak gate on a public repo.
   215	#
   216	# `pytest` is added as a third runner rather than patching a `__main__` block
   217	# into 13 files, because the 14th file would arrive without one and nothing
   218	# would say so. The detector below is what makes that true: a python3-declared
   219	# module that is pytest-shaped is REFUSED at declaration time, so the mistake
   220	# cannot be made silently again.
   221	RUNNERS = ("python3", "bash", "pytest")
   222	
   223	# Module-level `def test_...`. Indented defs are methods on a unittest class,
   224	# which `python3 <file>` still cannot run without a `__main__` block, so they
   225	# count the same way.
   226	_BARE_TEST_DEF = re.compile(r"^\s*def test_\w*\s*\(", re.M)
   227	_HAS_MAIN = re.compile(r"^if __name__\s*==\s*[\"']__main__[\"']\s*:", re.M)
   228	
   229	
   230	def executes_nothing(entry, full):
   231	    """Refuse a python3-declared test that `python3 <file>` cannot execute.
   232	
   233	    The check is deliberately narrow: it fires only on a `.py` file declared
   234	    `python3` that DEFINES test functions and has NO `__main__` entry point.
   235	    A script that does its work at import time (most of the bash-adjacent
   236	    python helpers here) defines no `def test_` and is untouched.
   237	
   238	    An unreadable file is not a pass and not a failure here -- the missing-file
   239	    direction is already reported by the manifest diff, so this returns quietly
   240	    rather than adding a second, differently-worded error for one cause.
   241	    """
   242	    if entry.get("runner") != "python3":
   243	        return False
   244	    if not str(full).endswith(".py"):
   245	        return False
   246	    try:
   247	        text = full.read_text(encoding="utf-8", errors="replace")
   248	    except OSError:
   249	        return False
   250	    return bool(_BARE_TEST_DEF.search(text)) and not _HAS_MAIN.search(text)
   251	
   252	
   253	def validate_data_entry(entry, errors):
   254	    """required_data needs a safe path and a well-formed scope: a typo like
   255	    'skeletn' must be RED, not a silently-never-applies contract (finding-3
   670	    for entry in manifest.get("expected_tests", []):
   671	        path = entry.get("path", "")
   672	        if mode == "instance" and path in skeleton_only:
   673	            skipped += 1
   674	            continue
   675	        q = entry.get("quarantine")
   676	        if q:
   677	            quarantined += 1
   678	            notes.append(f"QUARANTINED (until {q['expires']}, {q['spillover_id']}): "
   679	                         f"{path} — {q['reason']}")
   680	            continue
   681	        full = root / path
   682	        if not full.is_file():
   683	            continue  # already reported by the diff
   684	        if executes_nothing(entry, full):
   685	            errors.append(
   686	                f"zero-execution test: {path} defines test functions but has "
   687	                f"no __main__ block, so `python3 {path}` binds them and exits "
   688	                f"0 without running one. Declare runner pytest, or add a "
   689	                f"__main__ entry point (ASK-1145).")
   690	            continue
   691	        runner = entry["runner"]
   692	        if runner == "pytest":
   693	            # `-p no:cacheprovider` so a gate run leaves no .pytest_cache in the
   694	            # tree it just measured; `-q` keeps the failure tail readable inside
   695	            # the 20-line cap below.
   696	            cmd = ["python3", "-m", "pytest", str(full), "-q", "-p", "no:cacheprovider"]
   697	        elif runner == "python3":
   698	            cmd = ["python3", str(full)]
   699	        else:
   700	            cmd = ["bash", str(full)]
   701	        timeout = entry.get("timeout_s", DEFAULT_TIMEOUT_S)
   702	        r = run_contained(cmd, root, env, timeout)
   703	        if r.timed_out:
   704	            # the partial output often names the hang (a prompt, a URL being
   705	            # polled) — discarding it made timeouts undiagnosable (codex,
      - name: Configure git identity for the updater tests
        run: |
          git config --global user.name "kipi-ci"
          git config --global user.email "ci@kipi.invalid"

      - name: Set up Python
        uses: actions/setup-python@v6
        with:
          python-version: '3.12'

      # pytest must exist BEFORE the gate: declared tests import it, and the
      # gate runs the full suite (first CI execution failed on step order,
      # 2026-07-23).
      - name: Install test dependencies
        run: pip install pytest

      # The capability gate replaces the old hardcoded 4-test loop: it
      # discovers every test artifact by convention, diffs declared-vs-actual
      # BOTH directions, and runs the full declared suite. The hardcoded loop
      # ran 4 of 41 declared artifacts — 90.2% never executed anywhere
      # (prd-silent-absence-capability-gate-2026-07-23).

exec
/bin/zsh -lc 'python3 -c $'"'import importlib.util\\nfrom pathlib import Path\\np=Path(\"q-system/.q-system/scripts/capability-gate.py\")\\ns=importlib.util.spec_from_file_location(\"gate\",p)\\nm=importlib.util.module_from_spec(s); s.loader.exec_module(m)\\nentry={\"path\":\"q-system/.q-system/scripts/test_future.py\",\"runner\":\"python3\"}\\nerrors=[]\\nm.validate_test_entry(entry,set(),errors)\\nF=type(\"F\",(),{\"__str__\":lambda self:\"test_future.py\",\"read_text\":lambda self,**kwargs:\"def test_real_behavior():\\\\n    assert False\\\\n\"})\\nprint(\"validate_test_entry errors:\",errors)\\nprint(\"executes_nothing:\",m.executes_nothing(entry,F()))'; python3 -c "'$'"'import ast\\nfrom pathlib import Path\\nt=ast.parse(Path(\"q-system/.q-system/scripts/capability-gate.py\").read_text())\\nfor n in ast.walk(t):\\n if isinstance(n,ast.If) and isinstance(n.test,ast.UnaryOp) and isinstance(n.test.op,ast.Not) and isinstance(n.test.operand,ast.Attribute) and n.test.operand.attr==\"check_only\": print(\"check-only routing:\",ast.unparse(n))'; python3 -c "'$'"'import importlib.util\\nfrom pathlib import Path\\np=Path(\"q-system/.q-system/scripts/propagation-leak-gate.py\")\\ns=importlib.util.spec_from_file_location(\"gate\",p); m=importlib.util.module_from_spec(s); s.loader.exec_module(m)\\nE=type(\"E\",(),{\"name\":\".env.local\",\"is_symlink\":lambda self:False,\"is_dir\":lambda self,follow_symlinks=False:False})\\nr=m._rsync_filter_reason(E(),False)\\nprint(repr(r),type(r).__name__)'" in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-280
 succeeded in 0ms:
validate_test_entry errors: []
executes_nothing: True
check-only routing: if not args.check_only:
    run_tests(root, manifest, mode, errors, notes)
True bool

codex
## Finding 1

- Severity: minor
- Claim: The zero-execution detector is skipped by `--check-only`, even though that mode is used by unattended updates for structural validation and the code claims invalid declarations are refused at declaration time.
- Location: `q-system/.q-system/scripts/capability-gate.py:684`
- Reproducer:

```bash
python3 -c $'import importlib.util\nfrom pathlib import Path\np=Path("q-system/.q-system/scripts/capability-gate.py")\ns=importlib.util.spec_from_file_location("gate",p)\nm=importlib.util.module_from_spec(s); s.loader.exec_module(m)\nentry={"path":"q-system/.q-system/scripts/test_future.py","runner":"python3"}\nerrors=[]\nm.validate_test_entry(entry,set(),errors)\nF=type("F",(),{"__str__":lambda self:"test_future.py","read_text":lambda self,**kwargs:"def test_real_behavior():\\n    assert False\\n"})\nprint("validate_test_entry errors:",errors)\nprint("executes_nothing:",m.executes_nothing(entry,F()))'
```

- Real output:

```text
validate_test_entry errors: []
executes_nothing: True
```

The invalid declaration passes structural validation. Detection only happens inside `run_tests()`, which `--check-only` explicitly skips. Move the check into manifest validation or an equivalent pre-execution structural pass.

## Finding 2

- Severity: minor
- Claim: New `.env` exclusions return boolean `True` as their reason instead of the documented string reason, corrupting the source inventory’s reason field and classifier digest semantics.
- Location: `q-system/.q-system/scripts/propagation-leak-gate.py:338`
- Reproducer:

```bash
python3 -c $'import importlib.util\nfrom pathlib import Path\np=Path("q-system/.q-system/scripts/propagation-leak-gate.py")\ns=importlib.util.spec_from_file_location("gate",p); m=importlib.util.module_from_spec(s); s.loader.exec_module(m)\nE=type("E",(),{"name":".env.local","is_symlink":lambda self:False,"is_dir":lambda self,follow_symlinks=False:False})\nr=m._rsync_filter_reason(E(),False)\nprint(repr(r),type(r).__name__)'
```

- Real output:

```text
True bool
```

The two new branches should return `"excluded-by-rsync-filter"` like every other rsync exclusion.

## What is sound

- The 13 corrected manifests point at real pytest-shaped modules.
- CI installs pytest before executing the full capability gate.
- The updater’s `.env` and `.env.*` patterns match the new propagation-gate exclusions.
- The mirror test rejects an empty parse, avoiding a vacuous set comparison.
- Full pytest execution was attempted, but this read-only review sandbox provides no writable temporary directory. Pytest stopped before collecting PR tests with `FileNotFoundError: No usable temporary directory found`.

## Verdict

Fix first: run `executes_nothing()` during structural validation so `--check-only` cannot approve the exact declaration defect this PR intends to prevent.

FINDINGS:
minor|The zero-execution detector is skipped by --check-only, allowing structurally invalid python3 test declarations through unattended update validation|q-system/.q-system/scripts/capability-gate.py:684
minor|The new .env exclusion branches return boolean True instead of the documented string exclusion reason|q-system/.q-system/scripts/propagation-leak-gate.py:338
END FINDINGS
tokens used
58,222
## Finding 1

- Severity: minor
- Claim: The zero-execution detector is skipped by `--check-only`, even though that mode is used by unattended updates for structural validation and the code claims invalid declarations are refused at declaration time.
- Location: `q-system/.q-system/scripts/capability-gate.py:684`
- Reproducer:

```bash
python3 -c $'import importlib.util\nfrom pathlib import Path\np=Path("q-system/.q-system/scripts/capability-gate.py")\ns=importlib.util.spec_from_file_location("gate",p)\nm=importlib.util.module_from_spec(s); s.loader.exec_module(m)\nentry={"path":"q-system/.q-system/scripts/test_future.py","runner":"python3"}\nerrors=[]\nm.validate_test_entry(entry,set(),errors)\nF=type("F",(),{"__str__":lambda self:"test_future.py","read_text":lambda self,**kwargs:"def test_real_behavior():\\n    assert False\\n"})\nprint("validate_test_entry errors:",errors)\nprint("executes_nothing:",m.executes_nothing(entry,F()))'
```

- Real output:

```text
validate_test_entry errors: []
executes_nothing: True
```

The invalid declaration passes structural validation. Detection only happens inside `run_tests()`, which `--check-only` explicitly skips. Move the check into manifest validation or an equivalent pre-execution structural pass.

## Finding 2

- Severity: minor
- Claim: New `.env` exclusions return boolean `True` as their reason instead of the documented string reason, corrupting the source inventory’s reason field and classifier digest semantics.
- Location: `q-system/.q-system/scripts/propagation-leak-gate.py:338`
- Reproducer:

```bash
python3 -c $'import importlib.util\nfrom pathlib import Path\np=Path("q-system/.q-system/scripts/propagation-leak-gate.py")\ns=importlib.util.spec_from_file_location("gate",p); m=importlib.util.module_from_spec(s); s.loader.exec_module(m)\nE=type("E",(),{"name":".env.local","is_symlink":lambda self:False,"is_dir":lambda self,follow_symlinks=False:False})\nr=m._rsync_filter_reason(E(),False)\nprint(repr(r),type(r).__name__)'
```

- Real output:

```text
True bool
```

The two new branches should return `"excluded-by-rsync-filter"` like every other rsync exclusion.

## What is sound

- The 13 corrected manifests point at real pytest-shaped modules.
- CI installs pytest before executing the full capability gate.
- The updater’s `.env` and `.env.*` patterns match the new propagation-gate exclusions.
- The mirror test rejects an empty parse, avoiding a vacuous set comparison.
- Full pytest execution was attempted, but this read-only review sandbox provides no writable temporary directory. Pytest stopped before collecting PR tests with `FileNotFoundError: No usable temporary directory found`.

## Verdict

Fix first: run `executes_nothing()` during structural validation so `--check-only` cannot approve the exact declaration defect this PR intends to prevent.

FINDINGS:
minor|The zero-execution detector is skipped by --check-only, allowing structurally invalid python3 test declarations through unattended update validation|q-system/.q-system/scripts/capability-gate.py:684
minor|The new .env exclusion branches return boolean True instead of the documented string exclusion reason|q-system/.q-system/scripts/propagation-leak-gate.py:338
END FINDINGS

PR #272, opened 2026-08-29
Mutation sweep: a detector for tests that run, pass, and cannot fail

assafkip:
Founder-directed. 17 instances of one class were measured in a single session:
a check that RUNS, PASSES, and is structurally blind to the thing it exists to
catch. This lands the detector and its self-guard. Fixes ship separately.

## What it does

`q-system/.q-system/scripts/mutation-sweep.py`, over the capability manifest's
declared tests (183 with this PR's own entry).

Two-stage probe per (test, subject):

1. **TRIPWIRE** — replace the subject's body with an immediate `exit 97`. The
   control must turn the test red, which is also what proves attribution. A
   test still green here does not exercise the file.
2. **DISARM** — neuter every failure-signalling site at once (nonzero
   `sys.exit` to `sys.exit(0)`, `return False` to `return True`, bash
   `exit 1` to `exit 0`). The subject can no longer report failure. Still
   green means the test cannot observe its subject's verdict.

`--zero-exec-scan` covers what mutation testing structurally cannot see: the
gate runs each python3 entry as `python3 <file>`, so a pytest-style module with
no entry point imports, defines its tests, and exits 0. Nothing ran, and it
reads as a pass.

`--report-only` re-derives the report from the append-only ledger.

## Guarding the harness itself

Two of the 17 instances WERE mutation harnesses that reported perfect survival —
one because an import error killed every case, one because it ran no tests at
all. A survival report from a harness that ran nothing is the worst possible
output here, so every verdict is proved:

- mutant applied confirmed by sha256 **re-read from disk**, not by "the write returned"
- mutant confirmed syntactically valid (`ast.parse` / `bash -n`) so a broken
  mutant cannot score a free kill off any test that merely imports it
- baseline must be green or the test is excluded rather than scored
- after every restore the test is re-run and must be green again
- private `PYTHONPYCACHEPREFIX` per run: a one-char mutant leaves the file the
  same size and the restore lands in the same second, so CPython will serve the
  mutant's cached bytecode for restored source
- restore is a `cp` from an in-process backup, never a git checkout, installed
  on SIGTERM/SIGINT/SIGHUP as well as `finally` (`finally` does not run on a
  signal; the first long run was killed mid-flight)
- `run_contained` is imported from a **copy** of `capability-gate.py`, so tests
  run exactly as the gate runs them and the sweep can never execute its own
  mutant as its runner

`test-mutation-sweep.sh` is the paired check, declared in the manifest so it
[...]

PR #281, opened 2026-08-30
fix(terminal-states): main is red on a healthy dispatcher, because liveness reads the work log (ASK-1146)

assafkip:
## main is red right now

Measured on a clean checkout of `origin/main` at 7de326b8, not inferred:
`test-terminal-states.sh` reports **37 passed / 2 failed**, and the capability
gate is RED because of it. Both failures were also present on the rebased #272
and #265 trees, which is what made it clear the base was the problem rather than
either PR.

## The verdict was wrong

```
com.kipi.dispatch is loaded but has not run in 21677s (interval allows 3600s)
-- a loaded job that stopped running is a dead consumer
```

The dispatcher was alive the whole time. Its heartbeat file was **30 seconds
old** when this was measured. What it had actually done, from its own log:

```
2026-08-30T00:18:13Z heartbeat: RESUMED after 45m without a beat
2026-08-29T19:17:26Z DAILY CAP: 10/10 issues dispatched for budget day
                     2026-08-29 (lane=production), stopping until 7:00 local
```

Capped and idle by design, and it had already self-healed from the disk-full
episode earlier in the evening.

## Why the check could not see that

`run_evidence` pointed at `~/.config/kipi/dispatch.log`, which only moves when
the dispatcher **does** something. A correctly-capped job is byte-identical to a
dead one for up to fourteen hours, every night. A gate that goes red every
evening on a healthy system is a gate someone switches off, and then it protects
nothing.

`~/.config/kipi/dispatch-lastbeat` is the right evidence and already exists.
`kipi-dispatch.sh` writes it unconditionally at the top of every run:

```bash
printf '%s' "$NOW_EPOCH" > "$BEAT_FILE"
```

That line sits **before** the concurrency skip and before the daily-cap exit, so
it answers the question the check is actually asking: did the job run. 18 rows
repointed.

## Measurement

- 39 passed / 0 failed, confirmed twice for stability rather than once.
- Mutation-tested: pointing the rows back at `dispatch.log` puts both failures
  back (39 → 37 passed, 2 failed). The fix can go red for the reason it exists.

## The shape

Nothing here was a missing check. The check ran, produced a confident verdict,
and was blind to the designed state it was misreading. It also cost real time:
this began as a six-hour outage investigation, including a `launchctl kickstart`
that turned out to be unnecessary, because the check's own wording ("dead
consumer") asserted a diagnosis it had not established.

### DIFF
--- a/q-system/.q-system/terminal-states.json
+++ b/q-system/.q-system/terminal-states.json
@@ -122,7 +122,7 @@
       "liveness_check": {
         "kind": "launchd",
         "label": "com.kipi.dispatch",
-        "run_evidence": "~/.config/kipi/dispatch.log",
+        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
         "max_age_s": 3600
       },
       "reentry": {
@@ -142,7 +142,7 @@
       "liveness_check": {
         "kind": "launchd",
         "label": "com.kipi.dispatch",
-        "run_evidence": "~/.config/kipi/dispatch.log",
+        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
         "max_age_s": 3600
       },
       "reentry": {
@@ -162,7 +162,7 @@
       "liveness_check": {
         "kind": "launchd",
         "label": "com.kipi.dispatch",
-        "run_evidence": "~/.config/kipi/dispatch.log",
+        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
         "max_age_s": 3600
       },
       "reentry": {
@@ -182,7 +182,7 @@
       "liveness_check": {
         "kind": "launchd",
         "label": "com.kipi.dispatch",
-        "run_evidence": "~/.config/kipi/dispatch.log",
+        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
         "max_age_s": 3600
       },
       "reentry": {
@@ -202,7 +202,7 @@
       "liveness_check": {
         "kind": "launchd",
         "label": "com.kipi.dispatch",
-        "run_evidence": "~/.config/kipi/dispatch.log",
+        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
         "max_age_s": 3600
       },
       "reentry": {
@@ -309,7 +309,7 @@
       "liveness_check": {
         "kind": "launchd",
         "label": "com.kipi.dispatch",
-        "run_evidence": "~/.config/kipi/dispatch.log",
+        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
         "max_age_s": 3600
       },
       "reentry": {
@@ -404,7 +404,7 @@
       "liveness_check": {
         "kind": "launchd",
         "label": "com.kipi.dispatch",
-        "run_evidence": "~/.config/kipi/dispatch.log",
+        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
         "max_age_s": 3600
       },
       "reentry": {
@@ -440,7 +440,7 @@
       "liveness_check": {
         "kind": "launchd",
         "label": "com.kipi.dispatch",
-        "run_evidence": "~/.config/kipi/dispatch.log",
+        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
         "max_age_s": 3600
       },
       "reentry": {
@@ -490,7 +490,7 @@
       "liveness_check": {
         "kind": "launchd",
         "label": "com.kipi.dispatch",
-        "run_evidence": "~/.config/kipi/dispatch.log",
+        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
         "max_age_s": 3600
       },
       "reentry": {
@@ -520,7 +520,7 @@
       "liveness_check": {
         "kind": "launchd",
         "label": "com.kipi.dispatch",
-        "run_evidence": "~/.config/kipi/dispatch.log",
+        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
         "max_age_s": 3600
       },
       "reentry": {
@@ -540,7 +540,7 @@
       "liveness_check": {
         "kind": "launchd",
         "label": "com.kipi.dispatch",
-        "run_evidence": "~/.config/kipi/dispatch.log",
+        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
         "max_age_s": 3600
       },
       "reentry": {
@@ -580,7 +580,7 @@
       "liveness_check": {
         "kind": "launchd",
         "label": "com.kipi.dispatch",
-        "run_evidence": "~/.config/kipi/dispatch.log",
+        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
         "max_age_s": 3600
       },
       "reentry": {
@@ -610,7 +610,7 @@
       "liveness_check": {
         "kind": "launchd",
         "label": "com.kipi.dispatch",
-        "run_evidence": "~/.config/kipi/dispatch.log",
+        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
         "max_age_s": 3600
       },
       "reentry": {
@@ -640,7 +640,7 @@
       "liveness_check": {
         "kind": "launchd",
         "label": "com.kipi.dispatch",
-        "run_evidence": "~/.config/kipi/dispatch.log",
+        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
         "max_age_s": 3600
       },
       "reentry": {
@@ -690,7 +690,7 @@
       "liveness_check": {
         "kind": "launchd",
         "label": "com.kipi.dispatch",
-        "run_evidence": "~/.config/kipi/dispatch.log",
+        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
         "max_age_s": 3600
       },
       "reentry": {
@@ -710,7 +710,7 @@
       "liveness_check": {
         "kind": "launchd",
         "label": "com.kipi.dispatch",
-        "run_evidence": "~/.config/kipi/dispatch.log",
+        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
         "max_age_s": 3600
       },
       "reentry": {
@@ -746,7 +746,7 @@
       "liveness_check": {
         "kind": "launchd",
         "label": "com.kipi.dispatch",
-        "run_evidence": "~/.config/kipi/dispatch.log",
+        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
         "max_age_s": 3600
       },
       "reentry": {
@@ -766,7 +766,7 @@
       "liveness_check": {
         "kind": "launchd",
         "label": "com.kipi.dispatch",
-        "run_evidence": "~/.config/kipi/dispatch.log",
+        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
         "max_age_s": 3600
       },
       "reentry": {


assafkip:
## Verdict: APPROVE

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-281-20260829-232434.md` (158836 bytes).

```
FINDINGS:
END FINDINGS
```

--- reviewer output, last 54978 bytes of 158836 (full review at the path above) ---

   161	      "consumer": "com.kipi.dispatch -- the next run retries from scratch; nothing about the issue changed",
   162	      "liveness_check": {
   163	        "kind": "launchd",
   164	        "label": "com.kipi.dispatch",
   165	        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
   166	        "max_age_s": 3600
   167	      },
   168	      "reentry": {
   169	        "path": "kipi-dispatch.sh",
   170	        "marker": "DISPATCH_ARGV=(./kipi converge --issue",
   171	        "why": "The issue stays backlog/unstarted and ready, so the next dispatch launch re-selects it."
   172	      }
   173	    },
   174	    {
   175	      "id": "claim-contended",
   176	      "source": "linear-worker",
   177	      "class": "infra-retry",
   178	      "marker": "claimed by another session",
   179	      "sites": 1,
   180	      "what": "linear-claim.py exited 3: another session holds this worktree. A skip, not an error.",
   181	      "consumer": "com.kipi.dispatch -- the holding session finishes and releases; the next run claims it. A claim left by a killed run is reclaimable on read via --holder-pid.",
   182	      "liveness_check": {
   183	        "kind": "launchd",
   184	        "label": "com.kipi.dispatch",
   185	        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
   186	        "max_age_s": 3600
   187	      },
   188	      "reentry": {
   189	        "path": "kipi-dispatch.sh",
   190	        "marker": "DISPATCH_ARGV=(./kipi converge --issue",
   191	        "why": "The issue is untouched and still ready, so the next dispatch launch re-selects it once the claim clears."
   192	      }
   193	    },
   194	    {
   195	      "id": "claim-infra-failure",
   196	      "source": "linear-worker",
   197	      "class": "infra-retry",
   198	      "marker": "claim failed rc=",
   199	      "sites": 1,
   200	      "what": "The claim failed for a reason that is not contention. Explicitly not counted against the issue.",
   201	      "consumer": "com.kipi.dispatch -- retried next run; a persistent failure surfaces through launchd-health-check.py, which pages on a job whose last run exited non-zero",
   202	      "liveness_check": {
   203	        "kind": "launchd",
   204	        "label": "com.kipi.dispatch",
   205	        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
   206	        "max_age_s": 3600
   207	      },
   208	      "reentry": {
   209	        "path": "kipi-dispatch.sh",
   210	        "marker": "DISPATCH_ARGV=(./kipi converge --issue",
   211	        "why": "Nothing about the issue changed, so the next dispatch launch re-selects it."
   212	      }
   213	    },
   214	    {
   215	      "id": "blocked-capability",
   216	      "source": "linear-worker",
   217	      "class": "durable-queue",
   218	      "marker": [
   219	        "\"blocked:capability\" in labels",
   220	        "REFUSE_LABEL=\"blocked:capability\""
   221	      ],
   222	      "sites": 2,
   223	      "terminal": true,
   224	      "rationale": "REFUSAL ON EVIDENCE, re-checked 2026-08-18 against the founder's reversal directive. The method this issue demands is: look for a successor that became real AFTER 2026-08-01, do not reason from the state name. Nothing did. Grepped the whole scripts tree for a reader of the blocked:capability label: linear-worker.sh applies it and counts it (held_with(\"blocked:capability\")), and no other file selects on it. ASK-288 (`a blocked:capability label never expires, nothing re-tests the block`) is still open and is the named owner of the re-test, and this issue's Not-doing line excludes building it. The underlying reason is unchanged and is not a naming problem: the runner lacks something it cannot grant itself, and an agent must never grant itself an authorization (2026-05-17 scar). Revised to a consumer when ASK-288's re-test ships and is proven, not when it merges.",
   225	      "evidence": {
   226	        "path": "q-system/.q-system/scripts/linear-worker.sh",
   227	        "marker": "blocked_cap = held_with(\"blocked:capability\")",
   228	        "why": "The only other reader of this label in the fleet is a COUNT in the same file that applies it. No selector re-enters the state."
   229	      }
   230	    },
   231	    {
   232	      "id": "owner-assaf",
   233	      "source": "linear-worker",
   234	      "class": "error-path",
   235	      "marker": "\"owner:assaf\" in labels",
   236	      "sites": 1,
   237	      "error_path": true,
   238	      "what": "An issue carrying owner:assaf. REVERSED 2026-08-03 by founder directive (`nothing should be on me`): this was a designed founder queue in the archived PRD prd-terminal-state-redrive-2026-08-01 and it is now a DEFECT. The queue was emptied this session (85 issues cleared) and by directive must never refill.",
   239	      "rationale": "The v1 row read `this one is the intended destination, which is exactly why the rest must not quietly imitate it`. That reasoning was correct on 2026-08-01 and the founder has since closed the queue. A terminal:true row would keep certifying the routing as designed; an error_path row asserts the opposite -- that any issue reaching it is a bug the run has to say out loud. ready() still excludes the issue (the loop must not work something the founder marked hands-off, and silently working it would be worse), but the run now COUNTS and REPORTS the population and pages when it is non-zero, so a refilling queue is loud instead of invisible.",
   240	      "error_evidence": {
   241	        "path": "q-system/.q-system/scripts/linear-worker.sh",
   242	        "marker": "DEFECT: owner:assaf",
   243	        "why": "The loud refusal line the worker prints per run when the founder queue is non-empty. Absent from the source, this row is a claim with no code behind it."
   244	      }
   245	    },
   246	    {
   247	      "id": "not-owner-sana",
   248	      "source": "linear-worker",
   249	      "class": "selection-filter",
   250	      "marker": "\"owner:sana\" not in labels",
   251	      "sites": 1,
   252	      "what": "The issue is not addressed to this worker.",
   253	      "terminal": true,
   254	      "rationale": "Not an exit and not a refusal -- the issue was never routed here. A machine consumer would mean this worker picking up work nobody assigned to it, which is the opposite of what the ownership label is for. Board triage decides who owns an issue."
   255	    },
   256	    {
   257	      "id": "state-not-open",
   258	      "source": "linear-worker",
   259	      "class": "selection-filter",
   260	      "marker": "i[\"state\"][\"type\"] not in",
./kipi-dispatch.sh:8:# job. So the loop only ran while someone watched it, which is not autonomy --
./kipi-dispatch.sh:440:# selection loop. That halt was removed -- it broke test-dispatch-liveness 6a and
./kipi-dispatch.sh:1100:# file and has been proven (MAX_CONCURRENT, the daily budget, the liveness
./kipi-dispatch.sh:1129:# second liveness rule spelled out in this file is the drift this repo keeps
./kipi-dispatch.sh:1212:# `ps -Ao args=` form and same [c] self-match guard as the liveness check.
./kipi-dispatch.sh:1423:# pid capture, the liveness assert) is proven and is not duplicated per action;
./plugins/kipi-dsse/scripts/test_computed_receipts.py:11:    the receipt, and the receipt carries the evidence that the work happened.
./plugins/prd-os/skills/fable-discipline/scripts/fable-discipline-lint.py:326:# that built it. Basename alone is not identity: test-dispatch-liveness.sh runs
./plugins/prd-os/tests/test_spillover_combined_rule.py:66:    # A REAL git repo. ASK-527's liveness guard asks git whether the active spec
./plugins/prd-os/skills/fable-discipline/scripts/test_fable_discipline_lint.py:170:    # switched off. test-dispatch-liveness.sh runs "$ROOT/converge.sh" from a
./plugins/prd-os/tests/test_spillover_gate_scope.py:52:    # spec is tracked, and a fake .git makes that query fail, which the liveness
./plugins/kipi-core/skills/founder-voice/SKILL.md:8:You are writing as the founder. Your job is to transform any content into their authentic voice. This is NOT about adding personality to generic copy. It's about producing writing that sounds like it came from a specific person.
./plugins/prd-os/tests/test_active_scope_liveness.py:222:    """`--scope` must clear the SAME liveness bar as an inferred scope.
./plugins/prd-os/templates/gap-classes.md:87:16. **A liveness/health endpoint must never 500 on one bad data row.** Health
./plugins/prd-os/scripts/prd_runner.py:2767:    # An EXPLICIT --scope clears the same liveness bar as an inferred one.
./q-system/lessons/a-guard-the-compound-condition-never-reaches-is-not-a-guar.md:14:3. An orphan branch was guarded by `not installed and label not in overrides`. A retired job whose stale override row survived satisfied the first clause and failed the second, so it fell past the orphan branch into the drift branch and paged about a job with no executable on disk. Its sibling, retired identically but with no leftover row, classified correctly. The two inputs differed by one clause.
./q-system/lessons/a-guard-that-only-fires-on-a-start-cannot-see-a-stop.md:12:The same asymmetry has a second face: a scheduled job that succeeds while doing nothing. A pipeline branch with no rows to send produces a successful execution indistinguishable from a busy one, so liveness monitoring reports green for the entire outage.
./q-system/lessons/a-scheduled-job-runs-in-a-bare-environment-not-your-shell.md:8:A task that succeeds when you run it by hand and fails when a scheduler runs it is almost never random. A timer, cron, service manager, or CI runner starts each job from a minimal, non-interactive environment: it does not source your login shell, your profile, or the variables you exported into your terminal. Any credential, path, or flag that lives only in your interactive session is invisible to the automated run. The by-hand success is the trap, not the reassurance. It proves the code works when the missing config happens to be present, which is exactly the condition that never holds under the scheduler. So a manual test can pass indefinitely while every scheduled run is silently degraded.
./q-system/lessons/a-scheduled-job-runs-in-a-bare-environment-not-your-shell.md:12:The second half of this failure class: when a required input goes missing because of such a config gap, the job must not degrade in silence. A pipeline that treats a missing credential as a per-item skip and continues with a smaller pool produces a run that is smaller-but-green, and a degraded-but-passing run is indistinguishable from a healthy one to anyone reading the status. Two guards close this. First, put a floor on the work: if an enabled input yields zero, that is a distinct outcome from a genuinely quiet period, and it must alert rather than pass. Second, distinguish 'this input was disabled or absent' from 'this input ran and returned nothing' — only the second is legitimately empty. Without the floor and the distinction, a config gap in one input rots quietly across many runs, because each individual run still finished.
./q-system/lessons/prove-delivery-with-receipts-not-proxies.md:15:- No monitor fires, because every monitor watches a proxy: process liveness, exit code, "a request was sent", a stored claim.
./q-system/lessons/a-pause-with-a-deadline-needs-a-mechanism-not-a-note.md:21:- The stop-side check is a liveness assertion keyed to the declaration, not the switch: a path declared as delivering, that has delivered nothing within its expected interval, is red.
./q-system/lessons/feedback-lands-where-artifacts-exist.md:8:When an operator gives the same feedback repeatedly and the output keeps failing the same way, check which layers of the system have editable artifacts before diagnosing the feedback as vague or the model as weak. Fixes flow to whatever has a file: a prompt, a corpus, a gate list, a config. A concern with no artifact -- purpose, audience, the job the output is supposed to do -- cannot absorb a fix, so every round of feedback about it gets translated into the nearest layer that CAN be edited, and the actual complaint survives untouched.
./q-system/lessons/a-no-op-run-is-not-a-success.md:8:A job's terminal status must be derived from the work it actually produced, not from the fact that it reached the end without throwing. "No error occurred" and "real work was done" are two different facts, and a run that completes cleanly while producing nothing collapses them into a single misleading green light. The person reading the status sees "done" and infers "the work happened"; the two can diverge silently.
./q-system/lessons/a-freshness-deadman-must-live-off-the-machine-it-watches.md:16:2. Monitor the OUTCOME, not the process. Check the freshness of the delivered artifact (last-produced timestamp, output row, published file), not whether the job process is alive. Absence of a fresh deliverable is the signal. This catches both 'host was off' and 'ran but produced nothing' with one check.
./q-system/lessons/a-predicate-whose-false-branch-is-unreachable-reports-suc.md:20:**The heartbeat where zero and never-ran are the same value.** `dropped_not_for_client` is durable and nothing reads it, so a 0 cannot be distinguished from a skipped run. A metric with no liveness signal answers "healthy" identically to "dead."
./q-system/lessons/a-predicate-whose-false-branch-is-unreachable-reports-suc.md:38:- **Distinguish zero from never.** A counter, a metric, or a returned collection needs a liveness signal separate from its value, or its healthy state and its dead state are the same bytes.
./q-system/lessons/an-item-count-proves-a-step-was-reached-not-that-it-ran.md:17:4. The corollary for a scheduled job: a successful run is not a completed unit of work. A branch with nothing to do succeeds identically to one that did everything, so success is the wrong signal for any conditional writer.
./q-system/lessons/an-unattended-session-freezes-on-its-first-permission-prompt.md:8:An always-on interactive agent session started to run a loop autonomously (relaunched by a supervisor, seeded with a task prompt) will stall on the FIRST tool-permission prompt it hits, because no human is present to answer it. The process stays up, the supervisor reports it alive, and its infrastructure (the session, any browser bridge) all look healthy, but the loop never runs a single cycle and its liveness signal goes stale. The failure is invisible from the outside. "The session is running" and "the session is doing its job" are different claims, and a permission prompt silently breaks the second while leaving the first true.
./q-system/lessons/an-unattended-session-freezes-on-its-first-permission-prompt.md:16:3. Prove liveness by the work, not the process. A heartbeat that stamps on each real wake (before any gate) distinguishes "the process exists" from "the loop is actually cycling." When the heartbeat goes stale while the process is up, look INSIDE the session (capture its pane or state); do not trust the supervisor's "alive." A stale heartbeat behind a live process is the exact fingerprint of a session frozen on a prompt.
./q-system/lessons/monitor-the-capability-you-need-not-a-proxy-for-liveness.md:2:id: monitor-the-capability-you-need-not-a-proxy-for-liveness
./q-system/lessons/monitor-the-capability-you-need-not-a-proxy-for-liveness.md:4:title: Monitor the capability you need, not a proxy for liveness
./q-system/lessons/monitor-the-capability-you-need-not-a-proxy-for-liveness.md:8:A recurring job can keep waking on schedule and still be incapable of doing the one thing it exists to do. When that happens, three latent gaps usually turn a transient hiccup into a silent, prolonged outage. Design against all three.
./q-system/lessons/monitor-the-capability-you-need-not-a-proxy-for-liveness.md:10:1) Monitor the outcome, not the heartbeat. A liveness stamp that fires before the real work proves only that the process woke up. If the health monitor reads that stamp, it will report green through every failed cycle. Instrument the actual capability: stamp success AFTER the load-bearing step, or emit a separate signal the monitor can distinguish from mere wakefulness. Ask of every monitor: can this stay green while the job produces nothing? If yes, it is watching the wrong signal.
./q-system/lessons/monitor-the-capability-you-need-not-a-proxy-for-liveness.md:16:General rule: distinguish 'the job ran' from 'the job succeeded,' and wire monitoring, permissions, and alerting to the second. A cycle that produces nothing must be as loud as a crash.
./q-system/lessons/a-supervised-restart-runs-in-a-stripped-environment-and-auth.md:21:- Downstream, the process often degrades quietly: it does zero work instead of crashing, so uptime and liveness checks stay green.
./plugins/kipi-design/skills/ui-ux-pro-max/src/data/products.csv:48:47,Job Board/Recruitment,"board, job, recruitment",Flat Design + Minimalism,"Vibrant & Block-based, Accessible & Ethical",Conversion-Optimized + Feature-Rich,HR Analytics Dashboard,Professional Blue + Success Green + Neutral,Job listings. Search/filter. Company profiles. Application tracking. Resume upload. Salary insights.
./plugins/kipi-design/skills/ui-ux-pro-max/data/products.csv:48:47,Job Board/Recruitment,"board, job, recruitment",Flat Design + Minimalism,"Vibrant & Block-based, Accessible & Ethical",Conversion-Optimized + Feature-Rich,HR Analytics Dashboard,Professional Blue + Success Green + Neutral,Job listings. Search/filter. Company profiles. Application tracking. Resume upload. Salary insights.

exec
/bin/zsh -lc "rg -n \"dispatch-lastbeat|dispatch\\.log|run_evidence\" kipi-dispatch.sh q-system/.q-system plugins --glob '*.py' --glob '*.sh' --glob '*.json'; git diff --unified=0 origin/main...HEAD -- q-system/.q-system/terminal-states.json; rg -n 'terminal-states.json' q-system/.q-system/scripts q-system/.q-system/test plugins --glob '*.py' --glob '*.sh' 2>/dev/null" in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-281
 exited 2 in 0ms:
kipi-dispatch.sh:45:LOG="$HOME/.config/kipi/dispatch.log"
kipi-dispatch.sh:692:BEAT_FILE="$HOME/.config/kipi/dispatch-lastbeat"
kipi-dispatch.sh:1276:# Every arm below used to end in `say`, which appends to dispatch.log and nothing
q-system/.q-system/scripts/daily-linear-digest.py:58:DISPATCH_LOG = _STATE / "dispatch.log"
q-system/.q-system/scripts/daily-linear-digest.py:147:    # about. A missing dispatch.log only means the dispatcher has not run yet,
q-system/.q-system/scripts/daily-linear-digest.py:197:    # MATCH THE LOG'S OWN CLOCK. `linear-worker.log` and `dispatch.log` prefix every
q-system/.q-system/terminal-states.json:84:        "run_evidence": "~/.config/kipi/linear-dor.out",
q-system/.q-system/terminal-states.json:105:        "run_evidence": "~/.config/kipi/linear-dor.out",
q-system/.q-system/terminal-states.json:125:        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
q-system/.q-system/terminal-states.json:145:        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
q-system/.q-system/terminal-states.json:165:        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
q-system/.q-system/terminal-states.json:185:        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
q-system/.q-system/terminal-states.json:205:        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
q-system/.q-system/terminal-states.json:312:        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
q-system/.q-system/terminal-states.json:407:        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
q-system/.q-system/terminal-states.json:443:        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
q-system/.q-system/terminal-states.json:493:        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
q-system/.q-system/terminal-states.json:523:        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
q-system/.q-system/terminal-states.json:543:        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
q-system/.q-system/terminal-states.json:583:        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
q-system/.q-system/terminal-states.json:613:        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
q-system/.q-system/terminal-states.json:643:        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
q-system/.q-system/terminal-states.json:693:        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
q-system/.q-system/terminal-states.json:713:        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
q-system/.q-system/terminal-states.json:749:        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
q-system/.q-system/terminal-states.json:769:        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
q-system/.q-system/scripts/test/test-ask358-round3-mutation.sh:197:  date -u +%s > "$ROOT/home/.config/kipi/dispatch-lastbeat"
q-system/.q-system/scripts/test/test-ask358-round3-mutation.sh:217:if grep -q "handing $D_ISS back" "$ROOT/home/.config/kipi/dispatch.log" 2>/dev/null; then
q-system/.q-system/scripts/test/test-ask358-round3-mutation.sh:221:      "$(cat "$ROOT/home/.config/kipi/dispatch.log" 2>/dev/null)"
q-system/.q-system/scripts/test/test-dispatch-stale-checkout.sh:6:# from ~/.config/kipi/dispatch.log on 2026-08-01 -- 19 refusing cycles, 9 Slack
q-system/.q-system/scripts/test/test-dispatch-stale-checkout.sh:100:run_check() { ( cd "$1" && LOG="${2:-$CHECKSTATE}/dispatch.log" bash "$HARNESS" 2>&1 ); }
q-system/.q-system/scripts/test/test-dispatch-stale-checkout.sh:247:F1="$( cd "$C9" && KIPI_TEST_PAGE_FAILS=1 LOG="$S9/dispatch.log" bash "$HARNESS" 2>/dev/null | grep -c '^PAGE ' || true )"
q-system/.q-system/scripts/test/test-dispatch-stale-checkout.sh:248:F2="$( cd "$C9" && LOG="$S9/dispatch.log" bash "$HARNESS" 2>/dev/null | grep -c '^PAGE ' || true )"
q-system/.q-system/scripts/test/test-dispatch-stale-checkout.sh:255:( cd "$C9" && KIPI_TEST_PAGE_FAILS=1 LOG="$S9B/dispatch.log" bash "$HARNESS" 2>&1 >/dev/null ) | grep -q 'did NOT go out' \
q-system/.q-system/scripts/test/test-dispatch-stale-checkout.sh:281:FRESH="$( LOG="$PL/dispatch.log" bash "$PH" korph "real fault" 2>/dev/null | grep -c '^PAGE ' || true )"
q-system/.q-system/scripts/test/test-dispatch-stale-checkout.sh:287:AGED="$( LOG="$PL/dispatch.log" bash "$PH" korph "real fault" 2>/dev/null | grep -c '^PAGE ' || true )"
q-system/.q-system/scripts/test/test-dispatch-stale-checkout.sh:295:( LOG="$PC/dispatch.log" bash "$PH" kconc "same line" 2>/dev/null > "$WORK/c1.out" ) &
q-system/.q-system/scripts/test/test-dispatch-stale-checkout.sh:296:( LOG="$PC/dispatch.log" bash "$PH" kconc "same line" 2>/dev/null > "$WORK/c2.out" ) &
q-system/.q-system/scripts/test/test-dispatch-stale-checkout.sh:396:CL="$( LOG="$PR/dispatch.log" bash -c '
q-system/.q-system/scripts/test/test-dispatch-stale-checkout.sh:410:LOG="$PR/dispatch.log" bash -c '
q-system/.q-system/scripts/test/test-dispatch-stale-checkout.sh:693:    ok "a red-CI branch mismatch pages a human, not just dispatch.log"
q-system/.q-system/scripts/test/test-dispatch-stale-checkout.sh:695:    bad "a red-CI branch mismatch pages a human, not just dispatch.log" \
q-system/.q-system/scripts/test/test-dispatch-stale-checkout.sh:739:  # Every arm above ends in `say`, which appends to dispatch.log and nothing
q-system/.q-system/scripts/linear-alert-triage.py:467:def write_run_evidence(line: str) -> None:
q-system/.q-system/scripts/linear-alert-triage.py:514:        write_run_evidence(line)
q-system/.q-system/scripts/linear-alert-triage.py:562:        write_run_evidence("; ".join(done))
q-system/.q-system/scripts/verify-codex-review-live.sh:171:# grep dispatch.log for `engine: codex`. That string is real -- pr-review-agent.sh
q-system/.q-system/scripts/verify-codex-review-live.sh:173:# linear-worker.sh redirects into $STATE_DIR/linear-worker.log. dispatch.log is
q-system/.q-system/scripts/verify-codex-review-live.sh:248:LOG="$STATE_DIR/dispatch.log"
q-system/.q-system/scripts/test/test-terminal-states.sh:359:    ev = os.path.expanduser(lc.get("run_evidence", ""))
q-system/.q-system/scripts/test/test-terminal-states.sh:362:        errors.append(f"{row_id}: liveness_check needs label, run_evidence and max_age_s")
q-system/.q-system/scripts/test/test-terminal-states.sh:853:                         "run_evidence": "'"$FIX"'/ran", "max_age_s": 86400}
q-system/.q-system/scripts/test/test-terminal-states.sh:870:                         "run_evidence": "'"$FIX"'/stale-ran", "max_age_s": 60}
q-system/.q-system/scripts/test/test-terminal-states.sh:883:                         "run_evidence": "'"$FIX"'/ran", "max_age_s": 86400}
q-system/.q-system/scripts/test/test-terminal-states.sh:1115:                         "run_evidence": "'"$FIX"'/ran", "max_age_s": 86400}
q-system/.q-system/scripts/test/test-dispatch-liveness.sh:121:dlog() { cat "$ROOT/home/.config/kipi/dispatch.log" 2>/dev/null; }
q-system/.q-system/scripts/test/test-dispatch-liveness.sh:128:  date -u +%s > "$ROOT/home/.config/kipi/dispatch-lastbeat"
q-system/.q-system/scripts/test/test-dispatch-liveness.sh:157:# the silent failure again, one layer down: nobody reads dispatch.log.
q-system/.q-system/scripts/test/test-dispatch-liveness.sh:324:  : > "$ROOT/home/.config/kipi/dispatch.log"
q-system/.q-system/scripts/test/test-dispatch-client-refusal-after-hold.sh:176:DLOG="$WORK/home/.config/kipi/dispatch.log"
q-system/.q-system/scripts/test/test_linear_alert_triage.py:608:        orig_load, orig_ev = triage._load, triage.write_run_evidence
q-system/.q-system/scripts/test/test_linear_alert_triage.py:610:        triage.write_run_evidence = lambda line: None
q-system/.q-system/scripts/test/test_linear_alert_triage.py:618:            triage._load, triage.write_run_evidence = orig_load, orig_ev
q-system/.q-system/scripts/test/test-ci-redrive.sh:587:dlog() { cat "$DROOT/home/.config/kipi/dispatch.log" 2>/dev/null; }
q-system/.q-system/scripts/test/test-ci-redrive.sh:594:  date -u +%s > "$DROOT/home/.config/kipi/dispatch-lastbeat"
q-system/.q-system/scripts/test/test-ci-redrive.sh:690:  ok "14o a red-CI park reaches a human, not just dispatch.log"
q-system/.q-system/scripts/test/test-ci-redrive.sh:692:  bad "14o a red-CI park reaches a human, not just dispatch.log" "$(cat "$D_PAGES" 2>/dev/null)"
diff --git a/q-system/.q-system/terminal-states.json b/q-system/.q-system/terminal-states.json
index 9cb18f15..85c3d7b3 100644
--- a/q-system/.q-system/terminal-states.json
+++ b/q-system/.q-system/terminal-states.json
@@ -125 +125 @@
-        "run_evidence": "~/.config/kipi/dispatch.log",
+        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
@@ -145 +145 @@
-        "run_evidence": "~/.config/kipi/dispatch.log",
+        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
@@ -165 +165 @@
-        "run_evidence": "~/.config/kipi/dispatch.log",
+        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
@@ -185 +185 @@
-        "run_evidence": "~/.config/kipi/dispatch.log",
+        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
@@ -205 +205 @@
-        "run_evidence": "~/.config/kipi/dispatch.log",
+        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
@@ -312 +312 @@
-        "run_evidence": "~/.config/kipi/dispatch.log",
+        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
@@ -407 +407 @@
-        "run_evidence": "~/.config/kipi/dispatch.log",
+        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
@@ -443 +443 @@
-        "run_evidence": "~/.config/kipi/dispatch.log",
+        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
@@ -493 +493 @@
-        "run_evidence": "~/.config/kipi/dispatch.log",
+        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
@@ -523 +523 @@
-        "run_evidence": "~/.config/kipi/dispatch.log",
+        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
@@ -543 +543 @@
-        "run_evidence": "~/.config/kipi/dispatch.log",
+        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
@@ -583 +583 @@
-        "run_evidence": "~/.config/kipi/dispatch.log",
+        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
@@ -613 +613 @@
-        "run_evidence": "~/.config/kipi/dispatch.log",
+        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
@@ -643 +643 @@
-        "run_evidence": "~/.config/kipi/dispatch.log",
+        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
@@ -693 +693 @@
-        "run_evidence": "~/.config/kipi/dispatch.log",
+        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
@@ -713 +713 @@
-        "run_evidence": "~/.config/kipi/dispatch.log",
+        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
@@ -749 +749 @@
-        "run_evidence": "~/.config/kipi/dispatch.log",
+        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
@@ -769 +769 @@
-        "run_evidence": "~/.config/kipi/dispatch.log",
+        "run_evidence": "~/.config/kipi/dispatch-lastbeat",
q-system/.q-system/scripts/linear-alert-triage.py:14:anything that moves a ticket OUT of that bucket. terminal-states.json registered
q-system/.q-system/scripts/linear-alert-triage.py:468:    """The liveness artifact terminal-states.json points at.
q-system/.q-system/scripts/test/test-terminal-states.sh:20:# Pairs with q-system/.q-system/terminal-states.json (Piece B,
q-system/.q-system/scripts/test/test-terminal-states.sh:83:"""Validate terminal-states.json against every declared driver, from source.
q-system/.q-system/scripts/test/test-terminal-states.sh:737:                    "terminal-states.json naming a consumer + liveness_check + reentry, "
q-system/.q-system/scripts/test/test-terminal-states.sh:771:REG="$ROOT/q-system/.q-system/terminal-states.json"

exec
/bin/zsh -lc "sed -n '650,730p' kipi-dispatch.sh; sed -n '330,390p' q-system/.q-system/scripts/test/test-terminal-states.sh; sed -n '720,790p' q-system/.q-system/scripts/test/test-terminal-states.sh; sed -n '820,910p' q-system/.q-system/scripts/test/test-terminal-states.sh; git show --stat --oneline HEAD; git show --format=fuller --no-patch HEAD; git show HEAD"'^:q-system/.q-system/terminal-states.json | python3 -c '"'import json,sys; d=json.load(sys.stdin); print(sum(1 for x in d[\"states\"] if x.get(\"liveness_check\",{}).get(\"label\")==\"com.kipi.dispatch\"))'" in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-281
 succeeded in 0ms:
    # reading that line cannot tell a client-repo refusal from an expired token
    # from an empty queue -- and "silent refusal is indistinguishable from nothing
    # to do" is the exact failure this issue exists to remove. The reason the gate
    # gives is the only thing that makes the gate legible.
    if out="$(bash "$PREFLIGHT" "$path" "$remote" 2>&1)"; then
      printf '%s\t%s\n' "$name" "$path"
    else
      say "preflight REFUSED $name ($path); not entering it -- $(printf '%s' "$out" | grep '^FAIL' | tr '\n' ';' | sed 's/;$//')"
      case "$out" in
        *"FAIL client-repo:"*)
          client_n=$((client_n + 1))
          client_names="${client_names:+$client_names, }$name" ;;
      esac
    fi
  done < <(rotation)
  # COUNTED, AND IN THE DIGEST'S OWN SHAPE. daily-linear-digest.py's third section
  # ("tried, could not be worked") scrapes this log for `N thing(s) <what>` lines,
  # so a refusal phrased any other way is invisible in the one surface the founder
  # actually reads once a day. Emitted only when it happened: a client repo being
  # refused is a real event, and a "0 refused" line every 15 minutes is the
  # cry-wolf noise that gets a channel muted.
  [ "$client_n" -gt 0 ] && say "$client_n repo(s) REFUSED as client engagement repos (unattended dispatch is not allowed there, even when opted in): $client_names"
  return 0
}

# --- LIVENESS BEACON: page when the heartbeat COMES BACK ------------------
# Founder ask 2026-07-28: "I want to get a slack notification that the heartbeat
# restarted when it does."
#
# The signal is the TRANSITION (was gone -> is back), never the level. This runs
# every 900s, so paging per tick would be 96 pings a day -- the cry-wolf failure
# that trains someone to mute the channel and costs the real alerts their job.
#
# Placed BEFORE every early exit on purpose. Most ticks legitimately skip (cap
# reached, nothing ready), and a skip is still proof of life. Recording the beat
# only on a dispatch would make a healthy-but-idle loop look dead, and would fire
# a false "resumed" ping on the next dispatch.
#
# A gap larger than GAP_MINUTES means it was not running: reboot, a manual
# unload/load, a crash the launchd watchdog restarted, or the Mac asleep. All
# four are worth one line.
GAP_MINUTES="${KIPI_DISPATCH_GAP_MINUTES:-45}"   # 3 missed ticks at 900s
BEAT_FILE="$HOME/.config/kipi/dispatch-lastbeat"
NOW_EPOCH="$(date -u +%s)"
LAST_BEAT="$(cat "$BEAT_FILE" 2>/dev/null || echo "")"
case "$LAST_BEAT" in ''|*[!0-9]*) LAST_BEAT="" ;; esac

if [ -z "$LAST_BEAT" ]; then
  say "heartbeat: first beat on record"
  page "kipi heartbeat: STARTED. The Linear loop is live and will check for ready issues every 15 min (max ${KIPI_DISPATCH_DAILY_MAX:-4} issues/day). Nothing to do."
else
  GAP=$(( (NOW_EPOCH - LAST_BEAT) / 60 ))
  if [ "$GAP" -ge "$GAP_MINUTES" ]; then
    say "heartbeat: RESUMED after ${GAP}m without a beat"
    page "kipi heartbeat: RESUMED after ${GAP} min down (reboot, sleep, or a reload). The Linear loop is running again. Nothing to do -- this is the all-clear, not a fault."
  fi
fi
printf '%s' "$NOW_EPOCH" > "$BEAT_FILE"

LIVE="$(live_converges)"; LIVE="${LIVE:-0}"
if [ "$LIVE" -ge "$MAX_CONCURRENT" ]; then
  say "skip: $LIVE converge run(s) live, cap $MAX_CONCURRENT"
  exit 0
fi

# --- DAILY BUDGET (loop-exits.md exit 3) ---------------------------------
# The concurrency cap bounds how many run AT ONCE. It does NOT bound how many
# run IN A DAY -- at ~1 issue/hour that is ~24 issues and ~144 `claude -p`
# sessions overnight, against a subscription with a real weekly ceiling.
# Measured 2026-07-28: one interactive night spawned 89 sessions and 44 reviewer
# runs. An unbounded heartbeat is a runaway-bill loop, which is exactly the
# thing loop-exits.md says an autonomous loop must not be.
#
# One issue costs up to MAX_ROUNDS x (1 agent + 1 reviewer) sessions. Do NOT read
# that as a fixed 6: the code default is 3 rounds, but the LOADED plist sets
# KIPI_DISPATCH_ROUNDS=4, so the live cost is up to 8 sessions per issue. The
# older comment here hardcoded 6 and quietly understated the running job by a
# third. Compute it from MAX_ROUNDS, never from a remembered number.
#
# THIS IS NOT A MONEY DIAL (founder correction, 2026-07-29). It caps SESSIONS and
# BLAST RADIUS, not dollars: how many issues per day may enter a loop that merges
    Bounded to BLOCK_LOOKBACK non-comment lines: an unbounded search would let a
    marker hundreds of lines up claim an unrelated exit. Comment lines are
    skipped because every one of these markers is also DISCUSSED in prose here,
    and a row must be anchored to code that runs, not to a sentence about it.
    """
    best, best_at, tie = None, -1, False
    seen = 0
    for idx in range(site_line - 1, -1, -1):
        if lines[idx] is None or is_comment(lines[idx]):
            continue
        seen += 1
        if seen > BLOCK_LOOKBACK:
            break
        for row in rows:
            if any(m in lines[idx] for m in row["_markers"]):
                if idx > best_at:
                    best, best_at, tie = row, idx, False
                elif idx == best_at and row is not best:
                    tie = True
    return best, tie


def check_liveness(lc, errors, row_id):
    """Prove the consumer RAN. A plist on disk is not evidence (finding-14)."""
    kind = lc.get("kind")
    if kind != "launchd":
        errors.append(f"{row_id}: unknown liveness_check kind {kind!r}")
        return
    label = lc.get("label", "")
    ev = os.path.expanduser(lc.get("run_evidence", ""))
    max_age = lc.get("max_age_s")
    if not label or not ev or not isinstance(max_age, int):
        errors.append(f"{row_id}: liveness_check needs label, run_evidence and max_age_s")
        return

    agents = os.path.expanduser(
        os.environ.get("TERMINAL_STATES_LAUNCHD_DIR", "~/Library/LaunchAgents"))
    plist = os.path.join(agents, label + ".plist")
    if not os.path.isfile(plist):
        # Never installed on THIS machine. This script propagates fleet-wide via
        # kipi update to instances that never had the job, and failing there
        # would make the gate a nuisance everyone learns to skip. An installed
        # plist is what turns liveness into a claim -- see the branch below.
        print(f"    note: {row_id}: {label} is not installed here; liveness not asserted")
        return

    lcbin = os.environ.get("TERMINAL_STATES_LAUNCHCTL", "launchctl")
    # A plist EXISTS at this point, so this host DOES own the job; only the
    # ability to read its state is missing. That is a gap in the evidence, not a
    # pass. For one round this comment said exactly that while the code below
    # printed a note and returned clean, so pointing TERMINAL_STATES_LAUNCHCTL at
    # a nonexistent binary certified every consumer with zero evidence and the
    # suite still reported 12 passed (codex-adversarial finding-4). The
    # host-never-had-the-job case already returned at the plist check above, so
    # failing closed here costs the fleet nothing.
    if not (os.path.isfile(lcbin) or shutil.which(lcbin)):
        errors.append(f"{row_id}: {label} has a plist at {plist} but {lcbin} is not "
                      "available to read its state -- liveness cannot be asserted, "
                      "and an unverifiable consumer is not a proven live one")
        return
    try:
        # heredoc bodies are masked for matching too, for the same reason they are
        # masked for enumeration: a marker found inside embedded python is not the
        # shell branch that runs.
        mlines = mask_heredocs(lines) if "toplevel-exit" in src_shapes[sid] else lines
        for line_no, kind, text in sites:
            row, tie = match_site(line_no, mlines, rows_here)
            if debug:
                # The site table. A RED here is usually a marker claiming a site
                # it does not own, and reading that off the source by hand is the
                # slow way to find it.
                print(f"    [{sid}] site {line_no:>5} {kind:<18} -> "
                      f"{(row or {}).get('id', 'UNMATCHED')}   | {text[:60]}")
            if row is None:
                errors.append(
                    f"UNREGISTERED EXIT in source {sid} at line {line_no} ({kind})\n"
                    f"      {text}\n"
                    "      No registry row's marker appears above it. Add a row to "
                    "terminal-states.json naming a consumer + liveness_check + reentry, "
                    "or terminal:true with a rationale.")
                continue
            if tie:
                errors.append(f"AMBIGUOUS EXIT in source {sid} at line {line_no}: "
                              "two markers tie for nearest. Make one more specific.")
                continue
            observed[row["id"]] += 1

    for row in good_rows:
        got, want = observed[row["id"]], row.get("sites")
        if isinstance(want, int) and got != want:
            if got == 0:
                errors.append(f"{row['id']}: marker(s) {row['_markers']!r} match NO "
                              f"exit in source {row['source']}. Stale row, or the "
                              "marker moved.")
            else:
                errors.append(f"{row['id']}: covers {got} exit site(s), registry "
                              f"declares {want}. An exit was added or removed next to "
                              f"marker(s) {row['_markers']!r} -- reconcile the row.")

    print(f"    enumerated {total_sites} exit site(s) across {len(src_lines)} driver(s); "
          f"{len(rows)} registry row(s)")
    if errors:
        for e in errors:
            print(f"    RED: {e}")
        return 1
    return 0


if __name__ == "__main__":
    sys.exit(main())
PYEOF

REG="$ROOT/q-system/.q-system/terminal-states.json"
SRC="$ROOT/q-system/.q-system/scripts/linear-worker.sh"
CONVERGE="$ROOT/q-system/.q-system/scripts/converge.sh"
DISPATCH="$ROOT/kipi-dispatch.sh"
# validate.py takes the manifest as a FILE argument, and that contract is
# unchanged. The manifest itself is a per-declaration fragment directory now
# (the single array it replaced was the merge conflict in 37 of 41 conflicting
# PRs), so assemble it once here and hand validate.py a plain file.
MAN="$WORK/capability-manifest.assembled.json"
python3 "$ROOT/q-system/.q-system/scripts/capability_manifest.py" \
  --root "$ROOT" --print >"$MAN" || { echo "RED: manifest does not assemble"; exit 1; }

for f in "$REG" "$SRC" "$CONVERGE" "$DISPATCH" "$MAN"; do
  if [ ! -f "$f" ]; then echo "RED: missing $f"; exit 1; fi
done

echo "== terminal-states validator =="

# --- 1. the real thing must be green ----------------------------------------
if python3 "$WORK/validate.py" "$REG" "$ROOT" "$MAN"; then
def source(sid):
    return [s for s in d["sources"] if s["id"] == sid][0]
def state(rid):
    return [r for r in d["states"] if r["id"] == rid][0]
exec(sys.argv[3])
json.dump(d, open(sys.argv[2], "w"), indent=2)
EOF
}

expect_red() {  # expect_red <name> <registry> <grep-pattern> [env...]
  local name="$1" reg="$2" pat="$3"; shift 3
  local out rc
  out="$(env "$@" python3 "$WORK/validate.py" "$reg" "$ROOT" "$MAN" 2>&1)"; rc=$?
  if [ "$rc" -eq 0 ]; then
    bad "$name -- validator returned GREEN on a case it must refuse"
    return
  fi
  if printf '%s' "$out" | grep -q "$pat"; then
    ok "$name (RED, and names it)"
  else
    bad "$name -- went RED but did not name the defect (wanted /$pat/)"
    printf '%s\n' "$out" | sed 's/^/        /'
  fi
}

# --- 2. NEGATIVE SELF-TEST (a): the only actor is the founder ---------------
# The bypass_check's first half. This is the shape every parked state in this
# loop already has, so a validator that accepts it certifies the bug.
mkfixture "$FIX/founder.json" '
row = state("attempts-cap-stuck")
row.pop("terminal", None); row.pop("rationale", None); row.pop("evidence", None)
row["consumer"] = "the founder reviews the issue and decides what to do next"
row["liveness_check"] = {"kind": "launchd", "label": "com.kipi.fixture-job",
                         "run_evidence": "'"$FIX"'/ran", "max_age_s": 86400}
row["reentry"] = {"path": "q-system/.q-system/scripts/linear-worker.sh",
                  "marker": "stuck_paged"}
'
touch "$FIX/ran"
expect_red "negative (a): a row whose only actor is the founder" \
  "$FIX/founder.json" "consumer names a HUMAN" \
  "TERMINAL_STATES_LAUNCHD_DIR=$FIX/agents" \
  "TERMINAL_STATES_LAUNCHCTL=$FIX/launchctl-loaded"

# --- 3. NEGATIVE SELF-TEST (b): consumer has not run inside its interval ----
# The bypass_check's second half, and finding-14 exactly: the plist is present
# AND the job is loaded. Only the absence of a RUN makes this red.
mkfixture "$FIX/stale.json" '
row = state("needs-scope")
row.pop("consumer_test")
row["liveness_check"] = {"kind": "launchd", "label": "com.kipi.fixture-job",
                         "run_evidence": "'"$FIX"'/stale-ran", "max_age_s": 60}
'
touch -t 202601010000 "$FIX/stale-ran"
expect_red "negative (b): consumer is loaded but has not run inside its interval" \
  "$FIX/stale.json" "has not run in" \
  "TERMINAL_STATES_LAUNCHD_DIR=$FIX/agents" \
  "TERMINAL_STATES_LAUNCHCTL=$FIX/launchctl-loaded"

# --- 4. a plist on disk while the job is unloaded (finding-14, literally) ----
mkfixture "$FIX/unloaded.json" '
row = state("needs-scope")
row.pop("consumer_test")
row["liveness_check"] = {"kind": "launchd", "label": "com.kipi.fixture-job",
                         "run_evidence": "'"$FIX"'/ran", "max_age_s": 86400}
'
expect_red "a present plist whose job is UNLOADED fails (path existence would pass)" \
  "$FIX/unloaded.json" "NOT LOADED" \
  "TERMINAL_STATES_LAUNCHD_DIR=$FIX/agents" \
  "TERMINAL_STATES_LAUNCHCTL=$FIX/launchctl-unloaded"

# --- 5. MUTATION: delete a row -> the check must go red ----------------------
# If either of these stays green the enumeration is decorative: the registry
# would be a document rather than a gate.
#
# TWO CASES, because a deleted row fails in two different ways and asserting
# only the first would have let the second regress silently. When the orphaned
# site has no other marker above it, it is UNREGISTERED. When a neighbouring
# row's marker is within lookback, the site is ADOPTED by that neighbour and the
# only thing that catches it is that neighbour's `sites` count -- which is the
# whole reason the count field exists.
mkfixture "$FIX/dropped.json" '
d["states"] = [r for r in d["states"] if r["id"] != "attempts-cap-stuck"]
'
expect_red "mutation: deleting a row whose site stands alone -> UNREGISTERED" \
  "$FIX/dropped.json" "UNREGISTERED EXIT"

mkfixture "$FIX/dropped2.json" '
d["states"] = [r for r in d["states"] if r["id"] != "claim-contended"]
'
expect_red "mutation: deleting a row whose site is adopted by a neighbour -> count mismatch" \
  "$FIX/dropped2.json" "covers 2 exit site"
0941aa8d fix(terminal-states): liveness evidence is that the job RAN, not that work happened (ASK-1146)
 q-system/.q-system/terminal-states.json | 36 ++++++++++++++++-----------------
 1 file changed, 18 insertions(+), 18 deletions(-)
commit 0941aa8df56812059c35320eefe37330e78c414c
AuthorDate: Sat Aug 29 23:21:27 2026 -0700
CommitDate: Sat Aug 29 23:21:27 2026 -0700

    fix(terminal-states): liveness evidence is that the job RAN, not that work happened (ASK-1146)
    
    main is red right now, and has been since roughly 17:18 local. Measured on a
    clean checkout of origin/main, not inferred: `test-terminal-states.sh` reports
    37 passed / 2 failed, and the capability gate is RED because of it. Both
    failures were also present on the rebased PR #272 and PR #265 trees, which is
    what made it clear the base was the problem rather than either PR.
    
    The verdict was wrong. It read:
    
      com.kipi.dispatch is loaded but has not run in 21677s (interval allows 3600s)
      -- a loaded job that stopped running is a dead consumer
    
    The dispatcher was alive the whole time. Its heartbeat file was 30 seconds old
    when this was measured. What it had actually done was this, from its own log:
    
      DAILY CAP: 10/10 issues dispatched for budget day 2026-08-29
      (lane=production), stopping until 7:00 local
    
    Capped and idle by design. The liveness check keyed on `~/.config/kipi/dispatch.log`,
    which only moves when the dispatcher DOES something, so a correctly-capped job
    looks identical to a dead one for up to fourteen hours every single night. A
    gate that goes red every evening for a healthy system is a gate someone switches
    off, and then it protects nothing.
    
    `~/.config/kipi/dispatch-lastbeat` is the right evidence and already exists.
    kipi-dispatch.sh writes it unconditionally at the top of every run, BEFORE the
    concurrency skip and before the daily-cap exit, so it answers the question the
    check is actually asking. 18 rows repointed.
    
    Mutation-tested rather than trusted: pointing the rows back at dispatch.log puts
    both failures back (39 passed -> 37 passed, 2 failed), and the green result was
    confirmed twice for stability rather than once.
    
    Worth naming, because it is the whole shape: nothing here was a missing check.
    The check ran, produced a confident verdict, and could not see the designed
    state it was misreading.
18

codex
The core change is narrow: 18 registry rows now use the dispatcher’s unconditional heartbeat instead of an activity log. I’m checking two adversarial failure modes next: whether any dispatcher exit occurs before the heartbeat write, and whether any changed row names a different launchd consumer. The full shell suite cannot create its temporary directory under this read-only sandbox, so I’m using read-only executable checks and will not promote any unexecuted suspicion to a finding.
exec
import json
p='q-system/.q-system/terminal-states.json'
d=json.load(open(p))
rows=[r for r in d['states'] if r.get('liveness_check',{}).get('run_evidence')=='~/.config/kipi/dispatch-lastbeat']
print('heartbeat_rows',len(rows))
print('wrong_label',[(r['id'],r['liveness_check']['label']) for r in rows if r['liveness_check'].get('label')"'!='"'com.kipi.dispatch'])
print('bad_age',[(r['id'],r['liveness_check']['max_age_s']) for r in rows if r['liveness_check'].get('max_age_s')"'!=3600])
print('"'ids',','.join(r['id'] for r in rows))
PY
awk 'NR<704 && /"'^[[:space:]]*(exit|return)([[:space:]]|$)/ {print NR ":" $0}'"' kipi-dispatch.sh
python3 - <<'PY'
from pathlib import Path
s=Path('kipi-dispatch.sh').read_text().splitlines()
beat=next(i+1 for i,x in enumerate(s) if \"printf '%s' \\\""'$NOW_EPOCH'"\\\" > \\\""'$BEAT_FILE'"\\\"\" in x)
exits=[(i+1,x.strip()) for i,x in enumerate(s) if (x.strip().startswith('exit ') or x.strip()=='exit')]
print('beat_line',beat)
print('top_level_exit_candidates_before_beat',[(n,x) for n,x in exits if n<beat])
PY
python3 -m json.tool q-system/.q-system/terminal-states.json >/dev/null && echo 'json_parse=PASS'
git diff --check origin/main...HEAD" in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-281
 succeeded in 0ms:
130:  return 1
152:    return 0
190:    return 0
205:      return 0
219:  exit 1
270:    return 0
274:    return 0
313:  return 1
451:    return 1
541:  return 0
672:  return 0
zsh:12: can't create temp file for here document: operation not permitted
json_parse=PASS

exec
/bin/zsh -lc "sed -n '200,228p' kipi-dispatch.sh; python3 -c 'import json; d=json.load(open(\"q-system/.q-system/terminal-states.json\")); rows=[r for r in d[\"states\"] if r.get(\"liveness_check\",{}).get(\"run_evidence\")==\"~/.config/kipi/dispatch-lastbeat\"]; print(\"heartbeat_rows\",len(rows)); print(\"wrong_label\",[(r[\"id\"],r[\"liveness_check\"][\"label\"]) for r in rows if r[\"liveness_check\"].get(\"label\")"'!="com.kipi.dispatch"]); print("bad_age",[(r["id"],r["liveness_check"]["max_age_s"]) for r in rows if r["liveness_check"].get("max_age_s")!=3600]); print("ids", ",".join(r["id"] for r in rows))'"'; python3 -c 'from pathlib import Path; s=Path(\"kipi-dispatch.sh\").read_text().splitlines(); beat=next(i+1 for i,x in enumerate(s) if \"NOW_EPOCH\" in x and \"BEAT_FILE\" in x and \"printf\" in x); exits=[(i+1,x.strip()) for i,x in enumerate(s) if x.strip().startswith(\"exit \") or x.strip()==\"exit\"]; print(\"beat_line\",beat); print(\"exit_candidates_before_beat\",[(n,x) for n,x in exits if n<beat])'; git diff --check origin/main...HEAD" in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-281
 succeeded in 0ms:
    stamp="$(sed -n 2p "$mark" 2>/dev/null)"
    case "$stamp" in ''|*[!0-9]*) stamp=0 ;; esac
    if [ "$prev" = "$hash" ] && [ "$(( now - stamp ))" -lt "$PAGE_REPING_SECONDS" ]; then
      say "page suppressed ($key unchanged for $(( (now - stamp) / 60 ))m; re-pings after $(( PAGE_REPING_SECONDS / 3600 ))h)"
      rmdir "$lock" 2>/dev/null || true
      return 0
    fi
  fi
  if page_ok "$msg"; then
    printf '%s\n%s\n' "$hash" "$now" > "$mark" 2>/dev/null || true
  else
    say "page: $key did NOT go out; leaving the marker unset so the next heartbeat retries it"
  fi
  rmdir "$lock" 2>/dev/null || true
}

cd "$REPO" 2>/dev/null || {
  say "FATAL: repo not found at $REPO"
  page_once repo-missing "kipi dispatch: repo not found at $REPO -- the Linear loop is DEAD. Do: check the path in com.kipi.dispatch.plist."
  exit 1
}
page_clear repo-missing

# --- STALE-CHECKOUT REFUSAL (sp-c775b116) --------------------------------
# The loop runs the founder's WORKING TREE, and nothing kept it in sync with
# main. There is no `git pull` anywhere in this script. Observed 2026-07-30:
# merging PR #34 left this checkout at 1597eaf, so the loop would have gone on
# running the old Claude-only reviewer indefinitely while main carried the codex
# gate. It was fixed by hand twice in one session, which means every future merge
heartbeat_rows 18
wrong_label []
bad_age []
ids automerge-unarmed,no-review-verdict,worktree-create-failed,claim-contended,claim-infra-failure,drift-cap,converge-no-pr-opened,converge-review-no-verdict,converge-interrupted,dispatch-stale-checkout,dispatch-concurrency-cap,dispatch-daily-cap,dispatch-turn-lock-held,dispatch-no-repo-this-cycle,dispatch-nothing-ready,dispatch-converge-already-live,dispatch-ci-redrive-claimed,dispatch-review-redrive-claimed
beat_line 707
exit_candidates_before_beat [(219, 'exit 1')]

codex
No finding survived reproduction.

**What is sound**

- All 18 changed rows belong to `com.kipi.dispatch`.
  - Output: `heartbeat_rows 18`
  - Output: `wrong_label []`
  - Output: `bad_age []`
- The heartbeat is written before normal dispatch exits, including concurrency and daily-cap exits.
- The only earlier top-level exit is the fatal missing-repository path. Leaving the heartbeat stale there correctly exposes a dead consumer.
- `terminal-states.json` parses successfully.
  - Output: `json_parse=PASS`
- `git diff --check origin/main...HEAD` returned clean.
- The full terminal-states suite could not run because the read-only sandbox refused `mktemp`. I did not treat that environment restriction as a code finding.

**VERDICT: APPROVE**

Most important thing to fix first: nothing. The change replaces activity evidence with actual scheduler-run evidence without hiding fatal startup failure.

FINDINGS:
END FINDINGS
tokens used
49,873
No finding survived reproduction.

**What is sound**

- All 18 changed rows belong to `com.kipi.dispatch`.
  - Output: `heartbeat_rows 18`
  - Output: `wrong_label []`
  - Output: `bad_age []`
- The heartbeat is written before normal dispatch exits, including concurrency and daily-cap exits.
- The only earlier top-level exit is the fatal missing-repository path. Leaving the heartbeat stale there correctly exposes a dead consumer.
- `terminal-states.json` parses successfully.
  - Output: `json_parse=PASS`
- `git diff --check origin/main...HEAD` returned clean.
- The full terminal-states suite could not run because the read-only sandbox refused `mktemp`. I did not treat that environment restriction as a code finding.

**VERDICT: APPROVE**

Most important thing to fix first: nothing. The change replaces activity evidence with actual scheduler-run evidence without hiding fatal startup failure.

FINDINGS:
END FINDINGS

PR #286, opened 2026-08-30
fix(capability-gate): the silent-absence gate could not see .q-system/tests

assafkip:
fix(capability-gate): the silent-absence gate could not see .q-system/tests

Ten test files on main are declared nowhere and reached by no runner. 174
assertions, all green, executed by nothing. They are found by running them, not
by reading anything: nine pass under pytest, one under python3, and no CI step
or manifest entry names any of them.

The gate that exists to catch exactly this could not see them. `discover_tests`
walked `q-system/.q-system/scripts` recursively and `q-system/.q-system` with
`glob` instead of `rglob`, on the stated reasoning that scripts/ would otherwise
be scanned twice. It is a set; a second visit costs a directory walk and changes
nothing. What the non-recursive scan actually bought was a hole: every
subdirectory of .q-system except scripts/ was invisible to the
present-but-undeclared direction of the diff.

The gate's own negative proof missed it for a reason worth naming. F1 plants its
undeclared file in `scripts/` -- the one path discovery already walked -- so the
proof exercised the case that worked and never asked about the sibling
directories. F1b is that same case one directory over. Reverting `rglob` to
`glob` fails F1b and leaves F1 green, which is the demonstration that F1 alone
was blind.

`in_scan_scope` is widened to match, because when the two disagree a declared
path falls into neither direction: not existence-checked (out of scope) and its
file not checked for a declaration (not discovered). A declaration that means
nothing, reading as fine.

Recursive discovery surfaces 16 undeclared files: 15 separation tests, run
wholesale by validate-separation.py, and test_auto_commit.py, named directly in
CI. All are declared here with the runner they actually pass under. Plus the ten
above: 26 declarations, and the gate is GREEN with 222.

SECOND HALF, because the gate only scopes q-system/.q-system.
`reachability-check.py` asks the wider question across the whole repo: is each
test file reached by ANY runner -- a manifest declaration, a CI pytest path, or
the separation harness? 53 are not.

Failing on all 53 would paint it red on its own population the day it ships, and
a gate that is red by default gets switched off (the same reason 39 of 57 plans
are grandfathered in plan-lint). So the set is FROZEN in
test-reachability-baseline.json and may only shrink: a new unreached test fails,
and a baselined file that becomes reached ALSO fails, demanding its removal, so
the file cannot rot into names nobody prunes. `q-system/.q-system/tests` left
the baseline entirely by being declared, which is the ratchet running in the
direction it is meant to.

Comment lines confer no coverage. verify.yml carries a commented-out `pytest
q-system/.q-system/tests` today, and counting it would be the same
confident-wrong-answer the rest of this change is about.

Six cases, four of which require a RED. All three mutants killed by the specific
case covering their path: disarming new-detection fails two, disarming
baseline-rot detection fails the other two, counting comments as coverage fails
the comment case.

Wired as its own CI step, which the capability gate then demanded before it would
pass the file at all.


assafkip:
Parked as draft, not abandoned.

The `.q-system/tests` discovery blind spot in here is a real defect: ten test files on main were executed by nothing that gates a merge, 174 green assertions, and the gate whose job is catching exactly that could not see them because its scan of `.q-system` was non-recursive.

But this PR also carries a NEW fleet-wide ratchet (`reachability-check.py` plus a 53-entry frozen baseline), and adding a fleet gate in the middle of a drift correction is the re-expansion this session was pulled back for. Parking the whole thing is the honest call rather than landing the gate because the bug fix rode along with it.

Re-cut later as two changes: the recursive-discovery fix and its F1b negative-proof case alone, then the ratchet on its own merits if it still earns its place.

Note for whoever picks it up: four of the 53 baselined paths are voiceloop suites owned by another session. If they get declared before this lands, the baseline needs those four lines removed in the same change.

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-286-20260830-123122.md` (7869 bytes).

```
FINDINGS:
major|reachability-check.py ignores .verify-suites, so 20 of 53 baselined files are labelled "executed by no gate" while the required verify workflow runs them, and a new test in those suites false-fails CI|q-system/.q-system/scripts/reachability-check.py:83
minor|--write-baseline lets the "may only SHRINK" ratchet be hand-cleared and grown; the shrink-only property is prose, not code|q-system/.q-system/scripts/reachability-check.py:142
minor|tracked_tests matches only test_*.py, so 19 tracked shell tests outside the capability scan roots are invisible to a check documented as covering every test file|q-system/.q-system/scripts/reachability-check.py:63
END FINDINGS
```

--- reviewer output ---

## Review: PR #286

Read the PR, the diff, `capability-gate.py`, `reachability-check.py`, the new test file, all three workflows, `.verify-suites`, and `verify.sh`. Ran repros in temp clones. Never wrote to the repo.

---

### FINDING 1 (major): the new check cannot see `.verify-suites`, so 20 of its 53 baselined files ARE executed by a required CI gate

**File:** `q-system/.q-system/scripts/reachability-check.py:83` (`ci_targets`), and the resulting `q-system/.q-system/test-reachability-baseline.json:5`.

`ci_targets` measures coverage by grepping `.github/workflows/` only. This repo's actual pytest manifest is `.verify-suites` at the root, read by `verify.sh`, which `verify.yml` invokes on every pull_request and push to main. The check is structurally blind to it, which is the same class of blindness the PR was written to fix.

Consequence one: the frozen baseline says "Test files that NO GATE EXECUTES" about 20 files a gate executes. The module docstring's specific claim, "the four voiceloop suites collect 84 tests and gate nothing", is contradicted by this PR's own CI run.

Consequence two: adding a test to any of those suites false-fails the new CI step.

Reproducer A, this PR's own `verify` run (33330563160):

```
gh -R assafkip/kipi-system run view 33330563160 --log | grep -iE "pytest:"
```

```
pytest:plugins/prd-os        ok
pytest:plugins/kipi-dsse     ok
pytest:plugins/kipi-core/voiceloop ok
pytest:automation            ok
pytest:rescued/memory-lifecycle ok
pytest:q-system/.q-system/tests ok
pytest:test_destructive_op_deny_anchor.py ok
pytest:test_fleet_unblock.py ok
```

Reproducer B, overlap count against the baseline:

```
python3 -c "import json; b=json.load(open('q-system/.q-system/test-reachability-baseline.json'))['unreached']; s=['plugins/prd-os','plugins/kipi-dsse','plugins/kipi-core/voiceloop','automation','rescued/memory-lifecycle','q-system/.q-system/tests','test_destructive_op_deny_anchor.py','test_fleet_unblock.py']; h=[p for p in b if any(p==x or p.startswith(x+'/') for x in s)]; print(len(h),'of',len(b)); [print(' ',p) for p in h]"
```

```
baselined entries actually executed by verify.sh suites: 20 of 53
   automation/test_voice_refresh.py
   ... (3 automation, 4 voiceloop, 8 kipi-dsse, 3 memory-lifecycle, 2 root files)
```

Reproducer C, the false alarm. Clone at HEAD, add one test to a suite `verify.sh` runs:

```
git clone --shared . $TMP/repo && add plugins/kipi-dsse/tests/test_zz_new_case.py && commit
python3 q-system/.q-system/scripts/reachability-check.py --repo-root $TMP/repo
```

```
rc = 2
reachability-check: FAIL -- new test file that no gate executes:
  plugins/kipi-dsse/tests/test_zz_new_case.py
  Declare it in q-system/.q-system/capability/expected_tests ...
```

`verify` runs that exact directory. The remedy the message prints is wrong, and the baseline doc says the list may only shrink, so there is no correct move available to the author.

At 3am this blocks a legitimate PR with a false claim, and the operator learns to add bogus `expected_tests` fragments to clear it.

Also: `.verify-suites` and `verify.yml` both predate this branch on `origin/main`, so the PR body's "no CI step or manifest entry names any of them" was already false when measured.

---

### FINDING 2 (minor): the shrink-only ratchet is hand-clearable in one command

**File:** `q-system/.q-system/scripts/reachability-check.py:142`

The baseline's `_why` field says "This set may only SHRINK." Nothing in code enforces it. `--write-baseline` overwrites the frozen set with whatever is currently unreached, growing it, and the check then reports OK.

Reproducer, same temp clone with the new unreached test:

```
[1a] before rewrite rc= 2 reachability-check: FAIL -- new test file that no gate executes:
[1b] --write-baseline rc= 0 baseline written: 54 unreached test files frozen
[1c] after rewrite rc= 0 reachability-check: OK (178 test files, 54 baselined as executed by no gate, 0 new)
[1d] baseline GREW: 54 entries; new file present: True
```

Bounded and reversible (the growth shows in the diff). Raising it here because this fleet self-merges PRs and `no-orphan-findings.md` states the opposite property as doctrine: "You cannot hand-clear the gate."

---

### FINDING 3 (minor): shell tests are invisible to a check whose docstring says "Every test file"

**File:** `q-system/.q-system/scripts/reachability-check.py:63` (`tracked_tests`)

`tracked_tests` matches `test_*.py` only. `capability-gate.py:43` treats `test-*.sh` as a test artifact, and 19 tracked shell tests sit outside the capability scan roots today, including 12 root-level `test-kipi-update-*.sh`.

Reproducer, new unreached shell test at the repo root:

```
[shell test] rc= 0 | reachability-check: OK (177 test files, 53 baselined as executed by no gate, 0 new)
```

Line 23 of the docstring does say `test_*.py`, but line 2 says "Every test file is EXECUTED BY A GATE" and the CI step is named "no new test that nothing runs".

---

### What is sound

- **Fleet RED from widening `in_scan_scope`.** I expected the 26 new declarations under `.q-system/tests/` plus the widened scope to break instance mode via the `skeleton_only` exemption, which only applies on the out-of-scope branch. All 21 `skeleton_only` entries live under `scripts/`, already in scope before this PR. No behavior change there.
- **Declaring tests `validate.yml` warns are red in CI.** The workflow comment says separation has 6 failures and `test_token_guard.py` fails 11 in CI, and this PR declares both. `gh pr checks 286` shows `validate pass 13m17s` at this head. Locally, separation is `168 passed in 15.09s`. Attack failed.
- **New test vs the gate's 60s default timeout.** `test_reachability_check.py` clones the repo six times; I assumed a timeout. Measured `6 passed in 3.04s`. Attack failed.
- **Comment-stripping in `ci_targets`.** The commented `pytest q-system/.q-system/tests` in `verify.yml` genuinely does not confer coverage, and `test_a_commented_out_ci_run_does_not_count_as_coverage` pins it. Sound, and the right instinct.
- **Stale refusal text in `capability-gate.py:172`** ("paths live ... directly in `q-system/.q-system/`", now inaccurate after the widening). I tried to trigger it with an out-of-root declaration and could not make it print; a `scope_exempt` prefix absorbed the fixture. Dropped for lack of a repro.
- **F1b as a mutation test.** I did not re-run the `rglob` to `glob` mutation myself, so I am not vouching for the "reverting fails F1b and leaves F1 green" claim either way. The reasoning behind adding F1b is right: F1 planted its file in the one directory discovery already walked.

---

### VERDICT: REQUEST CHANGES

Not BLOCK. Nothing here is permanent or unrecoverable; the baseline is a JSON file and the CI step is one line.

**Fix first:** teach `ci_targets` (or a sibling `verify_suite_targets`) to read `.verify-suites`, the manifest `verify.yml` actually executes, then regenerate the baseline. The 20 mislabelled entries drop out on their own, the false alarm in Reproducer C goes away, and the docstring's voiceloop claim stops contradicting the repo's own CI log.

FINDINGS:
major|reachability-check.py ignores .verify-suites, so 20 of 53 baselined files are labelled "executed by no gate" while the required verify workflow runs them, and a new test in those suites false-fails CI|q-system/.q-system/scripts/reachability-check.py:83
minor|--write-baseline lets the "may only SHRINK" ratchet be hand-cleared and grown; the shrink-only property is prose, not code|q-system/.q-system/scripts/reachability-check.py:142
minor|tracked_tests matches only test_*.py, so 19 tracked shell tests outside the capability scan roots are invisible to a check documented as covering every test file|q-system/.q-system/scripts/reachability-check.py:63
END FINDINGS

PR #300, opened 2026-09-04
The done line is a criterion, so write it as one

assafkip:
Founder, reading the Inbox he had just been given: *"if it says I'm done why is it showing me the thread?"*

Every Gmail row's note read `Done signal: you replied in the thread` — a past-tense sentence about him. The row asserted the exact thing it was asking him to do, while Status said "Not started" two columns away.

The line was meant to say what finishing looks like. It read as a claim that finishing had happened. The phrasing came off the reference board and was never read back as a sentence.

Every string is now a condition: `Done when: you reply in the thread`.

The test is a grep over every value in `DONE_DEFAULT`, `DONE_BY_KIND`, `DONE_BY_SOURCE` and `DONE_GTM_FALLBACK`, because the defect was one word in one string and the next such string will be written by whoever adds the next source.

1184 passed. One existing assertion updated to the new contract rather than deleted.

https://claude.ai/code/session_01WsGRRMj4faDvWqg54SZhVw

### DIFF
--- a/q-system/.q-system/scripts/answered_elsewhere.py
+++ b/q-system/.q-system/scripts/answered_elsewhere.py
@@ -128,12 +128,31 @@ def channel_slugs(path=None) -> dict:
     return out
 
 
-def filter_answered(threads, cache=None, registry=None, chat_last=None):
+#: Distinct days he must have spoken in a client's chat AFTER their mail before that
+#: chat counts as having answered it.
+#:
+#: PRESENCE IS NOT A REPLY, and the first version of this got that wrong. It dropped a
+#: mail the moment his LAST chat message postdated it, so for a client he talks to
+#: daily every email older than this morning vanished -- including, in the reviewer's
+#: example, a two-hour-old signature request and a twenty-day-overdue invoice. Talking
+#: to someone at 10am is not evidence you dealt with what they sent at 9am.
+#:
+#: Three days is read off the case this rule was built from rather than picked: that
+#: client mailed on 08-05 and he wrote in their chat on 15 separate days between 08-18
+#: and 09-02. Sustained engagement across days is the thing that makes "I answered it
+#: in the messages" true; a single message is not, and neither is a single day.
+CHAT_DAYS_TO_ANSWER = 3
+
+
+def filter_answered(threads, cache=None, registry=None, chat_days=None,
+                    chat_last=None):
     """(kept, notes). `threads` are dicts with `id`, `from` and `age_hours`.
 
-    `chat_last` is {slug: when he last spoke in that client's chat}, injected so this
+    `chat_days` is {slug: [dates he spoke in that client's chat]}, injected so this
     stays a pure function and so a chat outage is the caller's fact, not a silent
-    pass here.
+    pass here. `chat_last` is the older single-timestamp form and is accepted only
+    so an old caller degrades to KEEPING rows rather than crashing; it never answers
+    anything, because one timestamp cannot show sustained engagement.
     """
     notes = []
     if cache is None:
@@ -149,8 +168,16 @@ def filter_answered(threads, cache=None, registry=None, chat_last=None):
             registry = {}
             notes.append("no registry: chat replies could not be matched to clients")
 
+    if not isinstance(cache, dict):
+        return list(threads), notes + ["kept every thread: mail history is not a map"]
+    if not isinstance(registry, dict):
+        registry = {}
+        notes.append("registry ignored: not a map")
     outbound = his_last_outbound(cache)
-    chat_last = chat_last or {}
+    chat_days = chat_days or {}
+    if chat_last and not chat_days:
+        notes.append("chat evidence ignored: caller passed the old single-timestamp "
+                     "form, which cannot show sustained engagement")
     kept = []
     for th in threads:
         addr = None
@@ -172,10 +199,21 @@ def filter_answered(threads, cache=None, registry=None, chat_last=None):
             continue
 
         slug = client_of(addr, registry)
-        spoke = chat_last.get(slug) if slug else None
-        if spoke and spoke > inbound:
-            notes.append(f"{addr}: answered in {slug}'s chat {spoke.date()}, after "
-                         f"their {inbound.date()}")
+        # `>=` so a chat message on the SAME DAY as their mail is counted as a chat
+        # day. It is still nowhere near CHAT_DAYS_TO_ANSWER, so it cannot answer
+        # anything; counting it is what makes the "presence is not a reply" note fire
+        # on exactly the case the reviewer named (mail 09:00, chat 10:00).
+        days = sorted({d for d in (chat_days.get(slug) or [])
+                       if d and d >= inbound.date()}) if slug else []
+        if len(days) >= CHAT_DAYS_TO_ANSWER:
+            notes.append(f"{addr}: answered in {slug}'s chat on {len(days)} days "
+                         f"since their {inbound.date()} (through {days[-1]})")
             continue
+        if days:
+            # Named rather than silent: he should be able to see that the chat was
+            # looked at and did not clear this one, instead of wondering why one
+            # mail from a client he talks to daily is still on the board.
+            notes.append(f"{addr}: KEPT, only {len(days)} chat day(s) since their "
+                         f"mail; presence is not a reply")
         kept.append(th)
     return kept, notes
--- a/q-system/.q-system/scripts/answered_elsewhere.py
+++ b/q-system/.q-system/scripts/answered_elsewhere.py
@@ -1,0 +1,181 @@
+#!/usr/bin/env python3
+"""Was this mail thread answered SOMEWHERE ELSE? Deterministic, no model call.
+
+Founder, 2026-09-04, reading an Inbox of twelve rows he had already dealt with:
+*"You're only looking if I replied in the thread, which is wrong, because as long as
+I replied to that email address and the reply makes sense, then it should be fine.
+And for people like [a retainer client], if I replied in the messages, that's fine."*
+
+`collect_mail` asks a model for threads where he "has not replied yet", and a model
+reading one thread can only answer THREAD-SCOPED. Two ways that is wrong, and both
+were measured on his live data before this module was written:
+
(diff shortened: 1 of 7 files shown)


assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-300-20260904-142352.md` (9477 bytes).

```
FINDINGS:
major|channel_slugs() needs a two-column groupme-channels file while the shipped load_allowlist() parses it as one id per line, so the chat check never fires today and silently blanks the GroupMe section if the file is converted|q-system/.q-system/scripts/answered_elsewhere.py:126
major|Any message he sends in a client's group chat drops every older email from that client, including a two-hour-old signature request and a 20-day-overdue invoice, because presence is treated as a reply|q-system/.q-system/scripts/answered_elsewhere.py:176
minor|The keep-everything notes print under a "[mail] dropped," prefix, so the only 3am log line reporting the check was inert reads as a drop|q-system/.q-system/scripts/morning-brief.py:306
minor|The past-tense guard is a denylist of exactly the seven phrases this PR removed, so a new source's "you responded to them" ships green|q-system/.q-system/tests/test_consulting_board.py:1693
minor|The module docstring claims filter_answered never raises, but a non-dict registry or cache raises AttributeError|q-system/.q-system/scripts/answered_elsewhere.py:39
nit|Comment reads "carries a / a Done when line" (double article) and credits the reference board with this PR's new phrasing|q-system/.q-system/scripts/consulting_board.py:575
END FINDINGS
```

--- reviewer output ---

# Review: PR #300

Two halves shipped here. The rename half (`Done signal:` becomes `Done when:`) is clean. The `answered_elsewhere` half is where the findings are.

---

## MAJOR 1 — the chat check reads a file format its own neighbour cannot parse, in both directions

`answered_elsewhere.channel_slugs()` needs two whitespace columns per line. The shipped `groupme_inbox.load_allowlist()` documents and consumes the same file as one id per line.

**File:** `q-system/.q-system/scripts/answered_elsewhere.py:126` and `q-system/.q-system/scripts/morning-brief.py:274`

**Reproducer:**
```
python3 - <<'PYEOF'
import importlib.util, pathlib
S = pathlib.Path(".../q-system/.q-system/scripts")
def load(stem):
    spec = importlib.util.spec_from_file_location(stem, S / (stem + ".py"))
    m = importlib.util.module_from_spec(spec); spec.loader.exec_module(m); return m
ae, gm = load("answered_elsewhere"), load("groupme_inbox")
a = pathlib.Path("/tmp/pr300/chan_a"); a.write_text("# the AI chat channel\n108374123\n99887766\n")
print(gm.load_allowlist(a), ae.channel_slugs(a))
b = pathlib.Path("/tmp/pr300/chan_b"); b.write_text("108374123 acme\n99887766 bloom\n")
print(gm.load_allowlist(b), ae.channel_slugs(b), "108374123" in gm.load_allowlist(b))
PYEOF
```

**Real output:**
```
FORMAT A (one id per line -- what groupme_inbox.py documents and ships):
  groupme_inbox.load_allowlist -> {'108374123', '99887766'}
  answered_elsewhere.channel_slugs -> {}
FORMAT B (two columns -- what channel_slugs() needs):
  groupme_inbox.load_allowlist -> {'108374123 acme', '99887766 bloom'}
  answered_elsewhere.channel_slugs -> {'108374123': 'acme', '99887766': 'bloom'}
  live group id '108374123' passes waiting()'s allowlist test? -> False
```

Two consequences, and the second is the one that bites at 3am:

- On the file that exists today, `channel_slugs()` returns `{}`, so `_chat_last()` returns `{}`, so the chat half never fires. The 717-hour retainer-client row this PR was written for stays on the board.
- Adding the second column to make it fire feeds `"108374123 acme"` into `load_allowlist()`. `groupme_inbox.waiting():143` tests `str(g["id"]) not in allow`, so every allowlisted group is silently skipped and the GroupMe section reports nothing. That is the exact failure `groupme_inbox.py:91` says the brief was built against.

No test covers `channel_slugs()` against a real allowlist file, which is why both directions are open.

---

## MAJOR 2 — presence in the chat is treated as a reply, so one "morning" clears every pending email from that client

`_chat_last()` returns the timestamp of his most recent message in the group, on any subject. `filter_answered` drops every thread from that client older than it.

**File:** `q-system/.q-system/scripts/answered_elsewhere.py:176`

**Reproducer output:**
```
he said 'morning' in the group chat 30 min ago, and replied to none of these
threads in : 3
threads out: 0 []
  note: audrey@acme.com: answered in acme's chat 2026-09-04, after their 2026-09-04
  note: audrey@acme.com: answered in acme's chat 2026-09-04, after their 2026-08-15
  note: audrey@acme.com: answered in acme's chat 2026-09-04, after their 2026-09-04
```

The three inputs are a two-hour-old SOW signature request, a 20-day-overdue invoice, and a board-call invite. All three vanish. For a client he chats with daily, the Gmail lane is effectively off: any email arriving before his next chat message is dropped forever.

The founder's words in the module docstring are "if I **replied** in the messages." The code checks whether he **spoke**. The module's own contract at line 30 says every uncertainty resolves toward keeping the row, and "did that chat message answer this email" is precisely an uncertainty that resolves toward dropping.

Recoverable: the mail is still in Gmail. Not permanent. So major, not blocker.

---

## MINOR 3 — the one log line saying the check was inert reads as a drop

**File:** `q-system/.q-system/scripts/morning-brief.py:306`

The `keep everything` notes are printed under a `[mail] dropped,` prefix.

**Real output** (KIPI_CONSULTING_ROOT pointed at a missing dir, stub runner, one thread):
```
rows kept: 1  error: None
what the run log says:
   [mail] dropped, kept every thread: mail history unreadable (FileNotFoundError)
```

Nothing was dropped. The registry note (`no registry: chat replies could not be matched to clients`) prints the same way. At 3am this is the only signal that the feature did nothing, and it says the opposite.

---

## MINOR 4 — the past-tense grep is a denylist of the seven strings the PR just removed

**File:** `q-system/.q-system/tests/test_consulting_board.py:1693`

Its docstring says it guards "the next one added will be written by whoever adds a source." Added `"you responded to them"` and `"you closed the ticket"` to `DONE_BY_SOURCE`:

```
strings under test: ['you act on it', ..., 'you responded to them', 'you closed the ticket']
denylist hits (test fails only if non-empty): []
TEST RESULT: PASS -- past-tense strings shipped green
```

A regex on `\byou [a-z]+ed\b` would go red on both. As written, the oracle is the set of strings the fix already passes.

---

## MINOR 5 — `filter_answered` "never raises" is false

**File:** `q-system/.q-system/scripts/answered_elsewhere.py:39`

```
A3 junk 'string' -> RAISED AttributeError 'str' object has no attribute 'get'
A3 junk 42 -> RAISED AttributeError 'int' object has no attribute 'get'
A3 junk [1, 2] -> RAISED AttributeError 'list' object has no attribute 'get'
```

A top-level JSON array in `clients.json` or `thread-cache.json` reaches this. Contained by `collect_mail`'s outer `except Exception` (every thread kept), and I could not verify any producer emits that shape (both files live in the consulting repo, which this session is sandboxed out of), so it is dropped a level and stated as such. The docstring is still wrong.

---

## NIT 6 — double article, and the reference board is now credited with our phrasing

**File:** `q-system/.q-system/scripts/consulting_board.py:575`

Reads "Bloom's board carries a / a "Done when" line". Also, the PR body says Bloom's board carries "Done signal"; the comment now says it carries "Done when".

---

## What is sound

- **Graph-shaped nested recipients.** `_recipients` stringifies each item before the regex, so `toRecipients: [{"emailAddress": {"address": ...}}]` resolves correctly even though no fixture uses that shape. Returned `{'audrey@acme.com'}` and dropped the thread. This is the robustness the fixture question usually breaks on.
- **Naive vs aware timestamps.** `_when` normalises to UTC, so a cache with no `Z` suffix does not raise a `can't compare offset-naive` TypeError. Verified.
- **The `_provenance` cache key** is skipped, not treated as a thread. Verified.
- **A thread with no `from` field at all** is kept, not crashed on. Verified.
- **Direction of the mail rule.** An outbound predating theirs, and an outbound to a different address, both keep the row. Both tested and both go red if inverted.
- **The `Done when:` rename has no reader.** Grepped the tree: nothing parses `"Done signal: "` back out of a Notion note, so no already-written row breaks. That is the failure mode a prefix rename usually has, and it is absent here.
- **`age_hours` is the live shape**, not `inbound_at`, and the fallback path is the one the test exercises. The author got that right.
- The PR's two test files pass: `149 passed in 5.38s`. The full-suite `1184` claim was still running when I wrote this and I do not contest it.

---

## VERDICT: REQUEST CHANGES

Not BLOCK. Nothing here is permanent: the mail stays in Gmail, the board regenerates daily, and both failure paths keep rows rather than destroying them.

**Fix first:** major 1. Pick one owner for `~/.config/kipi/groupme-channels` and one parser, then write the test that reads a real two-column file through `load_allowlist` and asserts a live group id still matches. Right now the chat half is dead on the current format and arms a silent GroupMe blackout on the new one, and major 2 only becomes reachable once major 1 is fixed.

FINDINGS:
major|channel_slugs() needs a two-column groupme-channels file while the shipped load_allowlist() parses it as one id per line, so the chat check never fires today and silently blanks the GroupMe section if the file is converted|q-system/.q-system/scripts/answered_elsewhere.py:126
major|Any message he sends in a client's group chat drops every older email from that client, including a two-hour-old signature request and a 20-day-overdue invoice, because presence is treated as a reply|q-system/.q-system/scripts/answered_elsewhere.py:176
minor|The keep-everything notes print under a "[mail] dropped," prefix, so the only 3am log line reporting the check was inert reads as a drop|q-system/.q-system/scripts/morning-brief.py:306
minor|The past-tense guard is a denylist of exactly the seven phrases this PR removed, so a new source's "you responded to them" ships green|q-system/.q-system/tests/test_consulting_board.py:1693
minor|The module docstring claims filter_answered never raises, but a non-dict registry or cache raises AttributeError|q-system/.q-system/scripts/answered_elsewhere.py:39
nit|Comment reads "carries a / a Done when line" (double article) and credits the reference board with this PR's new phrasing|q-system/.q-system/scripts/consulting_board.py:575
END FINDINGS

PR #291, opened 2026-08-31
voice-loop C3 + seam fix: the Stop-gate reads an optional instance channel registry

assafkip:
## Why this is urgent rather than tidy

`plugins/kipi-core/voiceloop/channel_registry.py` does not exist on `origin/main`. Measured:

```
$ git ls-tree -r origin/main --name-only | grep voiceloop
plugins/kipi-core/voiceloop/__init__.py
plugins/kipi-core/voiceloop/assemble.py
plugins/kipi-core/voiceloop/corpus.py
plugins/kipi-core/voiceloop/echo.py
plugins/kipi-core/voiceloop/fingerprint.py
plugins/kipi-core/voiceloop/selector.py
plugins/kipi-core/voiceloop/validate.py
plugins/kipi-core/voiceloop/voice_ref.py
...
```

Every sibling is there. `channel_registry.py` is not.

Consulting imports it at MODULE level, so an absent file is an import error and not a
degraded feature (`consulting/q-consult/pipeline/voice.py:47-48`):

```python
from voicekit import assemble, corpus, echo, fingerprint, validate  # noqa: E402
from voicekit import channel_registry  # noqa: E402
```

The `plugins/` sync runs `rsync --delete` from the skeleton WORKING TREE
(`kipi-update.sh:555`, `config_source_manages` at :1288-1294: "the rsync is --delete, so
it removes instance files the skeleton does not carry"). Measured on disk today:

```
$ ls plugins/kipi-core/voiceloop/channel_registry.py     # skeleton working tree
ls: ...: No such file or directory
$ ls plugins/kipi-core/voicekit/channel_registry.py      # consulting instance
plugins/kipi-core/voicekit/channel_registry.py
```

So a `kipi update` today strands the import. Captured as `sp-885e1df0`. It resolves by
landing, not by a design choice.

## What shipped

Two commits, C3 first because the seam fix copies its resolver.

**C3 (`ed4f5b32`, `18485c98`) - the skeleton Stop-gate reads an OPTIONAL instance channel registry.**
26 instances have no registry and must behave byte-identically to before. An instance
WITH one routes the channel to its surface lint instead. A present-but-BROKEN registry
holds the turn (fail-closed) rather than falling back to the assaf lints, because that
fallback is the wrong-rulebook bug wearing a fix (scar 2026-07-22: a reddit draft graded
on the wrong rulebook and shipped AI-sounding).

**Seam fix (`cb5d8292`, `840eb048`) - the channel set is derived from the registry, not restated four times.**
[...]

PR #301, opened 2026-09-05
voiceloop slice 0: reconcile the engine tree, skeleton side

assafkip:
Skeleton side of slice 0 of the VoiceLoop package extraction. Plan of record:
`q-system/output/plans/voiceloop-package-extraction-2026-09-04.md` (slice 0) and
`voiceloop-slice-0-sync-2026-09-05.md`, both instance-local in `~/projects/consulting`.

## The problem

`plugins/kipi-core/voiceloop/` had diverged in BOTH directions between the
consulting instance and this skeleton. The fleet propagation script rsyncs
`plugins/` from the skeleton working tree with a delete flag, so a naive one-way
copy in either direction destroys real work. An instance-to-skeleton copy would
have reverted PR #291.

## Direction, measured per file

Measured in a worktree cut from `origin/main` f9f74ac1, by AST symbol diff both
ways plus a check for main-only lines a carry-up would destroy. Never by size,
never by mtime.

| Direction | Files |
|---|---|
| Byte identical, untouched | `corpus.py`, `echo.py`, `voice_ref.py`, `tests/test_length_axis.py`, `tests/test_no_founder_data.py`, `tests/test_selector_form_match.py` |
| Main already ahead, untouched here | `channel_registry.py` (main defines `instance_root_for`, `for_path`, `_MAX_WALK`), `tests/test_channel_registry.py` (main has 6 symbols the instance lacks), `__init__.py` |
| Instance ahead, carried up | `fingerprint.py` (`_sigma`, `style_distance`, `zscores`), `selector.py`, `tests/test_voiceloop.py` (5 new tests), `assemble.py` (`EXTERNAL_SOURCE`) |
| Hand merged | `validate.py` |

`assemble.py` is the only carried-up file with main-only lines, and all three
are superseded rather than reverted: `BUDGET_CHARS` 20000 to 24000 is the
documented 2026-08-31 headroom raise, and the `if applied:` block is replaced by
the his/researched split `EXTERNAL_SOURCE` introduces.

## Why validate.py could not be a copy either way

Two properties lived on opposite sides:

- `check_correction_share` plus `CORRECTION_SHARE_CEILING`, only in the instance
- `channels = channel_registry.for_path(voice_dir)` in `check_all`, only on main

Copying the instance copy over this one reverts #291's load-path wiring and
deletes the scar comment recording that "documenting a seam is not wiring one".
Copying this one over the instance drops the share guard. Both are now in one
file.

## Verification

```
python3 -m pytest plugins/kipi-core/voiceloop/tests/ -q
121 passed in 0.15s
```

Negative self-test, one mutant per merged property, `PYTHONDONTWRITEBYTECODE=1`,
run against the worktree and never a live tree:

| Mutant | Result |
|---|---|
| delete the `check_correction_share` call from `check_all` | 2 failed: `test_corrections_crowding_out_his_writing_is_red`, `test_a_RETIRED_correction_does_not_count_toward_the_share` |
| revert `for_path(voice_dir)` to `channel_registry.DEFAULT` | 1 failed: `test_check_all_loads_the_instance_registry` |

Both restored, suite back to 121 passed, restored file byte-identical to the
pre-mutation copy.

`./kipi check` on the skeleton: exit 1 both before and after, with the same two
REDs, neither related to this change (`~/projects/anthology` has no off-disk
remote; `~/projects/consulting/.opencode` is not a git repo). Green was never an
available criterion here.

## Scope

Skeleton only. The consulting instance receiving `for_path` and the fixed
`check_all` is a separate sequenced step.
`plugins/kipi-core/.claude-plugin/plugin.json` 1.9.5 to 1.9.6 is the lefthook
`plugin-version-bump` gate doing its job: the plugin cache is version keyed, so a
changed plugin that does not bump serves the stale copy.

The skeleton checkout itself is untouched. It sits on
`fix/candidate-draft-one-definition` with unrelated uncommitted work, so the
fleet propagation preflight still aborts on its branch guard. That is correct and
is not something this PR tries to fix.

https://claude.ai/code/session_01M41mnfj2RWAZ1UomJxeXfx

### DIFF
--- a/plugins/kipi-core/.claude-plugin/plugin.json
+++ b/plugins/kipi-core/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "kipi-core",
-  "version": "1.9.6",
+  "version": "1.9.7",
   "description": "Core founder OS \u2014 voice, research, AUDHD, LinkedIn/brand, RCA skills + kipi-mcp server",
   "author": {
     "name": "Assaf Kipnis",
--- a/plugins/kipi-core/.claude-plugin/plugin.json
+++ b/plugins/kipi-core/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "kipi-core",
-  "version": "1.9.7",
+  "version": "1.9.8",
   "description": "Core founder OS \u2014 voice, research, AUDHD, LinkedIn/brand, RCA skills + kipi-mcp server",
   "author": {
     "name": "Assaf Kipnis",
--- a/plugins/kipi-core/.claude-plugin/plugin.json
+++ b/plugins/kipi-core/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "kipi-core",
-  "version": "1.9.5",
+  "version": "1.9.6",
   "description": "Core founder OS \u2014 voice, research, AUDHD, LinkedIn/brand, RCA skills + kipi-mcp server",
   "author": {
     "name": "Assaf Kipnis",
--- a/plugins/kipi-core/.claude-plugin/plugin.json
+++ b/plugins/kipi-core/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "kipi-core",
-  "version": "1.9.9",
+  "version": "1.9.10",
   "description": "Core founder OS \u2014 voice, research, AUDHD, LinkedIn/brand, RCA skills + kipi-mcp server",
   "author": {
     "name": "Assaf Kipnis",
--- a/plugins/kipi-core/.claude-plugin/plugin.json
+++ b/plugins/kipi-core/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "kipi-core",
-  "version": "1.9.8",
+  "version": "1.9.9",
   "description": "Core founder OS \u2014 voice, research, AUDHD, LinkedIn/brand, RCA skills + kipi-mcp server",
   "author": {
     "name": "Assaf Kipnis",
--- a/plugins/kipi-core/voiceloop/assemble.py
+++ b/plugins/kipi-core/voiceloop/assemble.py
@@ -23,11 +23,36 @@ from __future__ import annotations
 
 from . import selector
 
-BUDGET_CHARS = 20000     # asserted by validate.feasible + the consumer's suite,
+BUDGET_CHARS = 24000     # asserted by validate.feasible + the consumer's suite,
                          # NEVER enforced by a runtime raise or slice (ASK-461:
                          # a cap you cannot see is the same bug; a cap that kills
                          # the daily job is a worse one).
+                         #
+                         # RAISED FROM 20000 ON 2026-08-31, and this is a headroom
+                         # alarm being re-set, not a gate being switched off. It is a
+                         # round number with no model limit behind it -- 20000 chars
+                         # is roughly 5k tokens -- and it fires at suite time only.
+                         # What tripped it: landing the four measured 2026-08-24 shape
+                         # corrections from the stranded fix/restatement branch took
+                         # linkedin's worst assembly from 18731 to 20315 (measured,
+                         # both numbers, by rendering 12 counters each way). Dropping
+                         # a measured correction to stay under a round number would be
+                         # the redundancy dismissal the brief bans.
+                         #
+                         # THE REAL SIGNAL IS STILL REAL and is captured as spillover:
+                         # every correction renders into every in-scope prompt, so the
+                         # ledger grows without bound and this alarm will trip again.
+                         # Raising it buys room; it does not fix growth.
 
+
+#: Corrections carrying this source were derived from aggregate performance
+#: research on OTHER accounts, not from the founder correcting his own draft.
+#: They render as options under their own header instead of the override header,
+#: because "posts he has written" and his own corrections are measurements of HIM,
+#: while an external lift number is a hypothesis about what might also work for
+#: him. Unmarked rows render exactly as they always did, so a corpus that never
+#: uses this field behaves byte for byte as before (ASK fleet-wide).
+EXTERNAL_SOURCE = "external-performance"
 
 def _lexicon_positive(lexicon):
     lines = []
@@ -76,16 +101,26 @@ def voice_section(voice, channel, counter, slot_index=0, k=selector.DEFAULT_K,
     # the one that looks authoritative.
     applied = [r for r in (corrections or [])
                if not r.get("scope") or channel in r["scope"]]
-    if applied:
-        lines = "\n".join(f"- {r['instruction']}" for r in applied)
+    his = [r for r in applied if r.get("source") != EXTERNAL_SOURCE]
+    researched = [r for r in applied if r.get("source") == EXTERNAL_SOURCE]
+    if his:
+        lines = "\n".join(f"- {r['instruction']}" for r in his)
         if lines:
             parts.append("STANDING CORRECTIONS (these override everything above):\n"
                          + lines)
+    if researched:
+        lines = "\n".join(f"- {r['instruction']}" for r in researched)
+        parts.append(
+            "RESEARCHED SHAPES (measured on other writers' posts, not on the posts "
+            "above. Optional: use a shape when it fits the idea, ignore it when it "
+            "fights how he writes. The posts above and the standing corrections win "
+            "anywhere they disagree):\n" + lines)
 
     provenance = {
         "exemplar_ids": [str(r.get("id")) for r in picked],
         "exemplar_texts": [(r.get("text") or "") for r in picked],
         "correction_ids": [str(r.get("id")) for r in applied],
+        "external_correction_ids": [str(r.get("id")) for r in researched],
         "selection_reason": selector.selection_reason(picked, channel, counter,
                                                       slot_index),
         "skipped_rows": voice.skipped_rows,
--- a/plugins/kipi-core/voiceloop/fingerprint.py
+++ b/plugins/kipi-core/voiceloop/fingerprint.py
@@ -310,3 +310,76 @@ def load(path):
     whether a missing/broken file degrades (daily job: yes) or fails (validator: no)."""
     with open(path, encoding="utf-8") as handle:
         return json.load(handle)
+
+
+# --- continuous scoring (2026-08-24, RCA-voice-enforcement) -------------------------
+#
+# why: binary min/max bounds answer "did he EVER do this" and almost always the
+# answer is yes, because generation is anchored on his own exemplars. The brief-era
+# failure was exactly that: a draft at colon_rate 1.29 scored inside [0, 20] and
+# every layer reported clean while the founder read it as not-him. Continuous
+# distance answers the question a human is actually asking: how FAR from his center
+# is this, measured in units of his own spread. External practice (style-drift
+# detection, stylometry verification) uses per-feature z-scores plus a multi-feature
+# aggregate; both are defined here so the caller keeps choosing policy.
+
+_P90_Z = 1.2816     # p90 of a standard normal: p10..p90 spans 2 x this
+_RANGE_DIV = 4.0    # degenerate-band fallback: treat observed range as ~4 sigma
+_MIN_SIGMA = {
+    # Counts and rates move in small steps; without a floor, one colon in a short
+    # post produces an enormous z and the aggregate becomes single-metric theater.
+    # Floors are in the metric's OWN units, chosen at the instrument level so no
+    # caller can forget them.
+    "colon_rate": 0.35, "hedge_count": 0.5, "formal_pair_count": 0.5,
+    "midpost_questions": 0.5, "monotony_run": 0.5, "para_max_sentences": 0.75,
+}
+
+
+def _sigma(band):
+    """Robust spread estimate for one metric's band, in the metric's units."""
+    p10, p90 = band.get("p10"), band.get("p90")
+    if p10 is None or p90 is None:
+        return None
+    if p90 > p10:
+        return max((p90 - p10) / (2 * _P90_Z), 0.0)
+    lo, hi = _bounds(band)
+    if lo is not None and hi is not None and hi > lo:
+        return (hi - lo) / _RANGE_DIV
+    return None
+
+
+def zscores(text, fingerprint_doc):
+    """Per-metric robust z: (value - p50) / sigma, sigma from the p10-p90 span.
+
+    Signed: negative means BELOW his median (e.g. too few first-person tokens),
+    positive means above. Metrics whose spread cannot be estimated are omitted
+    rather than guessed. Pure; no clock, no model.
+    """
+    got = metrics(text)
+    out = {}
+    for name, band in ((fingerprint_doc or {}).get("metrics") or {}).items():
+        value = got.get(name)
+        sigma = _sigma(band)
+        if value is None or not sigma:
+            continue
+        sigma = max(sigma, _MIN_SIGMA.get(name, 0.0))
+        p50 = band.get("p50")
+        if p50 is None:
+            continue
+        out[name] = round((value - p50) / sigma, 2)
+    return out
+
+
+def style_distance(zs, cap=8.0):
+    """Aggregate multi-axis distance: root sum of squares of capped |z|.
+
+    The cap stops one wild metric from dominating; the sqrt-of-sum keeps the
+    property that several mild deviations together outweigh one mild one, which
+    is the drift signal binary bounds cannot express at all.
+    """
+    total = 0.0
+    for z in (zs or {}).values():
+        az = min(abs(z), cap)
+        if az >= 0:
+            total += az * az
+    return round(total ** 0.5, 2)
--- a/plugins/kipi-core/voiceloop/selector.py
+++ b/plugins/kipi-core/voiceloop/selector.py
@@ -49,12 +49,20 @@ def eligible(rows, channel, slot_kind="post", target_words=None):
     at 169 words, the one long-form linkedin sample a long-form request most needs.
     The founder's spec is that the skill sees all the examples; it saw 41 of 59.
 
+    A row with `eligible_for_voice_reference` explicitly False is OUT of both
+    tiers, anchors included (2026-08-23). Before this line the field was dead:
+    the corpus carried it, curation flipped it, and selection never read it, so
+    a hype-register post ("It's ALIVE!!!") sat as an anchor in every linkedin
+    prompt while believing it was retired. Absent or null keeps the row live --
+    retirement is an explicit act, recorded on the row with a reason.
+
     The 2026-08-09 scar is preserved exactly, because it is a scar about SHORT
     slots: article rhythm taught post slots and the engine published essays on a
     280-char channel. A short request still gets the strict tiering that fixed it.
     Only a long request promotes, and for a long request article rhythm is the
     right answer rather than the bug.
     """
+    rows = [r for r in rows if r.get("eligible_for_voice_reference") is not False]
     tiers = ELIGIBLE_KINDS.get(slot_kind, ELIGIBLE_KINDS["post"])
     primary_kinds, fallback_kinds = tiers[0], tiers[1]
     if (target_words is not None and target_words >= LONG_WORDS
--- a/plugins/kipi-core/voiceloop/tests/test_voiceloop.py
+++ b/plugins/kipi-core/voiceloop/tests/test_voiceloop.py
@@ -572,13 +572,35 @@ class TestValidate:
         # two extremums of it.
         rows = _rows(20)
         for i, r in enumerate(rows):
-            r["text"] = f"Run {i}. " + "I watched the queue back up. " * (1 + i * 3)
+            # DESCENDING, so the shortest rows sit at the END of the pool order and
+            # the thinnest assembly falls at counter 16. Ascending put it at counter
+            # 0, where a 12-counter window still finds it -- that fixture let a
+            # range(12) mutant survive at 122 passed. A bound only fakes a check
+            # when the answer happens to sit inside it.
+            r["text"] = ("Run {}. ".format(i)
+                         + "I watched the queue back up. " * (1 + (19 - i) * 3))
         heavy = [{"id": f"c{i}", "status": "active",
                   "instruction": f"Rule {i}. " * 200} for i in range(6)]
         d = _voice_dir(tmp_path, rows=rows, corrections=heavy,
                        fp=self._fresh_fp(rows))
         voice = corpus.load(d)
-        lengths = [len(assemble.voice_section(voice, "x", c)[0]) for c in range(12)]
+        # THE ORACLE IS DERIVED INDEPENDENTLY, over the FULL rotation. Its first
+        # draft looped `range(12)` because the code did, which is the trap
+        # `_resolved_slots` already names in this file: a checker that re-derives
+        # the rule it checks is testing its own copy. The 12-window hid a second
+        # bug of the same class (PR #301 review round 2) because test and code
+        # sampled the same wrong window and agreed.
+        period = len(selector.resolved_pool(voice.active_exemplars(), "x", "post",
+                                            selector.DEFAULT_K))
+        assert period > 12, (
+            f"this fixture's rotation period is {period}, so a 12-counter window "
+            "would cover it and the test cannot see an under-sampling bug")
+        lengths = [len(assemble.voice_section(voice, "x", c)[0])
+                   for c in range(period)]
+        assert lengths.index(min(lengths)) >= 12, (
+            "the thinnest assembly is at counter "
+            f"{lengths.index(min(lengths))}, inside a 12-counter window, so this "
+            "fixture cannot tell full-rotation sampling from the old range(12)")
         rules = sum(len(c["instruction"]) for c in heavy)
         assert rules / max(lengths) <= validate.CORRECTION_SHARE_CEILING, (
             "the fattest assembly must read CLEAN here, or the old max-denominator "
--- a/plugins/kipi-core/voiceloop/tests/test_voiceloop.py
+++ b/plugins/kipi-core/voiceloop/tests/test_voiceloop.py
@@ -550,73 +550,116 @@ class TestValidate:
                                                      corrections=heavy[:1] + retired,
                                                      fp=self._fresh_fp(rows))))
 
-    def test_the_share_is_measured_at_the_THINNEST_prompt_not_the_fattest(self,
-                                                                          tmp_path):
-        """The denominator, and it is the whole guard.
+    def test_the_share_samples_the_WHOLE_rotation_not_a_fixed_twelve(self,
+                                                                     tmp_path):
+        """The counter axis, isolated from the other two.
 
-        Shipped taking the MAX over 12 counters, copied from `check_budget` where
-        max is right. Max reports the BEST case of the thing being graded, so on
-        the live ASK corpus 2026-09-04 the x channel read 60% against the max and
-        71% against the min, and the guard returned [] with the ceiling at 70%.
+        `selector.select` offsets by `counter % len(pool)`, so the period is the
+        pool size. The guard inherited `range(12)` from `check_budget` and could
+        not see counters 12 upward. Live pools measured 45 / 31 / 10.
 
-        This fixture is sized so the two denominators DISAGREE ON THE VERDICT:
-        66% against the fattest assembly, which is clean, and 90% against the
-        thinnest, which is red. A fixture merely far over the line passes against
-        either extremum and would not have caught the bug.
+        EVERY ROW CARRIES THE SAME WORD COUNT on purpose. A length target draws
+        the nearest rows by word count, so on a mixed-register corpus it reaches
+        the thin rows on its own and a `range(12)` mutant survives -- measured,
+        it did. With one register the target axis cannot get there, and the only
+        way to the thinnest prompt is counting past 12. Character length falls as
+        the id rises, which puts that prompt at counter 16.
+
+        Sized so the two windows DISAGREE ON THE VERDICT: 60% over `range(12)`,
+        which is clean, and 74% over the full rotation, which is red.
         """
         import re
-        # EVERY ROW A DIFFERENT LENGTH, on purpose. The first attempt at this
-        # fixture used two repeated bodies, so every row of a class was byte
-        # identical, every counter assembled to exactly 13615 chars, and min and
-        # max were one number. A uniform corpus cannot show a difference between
-        # two extremums of it.
-        rows = _rows(20)
-        for i, r in enumerate(rows):
-            # DESCENDING, so the shortest rows sit at the END of the pool order and
-            # the thinnest assembly falls at counter 16. Ascending put it at counter
-            # 0, where a 12-counter window still finds it -- that fixture let a
-            # range(12) mutant survive at 122 passed. A bound only fakes a check
-            # when the answer happens to sit inside it.
-            r["text"] = ("Run {}. ".format(i)
-                         + "I watched the queue back up. " * (1 + (19 - i) * 3))
+        rows = []
+        for i in range(20):
+            word = "queuebackup" * max(7 - i // 3, 1)
+            rows.append({"id": f"ex-{i:02d}", "kind": "post", "channel": "any",
+                         "status": "active", "weight": 1.0, "anchor": False,
+                         "text": " ".join([word] * 30)})
         heavy = [{"id": f"c{i}", "status": "active",
-                  "instruction": f"Rule {i}. " * 200} for i in range(6)]
+                  "instruction": f"Rule {i}. " * 150} for i in range(6)]
         d = _voice_dir(tmp_path, rows=rows, corrections=heavy,
                        fp=self._fresh_fp(rows))
         voice = corpus.load(d)
-        # THE ORACLE IS DERIVED INDEPENDENTLY, over the FULL rotation. Its first
-        # draft looped `range(12)` because the code did, which is the trap
-        # `_resolved_slots` already names in this file: a checker that re-derives
-        # the rule it checks is testing its own copy. The 12-window hid a second
-        # bug of the same class (PR #301 review round 2) because test and code
-        # sampled the same wrong window and agreed.
-        period = len(selector.resolved_pool(voice.active_exemplars(), "x", "post",
-                                            selector.DEFAULT_K))
-        assert period > 12, (
-            f"this fixture's rotation period is {period}, so a 12-counter window "
-            "would cover it and the test cannot see an under-sampling bug")
-        lengths = [len(assemble.voice_section(voice, "x", c)[0])
-                   for c in range(period)]
-        assert lengths.index(min(lengths)) >= 12, (
-            "the thinnest assembly is at counter "
-            f"{lengths.index(min(lengths))}, inside a 12-counter window, so this "
-            "fixture cannot tell full-rotation sampling from the old range(12)")
         rules = sum(len(c["instruction"]) for c in heavy)
-        assert rules / max(lengths) <= validate.CORRECTION_SHARE_CEILING, (
-            "the fattest assembly must read CLEAN here, or the old max-denominator "
-            f"code passes this test too: {rules / max(lengths):.0%}")
+        # The oracle is derived INDEPENDENTLY of the code's own window. Its first
+        # draft looped `range(12)` because the code did, which is the trap
+        # `validate._resolved_slots` names in this file: a checker that re-derives
+        # the rule it checks is testing its own copy.
+        pool = selector.resolved_pool(voice.active_exemplars(), "x", "post",
+                                      selector.DEFAULT_K)
+        assert len({selector._words(r) for r in pool}) == 1, (
+            "this fixture must be a single length register, or a length target "
+            "reaches the thin rows and the counter window stops mattering")
+        assert len(pool) > 12, f"rotation period is {len(pool)}, nothing to miss"
+        lengths = [len(assemble.voice_section(voice, "x", c)[0])
+                   for c in range(len(pool))]
+        assert lengths.index(min(lengths)) >= 12, (
+            f"the thinnest prompt is at counter {lengths.index(min(lengths))}, "
+            "inside a 12-counter window, so this fixture cannot tell the two apart")
+        assert rules / min(lengths[:12]) <= validate.CORRECTION_SHARE_CEILING, (
+            "a 12-counter window must read CLEAN here, or the old code passes "
+            f"this test too: {rules / min(lengths[:12]):.0%}")
         assert rules / min(lengths) > validate.CORRECTION_SHARE_CEILING, (
-            f"the thinnest assembly must read RED: {rules / min(lengths):.0%}")
+            f"the full rotation must read RED: {rules / min(lengths):.0%}")
 
         problems = [p for p in validate.check_correction_share(voice)
                     if p.startswith("x:")]
         assert problems, (
-            "the guard graded the fattest prompt and called a corpus clean that is "
-            f"{rules / min(lengths):.0%} rules at its thinnest")
+            "the guard stopped at counter 11 and called a corpus clean that is "
+            f"{rules / min(lengths):.0%} rules at counter "
+            f"{lengths.index(min(lengths))}")
         reported = int(re.search(r" of (\d+) chars", problems[0]).group(1))
         assert reported == min(lengths), (
-            f"the guard divided by {reported}; the thinnest assembly is "
-            f"{min(lengths)} and the fattest is {max(lengths)}")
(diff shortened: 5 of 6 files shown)


assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-301-20260904-192852.md` (6199 bytes).

```
FINDINGS:
major|check_correction_share divides by the largest assembly, so it reports [] while the live x-channel corpus is at 71% against a 70% ceiling|plugins/kipi-core/voiceloop/validate.py:359
minor|the new eligible_for_voice_reference retirement filter has no test; deleting it leaves the suite at 121 passed|plugins/kipi-core/voiceloop/selector.py:65
minor|zscores, _sigma and style_distance have no caller and no test; all three mutants survive|plugins/kipi-core/voiceloop/fingerprint.py:351
nit|style_distance's `if az >= 0` false branch is unreachable because az is min(abs(z), cap)|plugins/kipi-core/voiceloop/fingerprint.py:383
nit|worst_text is assigned and never read|plugins/kipi-core/voiceloop/validate.py:356
END FINDINGS
```

--- reviewer output ---

# Review: PR #301, voiceloop slice 0

Five findings survived reproduction. One is a new guard that is already blind on the live corpus.

---

## MAJOR: the correction-share guard divides by the fattest prompt, so it misses the live breach today

`plugins/kipi-core/voiceloop/validate.py:359`

`check_correction_share` takes the max over 12 counters and uses that as the denominator. The failure it exists to catch, rules crowding out his writing, is worst at the THINNEST assembly, not the largest. Taking the max minimizes the share and reports the best case.

The loop looks copy-pasted from `check_budget` at line 319, where `max` is correct (largest assembly vs budget cap). The extremum came along with the loop.

Reproducer, run against the real producer corpus at `/Users/assafkipnis/projects/consulting/q-consult/voice`, read-only:

```
python3 -c "from voiceloop import assemble, corpus, validate, channel_registry; ..."
```

Real output:

```
CEILING: 0.7  BUDGET_CHARS: 24000

linkedin: rules=11618 chars
  assembly len across 12 counters: min=17995 max=21823
  share the GUARD reports (rules/MAX len): 53%
  share at the THINNEST counter (#9): 65%

x: rules=10154 chars
  assembly len across 12 counters: min=14293 max=16994
  share the GUARD reports (rules/MAX len): 60%
  share at the THINNEST counter (#6): 71%

check_correction_share on the live corpus: []
```

The `x` channel is over the 0.70 ceiling right now, on 1 of 12 prompts, and the guard says clean. Same file with the denominator swapped to the thinnest assembly, same live corpus:

```
["x: corrections are 10154 of 14293 chars (71%) of the largest assembly, over the
70% ceiling. Retire a superseded correction ..."]
```

Two knock-ons. The docstring's measured headroom ("linkedin 57%, x 60%... set at 0.70 so it does not fire today") is the max-denominator reading, so the documented margin is roughly 10 points more generous than the real one. And the message text says "of the largest assembly," which becomes wrong once the extremum is fixed.

Exemplar text on this corpus runs 18 to 9564 chars, so the counter-to-counter spread is structural, not a constructed input.

---

## MINOR: the new selector retirement filter has no test

`plugins/kipi-core/voiceloop/selector.py:65`

Mutation, temp copy of the tree, repo untouched:

```
[BASELINE]                                              rc=0  121 passed in 0.19s
[MUTANT A: eligible_for_voice_reference filter DELETED] rc=0  121 passed in 0.20s
```

The line is load-bearing on the real corpus: 10 of 161 rows carry `False` and are supposed to leave every prompt. Nothing goes red if it disappears.

The filter itself is correct. I checked the producer, not the fixture: `eligible_for_voice_reference` is emitted as a genuine bool, 140 True and 10 False, no string or `0` variant that `is not False` would let through.

---

## MINOR: `zscores`, `_sigma`, `style_distance` have no caller and no test

`plugins/kipi-core/voiceloop/fingerprint.py:338,351,373`

Three independent mutants, all survive:

```
[MUTANT B2: zscores() -> always {}]        rc=0  121 passed in 0.23s
[MUTANT D: _sigma() -> always None]        rc=0  121 passed in 0.20s
[MUTANT C: style_distance() -> always 0.0] rc=0  121 passed in 0.21s
```

`grep -rn "zscore\|style_distance\|_sigma" plugins/kipi-core/voiceloop/tests/` returns nothing, and a repo-wide grep finds no caller outside their own definitions. Fifty-eight lines of new scoring API shipping fleet-wide with nothing exercising it.

---

## NIT: unreachable false branch in `style_distance`

`plugins/kipi-core/voiceloop/fingerprint.py:383`

`az = min(abs(z), cap)`, then `if az >= 0`. Tested across `-1e9, -8.0, -0.5, 0.0, 0.5, 8.0, 1e9, inf, -inf`:

```
z values (incl. infinities) that reach the FALSE branch of `if az >= 0`: []
```

---

## NIT: dead variable

`plugins/kipi-core/voiceloop/validate.py:356` — `worst_text` is assigned and never read.

---

## What is sound

- **The EXTERNAL_SOURCE split is really covered.** Reverting it so external rules fall back under the override header kills a test: `rc=1  1 failed`. Not decoration.
- **The share guard's wiring into `check_all` is really covered.** `check_correction_share` stubbed to `return []` kills two tests: `rc=1  2 failed`. The PR's two claimed self-tests both hold.
- **Fixture shape matches the producer.** I read the live corpus rather than trusting the tests. Field types line up; no invented-fixture trap on the retirement field.
- **Backward compatibility holds.** An unmarked corrections corpus still renders one legacy block, and `external_correction_ids` comes back empty.
- **No divide-by-zero.** `if not worst_len: continue` covers the empty assembly.
- **The budget raise is honest.** Live linkedin max is 21823, so 20000 genuinely no longer fits; the raise is load-bearing, not a gate quietly switched off.
- **Nothing here runs at 3am.** Everything added is pure and suite-time. No launchd job, no writer, no concurrency surface.

---

## VERDICT: REQUEST CHANGES

Not BLOCK. Nothing here is permanent or unrecoverable, and the fix is one line.

**Fix first:** the denominator in `check_correction_share` (`validate.py:359`). Track the thinnest assembly, not the largest, and reword "largest assembly" in the message. A new guard that returns `[]` on a corpus already over its own ceiling teaches the operator the alarm is clean, which is the cost that lands later.

FINDINGS:
major|check_correction_share divides by the largest assembly, so it reports [] while the live x-channel corpus is at 71% against a 70% ceiling|plugins/kipi-core/voiceloop/validate.py:359
minor|the new eligible_for_voice_reference retirement filter has no test; deleting it leaves the suite at 121 passed|plugins/kipi-core/voiceloop/selector.py:65
minor|zscores, _sigma and style_distance have no caller and no test; all three mutants survive|plugins/kipi-core/voiceloop/fingerprint.py:351
nit|style_distance's `if az >= 0` false branch is unreachable because az is min(abs(z), cap)|plugins/kipi-core/voiceloop/fingerprint.py:383
nit|worst_text is assigned and never read|plugins/kipi-core/voiceloop/validate.py:356
END FINDINGS

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-301-20260904-193905.md` (9335 bytes).

```
FINDINGS:
major|check_pools counts raw rows, blind to the new eligible_for_voice_reference filter, so check_all returns empty on a corpus whose every prompt has zero exemplars|plugins/kipi-core/voiceloop/validate.py:126
major|check_correction_share samples only counters 0 through 11, so for pools over 12 rows it is not the rotation minimum and returns clean on a corpus whose true share is 0.917 against a 0.70 ceiling|plugins/kipi-core/voiceloop/validate.py:378
minor|zscores and style_distance ship with zero callers and zero tests; gutting both bodies leaves the 122-test suite green|plugins/kipi-core/voiceloop/fingerprint.py:351
minor|the colon_rate sigma floor of 0.35 is in the wrong scale for a per-100-words rate, so one colon in a short post yields z of 31.53, the single-metric theater the comment claims the floor prevents|plugins/kipi-core/voiceloop/fingerprint.py:333
nit|the az greater-or-equal-zero guard in style_distance has an unreachable false branch since az is min of abs z and cap, and zscores cannot emit NaN|plugins/kipi-core/voiceloop/fingerprint.py:383
END FINDINGS
```

--- reviewer output ---

# Review: PR #301, voiceloop slice 0

Read-only. Nothing in the repo was modified; every reproducer ran against a copy in `/tmp`.

Note on the evidence blocks below: the voice gate rejects raw program output, so each figure is reproduced as a labeled line. Every number is verbatim from the runs.

---

## MAJOR 1 — the new eligibility filter is invisible to the starvation check, so `check_all` goes fully silent on a zero-exemplar corpus

`selector.py:65` teaches `eligible()` to drop rows carrying `eligible_for_voice_reference: false`. `validate.check_pools` at `validate.py:126` still counts raw rows off `voice.active_exemplars()`, so it grades a pool the selector can't reach.

`check_rotation_headroom` hands the empty-pool case to it explicitly at `validate.py:259`: `if not pool: continue   # an empty pool is check_pools' story`. That story can no longer be told.

**File:** `plugins/kipi-core/voiceloop/validate.py:126`

Reproducer, six post rows all curated out:

```
Active post rows on disk: 6
Selector pool for linkedin: 0
Counter 0 gave exemplar_ids [] and no POSTS-HE-HAS-WRITTEN block
Counter 5 gave exemplar_ids [] and no POSTS-HE-HAS-WRITTEN block
Counter 11 gave exemplar_ids [] and no POSTS-HE-HAS-WRITTEN block
Result of validate.check_all() was the empty list
Verdict: SILENT, bug reproduced
```

The window, sweeping how many of six rows are flipped:

```
Flipped 0: reachable pool 6, check_pools silent, check_all returned 0 problems
Flipped 1: reachable pool 5, check_pools silent, check_all returned 0 problems
Flipped 2: reachable pool 4, check_pools silent, check_all returned 4 problems
Flipped 3: reachable pool 3, check_pools silent, check_all returned 4 problems
Flipped 4: reachable pool 2, check_pools silent, check_all returned 4 problems
Flipped 5: reachable pool 1, check_pools silent, check_all returned 4 problems
Flipped 6: reachable pool 0, check_pools silent, check_all returned 0 problems
```

`check_pools` is wrong at every level, and rotation-headroom masks it at 2 through 5. At 6 nothing fires: every prompt for that channel ships with no writing samples, and the validator says the corpus is healthy.

Honest caveat: no producer in *this* repo writes that field. `selector.py:52-57` says the live consulting corpus carries it and curation flips it, which is the shape I built from.

---

## MAJOR 2 — `check_correction_share` samples `range(12)`, so it is not the rotation minimum for any pool over 12 rows

`validate.py:378` copies `check_budget`'s 12-counter loop. For a pool of 20 the rotation has period 20, so counters 12 through 19 are never assembled. The comment above it claims the thinnest prompt a rotation produces.

**File:** `plugins/kipi-core/voiceloop/validate.py:378`

```
Channel x, reachable pool 20 rows, rotation period 20 counters
Counters the guard samples: range(12)
Thinnest over range(12): 6948 chars, share 0.691
Thinnest over the full rotation: 5237 chars, share 0.917
Ceiling constant CORRECTION_SHARE_CEILING: 0.70
Guard output for channel x: the empty list
```

The guard returns clean on a corpus whose real prompts are 0.917 rules by share, well over its own 0.70 ceiling. This is the same class as the bug commit `fb68c688` just fixed: right extremum now, wrong sample.

The new test can't see it. `test_the_share_is_measured_at_the_THINNEST_prompt_not_the_fattest` builds `_rows(20)` and computes its own oracle with `for c in range(12)`, the same window as the code under test. Pools of 12 or fewer are fully covered; `check_budget` at `validate.py:319` carries the same limitation and predates this PR.

---

## MINOR 3 — `zscores` and `style_distance` have zero callers and zero tests; both mutants survive

76 new lines in `fingerprint.py`. Grep across the repo finds only the definitions.

**File:** `plugins/kipi-core/voiceloop/fingerprint.py:351`

Mutation, both function bodies replaced with `raise AssertionError`, run in a `/tmp` copy:

```
MUTANT: zscores and style_distance replaced with raise
Suite result: 122 passed in 0.19s
```

The docstring at `fingerprint.py:314-320` names a live failure: a draft at colon_rate 1.29 scored inside the band and every layer reported clean. Nothing calls the fix, so that failure is still live.

Same mutation on the selector line:

```
MUTANT: the new eligible_for_voice_reference filter deleted from selector.eligible
Suite result: 122 passed in 0.18s
```

---

## MINOR 4 — the `colon_rate` sigma floor is 35x too small to do what its comment says

`_MIN_SIGMA` claims: "without a floor, one colon in a short post produces an enormous z and the aggregate becomes single-metric theater. Floors are in the metric's OWN units."

`colon_rate` is `text.count(":") * 100 / words` (`fingerprint.py:135`), a per-100-words rate whose step for a short post is around 12. The floor is 0.35.

**File:** `plugins/kipi-core/voiceloop/fingerprint.py:333`

```
Smoke run on an 8-word post with one colon:
Colon_rate z: 31.53
Sentence_mean z: -2.56
Style_distance aggregate: 8.4
```

One colon in an 8-word post gives z of 31.53, and the aggregate is 8.4, essentially the cap on that single metric. That is the single-metric theater the comment says the floor prevents. The other five floors are on raw counts and are fine.

---

## NIT 5 — unreachable false branch in `style_distance`

`if az >= 0:` after `az = min(abs(z), cap)`.

**File:** `plugins/kipi-core/voiceloop/fingerprint.py:383`

```
Probe of every input class, printing az and whether the branch is taken:
Z of -99 gave az 8.0, branch taken
Z of -1.0 gave az 1.0, branch taken
Z of 0.0 gave az 0.0, branch taken
Z of 1e+300 gave az 8.0, branch taken
Z of positive infinity gave az 8.0, branch taken
Z of negative infinity gave az 8.0, branch taken
Z of NaN gave az NaN, branch skipped, and this is the only False
```

`zscores` filters zero sigma and skips missing p50, so no NaN reaches it.

---

## What is sound

**Backward compatibility of the correction split is real, not asserted.** I hashed 36 assemblies (3 channels by 12 counters, plus provenance ids) against `origin/main` `f9f74ac1`'s `assemble.py` and `selector.py` on a corpus with no `source` field and a scoped correction. Identical:

```
Origin/main hash: eb4b0f9eb1c1f3578bd2939f51e7a008cc2432b06804f48d23a689e91938cdf5
PR 301 hash:      eb4b0f9eb1c1f3578bd2939f51e7a008cc2432b06804f48d23a689e91938cdf5
```

**Divide-by-zero in `check_correction_share` is genuinely closed.** `lengths = [n for n in lengths if n]` plus the `if not lengths: continue`. I tried to make one counter empty while another is not; assembly length is counter-independent for identity, pov and lexicon, so it is all-or-nothing and the guard skips correctly.

**The scope filter agrees between prompt and check.** `assemble.voice_section` and `check_correction_share` both read `voice.active_corrections()` and apply the same `scope` predicate, so the denominator counts what actually renders. External-source rows are counted, which matches the fact that they render.

**Concurrency and rollback.** Every module here is pure: no writes, no clock, no network. Two runs at once can't interfere, and the whole change reverts by reverting the commits.

**Ceiling not moved to clear the reading it exposed.** `CORRECTION_SHARE_CEILING` stays at 0.70 while the corrected arithmetic reads x at 0.71. That is the right call and the comment says so.

**The share guard is not the size guard restated.** I checked the fixture in `test_corrections_crowding_out_his_writing_is_red` actually sits inside budget; the test asserts it, and the assertion is load-bearing.

---

## VERDICT: REQUEST CHANGES

Two majors, neither permanent nor unrecoverable, so this is not a BLOCK. Everything is suite-time, pure, and revertible.

**Fix first:** MAJOR 2, `validate.py:378`. Sample the whole rotation, `range(max(12, len(pool)))`, not a fixed 12. It is a one-line change, and the test meant to pin the denominator currently computes its oracle with the same window as the code, so the test needs the fix too. MAJOR 1 is a second one-line change: route `check_pools` through `selector.resolved_pool` so it counts what the selector can reach, the way `_resolved_slots` already does.

FINDINGS:
major|check_pools counts raw rows, blind to the new eligible_for_voice_reference filter, so check_all returns empty on a corpus whose every prompt has zero exemplars|plugins/kipi-core/voiceloop/validate.py:126
major|check_correction_share samples only counters 0 through 11, so for pools over 12 rows it is not the rotation minimum and returns clean on a corpus whose true share is 0.917 against a 0.70 ceiling|plugins/kipi-core/voiceloop/validate.py:378
minor|zscores and style_distance ship with zero callers and zero tests; gutting both bodies leaves the 122-test suite green|plugins/kipi-core/voiceloop/fingerprint.py:351
minor|the colon_rate sigma floor of 0.35 is in the wrong scale for a per-100-words rate, so one colon in a short post yields z of 31.53, the single-metric theater the comment claims the floor prevents|plugins/kipi-core/voiceloop/fingerprint.py:333
nit|the az greater-or-equal-zero guard in style_distance has an unreachable false branch since az is min of abs z and cap, and zscores cannot emit NaN|plugins/kipi-core/voiceloop/fingerprint.py:383
END FINDINGS

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-301-20260904-195211.md` (7903 bytes).

```
FINDINGS:
major|correction-share and budget guards never sample target_words, the only shape voice_ref.py builds, so both read clean on an 88%-rules prompt|plugins/kipi-core/voiceloop/validate.py:387
minor|check_budget's range(12) misses the true maximum whenever the pool exceeds 12, and the BUDGET_CHARS raise in this PR was sized from that under-sampled measurement|plugins/kipi-core/voiceloop/validate.py:319
minor|the eligible_for_voice_reference filter has no test; deleting it leaves 122 of 122 green|plugins/kipi-core/voiceloop/selector.py:65
minor|a row retired from voice reference still feeds corpus_sha and the fingerprint bands drafts are scored against|plugins/kipi-core/voiceloop/validate.py:300
minor|zscores, style_distance and _sigma ship fleet-wide with zero callers and zero tests; all three mutants survive|plugins/kipi-core/voiceloop/fingerprint.py:351
nit|style_distance's `if az >= 0` false branch is unreachable because az is min(abs(z), cap)|plugins/kipi-core/voiceloop/fingerprint.py:381
END FINDINGS
```

--- reviewer output ---

## PR #301 review, adversarial pass

One-line state: the correction-share work is genuinely well tested (three mutants killed), and the guard it adds is blind on the one code path the package's CLI actually uses.

---

### FINDING 1 — major

**Claim:** `check_correction_share` and `check_budget` only ever sample `target_words=None`, but `voice_ref.py` declares `--words` as `required=True` and always passes it, which collapses the pool to the k nearest rows. The guard returns `[]` on a corpus whose real prompt is 88% corrections.

**File:** `plugins/kipi-core/voiceloop/validate.py:387` (and `voice_ref.py:54`, `voice_ref.py:80`)

**Why the axis matters:** `selector.select` at lines 169 to 178 replaces the whole pool with `ranked[:max(k,1)]` when `target_words` is set. On a three-register corpus, which `length_band`'s own docstring says is the live shape, a short target draws four tweets and the prompt collapses to a few hundred characters of his writing under thousands of characters of rules.

**Reproducer:** built a 12-row x corpus, six long-form and six tweet-length, ids interleaved, 3288 bytes of active corrections. Ran `validate.check_correction_share` against it, then rendered the prompt the CLI builds.

**Real output:**

```
ceiling 70%, correction bytes 3288

GUARD VERDICT for channel x: []
  (its sample: target_words=None, thinnest 7165 chars -> 46%)

WHAT voice_ref.py ACTUALLY BUILDS (--words is required=True):
  --words 25   thinnest prompt  3723 chars -> corrections  88%  OVER THE 70% CEILING
  --words 60   thinnest prompt  3723 chars -> corrections  88%  OVER THE 70% CEILING
  --words 150  thinnest prompt  3723 chars -> corrections  88%  OVER THE 70% CEILING
  --words 480  thinnest prompt 10607 chars -> corrections  31%  ok
```

**Honest limit:** the id interleaving is constructed, chosen so the `None` path reads clean. The underlying gap does not depend on the fixture: neither check passes `target_words` at any value, ever, on any corpus.

---

### FINDING 2 — minor

**Claim:** `check_budget`'s `range(12)` under-samples the rotation, so its comment "one rotation lap is enough to find the max" is false whenever the pool exceeds 12. The new comment 60 lines below states the live pools are 45, 31 and 10.

**File:** `plugins/kipi-core/voiceloop/validate.py:319`

**Extra weight:** this PR raises `BUDGET_CHARS` from 20000 to 24000, and `assemble.py:36` says that number came from "rendering 12 counters each way". The raise was sized against a measurement the adjacent new comment calls wrong.

**Reproducer:** 20-row corpus with descending body lengths so the fattest assembly lands at counter 16, `BUDGET_CHARS` placed between the 12-window max and the real max.

**Real output:**

```
rotation period (pool size): 20
max over the FULL rotation : 6494  (at counter 16)
max over check_budget's 12 : 4754

BUDGET_CHARS set to 5624 -- above the 12-window max, below the real max
check_budget says: []
reality          : largest assembly is 6494 chars, over budget by 870
```

---

### FINDING 3 — minor

**Claim:** the `eligible_for_voice_reference` filter, which changes what rides in every prompt fleet-wide, ships with zero test coverage. Deleting the line leaves the suite fully green.

**File:** `plugins/kipi-core/voiceloop/selector.py:65`

**Reproducer:** copied the package to a temp dir, deleted that one line, ran the suite.

**Real output:**

```
baseline                         rc=0  122 passed in 0.19s
drop-eligible-filter             rc=0  122 passed in 0.19s
```

---

### FINDING 4 — minor

**Claim:** retirement is half-wired. A row marked `eligible_for_voice_reference: false` leaves selection but still feeds `check_fingerprint_fresh`'s `corpus_sha` and therefore the fingerprint bands that drafts are scored against. The hype row the docstring names as the scar still defines his style bounds.

**File:** `plugins/kipi-core/voiceloop/validate.py:300`

**Real output:**

```
eligible_for_voice_reference=<absent>
  rows selection can reach      : 6   hype row present: True
  texts feeding the FINGERPRINT : 6   hype text present: True
  check_fingerprint_fresh's sha : 2f89faa371c377dbccdb24b9668618fe

eligible_for_voice_reference=False
  rows selection can reach      : 5   hype row present: False
  texts feeding the FINGERPRINT : 6   hype text present: True
  check_fingerprint_fresh's sha : 2f89faa371c377dbccdb24b9668618fe
```

Identical sha, so nothing ever asks for a regeneration.

---

### FINDING 5 — minor

**Claim:** `zscores`, `style_distance` and `_sigma` ship to every instance with zero callers and zero tests. All three mutants survive.

**File:** `plugins/kipi-core/voiceloop/fingerprint.py:351`

**Real output:**

```
zscores-emits-nothing            rc=0  122 passed in 0.19s
style-distance-always-zero       rc=0  122 passed in 0.20s
sigma-always-none                rc=0  122 passed in 0.19s
```

Repo-wide grep for `zscores` or `style_distance` outside `fingerprint.py` returns nothing.

---

### FINDING 6 — nit

**Claim:** `if az >= 0:` guards a branch that cannot be false, because `az = min(abs(z), cap)` on the line above.

**File:** `plugins/kipi-core/voiceloop/fingerprint.py:381`

**Real output:** `style_distance({'colon_rate': -99.0, 'hedge_count': 99.0}) = 11.31`, both signs accumulate, no input reaches the else.

---

## What is sound

- The correction-share arithmetic is real. Reverting `min` to `max` kills a test, reverting the full-rotation loop to `range(12)` kills a test, unwiring the check from `check_all` kills two. The negative self-test in the PR body reproduces exactly.
- The external-source split is real. `if False and researched` kills a test.
- Empty corpus, a correction with no `instruction` key, and a `scope` naming a channel that does not exist all return `[]` with no crash and no divide-by-zero. The `lengths = [n for n in lengths if n]` filter earns its comment.
- The rotation period derivation is correct: `select` offsets by `counter % len(pool)` and anchor choice is downstream of that offset, so the period really is the pool size.
- Provenance now filters once and feeds both the prompt and the receipt from the same list. `correction_ids` and `external_correction_ids` agreed on every case I tried.

## VERDICT: REQUEST CHANGES

Not BLOCK. Nothing here is permanent or unrecoverable; the worst case is a silent guard and a correction ledger a human prunes later.

**Fix first:** Finding 1. Both sampling bugs already fixed in this PR and both bugs found today are the same class, wrong sample of the assembly space, on the third and fourth round. The structural fix is one shared sampler that enumerates the counter rotation *and* the `target_words` registers the CLI can request, consumed by `check_budget` and `check_correction_share` alike, rather than a third patch to one loop.

FINDINGS:
major|correction-share and budget guards never sample target_words, the only shape voice_ref.py builds, so both read clean on an 88%-rules prompt|plugins/kipi-core/voiceloop/validate.py:387
minor|check_budget's range(12) misses the true maximum whenever the pool exceeds 12, and the BUDGET_CHARS raise in this PR was sized from that under-sampled measurement|plugins/kipi-core/voiceloop/validate.py:319
minor|the eligible_for_voice_reference filter has no test; deleting it leaves 122 of 122 green|plugins/kipi-core/voiceloop/selector.py:65
minor|a row retired from voice reference still feeds corpus_sha and the fingerprint bands drafts are scored against|plugins/kipi-core/voiceloop/validate.py:300
minor|zscores, style_distance and _sigma ship fleet-wide with zero callers and zero tests; all three mutants survive|plugins/kipi-core/voiceloop/fingerprint.py:351
nit|style_distance's `if az >= 0` false branch is unreachable because az is min(abs(z), cap)|plugins/kipi-core/voiceloop/fingerprint.py:381
END FINDINGS

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-301-20260904-200425.md` (5785 bytes).

```
FINDINGS:
major|check_correction_share pins slot_kind="post", so a comment slot at 97% corrections reads clean|plugins/kipi-core/voiceloop/validate.py:396
minor|check_budget samples only range(12) at target_words=None, a shape voice_ref.py never builds, and this PR raises BUDGET_CHARS off that measurement|plugins/kipi-core/voiceloop/validate.py:319
minor|the correction `source` field has no validator, so a typo renders an external shape under the STANDING CORRECTIONS override header|plugins/kipi-core/voiceloop/validate.py:106
minor|deleting the eligible_for_voice_reference filter leaves the suite 123/123 green|plugins/kipi-core/voiceloop/selector.py:65
minor|zscores and style_distance have no caller and no test in this repo; gutting both to constants leaves the suite green|plugins/kipi-core/voiceloop/fingerprint.py:351
nit|`if az >= 0` in style_distance has no reachable false branch for any non-negative cap|plugins/kipi-core/voiceloop/fingerprint.py:383
END FINDINGS
```

--- reviewer output ---

Review done. Six findings survived reproduction, one major.

## MAJOR: the share guard pins `slot_kind="post"`, and `comment` is the fourth axis of the same class

`plugins/kipi-core/voiceloop/validate.py:396` hardcodes `"post"`. `voice_ref.py:59` declares `--slot-kind {post,comment}` and `validate.SLOT_KINDS` in the same file calls both "every slot a caller can actually ask for". Three review rounds already fixed counter, then target_words, then the extremum. Slot kind is the one left.

Reproducer, run against the branch worktree, 12 long post rows plus 6 short comment rows:

```
x         slot_kind=post     thinnest= 19757  share= 63%  ceiling= 70%  clean
x         slot_kind=comment  thinnest= 12877  share= 97%  ceiling= 70%  RED
linkedin  slot_kind=post     thinnest= 19757  share= 63%  ceiling= 70%  clean
linkedin  slot_kind=comment  thinnest= 12877  share= 97%  ceiling= 70%  RED

check_correction_share() -> []
validate.SLOT_KINDS       = ('post', 'comment')
```

A corpus whose comment prompts are 97 percent rules reads green. This needs comment-kind rows to bite, and `corpus.py:29` records that a real founder-written Reddit comment was banked as an exemplar on 2026-08-29, so the input is one a producer emits.

## MINOR: `check_budget` grades a prompt shape production never builds

`validate.py:319` loops `range(12)` at `target_words=None`. Same argument the share guard just won three times. This PR raises `BUDGET_CHARS` from 20000 to 24000 off a measurement taken through that window.

```
BUDGET_CHARS = 24000
check_budget() -> []
largest assembly, the shape check_budget grades (12 counters, no target): 7525
largest assembly, the shape voice_ref.py --words 1382 actually builds: 26924
over budget by: 2924
```

## MINOR: `source` has no validator, so a typo promotes a researched shape into the override block

`check_corrections` (`validate.py:106` region) checks `class`, `status` and `scope` against vocabularies. `source` is unchecked, and `assemble.py:105` compares it by exact string.

```
check_corrections() -> []

STANDING CORRECTIONS (these override everything above):
- End on a colon line (typo).
...
external_correction_ids = ['good']
```

One underscore instead of one hyphen and an external hypothesis renders under "these override everything above", which is the inversion the split exists to prevent.

## MINOR: two mutants survive the suite

Copied the package to a temp tree, mutated the copy, ran pytest against the copy:

```
baseline (unmutated copy)                                  rc=0  123 passed in 0.20s
MUTANT A: delete the eligible_for_voice_reference filter   rc=0  123 passed in 0.20s
MUTANT B: gut zscores() and style_distance() to constants  rc=0  123 passed in 0.20s
restored copy                                              rc=0  123 passed in 0.22s
```

Mutant A is `selector.py:65`, a live-path filter deciding what rides in every prompt, whose own docstring names the scar it fixes. Mutant B is `fingerprint.py:351` and `fingerprint.py:373`, 76 new lines with zero callers and zero tests anywhere in this repo. Grep across the whole tree returns only their own definitions.

## NIT: unreachable false branch

`fingerprint.py:383`, `if az >= 0` where `az = min(abs(z), cap)`. False only if a caller passes a negative cap. Dead as written.

## What is sound

Four attacks the code survived.

The claim that `target_words = min(words)` reaches the thinnest prompt held. 300 random corpora, sweeping every distinct row length plus 0, 1 and 1000000 as targets: zero corpora where any other target produced a thinner prompt than the guard's two samples.

Cost is not a problem. On a 300-row corpus `check_correction_share` runs 0.52 seconds against `check_budget` at 0.00. Suite-time only, no runtime path, nothing at 3am touches it.

The zero-length guard is real. A channel that assembles to nothing is dropped by the `lengths = [n for n in lengths if n]` filter rather than dividing by zero. I fed it an empty pool and got a skip, not a crash.

Leaving the ceiling at 0.70 while the corrected arithmetic exposes x at 71 percent is a deliberate, documented decision by the voice owner. Nothing in this repo calls `check_all` outside tests, so nothing scheduled goes red on merge. Not a finding.

Two prose discrepancies, code is truth: the PR body says version 1.9.5 to 1.9.6 and the diff says 1.9.9; the body says 121 passed and the branch head passes 123.

## VERDICT: REQUEST CHANGES

Not BLOCK. Nothing here is permanent or unrecoverable, and the merged file is genuinely better than either side it reconciles.

Fix first: put `slot_kind` in the `check_correction_share` loop next to `counter` and `target_words`. It is the same one-line shape as the previous two rounds, and `validate.checked_slots()` already hands you the pairs.

FINDINGS:
major|check_correction_share pins slot_kind="post", so a comment slot at 97% corrections reads clean|plugins/kipi-core/voiceloop/validate.py:396
minor|check_budget samples only range(12) at target_words=None, a shape voice_ref.py never builds, and this PR raises BUDGET_CHARS off that measurement|plugins/kipi-core/voiceloop/validate.py:319
minor|the correction `source` field has no validator, so a typo renders an external shape under the STANDING CORRECTIONS override header|plugins/kipi-core/voiceloop/validate.py:106
minor|deleting the eligible_for_voice_reference filter leaves the suite 123/123 green|plugins/kipi-core/voiceloop/selector.py:65
minor|zscores and style_distance have no caller and no test in this repo; gutting both to constants leaves the suite green|plugins/kipi-core/voiceloop/fingerprint.py:351
nit|`if az >= 0` in style_distance has no reachable false branch for any non-negative cap|plugins/kipi-core/voiceloop/fingerprint.py:383
END FINDINGS

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-301-20260904-201318.md` (8415 bytes).

```
FINDINGS:
minor|The new eligible_for_voice_reference filter has no test: deleting the line leaves the suite 124/124 green|plugins/kipi-core/voiceloop/selector.py:65
minor|check_correction_share pins k=DEFAULT_K while voice_ref.py declares -k, so it reads CLEAN at 62% on a corpus whose real -k 2 prompt is 75% rules|plugins/kipi-core/voiceloop/validate.py:402
minor|The derived target ignores the LONG_WORDS pool promotion, falsifying the comment that no other target draws shorter rows; guard CLEAN while --words 150 emits an 86%-rules prompt|plugins/kipi-core/voiceloop/validate.py:409
minor|fingerprint.zscores and style_distance add 76 lines with no caller and no test; replacing both bodies with a raise leaves the suite green|plugins/kipi-core/voiceloop/fingerprint.py:351
minor|Neither new corpus field is schema-validated, so a typo'd source renders a researched shape under the override header and a string "false" keeps a retired row in every prompt|plugins/kipi-core/voiceloop/validate.py:80
nit|style_distance's `if az >= 0` false branch is unreachable because az = min(abs(z), 8.0)|plugins/kipi-core/voiceloop/fingerprint.py:383
END FINDINGS
```

--- reviewer output ---

# Review: PR #301 — voiceloop slice 0, skeleton side

Read-only. Nothing in the repo was modified; every mutation ran against a `shutil.copytree` copy in a temp dir.

---

## Findings

### 1. minor — the one behavior change in `selector.py` has zero test coverage

`plugins/kipi-core/voiceloop/selector.py:65`

The `eligible_for_voice_reference` filter is the line whose own docstring records the scar: the field was dead, curation flipped it, selection never read it, and a hype-register post sat as an anchor in every linkedin prompt. Deleting the line again leaves the suite fully green, so the next edit that re-deads it says nothing.

Reproducer (one of seven mutants run in one pass):

```
drop check_correction_share from check_all    -> 2 failed, 122 passed in 0.24s
counter window back to a fixed 12             -> 1 failed, 123 passed in 0.26s
slot_kind pinned back to post                 -> 1 failed, 123 passed in 0.25s
target_words back to None only                -> 1 failed, 123 passed in 0.28s
thinnest -> fattest (min back to max)         -> 3 failed, 121 passed in 0.28s
eligible_for_voice_reference filter removed   -> 124 passed in 0.25s
EXTERNAL_SOURCE split collapsed               -> 1 failed, 123 passed in 0.26s
```

Every other mutant goes red. That one survives.

### 2. minor — `check_correction_share` pins `k`, and `voice_ref.py` declares `-k`

`plugins/kipi-core/voiceloop/validate.py:402`

`voice_ref.py:56` declares `-k` and passes it to `assemble.voice_section` at line 82. The guard sweeps counter, target and slot_kind, all derived, and then hardcodes `selector.DEFAULT_K`. Fewer exemplars means a thinner prompt means a higher share. Same argument round 4 used for `--slot-kind`.

```
ceiling                    : 70%
check_correction_share says: CLEAN
  -k 4: thinnest prompt 10121 chars, corrections 6240 = 62% -> green
  -k 3: thinnest prompt 9237 chars, corrections 6240 = 68% -> green
  -k 2: thinnest prompt 8353 chars, corrections 6240 = 75% -> RED
  -k 1: thinnest prompt 7469 chars, corrections 6240 = 84% -> RED
```

Single length register on purpose, so the target axis cannot reach the thin prompt on its own. No in-repo caller passes a non-default `-k` today, which is why this is minor and not major.

### 3. minor — the derived target misses the `LONG_WORDS` promotion, and the comment claiming otherwise is false

`plugins/kipi-core/voiceloop/validate.py:409`, comment at 415-417

The comment says the shortest register is the target that collapses the pool hardest, "`length_band` ranks by distance, so no other target draws shorter rows." That holds only for a fixed pool. A target at or above `LONG_WORDS` promotes `article-excerpt` into the primary tier and **changes** the pool, so it can reach rows that were never in `base` and therefore never in the derived minimum.

```
check_correction_share says: CLEAN
guard samples target=None and target=1121 (its derived minimum)
--words 150 picks           : ['a-00', 'a-01', 'a-02', 'a-03']
--words 150 thinnest prompt : 8325 chars, corrections 7200 = 86% -> RED
```

Constructed corpus (posts longer than excerpts), so minor. The falsified invariant in the comment is the durable part.

### 4. minor — `fingerprint.zscores` and `style_distance` have no caller and no test

`plugins/kipi-core/voiceloop/fingerprint.py:351` and `:373`

76 new lines shipped into a fleet-propagated plugin. Grep finds no reference outside `fingerprint.py`. Inserting `raise AssertionError` as the first statement of both functions:

```
124 passed in 0.22s
```

`wiring-check.md` calls this dead weight. It needs a consumer or it needs to wait for one.

### 5. minor — neither new corpus field is schema-validated, and both fail silently in the wrong direction

`plugins/kipi-core/voiceloop/validate.py:80` (`check_corrections`) and `:27` (`check_exemplars`)

`check_corrections` allowlists `class` and `status`. It does not look at `source`. `check_exemplars` does not look at `eligible_for_voice_reference`. One character of drift and both features quietly invert.

```
corrections schema check  : CLEAN
exemplars schema check    : CLEAN
research rule under the OVERRIDE header? True
external_correction_ids   : []
row marked not-eligible still picked?   : True
its hype text in the prompt?            : True
```

Input was `source: "external_performance"` (underscore, not hyphen) and `eligible_for_voice_reference: "false"` (JSON string, not boolean). A research-derived shape gets presented to the model as the founder's standing override, and a retired row rides in the prompt, which is the scar each field was added to close.

### 6. nit — `style_distance` has an unreachable false branch

`plugins/kipi-core/voiceloop/fingerprint.py:383`

`az = min(abs(z), cap)` with `cap=8.0`, then `if az >= 0`. `abs()` is never negative. Constant wearing a conditional's syntax.

---

## What is sound

Attacks that failed, named:

- **The extremum fix is real.** `min` reverted to `max` goes red on 3 tests. Not decoration.
- **The three prior-round axes are genuinely derived, not literals.** Reverting counter to `range(12)`, slot_kind to `("post",)`, and target to `(None,)` each go red on exactly one test.
- **The wiring is pinned.** Deleting the `check_correction_share` call from `check_all` goes red on 2 tests.
- **The `EXTERNAL_SOURCE` split is pinned both ways.** Collapsing `researched` to `[]` goes red; the unmarked-corrections legacy test holds the byte-for-byte contract.
- **The target derivation is right within a fixed pool.** I checked the ordering claim: distance from the minimum is ascending order by value, so the k nearest really are the k smallest. Finding 3 is about the pool changing, not about that ranking.
- **`slot_index` is not a missing axis.** `selector.select` computes `offset = (counter + slot_index) % len(pool)`, so sweeping counter over the full period already covers every slot_index. I looked for this and it is not a hole.
- **No founder corpus in the tree.** `find` for `exemplars.jsonl` / `corrections.jsonl` returns nothing, so the public-repo leak boundary holds.
- **Cost is not a problem.** The sweep is bounded by pool size per slot_kind, pure string building, and the full suite runs in 0.22s.
- **Tried and dropped:** the ceiling being knowingly red on the live corpus at 71%. I looked for an in-repo consumer of `check_all` that would page someone or block a job and found none, so I have no reproducer and no finding. Worth knowing the instance sync is where that lands.

---

## VERDICT: APPROVE WITH NITS

Suite is 124 passed, not the 121 in the PR body. Version is 1.9.10, not the 1.9.6 the body claims. Neither changes anything.

**Fix first: finding 1.** A retirement mechanism whose own docstring records the scar of it being dead, shipped with no test that notices when it goes dead again, is the one thing here that costs a future session real time.

**On findings 2 and 3:** this is round 5 of one class. Four rounds have each patched one axis. The structural move is to stop enumerating axes by hand and make the sweep read `voice_ref.py`'s declared argument surface once, so a new CLI flag cannot arrive without the guard seeing it. Patching `-k` and then `LONG_WORDS` individually buys round 6.

FINDINGS:
minor|The new eligible_for_voice_reference filter has no test: deleting the line leaves the suite 124/124 green|plugins/kipi-core/voiceloop/selector.py:65
minor|check_correction_share pins k=DEFAULT_K while voice_ref.py declares -k, so it reads CLEAN at 62% on a corpus whose real -k 2 prompt is 75% rules|plugins/kipi-core/voiceloop/validate.py:402
minor|The derived target ignores the LONG_WORDS pool promotion, falsifying the comment that no other target draws shorter rows; guard CLEAN while --words 150 emits an 86%-rules prompt|plugins/kipi-core/voiceloop/validate.py:409
minor|fingerprint.zscores and style_distance add 76 lines with no caller and no test; replacing both bodies with a raise leaves the suite green|plugins/kipi-core/voiceloop/fingerprint.py:351
minor|Neither new corpus field is schema-validated, so a typo'd source renders a researched shape under the override header and a string "false" keeps a retired row in every prompt|plugins/kipi-core/voiceloop/validate.py:80
nit|style_distance's `if az >= 0` false branch is unreachable because az = min(abs(z), 8.0)|plugins/kipi-core/voiceloop/fingerprint.py:383
END FINDINGS

PR #302, opened 2026-09-05
knowledge-supply: prompt-conditioned reader for instance stores with receipts

assafkip:
## What

The read side of `kb-graph-guard`. Every instance store already had a writer and a freshness check; nothing read any of it when a question arrived. Measured before: 0 bytes injected for an entity question in every instance.

`knowledge_supply.py` resolves the entities and task class a question names, reads `graph.jsonl`, `relationships.md`, canonical, decisions, `commitments.jsonl`, `granola-cache.json`, `open-loops.json` and `last-handoff.md` through the source classes `knowledge-sources.json` declares for that class, and returns verbatim excerpts with `path:line`, a date and a status label (KNOWN / STALE / CONFLICTING / UNVALIDATED, from the one provenance vocabulary) plus a receipt naming every declared source as searched, empty, unreadable or absent. `knowledge-inject.py` (UserPromptSubmit, both settings files) renders it with a FULL / PARTIAL coverage line first. Zero bytes when no entity resolves; one line when an entity resolves and every declared store holds nothing. Always exits 0.

Plan: `q-system/output/plans/knowledge-supply-2026-09-04.md` (instance-local, not in this diff).

## Measured

- 55 tests; 31 mutations on copies each go red.
- Replay of the last 30 days of real questions, read-only. Largest instance: 373 questions, fire rate 1.9% before the first-name rule, 10.2% after (temporal 13, writing 14, lookup 10, commitment 1), payload median 815 chars, max 7,239, p95 21 ms. Consulting root: 1,758 questions, 2.3%, payload median 2,372, max 7,660, p95 12 ms (its transcripts are build instructions and its graph is empty).
- Hook-level probe in the exact settings command shape against the skeleton's own stores: exit 0, `hookEventName` present, first line named the absent graph, canonical lines with `path:line`, receipt written.
- Real-data defects the fixture hid, each fixed with a test: supersession on accumulative predicates, the live commitment state vocabulary, a low-tier line slipping into the gap after a higher-tier cut, bare first names, two more unvalidated marker forms.

## Review rounds

Three rounds of the Opus fallback reviewer (Codex timed out each time). Every finding was reproduced by the reviewer and fixed with a test that goes red under mutation on a copy. No finding class repeated across rounds.

- Round 1 (f0e8a370): per-class cap starved every entity after the first; 2-char aliases matched English words in store content; an all-corrupt commitments ledger read as empty under FULL; first names fired line-initial; the fire-alone manifest knob was inert.
- Round 2 (c88ecf8e): first-name expansion ran only when nothing else resolved; pinned items could run past the ceiling under cut=0; the provenance vocabulary was re-parsed per matched line.
- Round 3 (f6ba2da4, message in 1f1a06ea): the misses ledger grew unbounded from one pasted transcript (1,600 rows); separators and footer were outside the ceiling accounting; a resolved entity with nothing in any store injected an 800-char header; resolution was O(text x index), 7.1 s on a 109 KB paste against 6,000 entities.

- Round 4 (1f1a06ea, the cap): one major, three minors. The major (O(entities x store rows): ~65 resolved names kill the hook at its 5 s timeout with nothing injected and no receipt) is the same class as round 3's major. A repeated class means the fix shape was wrong, so 8fc5a23b is structural: a wall-clock deadline inside the supply pass (3.5 s default, `KNOWLEDGE_SUPPLY_DEADLINE_S`) that returns what was gathered plus a receipt naming where it stopped and which classes were never searched (a required one drops coverage to PARTIAL), and an entity cap of 12 with the dropped names in the receipt and header. The three minors are captured as spillover, not fixed here: sp-67d54572 (CONFLICTING assigned when the two newest rows agree), sp-c9b6401d (non-overlapping bigram scan in the misses ledger), sp-d830d71e (no per-item size cap on a pinned triple; `overflow` reports it).

- Round 5 (8fc5a23b): one major, three minors, all new. The major (single-token contact names fired sentence-initial; the guard sat on the first-name path only) repeated round 1 minor 4's class, so the loop stopped at the declared cap. Captured and promoted: ASK-1261 with a DoR.
- ASK-1261 (was #304, one review round per its DoR): one chokepoint `single_token_hit` for every single-token resolution with the initial-position rule inside it, pinned by a grep-the-tree test for a single call site and a three-path test; plus the six captured minors from rounds 4 and 5, each with a test. Its review round found three in-DoR items (unreadable rule decided from zero hits, casefold offset drift on the identifier path, the item cap cutting relationship blocks), fixed in b9ada7c5, and one beyond it, captured as sp-e103cfc5.
- Phase 2 (was #303): the recall producer and the write-side `src` / `alias_of` convention in the debrief template.

**Consolidated.** #303 and #304 were fast-forwarded into this branch (linear history, nothing rewritten) and closed, so this PR carries Phase 1, Phase 2 and ASK-1261 and merges once. 55 tests, 31 mutations red on copies. Auto-merge (squash) is armed; GitHub merges when `kipi/reviewer-approved` is green.

Codex has been out of workspace credits for every round (`.codex-failed` files: "Your workspace is out of credits"), so every verdict came from the Opus fallback in DEGRADED mode.

Commit f6ba2da4 carries a malformed message (two heredocs in one shell command collided). Force-push is off the table by the destructive-op contract, so 1f1a06ea is an empty commit with the intended message. Squash-merge collapses both.

## Checks

`settings-template-sync-check --check` 0 · `enforced-claim-lint --all` 0 · `hook_envelope_audit --hook` 0 · settings.json edited through `apply-claude-changes.sh` (hooks 55 to 56, gates held, tripwire re-baselined) · capability fragment added.

Fleet preview: the updater refuses to run from a feature branch, preview included (it fans `.claude/` and `plugins/` from the working tree). The leak gate reported clean. The preview runs after merge; the script sits under the same managed prefix as `lessons-inject.py`, and the template carries the line.

## Honest boundary

Proves the facts entered context, never that they were used (same boundary as lessons-inject). Supersession is decided only for state predicates, per project. Bare first names fire mid-sentence when they expand to exactly one entity; sentence-, line- and bullet-initial never. Only the first 12,000 chars of a question are scanned, and the receipt says when that happened. PARTIAL coverage at an instance whose meetings live at the consulting root is true, and the instance override manifest is the designed way to relax it.

## Spillover touched

sp-aed36c39 (no reader) resolves when this merges. Opened: sp-55fd5739 (instance self-entity not in the index). Still open from the audit: sp-4daf4890, sp-076fac26, sp-6ffdca75.


assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-302-20260904-210150.md` (9985 bytes).

```
FINDINGS:
major|Per-class cap runs across all entities before pinning, so every entity after the first gets zero facts while the header lists it as resolved|q-system/.q-system/scripts/knowledge_supply.py:986
major|entity_matches accepts 2-char aliases case-insensitively, so alias "DO" attributes every store line containing the word "do" to that person, including in outbound writing-class drafts|q-system/.q-system/scripts/knowledge_supply.py:391
major|Corrupt commitments.jsonl lines are counted into a key nothing reads, so an unreadable promise ledger reports present=True, problem=None under COVERAGE: FULL|q-system/.q-system/scripts/knowledge_supply.py:511
minor|First-name expansion fires on line-initial and bullet-initial tokens, contradicting the "sentence-initial stays out" comment and the docstring claim that a bare first name never fires|q-system/.q-system/scripts/knowledge_supply.py:365
minor|The manifest knob entity_kinds_that_fire_alone is inert for slug/client/capability because those kinds are hardcoded, and "loop" is a kind no code produces|q-system/.q-system/scripts/knowledge_supply.py:330
END FINDINGS
```

--- reviewer output ---

# Review: PR #302 — knowledge-supply

Read the diff, ran the suite (28 pass, 0.35s), and ran five independent reproducers against the shipped engine. Three findings survived reproduction at major, two at minor.

---

## MAJOR 1 — the per-class cap starves every entity after the first, and the pin comment claims otherwise

`q-system/.q-system/scripts/knowledge_supply.py:986`

`cap` is applied to `cls_items` **after** all entities have been concatenated in entity order, and **before** `assemble()` runs. Entity 1 eats the whole class budget, so entities 2..N have nothing left to pin. The `assemble()` docstring at line 785 says "the newest graph item per entity is pinned first so a cut can never drop the one fact most likely to be current." It can, and it does.

Repro: 14 graph rows for person 1, one each for persons 2 and 3, where both of theirs are the newest rows in the store.

```
Entities resolved: ['Dana Okafor', 'Priya Raman', 'Tomas Lind']
Coverage verdict : PARTIAL
Items for Dana Okafor : 12
Items for Priya Raman : 0
Items for Tomas Lind  : 0
Footer           : [knowledge-supply] delegated: lessons -> lessons-inject, voice -> voice-dna-loader. cut=4 receipt=q-fix/memory/.knowledge-supply-receipts.jsonl
Oldest kept      : Dana Okafor cares_about topic 2
```

Priya's 2026-09-01 pushback and Tomas's 2026-09-02 ownership are both newer than every line that got in. The rendered header names all three people as resolved, then hands the model facts about one. A model reading that concludes the store is empty for the other two. This is the PR's own motivating query shape ("who at 14 Peaks did we talk to and what did they push back on"), so multi-person is the normal case, not the corner.

The only signal is `cut=4`, an integer that names no entity.

---

## MAJOR 2 — a two-character alias matches common English words in store content

`q-system/.q-system/scripts/knowledge_supply.py:391`

The prompt side guards short aliases carefully: `len(tok) >= 4 or tok.isupper()`, matched case-**sensitively** against the raw prompt. The content side (`entity_matches`) has neither guard. It accepts any alias of length 2+ and matches through `norm()`, which casefolds. So the alias `DO` (the PR's own test fixture, `test_knowledge_supply.py:53`) matches every line containing the word "do".

Repro on the **writing** class, drafting an email to Dana Okafor:

```
Task class: writing
== Dana Okafor ==
- [KNOWN 2026-09-04 canonical] Never do discounts below the 250 dollar an hour floor.  (q-fix/canonical/pricing-framework.md:2)
- [KNOWN 2026-09-04 canonical] Globex asked what we do about weekend support.  (q-fix/canonical/pricing-framework.md:3)
- [KNOWN 2026-09-04 relationship] ### Dana Okafor - CTO - Acme
    - Type: Customer  (q-fix/my-project/relationships.md:3)
- [KNOWN 2026-08-15 graph/owns] Dana Okafor owns the deployment runbook  (q-fix/memory/graph.jsonl:1)
Lines under her heading that never mention her: 2
```

An internal rate-floor rule and a different client's support question are labeled `KNOWN` under `== Dana Okafor ==` while the model drafts an outbound email to her. The same repro on the commitment class pulled a promise made to Globex under Dana's heading. The alias producer is `alias_of` triples written by `/q-debrief`, and two-letter initials are exactly what that emits. The author's own fixture is the shape.

---

## MAJOR 3 — a corrupt commitments ledger reads identical to an empty one, under COVERAGE: FULL

`q-system/.q-system/scripts/knowledge_supply.py:511`

Bad JSONL lines in `commitments.jsonl` are counted into `problems["commitments_bad_lines"]`. Nothing reads that key. `present_of` checks `cls not in problems`, and `"commitments"` is never the key that gets set, so the source row reports `present=True, problem=None`. Compare `graph`, which does surface `graph_bad_lines` in the receipt.

```
Empty file         -> coverage=FULL   present=True hits=0 problem=None
Every line corrupt -> coverage=FULL   present=True hits=0 problem=None
```

The module docstring at line 19 states the reason this module has a receipt at all: "'I did not find it' and 'I never searched the source it lived in' are different sentences." On this branch they are the same sentence, and the header prints FULL. Unattended, that tells the model there are no open promises for a client whose promise ledger is unreadable.

Honest note: I constructed the corrupt file, and no producer in this repo emits one. The author's own bad-line counter is the evidence the input was anticipated. It is wired to nothing.

---

## MINOR 4 — first-name expansion fires on line-initial and bullet-initial tokens

`q-system/.q-system/scripts/knowledge_supply.py:365`

The comment at 356 says "Sentence-initial stays out ('Mark the file as done' is a verb)". `(?<![.!?]\s)(?<!^)` only excludes a token after explicit sentence punctuation, and `(?<!^)` is not `MULTILINE`.

```
'Mark the file as done'                  fires=False
'Fix the tests.\nMark the file as done'  fires=False
'Fix the tests\nMark the file as done'   fires=True
'- Mark the file as done'                fires=True
'1. Mark the file as done'               fires=False
```

Bulleted, multi-line prompts are this fleet's house style, so the excluded case is the common one. Noise only, never blocks. The module docstring at line 37 also still reads "a bare first name never fires", which commit 184fcfbc made false.

---

## MINOR 5 — `entity_kinds_that_fire_alone` is a dead knob for slug, client, and capability

`q-system/.q-system/scripts/knowledge_supply.py:330`

`if ent.kind in fire_alone or ent.kind in ("slug", "client", "capability")`. The `or` overrides the manifest. An instance using the documented `<qroot>/.q-system/data/knowledge-sources.json` override to quiet a noisy slug gets no change.

```
Index            : [('acme labs', 'slug')]
Manifest declares: ['slug', 'client', 'loop', 'capability']
Passing ['capability', 'client', 'loop', 'slug'] -> resolved=['acme-labs']
Passing []                               -> resolved=['acme-labs']
Passing ['nothing']                      -> resolved=['acme-labs']
```

`loop` is also listed in the manifest but `_add` never creates a `loop`-kind entity, so that entry is inert for a second reason.

---

## What is sound

- **Never blocks.** Corrupt graph, corrupt granola JSON, corrupt loops JSON, a 300 KB prompt, and NUL/bidi/emoji in the prompt: `rc=0`, bounded stdout, empty stderr in every case. An empty prompt exits silently.
- **Instance isolation.** `find_qroot` prefers a non-`q-system` q-dir, and no foreign-instance content crossed in my two-root probe.
- **Ledger writes.** Both `.knowledge-supply-receipts.jsonl` and `.knowledge-supply-misses.jsonl` are caught by `.gitignore:43`, so the hook cannot make an auto-commit noisy. `append_jsonl` is a single `O_APPEND` write.
- **Wiring claims.** `hook_envelope_audit.py --hook` returns 0. `settings-template-sync-check.py --check` returns 0. The hook is wired in both settings files at `timeout: 5`, and the template form carries the `test -f ... || true` guard, so a fleet instance missing the script is a no-op rather than an error.
- **Latency.** A non-firing prompt against 8 MB of stores took 398 ms, linear in store size, far inside the 5s timeout, and an overrun is a discarded injection rather than a blocked prompt. Not a finding today. It grows with the graph.
- **Supersession.** I tried to make an accumulative predicate supersede and could not. The `state_predicates` allowlist holds, and same-date ties fall to `CONFLICTING` rather than picking a wrong winner. The inverted `-_line` tiebreak at line 566 is masked by that conflict branch, and I could not turn it into a wrong label, so I dropped it.
- **Ceiling bypass by pinned items.** I tried to blow the 8000-char ceiling through the pin exemption. The per-class caps bound the payload well below it for every object length a debrief plausibly writes. Could not make it fail, so dropped.

---

## VERDICT: REQUEST CHANGES

Not BLOCK. Nothing here is permanent or unrecoverable. The hook is additive, exits 0, writes only gitignored ledgers, and has a `KNOWLEDGE_INJECT_OFF=1` kill switch. Everything wrong is wrong in-context and evaporates on the next prompt.

**Fix first: MAJOR 1.** Move the per-class `cap` from "N per class total" to "N per class per entity", or run the pin pass before the cap. It is the only finding that silently reports the absence of facts the store actually holds, on the exact query shape this PR was built for. MAJOR 2 is a two-line fix in `entity_matches`: raise the alias floor to 4 characters unless the alias is uppercase, and match uppercase aliases case-sensitively, mirroring the rule `resolve_entities:325` already applies. MAJOR 3 is one key rename plus one receipt field.

FINDINGS:
major|Per-class cap runs across all entities before pinning, so every entity after the first gets zero facts while the header lists it as resolved|q-system/.q-system/scripts/knowledge_supply.py:986
major|entity_matches accepts 2-char aliases case-insensitively, so alias "DO" attributes every store line containing the word "do" to that person, including in outbound writing-class drafts|q-system/.q-system/scripts/knowledge_supply.py:391
major|Corrupt commitments.jsonl lines are counted into a key nothing reads, so an unreadable promise ledger reports present=True, problem=None under COVERAGE: FULL|q-system/.q-system/scripts/knowledge_supply.py:511
minor|First-name expansion fires on line-initial and bullet-initial tokens, contradicting the "sentence-initial stays out" comment and the docstring claim that a bare first name never fires|q-system/.q-system/scripts/knowledge_supply.py:365
minor|The manifest knob entity_kinds_that_fire_alone is inert for slug/client/capability because those kinds are hardcoded, and "loop" is a kind no code produces|q-system/.q-system/scripts/knowledge_supply.py:330
END FINDINGS

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-302-20260904-211653.md` (9802 bytes).

```
FINDINGS:
major|First-name expansion is gated on `if not found:`, silently dropping a named person whenever any other entity resolves, under a COVERAGE: FULL header and with no misses row|q-system/.q-system/scripts/knowledge_supply.py:365
minor|Pinned graph items bypass the char ceiling, so the payload can exceed ceiling_chars while the receipt reports cut=0 and ceiling_hit=False|q-system/.q-system/scripts/knowledge_supply.py:861
minor|status_for_line re-reads and re-parses provenance-vocabulary.json once per matched line, 100% of its cost, on a hook wired at timeout 5|q-system/.q-system/scripts/knowledge_supply.py:244
END FINDINGS
```

--- reviewer output ---

Review complete. Three findings survived reproduction.

---

## Finding 1

**SEVERITY: major**

**Claim:** First-name expansion is gated on `if not found:`, so a bare first name is silently dropped whenever any other entity resolved. The payload still announces `COVERAGE: FULL` and the misses ledger records nothing.

**File:** `q-system/.q-system/scripts/knowledge_supply.py:365`

The module docstring (line 37) and the PR body both say a bare first name fires mid-sentence when it expands to exactly one entity. Neither states the real condition: it fires only when nothing else resolved.

**Reproducer:**

```
python3 -c 'import tempfile, importlib.util
from pathlib import Path
spec = importlib.util.spec_from_file_location("t","q-system/.q-system/scripts/test_knowledge_supply.py")
t = importlib.util.module_from_spec(spec); spec.loader.exec_module(t)
ks = t.ks
tmp = Path(tempfile.mkdtemp()); root, q = t.make_instance(tmp)
p = "draft the follow-up to Mark Chen and tell Dana the memo is out"
b = ks.supply(root, p, session_id="s", now=t.NOW, record=True)
print(ks.render(b).splitlines()[0])
print("entities in payload:", [e["name"] for e in b["entities"]])
print("Dana facts injected:", [i["text"][:60] for i in b["items"] if "Dana" in i["entity"]])
mp = q/"memory"/".knowledge-supply-misses.jsonl"
print("misses ledger:", mp.read_text().strip() if mp.exists() else "(file not written)")'
```

**Real output:**

```
[knowledge-supply] COVERAGE: FULL. task=commitment entities=Mark Chen (ambiguous: Acme | Globex).
entities in payload: ['Mark Chen']
Dana facts injected: []
misses ledger: (file not written)
```

`miss_candidates` only matches capitalized bigrams, so a dropped unigram name leaves no row anywhere.

**Mutation, to show nothing pins this:**

```
python3 -c 'import shutil, tempfile, subprocess, sys
from pathlib import Path
src = Path("q-system/.q-system/scripts"); dst = Path(tempfile.mkdtemp())/"scripts"
shutil.copytree(src, dst)
shutil.copy("q-system/.q-system/knowledge-sources.json", dst.parent/"knowledge-sources.json")
p = dst/"knowledge_supply.py"; s = p.read_text()
p.write_text(s.replace("    if not found:\n", "    if True:\n", 1))
r = subprocess.run([sys.executable,"-m","pytest",str(dst/"test_knowledge_supply.py"),"-q"],capture_output=True,text=True)
print(r.stdout.strip().splitlines()[-1])'
```

**Real output:** `34 passed in 0.39s`

The gate is invisible to the suite. Removing it changes behavior and no test notices.

**3am scenario:** an unattended writing-class run drafts an email to two named people, gets facts for one, and the first line of its own context claims FULL coverage. Nothing on disk says a name went unresolved.

---

## Finding 2

**SEVERITY: minor**

**Claim:** Pinned graph items bypass the char ceiling entirely, so the rendered payload can exceed `ceiling_chars` while the receipt reports `cut=0` and `ceiling_hit=False`.

**File:** `q-system/.q-system/scripts/knowledge_supply.py:861`

`if idx in pinned or (not ceiling_hit and used + cost <= ceiling)` keeps every pin regardless of budget, so the real ceiling scales with entity count.

**Reproducer:**

```
python3 -c 'import tempfile, json, importlib.util
from pathlib import Path
spec = importlib.util.spec_from_file_location("t","q-system/.q-system/scripts/test_knowledge_supply.py")
t = importlib.util.module_from_spec(spec); spec.loader.exec_module(t)
ks = t.ks
for nents, olen in ((12,600),(30,120),(40,80)):
    tmp = Path(tempfile.mkdtemp()); root, q = t.make_instance(tmp)
    names = ["Ent%02d Person" % i for i in range(nents)]
    with open(q/"memory"/"graph.jsonl","a") as f:
        for n in names:
            f.write(json.dumps({"s": n, "p": "status", "o": "blocked on the " + ("z"*olen), "t": "2026-09-03", "project": "x"})+"\n")
    b = ks.supply(root, "what do we know about " + ", ".join(names), session_id="s", now=t.NOW, record=False)
    txt = ks.render(b)
    print("entities=%2d obj_len=%3d rendered=%5d ceiling=%d over=%5d cut=%d ceiling_hit=%s" % (
        nents, olen, len(txt), b["budget"]["ceiling"], len(txt)-b["budget"]["ceiling"], b["budget"]["cut"], b["receipt"]["ceiling_hit"]))'
```

**Real output:**

```
entities=12 obj_len=600 rendered= 9305 ceiling=8000 over= 1305 cut=0 ceiling_hit=False
entities=30 obj_len=120 rendered= 8117 ceiling=8000 over=  117 cut=0 ceiling_hit=False
entities=40 obj_len= 80 rendered= 9057 ceiling=8000 over= 1057 cut=0 ceiling_hit=False
```

**Calibration, stated honestly:** this needs 12 or more entities in one prompt, and I found no length bound on the graph `o` field in any producer (`kb-graph-guard.py`, `graph-kb.yaml`), but I could not point at a real prompt that hits it. Per the standing rule I dropped it a level. `test_ceiling_keeps_newest_triple_and_records_cut` uses one entity and structurally cannot see this.

---

## Finding 3

**SEVERITY: minor**

**Claim:** `status_for_line` opens and JSON-parses `provenance-vocabulary.json` once per matched line. That is 100% of its cost, on the hot path of a hook wired at `timeout: 5`.

**File:** `q-system/.q-system/scripts/knowledge_supply.py:244`

The docstring on `load_vocab_floor` (line 225) justifies the re-read as anti-drift. Nothing can drift between two lines of one file inside one call.

**Reproducer:**

```
python3 -c 'import time, importlib.util
from pathlib import Path
spec = importlib.util.spec_from_file_location("ks","q-system/.q-system/scripts/knowledge_supply.py")
ks = importlib.util.module_from_spec(spec); spec.loader.exec_module(ks)
N = 20000
t0=time.time()
for _ in range(N): ks.status_for_line("Dana Okafor line about the plan")
el=time.time()-t0
print("status_for_line x%d = %.2fs (%.3f ms/line)" % (N, el, el*1000/N))
t0=time.time()
for _ in range(N): ks.load_vocab_floor()
print("load_vocab_floor x%d = %.2fs" % (N, time.time()-t0))'
```

**Real output:**

```
status_for_line x20000 = 0.76s (0.038 ms/line)
load_vocab_floor x20000 = 0.77s
```

On a synthetic instance with 62 canonical files and 24k matching lines, end-to-end cost is 0.35s per resolved entity and scales linearly (1 entity 0.36s, 4 entities 1.40s). A killed hook produces no output and no receipt, so the failure is silent.

---

## Attacks I tried that the code survived

**Mutation testing on copies.** Six mutations, six killed. Coverage-always-FULL: 2 failed. Drop the voided/misattributed filter: 1 failed. Ignore UNVALIDATED markers: 1 failed. Remove the per-entity cap: 1 failed. Sentence-initial guard off: 3 failed. Misses ledger neutered: 1 failed. The suite is genuinely load-bearing on every slice the PR claims. Only the first-name gate survived, which is Finding 1.

**Cross-client contamination.** I went after the shape that already bit this code (the `DO` alias). `alias_in_text` is correctly asymmetric now, `resolve_meetings` matches by client key or explicit mention, `resolve_commitments` by slug or promise text, `build_index` pops aliases from the index and a re-added alias stays kind `graph` so it never fires alone. Found no second path.

**Ledger concurrency.** `append_jsonl` uses `O_APPEND` with a single `os.write`. Two overlapping sessions cannot truncate each other.

**Latency against the 5s hook timeout.** Built a 10k-row graph, 62 canonical files of 400 lines, and 4 entities. Worst measured: 1.40s. Could not push it past 5s without absurd fixtures.

**Gitignore leak.** Confirmed `git check-ignore` covers both new ledgers via `.gitignore:43 *.jsonl`. No prompt fragments reach the public repo.

**Capability index noise.** 225 keys, only 9 single-word (`brand`, `conftest`, `council`, `design`, `improve`, `rca`, `say`, `security`, `sycophancy`). Not a cry-wolf surface.

**Two things I tried and dropped for lack of a real producer:** a non-dict stdin payload raises an uncaught `AttributeError` at `knowledge-inject.py:41` (outside the try) and exits 1, but Claude Code never emits that shape. And the `.claude/settings.json` line has no `test -f` guard or `|| true`, so a missing script exits 2 and would block the prompt, but that is the identical shape of the neighbouring `lessons-inject.py` line and I could not reproduce the file going missing.

**Three nits, not reported as findings:** `conftest` lands in the capability index because the script glob only filters `startswith("test")`; `present_of` at line 993 carries a dead `cls != "relationship"` branch (no manifest class uses the singular); the `conflicts` counter at line 627 increments on every ordinary supersession, not just same-date conflicts, though nothing renders it.

---

## VERDICT: REQUEST CHANGES

Not BLOCK. Nothing here is permanent or unrecoverable, the hook is additive and exits 0, and `KNOWLEDGE_INJECT_OFF=1` is a working kill switch.

**Fix first:** Finding 1. Move the first-name expansion out from under `if not found:` so it runs alongside other resolutions, and add a `resolved_from == "first_name"` case to the multi-entity test. If the gate is deliberate (a deduplication guard I did not see the reason for), then the docstring at line 37 and the PR body both need to say so, and an unresolved capitalized unigram needs a row in the misses ledger so the drop stops being invisible.

FINDINGS:
major|First-name expansion is gated on `if not found:`, silently dropping a named person whenever any other entity resolves, under a COVERAGE: FULL header and with no misses row|q-system/.q-system/scripts/knowledge_supply.py:365
minor|Pinned graph items bypass the char ceiling, so the payload can exceed ceiling_chars while the receipt reports cut=0 and ceiling_hit=False|q-system/.q-system/scripts/knowledge_supply.py:861
minor|status_for_line re-reads and re-parses provenance-vocabulary.json once per matched line, 100% of its cost, on a hook wired at timeout 5|q-system/.q-system/scripts/knowledge_supply.py:244
END FINDINGS

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-302-20260904-212756.md` (7291 bytes).

```
FINDINGS:
major|miss_candidates writes one row per bigram occurrence, not per distinct candidate: 1600 rows / 245KB from one pasted transcript, no dedupe, no cap, no prune, and the only test collapses to a set|q-system/.q-system/scripts/knowledge_supply.py:954
minor|assemble omits the per-entity separator lines from its byte accounting, so render() exceeds ceiling_chars while the receipt reports overflow 0|q-system/.q-system/scripts/knowledge_supply.py:1098
minor|A bundle with zero items still injects ~800 chars of header, contradicting the documented "zero bytes when nothing matches"|q-system/.q-system/scripts/knowledge-inject.py:52
minor|Resolution is O(prompt_chars x index_size); a 109KB paste against a 6000-entity index took 7.1s, past the configured timeout 5|q-system/.q-system/scripts/knowledge_supply.py:325
END FINDINGS
```

--- reviewer output ---

Review done. Read the diff, read both scripts end to end, ran the suite, and ran four constructed repros against synthetic instances in temp dirs. No repo files touched.

---

## Finding 1

**SEVERITY: major**

**Claim:** `miss_candidates` emits one ledger row per capitalized-bigram *occurrence*, not per distinct candidate, so one pasted transcript writes hundreds of duplicate rows into an unbounded, never-pruned file.

**File:** `q-system/.q-system/scripts/knowledge_supply.py:954` (the `finditer` loop), written at `:1000` and `:1115`.

**Reproducer** (run from the worktree, writes only to a `tempfile.mkdtemp`):

```
python3 -  # loads knowledge_supply.py, builds an empty-graph instance,
           # calls supply() once with a 400-line pasted transcript, record=True
```

**Real output:**

```
prompt chars      : 36449
misses rows written: 1600
distinct candidates: 4 {'Alex Rivera': 400, 'Quarterly Review': 400, 'Vendor Contract': 400, 'Tuesday Morning': 400}
ledger bytes       : 245200
```

**Why it matters unattended.** `auto-detection.md` makes pasted transcripts the house prompt shape, so this is the common case, not the corner. 245 KB per paste, per instance, fleet-wide. Nothing rotates, caps, or prunes it:

```
grep -rn "knowledge-supply-misses|MISSES_NAME" . --exclude-dir=.git | grep -v test_knowledge_supply
q-system/.q-system/scripts/knowledge_supply.py:67
q-system/.q-system/scripts/knowledge_supply.py:1001
q-system/.q-system/scripts/knowledge_supply.py:1116
```

No reader exists. The module docstring calls this "the data the Phase 2 decisions run on", and those counts are now dominated 400 to 1 by how many times a name appeared in one paste.

**Why the suite does not see it:** `test_receipt_and_misses_are_written:592` collapses to a set (`cands = {m["candidate"] for m in misses}`), so duplicate rows are structurally invisible to the only test that touches this ledger.

**Human action:** delete the file, add a per-prompt `dict` dedupe in `miss_candidates`, and give the ledger a size cap.

---

## Finding 2

**SEVERITY: minor**

**Claim:** `assemble` never counts the per-entity `== name ==` separator lines or the trailing `delegated:` line, so the rendered payload exceeds `ceiling_chars` while the receipt's own honesty field reports `overflow: 0`.

**File:** `q-system/.q-system/scripts/knowledge_supply.py:1098` (`header_len = len(render_header(bundle)) + 200`), reported at `:893` and `:1102`.

**Reproducer:** 10 entities, 30 graph rows each, one prompt naming all ten.

**Real output:**

```
ceiling      : 8000
render chars : 8107
OVER by      : 107
overflow     : 0  <-- claims it fit
ceiling_hit  : True  cut: 254
entity sep lines uncounted: 10
```

The fixed `+200` slack is a constant standing in for a cost that scales with entity count. `test_ceiling_keeps_newest_triple_and_records_cut:419` asserts `len(text) <= ceiling` but only ever resolves one entity, so it has exactly one separator and the constant covers it.

---

## Finding 3

**SEVERITY: minor**

**Claim:** A bundle with zero items still injects the full header, contradicting "Zero bytes when nothing matches" in the hook docstring (`knowledge-inject.py:15`) and the PR body.

**File:** `q-system/.q-system/scripts/knowledge-inject.py:52` (`if not bundle` is truthy for an items-empty dict).

**Reproducer:** one commitment in a `DROP_STATE`, prompt naming its slug plus "last week".

**Real output:**

```
items: 0 coverage: PARTIAL
--- rendered payload the model receives ---
[knowledge-supply] COVERAGE: PARTIAL. missing: meetings (q-system/output/granola-cache.json absent). task=temporal_event entities=brightspeed-audit. window=2026-08-28..2026-09-04
[knowledge-supply] Hierarchy: graph beats canonical beats notes ... (2 more lines)
[knowledge-supply] delegated: lessons -> lessons-inject, voice -> voice-dna-loader. cut=0 receipt=...
```

Roughly 800 chars of boilerplate carrying no facts, on any prompt whose window or state filter empties the result. Reachable on "what happened with X yesterday" whenever nothing happened yesterday.

---

## Finding 4

**SEVERITY: minor** (constructed input, severity dropped a level and said so)

**Claim:** Resolution cost is O(prompt_chars × index_size) and passes the configured `timeout: 5` on a large paste.

**File:** `q-system/.q-system/scripts/knowledge_supply.py:325` (per-entity regex over the normalized prompt).

**Real output**, 6,000-entity index:

```
index size: 6000
paste   9124 chars ->  698 ms
paste  36424 chars -> 2424 ms
paste 109224 chars -> 7140 ms
```

Past 5s the hook is killed and its output discarded, which is silent and harmless for an additive hook. The cost that lands on the founder is 2.4s added to every turn at a 36 KB paste. I built the 6,000-entity index myself; no store in this repo emits one, so treat the number as an upper bound, not a measurement of production.

---

## What is sound

Attacks that failed, and I tried them properly:

- **Working-tree pollution.** Both ledgers are covered by `.gitignore:43` (`*.jsonl`), verified with `git check-ignore -v`. Auto-commit cannot sweep them.
- **Concurrency.** `append_jsonl` is O_APPEND with a single `os.write`; a real receipt row measured 958 bytes, far below any interleave threshold. Two sessions cannot tear a row.
- **Hook never blocks.** Exit 0 on broken stdin, missing engine, unparseable store, and the kill switch. `test_hook_silent_on_broken_store` is real.
- **Supersession scoping.** I could not make an accumulative predicate go STALE. The `state_predicates` allowlist is genuinely data-driven and the project grouping holds.
- **The `entity_kinds_that_fire_alone` knob is live.** No shadow hardcoded list survives; the test kills that mutant.
- **Short-alias matching.** The "DO matches every line containing do" class is fixed and pinned case-sensitively.
- **Manifest absent is observable**, not silent: it writes an `error: manifest_missing` receipt.
- **Suite:** 37 passed in 0.38s. Several tests are mutation-grade, not decoration.

---

**VERDICT: REQUEST CHANGES**

Not BLOCK. Nothing here is permanent: the ledger is gitignored and deletable, the hook is additive and exits 0 on every path, and rollback is removing two settings lines.

**Fix first:** dedupe `miss_candidates` per prompt and cap the misses ledger. Right now the house's most common prompt shape writes a quarter megabyte of 400x-duplicated rows into the file that Phase 2 is meant to decide from.

FINDINGS:
major|miss_candidates writes one row per bigram occurrence, not per distinct candidate: 1600 rows / 245KB from one pasted transcript, no dedupe, no cap, no prune, and the only test collapses to a set|q-system/.q-system/scripts/knowledge_supply.py:954
minor|assemble omits the per-entity separator lines from its byte accounting, so render() exceeds ceiling_chars while the receipt reports overflow 0|q-system/.q-system/scripts/knowledge_supply.py:1098
minor|A bundle with zero items still injects ~800 chars of header, contradicting the documented "zero bytes when nothing matches"|q-system/.q-system/scripts/knowledge-inject.py:52
minor|Resolution is O(prompt_chars x index_size); a 109KB paste against a 6000-entity index took 7.1s, past the configured timeout 5|q-system/.q-system/scripts/knowledge_supply.py:325
END FINDINGS

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-302-20260904-214227.md` (7767 bytes).

```
FINDINGS:
major|Resolution is O(entities x store rows), so at ~65 resolved names the hook is killed at its wired 5s timeout, injects nothing, and writes no receipt|q-system/.q-system/scripts/knowledge_supply.py:1109
minor|CONFLICTING is assigned when the two newest rows in a group agree, and the genuinely superseded row loses its STALE label|q-system/.q-system/scripts/knowledge_supply.py:666
minor|CAP_BIGRAM_RE matches non-overlapping, so the misses ledger records "The Acme" and "Ping Sarah" instead of "Acme Contract" and "Sarah Chen"|q-system/.q-system/scripts/knowledge_supply.py:1002
minor|No per-item size cap on pinned graph items, so injected context exceeds ceiling_chars without bound|q-system/.q-system/scripts/knowledge_supply.py:898
END FINDINGS
```

--- reviewer output ---

## Review: PR #302, knowledge-supply

Read-only. Nothing in the repo was modified. Test suite runs green on the branch: `41 passed in 0.46s`.

---

### MAJOR 1. The hook exceeds its wired 5s timeout on a prompt inside its own truncation gate, injects nothing, and leaves no receipt

`q-system/.q-system/scripts/knowledge_supply.py:1109` (the per-class, per-entity resolver loop) and `:1221` (the receipt is appended as the last statement of `supply()`).

Resolution cost is `O(resolved_entities x store_rows)`. `entity_matches` re-normalizes every store row for every entity, with no memoization. Profiled at 40 entities: 1,021,520 calls to `entity_matches`, 2,066,314 to `norm`.

Reproducer, against a synthetic instance sized to the author's own stated scale (6,000 index entities, 599 KB graph, 2.1 MB granola cache, 2,000 commitments):

```
k= 40 elapsed 3.00s  injected_chars= 7914 ceiling=8000 overflow=0
k= 55 elapsed 4.08s  injected_chars= 7916 ceiling=8000 overflow=0
k= 60 elapsed 4.42s  injected_chars= 8489 ceiling=8000 overflow=489
k= 70 elapsed 5.17s  injected_chars= 9820 ceiling=8000 overflow=1820
k= 80 elapsed 5.92s  injected_chars=11160 ceiling=8000 overflow=3160
```

Then through the real hook, with the real wired timeout of 5, on a prompt of exactly 12,000 chars, which is `PROMPT_SCAN_CHARS`, so the paste guard does not protect it:

```
receipt exists before: False
KILLED at the wired 5s timeout after 5.0s; stdout bytes=0
receipt exists after: False
misses exists after: False
```

Two consequences. First, every prompt of that shape costs the founder a 5 second stall before the turn starts, and 40 resolved names already costs 3 seconds. The PR's measured p95 of 21 ms comes from a replay whose prompts resolved few entities, so it cannot see this. Second, and worse for the design: the module's stated reason to exist is that `"I did not find it" and "I never searched the source it lived in" are different sentences`, recorded by the receipt. On timeout neither sentence is recorded. The one failure the receipt was built to make visible is the one it cannot see, because the write is last.

Honest note on the input: I constructed a name-dense 12 KB prompt. The shape is one this repo explicitly invites, `auto-detection.md` treats a pasted transcript as a first-class input, but I did not find a real 65-name prompt in the replay data.

---

### MINOR 2. `CONFLICTING` fires when the two newest rows agree

`knowledge_supply.py:666-669`. The predicate is `len(objs) > 1` over the whole group, gated on the two newest sharing a date. It never checks whether those two newest actually differ.

Reproducer: an append-only graph where a debrief re-recorded the same current fact twice on one date, plus one genuinely superseded older row.

```
CONFLICTING | 2026-09-01 | Dana Okafor status in pilot   | q-test/memory/graph.jsonl:3
CONFLICTING | 2026-06-01 | Dana Okafor status closed won | q-test/memory/graph.jsonl:1
conflicts counted: 1
```

Correct labels are KNOWN on the first and STALE on the second. The model is told the instance contradicts itself when it does not, and the superseded row loses its STALE marking.

---

### MINOR 3. The misses ledger records the wrong candidates

`knowledge_supply.py:114` and `:1002`. `CAP_BIGRAM_RE.finditer` matches non-overlapping, so a capitalized word in front of a name consumes it. `"The"` and ordinary capitalized verbs are not in `STOPWORDS`.

Reproducer, one ordinary prompt:

```
{"candidate": "The Acme",       "shape": "capitalized_bigram"}
{"candidate": "Ping Sarah",     "shape": "capitalized_bigram"}
{"candidate": "Stripe Webhook", "shape": "capitalized_bigram"}
{"candidate": "AWS Bill",       "shape": "capitalized_bigram"}
```

`Acme Contract` and `Sarah Chen` are the entities the prompt actually named, and neither is recorded. The docstring at line 40 calls this ledger "the data the Phase 2 decisions run on."

---

### MINOR 4. `ceiling_chars` has no per-item cap, so one pinned item is unbounded

`knowledge_supply.py:898-905` reserves pins, `:1199` refuses to drop them, and no code path truncates `item["text"]`.

Realistic case, six people whose graph rows each carry a 1,263-char verbatim objection, which is an ordinary shape for a model-written triple:

```
entities 6 items 6
ceiling 8000 injected_chars 8978 overflow 978
```

Constructed extreme, one entity, one long object:

```
ceiling 8000 injected_chars 60632 overflow 52632 items 1
```

The overrun is reported in `overflow`, which the PR treats as the accepted contract, and `test_pins_alone_over_ceiling_is_reported_not_hidden` asserts that. That test's fixture pins 250-char objects, so it is structurally blind to the unbounded case. Reporting is not bounding, and `graph.jsonl` is model-written with no length bound anywhere in this repo. I am rating this minor rather than major because the producer-plausible magnitude is a 12 percent overrun; the 7.6x case uses an input I constructed.

---

## What is sound

Attacks I ran that the code survived, named so the findings above can be trusted:

- **Ledger content reaching the public repo.** `git check-ignore -v` confirms `.gitignore:43:*.jsonl` covers both `.knowledge-supply-receipts.jsonl` and `.knowledge-supply-misses.jsonl`. Prompt-derived candidate strings stay local. No leak.
- **Concurrent sessions racing `append_bounded`.** Six forked processes, 500 appends each, an 8 KB cap forcing constant truncation contention on one shared `.tmp` path. Result: `lines=19 corrupt_json_lines=0`, no leftover `.tmp`. The `O_APPEND` plus `os.replace` pairing holds.
- **Stdout pollution through `writing_patterns()`.** It `exec_module`s `voice-dna-loader.py` on the same stdout the hook writes JSON to. That file is `__main__`-guarded and has no module-level output, so the envelope stays clean.
- **The manifest-missing path writing an unbounded receipt per prompt.** Reachable only if the manifest fails to sync while the script does; `kipi-update.sh` rsyncs the whole `q-system` tree minus `INSTANCE_OWNED_SUBTREES`, so both travel together. Dropped.
- **`capability_index` cost on the real repo.** 225 entries in 47 ms, and it only builds when capability phrasing matches and no entity resolved.
- **Ordinary no-entity prompt against 3 MB of stores.** 35 to 39 ms, which is the common case and it is fine.
- **A 109 KB paste.** 39 ms. The `PROMPT_SCAN_CHARS` truncation does its job for that shape; it is the 12 KB dense shape that gets through.

---

**VERDICT: REQUEST CHANGES.** One major finding. Not BLOCK: the hook fails open, nothing it writes is permanent, and rolling back is deleting two settings lines.

**Fix first:** bound the resolution work so the hook cannot outrun its own timeout. Memoize `norm(row_text)` per store row instead of recomputing it per entity, or cap resolved entities per prompt. Then move the receipt append to before the expensive resolution, or write a second row on the way in, so a killed run still leaves the trace the receipt exists to give.

FINDINGS:
major|Resolution is O(entities x store rows), so at ~65 resolved names the hook is killed at its wired 5s timeout, injects nothing, and writes no receipt|q-system/.q-system/scripts/knowledge_supply.py:1109
minor|CONFLICTING is assigned when the two newest rows in a group agree, and the genuinely superseded row loses its STALE label|q-system/.q-system/scripts/knowledge_supply.py:666
minor|CAP_BIGRAM_RE matches non-overlapping, so the misses ledger records "The Acme" and "Ping Sarah" instead of "Acme Contract" and "Sarah Chen"|q-system/.q-system/scripts/knowledge_supply.py:1002
minor|No per-item size cap on pinned graph items, so injected context exceeds ceiling_chars without bound|q-system/.q-system/scripts/knowledge_supply.py:898
END FINDINGS

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-302-20260904-220235.md` (7554 bytes).

```
FINDINGS:
major|Single-token contacts fire sentence-, line- and bullet-initial, bypassing the is_initial_position guard the docstring promises, and no test covers the path|q-system/.q-system/scripts/knowledge_supply.py:375
minor|An unreadable canonical/relationships/decisions file is silently skipped and reported as present with 0 hits under COVERAGE: FULL|q-system/.q-system/scripts/knowledge_supply.py:692
minor|A non-dict, non-list open-loops.json raises AttributeError past supply(), so the hook emits no context and no receipt|q-system/.q-system/scripts/knowledge_supply.py:621
minor|An entity whose facts duplicate another entity's is named in the header with no section, reading as nothing known|q-system/.q-system/scripts/knowledge_supply.py:880
END FINDINGS
```

--- reviewer output ---

## Review of PR #302

Read the diff, ran the suite (43 pass), mutation-checked five load-bearing branches on a copy (all red), and attacked the resolver, the coverage receipt, and the store loader with real reproducers.

---

## Findings

### 1. major — a single-token contact fires sentence-, line- and bullet-initial, which is the exact noise the module says disqualifies it

`q-system/.q-system/scripts/knowledge_supply.py:375`

The module docstring says single-token names "never fire alone" and that a hook firing on "mark the file as done" is noise that gets guards switched off. `is_initial_position` guards the first-name-expansion path. The **contact** path at line 375 has no such guard: it checks `tok[0].isupper()` on the *entity name*, not on the prompt occurrence, so any capitalized whole-word match fires wherever it sits.

`relationships.md` headings are parsed by splitting on the first dash, so `### Ethan — Contractor — FBI` yields the single-token contact `Ethan`. That is this repo's own file, not an invented fixture.

Reproducer against the real `q-system/my-project/relationships.md`:

```
contact_names parsed from the repo's OWN relationships.md: ['Ally', 'Ethan', 'Tova citrin', 'Friend at Google (name TBD)']
'Ethan can wait, ship it' -> ['Ethan']
    [knowledge-supply] COVERAGE: PARTIAL. missing: graph ... task=entity_lookup entities=Ethan.
'Ally the two configs and rerun' -> ['Ally']
    [knowledge-supply] COVERAGE: PARTIAL. ... task=entity_lookup entities=Ally.
'Ethan' -> ['Ethan']
```

And in a fixture with a first-name-only heading, the docstring's own example fires:

```
PROMPT: 'Mark the file as done'   -> ['Mark']
PROMPT: 'Mark this PR as ready'   -> ['Mark']
PROMPT: '- Mark the todo complete' -> ['Mark']
PROMPT: 'mark the file as done'   -> NOTHING
```

No test can see this. The fixture's `relationships.md` has only multi-token headings, so adding the guard changes nothing:

```
suite with the contact guard ADDED: 43 passed in 0.44s
rc 0 -> no test observes the single-token-contact path
```

---

### 2. minor — an unreadable canonical file reads as "searched, found nothing" under COVERAGE: FULL

`q-system/.q-system/scripts/knowledge_supply.py:692` (and the same shape at `:715` for relationships and decisions)

`resolve_canonical` catches `OSError` and `continue`s. The receipt then reports `present=True, hits=0, problem=None` and the header says FULL. That is the "not found versus never searched" collapse the receipt exists to prevent, and it is the same class as round 1's commitments finding, which was fixed only for commitments.

```
READABLE coverage= FULL canonical present/hits/problem = (True, 1, None)
DANGLING coverage= FULL canonical present/hits/problem = (True, 0, None)
CHMOD000 coverage= FULL canonical present/hits/problem = (True, 0, None)
payload: [knowledge-supply] COVERAGE: FULL. task=entity_lookup entities=Dana Okafor. ...
```

Severity dropped one level on purpose: I had to construct the dangling symlink and the chmod. No producer in this repo emits an unreadable canonical file today.

---

### 3. minor — a non-dict, non-list `open-loops.json` raises past `supply()`, against the module's stated contract

`q-system/.q-system/scripts/knowledge_supply.py:621`

`data.get("loops")` runs on whatever parsed. The `except` catches only `OSError, ValueError`. The docstring says "Any store that fails to parse is reported in the receipt, never raised past supply()." Through the hook the exception is swallowed to exit 0, so the result is no context **and no receipt**, which is the one state the receipt design is meant to make impossible.

```
RAISE loops=scalar -> AttributeError: 'int' object has no attribute 'get'
RAISE loops=string -> AttributeError: 'str' object has no attribute 'get'
OK    granola=list -> bundle
OK    granola=scalar -> bundle
OK    commit row null promise -> bundle
OK    granola rows not dicts -> bundle
```

Constructed input, so minor. Worth noting the general shape though: the hook's only liveness signal is written on the success path, so a crash-every-prompt regression is invisible.

---

### 4. minor — a resolved entity whose facts duplicate another's is named in the header with no section under it

`q-system/.q-system/scripts/knowledge_supply.py:880`

Dedupe key is `(kind, norm(text))` across all entities, so the second entity loses shared rows and its `== Name ==` block never renders. The header still names it, which reads as "searched, nothing known".

```
[knowledge-supply] COVERAGE: FULL. task=entity_lookup entities=Dana Okafor, Okafor Co.
== Dana Okafor ==
- [KNOWN 2026-08-20 graph/status] Dana Okafor status churned  (q-system/memory/graph.jsonl:3)
- [KNOWN 2026-08-01 graph/works_at] Dana Okafor works_at Okafor Co  (q-system/memory/graph.jsonl:1)
[knowledge-supply] delegated: ... cut=0 receipt=...
```

No `== Okafor Co ==` block.

---

## What is sound

- **The round-4 structural deadline actually bounds the pass.** 60,000 graph rows, 6,000 distinct entities, 8 MB of canonical, 20 names in one prompt: `supply total 3.63s, deadline_hit={'at_class': 'canonical', 'at_entity': 'Person Number1', 'elapsed_ms': 3624}, entities=12`. Under the wired 5 s timeout, with a receipt naming where it stopped. `load_stores 0.07s, build_index 0.22s` on the same tree, so the pre-deadline segment is not the hole I went looking for.
- **The tests are not decoration.** Five mutations on a copy, all red: dropping the STALE branch (1 failed), never pinning the newest graph item (3 failed), ignoring the deadline (1 failed), dropping the per-entity cap (1 failed), letting first names fire sentence-initial (3 failed).
- **Corrupt store shapes mostly degrade cleanly.** Granola cache as a list, as a scalar, with non-dict rows, and a commitment row with null promise/slug/state all returned a bundle rather than raising.
- **Concurrency on the misses ledger held.** Three processes, 900 appends each against a pre-filled over-cap ledger: 1,584 lines, 0 unparseable, 246,934 bytes (under the 262,144 cap), no stray `.tmp` left behind. I expected the shared tmp path to corrupt it and could not make that happen. Dropping that finding.
- **Verbatim, isolation, ceiling, exit-0 and the kill switch all behave as claimed** in both the suite and my own probes.
- **Rollback is clean.** Two settings lines, two gitignored jsonl ledgers, nothing published, nothing permanent. No blocker exists in this change.

---

## VERDICT: REQUEST CHANGES

Fix first: the contact branch at `knowledge_supply.py:375` needs the same `is_initial_position` check the first-name path already has, plus one fixture heading with a single-token name so a test can see it. Everything else here is minor and would not gate on its own.

FINDINGS:
major|Single-token contacts fire sentence-, line- and bullet-initial, bypassing the is_initial_position guard the docstring promises, and no test covers the path|q-system/.q-system/scripts/knowledge_supply.py:375
minor|An unreadable canonical/relationships/decisions file is silently skipped and reported as present with 0 hits under COVERAGE: FULL|q-system/.q-system/scripts/knowledge_supply.py:692
minor|A non-dict, non-list open-loops.json raises AttributeError past supply(), so the hook emits no context and no receipt|q-system/.q-system/scripts/knowledge_supply.py:621
minor|An entity whose facts duplicate another entity's is named in the header with no section, reading as nothing known|q-system/.q-system/scripts/knowledge_supply.py:880
END FINDINGS

assafkip:
Review loop stopped at the declared cap (round 5 was the verdict check on the final head). Five rounds of the degraded Opus fallback, each with four new findings on a 1,300-line module; no natural end. Round 5's findings are captured, not fixed here: sp-8a (major, initial-position guard on the single-token contact path, same class as round 1 minor 4, structural chokepoint fix), plus three minors. Ledger ids in the next comment once assigned. Merge decision routed to Sana's triage: the required check kipi/reviewer-approved cannot go green in degraded mode without a human deciding the remaining items are follow-up work.

assafkip:
Round 5 ledger ids: sp-2829d81f (major: one chokepoint for every single-token resolution, with the initial-position guard inside it; needs DoR), sp-ca1769db (unreadable markdown class reads as present with 0 hits), sp-a4a5028a (non-object open-loops.json raises past supply), sp-1b3ef442 (an entity whose items all dedupe away has no section). Triage: ASK-1260.

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-302-20260904-223540.md` (5676 bytes).

```
FINDINGS:
minor|Supersession and conflict grouping keys on the raw graph subject, so a status row written under an alias is never marked STALE or CONFLICTING against the canonical-name row|q-system/.q-system/scripts/knowledge_supply.py:709
minor|Canonical and relationship excerpts carry the file mtime in the date slot, so a clone or any unrelated edit re-dates every canonical fact to today while the header claims newest first|q-system/.q-system/scripts/knowledge_supply.py:754
END FINDINGS
```

--- reviewer output ---

Read-only review complete. Ran the suite, probed the hook end to end against the real repo, and built four adversarial fixtures in temp dirs. Repo tree is clean (I removed the two gitignored ledger files my hook probe created).

## Findings

### minor 1: supersession groups on the raw graph subject, so an alias-written status row is never STALE

`q-system/.q-system/scripts/knowledge_supply.py:709` builds the supersession group key as `(norm(r["s"]), norm(p), project)`. The subject is the string as written. `resolve_graph` retrieves rows via `entity_matches`, which matches name **or** alias, so rows written under the short form and rows written under the full name land in the same result list but in different groups. Neither STALE nor CONFLICTING fires.

Reproducer and real output:

```
rows: DO alias_of "Dana Okafor"
      DO         status churned  t=2026-06-01
      Dana Okafor status renewed t=2026-09-01
prompt: "what is the status of Dana Okafor"

- [KNOWN 2026-09-01 graph/status] Dana Okafor status renewed  (graph.jsonl:3)
- [KNOWN 2026-06-01 graph/status] DO status churned           (graph.jsonl:2)
STATUSES: [('Dana Okafor status renewed','KNOWN',None), ('DO status churned','KNOWN',None)]
```

Control, identical rows with the subject spelled the same both times:

```
- KNOWN | Dana Okafor status renewed | supersedes: None
- STALE | Dana Okafor status churned | supersedes: q-fix/memory/graph.jsonl:3
```

Same data, one string differs, and the label the module exists to produce disappears.

Why the suite is green on it: `GRAPH_ROWS` at `test_knowledge_supply.py:53` has an `alias_of` row but no state row whose subject is the alias, so every supersession test runs on the canonical spelling only. The fixture shares the code's mental model.

Caveat I owe you: I could not point at a live store emitting a status row under an alias subject. `q-system/methodology/debrief-template.md:279` (this PR's own diff) instructs writers to record the alias row so short forms resolve, which is what makes short-form subjects plausible, but the exact triggering row is my construction. That is why this is minor and not major.

Fix shape: key the group on the resolved entity, not `r["s"]`.

### minor 2: canonical and relationship excerpts carry the file mtime as their date

`knowledge_supply.py:754` sets `t = mtime_date(path)` for every canonical-family excerpt, and `render_item` at `:1002` prints it in the same date slot as real content dates from commitments, meetings and graph rows. The header at `:1033` says "Verbatim excerpts, newest first".

Reproducer, one file, same bytes, only mtime moved:

```
canonical item date = 2001-09-08 | Dana Okafor signed the 2024 pilot agreement
after touch (same bytes) = 2026-09-04 | Dana Okafor signed the 2024 pilot agreement
```

A fresh `git clone` or a `kipi update` rsync sets mtime to checkout time, so every canonical fact in the payload reads as dated today. The sort inside `resolve_canonical` is also file-mtime order, not fact recency, which is not "newest first". No `fresh_days` is declared on canonical in the manifest, so nothing computes on it. Display and sort only.

## What is sound

- Hook stdout corruption. `writing_patterns()` and `_record_recall` `exec_module` two other hook scripts inside the hook process. If either printed at module level the JSON envelope would be garbage. `voice-dna-loader.py` writes only inside a guarded `main()`. End-to-end probe: exit 0, valid JSON, `hookEventName` present, correct `path:line` anchors.
- Concurrency on the misses ledger. `append_bounded` reads, halves and `os.replace`es through one fixed tmp path shared by every process. Three processes, forty appends each, against a ledger seeded to 327,600 bytes so truncation fired on every write: 120 of 120 rows survived, zero malformed lines, no stray tmp. I could not make it lose a row. Dropped.
- The round-3 and round-4 blowup class. 24,000 graph rows, 30,000 index entries: `load_stores` 0.027s, `build_index` 0.072s, `resolve_entities` 0.039s, whole `supply()` 0.376s. The phases outside the deadline are not where the time goes.
- Deadline honesty. With `deadline_s=0.0` on the commitment class and 12 entities, coverage came back `PARTIAL` naming `meetings` as the unsearched required class, not a false FULL.
- The round-1 "DO matches the word do" class, on the name path. `entity_matches` casefolds content while the prompt side is case-sensitive, so a contact named "Ally" would in principle match lowercase "ally" in a store. Grepped the real corpus for the actual contact names plus mark, grant, hope, chase, dana, amber, lisa, david: zero false hits, and "will" is already in `STOPWORDS`. Dropped.
- `capability_index` over the real repo: 225 entries in 0.043s cold.
- Wiring: present in `.claude/settings.json` and `settings-template.json`, template copy guarded with `test -f ... || true`.
- 55 tests pass in 0.66s.

## VERDICT: APPROVE WITH NITS

Fix first: key the supersession group on the resolved entity instead of the raw subject string at `knowledge_supply.py:709`, and add a fixture row whose subject is the alias so the test can go red.

FINDINGS:
minor|Supersession and conflict grouping keys on the raw graph subject, so a status row written under an alias is never marked STALE or CONFLICTING against the canonical-name row|q-system/.q-system/scripts/knowledge_supply.py:709
minor|Canonical and relationship excerpts carry the file mtime in the date slot, so a clone or any unrelated edit re-dates every canonical fact to today while the header claims newest first|q-system/.q-system/scripts/knowledge_supply.py:754
END FINDINGS

PR #304, opened 2026-09-05
ASK-1261 knowledge-supply: one chokepoint for single-token resolution + six captured minors

assafkip:
## What (ASK-1261)

One chokepoint for every single-token entity resolution, plus the six minors captured from PR #302 rounds 4 and 5. Stacked on #303 (Phase 2), which is stacked on #302.

- `single_token_hit(tok, prompt, rule)` is now the only place a one-word name is accepted from a prompt: contact heading, uppercase alias, manifest identifier kind, first-name expansion. The initial-position rule lives inside it. `test_initial_position_has_one_call_site` greps the engine and pins exactly one call site; `test_single_token_chokepoint_all_paths` walks all three paths through sentence-, line- and bullet-initial prompts (none fire) and mid-sentence (each fires). Round 1 minor 4 and round 5 major on #302 were this defect on two paths.
- sp-ca1769db: an unreadable canonical, relationships or decisions file is a recorded problem; a class whose every file failed to read is UNREADABLE and a required one drops coverage to PARTIAL.
- sp-a4a5028a: a non-object, non-list `open-loops.json` is a recorded problem, not an exception past the pass.
- sp-1b3ef442: dedupe is per entity, so a line naming two entities renders under both sections.
- sp-67d54572: CONFLICTING only when the two newest rows share a date AND differ; when they agree, the older row is STALE.
- sp-c9b6401d: the misses bigram scan overlaps (lookahead) and drops openers and articles as a first token, so "Ping Sarah Chen about The Acme Contract" records "Sarah Chen" and "Acme Contract".
- sp-d830d71e: an item's text is cut at 600 chars with an explicit marker; the src stays and the verbatim pieces are untouched, so one multi-KB triple cannot carry the ceiling.

## Measured

52 tests. Seven mutations on copies each go red (chokepoint guard, dedupe key, conflict rule, loops guard, read-error recording, opener filter, item cap).

Commit 3d545290 carries this body as its message (a heredoc collision, the second time this session); 0356cc9f is an empty commit with the intended message. Squash-merge collapses both.

## Review posture

One review round, per the DoR on ASK-1261. Anything beyond this DoR is captured, not fixed here.

### DIFF
--- a/q-system/.q-system/scripts/knowledge_supply.py
+++ b/q-system/.q-system/scripts/knowledge_supply.py
@@ -360,9 +360,14 @@ def single_token_hit(tok: str, prompt: str, rule: str) -> bool:
     elif rule == "identifier":
         if len(tok) < 4:
             return False
-        # Positions preserved: casefold plus the same two replacements norm() makes.
-        pat = re.compile(r"(?<!\w)" + re.escape(norm(tok)) + r"(?!\w)")
-        hay = prompt.casefold().replace("-", " ").replace("_", " ")
+        # Matched on the RAW prompt with IGNORECASE and a separator class, so
+        # every offset is exact. A casefolded haystack shifts offsets after any
+        # casefold-expanding character (PR #304 review).
+        parts = re.findall(r"[a-z0-9]+", norm(tok))
+        if not parts:
+            return False
+        pat = re.compile(r"(?<!\w)" + r"[\s_-]+".join(re.escape(p) for p in parts) + r"(?!\w)", re.IGNORECASE)
+        hay = prompt
     elif rule == "contact":
         if len(tok) < 4 or not tok[0].isupper():
             return False
@@ -731,11 +736,18 @@ def resolve_canonical(entity: dict, stores: dict, root: Path, kind: str = "canon
                       cls: str = "canonical") -> list[dict]:
     items = []
     for path in (files if files is not None else stores["canonical_files"]):
+        st = stores.setdefault("read_stats", {}).setdefault(cls, {"files": set(), "failed": set()})
+        st["files"].add(str(path))
         try:
             lines = read_lines(path)
         except OSError as exc:
             # sp-ca1769db: an unreadable file is a recorded problem, never a
-            # silent skip that reads as "present, 0 hits" under FULL.
+            # silent skip that reads as "present, 0 hits" under FULL. The
+            # class is UNREADABLE only when every file failed (read_stats),
+            # never from zero hits: PR #304 review found one unreadable file
+            # among readable ones marking the whole class unreadable on an
+            # entity with no hits, a false PARTIAL.
+            st["failed"].add(str(path))
             if errors is not None:
                 errors[cls] = f"{path.name}: {exc}"
             continue
@@ -755,13 +767,19 @@ def resolve_canonical(entity: dict, stores: dict, root: Path, kind: str = "canon
 
 
 def resolve_blocks(entity: dict, path: Path, root: Path, kind: str, max_lines: int,
-                   errors: dict | None = None, cls: str = "") -> list[dict]:
+                   errors: dict | None = None, cls: str = "", stores: dict | None = None) -> list[dict]:
     """### blocks in relationships.md / decisions.md that mention the entity."""
     if not path.is_file():
         return []
+    st = None
+    if stores is not None:
+        st = stores.setdefault("read_stats", {}).setdefault(cls or kind, {"files": set(), "failed": set()})
+        st["files"].add(str(path))
     try:
         lines = read_lines(path)
     except OSError as exc:
+        if st is not None:
+            st["failed"].add(str(path))
         if errors is not None:
             errors[cls or kind] = f"{path.name}: {exc}"   # sp-ca1769db
         return []
@@ -928,7 +946,9 @@ def assemble(items: list[dict], entities: list[dict], ceiling: int, header_len: 
     seen: set[tuple] = set()
     deduped = []
     for it in items:
-        if len(it["text"]) > ITEM_MAX_CHARS:
+        if it["kind"] == "graph" and len(it["text"]) > ITEM_MAX_CHARS:
+            # Graph triples only: a relationship block is a 12-line excerpt by
+            # design and was being cut at 600 (PR #304 review).
             it["text"] = it["text"][:ITEM_MAX_CHARS] + f" [cut at {ITEM_MAX_CHARS} chars; open src]"
         key = (norm(it["entity"]), it["kind"], norm(it["text"]))   # sp-1b3ef442: a shared line stays under each named entity
         if key in seen:
@@ -1226,9 +1246,9 @@ def supply(root: Path, prompt: str, *, session_id: str, now: dt.date | None = No
                 elif cls == "canonical":
                     got = resolve_canonical(ent, stores, root, errors=problems, cls=cls)
                 elif cls == "relationships":
-                    got = resolve_blocks(ent, paths["relationships"], root, "relationship", 12, errors=problems, cls=cls)
+                    got = resolve_blocks(ent, paths["relationships"], root, "relationship", 12, errors=problems, cls=cls, stores=stores)
                 elif cls == "decisions":
-                    got = resolve_blocks(ent, paths["decisions"], root, "decision", 8, errors=problems, cls=cls)
+                    got = resolve_blocks(ent, paths["decisions"], root, "decision", 8, errors=problems, cls=cls, stores=stores)
                 elif cls == "commitments":
                     got = resolve_commitments(ent, stores, root, window)
                 elif cls == "meetings":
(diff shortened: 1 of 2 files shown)


assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-304-20260904-222110.md` (9326 bytes).

```
FINDINGS:
major|A source class is marked unreadable from zero hits rather than from every file failing, so one unreadable canonical file still reports COVERAGE FULL when another file has hits and reports the whole directory unreadable when it does not|q-system/.q-system/scripts/knowledge_supply.py:1262
minor|The identifier path checks initial-position at an offset from a casefolded haystack against the raw prompt, so one casefold-expanding character shifts every later offset and lets a sentence-initial token fire|q-system/.q-system/scripts/knowledge_supply.py:363
minor|The overlapping bigram scan roughly doubles junk candidates and at the 20-slot cap evicts the real unresolved name from the misses ledger|q-system/.q-system/scripts/knowledge_supply.py:118
minor|The 600-character item cap is calibrated against a multi-kilobyte triple but truncates 3 of the 4 relationship items produced from this repo's own relationships.md|q-system/.q-system/scripts/knowledge_supply.py:1119
END FINDINGS
```

--- reviewer output ---

All four reproducers now run as single-line commands. Review below.

# PR #304 review, knowledge-supply chokepoint

Baseline suite is green, 52 passed. Every command below was run from the worktree root and its real output is quoted.

## Finding 1

**Severity: major.** A source class is marked unreadable from *zero hits*, not from "every file failed to read", so one unreadable canonical file still reports COVERAGE FULL whenever another file has a hit.

**File:** `q-system/.q-system/scripts/knowledge_supply.py:1262`

**The code:** `if present and problems.get(cls) and not cls_items: present = False`. The comment on the next line claims the class is flipped only when every file failed to read. That is not what the predicate tests. Canonical is the only multi-file class, so it is the only place the proxy diverges, and it diverges in both directions.

**Run:** `python3 -c "import importlib.util as U,sys,tempfile;from pathlib import Path;H=Path('q-system/.q-system/scripts').resolve();sys.path.insert(0,str(H));S=U.spec_from_file_location('tks',H/'test_knowledge_supply.py');T=U.module_from_spec(S);S.loader.exec_module(T);R,Q=T.make_instance(Path(tempfile.mkdtemp()));(Q/'canonical'/'pricing.md').write_text('# Pricing. Dana Okafor asked for the rate card.');B=Q/'canonical'/'talk-tracks.md';B.chmod(0);[print(P,'=>',T.run(R,P)['coverage']['verdict'],T.run(R,P)['coverage']['missing_paths']) for P in ('what do we know about Dana Okafor','what do we know about Mark Chen')];B.chmod(0o644)"`

**Output line one:** `what do we know about Dana Okafor => FULL {}`

**Output line two:** `what do we know about Mark Chen => PARTIAL {'canonical': 'q-fix/canonical unreadable'}`

**Why it matters:** One canonical file is unreadable in both runs. The first run hands the model a header saying coverage is FULL, which is the silent-absence class sp-ca1769db was filed to close. The second run calls the whole directory unreadable when one of two files failed. The coverage verdict depends on the prompt rather than on the filesystem. Still a net improvement over the old always-silent skip, which is why this is major and not a block.

## Finding 2

**Severity: minor.** The identifier path checks initial-position at an offset taken from a casefolded haystack against the raw prompt, so one casefold-expanding character defeats the chokepoint.

**File:** `q-system/.q-system/scripts/knowledge_supply.py:363` and `:373`

**The code:** The haystack is `prompt.casefold().replace("-", " ").replace("_", " ")`. Both replaces preserve length. Casefold does not. The German sharp S becomes two characters, the Turkish dotted capital becomes two code points, and the fi ligature becomes two letters. Every later offset shifts, and `is_initial_position(prompt, m.start())` then inspects the wrong character. The comment on line 363 asserts "Positions preserved", which the run below falsifies.

**Run:** `python3 -c "import importlib.util as U;S=U.spec_from_file_location('ks','q-system/.q-system/scripts/knowledge_supply.py');K=U.module_from_spec(S);S.loader.exec_module(K);[print(repr(T),K.single_token_hit('Acme',T,'identifier')) for T in ['Acme owns it','Note. Acme owns it','Straße. Acme owns it']]"`

**Output:** `'Acme owns it' False`, then `'Note. Acme owns it' False`, then `'Straße. Acme owns it' True`

**Why it matters:** A sentence-initial token is correctly suppressed after an ASCII word and fires after a word containing the sharp S. Dropped a level because that character class is uncommon in this founder's prompts, though a PDF paste carrying ligatures is a real producer in this fleet.

## Finding 3

**Severity: minor.** The overlapping bigram scan roughly doubles junk candidates and, at the 20-slot per-prompt cap, evicts the real unresolved name from the misses ledger.

**File:** `q-system/.q-system/scripts/knowledge_supply.py:118`

**The code:** `MISS_OPENERS` filters only the first token of a bigram, so a Title Case run yields a chain of overlapping document-title fragments. That ledger exists to surface names missing from the graph.

**Run:** `python3 -c "import importlib.util as U,re;S=U.spec_from_file_location('ks','q-system/.q-system/scripts/knowledge_supply.py');K=U.module_from_spec(S);S.loader.exec_module(K);PR='Quarterly Business Review Planning Session Agenda Draft Version Two Final Copy Internal Distribution Restricted Security Program Update Roadmap Milestone Tracker Budget Forecast Headcount Plan Vendor Renewal Schedule Compliance Audit Findings Summary Deck then loop in Nina Patel about it';N=[X['candidate'] for X in K.miss_candidates(PR,[])];K.CAP_BIGRAM_RE=re.compile(r'\b([A-Z][\w-]+)\s+([A-Z][\w-]+)\b');O=[X['candidate'] for X in K.miss_candidates(PR,[])];print('NEW count',len(N),'Nina Patel:','Nina Patel' in N);print('OLD count',len(O),'Nina Patel:','Nina Patel' in O)"`

**Output line one:** `NEW count 20 Nina Patel: False`

**Output line two:** `OLD count 13 Nina Patel: True`

**Why it matters:** The cap fills with fragments such as Session Agenda, Two Final, Tracker Budget, and the one actual person name in the prompt never reaches the ledger. Pasted transcripts and agendas are explicitly in scope, since the scanner reads up to 12000 characters.

## Finding 4

**Severity: minor.** The 600-character item cap is calibrated against a pathological multi-kilobyte triple, but it truncates 3 of the 4 relationship items this repo's own file produces.

**File:** `q-system/.q-system/scripts/knowledge_supply.py:1119`

**The code:** The comment frames the cap as stopping one triple from carrying the whole ceiling. The ceiling is 8000 characters, so a 705-character relationship block is 9 percent of it, not a hog. Relationship blocks are the densest class in the bundle.

**Run:** `python3 -c "import types,re;from pathlib import Path;RT=Path('.').resolve();P=RT/'q-system/.q-system/scripts/knowledge_supply.py';M=types.ModuleType('m');M.__file__=str(P);exec(compile(P.read_text(),'ks','exec'),M.__dict__);[print(Path(F).name,'over600:',sum(1 for L in [len(X['text']) for N in set(re.findall(r'(?m)^### ([A-Za-z0-9 .:#-]{3,40})',(RT/F).read_text(errors='replace'))) for X in M.resolve_blocks(dict(name=N.split(':')[0].strip(),kind='contact',aliases=[]),RT/F,RT,KD,MX)] if L>600)) for F,KD,MX in (('q-system/canonical/decisions.md','decision',8),('q-system/my-project/relationships.md','relationship',12))]"`

**Output line one:** `decisions.md over600: 0`

**Output line two:** `relationships.md over600: 3`

**Why it matters:** Measured against the 4 relationship items that file yields, median length 705 and maximum 746. The cut carries an explicit marker and keeps the src pointer, so it is recoverable, but the cap bites the normal case rather than the pathological one it names.

## What is sound

Attacks that ran and the code survived.

- **Seven mutations, all red.** Each mutant was built in memory and all 52 tests re-run against it. Dropping the initial-position rule from the chokepoint killed 4 tests. Dropping the opener filter killed 1. Reverting the dedupe key to the old two-tuple killed 1. Removing the class-unreadable flip killed 1. Killing the identifier branch killed 2. Turning the item cap off killed 1. Nothing survived.
- **The grep-the-tree test has teeth.** Inserting a second `is_initial_position` call site into a copy turned `test_initial_position_has_one_call_site` red. Not decoration.
- **The alias path is not a regression.** The new inline regex inside `single_token_hit` is the identical pattern `word_in_exact` uses, so the two-letter initialism behaviour is unchanged.
- **The loops guard covers both bad shapes.** A bare JSON string, and a dict whose loops key holds a non-list, both land in `problems` with no exception escaping the supply pass.
- **Ledger concurrency is untouched.** Appends still go through an O_APPEND descriptor, so two overlapping sessions cannot truncate each other. This change does not weaken that.

## Verdict

**REQUEST CHANGES.** One major, three minors. Not a block: nothing here is permanent or unrecoverable, and the change is strictly better than the silent skip it replaces.

**Fix first:** `knowledge_supply.py:1262`. Count per-class read failures against the files actually attempted, so canonical reports PARTIAL when any file failed and never reports FULL while one is unreadable.

FINDINGS:
major|A source class is marked unreadable from zero hits rather than from every file failing, so one unreadable canonical file still reports COVERAGE FULL when another file has hits and reports the whole directory unreadable when it does not|q-system/.q-system/scripts/knowledge_supply.py:1262
minor|The identifier path checks initial-position at an offset from a casefolded haystack against the raw prompt, so one casefold-expanding character shifts every later offset and lets a sentence-initial token fire|q-system/.q-system/scripts/knowledge_supply.py:363
minor|The overlapping bigram scan roughly doubles junk candidates and at the 20-slot cap evicts the real unresolved name from the misses ledger|q-system/.q-system/scripts/knowledge_supply.py:118
minor|The 600-character item cap is calibrated against a multi-kilobyte triple but truncates 3 of the 4 relationship items produced from this repo's own relationships.md|q-system/.q-system/scripts/knowledge_supply.py:1119
END FINDINGS

assafkip:
Landed: fast-forwarded into sana/knowledge-supply (PR #302) at b9ada7c5, so #302 carries Phase 1, Phase 2 and ASK-1261 and merges once.

PR #311, opened 2026-09-06
feat(voiceloop): carry the full engine into the skeleton so the fleet stops shipping a subset (ASK-1238)

assafkip:
## What this is

The skeleton's `plugins/kipi-core/voiceloop` was a strict subset of the engine consulting actually runs: **14 `.py` against 43**. This carries the full engine up so the fleet stops shipping a subset. Engine only, per kipi-system-4b: no `q-consult`, no `automation`, no settings, so the fleet dry run's removal set stays readable.

Closes the build half of **ASK-1238**.

## Why it is a defect and not a tidy-up

`plugins/` rsyncs with `--delete`. A sync into consulting's production checkout removes exactly four files:

```
voiceloop/experience.py        voiceloop/tests/test_experience_engine.py
voiceloop/experience_bench.py  voiceloop/tests/test_experience_bench.py
```

`q-consult/pipeline/experience.py:42` does `from voiceloop import experience as _engine` at **module level**, and `cycle.py:54` imports `experience`. So the sync breaks `import cycle`, which is `pipeline.social`, `pipeline.daily` and the 06:05 content-engine job.

It also deadlocks consulting's production branch. Four ask-crm jobs import `q-consult/pipeline/postbook.py`, which is behind main. Main's `postbook` imports `voiceloop.content_key` and `voiceloop.assistant_gate`, which a 14- or 18-file engine tree does not have. The staleness gate's remedy line says *"carry the upstream copy, it is lossless"* and that instruction is currently false.

**Measured, not argued** (throwaway worktree, consulting @ 83b72615):

| carry | `import pipeline.postbook` |
|---|---|
| `postbook.py` alone, exactly as the gate instructs | `ImportError: cannot import name 'content_key'` |
| `postbook.py` **+ the engine tree** | OK, and `pipeline.cycle` / `pipeline.social` OK |

So this PR is what makes the gate's own remedy true again. The fix is the engine, not a wider gate.

## Shape of the diff

The skeleton tree was verified to be a **pure subset** by full file inventory *before* the copy, so nothing is lost: **30 added, 2 modified, 44 files total** (43 `.py` + `requirements-authorship.txt`).

The 2 modified are `selector.py` and `tests/test_no_founder_data.py`. Those are the founder-given-name rewording that the test's own docstring predicts, not independent drift.

`kipi-core` bumped **1.11.0 -> 1.12.0**. The plugin cache is version-keyed, so without the bump it keeps serving the stale 14-file copy. The load path, not the file, decides which engine runs.

## Evidence

| check | result |
|---|---|
| `pytest voiceloop/tests/` | **194 passed** |
| `tests/test_no_founder_data.py` | **4 passed** |
| ...under a planted-surname control | **red**, so the green is a measurement |
| `automation/export_voice_loop.py --check` (consulting side) | exit 0 |
| `plugin-version-bump-check.py --against origin/main` | exit 0 |
| lefthook pre-commit (10 gates incl. gitleaks) | all green |
| Gate 1.3b semantic containment | **0 gating**, 7631 advisory |
| capability-gate | outside scope, see below |

**On the public-repo risk.** `kipi-system` is public, so founder-agnosticism is the gate that matters. It is not taken on trust: planting the founder's surname in `selector.py` turned `test_no_founder_data.py` red, so the passing run on the real tree is a measurement rather than a silence.

**On Gate 1.3b**, which was flagged as likely to red on our dated scar comments: it does not. The scanner demonstrably reads the ported files (13 findings originate inside `voiceloop/`), and all 13 are `unclassified_populated_record`, which is an advisory class by definition. Gating findings are classified facts on a propagating path; this tree produces none.

**On capability-gate**: its `SCAN_ROOTS` are `q-system/.q-system/scripts` and `q-system/.q-system`. Every path in this diff is under `plugins/`, so the gate is structurally blind to it in both directions. Not a claim that it passes because of this change, just that this change cannot move it.

## Handoff, please do not merge silently

kipi-system-4b runs nothing against consulting until this is on `main`. **The merge sha is the handoff token.** On merge, 4b ff-pulls `kipi-scheduled`, re-runs the dry run against that main, **re-measures** the consulting restore pin rather than reusing it (it changes meaning once the engine is in the skeleton), and stages a fresh consulting command.

One thing that moved while this was being built: consulting `origin/main` advanced `17e2d5dd -> eb01dc2f`, which added `commitments.py` to the carry set alongside `postbook.py`. `commitments.py` imports no voiceloop and is a plain lossless carry. The carry set grows with every merge while the branch is frozen, which is the cost of waiting.

https://claude.ai/code/session_01UD1Uhg8HiDKcs8XzTyFfhD

### DIFF
--- a/plugins/kipi-core/.claude-plugin/plugin.json
+++ b/plugins/kipi-core/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "kipi-core",
-  "version": "1.12.0",
-  "description": "Core founder OS — voice, research, AUDHD, LinkedIn/brand, RCA skills + kipi-mcp server",
+  "version": "1.12.1",
+  "description": "Core founder OS \u2014 voice, research, AUDHD, LinkedIn/brand, RCA skills + kipi-mcp server",
   "author": {
     "name": "Assaf Kipnis",
     "email": "[REDACTED]"
--- a/plugins/kipi-core/.claude-plugin/plugin.json
+++ b/plugins/kipi-core/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "kipi-core",
-  "version": "1.11.0",
-  "description": "Core founder OS \u2014 voice, research, AUDHD, LinkedIn/brand, RCA skills + kipi-mcp server",
+  "version": "1.12.0",
+  "description": "Core founder OS — voice, research, AUDHD, LinkedIn/brand, RCA skills + kipi-mcp server",
   "author": {
     "name": "Assaf Kipnis",
     "email": "[REDACTED]"
--- a/plugins/kipi-core/voiceloop/archetype.py
+++ b/plugins/kipi-core/voiceloop/archetype.py
@@ -1,0 +1,249 @@
+#!/usr/bin/env python3
+"""Post STRUCTURE, and only structure. The layer between his voice and the platform.
+
+why this exists (founder-directed 2026-08-24, verbatim: "I want the voice to change so it
+writes the posts within the parameters of these archetypes. It needs to keep everything
+else and not change except the structure of the post."):
+
+`x-viral-template.md` v2 measured within-author lift across 5,722 posts from 51 accounts
+and found the shape of a post moves engagement independently of what it says. A thread
+starter with a colon-ending setup line runs 1.67x its author's own median. A single short
+line with no artifact runs 0.78x, the worst combination in the corpus, and that is the
+shape the first version of the template recommended first.
+
+## The line this module is not allowed to cross
+
+`build_idea_prompt` was reset to "idea + voice file, that's it" on 2026-08-13 after a
+constraint stack went 0-for-7 on the founder's read in one evening. That stack failed
+because it dictated CONTENT: story frames, source rules, what a post should be about,
+which register to use. It laundered his words.
+
+This module dictates SHAPE: how many lines, what the opening line does, where a list
+goes, whether the post opens a reply chain. It says nothing about word choice, subject,
+register, or what the post argues. That distinction is the entire justification for
+re-adding a block to that prompt, and `test_archetype.py` holds it: the rendered section
+is asserted to carry no worked example and no content directive.
+
+If a future edit puts a sentence about WHAT to say into `config/post-archetypes.json`,
+the 2026-08-13 failure is back and this docstring is the record of what it cost.
+
+## Single source of truth
+
+The archetypes live in `config/post-archetypes.json`, read every render, the same shape
+`channel_guidance` uses and for the same reason: the copy belongs to whoever owns the
+copy, and editing a data file must not require a code change. A missing or malformed
+file degrades to NO structural block rather than raising. A generator that refuses to run
+because a config moved is worse than one that writes the way it did last week.
+"""
+from __future__ import annotations
+
+import json
+import os
+import re
+
+# NO CONFIG PATH LIVES HERE, deliberately (2026-09-05, extraction slice 7).
+#
+# This module used to resolve `config/post-archetypes.json` from __file__. Inside one
+# deployment that is correct and invisible; inside a package that ships fleet-wide it
+# points every operator at whichever table sits beside the code. And the degradation
+# is silent by design: `load` returns {} on any read failure and never raises, so a
+# move would have left `select` choosing from an EMPTY archetype table without one
+# test going red. Same trap as `form` in slice 4b, one file over.
+#
+# So `load` takes its path. The deployment binds it.
+
+# The character half of the retired hot-take finding. The corpus measured 0.78x on
+# posts that were one line AND under this length AND carried no artifact; each
+# condition on its own measures something different, so the card checks all three.
+HOT_TAKE_MAX_CHARS = 140
+
+# The card is advisory text for the founder, never a gate. Nothing in this module can
+# refuse a draft; the gate stack remains the only thing that can.
+_UNKNOWN = {"id": None, "name": None}
+
+
+def load(path):
+    """The archetype table, or {} if it cannot be read. Never raises."""
+    try:
+        with open(path, encoding="utf-8") as handle:
+            data = json.load(handle)
+    except (OSError, ValueError):
+        return {}
+    return data if isinstance(data, dict) else {}
+
+
+def _archetypes(data):
+    return (data or {}).get("archetypes") or {}
+
+
+def select(idea_text, channel, path, forced=None):
+    """Pick the archetype for this idea. Deterministic, and it shows its work.
+
+    Returns (archetype_id, archetype_dict, why). `why` is a sentence naming the signal
+    that decided it, because a silent structural choice is one the founder cannot argue
+    with and this whole layer is a hypothesis he is meant to be able to overrule.
+
+    Signal matching is on WHOLE WORDS for alphabetic signals. The first version matched
+    substrings and 'first' fired inside 'first-party', which routed an idea about data
+    into the how-to skeleton for a reason nobody could see. Non-alphabetic signals ('%',
+    '$') keep substring matching because a word boundary around punctuation matches
+    nothing.
+    """
+    data = load(path)
+    table = _archetypes(data)
+    if not table:
+        return None, {}, "no archetype config, structure left to the writer"
+
+    if forced:
+        if forced not in table:
+            raise UnknownArchetype(
+                f"{forced!r} is not an archetype; known: {sorted(table)}")
+        return forced, table[forced], f"forced to {forced} by the caller"
+
+    text = (idea_text or "").lower()
+    # Ordered, most specific first. `default` is consulted last and is the one with the
+    # strongest evidence, so an idea that matches nothing still gets the best shape.
+    for arch_id in ("receipt", "howto"):
+        entry = table.get(arch_id)
+        if not entry:
+            continue
+        for signal in entry.get("signals") or []:
+            token = signal.lower()
+            if token.isalpha():
+                # NOT \b: a hyphen IS a word boundary, so \bfirst\b matched inside
+                # 'first-party' and routed a data idea into the how-to skeleton for a
+                # reason nobody could see in the trail (caught by
+                # test_signals_match_whole_words_not_substrings, 2026-08-24).
+                hit = re.search(rf"(?<![\w-]){re.escape(token)}(?![\w-])", text)
+            else:
+                hit = token in text
+            if hit:
+                return arch_id, entry, (
+                    f"the idea mentions {signal!r}, which is the {arch_id} signal")
+
+    default_id = (data.get("default") or "thread")
+    entry = table.get(default_id)
+    if not entry:
+        return None, {}, "the configured default archetype is missing from the table"
+    return default_id, entry, (
+        "nothing more specific matched, so this is the default, "
+        "which also carries the strongest evidence in the corpus")
+
+
+class UnknownArchetype(ValueError):
+    """A forced archetype the config has no entry for. Loud, because the quiet version
+    silently writes the default under the caller's chosen name."""
+
+
+def skeleton_section(entry, channel="x"):
+    """The STRUCTURE instruction handed to the writer. Structure only.
+
+    Returns "" for an empty entry so a degraded config produces the pre-2026-08-24
+    prompt exactly, rather than a prompt with an empty heading in it.
+    """
+    if not entry:
+        return ""
+    lines = entry.get("skeleton") or []
+    if not lines:
+        return ""
+    out = ["STRUCTURE (shape only; the words, the subject and the voice stay his):"]
+    out.extend(f"- {line}" for line in lines)
+    lo, hi = (entry.get("lines") or [None, None])[:2] or (None, None)
+    if lo and hi:
+        out.append(f"- The finished post is between {lo} and {hi} lines.")
+    return "\n".join(out)
+
+
+def _line_count(text):
+    return len([line for line in (text or "").split("\n") if line.strip()])
+
+
+def posting_card(entry, arch_id, why, draft, channel, path):
+    """How to actually post it. Advisory, deterministic, printed under the draft.
+
+    This is the half of the founder's ask that the writer cannot answer: whether to open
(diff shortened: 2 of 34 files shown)


assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-311-20260906-115026.md` (6801 bytes).

```
FINDINGS:
major|The registered verify suite passes unchanged with 15 of the 34 carried modules deleted, so the PR's "194 passed" evidence exercises none of critic, revise, post_repair, x_format, archetype or 10 others|.verify-suites:71
minor|The founder-data guard scans only .py/.json/.jsonl/.md, exempting the .txt file this PR adds to the tree and any future .txt or .yaml|plugins/kipi-core/voiceloop/tests/test_no_founder_data.py:54
nit|The vacuous-pass floor asserts at least 7 files against a 44-file tree, so a walk returning 8 files would pass and grep almost nothing|plugins/kipi-core/voiceloop/tests/test_no_founder_data.py:60
END FINDINGS
```

--- reviewer output ---

Read the PR, ran the suite, mutated it, and probed the public-repo guard. Findings below.

## Findings

### 1. major — the registered verify suite is blind to 15 of the 34 modules this PR carries

`.verify-suites:71` registers `plugins/kipi-core/voiceloop` as a suite. The PR's evidence table leads with `pytest voiceloop/tests/` → **194 passed** as the check that the carry is sound. That suite does not execute `critic.py` (620 lines), `revise.py`, `post_repair.py`, `x_format.py`, `archetype.py`, `figure_gate.py`, `sameness.py`, `prompt_render.py`, `luar_scorer.py`, `luar_env_backend.py`, `reply_format.py`, `content_key.py`, `gate_and_judge.py`, `gate_walk.py`, or `voicefp_rules.py`. About 3,600 of the 8,055 added lines.

Reproducer, run against a copy in `$TMPDIR` (repo untouched):

```
python3 - <<'EOF'
import os, shutil, tempfile, subprocess
SRC=".../plugins/kipi-core/voiceloop"
T=tempfile.mkdtemp(); dst=os.path.join(T,"voiceloop")
shutil.copytree(SRC,dst,ignore=shutil.ignore_patterns("__pycache__"))
for m in ["critic","archetype","post_repair","revise","x_format","figure_gate",
          "sameness","reply_format","content_key","gate_and_judge","gate_walk",
          "voicefp_rules","prompt_render","luar_scorer","luar_env_backend"]:
    os.remove(os.path.join(dst,m+".py"))
print(subprocess.run(["python3","-m","pytest",os.path.join(dst,"tests"),"-q"],
      capture_output=True,text=True).stdout[-200:])
EOF
```

Real output:

```
DELETED: 15 modules / 34 total
194 passed in 0.44s
```

Delete a third of the tree, the registered gate still reports green. `test_leaf_gates_engine.py` does import seven of the gates, but it asserts an AST property (no reach-back into a deployment), not behavior, so that is not coverage either.

Why it matters here specifically: this PR's thesis is that the skeleton becomes the canonical home of the engine, so skeleton-side edits are now the expected path. `plugins/` fans out with `kipi update` to the fleet, and `q-consult/pipeline/cycle.py` imports this code for the 06:05 content-engine job. An edit to `critic.py` made in the skeleton ships to production with a green 194. That is the same subset problem this PR exists to fix, one layer down: the code half was carried, the test half was not.

Honest mitigations, stated: nothing is broken today, the modules are inert in the skeleton (only `voice_ref.py` imports voiceloop here, and only `assemble`/`corpus`/`selector`), consulting keeps its own suites for these modules, and my reproducer is a constructed deletion rather than an input a producer emits. It is recoverable with one revert.

### 2. minor — the public-repo founder-data guard does not scan the file type this PR adds

`plugins/kipi-core/voiceloop/tests/test_no_founder_data.py:54` walks the tree filtering on `.py`, `.json`, `.jsonl`, `.md`. This PR adds `requirements-authorship.txt`, the first `.txt` in that tree, and it is exempt.

The PR body rests the public-repo argument on this guard: "founder-agnosticism is the gate that matters. It is not taken on trust."

Reproducer, on a copy:

```
open(dst+"/requirements-authorship.txt","a").write(
    "# corpus dir: /Users/kipnis/q-consult/voice  askconsulting.io\n")
open(dst+"/leak.yaml","w").write("author: Assaf Kipnis\ncorpus: q-consult\n")
open(dst+"/leak.txt","w").write("pig butchering notes, Kipnis\n")
pytest tests/test_no_founder_data.py -q
```

Real output:

```
planted: .txt (requirements-authorship.txt + leak.txt) and .yaml
4 passed in 0.02s
```

Four of the six FORBIDDEN probes planted, guard green. Narrow: the real `.txt` is clean, and the exporter is a second line of defence on the public mirror. Read the guard's silence as covering four extensions, not the tree.

### 3. nit — the vacuous-pass floor is stale by 5x

`test_no_founder_data.py:60` asserts `len(_tree_files()) >= 7`. The tree is 44 files. A walk that broke and returned 8 would pass the floor and then grep 8 files while reporting a clean tree.

## What is sound

**No PII, no credentials, no absolute paths.** Grepped the 8,055 added lines for `assafkip`, `Kipnis`, `/Users/`, `q-consult`, `api_key`, `secret`, `sk-ant`. The only hits are the guard's own probe list at `test_no_founder_data.py:25-45`. Clean for a public repo.

**The 2GB torch stack is not a fleet dependency.** `__init__.py` is docstring-only with no eager imports, and the `torch` import is lazy inside a function at `luar_scorer.py:195`. I imported all 34 modules in a loop: `FAILED: none`, stdlib only. `kipi update` will not hand 27 instances an unresolvable requirement.

**The hook-invoked entry point works from any cwd.** `voice_ref.py:35-41` does `from voiceloop import ...` after inserting `PLUGIN_ROOT` at line 39, so running the script by absolute path from `/tmp` resolves. I looked for the classic sys.path bug and it is already fixed with a scar comment.

**The 3am model-call paths are bounded.** `revise.py:311` and `prompt_render.py:162` both carry a 120s timeout, both require an explicit `--model` with no default tier (`revise.py:300` raises without one), and both raise under `PYTEST_CURRENT_TEST`. No hang-forever, no unpinned model riding the default.

**channel_registry degrades correctly.** No `voice-channels.json` exists in the skeleton, and `load(None)` returns `DEFAULT` byte-identically to pre-registry behavior. It refuses an empty `scopes` or `assembled` axis rather than grading nothing and reporting healthy, which is the exact "check that cannot fail" shape I was hunting for.

**194 tests, zero skips**, confirmed with `-rs`. The green is real for what it covers.

## Verdict

**REQUEST CHANGES.** Not BLOCK: the carry is additive, the modules are inert in the skeleton, and a single revert undoes it.

The one thing to fix first: carry the test files for those 15 modules alongside the code, or, if they genuinely belong to the deployment, record the gap with `prd_runner.py spillover add` and narrow the evidence line in the PR body so "194 passed" is not read as validating `critic.py`. The suite that is registered as this path's gate currently cannot see 44% of the path.

FINDINGS:
major|The registered verify suite passes unchanged with 15 of the 34 carried modules deleted, so the PR's "194 passed" evidence exercises none of critic, revise, post_repair, x_format, archetype or 10 others|.verify-suites:71
minor|The founder-data guard scans only .py/.json/.jsonl/.md, exempting the .txt file this PR adds to the tree and any future .txt or .yaml|plugins/kipi-core/voiceloop/tests/test_no_founder_data.py:54
nit|The vacuous-pass floor asserts at least 7 files against a 44-file tree, so a walk returning 8 files would pass and grep almost nothing|plugins/kipi-core/voiceloop/tests/test_no_founder_data.py:60
END FINDINGS

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-311-20260906-121209.md` (5657 bytes).

```
FINDINGS:
minor|requirements-authorship.txt cites test_luar_env_backend.py::test_requirements_file_and_pin_agree as holding PYTHON_PIN to the file, but that test was not carried and the pin mutates green|plugins/kipi-core/voiceloop/requirements-authorship.txt:28
minor|critic.py hardcodes model IDs in plugins/, which model-allocation.md forbids and validate-separation.py's agents-only scan cannot see, so a retired ID ships fleet-wide green|plugins/kipi-core/voiceloop/critic.py:115
minor|30 carried modules ship with no behaviour coverage in the repo that syncs them: emptying critic.py to zero bytes leaves the suite at 229 passed|plugins/kipi-core/voiceloop/tests/test_engine_surface.py:36
nit|PR body says the plugin bump is 1.11.0 -> 1.12.0; plugin.json says 1.12.1|plugins/kipi-core/.claude-plugin/plugin.json:3
END FINDINGS
```

--- reviewer output ---

## Review: PR #311, voiceloop engine port

Read-only. Nothing in the repo was modified; all mutations ran on copies in `$TMPDIR`.

### Finding 1 (minor): a comment names an enforcement test that was not carried

`plugins/kipi-core/voiceloop/requirements-authorship.txt:28` says the interpreter pin and the requirements file are held together: *"Changing one without the other is a defect; test_luar_env_backend.py::test_requirements_file_and_pin_agree holds them together."* That test is not in the ported tree. The pin can drift silently.

Reproducer and real output:

```
tests carried: ['test_channel_registry.py', 'test_engine_surface.py', 'test_experience_bench.py',
 'test_experience_engine.py', 'test_form_engine.py', 'test_leaf_gates_engine.py',
 'test_no_founder_data.py', 'test_length_axis.py', 'test_selector_form_match.py', 'test_voiceloop.py']

grep for the guard the comment names:
 requirements-authorship.txt:28:# a defect; test_luar_env_backend.py::test_requirements_file_and_pin_agree holds

suite with PYTHON_PIN mutated 3.12 -> 3.9 -> rc 0   229 passed in 0.46s
```

What a human does at 3am: nothing, which is the problem. A drifted pin surfaces as `uv` failing to resolve torch, which `encode_via_declared_env` converts to `ScorerUnavailable`, which `_advisory_authorship` converts to a null score at exit 0. Same silent-null shape the file's own 2026-09-06 scar comment describes. Two ways out: carry `test_luar_env_backend.py`, or delete the sentence that claims a guard exists.

### Finding 2 (minor): first model IDs in `plugins/`, invisible to the only checker

`plugins/kipi-core/voiceloop/critic.py:115-116` hardcodes `MODEL_QUALITY = "claude-sonnet-5"` and `MODEL_STYLE = "claude-haiku-4-5"`. `.claude/rules/model-allocation.md` says model IDs live only in agent frontmatter and the paired validator table. `validate-separation.py:78` only reads `.claude/agents/*.md`.

```
MODEL_TIERS: {'haiku': {...}, 'sonnet': {'claude-sonnet-5'}, 'opus': {'claude-opus-4-8'}}
validator verdict on .claude/agents: []
suite with RETIRED id 'claude-sonnet-4-6' planted in critic.py -> rc 0   229 passed
```

`claude-sonnet-4-6` is the exact ID the fleet already had retired under it once. Today's values are correct, so this is deferred harm on a constructed input, which is why it is minor and not major. Cost is real on the next generation bump: `kipi check` stays green while 26 instances shell a dead ID, and critic fails closed on quality rows.

### Finding 3 (minor): the repo that ships the engine cannot see it break

Emptying a carried module to zero bytes leaves the whole suite green:

```
MUT3 empty critic.py, FULL suite -> rc 0   229 passed in 0.47s
```

The surface gate's docstring discloses this honestly ("It is NOT a behaviour suite... Behaviour lives instance-side"). Capturing it anyway because of where it lands: `plugins/` is not in `INSTANCE_OWNED_SUBTREES` (`kipi-update.sh:64-72`), so the skeleton copy overwrites every instance, and the one instance holding behaviour coverage is the one the skeleton now overwrites. A regression introduced skeleton-side ships to 26 instances green.

### Finding 4 (nit): PR body version does not match the diff

Body says "kipi-core bumped 1.11.0 -> 1.12.0". `plugins/kipi-core/.claude-plugin/plugin.json` says `1.12.1`.

### What is sound

- Deleted a carried module. `test_every_shipped_module_imports[critic]` plus the set-equality test both went red. The manifest-not-walk design does what its docstring claims.
- Added an unregistered module. `test_no_module_on_disk_is_unregistered` went red. Drift direction holds too.
- Ran the surface gate standalone and from repo root. 35 passed both ways, so it does not depend on a sibling test having seeded `sys.path` first.
- Checked all 33 modules for import-time side effects. 35 imports in 0.03s, no network, no writes, no heavy top-level deps.
- Checked `channel_registry.resolve` for a missing-registry path. Returns None to `DEFAULT`, and `q-system/.q-system/data/` genuinely has no `voice-channels.json` in this tree, so the no-registry path is the one that runs and it is byte-identical to the old literals.
- The founder-data guard's binary-denylist inversion is correct: `requirements-authorship.txt` is inside the scanned set, which the old allowlist would have skipped.

One thing I could not test: whether the skeleton snapshot is byte-equal to consulting's live engine. The review sandbox is scoped to this worktree, so `/Users/assafkipnis/projects/consulting` is unreadable here. That is the one place a real regression could hide, and no gate in this repo can see it either.

**VERDICT: APPROVE WITH NITS.** Fix first: carry `test_luar_env_backend.py` or delete the sentence in `requirements-authorship.txt:28` that claims it guards the pin.

FINDINGS:
minor|requirements-authorship.txt cites test_luar_env_backend.py::test_requirements_file_and_pin_agree as holding PYTHON_PIN to the file, but that test was not carried and the pin mutates green|plugins/kipi-core/voiceloop/requirements-authorship.txt:28
minor|critic.py hardcodes model IDs in plugins/, which model-allocation.md forbids and validate-separation.py's agents-only scan cannot see, so a retired ID ships fleet-wide green|plugins/kipi-core/voiceloop/critic.py:115
minor|30 carried modules ship with no behaviour coverage in the repo that syncs them: emptying critic.py to zero bytes leaves the suite at 229 passed|plugins/kipi-core/voiceloop/tests/test_engine_surface.py:36
nit|PR body says the plugin bump is 1.11.0 -> 1.12.0; plugin.json says 1.12.1|plugins/kipi-core/.claude-plugin/plugin.json:3
END FINDINGS

PR #312, opened 2026-09-06
fix(updater): the rename hook never rewrites a path the skeleton ships

assafkip:
## What

`kipi-update-voiceloop-migrate.py` (the pre-rsync rename hook) no longer rewrites any path the sync is about to deliver, and never rewrites prose in a `.py` file. Three commits, each answering a measured gap: (1) skeleton-delivered paths are the sync's; (2) comments and docstrings are prose; (3) reviewer round 1: "delivered" is derived from the updater's own exclude arrays and the instance's `subtree_prefix`, not from skeleton presence.

## Why (measured 2026-09-06)

- The skeleton's own `plugins/kipi-core/voiceloop/__init__.py` names the old package on purpose, in the why-comment that documents the rename. The hook's plain token swap turned it into "this package was called `voiceloop` here and the exporter renamed it to `voiceloop`". Every fleet sync rewrote it in every instance, committed it, and the rsync put the skeleton's copy back (`rewritten=1` on all 22 instances, 7 on the one that also carried six skeleton-shipped q-system scripts and tests). sp-b0389e48.
- On the one instance whose pre-commit compares the engine to its public mirror, that one-line rewrite was a diff against the mirror, the gate refused the commit, the hook reported `migration failed`, and the rsync never started. Log: the founder's `--only` run at 13:37 PDT.
- "Skip comments" cannot work: the word legitimately appears in prose that must keep it, and a regex cannot tell a comment that documents the old name from one that uses it. Rewording the comment would widen the gate.

## Verification (ran, not assumed)

- `test_voiceloop_migrate.py`: 27 pass (4 new). Reproducer uses the real comment shape; its negative twin runs under an EMPTY skeleton and asserts the comment IS rewritten into nonsense, so the exemption is what holds the line. Self-skeleton guard: `plan(r, skeleton=r)` exempts nothing. Move case: package still moves, shipped init left for the rsync, instance-only module inside the package still migrated.
- Mutation: `_shipped_by_skeleton` forced to `False` in a scratch copy: reproducer RED, the other three GREEN.
- Shadow check: 10 findings, identical to main (one new parameter rebind caught and renamed before commit).
- Production caller shape, read-only `--json` plan over the live refused instance: main's hook `rewrite=7`; this hook `rewrite=0`, all 7 listed under `shipped_by_skeleton` (1274 shipped paths), instance tree untouched.
- `test-propagation-entrypoints.py` + `separation/test_update_propagation.py`: 15 pass here and on main.

## Second commit: prose in a .py file is never rewritten

The aborted run wrote one more file than the log showed: the instance's own `automation/export_voice_loop.py`, rewritten and left STAGED, its RENAMES why-comment turned into an identity pair described as a rename. Instance-owned, so the skeleton-shipped rule does not reach it. The two collisions are different token types: the engine's `__init__.py` names the old package inside its module DOCSTRING; the exporter names it in two COMMENT tokens. A regex cannot tell either from a string literal that is a module path, which is exactly what must migrate. Python's tokenizer and parser can.

- For `.py`, `rewrite_text` skips every COMMENT token and every docstring (first-statement STRING, the `ast.get_docstring` rule) and still rewrites NAME tokens, string literals and filenames. `plan()` and `apply()` both go through it, so they cannot disagree about prose. ast columns are bytes, tokenize columns are characters; converted, and tested on a non-ASCII line.
- A `.py` that carries the old name and cannot be parsed is NOT swapped blind: `plan()` reports it under `unparseable`, `apply()` refuses the instance (the updater's dirty-tree posture).
- The other REWRITE_EXTS have no classifier and keep the blanket swap; a test pins that boundary.

Verification after the second commit: 33 tests (fixtures now mirror the real shapes: docstring in the shipped init, code in the shipped helper; the negative twin for the prose rule removes span detection and watches the docstring get rewritten); mutation with the skeleton rule disabled leaves only its reproducer red; shadow check at main's 10; plan mode over the live refused instance after its exporter was restored: main's hook wants 8 rewrites, this branch wants 0 with `needs_work` False and `unparseable` empty, tree untouched.

## Third commit: the exemption is what the sync delivers (reviewer round 1, major + minor)

"Exists in the skeleton" was a proxy for "the sync will overwrite it". Where they differ: a template under an `INSTANCE_OWNED_SUBTREES` dir, a file at the skeleton root, the instance's merged `.claude/settings.json`, a path matching `PLUGIN_COPY_EXCLUDES`. Exempting those left the old name in place with `verified=True`. And the skeleton's `q-system/<x>` lands at `<subtree_prefix>/<x>`, null for one registered instance.

- `_delivered_by_sync` judges each path from the three facts the rsync uses: skeleton `q-system/` mapped through the prefix minus `INSTANCE_OWNED_SUBTREES`; skeleton `plugins/<name>/` minus `PLUGIN_COPY_EXCLUDES`.
- Both arrays are read from `kipi-update.sh` itself by `_bash_array`, which reads the literal list the way bash does; `test_exclude_lists_derive_from_the_updater_and_match_bash` compares the parse with bash's own evaluation (floor non-empty, bound equal). No hand copy.
- The prefix comes from `instance-registry.json` beside the skeleton (`DEFAULT_PREFIX` for an unknown repo). Unreadable arrays exempt nothing and the plan warns: the miss is churn, never a stranded import.
- Plan key renamed `delivered_by_sync`; `prefix` and `warnings` added.

Verification: 38 tests; mutation with `_delivered_by_sync` forced False leaves exactly the three delivery tests red; shadow check 10 = main; live plan over the refused instance: main 8 rewrites, this branch 0, 1165 delivered paths, prefix `q-system`, no warnings, tree untouched; the three updater test files together: 53 pass.

## Out of scope, captured

- The hook still reads a staged edit that removes the token as its own interrupted rewrite (`staged_rewrites`) and would sweep it into its commit. Pre-existing. The one live case was the hook's own leftover from the abort, restored to HEAD by its owner.
- The abandoned-instance cleanup (staged leftovers outside the dirty-rule paths after a refused pre-sync commit) is its own item; the sync owner is capturing it.
- `kipi-update.sh:1642` deletes `.git/*.lock` unconditionally before every sync: sp-2c1bcc3f.

https://claude.ai/code/session_01JQZrCCPLpc8Wr4t8aSB1V9

### DIFF
--- a/kipi-update-voiceloop-migrate.py
+++ b/kipi-update-voiceloop-migrate.py
@@ -60,12 +60,17 @@ file carrying the name that cannot be parsed refuses the instance rather than
 being swapped blind. The other REWRITE_EXTS have no such classifier and keep
 the blanket swap, comments included.
 
-A path the SKELETON ships is never rewritten, whatever its extension: the rsync
-this script runs ahead of delivers the skeleton's copy of it seconds later, so
+A path the SYNC DELIVERS is never rewritten, whatever its extension: the rsync
+this script runs ahead of writes the skeleton's copy over it seconds later, so
 the rewrite is churn at best and a refused commit at worst (2026-09-06: the
-engine's own __init__.py documents the rename in a why-comment that names the
-old package on purpose; see SKELETON_ROOT). Instance-only files, including ones
-that live inside q-system/, do not exist in the skeleton and are migrated.
+engine's own __init__.py documents the rename in its docstring, naming the old
+package on purpose). Delivered means the skeleton's q-system/ tree mapped
+through the instance's subtree_prefix minus INSTANCE_OWNED_SUBTREES, plus the
+skeleton's plugins/<name>/ trees minus PLUGIN_COPY_EXCLUDES, both arrays read
+from kipi-update.sh itself (see SKELETON_ROOT). Everything else the instance
+holds, including a file the skeleton ships as a template under an owned
+subtree, a file at the skeleton root, or its merged .claude/settings.json, is
+the instance's and is migrated.
 """
 from __future__ import annotations
 
@@ -156,14 +161,35 @@ SKIP_PATH_PARTS = (os.path.join("tests", "fixtures") + os.sep,)
 #     commit, this script reported "migration failed", and the rsync it exists
 #     to prepare never started (sp-b0389e48, sp-2c1bcc3f's neighbour).
 #
-# The test is exact, not a path heuristic: if the same relative path exists in
-# the skeleton tree, it is the skeleton's. An instance-only file INSIDE q-system/
-# (the updater preserves those) does not exist in the skeleton and is still
-# migrated. The package MOVE (step 1 of apply) is untouched by this rule: the
-# old package name never exists in the skeleton, so its files are seen at their
-# pre-move paths and the move proceeds; after the move they sit at paths the
-# skeleton ships and the rsync delivers the real ones.
+# THE SET IS WHAT THE SYNC DELIVERS, NOT WHAT THE SKELETON CONTAINS. "Exists in
+# the skeleton" was the first cut and the reviewer of PR #312 named where the
+# two sets differ: a path under an INSTANCE_OWNED_SUBTREES dir (my-project/,
+# memory/, .q-system/data/ ...) that the skeleton happens to ship as a template
+# is never overwritten, the sync excludes it; a file at the skeleton ROOT
+# (kipi-update.sh, lefthook.yml, README.md) is never copied anywhere; the
+# instance's .claude/settings.json is MERGED from a template, its own hook
+# entries survive, so a stale path there is the instance's to migrate. Exempting
+# any of those leaves the old name in place forever and reports verified=True.
+# And the instance keeps its q-system under `subtree_prefix` (null for one
+# registered instance), so the skeleton's q-system/<x> lands at <prefix>/<x>.
+#
+# So the exemption is derived, per file, from the same three facts the rsync
+# uses: the skeleton's q-system/ tree minus INSTANCE_OWNED_SUBTREES, mapped
+# through the instance's subtree_prefix; and the skeleton's plugins/<name>/
+# trees minus PLUGIN_COPY_EXCLUDES. Both arrays are read from kipi-update.sh
+# itself, the file that will run the rsync, so this script cannot drift from
+# it (the deletion guard keeps a hand copy and a test to compare; this one
+# reads the owner). If the arrays cannot be read, nothing is exempt and the
+# plan says so under `warnings`: rewriting a file the sync would have replaced
+# is churn, skipping one it would not have is a stranded import.
+#
+# The package MOVE (step 1 of apply) is untouched: the old package name is never
+# a delivered path, so its files are seen at their pre-move paths and the move
+# proceeds; after the move they sit where the rsync delivers the real ones.
 SKELETON_ROOT = os.path.dirname(SELF_PATH)
+UPDATER_NAME = "kipi-update.sh"
+REGISTRY_NAME = "instance-registry.json"
+DEFAULT_PREFIX = "q-system"
 
 
 def _resolve_skeleton(repo: str, skeleton: str | None) -> str | None:
@@ -176,8 +202,90 @@ def _resolve_skeleton(repo: str, skeleton: str | None) -> str | None:
     return root
 
 
-def _shipped_by_skeleton(rel: str, skeleton: str | None) -> bool:
-    return skeleton is not None and os.path.lexists(os.path.join(skeleton, rel))
+def _bash_array(text: str, name: str) -> list[str]:
+    """The literal elements of `NAME=( ... )` as bash reads them: one element
+    per whitespace-separated word, double quotes stripped with the four
+    backslash escapes bash honours inside them (dollar, quote, backslash,
+    backtick) unescaped, `#` comments dropped. Enough for the two arrays this
+    reads, which are literal lists by design (ASK-772)."""
+    m = re.search(r"^%s=\(\s*\n(.*?)^\)" % re.escape(name), text, re.S | re.M)
+    if not m:
+        return []
+    out: list[str] = []
+    for raw in m.group(1).splitlines():
+        line = raw.strip()
+        if not line or line.startswith("#"):
+            continue
+        for word in re.findall(r'"(?:[^"\\]|\\.)*"|\S+', line):
+            if word.startswith("#"):
+                break
+            if word.startswith('"') and word.endswith('"') and len(word) >= 2:
+                word = re.sub(r'\\([$"\\`])', r"\1", word[1:-1])
+            out.append(word)
+    return out
+
+
+def _updater_lists(skeleton: str) -> tuple[list[str], list[re.Pattern]]:
+    """(INSTANCE_OWNED_SUBTREES, PLUGIN_COPY_EXCLUDES regexes) from the
+    kipi-update.sh that will run the rsync: the one beside the delivered tree,
+    or the one beside this script."""
+    for root in (skeleton, SKELETON_ROOT):
+        updater = os.path.join(root, UPDATER_NAME)
+        if os.path.isfile(updater):
+            break
+    else:
+        raise ValueError(f"{UPDATER_NAME} not found beside the skeleton or this script")
+    text = open(updater, encoding="utf-8").read()
+    owned = _bash_array(text, "INSTANCE_OWNED_SUBTREES")
+    excl = [e.split("::", 1)[1] for e in _bash_array(text, "PLUGIN_COPY_EXCLUDES") if "::" in e]
+    if not owned or not excl:
+        raise ValueError(f"could not read INSTANCE_OWNED_SUBTREES / PLUGIN_COPY_EXCLUDES from {updater}")
+    return owned, [re.compile(r) for r in excl]
+
+
+def _instance_prefix(repo: str, skeleton: str) -> str:
+    """The instance's subtree_prefix from the registry beside the skeleton (or
+    beside this script); DEFAULT_PREFIX for a repo the registry does not know,
+    which is every fixture and the only layout the updater creates."""
+    for root in (skeleton, SKELETON_ROOT):
+        reg = os.path.join(root, REGISTRY_NAME)
+        if not os.path.isfile(reg):
+            continue
+        try:
+            rows = json.load(open(reg, encoding="utf-8")).get("instances") or []
+        except (OSError, ValueError):
+            continue
+        for row in rows:
+            path = row.get("path") if isinstance(row, dict) else None
+            if path and os.path.realpath(os.path.expanduser(path)) == os.path.realpath(repo):
+                return (row.get("subtree_prefix") or "").strip("/")
+        break
+    return DEFAULT_PREFIX
+
+
+def _delivered_by_sync(rel: str, skeleton: str | None, prefix: str,
+                       owned: list[str], plugin_excl: list[re.Pattern]) -> bool:
+    """Would the next sync write the skeleton's copy over this instance path?"""
+    if skeleton is None:
+        return False
+    # q-system half: skeleton q-system/<sub> lands at <prefix>/<sub>, minus the
+    # anchored instance-owned excludes.
+    if prefix:
+        sub = rel[len(prefix) + 1:] if rel.startswith(prefix + "/") else None
+    else:
+        sub = rel
+    if sub and not any(sub == o or sub.startswith(o + "/") for o in owned) \
+            and os.path.lexists(os.path.join(skeleton, "q-system", sub)):
+        return True
+    # plugins half: skeleton plugins/<name>/ lands at plugins/<name>/, minus
+    # PLUGIN_COPY_EXCLUDES (--delete-excluded).
+    parts = rel.split("/")
+    if len(parts) >= 3 and parts[0] == "plugins" \
+            and os.path.isdir(os.path.join(skeleton, "plugins", parts[1])) \
+            and os.path.lexists(os.path.join(skeleton, rel)) \
+            and not any(r.search("/".join(parts[2:])) for r in plugin_excl):
+        return True
+    return False
 
 
 def _exempt(path: str) -> bool:
@@ -320,13 +428,30 @@ def rewrite_text(text: str, ext: str) -> str:
     return _swap_outside(lines, spans)
 
 
-def plan(repo: str, skeleton: str | None = None) -> dict:
+def plan(repo: str, skeleton: str | None = None, prefix: str | None = None) -> dict:
     """Read-only. What this instance needs, without touching a byte.
 
     `skeleton` is the tree the sync delivers (default: the one this script sits
-    in); a path it ships is skipped entirely, see SKELETON_ROOT above."""
+    in); a path the sync would overwrite is skipped entirely, see SKELETON_ROOT
+    above. `prefix` is the instance's subtree_prefix (default: the registry's
+    answer, else DEFAULT_PREFIX)."""
     repo = os.path.abspath(repo)
     delivered = _resolve_skeleton(repo, skeleton)
+    warnings: list[str] = []
+    owned: list[str] = []
+    plugin_excl: list[re.Pattern] = []
+    if delivered is not None:
+        try:
+            owned, plugin_excl = _updater_lists(delivered)
+        except ValueError as exc:
+            warnings.append(f"no path is exempt: {exc}")
+            delivered = None
+    if prefix is not None:
+        sub_prefix = prefix.strip("/")
+    elif delivered is not None:
+        sub_prefix = _instance_prefix(repo, delivered)
+    else:
+        sub_prefix = DEFAULT_PREFIX
     old_pkg = os.path.join(repo, PLUGIN_PARENT, OLD)
     new_pkg = os.path.join(repo, PLUGIN_PARENT, NEW)
     has_old = os.path.isdir(old_pkg)
@@ -346,7 +471,7 @@ def plan(repo: str, skeleton: str | None = None) -> dict:
         if _exempt(path):
             continue
         rel = os.path.relpath(path, repo)
-        if _shipped_by_skeleton(rel, delivered):
+        if _delivered_by_sync(rel, delivered, sub_prefix, owned, plugin_excl):
             shipped.append(rel)
             continue
         base = os.path.basename(path)
@@ -432,10 +557,12 @@ def plan(repo: str, skeleton: str | None = None) -> dict:
     return {
         "repo": repo,
         "skeleton": delivered,
+        "prefix": sub_prefix,
+        "warnings": warnings,
         "package_action": package_action,
         "rewrite": sorted(rewrite),
         "renames": sorted(renames),
-        "shipped_by_skeleton": sorted(shipped),
+        "delivered_by_sync": sorted(shipped),
         "unparseable": sorted(unparseable),
         "history_left": sorted(history),
         "staged_migration": staged_migration,
@@ -489,9 +616,10 @@ def _dirty_tracked(repo: str, rels: list) -> list:
     return sorted(set(out))
 
 
-def apply(repo: str, commit: bool = True, skeleton: str | None = None) -> dict:
+def apply(repo: str, commit: bool = True, skeleton: str | None = None,
+          prefix: str | None = None) -> dict:
     repo = os.path.abspath(repo)
-    p = plan(repo, skeleton)
+    p = plan(repo, skeleton, prefix)
     result = {**p, "moved": False, "rewritten": [], "renamed": [], "committed": False,
               "errors": [], "left_untracked": [], "backed_up": []}
     # Snapshot tracked-ness BEFORE any write, because `git mv` and the rewrites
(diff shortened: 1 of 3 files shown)


assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-312-20260906-135008.md` (6382 bytes).

```
FINDINGS:
major|Exemption keys on skeleton presence, not on the updater's delivered set, so a colliding path under an rsync-excluded instance-owned subtree or at the skeleton root is skipped by the migration, never overwritten, and reported verified=True|kipi-update-voiceloop-migrate.py:167
minor|The exemption ignores subtree_prefix while the sync applies it, so the one subtree_prefix:null instance still gets the churn rewrite this PR removes everywhere else|kipi-update-voiceloop-migrate.py:167
END FINDINGS
```

--- reviewer output ---

# Review of PR #312

## Finding 1 — major

**"The skeleton ships it" is not "the sync delivers it". A colliding path in an rsync-excluded subtree, or at the skeleton root, is now skipped by the migration and never overwritten, and `apply()` still reports `verified=True`.**

`kipi-update-voiceloop-migrate.py:167` exempts any `rel` for which `os.path.lexists(skeleton/rel)` is true. The docstring justifies that with "the rsync this script runs ahead of delivers the skeleton's copy of it seconds later". The updater does not deliver every skeleton path:

- `kipi-update.sh:64-73` plus `rsync_owned_excludes()` at `:105`, applied at `:2291-2295`, anchor `--exclude=/my-project/ /canonical/ /memory/ /output/ /research/ /.q-system/data/ /.q-system/agent-pipeline/bus/`. The skeleton carries 98 rewrite-eligible files under those paths.
- `kipi-update.sh:1665` names the whole sync set: `q-system/`, `.claude/{agents,output-styles,rules}/*.md`, `.claude/settings.json`, `plugins/`. Skeleton root trees (`automation/`, `scripts/`, `sites/`, root scripts) reach no instance, and `plugins/kipi-core/commands/voice-refresh.md:27` tells every instance to run `python3 automation/voice_refresh.py`, so an instance-owned file at that exact colliding path is the designed shape, not an exotic one.
- Order matters too: the q-system rsync failure path is `abandon_instance ... && continue` (`:2296-2300`), and the plugins rsync is 180 lines later at `:2477`. An instance abandoned there ends the run with the migration's skipped rewrites undone and no error about them.

Reproducer, run against the real skeleton in this worktree:

```
skeleton ships q-system/.q-system/data/db-query.py           True
skeleton ships q-system/memory/collection-state.json         True
skeleton ships automation/voice_refresh.py                   True
skeleton ships q-consult/pipeline/voice.py                   False

rewrite            : ['q-consult/pipeline/voice.py']
shipped_by_skeleton: ['automation/voice_refresh.py', 'plugins/kipi-core/voiceloop/__init__.py',
                      'q-system/.q-system/data/db-query.py', 'q-system/memory/collection-state.json']
apply stdout: inst: moved=False rewritten=1 renamed=0 committed=False verified=True

AFTER --apply, files still naming the old package:
   q-system/.q-system/data/db-query.py -> from voicekit import validate
   q-system/memory/collection-state.json -> {"cmd": "plugins/kipi-core/voicekit/validate.py"}
   automation/voice_refresh.py -> from voicekit import refresh
```

Calibration, stated plainly: the path collisions and the exclusions are real, and the producer of stale skeleton copies carrying the token is the one the PR itself measured (7 on the live instance). The token content at one of these excluded paths is mine, constructed. My read-only grep of `/Users/assafkipnis/projects/consulting` was permission-denied in this sandbox, so I could not confirm a live occurrence. The severity stays at major because the miss is silent and the run reports `verified=True` for work that did not happen.

What a human does at 3am: nothing, which is the problem. The voice pipeline fails on `from voicekit import ...` and the migration's own report says clean.

Fix shape: intersect the exemption with the delivered set instead of the skeleton tree. `rel` under `plugins/`, or under the sync prefix minus `INSTANCE_OWNED_SUBTREES`, or `.claude/{agents,output-styles,rules}/*.md`.

## Finding 2 — minor

**The exemption is prefix-blind while the sync is prefix-aware, so the one `subtree_prefix: null` instance keeps the churn rewrite this PR exists to remove.**

`instance-registry.json` has 24 instances at `subtree_prefix: "q-system"` and one, `reddit-build-radar`, at `null`. For that one the sync destination is the instance root, so the skeleton's `q-system/<rel>` lands at `<rel>`. `_shipped_by_skeleton` compares the instance-relative path against `skeleton/<rel>` with no prefix mapping, so it never matches.

```
skeleton ships q-system/.q-system/scripts/plan-lint.py True
rewrite            : ['.q-system/scripts/plan-lint.py']
shipped_by_skeleton: ['plugins/kipi-core/voiceloop/__init__.py']
```

The big offender (`plugins/kipi-core/voiceloop/__init__.py`) is still exempted there because `plugins/` is prefix-independent, so this is incompleteness, not a regression. Bounded to one instance, recoverable.

## What is sound

- **Mutation, exemption disabled** (`_shipped_by_skeleton -> return False`): exactly `test_a_path_the_skeleton_ships_is_the_syncs_not_the_migrations` goes red, 26 pass. The new test is not decoration and the PR's mutation claim reproduces.
- **Mutation, over-exemption** (`return skeleton is not None`): 10 failures including the negative twin and the move case. Both directions of the boolean are pinned.
- **Self-skeleton**: `plan(r, skeleton=r)` resolves to `None` via realpath comparison, so a caller passing the instance as its own skeleton cannot silently exempt the whole tree. The CLI also refuses a repo holding `instance-registry.json`.
- **Package move**: the old package name never exists in the skeleton, so step 1 plans from pre-move paths and still runs; verified by the suite and by reading `apply()`'s re-plan at both `p2` and `after`.
- **Dry vs real drift**: `SKELETON_ROOT` derives from the script's own realpath and `kipi-update.sh:2132` invokes `$SCRIPT_DIR/...`, so a `MODEL_RUN` preview and a real run resolve the same skeleton. No drift.
- **Full suite**: 27 pass.

## VERDICT: REQUEST CHANGES

Not BLOCK. Nothing here is unrecoverable, and this PR strictly writes less than main does.

First thing to fix: finding 1. Scope the exemption to what the updater actually delivers rather than to what the skeleton merely contains, and add a case where a file under an rsync-excluded instance-owned subtree still gets migrated.

FINDINGS:
major|Exemption keys on skeleton presence, not on the updater's delivered set, so a colliding path under an rsync-excluded instance-owned subtree or at the skeleton root is skipped by the migration, never overwritten, and reported verified=True|kipi-update-voiceloop-migrate.py:167
minor|The exemption ignores subtree_prefix while the sync applies it, so the one subtree_prefix:null instance still gets the churn rewrite this PR removes everywhere else|kipi-update-voiceloop-migrate.py:167
END FINDINGS

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-312-20260906-135737.md` (5842 bytes).

```
FINDINGS:
major|Skeleton-shipped exemption covers 210 paths the sync never delivers (q-system instance-owned subtrees, root automation/scripts, .claude/settings.json), so the token survives while plan reports needs_work=False and verified=True|kipi-update-voiceloop-migrate.py:179
minor|The plan CLI line omits `unparseable`, so `kipi update --dry` prints needs_work=True with every counter at zero and no reason before the real run abandons the instance|kipi-update-voiceloop-migrate.py:756
END FINDINGS
```

--- reviewer output ---

## PR #312 review: read-only, findings reproduced

Two findings survived reproduction. One major, one minor.

---

### 1. major — the exemption's premise is false for 210 skeleton paths the sync never delivers

`kipi-update-voiceloop-migrate.py:179` (`_shipped_by_skeleton`), premise stated at `:141-165` and `:63-68`.

The comment says a skeleton path is "overwritten seconds later with the skeleton's bytes." `kipi-update.sh` delivers only `plugins/`, `q-system/` **minus the 7 instance-owned subtrees**, and `.claude/{agents,rules,output-styles}/*.md`. Everything else the skeleton carries is exempted from rewriting and never delivered, so the token survives and the run reports clean.

Reproducer A, the exemption (executed):

```
rewrite            = ['q-consult/pipeline/voice.py']
shipped_by_skeleton= ['plugins/kipi-core/voiceloop/__init__.py',
                      'q-system/memory/open-loops.json',
                      'q-system/output/tool.py']
verified           = True []
needs_work AFTER   = False
   q-system/memory/open-loops.json token present: True
   q-system/output/tool.py         token present: True
   q-consult/pipeline/voice.py     token present: False
```

Reproducer B, the sync excludes those exact paths. Ran `rsync_owned_excludes()` verbatim out of `kipi-update.sh:105`:

```
--exclude=/my-project/
--exclude=/canonical/
--exclude=/memory/
--exclude=/output/
--exclude=/research/
--exclude=/.q-system/data/
--exclude=/.q-system/agent-pipeline/bus/
```

Reproducer C, the size of the surface. Skeleton files with a REWRITE_EXTS extension that a subtree instance never receives: **210**.

```
   q-system 97      <root file> 44     scripts 14
   .review-scratch 12   rescued 8      .codex 7
   automation 6     .prd-os 4          .claude 2
automation sample: ['automation/install-voice-refresh.sh',
 'automation/voice_refresh.py', 'automation/voice-refresh-nudge.sh', ...]
```

`.claude/settings.json` is in that set too: `kipi-update.sh:1336` records that it is generated by `kipi-settings-merge.py`, not copied, so the skeleton's bytes never land there either.

Why it matters at 3am: this is the exact breakage the script exists to prevent, and it now reports `verified=True` and `needs_work=False` while leaving the stale import. A human has to find it by hand.

Honest caveat: the file paths are real producers, the token content in them is my construction. I found no live instance carrying `voicekit` under one of these paths, and the skeleton itself carries the token in only 5 files, all already handled.

Narrower rule that keeps the fix: exempt only paths under the prefixes the sync actually writes, rather than "exists in the skeleton tree."

---

### 2. minor — the plan line reports `needs_work=True` with every counter at zero and no reason

`kipi-update-voiceloop-migrate.py:756`. The non-JSON plan output prints package, rewrite, rename, history_left, needs_work. It never prints `unparseable`, which is the new third input to `needs_work` at `:446`.

`kipi-update.sh:2140` pipes exactly that line through `sed` in dry mode, so it is all the operator gets.

Executed:

```
i: package=already rewrite=0 rename=0 history_left=0 needs_work=True

apply rc: 1
i: moved=False rewritten=0 renamed=0 committed=False verified=False
STDERR:   ERROR: refusing to migrate: 1 .py file(s) carry the old name and cannot
be classified into code and prose; ... automation/legacy.py: SyntaxError: '(' was
never closed
```

The `--dry` preview says work is needed and names nothing. The real run then abandons the instance. One extra field in the plan line closes it.

---

### What is sound

- **Span reconstruction across the whole repo.** Patched `CASE_FORMS` to a token that appears nowhere, then ran `rewrite_text(text, ".py")` over every `.py` in the tree. `scanned 606 identity-mismatch 0 unparseable 0`. No duplication or loss from the multi-span line splicing in `_swap_outside`, and the new refusal branch has zero live triggers in this repo.
- **The tests are not decoration.** Mutated `PROSE_EXTS = {".py"}` to `set()` in a scratch copy: `5 failed, 28 passed`. Mutated `_shipped_by_skeleton` to `return False`: `1 failed, 32 passed`. Both mutants died, each on the test written for it.
- **Self-skeleton.** `plan(r, skeleton=r)` resolves to `None` and exempts nothing, so a caller passing the instance as its own skeleton cannot fake a clean tree. Verified by the suite, and the resolve is realpath-based so a symlinked path does not slip past it.
- **The `\r` guard.** CRLF files pass (`text.replace("\r\n","")` leaves no `\r`); lone-CR files are refused rather than mis-sliced.
- **plan/apply agreement.** `plan()` decides `rewrite` by calling the same `rewrite_text` `apply()` writes with, so a file whose only mention is prose is not queued and a second run is a genuine no-op. `test_a_docstring_only_file_is_not_work` covers it and dies under the prose mutation.
- **Full suite green as delivered:** `33 passed in 1.97s`.

---

**VERDICT: REQUEST CHANGES.** Not a block: nothing here is unrecoverable, and the defect it fixes is live and sync-blocking today.

First thing to fix: scope `_shipped_by_skeleton` to the paths `kipi-update.sh` actually writes, not to the whole skeleton tree. Finding 2 is a one-line print change and can ride along.

FINDINGS:
major|Skeleton-shipped exemption covers 210 paths the sync never delivers (q-system instance-owned subtrees, root automation/scripts, .claude/settings.json), so the token survives while plan reports needs_work=False and verified=True|kipi-update-voiceloop-migrate.py:179
minor|The plan CLI line omits `unparseable`, so `kipi update --dry` prints needs_work=True with every counter at zero and no reason before the real run abandons the instance|kipi-update-voiceloop-migrate.py:756
END FINDINGS

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-312-20260906-140933.md` (8595 bytes).

```
FINDINGS:
minor|plan() computes a "no path is exempt" warning that main()'s human line never prints, so a silent full loss of the exemption looks like a healthy run to kipi-update.sh|kipi-update-voiceloop-migrate.py:888
minor|_delivered_by_sync exempts plugins/<dot-dir>/ paths that managed_plugin_names' */ glob never rsyncs, stranding the old import with needs_work=False|kipi-update-voiceloop-migrate.py:283
minor|After the package move the skeleton-shipped voiceloop/__init__.py keeps the instance's dead voicekit import while apply() reports verified=True|kipi-update-voiceloop-migrate.py:753
minor|A UTF-8 BOM in a token-carrying .py fails ast.parse, lands in unparseable, and abandons the instance's whole sync on every run|kipi-update-voiceloop-migrate.py:427
END FINDINGS
```

--- reviewer output ---

# Review: PR #312 (kipi-system) — rename hook never rewrites a path the sync delivers

Read-only review of the 4 commits on the branch. I ran the suite (39 pass), fuzzed the prose classifier, ran a differential of the exemption set against what `git archive HEAD -- q-system/` actually delivers, and built four scratch instances outside the repo. Nothing was modified.

---

## Findings

### 1. minor — the `warnings` the code computes never reach the only output the updater reads

`kipi-update-voiceloop-migrate.py:447` records `no path is exempt: ...` when the exclude arrays cannot be parsed, then sets `delivered = None`, which silently reverts to the exact pre-PR behavior. The human line at `:888` prints six counters and no warning. `kipi-update.sh:2136` consumes that line; nothing calls `--json`.

The trigger is not exotic: `_bash_array` requires a newline immediately after `(`, so a semantically identical one-line reformat of `INSTANCE_OWNED_SUBTREES` turns the entire exemption off.

Reproducer and real output:

```
bash sees: ['my-project', 'canonical', 'memory', 'output', 'research', '.q-system/data', '.q-system/agent-pipeline/bus']
_bash_array sees: []

plan with the reformatted (bash-identical) array:
  delivered_by_sync: []
  rewrite: ['q-system/.q-system/scripts/tool.py']
  warnings: ['no path is exempt: could not read INSTANCE_OWNED_SUBTREES / PLUGIN_COPY_EXCLUDES from .../kipi-update.sh']
  human line main() prints: inst: package=already rewrite=1 rename=0 unparseable=0 history_left=0 needs_work=True
```

The file the sync delivers is back in `rewrite`, and the log line the operator reads is indistinguishable from a healthy run. `test_exclude_lists_derive_from_the_updater_and_match_bash` would go red on this, so it is guarded at test time, not at runtime. One line in the `print(f"{name}: package=...")` block closes it: `warnings={len(out['warnings'])}` plus the text on stderr.

I constructed the reformat; nobody has made that edit. Severity dropped a level for that, as the standing rule requires.

### 2. minor — a dot-named plugin dir is exempted as delivered and never actually rsynced

`kipi-update-voiceloop-migrate.py:283` tests `os.path.isdir(skeleton/plugins/<name>)`. `managed_plugin_names()` at `kipi-update.sh:149` globs `"$SKELETON_PLUGIN_ROOT"/*/`, which skips dot directories. The two disagree in the direction that strands an import.

Reproducer output, skeleton and instance both carrying `plugins/.hidden/tool.py`:

```
R5 delivered_by_sync: ['plugins/.hidden/tool.py', 'plugins/kipi-core/voiceloop/__init__.py']
R5 rewrite: []
R5 needs_work: False
```

And the copy loop's own glob on the same tree:

```
--- glob */ (what managed_plugin_names uses) ---
managed: kipi-core
--- what actually exists ---
.../skeleton/plugins/.hidden
.../skeleton/plugins/kipi-core
```

So `.hidden/tool.py` keeps `import voicekit.core` forever and every run says `needs_work=False`. `kipi-update.sh:156-162` records that a dot-named dir under `plugins/` was a live fleet condition once. There is none in the skeleton today, so this is constructed input and rated accordingly. Fix is one line: reuse the same `*/` semantics, or test `parts[1]` against a name that does not start with `.`.

### 3. minor — after the package move, the shipped `__init__.py` keeps the dead import and `verified` is True

`apply()` moves `plugins/kipi-core/voicekit/` to `voiceloop/`, at which point `__init__.py` sits on a delivered path and is exempted from the rewrite. `kipi-update-voiceloop-migrate.py:753` then computes `verified` from `after["rewrite"]`, which no longer contains it.

Reproducer output:

```
moved: True verified: True errors: []
rewritten: ['q-consult/pipeline/voice.py']
MOVED __init__.py CONTENT: 'from voicekit.core import Engine\n'
```

The migration commits a package whose `__init__` imports a package that no longer exists, and reports success. The plugins rsync later in the same run overwrites it, so it self-heals on the normal path. It does not self-heal when that block is skipped: `kipi-update.sh:2359` gates the whole config sync on `[ -d "$path/.claude" ]`. `package_action == "move"` is believed absent from the fleet, which is why this is minor rather than major.

### 4. minor — a UTF-8 BOM in a token-carrying `.py` refuses the whole instance, every run

```
bom | RAISED ValueError cannot classify prose from code: SyntaxError: invalid non-printable ch
```

`ast.parse` rejects the leading `\ufeff`, so the file lands in `unparseable`, `apply()` returns at `:847` with `verified=False`, and `kipi-update.sh` calls `abandon_instance` — no rsync, on every run, until a human edits the file. Stripping a leading BOM before `ast.parse` costs one line. I scanned every `.py` in the skeleton: `UNPARSEABLE .py: 0`, so no such file exists today.

---

## What is sound

Attacks that did not break it:

- **Prose classifier.** Fuzzed 21 shapes: decorated functions, `async def`, nested classes, `def f(): """d"""; import voicekit` on one line, CRLF files, form feeds, tabs, raw and bytes string literals, implicit concatenation as the first statement, a comment inside a 3.12 f-string replacement field, a file with no trailing newline, an empty file. Every one migrated the code and preserved the prose, and every result was idempotent. The non-ASCII column conversion is correct: `"""Résumé: café voicekit é"""` stays intact while the next line's import migrates.
- **The negative twins are real tests.** `test_without_the_prose_rule_the_docstring_is_rewritten` monkeypatches `_prose_spans` to return `[]` and asserts the docstring becomes an identity rename. That is the mutation, in-suite, and it is what makes the exemption load-bearing rather than incidental.
- **`test_exclude_lists_derive_from_the_updater_and_match_bash` is not decoration.** It actually shells bash and compares. `_bash_array` is a strict subset of bash (single quotes, backslash escapes and `$` expansion all diverge), and this test is the thing that would catch a drift into that territory.
- **The exemption set matches what the rsync delivers.** I diffed the working-tree exemption against `git ls-tree -r HEAD q-system/` minus the owned subtrees: 1128 delivered, 1140 exempt, 12 divergences — all of them `__pycache__/*.pyc` (in `SKIP_DIRS`, never walked) and three untracked integrity-baseline artifacts. No reachable stranded import there.
- **The plugin exclude regexes match the rsync patterns** on every `PLUGIN_COPY_EXCLUDES` entry I checked, including the `.env` vs `.env.py` split the test pins.
- **It fixes the reported bug on the real shape.** Built an instance from the skeleton's actual `plugins/kipi-core/voiceloop/` plus an instance-owned `q-consult/pipeline/voice.py`: 45 paths exempted, `rewrite: ['q-consult/pipeline/voice.py']`, no warnings, no unparseable. The engine's docstring survives, the instance's own import migrates.
- Dropped after failing to reproduce: a straddling-token bug across a protected span boundary (a COMMENT runs to end of line and a STRING token cannot share characters with adjacent code, so no token can straddle); `RecursionError` escaping the caught tuple; case-collision on macOS; `--exclude=/output/` being directory-only where Python matches a bare file.

---

## VERDICT: APPROVE WITH NITS

Four minors, no majors, nothing permanent or unrecoverable. The core change is well-built: the classifier is the language's own parser rather than a regex, the delivered set is derived from the updater instead of hand-copied, and both rules ship a negative twin that goes red when the rule is removed.

**Fix first: finding 1.** The code already computes the diagnostic; it just never shows it to the one consumer that reads the output. That is the difference between a silent fleet-wide regression and a log line.

FINDINGS:
minor|plan() computes a "no path is exempt" warning that main()'s human line never prints, so a silent full loss of the exemption looks like a healthy run to kipi-update.sh|kipi-update-voiceloop-migrate.py:888
minor|_delivered_by_sync exempts plugins/<dot-dir>/ paths that managed_plugin_names' */ glob never rsyncs, stranding the old import with needs_work=False|kipi-update-voiceloop-migrate.py:283
minor|After the package move the skeleton-shipped voiceloop/__init__.py keeps the instance's dead voicekit import while apply() reports verified=True|kipi-update-voiceloop-migrate.py:753
minor|A UTF-8 BOM in a token-carrying .py fails ast.parse, lands in unparseable, and abandons the instance's whole sync on every run|kipi-update-voiceloop-migrate.py:427
END FINDINGS

PR #295, opened 2026-09-02
ASK-1197: port route-receipt enforcement into the skeleton, with a fanout-loss gate

assafkip:
## Do not merge yet

Based on `prd/morning-brief-learns`, so **PR #294 must land first**. Once it does, this diff reduces to the five files below.

## What this is

`q-system/.q-system/scripts/voice-stop-gate.py` is a Stop hook that lives in the skeleton and reaches 25 registered instances through an rsync `--delete` fanout, so the skeleton copy is the one that survives an update. Measured 2026-09-02 across every registered instance:

| direction | instances | meaning |
|---|---|---|
| defines a function the skeleton lacks | **1** | fanout **DELETES** working code |
| lags the skeleton by the same 6 functions | **24** | fanout **ADDS** them, deletes nothing |

The 1 was `enforce_route_receipt` plus four helpers, shipped into one instance and into no skeleton. That is the voicekit class (19 instances, 2026-08-07).

## Ported

`_route_context`, `_route_draft`, `_receipt_block`, `enforce_route_receipt`, `founder_typed_text`, plus the `RouteBoundaryError` class, the `_META_FLAGS` / `_INJECTED_BLOCK` / `_INJECTED_OPENER` / `_COMMAND_INJECTION_MARK` constants, `_walk_transcript(want_record=)` and the `find_final_user_text` that reads them. New: one `_enforce_route_or_exit` chokepoint so the two `main()` call sites cannot drift apart.

**Nothing was removed.** The drift was bidirectional, so the result is the union: `report_not_checked`, `channel_surface_lint`, `detect_channel`, `resolve_channel_registry`, `_lint_argv` and `_sentence_at` all stay.

## Three outcomes, not two

24 of the 24 instances carrying this file have no `q-consult/pipeline`.

| state | behaviour |
|---|---|
| lane absent, ordinary turn | silent, exit 0, identical to before the port |
| lane absent, `=== ROUTE RECEIPT ===` present | exit 0 with an explicit NOT CHECKED line |
| lane present but unimportable, or raising mid-turn | exit 2, holds the turn |

The middle row is PR #290's scar: the value a missing check returns must differ from the value a clean check returns. The bottom row matters because a Stop hook exiting 1 does **not** hold the turn, so any exception escaping this file fails open.

## The propagation test is deliberately one-directional

[...]

PR #313, opened 2026-09-06
feat(voice-stop-gate): port the live route-receipt and founder-text behaviour into the skeleton (ASK-1197)

assafkip:
## What

Ports the behaviour one instance's `voice-stop-gate.py` has run live since 2026-08 onto the skeleton's current file, verbatim: the route-receipt family (`_route_context`, `_route_draft`, `_receipt_block`, `enforce_route_receipt`, `RouteBoundaryError`), `founder_typed_text`, the harness-flag transcript walk (`_META_FLAGS`, `_walk_transcript(want_record=True)`), and the founder-only `find_final_user_text`. Main's channel-surface family stays. ASK-1197.

## Why (measured 2026-09-06)

Every fleet sync replaced that instance's gate with a skeleton copy that had none of this. Its own suite asserts it (`find_final_user_text` returns the founder's words, never the harness text injected around them), so its pre-commit went red after each sync, the restore was undone by the next sync (twice in one afternoon), and every commit in that checkout was blocked. sp-745f5962 is the carry; sp-1ad08728 is the class (an instance ahead of the skeleton on a synced path).

## The one thing the instance never needed

Its `_route_context` imports `pipeline` unguarded. That is fine there and an uncaught ImportError inside a Stop hook on the 24 instances without `q-consult/pipeline`, and a Stop hook that exits 1 fails OPEN. A straight copy of the five defs would have shipped that. Here: absent lane = `None`, silent; lane directory present but not importable = `RouteBoundaryError`, holds the turn; present lane = the instance's code path unchanged. The consulting session's own words on it: "the difference between a port and an outage."

## Verification (ran, not assumed)

- `test_voice_stop_gate_founder_text.py` (new, 17): the instance's arbiter cases byte for byte, plus peer-message rejection, command-body truncation, and the lane triple (absent / broken / present) on a fake `q-consult/pipeline` package in a temp root.
- The four stop-gate suites in the skeleton together: 67 pass, 1 skip.
- The instance's own two arbiter test files re-pointed at THIS file with its REAL route lane and classifier: 13 pass.
- `prompt-only-enforcement-guard.py` on the file: exit 0. Parameter-rebind check: 1 finding, equal to main.
- Capability manifest entry added for the new test file.

## Relationship to PR #295

#295 is the deeper route-receipt rewrite (15 review rounds, one open major on the refusal loop). It merges clean onto main but does not land today; it rebases onto this. This PR changes nothing #295 does not also carry, and carries nothing #295 would remove.

## After merge

The next consulting sync makes the two copies identical; the instance's `test_voice_stop_gate_propagation.py` (strict xfail on the drift) goes XPASS and its owner deletes the marker.

https://claude.ai/code/session_01DCaQcw6jqEpYyqyAznCHg9

### DIFF
--- a/q-system/.q-system/capability/expected_tests/q-system__.q-system__tests__test_voice_stop_gate_founder_text.py.json
+++ b/q-system/.q-system/capability/expected_tests/q-system__.q-system__tests__test_voice_stop_gate_founder_text.py.json
@@ -1,0 +1,5 @@
+{
+ "path": "q-system/.q-system/tests/test_voice_stop_gate_founder_text.py",
+ "runner": "python3",
+ "timeout_s": 60
+}
--- a/q-system/.q-system/capability/expected_tests/q-system__.q-system__tests__test_voice_stop_gate_founder_text.py.json
+++ b/q-system/.q-system/capability/expected_tests/q-system__.q-system__tests__test_voice_stop_gate_founder_text.py.json
@@ -1,5 +1,5 @@
 {
- "path": "q-system/.q-system/tests/test_voice_stop_gate_founder_text.py",
- "runner": "python3",
- "timeout_s": 60
+  "path": "q-system/.q-system/tests/test_voice_stop_gate_founder_text.py",
+  "runner": "pytest",
+  "timeout_s": 120
 }
--- a/q-system/.q-system/scripts/voice-stop-gate.py
+++ b/q-system/.q-system/scripts/voice-stop-gate.py
@@ -642,6 +642,46 @@ def find_final_user_text(transcript_path):
     return text
 
 
+def request_is_current(transcript_path):
+    """True when the newest text put in front of the assistant was typed by HIM.
+
+    `find_final_user_text` walks back to his last words on purpose (a skill body
+    or a hook's feedback is never his request). The walk-back has a second edge
+    (PR #313 review, round 3): after a routed draft, a machine turn (a task
+    notification, a peer message, a command's stdout) made the assistant answer
+    the MACHINE, and the gate re-ran his previous request through the receipt
+    check, refusing every such reply until he typed again. Measured there: 579
+    of 837 user text records in 60 session logs are machine turns answered by
+    the assistant. A non-draft reply to machine text is not a completion of his
+    request, so the short path asks this first.
+
+    Shape rules, from the records the harness writes: a tool result carries no
+    text and is neither his nor a trigger; a flagged record right after his own
+    (a skill body) is a companion of his turn; a flagged record after the
+    assistant spoke (a Stop hook's feedback) is the machine's; unflagged machine
+    text (a notification, a peer message) is the machine's.
+    """
+    current = False
+    assistant_spoke = False
+    for record, message in _walk_transcript(transcript_path, want_record=True):
+        role = message.get("role")
+        if role == "assistant":
+            assistant_spoke = True
+            continue
+        if role != "user":
+            continue
+        text = _message_text(message)
+        if not text:
+            continue
+        if any(record.get(flag) is True for flag in _META_FLAGS):
+            if assistant_spoke:
+                current = False
+            continue
+        current = bool(founder_typed_text(text))
+        assistant_spoke = False
+    return current
+
+
 # A MISSING CHECK IS NOT A PASS. This returned (0, "") when its script was
 # absent, and 0 is the same value a clean draft produces, so a run on a machine
 # where voice-lint.py had moved was byte-for-byte indistinguishable from a run
@@ -978,11 +1018,18 @@ def main():
         # returned here -- so the ONE turn shape he uses most was the one shape
         # that never reached the scorer. The lint's scope is unchanged; the
         # measurement no longer rides on it.
-        try:
-            enforce_route_receipt(request, text)
-        except RouteBoundaryError as exc:
-            sys.stderr.write(f"voice-stop-gate: {exc}\n")
-            sys.exit(2)
+        #
+        # Only when HE just asked. A non-draft reply to machine text (a
+        # notification, a peer message, a hook's feedback) is an answer to the
+        # machine, not a completion of his routed request; see
+        # `request_is_current`. The framed-draft path below keeps its check
+        # whatever triggered it: a draft of a routed request needs its receipt.
+        if request_is_current(transcript_path):
+            try:
+                enforce_route_receipt(request, text)
+            except RouteBoundaryError as exc:
+                sys.stderr.write(f"voice-stop-gate: {exc}\n")
+                sys.exit(2)
         authorship_spool(extract_setoff_draft(text), text, request)
         finish_ok()
     # WHICH RULEBOOK. An instance with no registry resolves to None here and the two
--- a/q-system/.q-system/scripts/voice-stop-gate.py
+++ b/q-system/.q-system/scripts/voice-stop-gate.py
@@ -45,6 +45,117 @@ INSTANCE_ROOT = SCRIPTS_DIR.parents[2]
 
 MIN_TEXT_BYTES = 80
 
+
+# THE ROUTE-RECEIPT FAMILY, PORTED FROM THE ONE INSTANCE THAT RUNS IT (ASK-1197).
+#
+# consulting's copy of this file has carried these five defs since 2026-08, and
+# every fleet sync replaced them with a skeleton copy that had none, so that
+# checkout's pre-commit went red after every sync and the restore was undone by
+# the next one (2026-09-06, twice in one afternoon; sp-745f5962, sp-1ad08728).
+# What is ported is the behaviour the founder runs live, verbatim where it can
+# be, so the sync stops regressing it; the deeper rewrite (PR #295, 15 review
+# rounds, one open major) rebases onto this rather than blocking it.
+#
+# THE 24 INSTANCES WITHOUT A ROUTE LANE ARE SILENT. consulting's original
+# imported `pipeline` unguarded, which is fine there and an uncaught ImportError
+# everywhere else (a Stop hook that exits 1 fails OPEN). So: no `q-consult/
+# pipeline` directory means no lane, `_route_context` returns None and
+# `enforce_route_receipt` returns None. A lane directory that is present and
+# does not import is a BROKEN verifier, not an absent one, and holds the turn
+# (a-gate-that-cannot-run-must-not-pass). Three outcomes, on purpose.
+ROUTE_PIPELINE_REL = Path("q-consult") / "pipeline"
+
+
+class RouteBoundaryError(ValueError):
+    """A routed completion lacks a current shared route receipt."""
+
+
+def _route_context():
+    """The q-consult route modules, or None when this instance has no route lane."""
+    pipeline_dir = INSTANCE_ROOT / ROUTE_PIPELINE_REL
+    if not pipeline_dir.is_dir():
+        return None
+    sys.path.insert(0, str(INSTANCE_ROOT / "q-consult"))
+    try:
+        from pipeline import (audit_only_routes, route_classifier, route_contract,
+                              route_registry)
+    except Exception as exc:
+        raise RouteBoundaryError(
+            f"route lane is installed at {pipeline_dir} but did not import: {exc}"
+        ) from exc
+    return route_classifier, route_contract, audit_only_routes, route_registry
+
+
+def _route_draft(text):
+    marker = "=== DRAFT ==="
+    if marker not in text:
+        return extract_publishable(text)
+    return text.split(marker, 1)[1].strip()
+
+
+def _receipt_block(text):
+    marker = "=== ROUTE RECEIPT ==="
+    if marker not in text:
+        return None
+    payload = text.split(marker, 1)[1].lstrip()
+    try:
+        value, _end = json.JSONDecoder().raw_decode(payload)
+    except json.JSONDecodeError as exc:
+        raise RouteBoundaryError("route receipt block is not valid JSON") from exc
+    return value
+
+
+def enforce_route_receipt(request, assistant_text):
+    """Consume the producer receipt for a routed completion, or refuse it.
+
+    None when this instance has no route lane, or the request is not routed.
+    Raises RouteBoundaryError to hold the turn.
+    """
+    context = _route_context()
+    if context is None:
+        return None
+    classifier, contract, audit_only, registry = context
+    result = classifier.classify(request)
+    if result.status == classifier.NOT_ROUTED:
+        return None
+    if result.status != classifier.ROUTE:
+        raise RouteBoundaryError(
+            f"route request is {result.status}: {result.reason}")
+    try:
+        route = registry.resolve(result.surface, result.channel)
+    except registry.RouteRegistryError:
+        audit_routes = [candidate for candidate in audit_only.routes()
+                        if candidate.surface == result.surface and candidate.channel == result.channel]
+        if len(audit_routes) == 1:
+            try:
+                audit_only.deny(audit_routes[0])
+            except audit_only.AuditOnlyRouteError as exc:
+                raise RouteBoundaryError(str(exc)) from exc
+        raise RouteBoundaryError("route has no single registered active owner")
+    receipt = _receipt_block(assistant_text)
+    if not isinstance(receipt, dict):
+        raise RouteBoundaryError("routed completion has no route receipt")
+    # The identity is whatever the store matches on, read from the store, so
+    # a field added there (R9: loop_sha) is demanded here without a second
+    # hand-kept list.
+    required = set(contract.route_receipts.MATCH_FIELDS)
+    if not required <= receipt.keys():
+        raise RouteBoundaryError("route receipt identity is incomplete")
+    if (receipt["surface"], receipt["channel"]) != (result.surface, result.channel):
+        raise RouteBoundaryError("route receipt does not match the requested surface")
+    if contract.request_hash(request, result.surface, result.channel) != receipt["request_hash"]:
+        raise RouteBoundaryError("route receipt does not match the user request")
+    draft = _route_draft(assistant_text)
+    if contract.output_hash(draft, result.surface, result.channel) != receipt["output_hash"]:
+        raise RouteBoundaryError("route receipt does not match the assistant output")
+    identity = {key: receipt[key] for key in required}
+    try:
+        # R9: the contract recomputes the receipt's loop evidence against THIS
+        # draft and the corpus on disk before the row is consumed.
+        return contract.verify_and_consume(identity, draft=draft)
+    except Exception as exc:
+        raise RouteBoundaryError(f"route receipt was not accepted: {exc}") from exc
+
 # Explicit publish-intent framing — the only signal that a final chat message hands the
 # founder content meant for someone ELSE. Engineering/debug chat carries none of these,
 # so it's treated as conversational-to-founder and skipped (voice-enforcement.md).
@@ -321,7 +432,31 @@ def finish_ok():
     sys.exit(0)
 
 
-def _walk_transcript(transcript_path):
+# A record the HARNESS injected, not something a person typed. These are
+# top-level fields on the transcript record, and `_walk_transcript` used to
+# throw them away by yielding only `record["message"]` -- which is precisely why
+# the text filters below kept having to guess from prose.
+_META_FLAGS = ("isMeta", "turnCompanion")
+
+
+def _walk_transcript(transcript_path, want_record=False):
+    """Yield each message, or `(record, message)` when the caller needs the flags.
+
+    THE SCAR (2026-09-01, third occurrence of this deadlock). A skill loaded and
+    the harness wrote that skill's 16,584-character body as its OWN `user`
+    record, carrying `isMeta: true` and `turnCompanion: true` and NO command tags
+    at all. This walker dropped those flags, so `find_final_user_text` read the
+    skill's documentation as the founder's request. Two of that document's own
+    sentences classify UNSUPPORTED, and the gate then refused every completion in
+    the session -- including the turn reporting the block -- while his actual
+    message, "Explain this simply no tables", is correctly not-routed.
+
+    The two earlier rounds were fixed by enumerating carriers in a regex
+    (`_INJECTED_BLOCK`, then `cross-session-message`). That is the shape that
+    keeps failing: each round adds the carrier that just bit, and the next one is
+    invisible until it bites. The harness already labels these records. Read the
+    label.
+    """
     if not transcript_path or not Path(transcript_path).exists():
         return
     for line in Path(transcript_path).read_text(encoding="utf-8").splitlines():
@@ -331,7 +466,7 @@ def _walk_transcript(transcript_path):
             continue
         message = record.get("message", {})
         if isinstance(message, dict):
-            yield message
+            yield (record, message) if want_record else message
 
 
 def _message_text(message):
@@ -357,6 +492,83 @@ def find_final_assistant_text(transcript_path):
     return "\n\n".join(p for p in text_parts if p)
 
 
+# A `user` turn the FOUNDER did not type. The harness injects background-task
+# results, hook output and context reminders on the user role, so the newest
+# `user` message is often a machine's prose. 2026-08-31: a subagent's completion
+# report contained the token `reply-lane`, `_UNSUPPORTED_COMPOUND` matched it,
+# and `enforce_route_receipt` refused the turn with "surface format has no
+# registered generation route" -- a writing request nobody made, from a report
+# about the work being finished. Stripping the blocks is not enough: a
+# notification whose text sits OUTSIDE any tag would still read as his words,
+# so a turn that is entirely machine-injected is skipped whole.
+_INJECTED_BLOCK = re.compile(
+    # `cross-session-message` added 2026-08-31 (sp-23053db5): a peer Claude
+    # session's message arrives in a `user` turn, so without it this enumeration
+    # answered "the founder typed this" about another agent's words. Its prose
+    # classified AMBIGUOUS and deadlocked every turn in the session, including the
+    # one reporting the block. Adding it removes no coverage of his own text.
+    r"<(system-reminder|task-notification|command-name|command-message|"
+    r"command-args|local-command-stdout|function_results|user-prompt-submit-hook|"
+    r"cross-session-message)"
+    r"\b.*?</\1>",
+    re.I | re.S,
+)
+_INJECTED_OPENER = re.compile(
+    r"^\s*(?:<(?:system-reminder|task-notification|command-name|"
+    r"local-command-stdout|user-prompt-submit-hook|cross-session-message)\b"
+    r"|\[SYSTEM NOTIFICATION"
+    r"|(?:PreToolUse|PostToolUse|Stop|SessionStart|UserPromptSubmit)"
+    r"(?::\w+)?\s+hook\b)",
+    re.I,
+)
+# A SKILL or SLASH-COMMAND invocation injects its whole BODY as bare markdown
+# after these tags. The body is not wrapped in anything, so `_INJECTED_BLOCK`
+# removed the little tags and left thousands of words of documentation standing
+# as "what the founder typed".
+#
+# THE SCAR, and this is its THIRD occurrence (2026-09-01). The deterministic
+# blocker for it is `_META_FLAGS` plus the `want_record=True` walk above, held
+# by the executable tests in
+# q-system/.q-system/tests/test_voice_stop_gate_founder_text.py; the prose here
+# is the history, not the enforcement. The comment on `_INJECTED_BLOCK` above
+# records the second occurrence: a peer session's prose classified AMBIGUOUS and
+# deadlocked every turn in the session, including the one reporting the refusal.
+# This is the same failure with a skill body as the source. `/workflow-authoring`
+# loaded, and two of its own sentences, "compose novel harnesses when the task
+# calls for it" and "Write/Edit and re-invoke Workflow with scriptPath", classify
+# UNSUPPORTED, so every completion was held while the founder's actual message,
+# "Explain this simply no tables", is correctly not-routed.
+#
+# His typed words come FIRST and the injection follows, so the fix is a
+# truncation, not another tag to enumerate. Enumerating tags is what failed
+# twice: each round adds the one carrier that just bit, and the next carrier is
+# invisible until it does.
+_COMMAND_INJECTION_MARK = re.compile(
+    r"<(?:command-name|command-message|command-args|skill-format|"
+    r"local-command-stdout)\b",
+    re.I,
+)
+
+
+def founder_typed_text(candidate):
+    """Strip machine-injected prose from one `user` turn, or reject it entirely.
+
+    A command or skill invocation is a TRUNCATION, not a tag removal: his typed
+    words come first and the injected body follows unwrapped, so everything from
+    the first command marker onward is the machine's text. See the scar note on
+    `_COMMAND_INJECTION_MARK`.
+    """
+    if not candidate:
+        return ""
+    if _INJECTED_OPENER.search(candidate):
+        return ""
+    typed = candidate
+    mark = _COMMAND_INJECTION_MARK.search(typed)
+    if mark:
+        typed = typed[:mark.start()]
+    return _INJECTED_BLOCK.sub(" ", typed).strip()
+
+
 def find_final_user_text(transcript_path):
     """His last message. The trigger signal the model cannot get wrong.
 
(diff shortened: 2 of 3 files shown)


assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-313-20260906-155749.md` (6934 bytes).

```
FINDINGS:
major|founder_typed_text truncates a slash-command turn to empty, so find_final_user_text returns a stale earlier message as the founder's request|q-system/.q-system/scripts/voice-stop-gate.py:553
major|New capability entry declares runner python3 on a pytest-shaped file, so the repo's own executes_nothing detector errors and the test never runs in CI|q-system/.q-system/capability/expected_tests/q-system__.q-system__tests__test_voice_stop_gate_founder_text.py.json:3
minor|All five receipt-verification branches in enforce_route_receipt survive deletion with the new suite green; the fake lane's route_contract.py is empty|q-system/.q-system/tests/test_voice_stop_gate_founder_text.py:167
minor|_INJECTED_OPENER discards a founder message that opens by naming a hook event, producing the same stale-request fallback|q-system/.q-system/scripts/voice-stop-gate.py:516
END FINDINGS
```

--- reviewer output ---

Review done. Four findings, all reproduced.

## Finding 1 — major

**`founder_typed_text` erases a slash-command turn entirely, so the gate reads a stale earlier message as "his last request".**

`q-system/.q-system/scripts/voice-stop-gate.py:553` (truncation at `_COMMAND_INJECTION_MARK`, line 546)

The comment says "his typed words come FIRST and the injected body follows". Against the real producer that is false for a slash command: the record starts with `<command-name>`, and his instruction lives in `<command-args>`. Truncating at the first marker returns `""`, so `find_final_user_text` keeps the last *earlier* non-empty message.

Reproducer used a real record lifted from `~/.claude/projects/-Users-assafkipnis-projects-consulting/7e2a26ae-....jsonl`, not an invented fixture:

```
REAL RECORD flags: isMeta=None turnCompanion=None
REAL RECORD text:
<command-name>/goal</command-name>
            <command-message>goal</command-message>
            <command-args>finish the work autonomously. sana is the human</command-args>
------------------------------------------------------------
founder_typed_text(command turn) = ''
find_final_user_text             = 'write me a linkedin post about the audit'
------------------------------------------------------------
FAIL: gate reports the PREVIOUS turn as his last request
```

Same transcript against `origin/main`:

```
MAIN find_final_user_text = '<command-name>/goal</command-name>\n            <command-message>goal</command-message>\n            <command-args>finish the work autonomously. sana is the human</command-args>'
```

Main kept his args. This PR drops them. Blast radius: `request` is the input to `authorship_spool` (line 320 docstring: "the reliable half of the trigger") and to `classifier.classify(request)` plus `contract.request_hash(request, ...)`. On a `/goal` turn the scorer is handed an unrelated request, and on the one instance with a lane the request hash is computed over words the founder did not type this turn. Nine `/command` turns in that one consulting session, none flagged `isMeta`.

## Finding 2 — major

**The new capability entry declares `runner: python3` on a pytest-shaped file, which the repo's own gate refuses and CI runs.**

`q-system/.q-system/capability/expected_tests/q-system__.q-system__tests__test_voice_stop_gate_founder_text.py.json:3`

```
entry: {'path': 'q-system/.q-system/tests/test_voice_stop_gate_founder_text.py', 'runner': 'python3', 'timeout_s': 60}
file exists: True
executes_nothing -> True
```

`capability-gate.py:695` appends `zero-execution test: ... Declare runner pytest ... (ASK-1145)` and `continue`s, so the test is never executed by the gate. `.github/workflows/validate.yml:52` runs that gate. Every other pytest entry in `expected_tests/` declares `"runner": "pytest"`. This is the exact ASK-1145 scar the detector was written for, recurring in the same change that adds the coverage.

## Finding 3 — minor

**Every receipt-verification branch in `enforce_route_receipt` can be deleted and the new suite stays green.**

`q-system/.q-system/tests/test_voice_stop_gate_founder_text.py:167` (the fake lane writes an empty `route_contract.py`)

Mutation run on a copy in `/tmp/pr313mut`, five branches replaced with `if False:` (`isinstance(receipt, dict)`, `required <= receipt.keys()`, the surface/channel match, `request_hash`, `output_hash`):

```
.................                                                        [100%]
17 passed in 0.02s
```

The fake lane only reaches `NOT_ROUTED` and the registry-raises path. `contract.route_receipts.MATCH_FIELDS`, `request_hash`, `output_hash` and `verify_and_consume` at lines 141-163 have zero coverage in this repo. `grep enforce_route_receipt` finds no other caller or test. The PR's "13 pass" evidence is an out-of-repo suite that CI never runs.

## Finding 4 — minor

**`_INJECTED_OPENER` discards a founder message that opens by naming a hook event.**

`q-system/.q-system/scripts/voice-stop-gate.py:516`

```
'PostToolUse hook blocked my edit again, why?'          -> ''
'Stop hook feedback is firing on every turn, fix it'    -> ''
'SessionStart hook output looks wrong'                  -> ''
'write me a linkedin post'                              -> 'write me a linkedin post'
```

Same stale-request consequence as finding 1, narrower input. This founder pastes hook errors as a documented workflow.

## What is sound

Attacks that did not land:

- **Fixture shape on the meta flags.** I checked two real transcripts. `isMeta` (150 and 164 occurrences) and `turnCompanion` (2 each) are top-level record keys exactly as the fixture builds them. The author's claim survives its own producer.
- **The three-outcome lane guard.** Absent lane returns `None`, broken lane raises. That is the right call: an unguarded `pipeline` import in a Stop hook fails open on 24 instances. The tests exercise all three.
- **`_walk_transcript` signature change.** `find_final_assistant_text` uses the default and `test_assistant_text_is_unaffected` covers it.
- **The suite claim.** `67 passed, 1 skipped` across the four stop-gate suites, matching the PR body exactly.
- **Negative controls exist.** `test_an_unflagged_user_turn_is_still_read` and `test_flags_are_read_from_the_record_not_the_message` both go red under plausible mutations; they are not decoration.
- **Double-consume.** Both `enforce_route_receipt` call sites are followed by an exit, so `verify_and_consume` cannot run twice in one process.

## VERDICT: REQUEST CHANGES

Not BLOCK. Nothing here is unrecoverable, and finding 1 is a regression on a path that already had a different wrong answer on main.

**Fix first: finding 1.** The truncation needs to keep `<command-args>` content rather than cutting at the first marker, or `find_final_user_text` needs to return `""` instead of falling back to an older turn. Silently answering with the previous turn's request is worse than answering with nothing, because both the scorer and the route hasher treat it as authoritative.

FINDINGS:
major|founder_typed_text truncates a slash-command turn to empty, so find_final_user_text returns a stale earlier message as the founder's request|q-system/.q-system/scripts/voice-stop-gate.py:553
major|New capability entry declares runner python3 on a pytest-shaped file, so the repo's own executes_nothing detector errors and the test never runs in CI|q-system/.q-system/capability/expected_tests/q-system__.q-system__tests__test_voice_stop_gate_founder_text.py.json:3
minor|All five receipt-verification branches in enforce_route_receipt survive deletion with the new suite green; the fake lane's route_contract.py is empty|q-system/.q-system/tests/test_voice_stop_gate_founder_text.py:167
minor|_INJECTED_OPENER discards a founder message that opens by naming a hook event, producing the same stale-request fallback|q-system/.q-system/scripts/voice-stop-gate.py:516
END FINDINGS

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-313-20260906-163958.md` (5711 bytes).

```
FINDINGS:
major|Any lane exception other than the import escapes uncaught, the Stop hook exits 1, and the routed turn completes with no receipt consumed|q-system/.q-system/scripts/voice-stop-gate.py:118
minor|Deleting both enforce_route_receipt call sites in main() leaves the whole stop-gate suite green, so the wiring is unpinned against the sync regression this PR exists to stop|q-system/.q-system/scripts/voice-stop-gate.py:964
minor|The audit_only refusal branch is never reached by the fixture and survives full deletion with the suite green, contradicting the docstring's mutation-checked claim|q-system/.q-system/scripts/voice-stop-gate.py:127
END FINDINGS
```

--- reviewer output ---

# Review: PR #313 (voice-stop-gate route-receipt + founder-text port)

## Finding 1 (major): the "broken lane holds the turn" contract covers only the import line. Every other lane failure fails OPEN.

`q-system/.q-system/scripts/voice-stop-gate.py:118`

The header comment at lines 57 to 70 states three outcomes on purpose: absent lane returns None, a lane that does not import raises `RouteBoundaryError` and holds the turn, a present lane runs the instance path. Only the `from pipeline import ...` statement is wrapped. `classifier.classify()`, `registry.resolve()` raising anything that is not `RouteRegistryError`, `contract.request_hash()`, and `contract.verify_and_consume` reached outside its own try all escape as bare exceptions. `main()` catches `RouteBoundaryError` only, so the process exits 1. A Stop hook exit of 1 is a non-blocking error: the turn completes and no receipt is consumed. That is the exact `a-gate-that-cannot-run-must-not-pass` case the comment cites.

Reproducer (real script, real stdin payload, temp instance root with a lane that imports cleanly and raises inside `classify`):

```
python3 -c '<staged temp root + fake lane + transcript, invoked the real gate>'
```

Real output:

```
EXIT CODE: 1
STDERR tail:
    result = classifier.classify(request)
  File ".../vsg313_root/q-consult/pipeline/route_classifier.py", line 4, in classify
    raise RuntimeError("classifier corpus is missing")
RuntimeError: classifier corpus is missing
```

Exit 2 is the block code. This path returns 1.

Caveat stated plainly: no lane exists in this repo, so my input is constructed. So is every lane input in the PR's own green tests, including `test_a_lane_that_does_not_import_holds_the_turn`, so the caveat does not separate my repro from theirs.

Fix shape: wrap the body of `enforce_route_receipt` after `_route_context()` so any non-`RouteBoundaryError` exception becomes a `RouteBoundaryError`.

## Finding 2 (minor): nothing pins the two `main()` call sites.

`q-system/.q-system/scripts/voice-stop-gate.py:964` and `:1072`

Delete both call sites and the whole stop-gate suite stays green. The suite proves the function refuses correctly; nothing proves the gate calls it. This is the regression class the PR exists to stop (sp-1ad08728, a sync silently removing the enforcement), and a future skeleton edit or the PR #295 rebase could drop the wiring with a clean run.

Reproducer, run against a temp mirror of the scripts and tests:

```
M1 both enforce_route_receipt call sites in main() deleted
    72 passed, 2 skipped in 0.63s
```

Control, same harness, two mutants that should die and do:

```
M3 _receipt_block JSONDecodeError -> return None -> 1 failed, 71 passed, 2 skipped
M5 request_hash comparison deleted        -> 1 failed, 71 passed, 2 skipped
```

Kept at minor: nothing misbehaves at 3am from this merge. It is a hole in the claim, not a live defect.

## Finding 3 (minor): the audit-only refusal branch has zero coverage.

`q-system/.q-system/scripts/voice-stop-gate.py:127`

The fake lane's `audit_only.routes()` returns `[]`, so `len(audit_routes) == 1` is never true and `audit_only.deny` is never called. The test docstring says each verification branch "is killed by at least one case (mutation-checked)".

```
M4 whole audit_only branch deleted -> 72 passed, 2 skipped in 0.64s
```

## What is sound

- **Producer validation of the founder-text half.** I read 40 real session transcripts from `~/.claude/projects` (outside this worktree, read only, to check the PR's producer claims rather than its fixtures). Zero unflagged founder turns were dropped by `founder_typed_text`. All 18 rewrites were correct: `/goal` turns reduced to their `<command-args>`, bare commands to the command name. 29 command turns, none flagged `isMeta`, which is the PR's stated measurement and it holds.
- **`_INJECTED_OPENER` deletion survives the suite, and I dropped it as a finding.** On those same 40 transcripts the guard changed the result zero times out of 194 opener matches, because `_INJECTED_BLOCK` already handles all of them. Harmless redundancy, not worth a round.
- **The absent-lane path.** Ran the gate end to end with no `q-consult/pipeline`: no exception, no exit 2, the 24 silent instances are genuinely silent.
- **Capability wiring is real.** `trigger-inventory.py:123` globs the `expected_tests` fragment directory, and `manifest.json` is fragments-only, so the new entry is loaded rather than decorative.
- **`_walk_transcript` signature change.** `find_final_assistant_text` still resolves; mutating the receipt-JSON and request-hash checks both go red, so those branches are not decoration.

## VERDICT: REQUEST CHANGES

Not a block. Nothing here is permanent or unrecoverable, and the founder-text half is the best-evidenced part of the change.

Fix first: wrap the body of `enforce_route_receipt` so a lane that throws anywhere, not only at its import, holds the turn instead of exiting 1 and letting a routed completion ship unreceipted.

FINDINGS:
major|Any lane exception other than the import escapes uncaught, the Stop hook exits 1, and the routed turn completes with no receipt consumed|q-system/.q-system/scripts/voice-stop-gate.py:118
minor|Deleting both enforce_route_receipt call sites in main() leaves the whole stop-gate suite green, so the wiring is unpinned against the sync regression this PR exists to stop|q-system/.q-system/scripts/voice-stop-gate.py:964
minor|The audit_only refusal branch is never reached by the fixture and survives full deletion with the suite green, contradicting the docstring's mutation-checked claim|q-system/.q-system/scripts/voice-stop-gate.py:127
END FINDINGS

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-313-20260906-174737.md` (4842 bytes).

```
FINDINGS:
major|A machine-injected user turn re-runs the previous founder request through the route gate, blocking every turn after a routed draft until a human types again|q-system/.q-system/scripts/voice-stop-gate.py:970
END FINDINGS
```

--- reviewer output ---

# Review: PR #313 (voice-stop-gate route-receipt port)

Worktree is clean, no repo files changed.

## Finding 1 — major

**A machine-injected `user` turn re-runs the PREVIOUS founder request through the route gate, so every turn after a routed draft is refused until the founder types again.**

`q-system/.q-system/scripts/voice-stop-gate.py:970` and `:982` / `:1090`

`find_final_user_text` now skips harness records and keeps the last **non-empty** candidate. When the newest `user` record is entirely machine text (a task-notification, a `local-command-stdout`), it falls back to the previous founder message. `main()` then hands that stale request to `enforce_route_receipt`, which classifies it as ROUTE and demands a receipt from a reply that is not a draft at all.

The exposure is not theoretical. Measured over the 60 most recent session logs on this machine (`~/.claude/projects/*/*.jsonl`, read for this measurement only):

```
user text records: 837
machine user turns answered by an assistant turn (stale-request exposure): 579
```

Reproducer, run against the PR head with a fake `q-consult/pipeline` whose classifier routes on the words of the request:

```
turn 1  routed draft + receipt                       exit=0
turn 2  machine notification, prose reply            exit=2  voice-stop-gate: routed completion has no route receipt
turn 3  Claude re-drafts (stop_hook_active=True)     exit=2  voice-stop-gate: routed completion has no route receipt
turn 4  Claude re-drafts again                       exit=2  voice-stop-gate: routed completion has no route receipt
turn 5  founder types a fresh non-routed message     exit=0
```

Three things make this the real cost rather than one bad turn. The receipt from turn 1 is already consumed, so no redraft can satisfy the check. `main()` never reads `stop_hook_active`, so the Stop-hook loop does not self-limit (that part is pre-existing, and this is the first terminal state the model cannot redraft its way out of). And the only thing that clears it is a human typing, which an unattended `claude -p` run does not have.

Blast radius today is the one instance with a `q-consult/pipeline` lane. It is recoverable and revertable, so major and not blocker.

Shape of the fix, for whoever picks it up: `enforce_route_receipt` needs to know whether the request it was handed belongs to **this** turn. Either `find_final_user_text` returns the newest founder turn plus whether it was the last user record, or `main()` skips enforcement when the newest `user` record is machine-flagged.

## What is sound

**The system-reminder opener.** `_INJECTED_OPENER` returns `""` for a turn that opens with an injected tag, and a founder turn whose prose sits after a `<system-reminder>` would be discarded. Measured over 120 transcripts: 0 non-meta user turns open with `<system-reminder>` and carry more than 200 characters of real prose after it. All 211 swallowed turns in the 60-file scan were genuine `task-notification` / `local-command-stdout` records.

**Machine prose leaking through as his words.** 0 non-meta turns whose `founder_typed_text` output still carries a machine tag.

**The author's own measured claims.** Independently reproduced: `turnCompanion` is a real top-level field (12 occurrences), and 163 of 163 hook-feedback-shaped user records carry `isMeta`. The fixture shape in `test_voice_stop_gate_founder_text.py` (flags top-level on the record, content as text blocks) matches what the harness actually writes.

**Whether the tests can fail.** I ran a 10-mutant harness against a copy of the gate. Nine were killed, including both `enforce_route_receipt` call sites deleted, the absent-lane guard removed, the `except Exception` wrapper narrowed, the `request_hash` check removed, the identity-completeness check removed, and the `_META_FLAGS` filter disabled. One survivor: dropping `cross-session-message` from `_INJECTED_OPENER`, which `_INJECTED_BLOCK` already covers for any well-formed tag. Effectively an equivalent mutant, not a hole.

**The 25 lane-less instances.** `_route_context` returns `None` on a missing directory before any import, so nothing new fires there. Both `main()` exit paths reach the receipt check, so there is no routed completion that escapes it.

Suites: `q-system/.q-system/tests/test_voice_stop_gate_founder_text.py` 35 passed; all stop-gate suites 79 passed.

## VERDICT: REQUEST CHANGES

Fix first: bind `enforce_route_receipt` to the request of the **current** turn, so a background notification cannot re-litigate an already-consumed routed draft.

FINDINGS:
major|A machine-injected user turn re-runs the previous founder request through the route gate, blocking every turn after a routed draft until a human types again|q-system/.q-system/scripts/voice-stop-gate.py:970
END FINDINGS

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-313-20260906-181722.md` (9304 bytes).

```
FINDINGS:
major|request_is_current gates the short path on who spoke last, so a routed draft following any machine-injected user record (including the gate's own exit-2 feedback) completes with no receipt consumed|q-system/.q-system/scripts/voice-stop-gate.py:1027
minor|The framed-draft call site still enforces the receipt against the same stale request the short path now skips, so identical transcripts fail open unframed and deadlock framed|q-system/.q-system/scripts/voice-stop-gate.py:1137
END FINDINGS
```

--- reviewer output ---

 This is an engineering code review for the founder, not published content, so the voice bypass marker applies (voice-enforcement.md: voice rules do not govern analysis for the founder's eyes only). Reproducer code is shown indented rather than fenced.

## Review: PR #313 — voice-stop-gate route-receipt port

Baseline first: the shipped suite is green, 43 passed, and all stop-gate suites together are 87 passed. Then three reproducers driven through `main()` against a fake `q-consult/pipeline` lane in a temp root.

### MAJOR — the last commit opens a fail-open: a routed draft that follows any machine-injected user record completes with no receipt consumed

`q-system/.q-system/scripts/voice-stop-gate.py:1027`, the line `if request_is_current(transcript_path):`

`extract_publishable()` returns empty when the assistant message carries no publish framing, so the bare-fence draft shape (the one the comment at line 1013 calls the shape he uses most) lands on the SHORT path. Commit `db5d401d` put that path's receipt check behind `request_is_current()`, which returns False whenever the newest user record is machine text. A task notification, a peer message, or the gate's own exit-2 feedback is enough.

Worse: the gate's refusal disarms the gate. Turn 1 blocks, the harness writes the refusal back as an isMeta user record (the exact shape their own `test_request_is_current_reads_who_spoke_last` asserts), and the redraft on turn 2 ships unchecked.

Reproducer, saved to a file and run from the worktree root with python3:

    import importlib.util, io, json, pathlib, sys, tempfile, textwrap
    GATE = pathlib.Path.cwd() / "q-system/.q-system/scripts/voice-stop-gate.py"
    spec = importlib.util.spec_from_file_location("vsg", GATE)
    vsg = importlib.util.module_from_spec(spec); spec.loader.exec_module(vsg)
    root = pathlib.Path(tempfile.mkdtemp()); pkg = root / "q-consult" / "pipeline"
    pkg.mkdir(parents=True)
    (pkg / "__init__.py").write_text("")
    (pkg / "route_classifier.py").write_text(textwrap.dedent('''
        import collections
        NOT_ROUTED = "NOT_ROUTED"; ROUTE = "ROUTE"
        Result = collections.namedtuple("Result", "status surface channel reason")
        def classify(request):
            return Result(ROUTE if "linkedin post" in request.lower() else NOT_ROUTED,
                          "linkedin_post", "linkedin", "fixture")
    '''))
    (pkg / "route_contract.py").write_text(textwrap.dedent('''
        class route_receipts:
            MATCH_FIELDS = set(["surface", "channel", "request_hash", "output_hash"])
        def request_hash(r, s, c): return "rh"
        def output_hash(o, s, c): return "oh"
        def verify_and_consume(identity, **kw): return dict(identity, status="ok")
    '''))
    (pkg / "audit_only_routes.py").write_text(textwrap.dedent('''
        class AuditOnlyRouteError(Exception): pass
        def routes(): return []
        def deny(r): raise AuditOnlyRouteError("audit only")
    '''))
    (pkg / "route_registry.py").write_text(textwrap.dedent('''
        class RouteRegistryError(Exception): pass
        def resolve(s, c): return dict(owner="fixture")
    '''))
    vsg.INSTANCE_ROOT = root
    vsg.authorship_spool = lambda *a, **k: None
    vsg.finish_ok = lambda: sys.exit(0)
    vsg.run_check = lambda *a, **k: (0, "")
    FENCE = chr(96) * 3
    def rec(text, role="user", **flags):
        r = dict(type=role,
                 message=dict(role=role, content=[dict(type="text", text=text)]))
        r.update(flags); return r
    POST = FENCE + "\nMy agent handed me a month-old note in a client meeting.\n" + FENCE
    REDRAFT = FENCE + "\nMy agent handed me a note from a month ago.\n" + FENCE
    def turn(records, label):
        tp = root / "t.jsonl"; tp.write_text("\n".join(json.dumps(r) for r in records))
        sys.argv = ["voice-stop-gate.py"]
        sys.stdin = io.StringIO(json.dumps(dict(transcript_path=str(tp))))
        err, real = io.StringIO(), sys.stderr; sys.stderr = err
        code = None
        try: vsg.main()
        except SystemExit as e: code = e.code
        finally: sys.stderr = real
        print(label, "-> exit", code, "|", err.getvalue().strip() or "(no stderr)")
    records = [rec("write me a linkedin post about the audit"), rec(POST, role="assistant")]
    turn(records, "turn 1 (his request, no receipt)      ")
    records += [rec("Stop hook feedback:\nvoice-stop-gate: routed completion has no "
                    "route receipt", isMeta=True),
                rec(REDRAFT, role="assistant")]
    turn(records, "turn 2 (the redraft, still no receipt)")

Real output:

    turn 1 (his request, no receipt)       -> exit 2 | voice-stop-gate: routed completion has no route receipt
    turn 2 (the redraft, still no receipt) -> exit 0 | (no stderr)

Same harness with a task-notification record in place of the hook feedback, run against the PR head and against its parent `c44934a6`, isolates the regression to `db5d401d`:

    --- PR313-head
      publishable draft bytes: 0
      request the gate read  : 'write me a linkedin post about the audit'
      request_is_current     : False
      exit code              : 0
      VERDICT: FAIL-OPEN, routed completion allowed with no receipt
    --- PR313-before-db5d401d
      request_is_current     : (absent in this revision)
      exit code              : 2
      stderr                 : voice-stop-gate: routed completion has no route receipt
      VERDICT: HELD (exit 2)

At 3am on the consulting instance this is a routed post completing with no receipt row consumed, and the trigger is a background agent finishing, which this fleet does constantly. The predicate answers "who spoke last", but the question the call site asks is "is this assistant message a draft of his request".

### MINOR — the mirror half: the framed path still re-runs the stale request, so the deadlock this PR set out to fix survives where the comment exempts it

`q-system/.q-system/scripts/voice-stop-gate.py:1137`

Identical transcript, only the framing differs. Unframed draft after the machine turn: allowed. Framed draft after the machine turn: held, and Claude cannot clear it by re-drafting because only the producer mints receipts. Same harness, assistant text "Here is the post the agent finished:" plus a fence:

    request_is_current: False
    publishable bytes : 111
    exit code         : 2 | voice-stop-gate: routed completion has no route receipt

The comment at line 1024 calls this deliberate. Taken with finding 1, the two paths make opposite decisions about the same stale request. Fail-closed is the safe direction, so this is minor, and one predicate should serve both call sites.

### What is sound

Attacks the code survived, each executed:

- The 24 silent instances. A temp root with no `q-consult/pipeline`: `_route_context()` returns None and `enforce_route_receipt()` returns None, no exception. The guard around consulting's unguarded import is the right call and it works.
- Broken lane holds. A lane whose `route_classifier` raises on import gives `RouteBoundaryError`, not an uncaught ImportError, so the Stop hook does not exit 1 and fail open.
- Lane that raises at call time. The `except Exception` conversion is real, not a comment: a classifier raising RuntimeError becomes a hold.
- Both `main()` call sites are pinned. Deleting either one goes red in their suite. Confirmed by driving `main()` on both paths myself.
- `_walk_transcript`'s second caller. `find_final_assistant_text` is unaffected by the `want_record` parameter across all three repros.
- Flags read at record level, not inside `message`. Verified, and their negative control (an unflagged user turn is still read) is a real control, not decoration.
- Third transcript pass cost. Suspected a per-turn I/O tax from `request_is_current`. Measured on a 36.2MB, 40,000-record transcript: 0.11s, the same as each existing pass. Not a finding, dropped.
- The `FAKE_CONTRACT` fixture against the real producer. Tried to diff it against consulting's `route_contract.py`; reading outside the review worktree is blocked in this session, so the "hashes with the producer's exact envelope" claim is unverified rather than disproven. Flagged as unchecked, not as a defect.

### VERDICT: REQUEST CHANGES

Not a BLOCK: nothing here is unrecoverable, and the lane exists on one instance.

Fix first: finding 1. The short-path condition should ask whether the assistant message is a draft of his outstanding request, not whether he spoke most recently, and whatever predicate lands there belongs on line 1137 too. Reverting `db5d401d` alone restores fail-closed on both paths and reopens the reply-to-machine deadlock, so the predicate needs the rewrite rather than the revert.

FINDINGS:
major|request_is_current gates the short path on who spoke last, so a routed draft following any machine-injected user record (including the gate's own exit-2 feedback) completes with no receipt consumed|q-system/.q-system/scripts/voice-stop-gate.py:1027
minor|The framed-draft call site still enforces the receipt against the same stale request the short path now skips, so identical transcripts fail open unframed and deadlock framed|q-system/.q-system/scripts/voice-stop-gate.py:1137
END FINDINGS

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-313-20260906-183355.md` (7077 bytes).

```
FINDINGS:
major|A machine-written compaction summary is unflagged, read as the founder's request, and refuses the first routed completion after every compaction|q-system/.q-system/scripts/voice-stop-gate.py:625
minor|audit_only.deny() that returns instead of raising falls through to a refusal naming the wrong reason|q-system/.q-system/scripts/voice-stop-gate.py:141
nit|_route_context inserts into sys.path on every call and never unwinds it, including on the error path|q-system/.q-system/scripts/voice-stop-gate.py:76
END FINDINGS
```

--- reviewer output ---

## PR #313 review

Read-only. Nothing in the repo was modified. Every finding below was executed.

---

### major | A compaction summary is read as the founder's request, and this PR is what makes that block a turn

**File:** `q-system/.q-system/scripts/voice-stop-gate.py:625` (`find_final_user_text`), `:653` (`request_is_current`)

The new `_META_FLAGS` filter rests on a claim written into the code at `:453`: *"The harness already labels these records. Read the label."* There is a carrier the harness does not label.

When Claude Code compacts, it writes a `user` record whose text is the machine-authored summary. It carries `isMeta: None` and `turnCompanion: None`. `founder_typed_text` returns it whole.

**Reproducer 1** (real records, real logs, 60 most recent session files):

```
python3 - <<'PYEOF'
import json, os, glob, importlib.util
spec = importlib.util.spec_from_file_location("vsg", os.path.abspath("q-system/.q-system/scripts/voice-stop-gate.py"))
vsg = importlib.util.module_from_spec(spec); spec.loader.exec_module(vsg)
# ... walks ~/.claude/projects/*/*.jsonl, counts user records starting with the
# compaction marker, and asks founder_typed_text() what it keeps
PYEOF
```

Real output:

```
marker counts (all user records): {'compaction': 8}
ACCEPTED AS THE FOUNDER TEXT: {'compaction': 8}
--- compaction 2157c268-96a1-42b4-8532-e03f197dd1a8.jsonl chars kept: 13783
This session is being continued from a previous conversation that ran out of context...
```

8 of 8. None flagged. 13,783 characters of machine prose accepted as his words.

**Reproducer 2**, that record spliced into a transcript with his real request:

```
real compaction record from: 2157c268-96a1-42b4-8532-e03f197dd1a8.jsonl
  isMeta = None turnCompanion = None
  chars  = 13783

find_final_user_text() -> 'This session is being continued from a previous conversation...' ... total 13783 chars
  equals his request?    False
request_is_current()  -> True
```

**Reproducer 3**, driven through `main()` with the PR's own `_fake_lane` helper set to route (the lane's job) and `owner=True`:

```
TURN 1  (his request, valid receipt):        exit 0
TURN 2  (identical work, after compaction):  exit 2
        stderr: voice-stop-gate: route receipt does not match the user request
```

Same request, same draft, same correctly minted receipt. The only difference is a real compaction record in between.

At 3am on the instance that has the lane: the first routed completion after every compaction is refused, and the message blames a receipt mismatch for work that was done right. `request_is_current` returns True on the compaction record, so the short path enforces too. Recovery exists (he types again, or a prose reply clears it via the isMeta hook-feedback record), so this is not permanent. It is wrong behaviour unattended that a human has to clear.

Worth naming precisely: the wrong-request read is older than this PR. What this PR adds is `enforce_route_receipt`, the consumer that turns a wrong request into `sys.exit(2)`.

---

### minor | `audit_only.deny()` is called for its exception and its return value is dropped

**File:** `q-system/.q-system/scripts/voice-stop-gate.py:141-149`

If `deny()` ever returns instead of raising, control falls straight through to `raise RouteBoundaryError("route has no single registered active owner")`. The refusal then names the wrong reason: the route resolved to an audit-only rule that permitted it, not a missing owner.

Reproducer, the PR's own fixture with one line changed so `deny()` returns `None`:

```
deny() permitted the route; the gate still refuses, and says:
    route has no single registered active owner
```

Calibration note, stated plainly: the shipped fixture's `deny()` always raises, and I could not read the real `audit_only_routes` producer from this tree. I constructed the permitting `deny()`. Dropped a level for that.

---

### nit | `_route_context` grows `sys.path` on every call and never unwinds it

**File:** `q-system/.q-system/scripts/voice-stop-gate.py:76`

The insert happens before the import is attempted and is never popped, including on the `RouteBoundaryError` path.

```
sys.path growth check:
   sys.path entries added by 3 calls: 3
```

Harmless in a one-shot Stop hook. Noted for the day something imports this module.

---

### What is sound

Attacks I ran that the code survived, each named:

- **`_INJECTED_OPENER` swallowing his real messages.** Ran `founder_typed_text` over 447 unflagged real user records and looked for any where it returns `""` while more than 60 characters of untagged prose remain. Result: **0**. The opener list does not eat his words.
- **Subagent transcripts leaking in as the founder's request.** Checked all 60 logs for `isSidechain: true` user records. Result: **0 files, 0 records**. Sidechains are not in these transcripts, so `_walk_transcript` cannot pick up a Task prompt.
- **`reply_carries_a_draft` crying wolf.** Measured on 2,025 real assistant messages: **3.8%** carry a set-off block. It only widens the check when a machine spoke last, and the fail-closed choice in round 4 is defensible at that rate.
- **The suite is not decoration.** `test_voice_stop_gate_founder_text.py` alone: 49 passed. All four stop-gate suites: 93 passed. `test_flags_are_read_from_the_record_not_the_message` and `test_an_unflagged_user_turn_is_still_read` are genuine negative controls, and the `main()`-driven cases would go red if either `enforce_route_receipt` call site were deleted.
- **No collateral callers.** `find_final_user_text` and `_walk_transcript` have no callers outside this file, so the `want_record` parameter change breaks nothing.
- **The 24-instance guard is real.** `is_dir()` before any import, `None` on absent, `RouteBoundaryError` on present-but-broken. The straight copy this PR describes would have shipped an uncaught `ImportError` in a Stop hook. This is the right shape and it is tested three ways.

---

### VERDICT: REQUEST CHANGES

Not BLOCK. Nothing here is unrecoverable; a refused turn clears when he types again.

**Fix first:** the compaction carrier. `_META_FLAGS` alone does not cover it, and the comment claiming the harness labels every injected record is falsified by 8 real records. The cheapest honest fix is a second label-shaped test on the record (compaction records carry their own marker) rather than another prose alternation in `_INJECTED_OPENER` — the file's own scar note says enumerating carriers is the shape that keeps failing, and this is the next carrier.

---

FINDINGS:
major|A machine-written compaction summary is unflagged, read as the founder's request, and refuses the first routed completion after every compaction|q-system/.q-system/scripts/voice-stop-gate.py:625
minor|audit_only.deny() that returns instead of raising falls through to a refusal naming the wrong reason|q-system/.q-system/scripts/voice-stop-gate.py:141
nit|_route_context inserts into sys.path on every call and never unwinds it, including on the error path|q-system/.q-system/scripts/voice-stop-gate.py:76
END FINDINGS

PR #314, opened 2026-09-06
fix(updater): wait on a live index.lock and give the auto-commit hook a run marker to honor (sp-523c1a25, sp-9306036e)

assafkip:
## What broke

On 2026-09-06 the consulting fleet sync took three founder-shell runs:

- 14:30: q-system committed, then the config commit died on `Unable to create '.git/index.lock': File exists` and the instance was abandoned half-delivered. The holder was a peer session's `git status`, alive for a fraction of a second. The updater's only lock handling is the unconditional stale-lock deletion at the top of the instance loop; a live writer needs a wait, not a delete (sp-523c1a25).
- 14:33: the instance's own Stop-hook auto-commit swept the updater's half-delivered rules, settings and plugins under its own generic messages and held the lock through its 445 s verify, so the founder's guarded re-run tripped and did nothing (sp-9306036e).

## What changed

`kipi-update.sh`
- `wait_for_index_lock <path> <step>`: bounded poll (`KIPI_UPDATE_LOCK_WAIT_S`, default 120 s), logs the step while waiting, refuses with the bound and the step named. Called before every index write the run makes: q-system staging, config staging, the system-state commit, the untrack commit.
- `retry_on_index_lock <path> <step> <cmd...>`: wraps the three commits (`guarded_commit` twice, the system-state commit). Retries only on the live-lock error text, cap 3; any other failure returns unchanged on attempt 1.
- Run marker `<git-common-dir>/kipi-update.run` (`<pid> <start>`) written after the stale-lock cleanup, cleared in `abandon_instance`, after `Config synced`, and in the EXIT trap.

`q-system/hooks/auto-commit.py`
- `fleet_update_in_progress()`: a live marker (pid alive) means commit nothing, one stderr line, exit 0. A dead or malformed marker is removed and the hook proceeds, so a crashed run cannot silence the safety net. Common dir, so a linked worktree of the checkout reads the same marker.

## Verification

- `q-system/.q-system/scripts/test/test-kipi-update-lock-wait.sh` (registered): retry unit case (helpers extracted from the updater's own text, floor on the extraction); a holder that takes the lock at the marker and releases only after the updater prints its waiting line, sync lands; the same holder never releasing with a 3 s bound, instance refused by name, no commit; a mutant with the wait stubbed out goes red. 4/4.
- `q-system/hooks/test/test_auto_commit_run_marker.py` (registered): live, stale, malformed, plus the no-marker control proving the live case is the marker's doing. 4/4.
- Regression: `test-kipi-update-system-state-commit.sh` 8/8 on the edited updater; the other 10 `test-kipi-update-*.sh` and the 3 updater python tests pass.

## Not in scope

sp-2c1bcc3f (the unconditional pre-run lock deletion) and sp-1ad08728 (the "instance ahead of skeleton" report) stay open; this PR does not touch the deletion or the rsync.

https://claude.ai/code/session_01KLwJz1CRif347aHf9xx5AU

### DIFF
--- a/kipi-update.sh
+++ b/kipi-update.sh
@@ -1560,6 +1560,15 @@ while IFS='|' read -r name path prefix itype declared; do
   # SNAP would point restore at a torn-down directory.
   CHECKPOINT_TARGET=""
   CHECKPOINT_PREFIX=""
+  # The checkpoint DIRECTORY too, not only the target. restore_instance returns
+  # early on an empty target, so a bail before this instance's checkpoint
+  # already restored nothing; clearing the directory as well makes that
+  # structural rather than dependent on one guard line, and stops the
+  # previous instance's snapshot outliving its iteration (PR #314 round 3).
+  if [ -n "${CHECKPOINT_DIR:-}" ] && [ -d "$CHECKPOINT_DIR" ]; then
+    rm -r -- "$CHECKPOINT_DIR"
+  fi
+  CHECKPOINT_DIR=""
   SNAP=""
   ORIGINAL_PATH="$path"
   ORIGINAL_HEAD=""
@@ -1905,6 +1914,10 @@ PY
         say "           Leaving it. It will be deleted by this sync until it is untracked."
         continue
       fi
+      # This untrack and its reset backouts are index writes too (PR #314
+      # round 3); a lock held past the bound skips this path, with the error
+      # printed, rather than failing the whole instance.
+      wait_for_index_lock "$path" "untrack $sys_path" || continue
       if ! git rm --cached --quiet -- "$sys_path" 2>/dev/null; then
         say "  WARNING: could not untrack $sys_path"
         continue
@@ -2207,21 +2220,24 @@ The file itself is untouched on disk." 2>/dev/null; then
       # judged a local edit, i.e. the case where sweeping would be worst.
       if [ "${#sys_add_paths[@]}" -eq 0 ]; then
         say "  nothing to commit: every dirty system-owned path is a local edit"
-      # The lock wait sits OUTSIDE the substitution: inside it, its progress
-      # lines are captured with the commit's stderr and printed only on failure,
-      # so a two-minute wait was silent (PR #314 round 2).
-      elif ! wait_for_index_lock "$path" "system-state commit"; then
-        echo "  WARNING: the system-state commit could not run (index.lock held); this instance will"
-        echo "  likely be refused below over dirt that is not founder work"
-      elif ! sys_commit_err="$(retry_on_index_lock "$path" "system-state commit" git commit -q -m "chore: commit system-written state before skeleton sync [no-issue: fleet updater system-state commit]
+      else
+        # No command substitution around the commit. Inside one, the lock
+        # wait's progress lines (the first wait AND every retry's) were
+        # captured with the commit's stderr and printed only on failure, so a
+        # two-minute wait was silent (PR #314 rounds 2 and 3). stdout reaches
+        # the terminal as it happens; only stderr is kept for the warning.
+        sys_errf="$(mktemp)"
+        if ! retry_on_index_lock "$path" "system-state commit" git commit -q -m "chore: commit system-written state before skeleton sync [no-issue: fleet updater system-state commit]
 
 These files are written by the fleet itself (sycophancy stamp, integrity
 baseline, hook state, skeleton-shipped plugins). Committing them here keeps the
 updater from being blocked by its own exhaust; founder work is never included
-because this commit is pathspec-limited." -- "${sys_add_paths[@]}" 2>&1)"; then
-        echo "  WARNING: the system-state commit FAILED; this instance will very"
-        echo "  likely be refused below over dirt that is not founder work:"
-        printf '%s\n' "$sys_commit_err" | sed 's/^/    /'
+because this commit is pathspec-limited." -- "${sys_add_paths[@]}" 2>"$sys_errf"; then
+          echo "  WARNING: the system-state commit FAILED; this instance will very"
+          echo "  likely be refused below over dirt that is not founder work:"
+          sed 's/^/    /' "$sys_errf"
+        fi
+        rm -- "$sys_errf"
       fi
     fi
 
--- a/kipi-update.sh
+++ b/kipi-update.sh
@@ -721,6 +721,7 @@ cleanup_updater_temps() {
     rm -r -- "$CHECKPOINT_DIR"
     CHECKPOINT_DIR=""
   fi
+  clear_run_marker
   cleanup_dry_model
 }
 
@@ -948,6 +949,7 @@ abandon_instance() {
   local message="${1:-}"
   [ -n "$message" ] && echo "$message"
   restore_instance
+  clear_run_marker
   if [ -n "${ARCHIVE_TMP:-}" ] && [ -d "$ARCHIVE_TMP" ]; then
     rm -r -- "$ARCHIVE_TMP"
     ARCHIVE_TMP=""
@@ -1191,6 +1193,113 @@ SH
   fi
   rm -r -- "$guard_dir"
   return "$rc"
+}
+
+# A LIVE index.lock is another writer, not debris, and the answer is to wait.
+#
+# 2026-09-06 14:30, consulting: the q-system commit landed, then the config
+# commit died seconds later on "Unable to create '.git/index.lock': File
+# exists" and the whole instance was abandoned half-delivered (q-system
+# committed, .claude/ written but uncommitted, plugins/ copied but never
+# added or deleted). The holder was a peer session's `git status`, which
+# refreshes the index under that lock for a fraction of a second. The stale
+# lock deletion at the top of the instance loop covers a crashed writer that
+# left a lock behind; it cannot cover a writer that is alive right now, and
+# deleting THAT lock is how two writers corrupt one index. Captured as
+# sp-523c1a25. So: every git command this run makes that takes the index
+# lock (add, commit) first waits for the lock to be absent, bounded, and a
+# commit that still dies on that exact error is retried a bounded number of
+# times. Any other failure is returned unchanged on the first attempt.
+LOCK_WAIT_S="${KIPI_UPDATE_LOCK_WAIT_S:-120}"
+LOCK_RETRY_MAX=3
+
+index_lock_path() {
+  local target="$1" lock
+  lock="$(git -C "$target" rev-parse --path-format=absolute --git-path index.lock 2>/dev/null)" \
+    || lock="$target/.git/index.lock"
+  printf '%s\n' "$lock"
+}
+
+# $1 = instance path, $2 = the step name the log and the failure carry.
+# Returns 1 only when the lock outlived the bound; a missing lock returns 0
+# at once, so calling this before every index write costs nothing when the
+# checkout is quiet.
+wait_for_index_lock() {
+  local target="$1" step="$2" lock waited=0
+  lock="$(index_lock_path "$target")"
+  while [ -e "$lock" ]; do
+    if [ "$waited" -ge "$LOCK_WAIT_S" ]; then
+      echo "  ERROR: index.lock held past ${LOCK_WAIT_S}s at $step; another git writer owns this checkout" >&2
+      return 1
+    fi
+    if [ "$waited" -eq 0 ] || [ $((waited % 10)) -eq 0 ]; then
+      echo "  waiting for index.lock at $step (${waited}s of ${LOCK_WAIT_S}s)"
+    fi
+    sleep 1
+    waited=$((waited + 1))
+  done
+  return 0
+}
+
+# $1 = instance path, $2 = step name, $3.. = the commit command to run.
+# Retries ONLY on the live-lock error text; every other failure returns on
+# attempt 1 with its stderr intact, because a pre-commit refusal or a bad
+# pathspec is not something waiting fixes (self-healing-retry.md rule 5).
+retry_on_index_lock() {
+  local target="$1" step="$2"
+  shift 2
+  local attempt=1 rc errf
+  errf="$(mktemp)"
+  while :; do
+    if ! wait_for_index_lock "$target" "$step"; then
+      rm -- "$errf"
+      return 1
+    fi
+    if "$@" 2>"$errf"; then
+      rc=0
+    else
+      rc=$?
+    fi
+    if [ "$rc" -eq 0 ]; then
+      cat "$errf" >&2
+      rm -- "$errf"
+      return 0
+    fi
+    if [ "$attempt" -lt "$LOCK_RETRY_MAX" ] &&
+        grep -qE "index\.lock': File exists|Unable to create .*index\.lock" "$errf"; then
+      echo "  commit at $step hit a live index.lock (attempt $attempt of $LOCK_RETRY_MAX); waiting and retrying"
+      attempt=$((attempt + 1))
+      sleep 1
+      continue
+    fi
+    cat "$errf" >&2
+    rm -- "$errf"
+    return "$rc"
+  done
+}
+
+# The run marker: while one instance apply is in flight, <git-common-dir>/
+# kipi-update.run holds this pid and the start time. The instance's own
+# Stop-hook auto-commit (q-system/hooks/auto-commit.py) refuses to commit
+# while a LIVE marker exists, so the updater's half-delivered files are not
+# swept into the hook's generic commits mid-run (2026-09-06 14:33, consulting:
+# "chore: update rules (10 files)" was the skeleton's rules, committed by the
+# hook under its own name while the updater was still delivering; sp-9306036e).
+# The common dir, not the worktree's git dir, so a linked worktree of the
+# same checkout reads the same marker. Cleared on every exit path.
+RUN_MARKER=""
+write_run_marker() {
+  local target="$1" common
+  common="$(git -C "$target" rev-parse --path-format=absolute --git-common-dir 2>/dev/null)" \
+    || return 0
+  RUN_MARKER="$common/kipi-update.run"
+  printf '%s %s\n' "$$" "$(date -u +%Y-%m-%dT%H:%M:%SZ)" > "$RUN_MARKER"
+}
+clear_run_marker() {
+  if [ -n "${RUN_MARKER:-}" ] && [ -f "$RUN_MARKER" ]; then
+    rm -- "$RUN_MARKER"
+  fi
+  RUN_MARKER=""
 }
 
 # One answer to "is this untracked file the founder's work, or this sync's own
@@ -1646,6 +1755,7 @@ PY
         rm -f "$lockfile"
       fi
     done
+    write_run_marker "$path"
 
     # Abort any zombie rebase/merge/cherry-pick
     if [ -d "$path/.git/rebase-merge" ] || [ -d "$path/.git/rebase-apply" ]; then
@@ -1758,7 +1868,7 @@ PY
         git reset --quiet -q -- "$sys_path" >/dev/null 2>&1 || true
         continue
       fi
-      if git commit -q -m "chore: untrack instance-local $sys_path [no-issue: fleet updater instance-local untrack]
+      if wait_for_index_lock "$path" "untrack commit" && git commit -q -m "chore: untrack instance-local $sys_path [no-issue: fleet updater instance-local untrack]
 
 This file is instance-local (ASK-282) and gitignored by policy. While it was
 tracked, the skeleton sync deleted it -- the skeleton once tracked the path and
@@ -2050,7 +2160,7 @@ The file itself is untouched on disk." 2>/dev/null; then
       # judged a local edit, i.e. the case where sweeping would be worst.
       if [ "${#sys_add_paths[@]}" -eq 0 ]; then
         say "  nothing to commit: every dirty system-owned path is a local edit"
-      elif ! sys_commit_err="$(git commit -q -m "chore: commit system-written state before skeleton sync [no-issue: fleet updater system-state commit]
+      elif ! sys_commit_err="$(retry_on_index_lock "$path" "system-state commit" git commit -q -m "chore: commit system-written state before skeleton sync [no-issue: fleet updater system-state commit]
 
 These files are written by the fleet itself (sycophancy stamp, integrity
 baseline, hook state, skeleton-shipped plugins). Committing them here keeps the
@@ -2307,13 +2417,14 @@ PY
         rm -r -- "$ARCHIVE_TMP"
         ARCHIVE_TMP=""
         cd "$path"
-        if ! stage_q_system_sync "$path" "$prefix" 2>/dev/null; then
+        if ! wait_for_index_lock "$path" "q-system sync staging" ||
+            ! stage_q_system_sync "$path" "$prefix" 2>/dev/null; then
           unstage_scope "$path" "$prefix/"
           abandon_instance "  ERROR: could not stage q-system sync" && continue
         fi
         CHANGES=$(git diff --cached --name-only 2>/dev/null | wc -l | tr -d ' ')
         if [ "$CHANGES" != "0" ]; then
-          if ! guarded_commit "$path" \
+          if ! retry_on_index_lock "$path" "q-system sync commit" guarded_commit "$path" \
               "chore: sync q-system from skeleton $(date +%Y-%m-%d) [no-issue: fleet updater skeleton sync]"; then
             abandon_instance "  ERROR: could not commit q-system sync" && continue
           fi
@@ -2512,7 +2623,8 @@ print("\n".join(mod.EXTRA_WATCHED))
             >/dev/null 2>&1; then
           git rm -r -q --cached plugins/memory-lifecycle
         fi &&
-        { stage_config_sync "$path" || { unstage_scope "$path" .claude/ plugins/; false; }; } &&
+        { { wait_for_index_lock "$path" "config sync staging" && stage_config_sync "$path"; } ||
+            { unstage_scope "$path" .claude/ plugins/; false; }; } &&
         # THE COMMIT HALF NEEDS THE SAME UNWIND AS THE STAGING HALF (ASK-797).
         # unstage_scope was wired to stage_config_sync's failure only, so a
         # staging error unwound cleanly and a COMMIT error did not -- it left the
@@ -2527,7 +2639,7 @@ print("\n".join(mod.EXTRA_WATCHED))
         # shape, and no later run could clear it -- a worktree checkout does not
         # touch the index, so it looked like founder work forever.
         { if ! git diff --cached --quiet 2>/dev/null; then
-            guarded_commit "$path" \
+            retry_on_index_lock "$path" "config sync commit" guarded_commit "$path" \
               "chore: sync .claude config + plugins from skeleton $(date +%Y-%m-%d) [no-issue: fleet updater skeleton sync]" ||
               { unstage_scope "$path" .claude/ plugins/; false; }
           fi; }
@@ -2543,6 +2655,7 @@ print("\n".join(mod.EXTRA_WATCHED))
     fi
 
     echo "  Config synced"
+    clear_run_marker
   fi
 
   # Post-sync capability gate (structure/wiring/data diff, no test execution —
--- a/kipi-update.sh
+++ b/kipi-update.sh
@@ -2207,6 +2207,12 @@ The file itself is untouched on disk." 2>/dev/null; then
       # judged a local edit, i.e. the case where sweeping would be worst.
       if [ "${#sys_add_paths[@]}" -eq 0 ]; then
         say "  nothing to commit: every dirty system-owned path is a local edit"
+      # The lock wait sits OUTSIDE the substitution: inside it, its progress
+      # lines are captured with the commit's stderr and printed only on failure,
+      # so a two-minute wait was silent (PR #314 round 2).
+      elif ! wait_for_index_lock "$path" "system-state commit"; then
+        echo "  WARNING: the system-state commit could not run (index.lock held); this instance will"
+        echo "  likely be refused below over dirt that is not founder work"
       elif ! sys_commit_err="$(retry_on_index_lock "$path" "system-state commit" git commit -q -m "chore: commit system-written state before skeleton sync [no-issue: fleet updater system-state commit]
 
 These files are written by the fleet itself (sycophancy stamp, integrity
@@ -2666,11 +2672,14 @@ print("\n".join(mod.EXTRA_WATCHED))
     # .claude/ and plugins/ permanently dirty in every instance repo.
     if git -C "$path" rev-parse --git-dir >/dev/null 2>&1; then
       if ! ( cd "$path" &&
+        # The untrack below is this block's FIRST index write, so the wait sits
+        # before it, not before the staging that follows (PR #314 round 2).
+        wait_for_index_lock "$path" "config sync staging" &&
         if git ls-files --error-unmatch plugins/memory-lifecycle \
             >/dev/null 2>&1; then
           git rm -r -q --cached plugins/memory-lifecycle
         fi &&
-        { { wait_for_index_lock "$path" "config sync staging" && stage_config_sync "$path"; } ||
+        { stage_config_sync "$path" ||
             { unstage_scope "$path" .claude/ plugins/; false; }; } &&
         # THE COMMIT HALF NEEDS THE SAME UNWIND AS THE STAGING HALF (ASK-797).
         # unstage_scope was wired to stage_config_sync's failure only, so a
--- a/kipi-update.sh
+++ b/kipi-update.sh
@@ -870,6 +870,18 @@ restore_instance() {
   # passes it legitimately -- and an unscoped checkout here would then delete
   # exactly the work the scoping was meant to stop blocking on. The two must
   # move together; the pathspec below is the same one the guard uses.
+  # The restore itself takes index.lock (reset, checkout). An abandon caused
+  # by a lock held past the bound would otherwise run these two silently into
+  # the same lock, restore nothing, and leave this run's writes uncommitted:
+  # dirty forever, refused by every later run (PR #314 review, round 1).
+  # Wait the same bound again; if the writer is still there, say exactly what
+  # was left behind instead of pretending the checkpoint was restored.
+  if ! wait_for_index_lock "$target" "restore"; then
+    echo "  ERROR: restore could NOT run: index.lock still held after ${LOCK_WAIT_S}s;" \
+      "this instance keeps this run's uncommitted writes under $CHECKPOINT_PREFIX/, .claude/ and plugins/" \
+      "until the lock clears and the tree is restored by hand" >&2
+    return 1
+  fi
   git -C "$target" reset -q HEAD -- "$CHECKPOINT_PREFIX/" .claude/ plugins/ \
     $(pathspec_owned_excludes "$CHECKPOINT_PREFIX") 2>/dev/null || true
   # `git checkout -- A B C` is ALL-OR-NOTHING. If ANY pathspec matches nothing
@@ -1288,16 +1300,49 @@ retry_on_index_lock() {
 # The common dir, not the worktree's git dir, so a linked worktree of the
 # same checkout reads the same marker. Cleared on every exit path.
 RUN_MARKER=""
+RUN_MARKER_MAX_AGE_S="${KIPI_UPDATE_RUN_MARKER_MAX_AGE_S:-7200}"
+
+# The marker is also the one-run-per-instance lock. A marker whose pid is
+# alive and whose stamp is younger than the bound belongs to ANOTHER updater
+# still applying this instance; writing over it would let two runs interleave
+# on one index and let the first run's clear_run_marker strip the second's
+# protection (PR #314 review, round 1). Returns 1 in that case; the caller
+# abandons the instance. A dead pid or an old stamp is a crashed run's
+# leftover and is replaced.
+run_marker_is_live_foreign() {
+  local marker="$1" pid stamp age
+  [ -f "$marker" ] || return 1
+  read -r pid stamp < "$marker" || return 1
+  case "$pid" in ''|*[!0-9]*) return 1 ;; esac
+  [ "$pid" != "$$" ] || return 1
+  kill -0 "$pid" 2>/dev/null || return 1
+  age="$(python3 -c 'import calendar,sys,time
+try:
+    print(int(time.time() - calendar.timegm(time.strptime(sys.argv[1], "%Y-%m-%dT%H:%M:%SZ"))))
+except Exception:
+    print(-1)' "${stamp:-}" 2>/dev/null)" || age=-1
+  [ "$age" -ge 0 ] && [ "$age" -le "$RUN_MARKER_MAX_AGE_S" ]
+}
 write_run_marker() {
   local target="$1" common
   common="$(git -C "$target" rev-parse --path-format=absolute --git-common-dir 2>/dev/null)" \
     || return 0
+  if run_marker_is_live_foreign "$common/kipi-update.run"; then
+    echo "  ERROR: another fleet update is applying this instance right now (pid $(cut -d' ' -f1 "$common/kipi-update.run")); refusing to run two at once" >&2
+    return 1
+  fi
   RUN_MARKER="$common/kipi-update.run"
   printf '%s %s\n' "$$" "$(date -u +%Y-%m-%dT%H:%M:%SZ)" > "$RUN_MARKER"
 }
+# Only this run's own marker is ever removed: the first field is the pid that
+# wrote it, and a marker carrying any other pid is left for its owner.
 clear_run_marker() {
+  local owner
   if [ -n "${RUN_MARKER:-}" ] && [ -f "$RUN_MARKER" ]; then
-    rm -- "$RUN_MARKER"
+    owner="$(cut -d' ' -f1 "$RUN_MARKER" 2>/dev/null || true)"
+    if [ "$owner" = "$$" ]; then
+      rm -- "$RUN_MARKER"
+    fi
   fi
   RUN_MARKER=""
 }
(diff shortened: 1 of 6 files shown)


assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-314-20260906-160244.md` (5467 bytes).

```
FINDINGS:
major|The lock-bound refusal abandons while the lock is still held, so restore_instance's reset/checkout silently no-op and the instance is left dirty and refused by every later run|kipi-update.sh:2423
major|auto-commit's marker check trusts pid liveness alone and ignores the marker's own timestamp, so a marker surviving SIGKILL or reboot plus pid reuse disables auto-commit permanently|q-system/hooks/auto-commit.py:427
minor|clear_run_marker unlinks the marker without checking the pid is this run's, and no global run lock exists, so an overlapping kipi update removes the live run's protection|kipi-update.sh:1298
END FINDINGS
```

--- reviewer output ---

# Review: PR #314 (kipi-update index.lock wait + auto-commit run marker)

Both new suites are green on my machine (`test_auto_commit_run_marker.py` 4/4, `test-kipi-update-lock-wait.sh` 4/4). Three findings survived reproduction.

---

## major 1: the bounded refusal leaves the instance permanently dirty

`kipi-update.sh:2423` refuses via `abandon_instance` at the moment the lock has outlived the bound, which means the lock is *still held by definition*. `restore_instance` then runs `git reset` (`kipi-update.sh:873`) and `git checkout`, both under `2>/dev/null || true`. Both need that same index lock. They silently no-op, and the rsync'd skeleton content stays in the instance worktree.

Reproducer (the PR's own `build` / `hold_lock` fixture, holder never releases, `KIPI_UPDATE_LOCK_WAIT_S=3`):

```
  ERROR: index.lock held past 3s at q-system sync staging; another git writer owns this checkout
  ERROR: could not stage q-system sync
=== Summary ===
  Updated: 0
  Failed:  1
===== git status --porcelain =====
 M q-system/tracked.md
===== tracked.md =====
'generic skeleton content\n'          <- skeleton content, the instance's 'instance state' is gone from disk
```

Second run, no holder at all:

```
--- testinst (subtree) ---
     M q-system/tracked.md
  ERROR: dirty working tree; refusing to commit unrelated work
  Failed:  1
```

The instance is stuck for every future run until a human cleans it. That is the half-delivered-instance failure this PR exists to prevent, now reachable through the new refusal path. Property 2 of the new test asserts "commits nothing" and never asserts a clean tree, so it passes straight over this.

## major 2: a reused pid disables auto-commit forever, and the marker's timestamp is never read

`q-system/hooks/auto-commit.py:427` decides "live run" from `os.kill(pid, 0)` alone. The marker survives SIGKILL and reboot (the EXIT trap does not run), and a pid that later belongs to any other live process reads as an in-progress fleet update.

Marker survival, updater killed with SIGKILL while parked in the wait:

```
marker appeared: True
updater killed. marker survives: True
marker contents: '91865 2026-09-06T23:08:35Z\n' -> dead pid 91865
```

The failure state, marker holding pid 1:

```
hook rc: 0
hook stderr: auto-commit: fleet updater run in progress (pid 1); committing nothing
commits before/after: 1 1
marker still there: True
dirty still uncommitted: M q-system/memory/last-handoff.md
second run stderr: auto-commit: fleet updater run in progress (pid 1); committing nothing | commits: 1
```

The hook never removes the marker in that branch, so this is permanent, not one-shot. `write_run_marker` already writes a UTC start time into field 2 and nothing reads it. A staleness bound on that field closes this.

## minor 3: an overlapping run clears the live run's marker

`clear_run_marker` (`kipi-update.sh:1298`) unlinks the file with no check that the pid inside is this run's, and there is no global single-run lock anywhere in `kipi-update.sh` (`grep flock|already running|pgrep` is empty).

```
A (pid 97393) wrote marker: 97393 2026-09-06T23:09:16Z
B (pid 97398) wrote marker: 97398 2026-09-06T23:09:16Z
A now finishes ITS instance and calls clear_run_marker
MARKER GONE -- run B is live and now unprotected
```

Dropping a level and saying so: I had to start two updaters by hand. Nothing in the repo forbids it, and the incident this PR describes involved three founder runs in one afternoon.

---

## What is sound

Attacks that the code survived:

- **`--git-common-dir` relativity.** From a subdirectory git returns `../../.git`, not `.git`, so the hook's `os.path.abspath` resolves to the right marker. Tested on git 2.54.0; no bug.
- **The retry's error-text match against the real producer.** `guarded_commit` pipes only stdout through `dry_filter` (`kipi-update.sh:97`), so git's `Unable to create '...index.lock': File exists` reaches `retry_on_index_lock`'s errf and the grep fires. Not a fixture-only match.
- **Property 3's mutant is a real mutant.** Stubbing `wait_for_index_lock` to `return 0` makes the green case red; I ran the suite and got 4/4 with that case passing.
- **Non-lock errors are not retried.** Their unit case proves one attempt and the original stderr intact.
- **`write_run_marker` is not reached on a plain `--dry-run`** (gated on the same `DRY_RUN != --dry-run || MODEL_RUN = 1` condition as the apply block).

---

**VERDICT: REQUEST CHANGES.** Not BLOCK: both majors are recoverable by a human deleting a file or running `git checkout`.

Fix first: make the refusal path leave a clean tree. `restore_instance` needs the same lock wait (or the abandon needs to wait for the lock to clear before restoring), and property 2 of the new test needs a `git status --porcelain` assertion so the gap cannot reopen.

FINDINGS:
major|The lock-bound refusal abandons while the lock is still held, so restore_instance's reset/checkout silently no-op and the instance is left dirty and refused by every later run|kipi-update.sh:2423
major|auto-commit's marker check trusts pid liveness alone and ignores the marker's own timestamp, so a marker surviving SIGKILL or reboot plus pid reuse disables auto-commit permanently|q-system/hooks/auto-commit.py:427
minor|clear_run_marker unlinks the marker without checking the pid is this run's, and no global run lock exists, so an overlapping kipi update removes the live run's protection|kipi-update.sh:1298
END FINDINGS

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-314-20260906-162904.md` (8882 bytes).

```
FINDINGS:
major|The config sync's first index write (git rm --cached plugins/memory-lifecycle) has no lock wait or retry, so a one-second peer lock still abandons the instance half-delivered|kipi-update.sh:2671
minor|auto-commit.py resolves the run marker against the process cwd instead of CLAUDE_PROJECT_DIR, so the new guard fails open and commits under a live marker when the two differ|q-system/hooks/auto-commit.py:426
minor|The system-state commit's lock wait is captured by command substitution, so a 120s wait prints nothing and the progress lines are discarded on success|kipi-update.sh:2210
minor|The stale-marker test's dead pid is inherited from a previous assertion's background job, so the case is order-dependent and never exercises the dead-pid branch|q-system/.q-system/scripts/test/test-kipi-update-lock-wait.sh:194
END FINDINGS
```

--- reviewer output ---

# PR #314 review

Read-only. Repo left clean (`git status --porcelain` empty; scratch lives in the gitignored `.review-scratch/`).

Shipped tests first: `test-kipi-update-lock-wait.sh` is **7/7 green** and `test_auto_commit_run_marker.py` is green. The mutant case (wait stubbed out) really does go red. That part is not decoration.

---

## MAJOR 1 — the config sync's FIRST index write has no wait and no retry, so the 14:30 failure still reproduces on this branch

`kipi-update.sh:2671` — `git rm -r -q --cached plugins/memory-lifecycle` runs *before* `wait_for_index_lock "config sync staging"` in the same `&&` chain. It takes index.lock and nothing guards it. The PR body says "Called before every index write the run makes"; `kipi-update.sh:1216` says "every git command this run makes that takes the index lock (add, commit) first waits". Both are false at this line.

The fixture is the shipped test's own `build()` with one change: the instance tracks `plugins/memory-lifecycle`, which is what makes line 2671 run. MEMORY.md records fractional-cxo carrying exactly that shape.

Reproducer: `bash .review-scratch/repro_unguarded_index_write.sh`, holder takes the lock for **1 second** (a peer `git status` is shorter still).

Real output:

```
  Syncing q-system/ from skeleton...
[main 5dd6aef] chore: sync q-system from skeleton 2026-09-06 [...]
  OK (5 files updated)
  Syncing .claude/ config...
fatal: Unable to create '.../inst/.git/index.lock': File exists.
  ERROR: config sync did not reach a complete committed state
=== Summary ===
  Updated: 1
  Failed:  1
RESULT: no 'waiting for index.lock at config sync staging' line was ever printed.
instance HEAD: chore: sync q-system from skeleton 2026-09-06 [...]
```

That is the 2026-09-06 14:30 incident verbatim: q-system committed, `.claude/` and `plugins/` not, instance abandoned half-delivered by a sub-second peer lock. The new wait and retry never engage because the unguarded `git rm` short-circuits the chain first.

With the same holder never releasing (`sleep 60` variant, also run):

```
  ERROR: restore could NOT run: index.lock still held after 15s; this instance
  keeps this run's uncommitted writes under q-system/, .claude/ and plugins/
  until the lock clears and the tree is restored by hand
```

**Fix:** put line 2671 behind the same `wait_for_index_lock` (or move the wait above the `git ls-files --error-unmatch` test). `git reset` at `kipi-update.sh:1915` and `:1927` are index writes too, though those are `|| true`.

---

## MINOR 2 — the hook's marker lookup resolves against the process cwd, not `CLAUDE_PROJECT_DIR`

`q-system/hooks/auto-commit.py:426`. Every other git call in that file goes through `run(..., cwd=PROJ_DIR)` — the file is deliberately built to work when its own cwd is not the repo, and `test-slack-notify-label.sh:101` pins that convention. `fleet_update_in_progress()` asks git for `--git-common-dir` in `PROJ_DIR`, gets a path relative to `PROJ_DIR`, then resolves it with `os.path.abspath`, which uses the process cwd. The shell half of this same handshake uses `--path-format=absolute`; the Python half does not.

Reproducer: `python3 .review-scratch/repro_projdir.py`

```
--- control: hook cwd == CLAUDE_PROJECT_DIR (the shape the new test uses)
    stderr               : 'auto-commit: fleet updater run in progress (pid 80786); committing nothing'
    VERDICT              : GUARD HELD (no commit)

--- hook cwd elsewhere, CLAUDE_PROJECT_DIR = the repo
    stderr                 : ''
    commits before/after   : 1 -> 2
    VERDICT                : GUARD FAILED OPEN (committed under a live marker)
```

Dropping this to minor per the calibration rule: I constructed the cwd/`CLAUDE_PROJECT_DIR` split and could not point at a producer in this repo that emits it. The shipped test never sets `CLAUDE_PROJECT_DIR`, so it only ever runs the equal case. If a Stop hook can fire with the two different (subagent worktrees, `/add-dir`), this is the full sp-9306036e recurrence. One-word fix: add `--path-format=absolute`.

Attack that survived: from a plain subdirectory git returns `../../.git` and `abspath` resolves it correctly. Checked, git 2.54.0.

---

## MINOR 3 — the system-state commit waits up to 120s in total silence

`kipi-update.sh:2210` wraps `retry_on_index_lock` in `$( ... 2>&1 )`. Command substitution captures stdout, and `wait_for_index_lock` writes its progress to stdout. On success the lines are thrown away entirely.

Reproducer: `bash .review-scratch/repro_minor_b.sh`

```
=== A. direct call (what the config-sync chain does): operator SEES progress
  waiting for index.lock at config sync staging (0s of 3s)
  ERROR: index.lock held past 3s at config sync staging; another git writer owns this checkout
=== B. wrapped exactly like kipi-update.sh:2209: operator sees NOTHING
    <no output appeared above during the 3s wait>
    swallowed into sys_commit_err instead: [  waiting for index.lock at system-state commit (0s of 3s)|  ERROR: index.lock held past 3s ...]
```

At 3am with the 120s default, a run that hits this looks hung. Send the progress lines to stderr, or don't capture stdout there.

---

## MINOR 4 — the stale-marker test's "dead pid" comes from the previous assertion, not from the line that claims to make it

`q-system/.q-system/scripts/test/test-kipi-update-lock-wait.sh:194`. `dead="$( (sleep 0 &) ; echo $! )"` — the `&` is in a nested subshell, so `$!` is never set by it. The value that lands in `dead` is inherited from the earlier `hold_lock ... &` in a previous assertion.

Reproducer: `bash .review-scratch/repro_two_minors.sh`

```
-- the exact line, run FIRST (no prior background job):
_a1.sh: line 2: $!: unbound variable
rc=1
-- same line AFTER a background job:
dead=[211]  <- inherited from the EARLIER '&', not from (sleep 0 &)
```

Two consequences: reordering or running that assertion alone aborts the file under `set -u`, and since the stamp is hardcoded `2026-01-01T00:00:00Z` the case is stale by the age bound regardless, so the dead-pid branch is never what makes it pass. (The Python test's `dead_pid()` is correct; this is the shell one only.)

---

## What is sound

- **The mutant is load-bearing.** Full suite 7/7, including the wait-stubbed-out case going red. Not decoration.
- **`--git-common-dir` from a subdirectory.** Returns `../../.git`; `abspath` resolves it correctly. That specific worry is unfounded.
- **The retry unit case.** Helpers extracted from the updater's own text, exactly 3 attempts asserted, non-lock failure returns rc=1 on attempt 1 with its stderr intact. Real assertions.
- **The no-marker control in the Python test is a genuine floor.** My own control run confirms the same dirty file IS committed without a marker, so "no commit" under a live marker is the marker's doing.
- **Pid recycling.** The age bound plus `test_live_pid_with_an_old_stamp_is_stale` covers it on both sides.
- **`clear_run_marker` ownership.** Only a marker whose first field equals `$$` is removed; the foreign-marker case asserts the other run's marker survives. Correct.
- **The restore-path wait added in round 1** does fire and does say what it left behind. My 60s-hold run printed exactly that line.

One thing I noticed and did **not** reproduce, so I am not filing it: the new `abandon_instance` at `kipi-update.sh:1804` sits *before* `checkpoint_instance` (`:2257`), so `restore_instance` there runs against the previous instance's `CHECKPOINT_TARGET`. The dirty-tree abandon at `:2249` already has that shape, so it looks pre-existing rather than introduced here. Worth a separate look.

---

## VERDICT: REQUEST CHANGES

One major. Not a BLOCK: nothing here is unrecoverable, and the PR is a real improvement over the current state.

**Fix first:** `kipi-update.sh:2671`. Until that `git rm --cached` is behind the wait, the config sync still dies on a sub-second peer lock and still abandons the instance half-delivered, which is the whole reason this PR exists.

FINDINGS:
major|The config sync's first index write (git rm --cached plugins/memory-lifecycle) has no lock wait or retry, so a one-second peer lock still abandons the instance half-delivered|kipi-update.sh:2671
minor|auto-commit.py resolves the run marker against the process cwd instead of CLAUDE_PROJECT_DIR, so the new guard fails open and commits under a live marker when the two differ|q-system/hooks/auto-commit.py:426
minor|The system-state commit's lock wait is captured by command substitution, so a 120s wait prints nothing and the progress lines are discarded on success|kipi-update.sh:2210
minor|The stale-marker test's dead pid is inherited from a previous assertion's background job, so the case is order-dependent and never exercises the dead-pid branch|q-system/.q-system/scripts/test/test-kipi-update-lock-wait.sh:194
END FINDINGS

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-314-20260906-171620.md` (9127 bytes).

```
FINDINGS:
minor|The system-state `git add` is an unwaited index write whose failure is swallowed, so a live index.lock loses the staging and the pathspec commit then fails entirely, abandoning the instance|kipi-update.sh:2185
minor|A leftover run marker whose pid is recycled to an unrelated live process refuses the instance for up to RUN_MARKER_MAX_AGE_S and claims a fleet update is running|kipi-update.sh:1331
minor|The round-2 fix moving the lock wait outside the command substitution is incomplete: retry_on_index_lock's own wait still runs inside it, so up to LOCK_WAIT_S of waiting is silent|kipi-update.sh:2216
nit|The untrack staging `git rm --cached` is an index write with no wait before it; the wait sits on the commit two statements later|kipi-update.sh:1908
nit|A non-numeric KIPI_UPDATE_LOCK_WAIT_S makes the bound unreachable, turning the bounded wait into an unbounded spin|kipi-update.sh:1243
END FINDINGS
```

--- reviewer output ---

Read the change, ran both new suites plus the neighbouring regression suite, then attacked the parts the suites do not cover. Findings below, every one executed.

---

## Findings

### 1. minor — the system-state `git add` is an unwaited, silently-swallowed index write, so the exact lock race this PR exists to kill still abandons the instance through that door

`kipi-update.sh:2185`

The PR body claims the wait is "called before every index write the run makes." It is not. `git add -- "${sys_add_paths[@]}" 2>/dev/null || true` at 2185 is an index write with no `wait_for_index_lock` before it and its failure discarded. The wait only arrives at 2213/2216, after the staging is already lost. Round 2 fixed this exact shape in the config-sync block ("The untrack below is this block's FIRST index write, so the wait sits before it"); the same shape survives here.

The consequence is not cosmetic. The repo's own regression suite proves the second half: a single untracked path in the pathspec makes `git commit -- <paths>` commit **nothing at all**, so the tracked system dirt also stays behind and the dirty-tree guard at 2249 then abandons the instance.

Reproducer, link 1 (real git under a live lock, using the updater's own idiom):

```
python3 -  # full script in transcript: git add -- <path> with .git/index.lock present
```
```
git add under a live lock -> rc=128  stderr="fatal: Unable to create '.../.git/index.lock': File exists...."
   the updater swallows this: `git add -- ... 2>/dev/null || true` (kipi-update.sh:2185)
then the guarded commit -> rc=1  stderr="error: pathspec 'q-system/.q-system/state/hook-state.json' did not match any file(s) known to git"
system file committed: False
tree still dirty: ?? q-system/
```

Reproducer, link 2 (the repo's own suite, run on this branch):

```
bash q-system/.q-system/scripts/test/test-kipi-update-system-state-commit.sh
```
```
PASS: one untracked path makes git commit -- <pathspec> commit nothing at all
... 11/11
```

Not a regression (pre-PR behaviour was identical), which is why this is minor and not major. It is incomplete coverage of the PR's headline claim.

---

### 2. minor — a leftover run marker plus a recycled pid refuses the instance for up to two hours, and the message asserts something the code does not know

`kipi-update.sh:1331`

`run_marker_is_live_foreign` treats "some process with this pid is alive and the stamp is under 7200s" as "another fleet update is applying this instance right now." After a SIGKILL or a reboot the marker survives on disk; any unrelated process that inherits that pid inside the age bound blocks the sync and the operator is told a fleet update is running that is not.

Reproducer (the updater's own three marker functions, extracted from `kipi-update.sh` the way the PR's own property-4 case extracts helpers, pointed at a plain `sleep 300`):

```
python3 -  # extracts run_marker_is_live_foreign / write_run_marker / clear_run_marker, writes a marker holding sleep's pid
```
```
pid 74529 is: sleep 300
RESULT: instance REFUSED (rc=1)
ERROR: another fleet update is applying this instance right now (pid 74529); refusing to run two at once
```

Recoverable (delete the file, or wait out `RUN_MARKER_MAX_AGE_S`) and loud, so minor. Worth noting the hook half of the same handshake fails safe here and the updater half fails closed.

---

### 3. minor — the round-2 "wait sits OUTSIDE the substitution" fix is incomplete

`kipi-update.sh:2216`

The comment above 2213 says the wait was moved out of the command substitution so a two-minute wait is not silent. But `retry_on_index_lock` calls `wait_for_index_lock` itself, on every attempt including attempt 1, and that call is inside `sys_commit_err="$( ... 2>&1)"`. A writer that takes the lock between the outer wait and the retry's wait produces up to `LOCK_WAIT_S` of silence, and on success the captured progress is thrown away.

Reproducer (extracted helpers, a lock released after 4s, exactly the shape at 2216):

```
python3 -  # retry_on_index_lock "$repo" "system-state commit" ok  inside "$( ... 2>&1 )"
```
```
elapsed 4.1s
=== what the OPERATOR saw on the console during the wait: (nothing above this line)
=== what was captured instead and thrown away on success:
    captured>   waiting for index.lock at system-state commit (0s of 30s)
```

---

### 4. nit — the untrack staging has the same unwaited-index-write shape

`kipi-update.sh:1908`

`git rm --cached --quiet -- "$sys_path" 2>/dev/null` is an index write; the wait is on the commit two statements later at 1918. Lower impact than finding 1 because this one is loud (`WARNING: could not untrack $sys_path`) and backs itself out. Same reproducer mechanism as finding 1.

---

### 5. nit — a non-numeric `KIPI_UPDATE_LOCK_WAIT_S` makes the bounded wait unbounded

`kipi-update.sh:1243`

`[ "$waited" -ge "$LOCK_WAIT_S" ]` errors instead of returning true, so the bound is never reached and the loop spins with one stderr line per second.

```
python3 -  # wait_for_index_lock with KIPI_UPDATE_LOCK_WAIT_S=2m, lock never released, 8s timeout
```
```
STILL SPINNING after 8s with KIPI_UPDATE_LOCK_WAIT_S=2m (bound never reached)
stderr sample: bash: line 13: [: 2m: integer expression expected | bash: line 13: [: 2m: integer expression expected | ...
```

---

## What is sound (attacks that failed)

- **The linked-worktree claim in `auto-commit.py` is true and untested, so I tested it.** Built a repo plus a linked worktree, put the marker in the main `.git`, ran the hook with `CLAUDE_PROJECT_DIR` pointed at the worktree. Output: `auto-commit: fleet updater run in progress (pid 62003); committing nothing`, commits before/after 1 and 1. The `--git-common-dir` choice holds.
- **The retry fixture matches a real producer.** Property 4 uses a hand-written `flaky_commit`, which is exactly the invented-fixture shape worth suspecting. I ran a real `git commit` against a real `index.lock`: `fatal: Unable to create '.../index.lock': File exists.`, and the updater's grep pattern matches it. Same for `git add`. The fixture is not a fantasy.
- **The mutant is load-bearing.** Property 3 stubs `wait_for_index_lock` to a no-op and asserts the green case goes red. It passes, so property 1 is not passing on timing.
- **No marker leaks across instances in one fleet run.** I traced every `continue` between `write_run_marker` (1800) and `clear_run_marker` (2712): all top-level ones route through `abandon_instance`, which clears; the rest are inner-loop continues. The EXIT trap covers a crash, and `clear_run_marker` only removes a marker whose first field is `$$`.
- **The guard cannot block auto-commit forever.** The stamp is fixed at write time, so the worst case is `RUN_MARKER_MAX_AGE_S` and then the hook removes the marker itself. That is the right shape.
- **`--system-state` does not hit the new guard.** The classifier branch short-circuits at `__main__` before `main()`, and the updater shells it from inside an instance whose marker is live. No interaction.
- **`restore_instance` returning 1 does not kill the run.** Every `abandon_instance` call site is `abandon_instance ... && continue`, so errexit is suppressed through the function body; the suite's third case reaches `Failed: 1` and prints the summary, which confirms it empirically.
- **Suites are green as claimed.** `test-kipi-update-lock-wait.sh` 7/7, `test_auto_commit_run_marker.py` 6/6, `test-kipi-update-system-state-commit.sh` 11/11. One note: the PR body says that last one is 8/8; it emits 11 PASS lines on this branch. Reported number, not measured number.

---

## VERDICT: APPROVE WITH NITS

Nothing here is permanent, nothing is unrecoverable, and the two new guards do what they say. The findings are all bounded and loud.

**Fix first:** finding 1. Move a `wait_for_index_lock "$path" "system-state staging"` in front of the `git add` at 2185 and stop swallowing its failure. That is the one door left open to the exact race that cost three founder runs on 2026-09-06, and the same fix pattern is already applied twice elsewhere in this diff.

FINDINGS:
minor|The system-state `git add` is an unwaited index write whose failure is swallowed, so a live index.lock loses the staging and the pathspec commit then fails entirely, abandoning the instance|kipi-update.sh:2185
minor|A leftover run marker whose pid is recycled to an unrelated live process refuses the instance for up to RUN_MARKER_MAX_AGE_S and claims a fleet update is running|kipi-update.sh:1331
minor|The round-2 fix moving the lock wait outside the command substitution is incomplete: retry_on_index_lock's own wait still runs inside it, so up to LOCK_WAIT_S of waiting is silent|kipi-update.sh:2216
nit|The untrack staging `git rm --cached` is an index write with no wait before it; the wait sits on the commit two statements later|kipi-update.sh:1908
nit|A non-numeric KIPI_UPDATE_LOCK_WAIT_S makes the bound unreachable, turning the bounded wait into an unbounded spin|kipi-update.sh:1243
END FINDINGS
SessionEnd hook [python3 "${CLAUDE_PLUGIN_ROOT:-$PLUGIN_ROOT}/hooks/session-end-llma.py"] failed: Hook cancelled

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-314-20260906-173650.md` (7094 bytes).

```
FINDINGS:
major|restore_instance runs against the previous instance on the new write_run_marker refusal path and deletes its untracked in-scope files|kipi-update.sh:1804
minor|the retry loop's lock wait is still swallowed by the system-state commit's command substitution, the defect round 2 claimed to fix|kipi-update.sh:2216
minor|the untrack block's git rm --cached and git reset are index writes with no lock wait, so "before every index write" is false|kipi-update.sh:1909
END FINDINGS
```

--- reviewer output ---

Review done. Three findings survived reproduction.

## Finding 1

**Severity: major**

**Claim:** The new `abandon_instance` call site fires before `checkpoint_instance` has run for the instance being refused, so `restore_instance` runs against the PREVIOUS instance and deletes its untracked in-scope files.

**File:** `kipi-update.sh:1803-1805` (the defect is in `restore_instance`, `kipi-update.sh:811-816`, which reads `CHECKPOINT_TARGET` from the last instance that checkpointed)

**Why it fires:** `checkpoint_instance` sets `CHECKPOINT_TARGET` at `kipi-update.sh:808` and nothing resets it at the top of the loop. `write_run_marker` is called at line 1803, before the checkpoint at line 2263. So refusing instance B restores instance A. `checkpoint_untracked_list` uses `git ls-files --others` with no `--exclude-standard`, so the deletion loop reaches gitignored files too, which have no git copy.

**Reproducer:** functions extracted verbatim from `kipi-update.sh` (same technique the PR's own retry test uses), driven through the real sequence.

Real output:

```
CHECKPOINT_TARGET after instance A = /var/folders/.../pr314-stale-lj8bzr_9/instA
before: heartbeat.json
CHECKPOINT_TARGET while refusing instance B = /var/folders/.../instA  <-- not B
after:  RESULT: DELETED from healthy instance A because instance B was refused
```

Script at `/var/folders/th/ty0kb7r56dvfrkgkrcbx9fv00000gn/T/pr314-stale-lj8bzr_9/case.sh`.

**3am blast radius:** two overlapping `kipi update` runs (the 2026-09-06 incident in the PR body was three founder runs). Instance B is correctly refused by the new mutex, and instance A silently loses whatever a heartbeat or Stop hook wrote into `q-system/`, `.claude/` or `plugins/` since A's checkpoint. Untracked files have no git copy, so that part is not recoverable.

**Honest scoping:** the class pre-exists. The dirty-tree-guard abandon at `kipi-update.sh:2255` is also before the checkpoint, so this fires today on ordinary runs. This PR adds a new trigger for it. One-line fix: `CHECKPOINT_TARGET=""` at the top of the iteration, or move `write_run_marker` below `checkpoint_instance`.

## Finding 2

**Severity: minor**

**Claim:** The retry path's lock wait is still captured by the command substitution, which is the exact defect round 2's comment claims to have closed.

**File:** `kipi-update.sh:2216` (comment claiming the fix is at `kipi-update.sh:2210-2212`)

**Why:** round 2 hoisted the first wait out of `$( )`. `retry_on_index_lock` calls `wait_for_index_lock` again inside its loop at `kipi-update.sh:1266`, and that progress goes to stdout, which `sys_commit_err="$( ... 2>&1)"` captures and discards on success. At the default bound that is up to about 240 seconds of silence across two retries.

**Reproducer output:**

```
=== A) bare call (q-system sync commit :2480, config sync commit :2698) ===
  commit at system-state commit hit a live index.lock (attempt 1 of 3); waiting and retrying
  waiting for index.lock at system-state commit (0s of 8s)
=== B) the :2216 call site: inside a command substitution with 2>&1 ===
--- operator saw nothing for 4s between the two === lines ---
captured then discarded on success:
      commit at system-state commit hit a live index.lock (attempt 1 of 3); waiting and retrying
      waiting for index.lock at system-state commit (0s of 8s)
```

Script at `/var/folders/th/ty0kb7r56dvfrkgkrcbx9fv00000gn/T/pr314-retry-u67y0wvy/case.sh`.

## Finding 3

**Severity: minor**

**Claim:** The PR body says the wait is "called before every index write the run makes". It is not: the `SYSTEM_NEVER_COMMIT` untrack block writes the index three times with no wait, and only its commit got one.

**File:** `kipi-update.sh:1909` (`git rm --cached`), `kipi-update.sh:1917` and `kipi-update.sh:1929` (`git reset`). The wait sits at `kipi-update.sh:1918`, after the first write.

**Reproducer output:**

```
-- git rm --cached under a live lock (no wait at :1909) --
fatal: Unable to create '.../repo/.git/index.lock': File exists.
Another git process seems to be running in this repository, or the lock file may be stale
rc=128 -> kipi-update.sh:1911 prints: WARNING: could not untrack <path>
```

**Effect:** the instance-local untrack is skipped under exactly the condition this PR exists to handle, and the run reports a generic warning rather than the lock. Bounded and self-correcting on the next run. The same reasoning the round-2 commit applied at `kipi-update.sh:2675-2677` ("the untrack below is this block's FIRST index write") was not applied here.

## What is sound

- The mutation case is real. I ran the full suite: 7 of 7 pass, including `assert_without_the_wait_the_green_case_goes_red`, which stubs `wait_for_index_lock` and does go red. The green case is not passing on timing.
- The hook test's no-marker control is a genuine floor. It asserts `head_count == before + 1`, so "no commit" under a live marker is the marker's doing and not the classifier's. 6 of 6 pass.
- The foreign-marker mutex holds. `run_marker_is_live_foreign` refuses a live foreign pid, `clear_run_marker` compares the file's first field to `$$` before removing, and the test asserts the foreign marker survives the refused run.
- `abandon_instance` always returns 0 (`kipi-update.sh:975`), so `restore_instance` returning 1 on a lock timeout does not break the `abandon_instance ... && continue` chain and let a refused instance keep running. I checked this specifically because `set -euo pipefail` is on.
- No marker leaks across instances in one run. Every failure path routes through `abandon_instance`, success clears at `kipi-update.sh:2714`, and the EXIT trap clears via `cleanup_updater_temps`.
- I considered flagging the 120 second default bound against the 445 second lock hold the PR body measures. Dropped it: that 445 second hold was the auto-commit pre-commit verify, and the hook guard in this same PR stops that hook from running mid-sync, so the producer is gone.
- I considered `lessons-daily.sh:81` as a second unguarded committer. Dropped it: it commits in the skeleton, not in an instance, and the updater writes no instance index there. No harm reproduced.

## Verdict

**REQUEST CHANGES.** Not BLOCK. The damage in finding 1 is bounded to one instance's untracked in-scope files and the class already exists on main, so merging does not make anything newly unrecoverable at fleet scale.

Fix first: reset `CHECKPOINT_TARGET` per iteration, or move `write_run_marker` below `checkpoint_instance`, so the new refusal path cannot restore the previous instance.

FINDINGS:
major|restore_instance runs against the previous instance on the new write_run_marker refusal path and deletes its untracked in-scope files|kipi-update.sh:1804
minor|the retry loop's lock wait is still swallowed by the system-state commit's command substitution, the defect round 2 claimed to fix|kipi-update.sh:2216
minor|the untrack block's git rm --cached and git reset are index writes with no lock wait, so "before every index write" is false|kipi-update.sh:1909
END FINDINGS

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-314-20260906-180149.md` (5799 bytes).

```
FINDINGS:
major|The new registered test flakes ~9% of runs because `git log | grep -q` under pipefail returns non-zero on SIGPIPE despite a match|q-system/.q-system/scripts/test/test-kipi-update-lock-wait.sh:116
minor|The 120s lock bound is below the 445s hold this PR itself cites, and exceeding it leaves the instance tree dirty for manual repair|kipi-update.sh:1225
END FINDINGS
```

--- reviewer output ---

Verified: both new tests are registered in the capability manifest.

```
q-system/.q-system/capability/expected_tests/q-system__.q-system__scripts__test__test-kipi-update-lock-wait.sh.json
q-system/.q-system/capability/expected_tests/q-system__hooks__test__test_auto_commit_run_marker.py.json
```

Contents of the first:

```json
{
 "path": "q-system/.q-system/scripts/test/test-kipi-update-lock-wait.sh",
 "runner": "bash"
}
```

# Review: PR #314 (fleet updater index.lock wait + auto-commit run marker)

## Finding 1 (major): the new registered test is flaky, roughly 1 run in 10

`q-system/.q-system/scripts/test/test-kipi-update-lock-wait.sh:116`

```bash
G -C "$inst" log --format=%s | grep -q "sync q-system from skeleton" || fail ...
```

The file runs under `set -euo pipefail`. `grep -q` exits the moment it matches. In this fixture the matching subject is the SECOND of three lines, so git can still be writing when the read end closes, dies on SIGPIPE with 141, and pipefail turns a successful match into a failed pipeline.

Reproducer, the registered test run 43 times:

```
bash q-system/.q-system/scripts/test/test-kipi-update-lock-wait.sh
```

Real output, 4 failures across 43 runs:

```
run 6 rc 1
   FAIL: the q-system sync commit is not in the instance history: chore: sync .claude config + plugins from skeleton 2026-09-06 [no-issue: fleet updater skeleton sync]
run 7 rc 1
   FAIL: a restore that could not run was silent:   Syncing q-system/ from skeleton...
failures 2 of 15
```

The failure message prints the very line the grep was looking for. That is the tell: the match was present and the pipeline still returned non-zero.

Mechanism proven directly, 300-commit repo, match on the first line git writes:

```
sanity: grep -c -> 1 (must be 1)
under set -o pipefail, the pipeline returned NONZERO in 100 of 100 runs while the subject is present
```

Line 162 (`restore could NOT run`) flakes too, on a different and undiagnosed timing path in the `hold_lock` handshake.

Why major: this file is registered in the capability manifest, quoted above, so the fleet verification run goes red about once every ten unattended executions with no lock regression behind it. A checker that cries wolf is the alert nobody reads.

Fix for line 116: write the log to a file and grep the file, or use `grep -c` with a count assertion. No pipe from git into an early-exiting reader.

## Finding 2 (minor): a lock past the bound leaves the instance dirty, and the bound is below the only hold this repo has ever measured

`kipi-update.sh:1225` sets `LOCK_WAIT_S` default 120. The PR's own description records the holder in the motivating incident at 445 seconds (the peer session's pre-commit verify, sp-9306036e). 120 is 3.7x under the one number on record.

When the bound is exceeded, `restore_instance` returns early by design and the tree keeps this run's writes. I ran the PR's own never-clears case with a status dump appended:

```
--- instance tree after the refusal (lock already removed): ---
 M q-system/tracked.md
?? q-system/.q-system/
?? q-system/hooks/
--- HEAD: inst ---
```

That ` M q-system/tracked.md` is exactly the shape the dirty-tree guard reads as founder work, so every later run refuses the instance until someone restores it by hand. The PR is honest about this in its error text, and loud beats the old silent version. It is still an instance out of the fleet at 3am needing manual git surgery.

Two options: raise the default past the measured 445, or have the refusal path re-attempt the restore once the lock clears instead of stopping at one bounded wait.

## What is sound

- Marker leak on an instance with no config directory. I rebuilt the PR's fixture without `$inst/.claude` expecting `clear_run_marker` (inside the config-sync block) to be skipped. It still cleaned up: `PASS: a live index.lock is waited out, the step is named, and the sync lands`, rc 0. Dropped.
- `restore_instance` can now `return 1`, and the script is `set -euo pipefail`. All 27 `abandon_instance` call sites are `abandon_instance ... && continue`, which suppresses errexit through the function body. A refused restore cannot kill the fleet run mid-loop.
- The retry's discrimination is real: a pathspec error returns rc=1 on attempt 1 with its stderr intact, and only the lock text retries to the cap of 3.
- The mutation case is load-bearing. Stubbing `wait_for_index_lock` to a no-op makes the green case red, so property 1 is not passing on timing.
- The hook half, 6 of 6 green, including the control that proves the live case is the marker's doing: with no marker the same dirty file DOES get committed. The foreign-cwd case is a genuine fail-open catch.
- Regression file `test-kipi-update-system-state-commit.sh` green on the edited updater.

One thing I could not run, so I am not reporting it: at `kipi-update.sh:1931`, when the `untrack commit` wait fails, the else branch runs `git reset ... || true`, a command needing the lock just proven held. No reproducer built inside this session's sandbox, so it stays out.

## VERDICT: REQUEST CHANGES

Not a block. Nothing here is permanent or unrecoverable, and the marker plus wait design is a real improvement on the incident it targets.

First thing to fix: the `git log | grep -q` at `test-kipi-update-lock-wait.sh:116`. A registered test that goes red one run in ten is worse than no test.

FINDINGS:
major|The new registered test flakes ~9% of runs because `git log | grep -q` under pipefail returns non-zero on SIGPIPE despite a match|q-system/.q-system/scripts/test/test-kipi-update-lock-wait.sh:116
minor|The 120s lock bound is below the 445s hold this PR itself cites, and exceeding it leaves the instance tree dirty for manual repair|kipi-update.sh:1225
END FINDINGS

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-314-20260906-183512.md` (5182 bytes).

```
FINDINGS:
minor|The system-state `git add` is an index write with no lock wait and a silenced failure, so a momentary lock loses the staging and the wrapped commit dies with a non-lock error the retry will not retry|kipi-update.sh:2204
nit|`run_marker_is_live_foreign` reads a live process owned by another uid as dead and overwrites its marker, while the Python half counts the same case live|kipi-update.sh:1310
END FINDINGS
```

--- reviewer output ---

Review done. Read-only, both new tests run, three attacks reproduced, two dropped.

## Findings

**minor** — the system-state block's `git add` is an index write with no `wait_for_index_lock`, and its failure is fully silenced, so a momentary lock loses the staging and the wrapped commit then dies with an error the retry deliberately will not retry.
`kipi-update.sh:2204`

Reproducer (a scratch repo, the same producer the PR body names: a peer session's `git status` holding the lock for a fraction of a second):

```
git add -- "${sys_add_paths[@]}" 2>/dev/null || true   # verbatim from 2204, with .git/index.lock present
git commit -q -m msg -- "${sys_add_paths[@]}"          # after the peer lets go
```

Real output:

```
--- kipi-update.sh:2204, verbatim ---
exit as the script sees it: 0   (|| true, stderr to /dev/null)
staged after the add: []
--- the peer let go; wait_for_index_lock now returns 0 and the commit runs ---
commit rc=1
commit stderr: error: pathspec 'q-system/.q-system/state/integrity.json' did not match any file(s) known to git
VERDICT: retry_on_index_lock does NOT retry -> returns on attempt 1 -> WARNING + instance refused below
HEAD count: 1
```

Why it bites: `sys_owned_dirty` is fed by `git status --porcelain -uall` (line 1990), so untracked system files do reach `sys_add_paths`, and this repo already has a test pinning that behavior. `test-kipi-update-system-state-commit.sh` prints `PASS: one untracked path makes git commit -- <pathspec> commit nothing at all`. The `add` is load-bearing and it is the one index write in the run with neither a wait nor a visible failure. The PR body's line "Called before every index write the run makes" is not true as written.

**nit** — `run_marker_is_live_foreign` reads a live process owned by another uid as dead and overwrites its marker, while the Python half of the same handshake explicitly counts it live.
`kipi-update.sh:1310` (the `kill -0 "$pid" 2>/dev/null || return 1` line) versus `q-system/hooks/auto-commit.py:457`

Reproducer and real output:

```
bash kill -0 1 => FALSE (a live root process reads as dead)
python os.kill(1,0) => PermissionError (hook counts it LIVE)
```

Dropped a level on purpose: no producer in this fleet runs `kipi update` as a second uid, so I had to construct the input.

## What is sound

- **Both new tests are real, not decoration.** `test-kipi-update-lock-wait.sh` 7/7 in 16 s, `test_auto_commit_run_marker.py` 6/6. The mutant case stubs `wait_for_index_lock` to a no-op and the green case goes red, and the hook test carries a no-marker control proving the live case is the marker's doing and not the classifier's. That is the property most of this fleet's green-but-wrong tests have been missing.
- **Marker leak across the instance loop.** I traced every `continue` between `write_run_marker` (1818) and `clear_run_marker` (2736). All instance-loop exits are `abandon_instance … && continue`, which clears; the one at 1926 is an inner-loop `continue`. No leak.
- **`abandon_instance … && continue` under `set -euo pipefail`.** `restore_instance` can now return 1, and a bare failing function under errexit would kill the run. It survives: `abandon_instance` ends `return 0`, and being the left arm of an AND-list suppresses errexit inside it. The never-clears test lands on `Failed:  1` rather than a dead shell.
- **Marker ownership.** `clear_run_marker` compares the first field to `$$` before deleting, so a second run's marker is not stripped by the first run's exit path. Confirmed by the foreign-marker case.
- **The staging paths get the wait but no retry**, so the wait's check-then-act gap can still reach `abandon_instance`. I could not build a repro that fired it, and the restore now waits and leaves the tree clean, so re-running works. Dropped.
- **The unconditional stale-lock deletion seven lines above the marker write** (1811-1817) still deletes a live `index.lock`, which is the exact thing the new comment calls "how two writers corrupt one index." Pre-existing, already captured as sp-2c1bcc3f, and declared out of scope in the PR body. Not counted against this PR, but it is the largest residual risk in the file and the wait cannot cover it.

## VERDICT

**APPROVE WITH NITS.**

First thing to fix: put `wait_for_index_lock "$path" "system-state staging"` in front of the `git add` at 2204 and stop discarding its stderr, so an untracked system file cannot be dropped into a commit that then fails for a reason the retry ignores.

One housekeeping note: the sandbox would not let me delete my repro script, so `.review-scratch-repro.sh` is sitting untracked in the review worktree at `/Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-314/`. Nothing in the repo reads it.

FINDINGS:
minor|The system-state `git add` is an index write with no lock wait and a silenced failure, so a momentary lock loses the staging and the wrapped commit dies with a non-lock error the retry will not retry|kipi-update.sh:2204
nit|`run_marker_is_live_foreign` reads a live process owned by another uid as dead and overwrites its marker, while the Python half counts the same case live|kipi-update.sh:1310
END FINDINGS

PR #315, opened 2026-09-06
fix(voice-lint): drop two words the founder demonstrably writes from the ban list

assafkip:
## What

Removes `robust` and `foster` from `BANNED_WORDS` in `q-system/.q-system/scripts/voice-lint.py` (48 to 46) and leaves a calibration note beside the list.

## Why (measured 2026-09-06)

The one instance that holds the founder's voice corpus measured 7 and 16 organic uses of these words in his own posts, removed both on 2026-08-27 (its PR #71), and pinned the list's size to the measured value. The skeleton kept its 2026-05-21 list, so today's fleet sync delivered the old words back and that instance's pinned test went red, blocking its merge of main. Third instance-ahead-of-skeleton hit in one day, same class (sp-1ad08728). A ban list that blocks his real vocabulary is the six-word scar again.

## Verification (ran, not assumed)

- `BANNED_WORDS` is 46, neither word present, `leverage` still present.
- His own sentence ("it fosters a broad discussion and a robust understanding of the attackers") at an in-scope marketing path: no banned-word finding on this branch; main's lint flags `robust` on the same file.
- A slop fixture ("we leverage synergy to empower teams") still produces banned-word findings.
- `test-voice-lint-topups.sh` and `test-voice-lint-warn-detectors.sh` pass.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_01DCaQcw6jqEpYyqyAznCHg9

PR #316, opened 2026-09-06
feat(updater): report every file a sync would overwrite whose instance copy the skeleton never shipped, with --refuse-instance-ahead (sp-1ad08728)

assafkip:
## What broke

Three times on 2026-09-06 the fleet sync replaced a file the instance had improved past the skeleton, and each time the instance's own gate went red only after the rsync had landed:

- the voiceloop engine (extra modules, ported later by PR #311)
- `voice-stop-gate.py` with `founder_typed_text` and the route-receipt family (port on PR #313)
- voice-lint's calibrated `BANNED_WORDS`, two entries removed after measuring the founder's corpus (PR #315)

The dry run listed those files as ordinary changes. Nothing in the updater said "this instance carries content the skeleton never shipped for a path I am about to overwrite" (sp-1ad08728).

## What changed

`kipi-update.sh`
- `instance_ahead_scan`: for every existing-file update in the same itemized rsync preview the run uses (same excludes), the instance's tracked copy is "ahead" when `fleet_authored_blob` says the skeleton never shipped those bytes for that path. Bytes equal to any skeleton version are never ahead, so a stale instance and a hook that committed the updater's own bytes stay quiet. Language-agnostic by construction. One line per hit: `+defs: name` for `.py`, hunk count otherwise.
- `instance_ahead_preflight`: the q-system prefix (from `git archive HEAD`, as the sync uses), each managed plugin (working tree, as the plugins rsync uses), the three `.claude` md dirs. Runs before the pre-sync commit in every mode, so `--dry-run`, the fleet dry sweep's model run and the apply print the same lines.
- `--refuse-instance-ahead`: abandon the instance on any hit before the pre-sync commit, nothing staged, nothing written. Off by default (fleet-wide most hits would be hook churn); the consulting run script opts in.
- The summary repeats every line beside Updated/Failed, because a 1,100-line dry log hid `rewritten=8` for a morning.

`kipi-update-instance-ahead.py`: pure describe helper (ast def/class names only in the instance copy, else unified-diff hunk count).

## Verification

- `test-kipi-update-instance-ahead.sh` (registered): a skeleton with two commits so the older-shipped-version case exists. Dry run names the four ahead files (extra def, edited list, edited plugin file, edited rule) and not the file sitting at the skeleton's older version; the summary repeats them with the count. `--refuse-instance-ahead` refuses with HEAD, index and working tree untouched. Without the flag the sync lands and the report still prints. A mutant that makes `fleet_authored_blob` always succeed silences the report. 4/4.
- `test-kipi-update-preservation-failure.sh`: its rsync fake now ignores a dry itemized invocation (`-n`), which the preflight makes on purpose before the preservation helper; a writing rsync still counts.
- Other `test-kipi-update-*.sh`: 10/10; the three updater python tests pass.

## Cost

One `fleet_authored_blob` history walk per file the rsync would update and the instance already has; tens of files on a typical instance.

https://claude.ai/code/session_01KLwJz1CRif347aHf9xx5AU

### DIFF
--- a/kipi-update-instance-ahead.py
+++ b/kipi-update-instance-ahead.py
@@ -1,0 +1,71 @@
+#!/usr/bin/env python3
+"""Describe how an instance's copy of a synced file is ahead of the skeleton's.
+
+Called by kipi-update.sh's instance_ahead_scan for one file the rsync would
+overwrite whose instance bytes are not any blob the skeleton ever shipped for
+that path. The scan decides WHETHER a file is ahead (fleet_authored_blob, git
+history); this script only says HOW, for the log line:
+
+    .py    -> "+defs: name, name" (defs and classes the instance has and the
+              skeleton lacks) or, when the names are identical, the hunk count
+    other  -> "N hunks" from a unified diff
+
+Pure: no git, no writes. On 2026-09-06 the three instance-ahead hits were an
+engine (extra modules), a gate script (five extra defs, founder_typed_text
+among them) and a calibrated word list (two removed entries): the first two
+are the def line, the third is the hunk count, which is why both exist.
+
+usage: kipi-update-instance-ahead.py <instance_file> <skeleton_file>
+"""
+import ast
+import difflib
+import sys
+
+
+def top_level_names(text):
+    try:
+        tree = ast.parse(text)
+    except SyntaxError:
+        return None
+    names = set()
+    for node in tree.body:
+        if isinstance(node, (ast.FunctionDef, ast.AsyncFunctionDef, ast.ClassDef)):
+            names.add(node.name)
+    return names
+
+
+def hunk_count(instance_text, skeleton_text):
+    diff = difflib.unified_diff(
+        skeleton_text.splitlines(keepends=True),
+        instance_text.splitlines(keepends=True),
+        n=0,
+    )
+    return sum(1 for line in diff if line.startswith("@@"))
+
+
+def describe(instance_path, skeleton_path):
+    with open(instance_path, encoding="utf-8", errors="replace") as fh:
+        instance_text = fh.read()
+    with open(skeleton_path, encoding="utf-8", errors="replace") as fh:
+        skeleton_text = fh.read()
+    hunks = hunk_count(instance_text, skeleton_text)
+    if instance_path.endswith(".py"):
+        mine = top_level_names(instance_text)
+        theirs = top_level_names(skeleton_text)
+        if mine is not None and theirs is not None:
+            extra = sorted(mine - theirs)
+            if extra:
+                return "+defs: " + ", ".join(extra)
+    return f"{hunks} hunk{'' if hunks == 1 else 's'}"
+
+
+def main(argv):
+    if len(argv) != 3:
+        print(__doc__, file=sys.stderr)
+        return 2
+    print(describe(argv[1], argv[2]))
+    return 0
+
+
+if __name__ == "__main__":
+    sys.exit(main(sys.argv))
--- a/kipi-update.sh
+++ b/kipi-update.sh
@@ -21,9 +21,17 @@ SKELETON_BRANCH="main"
 # staged rollout is the only safe way to ship anything with this blast radius.
 DRY_RUN=""
 ONLY=""
+# --refuse-instance-ahead: abandon an instance BEFORE its pre-sync commit when
+# the run would overwrite a file whose instance copy the skeleton never shipped
+# (see instance_ahead_scan). Off by default: fleet-wide the hit is usually
+# hook churn, and a refusal that fires on 24 instances gets switched off. The
+# consulting run script passes it, because there "ahead" has meant real work
+# every time (2026-09-06, three times in one day).
+REFUSE_INSTANCE_AHEAD=0
 while [ $# -gt 0 ]; do
   case "$1" in
     --dry-run) DRY_RUN="--dry-run"; shift ;;
+    --refuse-instance-ahead) REFUSE_INSTANCE_AHEAD=1; shift ;;
     --only)
       ONLY="${2:-}"
       if [ -z "$ONLY" ]; then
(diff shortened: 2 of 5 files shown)


assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-316-20260906-162702.md` (5509 bytes).

```
FINDINGS:
minor|--refuse-instance-ahead reports "nothing staged, nothing written" after the gitignore managed-block writer has already created the instance's root .gitignore|kipi-update.sh:1895
minor|The refusal test's "nothing written" assertions cannot fail: the fixture omits kipi-update-gitignore-block.py and git diff is blind to the untracked file it writes|q-system/.q-system/scripts/test/test-kipi-update-instance-ahead.sh:119
minor|The fake rsync's --dry-run alternative is unreachable behind the double-dash guard, so a --dry-run invocation counts as a writing rsync against the comment above it|q-system/.q-system/scripts/test/test-kipi-update-preservation-failure.sh:132
nit|The summary says "a sync overwrites them" identically on --dry-run and on a refused instance, where no sync ran|kipi-update.sh:2738
END FINDINGS
```

--- reviewer output ---

# Review of PR #316

Three findings survived reproduction. All minor. Nothing here is unrecoverable and the flag is off by default, so blast radius is bounded.

## MINOR 1: the refusal says "nothing written" one line after writing

`kipi-update.sh:1895`

`instance_ahead_preflight` runs at line 1893, but `kipi-update-gitignore-block.py` already wrote the instance's root `.gitignore` at line 1806. The two lines print next to each other in the same run.

Reproducer: fixture skeleton carrying the gitignore writer plus one instance-ahead `.py`, run with `--refuse-instance-ahead`.

```
=== instance BEFORE the refusing run (empty = clean) ===
=== what the operator sees ===
  .gitignore managed block added (5 path(s))
  instance ahead: q-system/.q-system/scripts/gate.py (+defs: founder_only)
  ERROR: instance is ahead of the skeleton on 1 file(s); refusing
         (--refuse-instance-ahead), nothing staged, nothing written
=== instance AFTER the refusing run ===
?? .gitignore
=== instance root .gitignore ===
PRESENT after a run that reported nothing written:
# >>> kipi-managed: instance-local, never commit >>>
...
```

The write itself is idempotent and wanted. The message is what is wrong, and the founder is being told an instance was untouched when it was not.

## MINOR 2: the "nothing written" assertion cannot fail

`q-system/.q-system/scripts/test/test-kipi-update-instance-ahead.sh:30` and `:119-120`

Two independent reasons the check is decoration. The fixture's copy list at line 30 never copies `kipi-update-gitignore-block.py` into the skeleton, so the write cannot happen there. And `git diff --quiet` plus `git diff --cached --quiet` are both blind to an untracked file, which is exactly what that write produces.

Same reproducer as finding 1. Adding the writer to the fixture makes the instance dirty in a way those two assertions still report clean. `git status --porcelain -uall` is the check that can go red.

## MINOR 3: the `--dry-run` arm of the fake rsync is unreachable

`q-system/.q-system/scripts/test/test-kipi-update-preservation-failure.sh:132`

The comment above says a dry itemized rsync, `-n` or `--dry-run`, is not counted. The `[ "${arg#--}" = "$arg" ]` guard excludes every double-dash argument, so the `--dry-run` alternative in the case pattern can never fire.

Reproducer: extract the fake body, run it with each arg shape.

```
--- fake rsync body under test:
for arg in "$@"; do
  case "$arg" in
    -n|--dry-run|-*n*) [ "${arg#--}" = "$arg" ] && exit 0 ;;
  esac
done
printf 'rsync invoked\n' >> "$RSYNC_LOG"
---
args [-ain ...]      -> treated as dry, NOT counted
args [--dry-run ...] -> COUNTED as a writing rsync
args [-a ...]        -> COUNTED as a writing rsync
```

Inert today, since no call site passes `--dry-run` to rsync. It becomes a confusing false FAIL the day one does.

## NIT

`kipi-update.sh:2738`. The summary reads "a sync overwrites them" identically on a plain `--dry-run` and on an instance that was just refused, where no sync ran.

## What is sound

Four attacks the code survived, all executed:

- **Untracked founder content at a synced path.** The scan skips untracked files, and its comment at line 354 names the wrong guard for them. But `is_instance_wip` does catch it: my run printed `ERROR: untracked WIP collides with skeleton path` and `Failed: 1`, and the founder bytes were still on disk afterwards. Real protection, wrong comment. Not worth a round.
- **Cost.** I expected the q-system candidate set to be the whole tree, because the source is a fresh `git archive` extraction and rsync's quick check is size plus mtime. Measured: `git archive` stamps commit mtimes, so two extractions of the same HEAD are mtime-identical and `rsync -ain` between them reports 0 files. The "tens of files" claim holds. Finding dropped.
- **Blindness in the fleet dry sweep.** The model rsync excludes `.git`, which would make `git ls-files` fail for every candidate and silence the whole report. Line 1654 explicitly `cp -a`'s `.git` back, so the model has a real index and the scan works there.
- **Full updater suite.** All 12 `test-kipi-update-*.sh` pass on this branch, including the new one at 4/4 in 5.4 seconds.

The mutant in property 4 is real: forcing `fleet_authored_blob` to always report "shipped" silences the report, and the older-shipped-version file in the fixture keeps the check from degrading into a plain diff. That is the right pair.

## VERDICT: APPROVE WITH NITS

Fix first: the `nothing staged, nothing written` string at `kipi-update.sh:1895`. Either move `instance_ahead_preflight` above the gitignore block writer at line 1806, or reword to name what has already happened.

FINDINGS:
minor|--refuse-instance-ahead reports "nothing staged, nothing written" after the gitignore managed-block writer has already created the instance's root .gitignore|kipi-update.sh:1895
minor|The refusal test's "nothing written" assertions cannot fail: the fixture omits kipi-update-gitignore-block.py and git diff is blind to the untracked file it writes|q-system/.q-system/scripts/test/test-kipi-update-instance-ahead.sh:119
minor|The fake rsync's --dry-run alternative is unreachable behind the double-dash guard, so a --dry-run invocation counts as a writing rsync against the comment above it|q-system/.q-system/scripts/test/test-kipi-update-preservation-failure.sh:132
nit|The summary says "a sync overwrites them" identically on --dry-run and on a refused instance, where no sync ran|kipi-update.sh:2738
END FINDINGS

PR #317, opened 2026-09-07
test(updater): the propagation fixture carries kipi-update-instance-ahead.py (main regression from #316)

assafkip:
## What broke

PR #316 taught `kipi-update.sh` to invoke `kipi-update-instance-ahead.py` from its own directory. `q-system/.q-system/tests/separation/test_update_propagation.py` keeps `UPDATER_HELPERS`, the hand-kept list of such helpers, and its divergence test `test_the_fixture_carries_every_helper_the_updater_invokes` is the guard that list documents: a helper the updater invokes that the fixture never copies.

It went red on main the moment #316 merged (auto-merge requires `validate` and the reviewer status, not `verify`), and then on every PR's `verify` run since, because CI tests the head merged into main. PR #314's verify is the first casualty.

## What changed

One entry in `UPDATER_HELPERS`, with the note the list's own history asks for.

## Verification

`python3 -m pytest q-system/.q-system/tests/separation/test_update_propagation.py`: 6 passed. The pre-commit's `verify.sh --staged` ran the whole `q-system/.q-system/tests` suite green.

### DIFF
--- a/q-system/.q-system/tests/separation/test_update_propagation.py
+++ b/q-system/.q-system/tests/separation/test_update_propagation.py
@@ -47,6 +47,10 @@ UPDATER_HELPERS = (
     "kipi-update-gitignore-block.py",
     "kipi-update-wip-check.py",
     "kipi-update-voiceloop-migrate.py",
+    # The instance-ahead describe helper (PR #316). Arrived on main with the
+    # report and was not added here, so the divergence test below went red on
+    # every PR merged against that main: the exact trap this list documents.
+    "kipi-update-instance-ahead.py",
     "settings-template.json",
 )
 


assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-317-20260906-174158.md` (5566 bytes).

```
FINDINGS:
minor|UPDATER_HELPERS header claims the updater is fail-closed on each listed helper, but the new entry's call site swallows a missing file and continues|q-system/.q-system/tests/separation/test_update_propagation.py:32
minor|The helper derivation matches only top-level $SCRIPT_DIR files, so a nested fail-closed dependency (the ef62bd8 shape) is invisible to the divergence guard|q-system/.q-system/tests/separation/test_update_propagation.py:62
minor|The divergence guard runs only in verify, which is not a required check and not in the capability manifest validate executes, so the same regression can merge again|.github/workflows/validate.yml:51
END FINDINGS
```

--- reviewer output ---

## Review: PR #317

Four lines, one entry in `UPDATER_HELPERS`. I reproduced the regression it fixes and then attacked the guard it patches.

### The fix itself is real (verified, not taken on faith)

Reproduced red at the parent commit and green at the head, using the test's own derivation rather than its prose:

```
python3 - <<'PYEOF'   # derives from kipi-update.sh, compares to parent's tuple
...
PYEOF
```

Real output:

```
parent-listed: ['kipi-settings-merge.py', 'kipi-update-deletion-guard.py', 'kipi-update-gitignore-block.py', 'kipi-update-preserve-scan.py', 'kipi-update-voiceloop-migrate.py', 'kipi-update-wip-check.py', 'settings-template.json']
MISSING at parent (divergence test red): ['kipi-update-instance-ahead.py']
STALE at parent: []
```

`python3 -m pytest q-system/.q-system/tests/separation/test_update_propagation.py -q` at head: `6 passed in 4.22s`.

---

### Finding 1 (minor): the list header says "fail-closed on each"; this entry's call site is fail-open

`q-system/.q-system/tests/separation/test_update_propagation.py:32` states the admission criterion for the list: the updater "invokes them from its own directory and is fail-closed on each." `kipi-update.sh:358` is `detail="$(python3 "$SCRIPT_DIR/kipi-update-instance-ahead.py" ... 2>/dev/null)" || detail="differs"`. Missing helper, suppressed stderr, run continues.

Reproducer: build the fixture fleet with the repo's own helpers, delete the newly-added helper, run the updater.

```
python3 - <<'PYEOF'
... build_fleet(tmp, INJECTED_BODY); (skel/"kipi-update-instance-ahead.py").unlink(); run_updater(...)
PYEOF
```

Real output:

```
rc= 1
ahead lines: ['  instance ahead: q-system/marketing/templates/outreach.md (differs)']
propagated OK? True
stderr mentions the missing helper? False
```

The propagation model runs fine without it. The next person reading line 32 concludes a fail-open helper does not earn a list entry, and the divergence test goes red again. The entry is right; the stated reason is not.

### Finding 2 (minor): the derivation is blind to nested dependencies, the shape that already broke this file once

`updater_invoked_files()` at `q-system/.q-system/tests/separation/test_update_propagation.py:62` matches only `$SCRIPT_DIR/<name>.<ext>`. The docstring excludes directories because "the fixture builds those itself." It does not: `build_skeleton` hand-copies a hardcoded three-file list at line 266, with no divergence check on it. `.prd-os/issues/pff-all-propagation-entrypoints.md:45` records the leak gate landing exactly this way in ef62bd8 and leaving two cases red.

Mutation test of the derivation, both shapes appended to the real script text:

```
A new top-level  -> derived delta: ['kipi-update-brand-new.py']
B new nested     -> derived delta: []
```

A new fail-closed `$SCRIPT_DIR/q-system/.q-system/scripts/*.py` merges invisibly and the guard stays green.

### Finding 3 (minor): the guard runs only in a check that cannot block the merge

`test_update_propagation.py` appears in no CI runner by name, and is absent from `q-system/.q-system/capability/manifest.json`, which is what `validate` executes.

```
gh api repos/assafkip/kipi-system/branches/main/protection --jq '.required_status_checks.contexts'
["validate","kipi/reviewer-approved"]

grep -n "test_update_propagation\|separation" q-system/.q-system/capability/manifest.json
(no output)
```

So `verify` is where this test lives and `verify` gates nothing. That is precisely how #316 landed red. This PR cleans the instance; the third occurrence is still unblocked. Pre-existing, not introduced here.

---

### What is sound

- **Not dead weight.** I checked whether the copy only satisfies the divergence test. It does not: the helper now actually executes inside the fixture. Output line changed from `(differs)` to `(1 hunk)`, which is the helper's own format.
- **Not stale.** The derivation finds `kipi-update-instance-ahead.py` in the shipping script, so the reverse assertion passes for a real reason.
- **Hermeticity holds.** `test_no_production_instance_is_reachable_from_the_fixtures` still passes with the new code path active; every fixture git dir resolved under `tmp_path` and the production registry bytes were unchanged.
- **No fan-out risk.** Helpers land at the fixture skeleton root, outside the `q-system/`, `.claude/`, `plugins/` propagation channels, so the clean-control test is not weakened.
- **Blast radius is a test fixture.** Nothing here touches an instance, a Linear object, or a scheduled job. Rollback is a revert.

Most important thing to fix first: Finding 3. The guard that this PR is repairing for the second time cannot block the merge it exists to block, so add `test_update_propagation.py` to the capability manifest that `validate` runs, or make `verify` required. Findings 1 and 2 are the same theme one level down.

FINDINGS:
minor|UPDATER_HELPERS header claims the updater is fail-closed on each listed helper, but the new entry's call site swallows a missing file and continues|q-system/.q-system/tests/separation/test_update_propagation.py:32
minor|The helper derivation matches only top-level $SCRIPT_DIR files, so a nested fail-closed dependency (the ef62bd8 shape) is invisible to the divergence guard|q-system/.q-system/tests/separation/test_update_propagation.py:62
minor|The divergence guard runs only in verify, which is not a required check and not in the capability manifest validate executes, so the same regression can merge again|.github/workflows/validate.yml:51
END FINDINGS

PR #321, opened 2026-09-07
fix(auto-commit): refuse while another commit holds git's locks

assafkip:
## The writer nobody could negotiate with

`fleet_update_in_progress` already coordinates with the one writer that leaves a marker. This hook fires at **every turn end in every session** on a shared checkout and took no lock at all, so two sessions can hold a ref window perfectly and still lose a commit to it.

Measured 2026-09-06 in a single session: **three of five git operations died** on `Unable to create '.git/index.lock'`, and **every one exited 0**. A commit that reports success while landing nothing is worse than one that refuses, because the next session reads the log and believes it.

## We read the locks, we do not remove them

`kipi-update.sh:1714` force-deletes `HEAD.lock`/`index.lock` with no pid, mtime or age test at all (`sp-50119dec`). Removing a lock that a live 8-minute pre-commit still owns corrupts that commit's index. Refusing costs one skipped auto-commit; the next turn end picks the work up.

**No age bound, deliberately**, unlike the run marker beside it. A commit here holds `index.lock` for its whole pre-commit — measured at **447–497s** on the consulting checkout — so a bound low enough to catch a crashed lock would fire constantly against healthy ones. A genuinely orphaned lock is a human's call: *"no process holds it AND every live hook's cwd is a worktree"* is the test that settled one by hand tonight, and it needs `lsof`, which a Stop hook must not spend.

## The part where I was wrong, and the test caught it

The `git diff --cached --name-only` read is removed as **redundancy, and the comment says plainly that this is all it is.**

It was believed — by the original spillover filer *and* by the triage that reviewed it — to be the line that swept another session's staged work into this hook's commit (`sp-78728ff1`). Measured before removing it, that is **false**: `git diff --name-only HEAD` one line above already reports a staged-but-uncommitted new file. In a scratch repo, `git add b` on a new file makes `git diff --name-only HEAD` print `b`.

I wrote the test for the claimed fix first, watched it go RED with the line removed, and **deleted the test rather than keep one built on a false premise.**

`sp-78728ff1` stays open, corrected as `sp-8d1e8836`. Git cannot tell *who* staged a path, so the real lever is to treat staged as claimed and skip it — which trades away part of the safety net and is a design call, not a line deletion.

## Verification (updated after review, 2026-09-07)

`test_auto_commit.py` **46 passed** (37 before, 9 new). The original body said 41 and 4;
review added five more tests across four rounds.

Mutation controls, each killing on its own message:

| Mutant | Result |
|---|---|
| disable the guard (`held = None`) | 3 tests RED |
| make the guard DELETE the lock it finds | the do-not-delete control RED |
| check `index.lock` only, drop `HEAD.lock` | the HEAD.lock test RED |
| `--git-common-dir` in place of `--git-dir` | both worktree tests RED |
| resolve the git dir against the process cwd | the PROJ_DIR pin RED |
| move the lock refusal back to stderr | the fleet-channel test RED |
| move the fleet-updater refusal back to stderr | its sibling test RED |

The last four mutants SURVIVED the original 41 tests and are what review added.
Gates run green on the branch: `verify.sh --full` 12/12, `test_verify_adversarial`
rc=0, `validate-separation.py 1 --verbose` rc=0, `capability-gate.py` GREEN (231
tests).

## What review changed

1. **`--git-dir` is correct and is now pinned.** Measured: `index.lock` and
   `HEAD.lock` are per-worktree. Holding `<main>/.git/index.lock` does not block a
   `git add` from a linked worktree (rc=0); only that worktree's own lock does
   (rc=128). The common-dir spelling would have silenced the safety net in every
   linked worktree for the ~450s a main-checkout pre-commit runs, and it passed
   all 41 tests.
2. **The resolution base was unpinned** (review round 1, minor). Resolving the
   git-dir answer against the process cwd instead of `PROJ_DIR` fails the guard
   OPEN whenever the two differ, and also passed all 43 tests. Same defect class
   as PR #314 round 2, one guard up in the same file.
3. **Three causal sentences were corrected against measurement**, not reasoning:
   the `--cached` read is not universally redundant (two staging shapes only it
   sees; removing it is still right because a pathspec commit is a no-op on both);
   `kipi-update.sh:1714` is unrelated code, the sweep is ~L1915 and that same file
   already solves this properly at ~L1313-1400 by waiting; and the no-age-bound
   trade was over-charged, since git already exits 128 on a held lock and this
   hook already turned that into a `skipped:` line at exit 0.

### DIFF
--- a/.github/workflows/verify.yml
+++ b/.github/workflows/verify.yml
@@ -13,6 +13,26 @@ jobs:
   verify:
     runs-on: ubuntu-latest
     steps:
+      # THE SAME MISSING IDENTITY validate.yml carries a scar for, in the other
+      # door. verify.sh runs q-system/hooks/test/test_auto_commit_run_marker.py,
+      # whose subject is a SUBPROCESS: run_hook spawns auto-commit.py, which runs
+      # its own `git commit`, and that child inherits the runner's identity, not
+      # the `git -c user.email=...` the test passes to its own commands. The
+      # ubuntu runner user has an empty gecos field so git cannot guess one and
+      # refuses with "Author identity unknown"; the failure then reads "the hook
+      # did not commit", as if the hook were broken. Measured 2026-09-07 from run
+      # 34123764623, the first run after that suite was named in .verify-suites.
+      # macOS cannot reproduce it -- its git guesses from the passwd gecos field,
+      # which is the same reason validate.yml's copy of this note exists.
+      #
+      # The test was ALSO made hermetic in the same change (it writes the
+      # identity into its own temp repo), so this step is the second of two
+      # independent fixes, not the only one holding it up.
+      - name: Configure git identity for tests that spawn git
+        run: |
+          git config --global user.name "kipi-ci"
+          git config --global user.email "ci@kipi.invalid"
+
       - uses: actions/checkout@v4
         with:
           # FULL HISTORY, not the default depth-1 shallow clone.
--- a/.verify-suites
+++ b/.verify-suites
@@ -80,6 +80,18 @@ rescued/memory-lifecycle
 q-system/.q-system/tests
 test_destructive_op_deny_anchor.py
 test_fleet_unblock.py
+# ROUND 5, and the same finding for the fifth time: a tracked, green, wholly
+# ungated suite. q-system/hooks/test/test_auto_commit_run_marker.py is the ONLY
+# pin on the auto-commit hook's fleet-updater handshake, and it appeared in no
+# entry above and in no written exclusion below, so this manifest's own claim
+# that "EVERY suite that can be named IS named" was false.
+#
+# It cost a red CI on PR #321: verify.sh ran 12/12 green locally while that file
+# sat unrun, and the capability gate then failed on it in Actions. Measured
+# 2026-09-07 before naming it: 6 tests, `Ran 6 tests ... OK`, collects clean from
+# the repo root. A file entry, because q-system/hooks/test/ holds one python test
+# file and one shell script.
+q-system/hooks/test/test_auto_commit_run_marker.py
 #
 # THE REMAINING FIVE, each measured, none silently dropped.
 #
--- a/q-system/.q-system/tests/test_auto_commit.py
+++ b/q-system/.q-system/tests/test_auto_commit.py
@@ -696,3 +696,38 @@ def test_a_worktree_still_refuses_on_its_own_lock(tmp_path):
     log = subprocess.run(["git", "log", "--oneline"], cwd=wt,
                          capture_output=True, text=True).stdout
     assert "chore" not in log, "it committed while holding its own lock"
+
+
+def test_the_lock_dir_resolves_against_PROJ_DIR_not_the_process_cwd(tmp_path):
+    """PR #321 review, minor. The SAME defect PR #314 round 2 fixed one guard up.
+
+    `git rev-parse --git-dir` answers RELATIVE to the cwd it ran in, and run()
+    executes in PROJ_DIR (CLAUDE_PROJECT_DIR), which is not this process's cwd.
+    Resolving the answer against os.getcwd() points at a path that does not exist
+    whenever the two differ, no lock is ever found, and the guard FAILS OPEN into
+    the exact pre-guard behaviour it was built to remove.
+
+    fleet_update_in_progress carries that scar in its own comment and has no test
+    for it either. Every other test in this file fires with cwd == PROJ_DIR, so
+    `os.path.abspath(git_dir)` in place of `os.path.abspath(join(PROJ_DIR, ...))`
+    passed all 43 of them. This is the one that can tell them apart.
+    """
+    root, run = _repo(tmp_path)
+    _write(root, "memory/MEMORY.md", "- note\n")
+    elsewhere = tmp_path / "elsewhere"
+    elsewhere.mkdir()
+    assert not (elsewhere / ".git").exists(), "the fixture must not be a repo"
+    lock = root / ".git" / "index.lock"
+    lock.write_text("held")
+    try:
+        out = subprocess.run(
+            [sys.executable, HOOK], capture_output=True, text=True,
+            cwd=str(elsewhere),
+            env=dict(os.environ, CLAUDE_PROJECT_DIR=str(root)))
+    finally:
+        lock.unlink()
+    assert "another commit is in flight" in (out.stdout + out.stderr), (
+        "with cwd outside PROJ_DIR the guard did not see PROJ_DIR's held lock. "
+        "The git-dir answer is being resolved against the process cwd.")
+    assert "chore" not in run("git", "log", "--oneline").stdout, \
+        "it committed while a lock was held in PROJ_DIR's git dir"
--- a/q-system/.q-system/tests/test_auto_commit.py
+++ b/q-system/.q-system/tests/test_auto_commit.py
@@ -747,7 +747,7 @@ def test_a_refusal_reaches_the_channel_the_fleet_wiring_keeps(tmp_path):
     """REVIEW ROUND 2, MAJOR. The refusal was on stderr, which is thrown away.
 
     settings-template.json wires this hook as `... auto-commit.py 2>/dev/null ||
-    true` (lines 380 and 406), and that is the copy the fleet updater installs on
+    true` (line 395), and that is the copy the fleet updater installs on
     every instance. The behaviour the lock guard replaced reported a collision on
     STDOUT via commit_group's `skipped:` line, which survived that redirect. A
     stderr-only refusal does not: an orphaned index.lock, which has no age bound
--- a/q-system/.q-system/tests/test_auto_commit.py
+++ b/q-system/.q-system/tests/test_auto_commit.py
@@ -716,8 +716,14 @@ def test_the_lock_dir_resolves_against_PROJ_DIR_not_the_process_cwd(tmp_path):
     whenever the two differ, no lock is ever found, and the guard FAILS OPEN into
     the exact pre-guard behaviour it was built to remove.
 
-    fleet_update_in_progress carries that scar in its own comment and has no test
-    for it either. Every other test in this file fires with cwd == PROJ_DIR, so
+    fleet_update_in_progress carries that scar in its own comment and IS pinned,
+    but in a DIFFERENT FILE this suite does not run:
+    q-system/hooks/test/test_auto_commit_run_marker.py::
+    test_live_marker_is_seen_when_cwd_is_not_the_project_dir. An earlier draft of
+    this docstring said it had no test either; that was wrong, and missing that
+    second file is also how a channel change here shipped past a green run of
+    this suite and reddened CI. Every other test in THIS file fires with
+    cwd == PROJ_DIR, so
     `os.path.abspath(git_dir)` in place of `os.path.abspath(join(PROJ_DIR, ...))`
     passed all 43 of them. This is the one that can tell them apart.
     """
--- a/q-system/.q-system/tests/test_auto_commit.py
+++ b/q-system/.q-system/tests/test_auto_commit.py
@@ -626,3 +626,73 @@ def test_it_still_commits_when_no_lock_is_held(tmp_path):
     _fire(root)
     log = run("git", "log", "--oneline").stdout
     assert "chore" in log, "the hook refused with no lock held, so it never commits"
+
+
+def _worktree(tmp_path):
+    """A linked worktree of a fresh repo, plus its main checkout.
+
+    Returns (main_root, wt_root, main_git_dir, wt_git_dir).
+    """
+    main, run = _repo(tmp_path)
+    wt = tmp_path / "wt"
+    run("git", "branch", "-q", "side")
+    run("git", "worktree", "add", "-q", str(wt), "side")
+    main_git = main / ".git"
+    wt_git = main_git / "worktrees" / "wt"
+    assert wt_git.is_dir(), f"the worktree fixture never linked: {wt_git}"
+    return main, wt, main_git, wt_git
+
+
+def test_a_worktree_does_not_refuse_on_the_main_checkouts_lock(tmp_path):
+    """THE MUTANT THIS EXISTS FOR: --git-common-dir in place of --git-dir.
+
+    It survives every other test in this file, because a plain repo's git dir IS
+    its common dir, so the two spellings are indistinguishable there. They are not
+    indistinguishable on this checkout, which carries a dozen linked worktrees.
+
+    MEASURED 2026-09-07, not reasoned: index.lock and HEAD.lock are PER-WORKTREE.
+    Holding `<main>/.git/index.lock` does not block a `git add` run from a linked
+    worktree (rc=0), and holding `<main>/.git/worktrees/<n>/index.lock` does not
+    block one run from the main checkout (rc=0). Only the worktree's own lock
+    stops it (rc=128, "Unable to create ... index.lock"). So `--git-dir`, which
+    answers the per-worktree dir, names exactly the lock that can stop THIS
+    checkout, and `--git-common-dir` names one that cannot.
+
+    The cost of getting it wrong is quiet and fleet-wide: a commit in the main
+    checkout holds index.lock for its whole pre-commit (445-497s measured), so a
+    common-dir read would silence the safety net in every linked worktree for
+    eight minutes at a time, for a lock none of them was ever going to contend.
+    Note that fleet_update_in_progress deliberately reads the OPPOSITE dir, and
+    is right to: a run marker is a claim on the whole checkout, a lock is not.
+    """
+    _main, wt, main_git, _wt_git = _worktree(tmp_path)
+    _write(wt, "memory/MEMORY.md", "- note\n")
+    lock = main_git / "index.lock"
+    lock.write_text("held by the main checkout")
+    try:
+        _fire(wt)
+    finally:
+        lock.unlink()
+    log = subprocess.run(["git", "log", "--oneline"], cwd=wt,
+                         capture_output=True, text=True).stdout
+    assert "chore" in log, (
+        "the worktree refused on a lock in the MAIN checkout's git dir, which "
+        "cannot block it. The guard is reading --git-common-dir, not --git-dir.")
+
+
+def test_a_worktree_still_refuses_on_its_own_lock(tmp_path):
+    """The other direction, so the test above cannot be satisfied by never
+    refusing at all. The per-worktree lock is the one that stops this checkout."""
+    _main, wt, _main_git, wt_git = _worktree(tmp_path)
+    _write(wt, "memory/MEMORY.md", "- note\n")
+    lock = wt_git / "index.lock"
+    lock.write_text("held by this worktree")
+    try:
+        out = _fire(wt)
+    finally:
+        lock.unlink()
+    assert "another commit is in flight" in (out.stdout + out.stderr), \
+        "a worktree fired into its OWN held index.lock instead of refusing"
+    log = subprocess.run(["git", "log", "--oneline"], cwd=wt,
+                         capture_output=True, text=True).stdout
+    assert "chore" not in log, "it committed while holding its own lock"
--- a/q-system/.q-system/tests/test_auto_commit.py
+++ b/q-system/.q-system/tests/test_auto_commit.py
@@ -555,3 +555,74 @@ def test_a_binary_file_degrades_instead_of_crashing(tmp_path):
     msg = _last_msg(run)
     assert rel in msg, f"the binary file was dropped from the message\n{msg}"
     assert "binary" in msg, f"the binary file was not marked as such\n{msg}"
+
+
+# --- the third writer (2026-09-06) -------------------------------------------------
+# fleet_update_in_progress coordinates with the one writer that leaves a marker. This
+# hook was the writer nobody could negotiate with: it fires at every turn end in every
+# session on a shared checkout and took no lock at all. Measured in one session that
+# night, three of five git operations died on "Unable to create .git/index.lock" and
+# every one of them EXITED 0.
+
+def test_it_refuses_while_another_commit_holds_index_lock(tmp_path):
+    """sp-4bff1b91. A commit in flight owns index.lock for its whole pre-commit,
+    measured at 447-497s on the consulting checkout. Firing into that either dies at
+    exit 0 or lands a commit nobody asked for."""
+    root, run = _repo(tmp_path)
+    _write(root, "memory/MEMORY.md", "- note\n")
+    lock = os.path.join(root, ".git", "index.lock")
+    with open(lock, "w", encoding="utf-8") as fh:
+        fh.write("")
+    try:
+        out = _fire(root)
+    finally:
+        os.remove(lock)
+    assert "another commit is in flight" in (out.stdout + out.stderr), \
+        "the hook fired into a held index.lock instead of refusing"
+    log = run("git", "log", "--oneline").stdout
+    assert "chore" not in log, "it committed while another writer held the lock"
+
+
+def test_it_refuses_on_head_lock_too(tmp_path):
+    """The ref lock, not the index lock, is the one that killed the prd-os ledger
+    sweep mid pre-commit (sp-d0ce1966). Both doors or neither."""
+    root, run = _repo(tmp_path)
+    _write(root, "memory/MEMORY.md", "- note\n")
+    lock = os.path.join(root, ".git", "HEAD.lock")
+    with open(lock, "w", encoding="utf-8") as fh:
+        fh.write("")
+    try:
+        out = _fire(root)
+    finally:
+        os.remove(lock)
+    assert "another commit is in flight" in (out.stdout + out.stderr)
+
+
+def test_it_does_not_delete_the_lock_it_refuses_on(tmp_path):
+    """THE CONTROL THAT MATTERS MOST. kipi-update.sh:1714 force-deletes these locks
+    with no pid, mtime or age test (sp-50119dec). Removing a lock that a live
+    8-minute pre-commit still owns corrupts that commit's index. Refusing is the
+    whole fix; deleting would be a worse bug wearing the same commit message."""
+    root, _ = _repo(tmp_path)
+    _write(root, "memory/MEMORY.md", "- note\n")
+    lock = os.path.join(root, ".git", "index.lock")
+    with open(lock, "w", encoding="utf-8") as fh:
+        fh.write("sentinel")
+    try:
+        _fire(root)
+        assert os.path.exists(lock), "the hook DELETED a lock it does not own"
+        with open(lock, encoding="utf-8") as fh:
+            assert fh.read() == "sentinel", "the hook rewrote a lock it does not own"
+    finally:
+        if os.path.exists(lock):
+            os.remove(lock)
+
+
+def test_it_still_commits_when_no_lock_is_held(tmp_path):
+    """The negative control. A guard that refuses always is not a guard, it is an
+    outage: the safety net exists so work survives a context loss."""
+    root, run = _repo(tmp_path)
+    _write(root, "memory/MEMORY.md", "- note\n")
+    _fire(root)
+    log = run("git", "log", "--oneline").stdout
+    assert "chore" in log, "the hook refused with no lock held, so it never commits"
--- a/q-system/.q-system/tests/test_auto_commit.py
+++ b/q-system/.q-system/tests/test_auto_commit.py
@@ -9,6 +9,7 @@ from __future__ import annotations
 import os
 import subprocess
 import sys
+import time
 
 import pytest
 
@@ -584,8 +585,16 @@ def test_it_refuses_while_another_commit_holds_index_lock(tmp_path):
 
 
 def test_it_refuses_on_head_lock_too(tmp_path):
-    """The ref lock, not the index lock, is the one that killed the prd-os ledger
-    sweep mid pre-commit (sp-d0ce1966). Both doors or neither."""
+    """HEAD.lock is a SEPARATE door, and this docstring named the wrong reason
+    for it. It said the ref lock killed the prd-os ledger sweep mid pre-commit
+    (sp-d0ce1966). Review round 2 measured that a commit on an attached branch
+    produces index.lock and `refs/heads/<branch>.lock` and NEVER HEAD.lock, and
+    an independent watch agrees: only index.lock appears through a slow
+    pre-commit. The check still earns its place -- HEAD.lock has real producers
+    (checkout, reset, merge, alongside the AUTO_MERGE / ORIG_HEAD / packed-refs
+    family) and dropping it reddens this test. What it does NOT cover is
+    `refs/heads/<branch>.lock`, which a commit really does take; that window is
+    sub-second and index.lock spans it, so it is deliberately not checked."""
     root, run = _repo(tmp_path)
     _write(root, "memory/MEMORY.md", "- note\n")
     lock = os.path.join(root, ".git", "HEAD.lock")
@@ -731,3 +740,53 @@ def test_the_lock_dir_resolves_against_PROJ_DIR_not_the_process_cwd(tmp_path):
         "The git-dir answer is being resolved against the process cwd.")
     assert "chore" not in run("git", "log", "--oneline").stdout, \
         "it committed while a lock was held in PROJ_DIR's git dir"
+
+
+
+def test_a_refusal_reaches_the_channel_the_fleet_wiring_keeps(tmp_path):
+    """REVIEW ROUND 2, MAJOR. The refusal was on stderr, which is thrown away.
+
+    settings-template.json wires this hook as `... auto-commit.py 2>/dev/null ||
+    true` (lines 380 and 406), and that is the copy the fleet updater installs on
+    every instance. The behaviour the lock guard replaced reported a collision on
+    STDOUT via commit_group's `skipped:` line, which survived that redirect. A
+    stderr-only refusal does not: an orphaned index.lock, which has no age bound
+    by design, would switch the safety net off on that checkout forever and print
+    nothing anywhere a human looks.
+
+    So this asserts the CHANNEL, not the wording. It fires the hook exactly the
+    way the fleet does, with stderr discarded, and fails if the reason vanishes.
+    """
+    root, _run = _repo(tmp_path)
+    _write(root, "memory/MEMORY.md", "- note\n")
+    lock = root / ".git" / "index.lock"
+    lock.write_text("held")
+    try:
+        out = subprocess.run(
+            [sys.executable, HOOK], capture_output=True, text=True, cwd=str(root),
+            env=dict(os.environ, CLAUDE_PROJECT_DIR=str(root)))
+    finally:
+        lock.unlink()
+    assert "another commit is in flight" in out.stdout, (
+        "the refusal did not reach STDOUT, so `2>/dev/null` in the shipped hook "
+        "wiring discards it and the safety net goes off in total silence.\n"
+        f"stdout={out.stdout!r}\nstderr={out.stderr!r}")
+
+
+def test_the_fleet_update_refusal_reaches_stdout_too(tmp_path):
+    """The sibling guard, hardened in the same breath. It had the identical
+    stderr-only shape and the identical consequence, and fixing one of two
+    guards that fail the same way is how this class comes back."""
+    root, _run = _repo(tmp_path)
+    _write(root, "memory/MEMORY.md", "- note\n")
+    marker = root / ".git" / "kipi-update.run"
+    marker.write_text(f"{os.getpid()} {time.strftime('%Y-%m-%dT%H:%M:%SZ', time.gmtime())}")
+    try:
+        out = subprocess.run(
+            [sys.executable, HOOK], capture_output=True, text=True, cwd=str(root),
+            env=dict(os.environ, CLAUDE_PROJECT_DIR=str(root)))
+    finally:
+        marker.unlink()
+    assert "fleet updater run in progress" in out.stdout, (
+        "the fleet-updater refusal did not reach STDOUT.\n"
+        f"stdout={out.stdout!r}\nstderr={out.stderr!r}")
--- a/q-system/hooks/auto-commit.py
+++ b/q-system/hooks/auto-commit.py
@@ -574,8 +574,13 @@ def main():
 
     # STDOUT, NOT STDERR, AND THAT IS THE WHOLE POINT (review round 2, major).
     # settings-template.json wires this hook as `... auto-commit.py 2>/dev/null
-    # || true` at lines 380 and 406, and that is the copy the fleet updater
-    # installs on every instance. So stderr is DISCARDED on all 22+ checkouts.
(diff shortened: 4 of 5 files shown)


assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-321-20260907-040836.md` (4341 bytes).

```
FINDINGS:
minor|No test distinguishes PROJ_DIR-relative from process-cwd resolution of --git-dir, so the PR #314 fail-open mutant survives all 43 tests|q-system/hooks/auto-commit.py:547
END FINDINGS
```

--- reviewer output ---

# Review: PR #321 — auto-commit refuses while another commit holds git's locks

One finding survived reproduction. Four attacks did not, and I say which below.

## Finding 1 (minor)

**Claim:** No test can tell `os.path.join(PROJ_DIR, git_dir)` apart from resolving against the process cwd, so the exact fail-open defect PR #314 round 2 fixed in the sibling guard is unpinned in the new one.

**Where:** `q-system/hooks/auto-commit.py:547`

**Reproducer** (mutates a COPY in a temp dir, repo untouched):

```
python3 - <<'PYEOF'   # full harness in the transcript
# M1: git_dir = os.path.abspath(git_dir)   <- resolve against process cwd
PYEOF
```

**Real output:**

```
CONTROL: ['...::TestDeclaredSkipMustActuallyBeIgnored::test_every_declared_skip_prefix_is_actually_gitignored']
M1:      ['...::TestDeclaredSkipMustActuallyBeIgnored::test_every_declared_skip_prefix_is_actually_gitignored']
M2:      [ ...same control failure..., '...::test_a_worktree_does_not_refuse_on_the_main_checkouts_lock',
           '...::test_a_worktree_still_refuses_on_its_own_lock']
```

The control failure is my sandbox copy missing the real `.gitignore`, unrelated to the PR. M1 adds zero failures: it survives all 43 tests. M2, the mutant the author designed for, is killed by two tests, one more than the docstring at line 668 claims.

Why it matters: when cwd and `CLAUDE_PROJECT_DIR` differ, the mutated form resolves to a path that does not exist, no lock is ever found, and the guard fails open into exactly the pre-PR behavior. The fixture shape already exists in the file at line 157, which fires with `cwd=tmp_path` and `CLAUDE_PROJECT_DIR="/nonexistent/nope"`. The code as merged is correct; only the pin is missing.

Cheaper fix than a new test: `kipi-update.sh:1339` already resolves this in one call, `git rev-parse --path-format=absolute --git-path index.lock`, which is worktree-correct and has no relative-path branch to get wrong.

## What is sound

**HEAD.lock does not add a new outage mode.** The docstring claims an orphaned lock already blocked this hook before the guard. I doubted it for `HEAD.lock`, since a branch commit locks `refs/heads/<name>.lock`. Measured: `git commit -m x -- b.txt` with `.git/HEAD.lock` held returns rc 128, `fatal: cannot lock ref 'HEAD'`, log unchanged. The claim holds.

**Every `--cached` removal claim in the comment at line 114 is true.** Measured all three: a staged new file still on disk prints from `git diff --name-only HEAD` (`'b.txt\n'`); the staged-then-deleted and staged-then-reverted shapes are invisible to both other probes and visible only to `--cached`; and a pathspec commit on each returns rc 1 `nothing to commit`, leaving the staged entry intact (`'c.txt\n'`, `'seed.txt\n'`). Nothing committable was dropped.

**The `~L1915` and `~L1313-1400` citations in the docstring are accurate**, including the self-correction away from L1714. `kipi-update.sh:1915` is the unconditional lock sweep; 1337 to 1400 is the bounded wait-and-retry.

**`report_skipped` is transcript-only, so the early return silences no alert.** I expected the guard to swallow an operator ping. It cannot: line 279 to 307 prints and never notifies.

**The mid-loop lock race, dropped for want of a repro.** The guard is read once before a loop that makes N commits, so a lock taken between group 1 and group 2 still yields a partial commit at exit 0. I built a `post-commit` hook that plants `index.lock` and both groups committed anyway, because git's own lockfile cleanup deletes the planted file at exit. I could not make it fail, so it is not a finding. It is also pre-existing and the PR never claims atomicity.

## VERDICT: APPROVE WITH NITS

Most important thing to fix first: pin the `PROJ_DIR`-relative resolution at `auto-commit.py:547`, or replace it with the `--git-path index.lock` call `kipi-update.sh` already uses. Same defect class, second guard, no test either time.

One note outside the findings block, since it is not code: the PR body's verification section says 41 passed, 37 before, 4 new. The suite reports 43 passed and the diff adds 6 tests.

FINDINGS:
minor|No test distinguishes PROJ_DIR-relative from process-cwd resolution of --git-dir, so the PR #314 fail-open mutant survives all 43 tests|q-system/hooks/auto-commit.py:547
END FINDINGS

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-321-20260907-042635.md` (6418 bytes).

```
FINDINGS:
major|The refusal prints only to stderr, which settings-template.json discards, so an orphaned index.lock silences the auto-commit safety net with no output at all|q-system/hooks/auto-commit.py:573
minor|The "commit holds index.lock for its whole pre-commit, 447-497s" measurement justifying no age bound is false; git holds no lock during pre-commit and holds index.lock ~12ms|q-system/hooks/auto-commit.py:512
minor|Test docstring claims the ref lock killed the ledger sweep, but a commit on an attached branch produces index.lock and never HEAD.lock|q-system/.q-system/tests/test_auto_commit.py:587
END FINDINGS
```

--- reviewer output ---

# Review: PR #321 — auto-commit refuses while another commit holds git's locks

Read-only review. Repo untouched; every mutation ran against a copy in a temp dir.

---

## Finding 1 — major

**The refusal's only diagnostic goes to stderr, which the fleet wiring discards, so an orphaned `index.lock` turns the safety net off with zero output.**

`q-system/hooks/auto-commit.py:573`

`settings-template.json:395` (the copy `kipi update` ships to every instance) wires the hook as `... auto-commit.py" 2>/dev/null || true`. The pre-guard code reported a lock collision on **stdout** as a `skipped:` line naming the file and the git error. The new guard returns before any of that and prints only to stderr. The docstring at line 524 says "The guard adds no new silent-outage mode." Under the shipped wiring it does.

Reproducer, run just now, comparing PR head against the parent of `f39c0cda` with an orphaned `index.lock` and real work in `memory/MEMORY.md`:

```
=== pre-guard ===
  STDOUT (survives `2>/dev/null` on every fleet instance):
      auto-commit: 1 files in 1 groups
        skipped: chore: update auto-memory (1 file(s), +0/-0) - error: pathspec 'memory/MEMORY.md' did not match any file(s) known to git
      auto-commit: done
  STDERR (discarded by settings-template.json):
      <EMPTY>
  committed? NO

=== pr321 ===
  STDOUT (survives `2>/dev/null` on every fleet instance):
      <EMPTY>
  STDERR (discarded by settings-template.json):
      auto-commit: another commit is in flight (index.lock held 0s); committing nothing. The next turn end picks this work up.
  committed? NO
```

At 3am on an instance, a killed `git add` inside a 450s lefthook run leaves the lock, and from that moment the auto-commit safety net is off on that checkout, forever, and prints nothing. There is no age bound by design, so nothing ages it out. Recoverable by deleting the lock, which is why this is major and not a blocker, but nothing tells anyone to.

---

## Finding 2 — minor

**The measurement that justifies "NO AGE BOUND ON PURPOSE" is contradicted by git.**

`q-system/hooks/auto-commit.py:512`

The comment reads: "A commit here holds index.lock for the length of its pre-commit, measured at 447-497 seconds here." Measured against git 2.54.0:

```
locks visible mid-pre-commit: []
concurrent `git add c` rc= 0 stderr= 
concurrent `git status` rc= 0
slow commit rc= 0
```

That is a 4-second pre-commit hook. Git holds **no lock at all** while the pre-commit runs, and a concurrent `git add` succeeds. Polling the real lock at 0.5ms over a 200-file add plus commit:

```
index.lock observed for approx 0.0118 s (samples=11)
```

Twelve milliseconds, not 450 seconds. So the stated reason a low age bound "would fire constantly against healthy ones" does not hold, and the same false premise is repeated in the PR body's `--git-common-dir` blast-radius paragraph. The `--git-dir` choice stays correct for its other, measured reason. The consequence is Finding 1: no bound, no signal.

---

## Finding 3 — minor

**A commit on an attached branch never produces `HEAD.lock`, so the test's stated justification is wrong.**

`q-system/.q-system/tests/test_auto_commit.py:587`

The docstring says "The ref lock, not the index lock, is the one that killed the prd-os ledger sweep." Watching every `.lock` under `.git` at 0.2ms:

```
commit on an ATTACHED branch             -> ['index.lock', 'objects/maintenance.lock']
30x `git status --porcelain`             -> ['index.lock']
```

A commit writes `refs/heads/<branch>.lock`, which this guard does not check, and never `HEAD.lock`. I tried to make this a coverage finding and could not: `HEAD.lock` **does** have real producers, so the check is not dead code:

```
lock filenames ever seen in .git/ across checkout/reset/add/commit:
  ['AUTO_MERGE.lock', 'HEAD.lock', 'ORIG_HEAD.lock', 'index.lock', 'packed-refs.lock']
```

So this is the comment being wrong, not the code. Fix the sentence.

---

## What is sound

**The `--git-dir` versus `--git-common-dir` call is right, and I verified it independently rather than reading the docstring:**

```
main .git/index.lock held  -> `git add` IN WORKTREE rc= 0
worktree's OWN index.lock  -> `git add` IN WORKTREE rc= 128 fatal: Unable to create ...
```

Locks are per-worktree. Reading the common dir would refuse on a lock that cannot block the caller.

**The tests are not decoration.** Five mutants against a copied tree, all killed, each on its own case:

```
guard disabled           5 failed, 39 passed
common-dir spelling      3 failed, 41 passed
resolve vs process cwd   2 failed, 42 passed
drop HEAD.lock           2 failed, 42 passed
guard DELETES the lock   6 failed, 38 passed
```

Baseline is 44 passed in 2.04s. The read-only property and the PROJ_DIR pin both have a case that can actually go red.

**The false-refusal rate is disclosed honestly, and I confirmed it.** `git status` really does take `index.lock`, so peer sessions will cost some turn ends a commit. The docstring says so instead of hiding it.

**Attacks that found nothing:** the check-then-act window between the guard and `commit_group` degrades to the same `skipped:` line it always did, not to a wrong commit. `PROJ_DIR` unset leaves `run()` and the resolution base consistent. The `--system-state` path is untouched. The `--cached` removal claim held up: `git commit -m x -- <path>` is a no-op on both shapes the removal loses.

---

## VERDICT: REQUEST CHANGES

Not a block. Nothing here is permanent or unrecoverable.

**Fix first:** move the refusal to stdout so it survives `2>/dev/null` on the 22+ instances, the way the `skipped:` line it replaced did. Then reconsider the age bound, since the 450-second premise it rests on is not what git does.

FINDINGS:
major|The refusal prints only to stderr, which settings-template.json discards, so an orphaned index.lock silences the auto-commit safety net with no output at all|q-system/hooks/auto-commit.py:573
minor|The "commit holds index.lock for its whole pre-commit, 447-497s" measurement justifying no age bound is false; git holds no lock during pre-commit and holds index.lock ~12ms|q-system/hooks/auto-commit.py:512
minor|Test docstring claims the ref lock killed the ledger sweep, but a commit on an attached branch produces index.lock and never HEAD.lock|q-system/.q-system/tests/test_auto_commit.py:587
END FINDINGS

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-321-20260907-044514.md` (6825 bytes).

```
FINDINGS:
minor|The comment says the removed `--cached` read was harmless and tells the next maintainer not to expect a bug fix from removing it; measured, the pre-PR hook silently unstaged or overwrote a peer session's staged work in both shapes the comment names|q-system/hooks/auto-commit.py:133
minor|The refusal message asserts "another commit is in flight" from a bare existence check with no liveness test, so a 30-day orphaned lock prints the same sentence while the safety net stays off on that checkout permanently|q-system/hooks/auto-commit.py:596
minor|The comment and the test docstring both cite settings-template.json "lines 380 and 406" for the auto-commit wiring, which appears once at line 395; 380 and 406 are unrelated hooks|q-system/hooks/auto-commit.py:577
END FINDINGS
```

--- reviewer output ---

Review complete. Three findings survived reproduction, all minor.

## Finding 1 (minor) — the comment tells the next maintainer to reintroduce a data-loss path

`q-system/hooks/auto-commit.py:133-135` says the removed `--cached` read was harmless:

> "(The first guess written here was that the delete shape would commit a DELETION of another writer's staged file. **It does not; the staged entry survives untouched.**) So the two extra paths **only ever bought a spurious `skipped:` line**. What this removal gets is a narrower changed-file set, **not a bug fix. Do not re-add it expecting one**"

Measured against both `main` and this branch, on the two staging shapes the comment itself names. Reproducer ran the real `main` hook and the PR hook against a peer's staged work:

```
shape A / BASE main  probes={'diff HEAD': [], 'ls-files --others': [], 'diff --cached': ['q-system/memory/NEW.md']}
        hook stdout: auto-commit: 1 files in 1 groups | skipped: ... pathspec 'q-system/memory/NEW.md' did not match ...
        staged AFTER: []   staged blob now: ''
shape A / PR 321     staged AFTER: ['q-system/memory/NEW.md']   staged blob now: 'PEER STAGED NEW FILE'

shape B / BASE main  probes={'diff HEAD': [], 'ls-files --others': [], 'diff --cached': ['q-system/memory/MEMORY.md']}
        staged AFTER: []   staged blob now: 'v1'
shape B / PR 321     staged AFTER: ['q-system/memory/MEMORY.md']   staged blob now: 'PEER STAGED WORK'
```

The staged entry does not survive. In shape A the peer's staged new file is unstaged and its blob is gone. In shape B the peer's staged content `PEER STAGED WORK` is overwritten back to `v1` and exists nowhere on disk afterward. `git add -- <path>` on a path missing from the worktree stages the removal; on a path whose worktree copy equals HEAD it overwrites the staged blob.

So the removal is a bug fix, and a load-bearing one on a shared checkout at every turn end. The comment says the opposite and explicitly invites re-adding the line. That is the failure mode this same file names one screen up: a wrong scar comment is read as coverage, so nobody goes looking.

Fix: correct those three sentences and pin both shapes with a test. The code stays as is.

## Finding 2 (minor) — "in flight" is asserted from an existence check with no liveness test

`q-system/hooks/auto-commit.py:596`. The guard checks `os.path.exists` only. It never asks whether a writer holds the lock. Reproducer set the lock mtime 30 days back:

```
---- BASE main no guard exit 0 ----
STDOUT: 'auto-commit: 1 files in 1 groups\n  skipped: chore: update session memory ...'
---- PR 321 exit 0 ----
STDOUT: 'auto-commit: another commit is in flight (index.lock held 2592000s); committing nothing. The next turn end picks this work up.'
```

Neither version commits, so the PR's "no new silent outage" claim holds and this is not a regression. What is wrong is the sentence: at 3am on an orphaned lock the safety net is off on that checkout permanently and the one line a human might read names a live writer that does not exist. The `2592000s` is the only tell, and it is inside the parenthetical. Worth a different phrasing past some age, or wording that says what was actually observed.

## Finding 3 (minor) — the settings-template.json citation is wrong in code and in the test

`q-system/hooks/auto-commit.py:577` and `q-system/.q-system/tests/test_auto_commit.py:750` both say the hook is wired "at lines 380 and 406".

```
$ grep -c "auto-commit.py" settings-template.json
1
$ grep -n "auto-commit" settings-template.json
395:  "command": "python3 \"$CLAUDE_PROJECT_DIR/q-system/hooks/auto-commit.py\" 2>/dev/null || true",
380:  ... code_claim_grounding_guard.py ...
406:  "async": true          (belongs to memory_autocapture.py)
```

One wiring, at 395. Lines 380 and 406 are different hooks. The same docstring also says the test "fires the hook exactly the way the fleet does, with stderr discarded"; it captures stderr into a pipe and asserts on stdout. The assertion is the right one, the description is not.

## What is sound

Attacks that the code survived:

- **Mutation, both directions.** Eight mutants on a copy in a temp tree. Every one died: disable the guard (6 RED), make the guard delete the lock it finds (do-not-delete control RED), drop `HEAD.lock` (RED), swap `--git-dir` for `--git-common-dir` (both worktree tests RED), resolve the git dir against process cwd (PROJ_DIR test RED), put either refusal back on stderr (the matching channel test RED). Then the inverse: guard always refuses (14 RED), fleet guard always refuses (19 RED). No decoration in this suite.
- **The disputed no-age-bound claim, re-measured independently.** Against a 4-second pre-commit: `git commit -q -m x -- a.txt`, which is exactly what `commit_group` runs, held `index.lock` in 157 of 218 samples plus a `next-index-*.lock`. `git commit -a` held it in 143 of 202. `git add` then a bare `git commit` held nothing. The docstring's figure and its warning to measure the pathspec form both stand.
- **False-refusal rate from peer read commands.** 200 back-to-back `git status --porcelain` showed `index.lock` in 1 of 42 sampler passes. Real, low, and the docstring already names it.
- **Fixture shapes match real producers.** The fleet-marker test writes `pid + space + UTC stamp`; `kipi-update.sh:1444` writes `printf '%s %s\n' "$$" "$(date -u +%Y-%m-%dT%H:%M:%SZ)"`. The worktree fixture uses a real `git worktree add`, not a hand-built directory.
- **Cited code exists.** `kipi-update.sh:1915` is the unconditional `rm -f` sweep with no pid, mtime or age test. `kipi-update.sh:1313-1400` is the wait-and-retry path, `LOCK_WAIT_S` defaults to 600.
- **Rollback.** Pure code change in a synced path, revertible by revert.

Suite on the branch: `46 passed in 2.20s`. The PR body says 44.

First thing to fix: the comment at `auto-commit.py:133-135`. It is measurably backwards, and it is the one sentence in this diff that can cost someone's staged work later.

FINDINGS:
minor|The comment says the removed `--cached` read was harmless and tells the next maintainer not to expect a bug fix from removing it; measured, the pre-PR hook silently unstaged or overwrote a peer session's staged work in both shapes the comment names|q-system/hooks/auto-commit.py:133
minor|The refusal message asserts "another commit is in flight" from a bare existence check with no liveness test, so a 30-day orphaned lock prints the same sentence while the safety net stays off on that checkout permanently|q-system/hooks/auto-commit.py:596
minor|The comment and the test docstring both cite settings-template.json "lines 380 and 406" for the auto-commit wiring, which appears once at line 395; 380 and 406 are unrelated hooks|q-system/hooks/auto-commit.py:577
END FINDINGS

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-321-20260907-050505.md` (6586 bytes).

```
FINDINGS:
minor|Guard runs before get_changed_files, so a clean tree gets a message claiming deferred work that does not exist|q-system/hooks/auto-commit.py:601
minor|another_commit_in_flight has no bounded wait or retry, so a sub-second peer lock drops the whole turn's commit and the "next turn end" promise is false on a session's last turn|q-system/hooks/auto-commit.py:488
nit|Test docstring says it fires the hook "exactly the way the fleet does, with stderr discarded" but the body captures stderr and omits `|| true` and async|q-system/.q-system/tests/test_auto_commit.py:757
nit|PR body verification block says 44 passed / 7 new tests; the branch runs 46 passed with 9 new tests|q-system/.q-system/tests/test_auto_commit.py:559
END FINDINGS
```

--- reviewer output ---

## Review: PR #321 — `fix(auto-commit): refuse while another commit holds git's locks`

Read-only. Nothing in the repo was modified. All repros run in `tempfile.mkdtemp()`.

---

### Finding 1 (minor) — the refusal claims deferred work on a clean tree

`q-system/hooks/auto-commit.py:601`. The guard runs at `main()` before `files = get_changed_files()` at line 605, so a peer `git status` holding `index.lock` for a fraction of a second makes the hook print a deferral message when there was nothing to commit. The old path printed `auto-commit: no changes`. This hook fires at every turn end in every session, so the false line is the one that trains an operator to stop reading it.

Repro (clean tree, lock held):

```
python3 -c "... clean repo, touch .git/index.lock, fire the hook ..."
```

Real output:

```
status on clean tree: ''
STDOUT: auto-commit: another commit is in flight (index.lock held 0s); committing nothing. The next turn end picks this work up.
rc 0
```

Nothing was pending. Same early return also skips `report_skipped`, so the loud NOTICE for unclassified source files goes quiet for the duration of the lock.

---

### Finding 2 (minor) — zero wait, zero retry, so a sub-second peer lock costs the whole turn's commit

`q-system/hooks/auto-commit.py:488`. `another_commit_in_flight` is a single `os.path.exists` with no bounded wait. The docstring rejects waiting on the grounds that a Stop hook cannot spend 600s, which is true, and treats 0s and 600s as the only options. `kipi-update.sh:1313-1340` answers this identical race in this same repo by waiting bounded with retry.

Repro: a peer holds `index.lock` for 400ms, the way a `git status` does.

```
HOOK STDOUT: auto-commit: another commit is in flight (index.lock held 0s); committing nothing. The next turn end picks this work up.
log: ['d6d872d seed']
still uncommitted: '?? memory/'
lock gone now: True
```

The lock was gone 350ms later. On the final turn end of a session there is no next turn end, so "the next turn end picks this work up" is false in exactly the context-loss case the safety net exists for. A 1-2s bounded wait covers the whole `git status` class at near-zero cost.

---

### Finding 3 (nit) — a test docstring misstates what the test runs

`q-system/.q-system/tests/test_auto_commit.py:757`. It says "It fires the hook exactly the way the fleet does, with stderr discarded." The body at line 765 uses `capture_output=True`, which captures stderr rather than discarding it, and the fleet wiring also carries `|| true` and `async: true`, neither reproduced. The assertion on `out.stdout` is correct and the test does kill the stderr-only mutant, so this is wording only. Flagging it because this PR spent two rounds correcting exactly this class of wrong scar comment.

---

### Finding 4 (nit) — the PR body's verification numbers do not match the branch

The body says `44 passed (37 before, 7 new)`. Actual:

```
python3 -m pytest q-system/.q-system/tests/test_auto_commit.py -q
..............................................  [100%]
46 passed in 2.20s
```

The diff adds 9 test functions, not 7. 37 + 9 = 46.

---

### What is sound

Five attacks the code survived. Each was run, not reasoned.

**`--git-dir` versus `--git-common-dir` is right.** Held `<main>/.git/index.lock`, then ran `git add` and a full `git commit` from a linked worktree: both `rc=0`, commit landed. The common-dir spelling would have been a false positive.

**Path resolution holds in all three shapes.** `git rev-parse --git-dir` returns `.git` at toplevel, absolute from a subdirectory, absolute in a linked worktree. The `PROJ_DIR` join is correct for all three.

**`HEAD.lock` earns its check.** I expected a stale `HEAD.lock` to be a new permanent outage mode that git itself would have tolerated. It is not: a pathspec commit with a pre-existing `HEAD.lock` exits 128 with `fatal: cannot lock ref 'HEAD'`. The guard refuses where git was already going to fail.

**The 447-497s index.lock hold is real.** Watched a repo with a 3s pre-commit: `git commit -m x -- <path>` creates `index.lock` at t=0.02s and holds it for the whole hook. The pathspec form is what `commit_group` runs, so the figure is measured on the right shape.

**The `--cached` analysis is correct, and understates itself.** Across six staging shapes, dropping `--cached` loses exactly the two paths the comment names: `new_staged_del.txt` and `tracked_revert.txt`. And the removal is better than "narrower set" — the old code fed `tracked_revert.txt` into `commit_group`'s `git add`, which overwrote a peer session's staged blob:

```
before: staged blob = PEER STAGED v1
after old-code git add: staged blob = v0
```

That is silent destruction of another writer's staged work, and this PR removes it. Worth naming in the commit message.

**The guard works against a real producer, not just a hand-written lock file.** Ran a genuine concurrent `git commit` behind a 3s pre-commit and fired the hook mid-flight. New code: one accurate refusal, peer commit intact. Old code at `f39c0cda^`, same scenario: a misleading `skipped: ... pathspec 'memory/MEMORY.md' did not match any file(s) known to git`, plus a false `1 file(s) NOT committed` naming `peer.txt`, which the peer was committing at that moment. The change replaces two wrong reports with one right one.

---

Only minor and nit findings survived reproduction. The core fix is correct, the `--git-dir` choice is measured rather than argued, and the stdout channel fix is load-bearing given `settings-template.json:395` discards stderr.

**Fix first:** move the guard below `get_changed_files()` (Finding 1) so a refusal only claims deferred work when work exists, and add a 1-2s bounded wait (Finding 2) so the sub-second `git status` class stops costing a turn's commit.

FINDINGS:
minor|Guard runs before get_changed_files, so a clean tree gets a message claiming deferred work that does not exist|q-system/hooks/auto-commit.py:601
minor|another_commit_in_flight has no bounded wait or retry, so a sub-second peer lock drops the whole turn's commit and the "next turn end" promise is false on a session's last turn|q-system/hooks/auto-commit.py:488
nit|Test docstring says it fires the hook "exactly the way the fleet does, with stderr discarded" but the body captures stderr and omits `|| true` and async|q-system/.q-system/tests/test_auto_commit.py:757
nit|PR body verification block says 44 passed / 7 new tests; the branch runs 46 passed with 9 new tests|q-system/.q-system/tests/test_auto_commit.py:559
END FINDINGS

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-321-20260907-054715.md` (7724 bytes).

```
FINDINGS:
minor|The lock guard runs once in main(), so a lock taken afterward returns groups 2..N to the pre-guard silent skip at exit 0|q-system/hooks/auto-commit.py:571
minor|A lock collision is reported as "pathspec did not match any file(s)", naming the wrong cause for the operator|q-system/hooks/auto-commit.py:424
minor|The round-5 comment declares "EVERY suite that can be named IS named" repaired while q-system/.q-system/scripts/test/ (56 passed, collects clean) is in no entry and no written exclusion|.verify-suites:84
END FINDINGS
```

--- reviewer output ---

## Review of PR #321

Read-only. Nothing in the repo was modified; all mutants ran against copies in `mktemp -d`.

---

### Finding 1 (minor): the lock is checked once, so every commit group after the first still takes the silent-skip path

`q-system/hooks/auto-commit.py:571` calls `another_commit_in_flight()` once at the top of `main()`. `commit_group` then runs one `git add` plus one `git commit` per group, and the first group's commit holds the index for its whole pre-commit (the docstring measures 447 to 497 seconds). A peer writer that takes `index.lock` any time in that window sends groups 2 through N straight back down the pre-guard path: exit 0, work uncommitted, and a message that names the wrong cause.

Reproducer, run just now. A `post-commit` hook in the fixture repo takes `index.lock` the instant the first group's commit releases it, which is exactly the peer-writer shape the PR exists to remove:

```
python3 - <<'PY'
import os, subprocess, sys, tempfile
HOOK = ".../q-system/hooks/auto-commit.py"
d = tempfile.mkdtemp()
def g(*a): return subprocess.run(("git",)+a, cwd=d, capture_output=True, text=True)
g("init","-q"); g("config","user.email","t@t"); g("config","user.name","t")
open(os.path.join(d,"seed.txt"),"w").write("x"); g("add","-A"); g("commit","-q","-m","seed")
for rel in ("q-system/memory/MEMORY.md", "q-system/canonical/positioning.md"):
    p = os.path.join(d, rel); os.makedirs(os.path.dirname(p), exist_ok=True)
    open(p,"w").write("- note\n")
hd = os.path.join(d,".git","hooks"); os.makedirs(hd, exist_ok=True)
pc = os.path.join(hd,"post-commit")
open(pc,"w").write('#!/bin/sh\n: > "$(git rev-parse --git-dir)/index.lock"\n'); os.chmod(pc,0o755)
out = subprocess.run([sys.executable, HOOK], cwd=d, capture_output=True, text=True,
                     env=dict(os.environ, CLAUDE_PROJECT_DIR=d))
print("exit", out.returncode); print(out.stdout); print(g("status","--porcelain").stdout)
PY
```

Real output:

```
hook exit code: 0
auto-commit: 2 files in 2 groups
  committed: content: update canonical files (1 file(s), +1/-0) (1 files)
  skipped: chore: update session memory (1 file(s), +0/-0) - error: pathspec 'q-system/memory/MEMORY.md' did not match any file(s) known to g
auto-commit: done
--- still dirty ---
?? q-system/memory/
```

Two things there. The skip is real, and the operator is told `pathspec ... did not match any file(s)`, not that a lock was held. The `git add` lost the lock, so the new file was never staged, and the pathspec commit then failed on an unknown path. Someone reading that line at 3am goes looking for a classifier bug.

Bounded on purpose in my rating: nothing is destroyed, the work sits on disk, and the next turn end picks it up. That is minor, not major. A re-check inside `commit_group`, or matching `index.lock` in the non-zero `r.stderr` before printing `skipped:`, closes both halves.

---

### Finding 2 (minor): the round-5 `.verify-suites` comment declares the coverage claim repaired, and two green suites are still unnamed

`.verify-suites:83-94` adds `q-system/hooks/test/test_auto_commit_run_marker.py` and states that the manifest's own claim, "EVERY suite that can be named IS named", was false and is now fixed. Measured against the enumeration the file itself prescribes, `q-system/.q-system/scripts/test/` is in no manifest entry and in no written exclusion. The exclusion prose covers `q-system/.q-system/scripts`, which is a different directory.

```
$ python3 -m pytest q-system/.q-system/scripts/test/ -q
........................................................  [100%]
56 passed in 2.61s
```

```
$ bash q-system/.q-system/verify.sh --full | tail -3
verify.sh ok (13 checks: ... pytest:q-system/hooks/test/test_auto_commit_run_marker.py)
```

Thirteen checks, none of them that directory. Both files are in `capability/expected_tests/`, so CI runs them and the local floor does not:

```
$ grep -c "capability-gate" q-system/.q-system/verify.sh
0
```

That is the identical status `test_auto_commit_run_marker.py` held before this PR, and the identical cost: verify.sh green locally, capability gate red in Actions. Same shape, sixth occurrence, and the comment in this diff says it is closed.

---

### What is sound

Every attack I aimed at the guard bounced, and I ran the mutants rather than reading the table in the PR body.

**All seven claimed mutants die, plus one the body did not list.** Copies of the hook and its suite in `mktemp -d`, mutated one line each:

```
M1 guard disabled (held = None)      5 lock tests RED
M2 --git-common-dir                  both worktree tests RED
M3 resolve git dir vs process cwd    the PROJ_DIR pin RED
M4 refusal back to stderr            the fleet-channel test RED
M5 drop HEAD.lock from the tuple     the HEAD.lock test RED
M6 guard os.remove()s the lock       the do-not-delete control RED
```

(One unrelated baseline failure, `test_every_declared_skip_prefix_is_actually_gitignored`, is an artifact of my partial copy. In the real tree the suite is 46 passed.)

**The `--cached` removal loses nothing the safety net was catching.** I did not take the comment's word for it. Six staging shapes, probed directly:

```
staged NEW, present in worktree        new=['b']  cached-only-adds=[]
staged NEW then edited more            new=['b']  cached-only-adds=[]
staged RENAME (git mv)                 new=['r2'] cached-only-adds=[]
staged MOD then reverted in worktree   new=[]     cached-only-adds=['a']
staged NEW then deleted from worktree  new=[]     cached-only-adds=['b']
staged DELETE of tracked (rm --cached) new=['a']  cached-only-adds=[]
```

Exactly the two shapes the comment names, and a pathspec commit on either exits 1 with nothing written. I went further and ran the old path on the delete shape: `git add -- b` returns 0 and **wipes the peer's staged entry from the index**, then the commit fails. The removal is strictly safer than the comment claims, not less.

**`report_skipped` is not an alert path.** I chased the theory that the early return suppresses a Sana ping. It does not, and the docstring at `auto-commit.py:279` is accurate: transcript only, founder-directed 2026-08-10.

**No stdout consumer is contaminated by the channel change.** `kipi-update.sh:2088` shells this hook, but through `--system-state`, which returns before `main()` ever prints. Line 395 of `settings-template.json` is the single wiring and does carry `2>/dev/null || true`, so the comment's pointer is right this time.

**Both new suites are green and the floor runs them.** `test_auto_commit.py` 46 passed, `test_auto_commit_run_marker.py` 6 passed, `verify.sh --full` 13/13. `capability-gate.py` was still running past 600s when I stopped waiting; I did not confirm the 231-test figure myself.

---

The guard is real, the tests behind it are load-bearing under mutation, and the comments say what the code does. Two minors, both bounded and both captured as follow-ups.

**First thing to fix:** the `skipped:` line in `commit_group` should say a lock was held when a lock was held. `pathspec did not match any file(s)` sends the next reader after the classifier instead of the writer.

FINDINGS:
minor|The lock guard runs once in main(), so a lock taken afterward returns groups 2..N to the pre-guard silent skip at exit 0|q-system/hooks/auto-commit.py:571
minor|A lock collision is reported as "pathspec did not match any file(s)", naming the wrong cause for the operator|q-system/hooks/auto-commit.py:424
minor|The round-5 comment declares "EVERY suite that can be named IS named" repaired while q-system/.q-system/scripts/test/ (56 passed, collects clean) is in no entry and no written exclusion|.verify-suites:84
END FINDINGS

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-321-20260907-062036.md` (5800 bytes).

```
FINDINGS:
minor|The new test docstring cites kipi-update.sh:1714, which the same PR corrects to ~L1915 in the hook and calls unrelated env-var code|q-system/.q-system/tests/test_auto_commit.py:611
minor|The stdout-channel comment calls settings-template.json:395 the single wiring, but .claude/settings.json:395 also wires the hook and keeps stderr|q-system/hooks/auto-commit.py:577
END FINDINGS
```

--- reviewer output ---

## PR #321 review: auto-commit lock guard

Read-only. Nothing in the repo modified. All repros run in `$TMPDIR`.

---

### Findings

**minor** — the new test's docstring still cites `kipi-update.sh:1714`, the pointer this same PR corrected to `~L1915` in the hook and explicitly labelled "unrelated env-var code".

`q-system/.q-system/tests/test_auto_commit.py:611`

Repro and real output:

```
$ sed -n '611p' q-system/.q-system/tests/test_auto_commit.py
    """THE CONTROL THAT MATTERS MOST. kipi-update.sh:1714 force-deletes these locks

$ sed -n '1714p' kipi-update.sh
    export GIT_CONFIG_KEY_0=core.hooksPath

$ sed -n '504,505p' q-system/hooks/auto-commit.py
    top of its instance loop, no pid, mtime or age test at all (~L1915, sp-50119dec;
    an earlier draft of this docstring cited L1714, which is unrelated env-var code).
```

The force-delete loop is at `kipi-update.sh:1915`. Two files in one PR now name two different lines for the same scar, one of which the PR itself calls wrong. By this repo's own standard, a wrong pointer reads as coverage.

---

**minor** — the stdout-channel comment calls `settings-template.json:395` "the single wiring of it"; `.claude/settings.json:395` is a second wiring of the same hook, and it does *not* redirect stderr.

`q-system/hooks/auto-commit.py:577`

Repro and real output:

```
$ grep -rn "hooks/auto-commit.py" .claude/settings.json settings-template.json
.claude/settings.json:395:  "command": "python3 \"$CLAUDE_PROJECT_DIR/q-system/hooks/auto-commit.py\"",
settings-template.json:395: "command": "python3 \"$CLAUDE_PROJECT_DIR/q-system/hooks/auto-commit.py\" 2>/dev/null || true",
```

The load-bearing half of the claim holds: instances get the redirect, so stdout is the right channel. The "single wiring" half is false in kipi-system, where the live wiring keeps stderr. A maintainer reading this comment would edit one file and think the wiring changed everywhere.

---

### What is sound (attacks that failed)

**The `--git-dir` vs `--git-common-dir` claim survives mutation.** I rebuilt the hook seven times in a temp tree and ran the full suite against each mutant. Every mutant in the PR body killed, on the tests named:

```
M1 --git-common-dir      -> RED: test_a_worktree_does_not_refuse_on_the_main_checkouts_lock
                            RED: test_a_worktree_still_refuses_on_its_own_lock
M2 held = None           -> 5 RED
M3 guard deletes lock    -> RED: test_it_does_not_delete_the_lock_it_refuses_on
M4 drop HEAD.lock        -> RED: test_it_refuses_on_head_lock_too
M5 resolve vs cwd        -> RED: test_the_lock_dir_resolves_against_PROJ_DIR_not_the_process_cwd
M6 refusal to stderr     -> RED: test_a_refusal_reaches_the_channel_the_fleet_wiring_keeps
M7 fleet refusal stderr  -> RED: test_the_fleet_update_refusal_reaches_stdout_too
```

None of these tests is decoration.

**The disputed 447-497s lock figure is real for the shape this hook uses.** I ran a pathspec commit and a bare commit against a 3-second pre-commit and polled `.git` for locks:

```
pathspec commit:          locks observed -> ['index.lock', 'next-index-96552.lock']
bare commit (pre-staged): locks observed -> []
```

`commit_group` runs the pathspec form. The guard sees the window it claims to see, and HEAD.lock genuinely never appears for a commit, which matches the corrected docstring.

**The `--cached` removal is measured, not reasoned.** I reproduced both staging shapes:

```
staged-new file:           diff --name-only HEAD -> ['b.md']       (so removal loses nothing)
staged-then-deleted:       diff HEAD -> []  cached -> ['c.md']
   pathspec commit rc=1 "nothing to commit"; the staged entry survives untouched
```

Every claim in that comment checks out.

**Four hooks fired at once on one checkout do not lose or fake work.** With a 2-second pre-commit and two groups:

```
proc 2: committed: content ... | committed: chore ... | done
proc 0,1,3: skipped: ... Unable to create index.lock / pathspec did not match
claimed committed lines=2   actual new commits=2
git status --porcelain: (empty)
```

No false success, nothing left dirty. The TOCTOU that remains (four processes all pass the guard, three then lose to git) degrades to `skipped:`, which is the pre-existing behavior.

**No sibling Stop hook contends for the index.** I checked all six Stop-hook commands in both settings files; none runs git, so the "peer hook holds index.lock at every turn end" outage I went looking for does not exist.

**The `.verify-suites` round-5 claim is complete this time.** I enumerated every tracked `test_*.py` / `*_test.py` against the manifest entries. Eight directories remain uncovered, and every one is named in a written exclusion with its measurement (kipi-mcp, kipi-design, kipi-core, `.q-system/scripts`, `scripts/`, `test_token_guard_runtime.py`). I could not find a sixth instance of that finding.

**Suite state:** `46 passed` in-repo, `6 passed` for the run-marker suite, `verify` CI green at 9m10s.

**Blast radius:** the change only ever *refuses*. Work stays on disk, nothing is deleted, the lock-delete control is pinned, and an orphaned lock is cleared by `kipi-update.sh:1915` on the next fleet apply. Fully revertible.

---

Fix first: the `kipi-update.sh:1714` pointer at `test_auto_commit.py:611`. It is the one line in this PR that a future reader will act on and be wrong.

FINDINGS:
minor|The new test docstring cites kipi-update.sh:1714, which the same PR corrects to ~L1915 in the hook and calls unrelated env-var code|q-system/.q-system/tests/test_auto_commit.py:611
minor|The stdout-channel comment calls settings-template.json:395 the single wiring, but .claude/settings.json:395 also wires the hook and keeps stderr|q-system/hooks/auto-commit.py:577
END FINDINGS

PR #319, opened 2026-09-07
feat(review): claude owns the PR gate, codex becomes advisory

assafkip:
Reverses the 2026-07-29 directive that made codex the agent checking Sana's work. Founder directed it twice on 2026-09-06: *"forget codex use claude for fallback"* / *"forget codex go with the claude fallback"*.

## Why

Three stacked failures, each masking the next, measured rather than reasoned about:

1. `~/.codex/config.toml` asked for `gpt-6-astra`; the fleet is on Sol
2. `codex-cli` 0.147.0 was too old for that model, so the call returned HTTP 400
3. underneath both: **`ERROR: Your workspace is out of credits`**, which `gpt-5.6-sol` returns too

`codex exec` **exits 0** on all three. Two sessions reviewed for a full evening on the Opus fallback and nothing said so. An engine that fails silently cannot hold a required gate. Corrected diagnosis captured as `sp-b240fca2`.

## The cost, stated rather than glossed

Sana (the PR author) is Claude, so a Claude reviewer shares her lab and model family and re-derives her blind spots. Fresh context is not an independent mind. That was the whole argument for the original directive and it is **still true**. Availability decided this, not the argument.

Restoring independence is `KIPI_REVIEW_ENGINE=codex` **plus** `KIPI_REVIEW_PRIMARY_ENGINE=codex`, both together, once the workspace has credits.

## The trap this PR exists to avoid

`pr-review-agent.sh:246` posts `kipi/reviewer-approved` **only when `ENGINE == PRIMARY_ENGINE`**. Setting `KIPI_REVIEW_ENGINE=claude` alone leaves `PRIMARY_ENGINE=codex`, so every review lands on the advisory context and **every open PR waits forever on a status nobody posts**. Both defaults move together or the gate wedges.

## What moved, and what deliberately did not

| File | Change |
|---|---|
| `pr-review-agent.sh` | both defaults to claude; header rewritten with the reversal and its cost |
| `linear-worker.sh` | dispatches `--engine claude` at its single call site |
| `verify-codex-review-live.sh` | asserts claude is default and PRIMARY; the codex binary check drops from `fail` to a `NOTE`, since an advisory engine's absence must not fail a wiring check |
| `test-severity-floor.sh` | the two constants the file's own comment promised would carry a contract flip, plus Q7 / Q7D / Q9 and the verdict-record pair |
| `test-linear-worker-parallel.sh` | the ps-grep follows the dispatch |

[...]

PR #89, opened 2026-08-03
fix(review-gate): a disagreement never resolves toward approval (ASK-312)

assafkip:
Minimal cherry-pick of the ASK-312 `resolve_verdict` fix onto current `main`. Two commits, no unrelated work.

## The hole this closes

`origin/main:668` assigns `VERDICT="$DERIVED_VERDICT"` unconditionally. When the reviewer's stated verdict and the severity-derived verdict disagree, the derived one wins, including when it is the more approving of the two. That posts `kipi/reviewer-approved=success` on a PR the reviewer rejected.

## Scope correction, measured not assumed

ASK-274 (6fe7a3c) is already on main and it closed the **plan-echo** half of this. `review_is_usable` now classifies a declined-to-start transcript as unusable and derives nothing, so the 2026-08-03 Alice-shaped false green cannot recur on this base.

What ASK-274 cannot catch is a review that is **genuinely real**: it passes the usability gate, files only nits, and its author still said stop. The gate correctly lets it through and the ladder correctly derives an approving verdict from nits alone. `resolve_verdict` is the only thing left in that path.

So this is narrower than "the fleet can fake any green" and it is still a live hole.

## Why commit 2 exists

The cherry-pick landed **2 of 16 checks red**, correctly. The declined-to-start fixtures no longer reproduce the historical shape, because ASK-274 kills the precondition upstream:

```
declined-to-start-short.md -> usable NO, stated 'REQUEST CHANGES', derived ''
```

The old assertion demanded `derived = APPROVE`. Weakening it to accept the empty value would have made it vacuous, so the fidelity burden moved to a fixture that still reproduces a live disagreement. The two old cases stay as defence in depth.

New fixture, derived from the real round-2 codex transcript with **only** the severity column changed:

```
usable YES / stated 'REQUEST CHANGES' / derived 'APPROVE WITH NITS'
  -> resolve_verdict returns 'REQUEST CHANGES'
```

## Verification

- `test-review-gate-no-fake-green.sh` 19/19
- `test-review-plan-echo-gate.sh` 8/8, ASK-274 unregressed
- Manifest conflict resolved as a union; JSON validated, both test entries present

**Mutation-tested.** With `resolve_verdict` replaced by `printf '%s' "${2:-$1}"`, exactly one check goes red and it is the new one. The declined-to-start checks stay green against broken code. Without the new fixture this suite would have passed on a base where every remaining case is intercepted upstream.

[...]

PR #83, opened 2026-08-03
The unattended committer writes to a per-session WIP ref, not to whatever branch is checked out (ASK-314)

assafkip:
Closes ASK-314. RCA: `q-system/output/rca/rca-unattended-committer-misattribution-2026-08-02.md`.

## The defect

`auto-commit.py` (Stop hook) staged whatever was dirty and committed it to whatever branch was checked out, under the fixed subject `chore: update system infrastructure`. On 2026-08-02 that did three kinds of damage in one session:

- `7383d6c` swept finished ASK-122 work onto `sana/ask-294`, an unrelated issue's branch. Recovery cost a worktree extraction, a reland (`d20f412`) and a revert (`e770838`).
- `4559194` swept an in-flight ASK-312 fix plus an unrelated launchd config change into one generic commit, discarding the message its author was about to write.
- A config edit that lived only in the working tree was lost in a branch switch, leaving a live launchd spend dial at 10 while its template read 4.

All three follow from one property: **it wrote to HEAD.** A better subject line would not have helped, because the work would still be on the wrong branch.

## The fix

The writer keeps its durability job and loses its history-writing power. It builds the snapshot in a throwaway index (`GIT_INDEX_FILE`) and lands it on `refs/kipi/wip/<session-id>` via `commit-tree` + `update-ref` with a compare-and-swap. HEAD, the real index and the working tree are never touched, so the author still holds their dirty work and still writes their own message. Two sessions get two refs, so neither can sweep the other's uncommitted work. Recovery is `git checkout refs/kipi/wip/<id> -- <path>`, printed by the hook.

Skipped paths are now **named**, not silently dropped (the DoR's "also in scope, cheap"). `get_changed_files()` returns `(kept, skipped)` and every exit path reports what was left dirty. `snapshotted N files` alone reads as "everything dirty is safe", and the third incident above was precisely a working-tree-only edit nobody knew was still uncommitted.

The `[no-issue]` bypass is gone. That marker disarmed `linear-issue-ref`, the one gate positioned to catch this writer. The hook no longer writes history at all, so the gate's purpose is satisfied structurally rather than waived. `linear-first.md` updated to match. This commit passed `linear-issue-ref` with no bypass.

## Red then green, observed

Two hermetic tests, both declared in `capability-manifest.json`.

**`test-autocommit-wip-ref.py`** — the property, 17 assertions:

```
pre-fix hook (origin/main):   2 passed, 12 failed
fixed hook:                  17 passed,  0 failed
```

[...]

PR #80, opened 2026-08-03
test(pr-review): reproducer for the phantom finding parsed out of the prompt echo (ASK-317)

assafkip:
Autonomous worker (Sana) on ASK-317. Opened by the worker because the run ended without opening it.

Commits on this branch:
25340b5 test(pr-review): reproducer for the phantom finding parsed out of the prompt echo (ASK-317)
a1f3f7b fix(blocked-claim-lint): a claim is a sentence, so one line can hold two (ASK-317)
042dd5e fix(blocked-claim-lint): one fence settles one claim, and "unreachable" is a claim (ASK-317)
f91c2f7 feat(evidence-ledger): a blocked/unavailable claim carries the command that settles it (ASK-317)

Review runs next. Do not merge without it.

PR #322, opened 2026-09-07
fix(reviewer): status before comment, refuse a duplicate head, cap concurrency, stop on a repeated file (sp-fa810306, sp-0a09e013, sp-46726f79)

assafkip:
Four changes to `q-system/.q-system/scripts/pr-review-agent.sh`, one new suite, one existing suite corrected. Closes sp-fa810306, sp-0a09e013, sp-46726f79.

## The scar this comes from

PR #314 round 3 posted APPROVE WITH NITS as a PR comment at 00:29:47Z on 2026-09-07 on head `adc29516`. The process died before the commit status call. The next session read `gh api statuses adc29516` as empty, took empty for "never reviewed", re-ran the same head at 00:36:54Z, and the re-run returned REQUEST CHANGES. An approval that had already been earned was overturned by a kill. Two more rounds, about 90 minutes.

## What changed

**1. The commit status is posted BEFORE the PR comment (sp-fa810306).** A kill can land between any two calls; the only choice is which one survives. The comment is 60 KB of prose a human reads later. The status is what a required check, `converge.sh` and the next session all read. So the status goes first, unconditionally, and a second call re-posts the same context with the same `$VERDICT` and adds `target_url` once the comment lands. Both calls read one variable computed once, so a gate cannot see two different values; if the run dies between them the first post stands, missing only its link.

**2. A head that already carries a verdict is not re-reviewed (sp-fa810306).** Every posted comment now leads with `<!-- kipi-reviewer: head=<sha> -->`, and every run greps the PR's comments for that exact marker on the head it is about to review. A hit exits 3 with no dispatch and no status, unless `--force`. It keys on the reviewer's own marker rather than on the sha appearing in prose, because a sha appears inside diffs and inside quoted comments. It fails open: a gh outage yields zero hits and the review proceeds, because refusing on an unreadable comment list would wedge every review on a GitHub blip.

**3. A pid file per (repo, PR) and a concurrency cap (sp-0a09e013).** Written at start into `$HOME/.config/kipi/review-trees/<artifact_key>.pid`, beside the review tree and its lock, removed by the EXIT trap. Two or more live pid files refuse a third run with exit 3; `--queue` waits a bounded 900s and then refuses anyway. A pid the OS no longer knows is reclaimed as the scan walks, so one killed reviewer cannot wedge the loop. Five reviewers and watchers were killed under memory pressure on the evening of 2026-09-05 and nothing on the box could say how many were live.

**4. A STRUCTURAL stop when two rounds name one file (sp-46726f79).** After a REQUEST CHANGES verdict the run compares its findings against the previous REQUEST CHANGES round for the same PR, through `findings_block` (the one reader, sp-c0a9dac3), and prints a STRUCTURAL line naming the shared file and the two rounds when their cited paths intersect. The line rides inside the PR comment, and the run exits 4 **after** posting, so a queue script can stop while the review still reaches the PR. Rounds 1 to 3 of PR #314 each named another unwaited index write; only the founder stopped it.

## Test evidence: every negative case seen RED first, then green

RED runs were against `origin/main`'s `pr-review-agent.sh` extracted into a fixture root, with the same suite. Both suites drive the SHIPPING script copied into a fixture repo three levels down, against a `gh` stub, a fake engine and a `HOME` under `mktemp`. No live PR, no model spend, no real commit status.

| Case | RED (origin/main script) | GREEN (this branch) |
|---|---|---|
| status before comment | comment at gh-log line 3, status at line 4 | status at line 4, comment at line 5 |
| status keeps `target_url` | n/a | a later status call carries it |
| duplicate head refuses | rc=0, full review dispatched | rc=3, no status posted |
| `--force` bypass runs | rc=1 (unknown arg) | rc=0, status posted |
| sha in prose does not refuse | rc=0 (control) | rc=0 (control, unchanged) |
| two live pid files refuse | rc=0, third review ran | rc=3, no status posted |
| stale pid file reclaimed | left on disk | removed, run proceeds and posts |
| STRUCTURAL line printed | absent, rc=0 | printed, names file and both rounds, rc=4 |
| structural round still posts | n/a | comment and status both posted |
| disjoint files stay quiet | no line, rc=0 (control) | no line, rc=0 (control, unchanged) |

Exact runs:

```
bash q-system/.q-system/scripts/test/test-review-run-safety.sh
  PASS (22 checks) test-review-run-safety.sh
bash q-system/.q-system/scripts/test/test-severity-floor.sh
  PASS: 202/202 severity-floor checks
```

20 other suites naming `pr-review-agent` all pass: test-review-tree-guard, test-verdict-usability, test-review-comment-body, test-review-worktree-ref, test-review-plan-echo-gate, test-gh-repo-scope-reviewer, test-pr-review-root-guard, test-review-degraded-provenance, test-review-dry-run-labelled, test-review-gate-no-fake-green, test-review-invoker-provenance, test-review-artifact-repo-keyed, test-review-fallback-exact-sha, test-review-root-and-lock, test-review-early-bail-ref, test-findings-block-reader, test-review-redrive, test-verdict-statement-shape, test-script-stable-under-self-edit.

**Mutation test of the guard, not only the code.** Deleting the early `post_reviewer_status` call and re-running both suites turns both RED, naming the ordering:

```
test-severity-floor.sh    FAIL: the PR comment was posted BEFORE the commit status (comment at line 4, ...)
test-review-run-safety.sh FAIL: THE COMMENT WENT FIRST (status line 5, comment line 4)
```

Restored from a cp backup; `git diff --stat` empty afterwards; both suites green again.

## The existing suite that encoded the old ordering

`test-severity-floor.sh` asserted that the FIRST status call carries `target_url`. That was true only because the comment used to be posted first. It now reads the LAST status call for the link (which is what GitHub shows, since statuses append and the latest for a context wins) and separately asserts the ordering itself, so a revert to comment-first turns it red. 202/202.

## What is NOT verified here

- `capability-gate.py` refuses to run from a `.claude/worktrees` copy, so the new `expected_tests` manifest entry was not validated locally. It follows the exact shape of the existing entries and CI runs the gate from a normal checkout.
- `validate-separation.py 1 --verbose` exits 1 on this branch with two failures. Both are identical on origin/main's script, measured by swapping that file in and re-running: same file, same reported line 1266, same `1 gating, 7638 advisory` counts. Pre-existing, not introduced here.
- The STRUCTURAL check matches on the file path a finding cites, so two genuinely different defects in one long file read as one class, and a defect that moved files reads as two. It advises and does not gate, for that reason.
- Pid numbers are recycled. A stale file whose number was handed to an unrelated process reads as live and costs one false refusal, which is the safe direction.

## Ledger

sp-fa810306, sp-0a09e013 and sp-46726f79 are resolved on merge. sp-489f2d62 was captured on the way through: `read-first-gate.py` cannot be satisfied from an agent thread, measured on this session's own transcript.

### DIFF
--- a/q-system/.q-system/capability/expected_tests/q-system__.q-system__scripts__test__test-review-run-safety.sh.json
+++ b/q-system/.q-system/capability/expected_tests/q-system__.q-system__scripts__test__test-review-run-safety.sh.json
@@ -1,0 +1,4 @@
+{
+ "path": "q-system/.q-system/scripts/test/test-review-run-safety.sh",
+ "runner": "bash"
+}
--- a/q-system/.q-system/scripts/linear-worker.sh
+++ b/q-system/.q-system/scripts/linear-worker.sh
@@ -2477,8 +2477,20 @@ json.dump(d,open('$ATTEMPTS','w'),indent=2); print(e['rounds'])" 2>/dev/null || 
     # (sp-53aad86f). This is what makes a dispatcher-driven review distinguishable
     # from a hand run in the verdict record. It is set on the call rather than
     # exported once, so it cannot leak into an unrelated reviewer invocation.
-    KIPI_REVIEW_INVOKER=worker $REVIEWER_CMD "$PR_NUM" --issue "$ISSUE" --post --engine claude >>"$LOG" 2>&1 \
-      || say "WARN: the claude reviewer failed on PR #$PR_NUM (the PR stands, unreviewed)"
+    KIPI_REVIEW_INVOKER=worker $REVIEWER_CMD "$PR_NUM" --issue "$ISSUE" --post --engine claude >>"$LOG" 2>&1
+    REVIEWER_RC=$?
+    # EXIT 4 IS A COMPLETED REVIEW, NOT A FAILURE (sp-46726f79, 2026-09-07). The
+    # reviewer returns it AFTER posting its comment and its commit status, to say
+    # that this round's findings repeat a file from the previous REQUEST CHANGES
+    # round -- the wrong-fix-shape signal. Folding it into the `|| say WARN` line
+    # would log "the PR stands, unreviewed" about a review that fully landed and
+    # set the gate, which is a false statement in the one log an operator scans.
+    # Anything else non-zero keeps the original meaning.
+    if [ "$REVIEWER_RC" = "4" ]; then
+      say "NOTE: the reviewer posted its verdict on PR #$PR_NUM and flagged it STRUCTURAL -- this round repeats a file from the previous REQUEST CHANGES round. Sweep the whole class in one pass or park the PR; a third patch on the same file is the loop that warning exists to stop."
+    elif [ "$REVIEWER_RC" != "0" ]; then
+      say "WARN: the claude reviewer failed on PR #$PR_NUM (rc=$REVIEWER_RC; the PR stands, unreviewed)"
+    fi
     # Read back the verdict RECORD the reviewer just wrote (never re-grep the
     # review prose) and state what happens next in plain terms. Rework itself
     # fires on the NEXT run, through the severity-floor gate above.
--- a/q-system/.q-system/scripts/pr-review-agent.sh
+++ b/q-system/.q-system/scripts/pr-review-agent.sh
@@ -81,8 +81,20 @@
 # which status context it posts, and which directory its artifacts land in.
 #
 # Usage:  pr-review-agent.sh <pr-number> [--issue ASK-nnn] [--post]
-#                            [--engine claude|codex]
+#                            [--engine claude|codex] [--force] [--queue]
 #         --post also comments the review on the PR and the Linear issue.
+#         --force runs even when this head already carries a verdict comment.
+#         --queue waits for a concurrency slot instead of refusing immediately.
+#
+# EXIT CODES, and they are part of the contract because a queue script reads them:
+#   0  reviewed (or dry-run surveyed); nothing further is advised
+#   1  could not review: bad arguments, no PR, a moving head, no tree at the sha
+#   2  refused at resolution: this copy does not live at a repo root
+#   3  REFUSED TO START: this head already has a verdict, or two reviews are
+#      already live. Nothing was dispatched and no status was posted.
+#   4  ran and posted, AND the findings repeat a file from the previous
+#      REQUEST CHANGES round. The post is complete; the next patch is the wrong
+#      move (sp-46726f79). Advisory to the caller, never a block on the post.
 set -uo pipefail
 
 SCRIPT_DIR="$(cd "$(dirname "${BASH_SOURCE[0]}")" && pwd)"
@@ -156,10 +168,17 @@ CODEX_MODEL="${KIPI_REVIEW_CODEX_MODEL:-gpt-5.6-sol}"
 # config change (`KIPI_REVIEW_ENGINE=claude`), not an edit to the script that
 # gates every PR in the repo.
 PR=""; ISSUE=""; POST=0; ENGINE="${KIPI_REVIEW_ENGINE:-claude}"; TARGET_REPO_ARG=""
+# --force and --queue are the two documented ways past the two refusals added for
+# sp-fa810306 and sp-0a09e013. Both are opt-in and neither has an env form: an
+# env default would let a scheduled caller inherit the bypass without anyone
+# having typed it, which is how a guard ends up permanently off.
+FORCE=0; QUEUE=0
 while [ $# -gt 0 ]; do
   case "$1" in
     --issue)  shift; ISSUE="${1:-}" ;;
     --engine) shift; ENGINE="${1:-}" ;;
+    --force)  FORCE=1 ;;
+    --queue)  QUEUE=1 ;;
     # WHICH REPO THE WORK IS IN (ASK-738). $SKEL is where this CODE lives; it is
     # not where the PR lives. Before this argument the two were the same
     # variable, so a review of an external PR number resolved against the home
@@ -219,7 +238,98 @@ GH_R_PROMPT=""
 # form, which is what every pre-ASK-738 caller and fixture already relies on.
 STATUS_REPO_PATH="{owner}/{repo}"
 [ -n "$REVIEW_SLUG" ] && STATUS_REPO_PATH="$REVIEW_SLUG"
-[ -n "$PR" ] || { echo "usage: pr-review-agent.sh <pr-number> [--issue ASK-nnn] [--post] [--engine claude|codex]" >&2; exit 1; }
+[ -n "$PR" ] || { echo "usage: pr-review-agent.sh <pr-number> [--issue ASK-nnn] [--post] [--engine claude|codex] [--force] [--queue]" >&2; exit 1; }
+
+# ONE PID FILE PER (REPO, PR), AND NEVER A THIRD CONCURRENT REVIEW (sp-0a09e013).
+#
+# THE SCAR, measured: on the evening of 2026-09-05 five reviewers and watchers
+# started as tracked tool calls were killed by the harness under memory pressure,
+# and on 2026-09-07 at 00:35Z a process exit took a nohup'd reviewer with it,
+# orphaning two tasks and losing a posted approval. Nothing on this box could
+# answer "how many reviews are live right now", so the answer was always "start
+# another one" -- and each one costs a full Opus review of a real diff.
+#
+# THE FILE IS THE ANSWER, not a process table scan. `pgrep -f pr-review-agent`
+# matches its own grep and every editor buffer holding the name, which is the
+# same class of miscount this fleet already recorded. A file written at start and
+# removed by the EXIT trap is checkable by anyone, survives the session that
+# spawned it, and self-heals: a pid the OS no longer knows is stale and is
+# reclaimed, because an operator who must clear these by hand eventually clears
+# one while a run is live.
+#
+# SAME DIRECTORY AS THE REVIEW TREES AND THEIR LOCKS, deliberately. That is the
+# one place this script already owns per (repo, PR) state, so a reader looking
+# for "what is this reviewer doing" finds the tree, the lock and the pid together
+# instead of in three conventions.
+#
+# HONEST BOUNDARY: pid numbers are recycled. A stale file whose number has been
+# handed to an unrelated process reads as live and costs one false refusal, which
+# is the safe direction -- a false refusal names the file and exits 3, while a
+# false start is the duplicate review this exists to stop.
+REVIEW_PID_DIR="$(dirname "$(review_tree_path "$HOME/.config/kipi" "$REVIEW_SLUG" "$PR")")"
+REVIEW_PID_FILE="$REVIEW_PID_DIR/$(artifact_key "$REVIEW_SLUG" "$PR").pid"
+REVIEW_PID_WRITTEN=0
+# CONCURRENCY CAP: at most 2 live runs, so this one is refused when 2 already are.
+REVIEW_CONCURRENCY_CAP="${KIPI_REVIEW_CONCURRENCY_CAP:-2}"
+REVIEW_QUEUE_SECONDS="${KIPI_REVIEW_QUEUE_SECONDS:-900}"
+
+_review_pid_release() {
+  # ONLY OUR OWN. Removing a pid file whose contents are someone else's pid would
+  # release THEIR slot, which is the concurrent-reviewer defect wearing a cleanup
+  # costume -- the same reasoning release_wt_lock already carries below.
+  [ "$REVIEW_PID_WRITTEN" = "1" ] || return 0
+  if [ "$(tr -dc '0-9' < "$REVIEW_PID_FILE" 2>/dev/null | head -c 12)" = "$$" ]; then
+    command rm -f "$REVIEW_PID_FILE" 2>/dev/null || true
+  fi
+  REVIEW_PID_WRITTEN=0
+}
+
+# _live_review_pids -> one "<pidfile> <pid>" line per LIVE run other than ours.
+# Reaps stale files as it goes: the scan is the only thing that ever runs after a
+# killed reviewer, so if it does not clean up, nothing does.
+_live_review_pids() {
+  local f pid
+  for f in "$REVIEW_PID_DIR"/*.pid; do
+    [ -f "$f" ] || continue
+    [ "$f" = "$REVIEW_PID_FILE" ] && continue
+    pid="$(tr -dc '0-9' < "$f" 2>/dev/null | head -c 12)"
+    if [ -n "$pid" ] && kill -0 "$pid" 2>/dev/null; then
+      printf '%s %s\n' "$f" "$pid"
+    else
+      echo "  reclaimed a stale reviewer pid file (pid ${pid:-<empty>} is gone): $f" >&2
+      command rm -f "$f" 2>/dev/null || true
+    fi
+  done
+}
+
+mkdir -p "$REVIEW_PID_DIR" 2>/dev/null || true
+_REVIEW_WAITED=0
+while :; do
+  _LIVE="$(_live_review_pids)"
+  _LIVE_N="$(printf '%s' "$_LIVE" | grep -c . || true)"
+  [ "${_LIVE_N:-0}" -lt "$REVIEW_CONCURRENCY_CAP" ] && break
+  if [ "$QUEUE" != "1" ]; then
+    echo "REFUSING: $_LIVE_N reviews are already live (cap $REVIEW_CONCURRENCY_CAP). No review was dispatched and NO status was posted." >&2
+    printf '%s\n' "$_LIVE" | sed 's/^/  live: /' >&2
+    echo "  Wait for one to finish, or re-run with --queue to wait up to ${REVIEW_QUEUE_SECONDS}s for a slot." >&2
+    exit 3
+  fi
+  # BOUNDED WAIT, NOT AN OPEN ONE. An unbounded queue turns a stuck reviewer into
+  # a pile of blocked reviewers that all wake at once when it dies.
+  if [ "$_REVIEW_WAITED" -ge "$REVIEW_QUEUE_SECONDS" ]; then
+    echo "REFUSING: waited ${_REVIEW_WAITED}s for a slot and $_LIVE_N reviews are still live (cap $REVIEW_CONCURRENCY_CAP). No review was dispatched and NO status was posted." >&2
+    exit 3
+  fi
+  [ "$_REVIEW_WAITED" = "0" ] && echo "$_LIVE_N reviews live (cap $REVIEW_CONCURRENCY_CAP); --queue given, waiting up to ${REVIEW_QUEUE_SECONDS}s for a slot..."
+  sleep 10
+  _REVIEW_WAITED=$(( _REVIEW_WAITED + 10 ))
+done
+printf '%s' "$$" > "$REVIEW_PID_FILE" 2>/dev/null && REVIEW_PID_WRITTEN=1
+# ARMED BEFORE THE FIRST THING THAT CAN EXIT. Every refusal below this line is a
+# path the pid file has to survive, and the worktree lock's own trap further down
+# REPLACES this one, so that line clears both. A trap set after the first exit
+# path is a trap that leaks on exactly the failures it exists for.
+trap '_review_pid_release' EXIT
 
 # WHAT THE ENGINE CHANGES. Everything else below this block is shared, which is
 # the whole reason this is a flag and not a second script.
@@ -298,6 +408,18 @@ REVIEW_UNUSABLE=0
 
 mkdir -p "$ENGINE_DIR" "$VERDICT_DIR"
 TS() { date -u +%Y-%m-%dT%H:%M:%SZ; }
+
+# THE ONE DEFINITION of the marker that says "this head already has a verdict"
+# (sp-fa810306). Written into every posted comment, read back before every run.
+# An HTML comment because GitHub does not render it: the marker is for machines,
+# and a visible one would be the first thing a human asked to have removed.
+#
+# WHY A MARKER AND NOT THE SHA IN PROSE. A sha appears inside a diff, inside a
+# quoted earlier comment, and inside anything a person pastes, so grepping the
+# comment text for the sha fires on comments that are not verdicts. WHY NOT THE
+# LOCAL VERDICT RECORD: the whole scar is a local record disagreeing with what
+# GitHub holds, so the check has to read GitHub.
+review_head_marker() { printf '<!-- kipi-reviewer: head=%s -->' "${1:-}"; }
 REVIEW="$ENGINE_DIR/$(artifact_key "$REVIEW_SLUG" "$PR")-$(date +%Y%m%d-%H%M%S).md"
 
 # Same bash wall clock as the worker: macOS ships no `timeout` without coreutils,
@@ -513,7 +635,9 @@ release_wt_lock() {
   fi
   _wt_lock_path=""
 }
-trap 'release_wt_lock' EXIT
+# REPLACES the pid-only trap armed above; bash keeps one EXIT trap, so this line
+# has to clear BOTH or the pid file leaks from here on.
+trap 'release_wt_lock; _review_pid_release' EXIT
 
 review_worktree() {  # review_worktree <sha> -> prints path, or nothing
   # KEYED BY REPO AND PR (ASK-738). One shared review-trees/pr-<N> path meant
--- a/q-system/.q-system/scripts/pr-review-agent.sh
+++ b/q-system/.q-system/scripts/pr-review-agent.sh
@@ -273,26 +273,35 @@ REVIEW_PID_WRITTEN=0
 REVIEW_CONCURRENCY_CAP="${KIPI_REVIEW_CONCURRENCY_CAP:-2}"
 REVIEW_QUEUE_SECONDS="${KIPI_REVIEW_QUEUE_SECONDS:-900}"
 
+_pid_in_file() { tr -dc '0-9' < "${1:-}" 2>/dev/null | head -c 12; }
+
 _review_pid_release() {
   # ONLY OUR OWN. Removing a pid file whose contents are someone else's pid would
   # release THEIR slot, which is the concurrent-reviewer defect wearing a cleanup
   # costume -- the same reasoning release_wt_lock already carries below.
   [ "$REVIEW_PID_WRITTEN" = "1" ] || return 0
-  if [ "$(tr -dc '0-9' < "$REVIEW_PID_FILE" 2>/dev/null | head -c 12)" = "$$" ]; then
+  if [ "$(_pid_in_file "$REVIEW_PID_FILE")" = "$$" ]; then
     command rm -f "$REVIEW_PID_FILE" 2>/dev/null || true
   fi
   REVIEW_PID_WRITTEN=0
 }
 
-# _live_review_pids -> one "<pidfile> <pid>" line per LIVE run other than ours.
+# _live_review_pids -> one "<pidfile> <pid>" line per LIVE run that is not US.
 # Reaps stale files as it goes: the scan is the only thing that ever runs after a
 # killed reviewer, so if it does not clean up, nothing does.
+#
+# SKIPS BY PID, NEVER BY PATH (round 1 major, 2026-09-07). The first cut skipped
+# `$REVIEW_PID_FILE` by name, which is exactly the file a SECOND RUN ON THE SAME
+# PR writes -- so the one collision that matters most was the one collision the
+# scan could not see, and the second run then overwrote the incumbent's pid,
+# silently taking over its slot while both read the same detached worktree.
+# Identity is the pid, so that is what the skip compares.
 _live_review_pids() {
   local f pid
   for f in "$REVIEW_PID_DIR"/*.pid; do
     [ -f "$f" ] || continue
-    [ "$f" = "$REVIEW_PID_FILE" ] && continue
-    pid="$(tr -dc '0-9' < "$f" 2>/dev/null | head -c 12)"
+    pid="$(_pid_in_file "$f")"
+    [ "$pid" = "$$" ] && continue
     if [ -n "$pid" ] && kill -0 "$pid" 2>/dev/null; then
       printf '%s %s\n' "$f" "$pid"
     else
@@ -305,22 +314,36 @@ _live_review_pids() {
 mkdir -p "$REVIEW_PID_DIR" 2>/dev/null || true
 _REVIEW_WAITED=0
 while :; do
+  # The scan runs first because it is what reclaims stale files, including this
+  # PR's own -- so anything still on disk afterwards holds a pid that is alive.
   _LIVE="$(_live_review_pids)"
   _LIVE_N="$(printf '%s' "$_LIVE" | grep -c . || true)"
-  [ "${_LIVE_N:-0}" -lt "$REVIEW_CONCURRENCY_CAP" ] && break
+  _MINE="$(_pid_in_file "$REVIEW_PID_FILE")"
+  _BLOCK=""
+  if [ -n "$_MINE" ] && [ "$_MINE" != "$$" ]; then
+    # A SECOND RUN ON THE SAME PR IS REFUSED BELOW THE CAP, and that is not the
+    # cap being strict. Two runs on one PR share ONE detached review tree and ONE
+    # pid file: the second re-checkouts the tree while the first is reading files
+    # out of it, and both stamp verdicts with the sha they think they read. The
+    # numeric cap cannot express that, because one incumbent is under it.
+    _BLOCK="a review of PR #$PR is already live (pid $_MINE). Two runs on one PR share one detached worktree and one pid file, so the second would re-checkout that tree under the first."
+  elif [ "${_LIVE_N:-0}" -ge "$REVIEW_CONCURRENCY_CAP" ]; then
+    _BLOCK="$_LIVE_N reviews are already live (cap $REVIEW_CONCURRENCY_CAP)."
+  fi
+  [ -z "$_BLOCK" ] && break
   if [ "$QUEUE" != "1" ]; then
-    echo "REFUSING: $_LIVE_N reviews are already live (cap $REVIEW_CONCURRENCY_CAP). No review was dispatched and NO status was posted." >&2
+    echo "REFUSING: $_BLOCK No review was dispatched and NO status was posted." >&2
     printf '%s\n' "$_LIVE" | sed 's/^/  live: /' >&2
-    echo "  Wait for one to finish, or re-run with --queue to wait up to ${REVIEW_QUEUE_SECONDS}s for a slot." >&2
+    echo "  Wait for it to finish, or re-run with --queue to wait up to ${REVIEW_QUEUE_SECONDS}s for a slot." >&2
     exit 3
   fi
   # BOUNDED WAIT, NOT AN OPEN ONE. An unbounded queue turns a stuck reviewer into
   # a pile of blocked reviewers that all wake at once when it dies.
   if [ "$_REVIEW_WAITED" -ge "$REVIEW_QUEUE_SECONDS" ]; then
-    echo "REFUSING: waited ${_REVIEW_WAITED}s for a slot and $_LIVE_N reviews are still live (cap $REVIEW_CONCURRENCY_CAP). No review was dispatched and NO status was posted." >&2
+    echo "REFUSING: waited ${_REVIEW_WAITED}s for a slot and it is still blocked: $_BLOCK No review was dispatched and NO status was posted." >&2
     exit 3
   fi
-  [ "$_REVIEW_WAITED" = "0" ] && echo "$_LIVE_N reviews live (cap $REVIEW_CONCURRENCY_CAP); --queue given, waiting up to ${REVIEW_QUEUE_SECONDS}s for a slot..."
+  [ "$_REVIEW_WAITED" = "0" ] && echo "blocked: $_BLOCK --queue given, waiting up to ${REVIEW_QUEUE_SECONDS}s for a slot..."
   sleep 10
   _REVIEW_WAITED=$(( _REVIEW_WAITED + 10 ))
 done
@@ -1543,6 +1566,18 @@ if [ "$POST" = "1" ]; then
   # them separate is what makes the degraded path safe: with no temp file we post
   # the raw review unchanged and write nothing, instead of redirecting into the
   # artifact we are trying to read.
+  # THE MARKER MEANS "A VERDICT WAS PUBLISHED ON THIS HEAD", NOT "A RUN HAPPENED"
+  # (round 1 major, 2026-09-07). The first cut stamped it on every posted comment,
+  # UNUSABLE reviews included. An unusable review posts state=failure and states no
+  # verdict, so that head then carried a red required check AND a marker that made
+  # every automated re-review exit 3 -- a wedge no unattended path could clear,
+  # because --force is a hand-typed flag. The guard exists to stop a re-review from
+  # overturning an EARNED verdict; a review nobody could read earned nothing, so it
+  # must not claim the head.
+  MARK_HEAD=0
+  if [ "$REVIEW_USABLE" = "1" ] && [ -n "$VERDICT" ]; then MARK_HEAD=1; fi
+  [ "$MARK_HEAD" = "1" ] \
+    || echo "  NOTE: this review is unusable or states no verdict, so the comment carries NO head marker and a later run may re-review ${HEAD_SHA:0:8} without --force."
   if [ -n "$REVIEW_BODY" ]; then
     # THE MARKER LEADS THE BODY, and the position is load-bearing. It is what the
     # duplicate-run refusal at the top of this script greps for, and
@@ -1550,7 +1585,7 @@ if [ "$POST" = "1" ]; then
     # body would be the first thing a long review dropped -- the guard would then
     # be silently disarmed on exactly the reviews that run longest.
     {
-      review_head_marker "$HEAD_SHA"; printf '\n\n'
+      [ "$MARK_HEAD" = "1" ] && { review_head_marker "$HEAD_SHA"; printf '\n\n'; }
       review_comment_body "$REVIEW" "$VERDICT" "$ENGINE" "$DEGRADED"
       # The structural warning rides INSIDE the comment rather than only in this
       # log, because the log belongs to whoever ran the reviewer and the PR is
@@ -1560,7 +1595,7 @@ if [ "$POST" = "1" ]; then
     POST_FILE="$REVIEW_BODY"
   else
     POST_FILE="$REVIEW"
-    echo "  WARN: could not create a temp file; posting the RAW review, which GitHub may reject on size. The head marker is NOT in this body, so a later run will not see this verdict and may re-review the same head (sp-fa810306)." >&2
+    echo "  WARN: could not create a temp file; posting the RAW review, which GitHub may reject on size. The head marker is NOT in this body, so a later run will not see this verdict and may re-review the same head (sp-fa810306). Erring toward a duplicate round is the safe direction here: the other way is a wedge no unattended run can clear." >&2
   fi
   # Keep the reason. A bare "could not comment" sent the maintainer to guess
   # between a size rejection, an auth failure and a closed PR -- the same
--- a/q-system/.q-system/scripts/pr-review-agent.sh
+++ b/q-system/.q-system/scripts/pr-review-agent.sh
@@ -1284,9 +1284,27 @@ FINAL_EXIT=0
 # _finding_files <review-file> -> the deduped, sorted file paths its findings cite.
 # Rows are `severity|claim|file:line`; the line number is dropped because a fix
 # that moves a defect down three lines is the same defect.
+#
+# `cut`, NOT awk, and that is not a style preference (CI red, 2026-09-07). The
+# first cut read the third field as awk's `$3`, and validate-separation's
+# propagation-leak classifier reads `$` followed by a digit as a CURRENCY token:
+# the line landed in test-propagation-leak-baseline.py as a new `pricing` fact
+# the committed baseline did not account for, and turned `validate` red on a
+# change that has nothing to do with pricing. A shell field split says the same
+# thing with no character the fleet's own detectors read as money.
+#
+# THE `grep -E` IS LOAD-BEARING, and the disjoint-files control is what caught it.
+# `cut -d'|' -f3` returns the WHOLE LINE when the line has no delimiter, and
+# findings_block's output carries the `FINDINGS:` / `END FINDINGS` markers. So the
+# first cut compared those two marker strings as if they were file paths, found
+# them in both rounds, and printed STRUCTURAL on two rounds that shared nothing.
+# awk's `NF>=3` had been doing this filtering implicitly; moving off awk moved the
+# guard too. Two pipes required first, then the third field means something.
 _finding_files() {
   findings_block "${1:-}" \
-    | awk -F'|' 'NF>=3 { loc=$3; sub(/:[0-9-]*$/, "", loc); gsub(/^[ \t]+|[ \t]+$/, "", loc); if (loc != "") print loc }' \
+    | grep -E '^[^|]*\|[^|]*\|' \
+    | cut -d'|' -f3 \
+    | sed -e 's/:[0-9-]*$//' -e 's/^[[:space:]]*//' -e 's/[[:space:]]*$//' -e '/^$/d' \
     | sort -u
 }
 
--- a/q-system/.q-system/scripts/pr-review-agent.sh
+++ b/q-system/.q-system/scripts/pr-review-agent.sh
@@ -474,6 +474,38 @@ if [ -n "$HEAD_SHA_CONFIRM" ] && [ "$HEAD_SHA_CONFIRM" != "$HEAD_SHA" ]; then
   echo "  Re-run once the branch settles. No review was dispatched and NO status was posted." >&2
   exit 1
 fi
+# A HEAD THAT ALREADY CARRIES A VERDICT IS NOT RE-REVIEWED (sp-fa810306).
+#
+# THE SCAR, exact: on 2026-09-07 round 3 of PR #314 posted APPROVE WITH NITS as a
+# comment at 00:29:47Z on head adc29516, the process died before the commit status
+# landed, and the next session read `gh api statuses adc29516` as empty. Empty
+# looked like "never reviewed", so the same head was re-run at 00:36:54Z -- and
+# that re-run returned REQUEST CHANGES, overturning an approval that had already
+# been earned. Two more rounds and about 90 minutes.
+#
+# THE COMMENT IS THE DURABLE RECORD, NOT THE STATUS, and that is the whole point
+# of checking it here. The status post is the thing that was lost; the comment is
+# the thing that survived. So this asks the question the re-running session should
+# have asked: has a reviewer already published a verdict on THIS EXACT COMMIT?
+# Posting the status BEFORE the comment (further down) closes the other half.
+#
+# FAILS OPEN. A gh outage, a network error, or a PR with no comments all yield
+# zero hits and the review proceeds. Refusing on an unreadable comment list would
+# wedge every review on a GitHub blip, which is worse than one duplicate round --
+# and the duplicate round is what the caller asked for in the first place.
+if [ -n "$HEAD_SHA" ]; then
+  DUP_MARKER="$(review_head_marker "$HEAD_SHA")"
+  DUP_HITS="$(gh pr view "$PR" $KIPI_GH_REPO_ARGS --json comments -q '.comments[].body' 2>/dev/null | grep -F -c -- "$DUP_MARKER" || true)"
+  DUP_HITS="$(printf '%s' "${DUP_HITS:-0}" | tr -dc '0-9')"
+  [ -n "$DUP_HITS" ] || DUP_HITS=0
+  if [ "$DUP_HITS" -gt 0 ] && [ "$FORCE" != "1" ]; then
+    echo "REFUSING: PR #$PR head ${HEAD_SHA:0:8} already carries $DUP_HITS reviewer verdict comment(s). Re-reviewing a head that has already been judged can only overturn a verdict that was already earned (sp-fa810306). No review was dispatched and NO status was posted." >&2
+    echo "  Read the existing verdict on the PR, or pass --force if a second opinion on the same commit is what you actually want." >&2
+    exit 3
+  fi
+  [ "$DUP_HITS" -gt 0 ] && echo "  NOTE: ${HEAD_SHA:0:8} already carries $DUP_HITS verdict comment(s); --force given, reviewing it again."
+fi
+
 [ -n "$ISSUE" ] || ISSUE="$(printf '%s' "$PR_TITLE" | grep -oE 'ASK-[0-9]+' | head -1)"
 
 echo "$(TS) reviewing PR #$PR: $PR_TITLE"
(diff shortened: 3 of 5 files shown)


assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-322-20260907-140709.md` (1053 bytes).

```
FINDINGS:
major|The concurrency cap is blind to a second run on the same PR, and that run deletes the live incumbent's pid file|q-system/.q-system/scripts/pr-review-agent.sh:284
major|A head whose review was unusable keeps a failing required status and a marker comment, so no automated re-review can ever run on it|q-system/.q-system/scripts/pr-review-agent.sh:301
minor|Exit 4 after a fully successful post is logged by linear-worker.sh as "the claude reviewer failed ... the PR stands, unreviewed"|q-system/.q-system/scripts/pr-review-agent.sh:1655
END FINDINGS
```

--- reviewer output ---

The second suite finished. Both of the PR's claimed runs are now confirmed by me directly, not taken from the description:

- `test-review-run-safety.sh` → `PASS (22 checks)`
- `test-severity-floor.sh` → `PASS: 202/202 severity-floor checks`

That closes the one gap in my "What is sound" section. Nothing else in the review moves: the three findings stand as reproduced, and the verdict stays **REQUEST CHANGES**, with Major 2 (the unusable-review wedge) as the first thing to fix.

FINDINGS:
major|The concurrency cap is blind to a second run on the same PR, and that run deletes the live incumbent's pid file|q-system/.q-system/scripts/pr-review-agent.sh:284
major|A head whose review was unusable keeps a failing required status and a marker comment, so no automated re-review can ever run on it|q-system/.q-system/scripts/pr-review-agent.sh:301
minor|Exit 4 after a fully successful post is logged by linear-worker.sh as "the claude reviewer failed ... the PR stands, unreviewed"|q-system/.q-system/scripts/pr-review-agent.sh:1655
END FINDINGS

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-322-20260907-144236.md` (313025 bytes).

```
FINDINGS:
major|A rejected commit-status POST still publishes a duplicate-review marker, permanently refusing unattended retries until a human forces the review|q-system/.q-system/scripts/pr-review-agent.sh:1595
major|Any public commenter can forge the unauthenticated head marker and suppress unattended review of that commit|q-system/.q-system/scripts/pr-review-agent.sh:521
END FINDINGS
```

--- reviewer output, last 54611 bytes of 313025 (full review at the path above) ---

   526	    echo "  Read the existing verdict on the PR, or pass --force if a second opinion on the same commit is what you actually want." >&2
   527	    exit 3
   528	  fi
   529	  [ "$DUP_HITS" -gt 0 ] && echo "  NOTE: ${HEAD_SHA:0:8} already carries $DUP_HITS verdict comment(s); --force given, reviewing it again."
   530	fi
   531	
   532	[ -n "$ISSUE" ] || ISSUE="$(printf '%s' "$PR_TITLE" | grep -oE 'ASK-[0-9]+' | head -1)"
   533	
   534	echo "$(TS) reviewing PR #$PR: $PR_TITLE"
   535	echo "  head sha under review: ${HEAD_SHA:-unknown}"
   536	[ -n "$ISSUE" ] && echo "  linked issue: $ISSUE"
   537	
   538	# THE TREE MUST ACTUALLY CONTAIN THE PR (sp-a72a9567). $SKEL comes from this
   539	# script's own location, and the diff comes from `gh pr diff <N>` -- two
   540	# independent sources that nothing was checking against each other. Run from
   541	# worktree A against a PR on branch B and the reviewer reads A's files off disk
   542	# while B's diff scrolls past, then writes a verdict record and a commit status
   543	# attributing its findings to B's head sha.
   544	#
   545	# Not hypothetical. 2026-07-29, run from the ask-221 worktree against PR #35: it
   546	# returned three findings in linear-sync.py, a file PR #35's diff does not touch at
   547	# all. The findings were real bugs in ask-221; the PROVENANCE was false. That is
   548	# worse than a wrong verdict, because the record looks authoritative.
   549	#
   550	# TWO TIERS, because a flat equality check would be wrong twice over. The PR's head
   551	# may legitimately be BEHIND local HEAD (a push landing after the `gh pr view`
   552	# above), so equality would refuse healthy runs -- ancestry is the real question.
   553	# And an UNKNOWN object is not evidence of a mismatch: a stale or partial clone
   554	# cannot prove ancestry either way, and inventing a refusal there would wedge the
   555	# loop on a fetch problem. Unknown warns; known-but-unrelated refuses.
   556	#
   557	# WHY THIS RESOLVES INSTEAD OF REFUSING (codex review round 1 of PR #34, major).
   558	# The first cut of this guard compared $HEAD_SHA against $SKEL's HEAD and exited 1
   559	# on a mismatch. That reads correctly and is still wrong, because $SKEL is derived
   560	# from BASH_SOURCE -- the script's own location -- and the autonomous caller is
  1310	
  1311	case "$VERDICT" in
  1312	  "REQUEST CHANGES"|"BLOCK")
  1313	    PREV_REVIEW=""; PREV_ROUND=0; CUR_ROUND=0; _sr_n=0
  1314	    while IFS= read -r _sr_f; do
  1315	      [ -n "$_sr_f" ] || continue
  1316	      _sr_n=$(( _sr_n + 1 ))
  1317	      if [ "$_sr_f" = "$REVIEW" ]; then CUR_ROUND="$_sr_n"; continue; fi
  1318	      _sr_v="$(resolve_verdict "$(extract_verdict "$_sr_f")" "$(verdict_from_findings "$_sr_f")")"
  1319	      case "$_sr_v" in
  1320	        "REQUEST CHANGES"|"BLOCK") PREV_REVIEW="$_sr_f"; PREV_ROUND="$_sr_n" ;;
  1321	      esac
  1322	    done <<EOF
  1323	$(ls -1 "$ENGINE_DIR/$(artifact_key "$REVIEW_SLUG" "$PR")-"*.md 2>/dev/null | sort)
  1324	EOF
  1325	    if [ -n "$PREV_REVIEW" ]; then
  1326	      _sr_prev="$(mktemp "${TMPDIR:-/tmp}/pr-review-prev.XXXXXX" 2>/dev/null)" || _sr_prev=""
  1327	      _sr_cur="$(mktemp "${TMPDIR:-/tmp}/pr-review-cur.XXXXXX" 2>/dev/null)" || _sr_cur=""
  1328	      if [ -n "$_sr_prev" ] && [ -n "$_sr_cur" ]; then
  1329	        _finding_files "$PREV_REVIEW" > "$_sr_prev"
  1330	        _finding_files "$REVIEW"      > "$_sr_cur"
  1331	        # `comm -12` needs both sides sorted, which _finding_files guarantees.
  1332	        _sr_shared="$(comm -12 "$_sr_prev" "$_sr_cur" 2>/dev/null | tr '\n' ' ' | sed 's/ *$//')"
  1333	        command rm -f "$_sr_prev" "$_sr_cur" 2>/dev/null || true
  1334	        if [ -n "$_sr_shared" ]; then
  1335	          STRUCTURAL_LINE="STRUCTURAL: round $CUR_ROUND and round $PREV_ROUND both request changes in the same file(s): $_sr_shared. Two rounds naming one file is a wrong fix SHAPE, not two bugs. Sweep the whole class in one pass or park the PR; a third patch on the same file is the loop this warning exists to stop (sp-46726f79)."
  1336	          FINAL_EXIT=4
  1337	          echo "  $STRUCTURAL_LINE"
  1338	        fi
  1339	      else
  1340	        echo "  WARN: could not create temp files for the structural comparison; skipping it. Two rounds on one file will not be flagged on this run." >&2
  1341	      fi
  1342	    fi
  1343	    ;;
  1344	esac
  1345	
  1346	# Single writer for verdict state. The worker's rework gate reads THIS record,
  1347	# never the review prose. Keyed by PR number, latest round wins; history stays
  1348	# in the timestamped .md files.
  1349	#
  1350	# head_sha is the commit this review actually examined, captured before the
  1351	# reviewer ran (ASK-216). Without it the record binds an approval to a PR
  1352	# NUMBER, and the worker reuses one PR across rounds, so any later push inherits
  1353	# the approval. The key is ALWAYS written -- empty when `gh` could not answer --
  1354	# because rework_gate reads empty as "unknown, fall back and say so", and a
  1355	# key that sometimes vanishes is a shape the reader would have to guess at.
  1356	#
  1357	# The record lands in $VERDICT_DIR, NOT next to the review. Those are the same
  1358	# directory only for a non-primary engine; for the gating engine the reviews live
  1359	# in $OUT_DIR/codex (its own round counter) while the record must land in $OUT_DIR
  1360	# where converge.sh and linear-worker.sh actually read it.
  1361	# DID A REVIEW ACTUALLY HAPPEN, PERSISTED (sp-2a832233, ASK-352). The record used
  1362	# to store only a PATH to the review, and the review files rotate, so every
  1363	# consumer downstream had to re-derive usability from a file it does not own --
  1364	# or, in practice, guess from the verdict.
  1365	#
  1366	# THE VERDICT DOES NOT ANSWER IT. Measured across all 79 records on 2026-08-03:
  1367	# 13 were unusable and they carry the whole range of verdicts. `APPROVE` on 11 of
  1368	# them (all merged); `REQUEST CHANGES` on #80 and #83; empty on #89. The
  1369	# REQUEST CHANGES pair is the expensive one: the reviewer's `stated` verdict was
  1370	# read out of the PROMPT'S OWN echoed grading rule, so a record that says an
  1371	# objection was raised is indistinguishable from one where nobody read the code.
  1372	# Both post `state: failure`, and a selector that sees only `failure` sends a
  1373	# never-reviewed PR to REWORK with no findings to work from.
  1374	#
  1375	# ASKED HERE, NOT REUSED FROM $REVIEW_UNUSABLE. All three dispatch paths now set
  1376	# that flag (the `ENGINE != codex` path joined them in the review of PR #319; it
  1377	# previously never evaluated usability at all, which is what let a truncated
  1378	# claude stream green the required gate once claude became primary). It is still
  1379	# asked again here rather than reused, because $REVIEW_UNUSABLE is the flag the
  1380	# VERDICT was computed from and this key is a claim about the FILE -- deriving one
  1381	# from the other would make the record unable to disagree with the gate, and that
  1382	# disagreement is exactly the signal this key exists to preserve.
  1383	#
  1384	# RECORD-ONLY, DELIBERATELY. This changes no gate. $VERDICT is computed above and
  1385	# is not touched here, so no PR's outcome moves on this commit; the consumer that
  1386	# acts on the key is the selector (review-redrive.py), which is a separate change
  1387	# with its own cap. Widening a gate as a side effect of adding a field is how a
  1388	# fleet-wide refusal ships unannounced.
  1389	if review_is_usable "$REVIEW"; then REVIEW_USABLE=1; else REVIEW_USABLE=0; fi
  1390	
  1391	# WHICH MODEL ACTUALLY WROTE THIS REVIEW (sp-8379cd52). `engine` is the FLAG the
  1392	# run was invoked with, not the author. On the DEGRADED path codex never answered
  1393	# and Opus wrote the review, yet the record still said `"engine": "codex"` -- so
  1394	# the human-facing surfaces told the truth (the status description and the Linear
  1395	# comment both say DEGRADED out loud) while the MACHINE-READABLE record that
  1396	# converge.sh:36 and linear-worker.sh:76 gate on claimed a second lab reviewed
  1397	# code that second lab never saw. Measured 2026-08-02 on PR #66 and #67 during a
  1398	# codex out-of-credits outage: both records read `engine: codex`, both reviews
  1399	# were Opus. That is the sp-a72a9567 false-provenance shape aimed at the gating
  1400	# reader instead of the human one, which is the worse direction.
  1401	#
  1402	# DERIVED, NEVER BRANCHED. This reads existing state and adds nothing to the
  1403	# control flow above -- deliberately. The fallback trigger is the one path in this
  1404	# script where a wrong edit posts an unearned green, so the provenance fix is not
  1405	# allowed to touch it. $DEGRADED is set at exactly one place (the outage branch),
  1406	# so deriving from it cannot disagree with what actually ran.
  1407	REVIEWED_BY="$CODEX_MODEL"
  1408	[ "$ENGINE" = "claude" ] && REVIEWED_BY="$CLAUDE_MODEL"
  1409	[ "$DEGRADED" = "1" ] && REVIEWED_BY="$CLAUDE_MODEL"
  1410	# `set -e` IS OFF IN THIS SCRIPT (line 64 is `set -uo pipefail`) and these two
  1411	# lines depend on that. Under `set -e` a false `[ ... ] && assign` is an AND-list
  1412	# whose final status is 1, which exits the shell -- so turning on -e here would
  1413	# abort every healthy codex review right before its record is written. If -e is
  1414	# ever added, these become if/fi first.
  1415	#
  1416	# IT SITS BELOW review_is_usable ON PURPOSE. test-review-degraded-provenance.sh
  1417	# extracts this block by awk range, anchored `^REVIEWED_BY="\$CODEX_MODEL"$` ..
  1418	# `^PY$`, and executes it in a bare subshell to drive the SHIPPED writer instead
  1419	# of a copy. Moving this above the `if review_is_usable` line pulls that function
  1420	# call into the extracted range, where it is undefined -- the writer would die,
  1421	# no record would be written, and the suite would report a break in the test
  1422	# rather than the defect. Keep the derivation adjacent to the python3 call.
  1423	
  1424	# The record path comes from the ONE resolver (repo-slug-lib.sh), passed in as
  1425	# argv 16 rather than rebuilt in python -- a second place that knows the naming
  1426	# rule is a second writer, which is the defect class this repo keeps finding.
  1427	python3 - "$PR" "$ISSUE" "$VERDICT" "$REVIEW" "$(TS)" "$STATED_VERDICT" "$DERIVED_VERDICT" "$ROUND" "$HEAD_SHA" "$VERDICT_DIR" "$ENGINE" "$INVOKER" "$REVIEW_USABLE" "$REVIEWED_BY" "$DEGRADED" "$(verdict_record_write_path "$VERDICT_DIR" "$REVIEW_SLUG" "$PR")" <<'PY'
  1428	import json, sys
  1429	(pr, issue, verdict, review, ts, stated, derived, rnd, head_sha, verdict_dir,
  1430	 engine, invoker, usable, reviewed_by, degraded) = sys.argv[1:16]
  1431	out = sys.argv[16]
  1432	json.dump({"pr": int(pr), "issue": issue, "verdict": verdict,
  1433	           "stated": stated, "derived": derived,
  1434	           "source": "findings" if derived else "prose",
  1435	           "engine": engine,
  1436	           # reviewed_by is the model that produced the prose; engine is the flag
  1437	           # the run was asked for. On the fallback those disagree, and that
  1438	           # disagreement IS the record of the outage.
  1439	           "reviewed_by": reviewed_by,
  1440	           "degraded": degraded == "1",
  1441	           "invoker": invoker,
  1442	           # A real boolean, not "1"/"0". A JSON string "0" is TRUTHY in every
  1443	           # consumer language here, so a truthiness read of the wrong shape
  1444	           # would call every phantom review usable -- the exact inversion this
  1445	           # key exists to prevent.
  1446	           "usable": usable == "1",
  1447	           "round": int(rnd), "review": review, "head_sha": head_sha,
  1448	           "ts": ts}, open(out, "w"), indent=2)
  1449	PY
  1450	
  1451	# Severity floor, capture half: APPROVE WITH NITS is a TERMINAL state -- the
  1452	# loop stops reworking -- so each minor must land in the spillover ledger or it
  1453	# evaporates (no-orphan-findings.md). On REQUEST CHANGES the minors ride along
  1454	# in the review, which is the spec for the next rework pass; capturing them
  1455	# there too would double-file them.
  1456	if [ "$VERDICT" = "APPROVE WITH NITS" ] && [ -n "$ISSUE" ]; then
  1457	  CAPTURED=0
  1458	  MINOR_COUNT=0
  1459	  while IFS='|' read -r _sev claim loc; do
  1460	    [ -n "$claim" ] || continue
  1461	    MINOR_COUNT=$((MINOR_COUNT+1))
  1462	    python3 "$SKEL/plugins/prd-os/scripts/prd_runner.py" spillover add \
  1463	      --source "$ISSUE" --desc "PR #$PR ${MINOR_TAG}review minor: $claim ($loc)" >/dev/null 2>&1 \
  1464	      && CAPTURED=$((CAPTURED+1))
  1465	  done <<EOF
  1466	$(extract_minor_findings "$REVIEW")
  1467	EOF
  1468	  echo "  minors captured as spillover: $CAPTURED of $MINOR_COUNT"
  1469	fi
  1470	
  1471	# The verdict as a COMMIT STATUS on the sha the reviewer read (ASK-217).
  1472	#
  1473	# WHY THIS EXISTS: the verdict record above is a LOCAL file. GitHub cannot see
  1474	# it, so no platform mechanism can gate on it, so every approved PR ends its
  1475	# life waiting on a human. Every prior-art integrator (merge queue, Bors,
  1476	# Mergify, Kodiak) has one shape: every precondition is a required status check
  1477	# and the platform does the merging. pr-receipt-gate.py is already a CI step;
  1478	# this was the one piece still stuck on disk.
  1479	#
  1480	# WHY A STATUS AND NOT A PR REVIEW: a commit status needs no second identity. A
  1481	# PR *review* would deadlock -- this agent runs as the account that authors
  1482	# these PRs, and GitHub forbids self-approval. Proven live on PR #23, 2026-07-27.
  1483	#
  1484	# ABSENT IS NOT APPROVED, and that is the point. A reviewer that fails or times
  1485	# out exits well above this, before the verdict is even computed, so no status is
  1486	# posted at all. Once this context becomes a REQUIRED check, "absent" is what
  1487	# holds the PR -- the safe direction. Nothing on an error path here invents one.
  1488	post_reviewer_status() {
  1489	  local sha="$1" verdict="$2" target="$3"
  1490	  # The context is the ENGINE's slot: kipi/reviewer-approved for claude,
  1491	  # kipi/codex-approved for the independent second opinion. Two contexts, one
  1492	  # writer each, so a gate can require either or both without either engine
  1493	  # being able to answer for the other.
  1494	  local context="$STATUS_CONTEXT" state="failure" desc
  1495	  # ONE reader of the verdict. $VERDICT is the derived-over-stated value already
  1496	  # written to the record; re-grepping the review prose here would be a second
  1497	  # reader with its own semantics, which is the defect class this repo keeps
  1498	  # finding. Anything that is not an approval -- including an unstated verdict --
  1499	  # is a failure, so an unparseable review cannot pass a gate by accident.
  1500	  case "$verdict" in
  1501	    "APPROVE"|"APPROVE WITH NITS") state="success" ;;
  1502	  esac
  1503	  desc="${verdict:-unstated: no verdict parsed from the review}"
  1504	  # SAY IT IN THE SLOT ITSELF. The page fires once on the transition and is gone;
  1505	  # the status description is what a human reads on the PR weeks later. Without
  1506	  # this marker a green kipi/codex-approved is indistinguishable from a real
  1507	  # second opinion, and the whole point of this engine is that it is not Claude.
  1508	  [ "$DEGRADED" = "1" ] && desc="DEGRADED (codex down, Opus fallback): $desc"
  1509	  desc="$(printf '%.140s' "$desc")"
  1510	  # STATUS_REPO_PATH, not {owner}/{repo}: `gh api` resolves those placeholders
  1511	  # from the CWD repo and accepts no -R, so an unattended run posted
  1512	  # kipi/reviewer-approved onto the home repo for a review of another one.
  1513	  local args=(api -X POST "repos/$STATUS_REPO_PATH/statuses/$sha"
  1514	              -f "state=$state" -f "context=$context" -f "description=$desc")
  1515	  # Link only a real URL. The PR comment just above is what --post creates; when
  1516	  # that failed there is nothing to link, and a local file path is not a URL.
  1517	  case "$target" in https://*) args+=(-f "target_url=$target") ;; esac
  1518	  if gh "${args[@]}" >/dev/null 2>&1; then
  1519	    echo "  commit status posted: $context=$state on $sha"
  1520	  else
  1521	    echo "  WARN: could not post commit status '$context' (state=$state) on sha $sha; the review is recorded but NO gate moved" >&2
  1522	  fi
  1523	}
  1524	
  1525	if [ "$POST" = "1" ]; then
  1526	  COMMENT_URL=""
  1527	  # THE STATUS GOES FIRST, BEFORE THE COMMENT (sp-fa810306). This ordering is the
  1528	  # fix, and it is the whole reason this block moved.
  1529	  #
  1530	  # THE SCAR, exact: on 2026-09-07 round 3 of PR #314 posted its APPROVE WITH NITS
  1531	  # comment at 00:29:47Z and the process died before the commit status call. The
  1532	  # approval existed as prose and did not exist as a gate, so the next session read
  1533	  # `gh api statuses adc29516` as empty, re-ran the same head, and the re-run
  1534	  # returned REQUEST CHANGES. An earned approval was overturned by a kill.
  1535	  #
  1536	  # WHICH HALF TO LOSE. A kill can land between any two calls; the only choice is
  1537	  # which one survives it. The comment is 60 KB of prose a human reads later; the
  1538	  # status is the one thing a required check, converge.sh and the next session all
  1539	  # read. Losing prose costs a re-read of the review file on disk. Losing the
  1540	  # status costs a whole re-review that can overturn the verdict. So the status
  1541	  # goes first, unconditionally.
  1542	  #
  1543	  # POSTED TWICE, ON PURPOSE, AND NOT A SECOND WRITER. Commit statuses append and
  1544	  # the latest wins, so the second call below re-posts the SAME context with the
  1545	  # SAME $VERDICT and adds only target_url once the comment URL exists. The value
  1546	  # a gate reads cannot differ between the two calls, because both read $VERDICT,
  1547	  # which was computed once, above. If the run dies between them the first post
  1548	  # stands, missing only its link.
  1549	  if [ -n "$HEAD_SHA" ]; then
  1550	    post_reviewer_status "$HEAD_SHA" "$VERDICT" ""
  1551	  else
  1552	    echo "  no head sha for PR #$PR: posting NO commit status (a status on a guessed sha looks authoritative)"
  1553	  fi
  1554	  # POST THE RENDERED REVIEW, NEVER THE RAW FILE (sp-48688b24). `--body-file
  1555	  # "$REVIEW"` sent the codex agent's entire stdout. Measured on disk 2026-07-30,
  1556	  # four real rounds: 435,280 / 519,377 / 278,439 / 197,279 bytes. Three were
  1557	  # rejected; only the 197,279-byte round landed (as a 197,208-character comment
  1558	  # on PR #46). So the old failure was SIZE-DEPENDENT, not universal -- worth
  1559	  # stating because the first two write-ups of this defect, mine included, both
  1560	  # claimed it failed every time and were wrong.
  1561	  #
  1562	  # THE CAP IS DELIBERATELY CONSERVATIVE, NOT TUNED. The observed ceiling sits
  1563	  # somewhere between 197,279 and 278,439 bytes, while a reproduced rejection
  1564	  # reported `Body is too long (maximum is 65536 characters) (addComment)`. Those
  1565	  # two facts do not agree, so the limit is path-dependent and I do not know which
  1566	  # path a future gh version takes. 60,000 is under BOTH, which makes the comment
  1567	  # succeed regardless of which limit applies. Tuning it upward would trade a
  1568	  # guaranteed delivery for a longer transcript nobody reads.
  1569	  # EXPLICIT XXXXXX TEMPLATE, because `mktemp -t name` is not portable. BSD
  1570	  # mktemp (macOS) appends the random suffix itself; GNU mktemp (the Linux CI
  1571	  # runner) rejects a template with fewer than three X's. My first cut used the
  1572	  # BSD form, passed 14/14 locally, and turned `validate` red on the PR -- the
  1573	  # body file was never created, so --body-file got an empty path. Nothing on
  1574	  # this machine could have caught it; the runner is the other OS.
  1575	  # NEVER FALL BACK TO $REVIEW ITSELF (codex round 4, minor). My first fallback was
  1576	  # `|| REVIEW_BODY="$REVIEW"`, which then ran
  1577	  # `review_comment_body "$REVIEW" > "$REVIEW_BODY"` -- the same path as input and
  1578	  # output. The `>` truncates the review before the renderer reads it, so a mktemp
  1579	  # failure would DESTROY the only copy of a review that cost 8-13 minutes of codex
  1580	  # time, and post a self-copy of the wreckage. A degraded path may post something
  1581	  # worse; it may never eat the artifact.
  1582	  REVIEW_BODY="$(mktemp "${TMPDIR:-/tmp}/pr-review-comment.XXXXXX" 2>/dev/null)" || REVIEW_BODY=""
  1583	  # POST_FILE is what gh sends; REVIEW_BODY is only ever a file WE created. Keeping
  1584	  # them separate is what makes the degraded path safe: with no temp file we post
  1585	  # the raw review unchanged and write nothing, instead of redirecting into the
  1586	  # artifact we are trying to read.
  1587	  # THE MARKER MEANS "A VERDICT WAS PUBLISHED ON THIS HEAD", NOT "A RUN HAPPENED"
  1588	  # (round 1 major, 2026-09-07). The first cut stamped it on every posted comment,
  1589	  # UNUSABLE reviews included. An unusable review posts state=failure and states no
  1590	  # verdict, so that head then carried a red required check AND a marker that made
  1591	  # every automated re-review exit 3 -- a wedge no unattended path could clear,
  1592	  # because --force is a hand-typed flag. The guard exists to stop a re-review from
  1593	  # overturning an EARNED verdict; a review nobody could read earned nothing, so it
  1594	  # must not claim the head.
  1595	  MARK_HEAD=0
  1596	  if [ "$REVIEW_USABLE" = "1" ] && [ -n "$VERDICT" ]; then MARK_HEAD=1; fi
  1597	  [ "$MARK_HEAD" = "1" ] \
  1598	    || echo "  NOTE: this review is unusable or states no verdict, so the comment carries NO head marker and a later run may re-review ${HEAD_SHA:0:8} without --force."
  1599	  if [ -n "$REVIEW_BODY" ]; then
  1600	    # THE MARKER LEADS THE BODY, and the position is load-bearing. It is what the
  1601	    # duplicate-run refusal at the top of this script greps for, and
  1602	    # review_comment_body truncates from the END, so a marker appended after the
  1603	    # body would be the first thing a long review dropped -- the guard would then
  1604	    # be silently disarmed on exactly the reviews that run longest.
  1605	    {
  1606	      [ "$MARK_HEAD" = "1" ] && { review_head_marker "$HEAD_SHA"; printf '\n\n'; }
  1607	      review_comment_body "$REVIEW" "$VERDICT" "$ENGINE" "$DEGRADED"
  1608	      # The structural warning rides INSIDE the comment rather than only in this
  1609	      # log, because the log belongs to whoever ran the reviewer and the PR is
  1610	      # where the next person decides whether to patch again.
  1611	      [ -n "$STRUCTURAL_LINE" ] && printf '\n**%s**\n' "$STRUCTURAL_LINE"
  1612	    } >"$REVIEW_BODY"
  1613	    POST_FILE="$REVIEW_BODY"
  1614	  else
  1615	    POST_FILE="$REVIEW"
  1616	    echo "  WARN: could not create a temp file; posting the RAW review, which GitHub may reject on size. The head marker is NOT in this body, so a later run will not see this verdict and may re-review the same head (sp-fa810306). Erring toward a duplicate round is the safe direction here: the other way is a wedge no unattended run can clear." >&2
  1617	  fi
  1618	  # Keep the reason. A bare "could not comment" sent the maintainer to guess
  1619	  # between a size rejection, an auth failure and a closed PR -- the same
  1620	  # discard-the-reason defect PR #46 fixed one call lower down.
  1621	  COMMENT_ERR="$(mktemp "${TMPDIR:-/tmp}/pr-review-comment-err.XXXXXX" 2>/dev/null)" || COMMENT_ERR=/dev/null
  1622	  if COMMENT_URL="$(gh pr comment "$PR" $KIPI_GH_REPO_ARGS --body-file "$POST_FILE" 2>"$COMMENT_ERR")"; then
  1623	    echo "  posted to PR #$PR ($(wc -c <"$POST_FILE" | tr -d ' ') bytes rendered from $(wc -c <"$REVIEW" | tr -d ' '))"
  1624	  else
  1625	    COMMENT_URL=""
  1626	    echo "  WARN: could not comment on PR #$PR: $(tr '\n' ' ' <"$COMMENT_ERR" | cut -c1-300)" >&2
  1627	    echo "  WARN: the review is on disk at $REVIEW but NO human-readable copy reached the PR" >&2
  1628	  fi
  1629	  rm -f "$REVIEW_BODY" "$COMMENT_ERR"
  1630	  # THE SECOND POST IS AN ENRICHMENT, NOT THE POST. The gate already moved above,
  1631	  # before the comment, so this call exists only to attach target_url to a status
  1632	  # that is already correct. It is skipped when there is no URL to attach: a
  1633	  # re-post carrying nothing new would be two writes for one fact, and the first
  1634	  # one already said everything a gate reads.
  1635	  #
  1636	  # No sha, no status -- the same rule as above. A status on a guessed commit is
  1637	  # worse than none because it looks authoritative (ASK-216).
  1638	  if [ -n "$HEAD_SHA" ] && [ -n "$COMMENT_URL" ]; then
  1639	    post_reviewer_status "$HEAD_SHA" "$VERDICT" "$COMMENT_URL"
  1640	  fi
  1641	  if [ -n "$ISSUE" ]; then
  1642	    # THE REVIEWER'S HALF OF THE CONVERSATION (ASK-221, founder directive
  1643	    # 2026-07-29: Sana and codex talk to each other in the issue's comments).
  1644	    #
  1645	    # This used to post a one-line summary, which is a NOTIFICATION, not a turn in
  1646	    # a conversation: Sana had nothing to answer because the findings themselves
  1647	    # only ever landed on the PR. Carrying the actual findings block onto the issue
  1648	    # is what makes a reply possible, and the issue is the one surface both agents
  1649	    # can see (the worker reads PR comments; the founder reads Linear).
  1650	    #
  1651	    # Attributed to "$ENGINE-reviewer", never a bare "reviewer": the whole point of
  1652	    # the flip is that the checker is not Claude, so a thread that cannot tell you
  1653	    # WHICH engine spoke loses the only fact that matters. It also gives Sana a
  1654	    # string to filter on (`linear-sync.py comments --agent codex-reviewer`).
  1655	    # Through the ONE reader (sp-c0a9dac3). Its own sed range here was the third
  1656	    # copy of that extraction, so the Linear comment could carry a DIFFERENT set of
  1657	    # findings than the verdict was derived from -- Sana would be answering findings
  1658	    # that never set the gate, on a review whose gate came from findings she never saw.
  1659	    REVIEW_FINDINGS="$(findings_block "$REVIEW")"
  1660	    [ -n "$REVIEW_FINDINGS" ] || REVIEW_FINDINGS="(no findings block parsed from this review)"
  1661	    # `|| true` HERE WAS A SILENT DROP (codex round 2 of PR #34, minor;
  1662	    # sp-583dc1a0). Every other failure on this path says so out loud -- the PR
  1663	    # comment warns, and a failed commit status warns that NO gate moved -- but a
  1664	    # failed Linear post printed nothing, discarded the reason down /dev/null, and
  1665	    # the run still exited 0 and printed `done`. Linear is the ONE surface Sana
  1666	    # reads, so losing it silently means she never answers findings she was never
  1667	    # shown, and the loop looks healthy while the conversation never happens. Not
  1668	    # hypothetical: the round-2 run on 2026-07-30 lost its PR comment
  1669	    # (`WARN: could not comment on PR`) and only the loud branch revealed it.
  1670	    #
  1671	    # STILL EXIT 0, deliberately. The gate above is already correctly set from a
  1672	    # review that really ran; making the run fail here would make the worker log
  1673	    # `codex reviewer failed` for a review that succeeded, which trades a silent
  1674	    # drop for a false alarm. Loud plus a page is the fix, not a non-zero exit.
  1675	    SYNC_ERR="$(python3 "$SYNC" progress "$ISSUE" \
  1676	      "Review of PR #$PR complete ($ENGINE engine$([ "$DEGRADED" = "1" ] && printf ', DEGRADED: codex down, Opus fallback')). Verdict: ${VERDICT:-unstated}. Reviewer: Meta senior-staff persona, fresh eyes, every finding required to ship an executed reproducer.
  1677	
  1678	Sana: reply to this comment on THIS issue. For each finding, either the file:line that already handles it, or what you changed. Findings below." \
  1679	      --agent "$ENGINE-reviewer" --evidence "$REVIEW_FINDINGS" 2>&1 >/dev/null)"
  1680	    if [ $? -eq 0 ]; then
  1681	      echo "  review posted to $ISSUE as $ENGINE-reviewer (findings included)"
  1682	    else
  1683	      echo "  WARN: could not post the review to $ISSUE as $ENGINE-reviewer. The gate is set from a review nobody on the issue can see, so Sana has no findings to answer. Reason: ${SYNC_ERR:-(no output)}" >&2
  1684	      # THE PAGE IS BEST-EFFORT AND SAYS SO (codex round 1 of PR #46, major 2).
  1685	      # slack-notify.sh no-ops silently when no webhook is configured -- that is
  1686	      # deliberate per founder-notifications.md, so callers never break -- which
  1687	      # means a zero exit here does NOT prove delivery. Claiming "paged" would be
  1688	      # the same overclaim this commit is removing one layer down. So: attempt it,
  1689	      # record what came back, and leave the stderr WARN above as the one record
  1690	      # that is always written.
  1691	      NOTIFY_OUT="$(bash "$NOTIFY" "reviewer: PR #$PR review did NOT reach $ISSUE ($ENGINE engine, verdict ${VERDICT:-unstated}). The gate moved but the findings are not on the issue, so the rework conversation cannot start." 2>&1)"
  1692	      NOTIFY_RC=$?
  1693	      if [ "$NOTIFY_RC" -ne 0 ]; then
  1694	        echo "  WARN: the page about that loss ALSO failed (rc=$NOTIFY_RC${NOTIFY_OUT:+: $NOTIFY_OUT}). This loss is recorded ONLY in this log." >&2
  1695	      else
  1696	        echo "  page attempted for the lost review (delivery not confirmable: the notifier no-ops silently when unconfigured)" >&2
  1697	      fi
  1698	    fi
  1699	  fi
  1700	fi
  1701	
  1702	echo "$(TS) done$DRY_NOTE"
  1703	# EXIT 4 MEANS "IT ALL WORKED AND THE NEXT PATCH IS THE WRONG MOVE" (sp-46726f79).
  1704	# It is reached only AFTER the post, so a caller that treats non-zero as failure
  1705	# still has its comment and its commit status. Callers that care read the code;
  1706	# callers that do not are no worse off than before, because the outward side
  1707	# effects are identical either way.
  1708	exit "${FINAL_EXIT:-0}"
  2460	    # is stated EXPLICITLY here rather than inherited from the reviewer's default,
  2461	    # because which model checks this fleet's work is the kind of fact that must be
  2462	    # readable at the call site, not two files away.
  2463	    #
  2464	    # ONE call, not two. It was claude-then-codex, then codex-only, and since
  2465	    # 2026-09-06 claude-only: claude owns kipi/reviewer-approved and writes the one
  2466	    # verdict record every gate below reads, so a second codex pass would only burn
  2467	    # spend and post an advisory status nobody gates on.
  2468	    #
  2469	    # THERE IS NO FALLBACK IN THIS DIRECTION, and the sentence here used to claim
  2470	    # one. The Opus fallback and the DEGRADED marking hang off the reviewer's codex
  2471	    # branch, so with claude PRIMARY nothing stands behind a claude outage. That is
  2472	    # the SAFE direction, not a gap: the reviewer exits non-zero and posts NO status,
  2473	    # reviewer-floor turns an absent verdict into a red required context, and the PR
  2474	    # holds. A codex fallback would be worse than none -- codex is out of credits and
  2475	    # fails at EXIT 0, so it would fill the required gate with nothing.
  2476	    # LABEL THE INVOKER HERE, at the one place the scheduled path runs the reviewer
  2477	    # (sp-53aad86f). This is what makes a dispatcher-driven review distinguishable
  2478	    # from a hand run in the verdict record. It is set on the call rather than
  2479	    # exported once, so it cannot leak into an unrelated reviewer invocation.
  2480	    KIPI_REVIEW_INVOKER=worker $REVIEWER_CMD "$PR_NUM" --issue "$ISSUE" --post --engine claude >>"$LOG" 2>&1
  2481	    REVIEWER_RC=$?
  2482	    # EXIT 4 IS A COMPLETED REVIEW, NOT A FAILURE (sp-46726f79, 2026-09-07). The
  2483	    # reviewer returns it AFTER posting its comment and its commit status, to say
  2484	    # that this round's findings repeat a file from the previous REQUEST CHANGES
  2485	    # round -- the wrong-fix-shape signal. Folding it into the `|| say WARN` line
  2486	    # would log "the PR stands, unreviewed" about a review that fully landed and
  2487	    # set the gate, which is a false statement in the one log an operator scans.
  2488	    # Anything else non-zero keeps the original meaning.
  2489	    if [ "$REVIEWER_RC" = "4" ]; then
  2490	      say "NOTE: the reviewer posted its verdict on PR #$PR_NUM and flagged it STRUCTURAL -- this round repeats a file from the previous REQUEST CHANGES round. Sweep the whole class in one pass or park the PR; a third patch on the same file is the loop that warning exists to stop."
  2491	    elif [ "$REVIEWER_RC" != "0" ]; then
  2492	      say "WARN: the claude reviewer failed on PR #$PR_NUM (rc=$REVIEWER_RC; the PR stands, unreviewed)"
  2493	    fi
  2494	    # Read back the verdict RECORD the reviewer just wrote (never re-grep the
  2495	    # review prose) and state what happens next in plain terms. Rework itself
  2496	    # fires on the NEXT run, through the severity-floor gate above.
  2497	    FINAL_VERDICT="$(verdict_from_record "$(verdict_record_path "$REVIEWS_DIR" "$TARGET_SLUG" "$PR_NUM")")"
  2498	    # RE-GATE THE RECORD BEFORE REPORTING ON IT (PR #30 review round 2, major 3).
  2499	    # The reviewer above can fail -- it is a `|| say WARN` line, not a hard stop --
  2500	    # and when it does, this read returns the SAME record the gate at the top of
  2501	    # this run already refused to trust. Reporting "converged ... waits on founder
  2502	    # merge" off it put that line two lines below "the code at the head was never
  2503	    # reviewed", and the closing line is the one an operator scans. Unreachable
  2504	    # before ASK-219 (gate 10 skipped the issue before step 5 could run), which is
  2505	    # why arming exit 40 is what exposed it.
  2506	    #
  2507	    # Both shas are re-read rather than reused: a round that pushed moved the head,
  2508	    # so the values from the top of the loop describe a state that no longer
  2509	    # exists. The gate is the ONE reader of the comparison -- deriving it here
  2510	    # would be a second reader with drifting semantics.
  2511	    FINAL_REVIEWED_SHA="$(head_sha_from_record "$(verdict_record_path "$REVIEWS_DIR" "$TARGET_SLUG" "$PR_NUM")")"
  2512	    FINAL_CURRENT_SHA="$(pr_head_sha "$PR_NUM")"
  2513	    # AND THE GATE'S NOTE IS SAID, NOT SWALLOWED (PR #30 review round 3, minor 2).
  2514	    # converge.sh's own call site states the rule this line broke: "Swallowing it
  2515	    # would silently grandfather the blind spot it announces." The reviewer always

exec
/bin/zsh -lc "nl -ba q-system/.q-system/scripts/pr-review-agent.sh | sed -n '495,555p'; rg -n \"comments.*body|review_head_marker|KIPI_REVIEW_CONCURRENCY|REVIEW_PID\" q-system/.q-system/scripts/test/test-review-run-safety.sh" in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-322
 succeeded in 0ms:
   495	  echo "REFUSING: PR #$PR's head moved between two reads (${HEAD_SHA:0:8} then ${HEAD_SHA_CONFIRM:0:8})." >&2
   496	  echo "  Something is pushing to this branch right now. Reviewing either sha risks a green status on code the reviewer did not read." >&2
   497	  echo "  Re-run once the branch settles. No review was dispatched and NO status was posted." >&2
   498	  exit 1
   499	fi
   500	# A HEAD THAT ALREADY CARRIES A VERDICT IS NOT RE-REVIEWED (sp-fa810306).
   501	#
   502	# THE SCAR, exact: on 2026-09-07 round 3 of PR #314 posted APPROVE WITH NITS as a
   503	# comment at 00:29:47Z on head adc29516, the process died before the commit status
   504	# landed, and the next session read `gh api statuses adc29516` as empty. Empty
   505	# looked like "never reviewed", so the same head was re-run at 00:36:54Z -- and
   506	# that re-run returned REQUEST CHANGES, overturning an approval that had already
   507	# been earned. Two more rounds and about 90 minutes.
   508	#
   509	# THE COMMENT IS THE DURABLE RECORD, NOT THE STATUS, and that is the whole point
   510	# of checking it here. The status post is the thing that was lost; the comment is
   511	# the thing that survived. So this asks the question the re-running session should
   512	# have asked: has a reviewer already published a verdict on THIS EXACT COMMIT?
   513	# Posting the status BEFORE the comment (further down) closes the other half.
   514	#
   515	# FAILS OPEN. A gh outage, a network error, or a PR with no comments all yield
   516	# zero hits and the review proceeds. Refusing on an unreadable comment list would
   517	# wedge every review on a GitHub blip, which is worse than one duplicate round --
   518	# and the duplicate round is what the caller asked for in the first place.
   519	if [ -n "$HEAD_SHA" ]; then
   520	  DUP_MARKER="$(review_head_marker "$HEAD_SHA")"
   521	  DUP_HITS="$(gh pr view "$PR" $KIPI_GH_REPO_ARGS --json comments -q '.comments[].body' 2>/dev/null | grep -F -c -- "$DUP_MARKER" || true)"
   522	  DUP_HITS="$(printf '%s' "${DUP_HITS:-0}" | tr -dc '0-9')"
   523	  [ -n "$DUP_HITS" ] || DUP_HITS=0
   524	  if [ "$DUP_HITS" -gt 0 ] && [ "$FORCE" != "1" ]; then
   525	    echo "REFUSING: PR #$PR head ${HEAD_SHA:0:8} already carries $DUP_HITS reviewer verdict comment(s). Re-reviewing a head that has already been judged can only overturn a verdict that was already earned (sp-fa810306). No review was dispatched and NO status was posted." >&2
   526	    echo "  Read the existing verdict on the PR, or pass --force if a second opinion on the same commit is what you actually want." >&2
   527	    exit 3
   528	  fi
   529	  [ "$DUP_HITS" -gt 0 ] && echo "  NOTE: ${HEAD_SHA:0:8} already carries $DUP_HITS verdict comment(s); --force given, reviewing it again."
   530	fi
   531	
   532	[ -n "$ISSUE" ] || ISSUE="$(printf '%s' "$PR_TITLE" | grep -oE 'ASK-[0-9]+' | head -1)"
   533	
   534	echo "$(TS) reviewing PR #$PR: $PR_TITLE"
   535	echo "  head sha under review: ${HEAD_SHA:-unknown}"
   536	[ -n "$ISSUE" ] && echo "  linked issue: $ISSUE"
   537	
   538	# THE TREE MUST ACTUALLY CONTAIN THE PR (sp-a72a9567). $SKEL comes from this
   539	# script's own location, and the diff comes from `gh pr diff <N>` -- two
   540	# independent sources that nothing was checking against each other. Run from
   541	# worktree A against a PR on branch B and the reviewer reads A's files off disk
   542	# while B's diff scrolls past, then writes a verdict record and a commit status
   543	# attributing its findings to B's head sha.
   544	#
   545	# Not hypothetical. 2026-07-29, run from the ask-221 worktree against PR #35: it
   546	# returned three findings in linear-sync.py, a file PR #35's diff does not touch at
   547	# all. The findings were real bugs in ask-221; the PROVENANCE was false. That is
   548	# worse than a wrong verdict, because the record looks authoritative.
   549	#
   550	# TWO TIERS, because a flat equality check would be wrong twice over. The PR's head
   551	# may legitimately be BEHIND local HEAD (a push landing after the `gh pr view`
   552	# above), so equality would refuse healthy runs -- ancestry is the real question.
   553	# And an UNKNOWN object is not evidence of a mismatch: a stale or partial clone
   554	# cannot prove ancestry either way, and inventing a refusal there would wedge the
   555	# loop on a fetch problem. Unknown warns; known-but-unrelated refuses.
43:unset KIPI_TARGET_REPO KIPI_REVIEW_ENGINE KIPI_REVIEW_CONCURRENCY_CAP KIPI_REVIEW_QUEUE_SECONDS 2>/dev/null || true

exec
/bin/zsh -lc 'HEAD_SHA=0123456789abcdef; DUP_MARKER=""; ATTACKER_BODY="I am not the reviewer. $DUP_MARKER"; DUP_HITS="$(printf '"'%s\\n' \""'$ATTACKER_BODY" | grep -F -c -- "$DUP_MARKER" || true)"; printf '"'comment_author=untrusted-user\\nprojected_body=%s\\nDUP_HITS=%s\\n' \""'$ATTACKER_BODY" "$DUP_HITS"; if [ "$DUP_HITS" -gt 0 ]; then printf '"'RESULT=REFUSING before dispatch\\n'; fi" in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-322
 succeeded in 0ms:
comment_author=untrusted-user
projected_body=I am not the reviewer. 
DUP_HITS=1
RESULT=REFUSING before dispatch

codex
A second path survived inspection: when the first status POST is rejected, the script still publishes a usable comment with the duplicate marker. The next unattended run then refuses the same head, so the required status can remain absent until a human forces a review. I’m reproducing that state transition and checking the shipped tests for a retry control.
exec
/bin/zsh -lc "POST_STATUS_RC=1; REVIEW_USABLE=1; VERDICT='APPROVE'; HEAD_SHA=0123456789abcdef; if [ \""'$POST_STATUS_RC" -ne 0 ]; then echo "WARN: status post rejected"; fi; MARK_HEAD=0; if [ "$REVIEW_USABLE" = 1 ] && [ -n "$VERDICT" ]; then MARK_HEAD=1; fi; COMMENT_BODY='"''; if [ \""'$MARK_HEAD" = 1 ]; then COMMENT_BODY=""; fi; DUP_HITS="$(printf '"'%s\\n' \""'$COMMENT_BODY" | grep -F -c -- "" || true)"; printf '"'first_status_posts=0\\ncomment_marker=%s\\nnext_run_DUP_HITS=%s\\n' \""'$COMMENT_BODY" "$DUP_HITS"; if [ "$DUP_HITS" -gt 0 ]; then echo '"'next_run=REFUSED rc=3'; fi; rg -n \"fail-status|head marker|re-review\" q-system/.q-system/scripts/test/test-review-run-safety.sh q-system/.q-system/scripts/test/test-severity-floor.sh | tail -30" in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-322
 succeeded in 0ms:
WARN: status post rejected
first_status_posts=0
comment_marker=
next_run_DUP_HITS=1
next_run=REFUSED rc=3
q-system/.q-system/scripts/test/test-severity-floor.sh:602:  || fail "the rebased diff was not re-reviewed; a force-push nobody looked at ships under the OLD verdict"
q-system/.q-system/scripts/test/test-severity-floor.sh:603:ok "a rebase round's resulting diff is re-reviewed (round recorded)"
q-system/.q-system/scripts/test/test-severity-floor.sh:740:# re-review at the new head; never merge, never auto-approve).
q-system/.q-system/scripts/test/test-severity-floor.sh:763:# re-review first, then the fresh record decides whether it is a rebase round.
q-system/.q-system/scripts/test/test-severity-floor.sh:765:  "drift outranks DIRTY: re-review at the new head before any rebase round"
q-system/.q-system/scripts/test/test-severity-floor.sh:769:# re-reviews forever, burning model budget and writing a permanent Linear
q-system/.q-system/scripts/test/test-severity-floor.sh:772:  "a matching sha stays terminal (a converged PR does not re-review forever)"
q-system/.q-system/scripts/test/test-severity-floor.sh:789:# drift would re-review every converged PR on the board at once. So absent falls
q-system/.q-system/scripts/test/test-severity-floor.sh:795:      Reading absent-as-drift re-reviews every pre-ASK-216 PR on the board at once."
q-system/.q-system/scripts/test/test-severity-floor.sh:796:ok "absent head_sha falls back to today's behaviour (no mass re-review)"
q-system/.q-system/scripts/test/test-severity-floor.sh:805:# toward terminal, because a manufactured re-review round costs every PR in the
q-system/.q-system/scripts/test/test-severity-floor.sh:809:  || fail "an unreadable current head manufactured a re-review round (got $GOT, want 10)"
q-system/.q-system/scripts/test/test-severity-floor.sh:812:ok "an unreadable current head does not manufacture a re-review round, and says so"
q-system/.q-system/scripts/test/test-severity-floor.sh:838:ok "record -> gate chain: approved record + moved head -> re-review, not merge"
q-system/.q-system/scripts/test/test-severity-floor.sh:932:# $4 "fail-status"  => the status POST exits non-zero
q-system/.q-system/scripts/test/test-severity-floor.sh:967:  "api -X")     [ "$mode" = "fail-status" ] && exit 1
q-system/.q-system/scripts/test/test-severity-floor.sh:1092:mk_status_stubs "$N4" "$APPROVE_REVIEW" "$SHA_A" fail-status
q-system/.q-system/scripts/test/test-severity-floor.sh:1096:      loop's hand-off; losing it to a transient GitHub error costs a full re-review."
q-system/.q-system/scripts/test/test-severity-floor.sh:1329:# every approved PR on the board re-reviews forever, burning model budget and
q-system/.q-system/scripts/test/test-severity-floor.sh:1339:      This fix must not turn every approved PR into an endless re-review. It said:
q-system/.q-system/scripts/test/test-severity-floor.sh:1365:# would re-review the entire board at once.
q-system/.q-system/scripts/test/test-severity-floor.sh:1374:      Absent is not drift; reading it that way re-reviews every PR on the board at once."
q-system/.q-system/scripts/test/test-severity-floor.sh:1382:# Same posture as ASK-212's empty merge state. A manufactured re-review round
q-system/.q-system/scripts/test/test-severity-floor.sh:1461:# unreviewed-code path wearing a rebase coat, so the re-review has to win and the
q-system/.q-system/scripts/test/test-severity-floor.sh:1482:      rebase budget on re-reviews makes a real conflict un-dispatchable later."
q-system/.q-system/scripts/test/test-severity-floor.sh:1568:grep -qi "re-review round" "$W2/prompt.txt" \
q-system/.q-system/scripts/test/test-severity-floor.sh:1573:ok "the drift round's prompt states that this is a re-review round"
q-system/.q-system/scripts/test/test-severity-floor.sh:2897:  && fail "arming at the gate re-reviewed an already-approved PR. Every scheduled run would pay
q-system/.q-system/scripts/test/test-severity-floor.sh:4020:      The operator needs a re-review, not a git command: $(cat "$W2/pages.txt")"
q-system/.q-system/scripts/test/test-severity-floor.sh:4054:      is that a manufactured re-review round costs the whole fleet at once. It said:

exec
/bin/zsh -lc "bash -n q-system/.q-system/scripts/pr-review-agent.sh q-system/.q-system/scripts/linear-worker.sh q-system/.q-system/scripts/test/test-review-run-safety.sh q-system/.q-system/scripts/test/test-severity-floor.sh; printf 'bash_n_rc=%s\\n' \""'$?"; git diff --check origin/main...HEAD; printf '"'diff_check_rc=%s\\n' \""'$?"; HEAD_SHA=0123456789abcdef; MARKER=""; ATTACKER_BODY="I am not the reviewer. $MARKER"; HITS="$(printf '"'%s\\n' \""'$ATTACKER_BODY" | grep -F -c "$MARKER" || true)"; printf '"'author=untrusted-user\\nhits=%s\\ndecision=%s\\n' \""'$HITS" "$([ "$HITS" -gt 0 ] && echo '"'REFUSE rc=3')\"; STATUS_POSTED=0; REVIEW_USABLE=1; VERDICT=APPROVE; MARK_HEAD=0; if [ \""'$REVIEW_USABLE" = 1 ] && [ -n "$VERDICT" ]; then MARK_HEAD=1; fi; BODY="$([ "$MARK_HEAD" = 1 ] && printf '"'%s' \""'$MARKER")"; HITS="$(printf '"'%s\\n' \""'$BODY" | grep -F -c "$MARKER" || true)"; printf '"'status_posts=%s\\nposted_body=%s\\nnext_hits=%s\\nnext_decision=%s\\n' \""'$STATUS_POSTED" "$BODY" "$HITS" "$([ "$HITS" -gt 0 ] && echo '"'REFUSE rc=3')\"" in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-322
 succeeded in 0ms:
bash_n_rc=0
diff_check_rc=0
author=untrusted-user
hits=1
decision=REFUSE rc=3
status_posts=0
posted_body=
next_hits=1
next_decision=REFUSE rc=3

codex
## Findings

### MAJOR

A rejected commit-status POST still produces a duplicate-review marker, causing every later unattended run to refuse the same head while its required status remains absent.

- Confidence: Confirmed
- Location: `q-system/.q-system/scripts/pr-review-agent.sh:1595`
- Human action at 3am: rerun with `--force` after GitHub recovers, or manually restore the missing status.
- Root cause: `MARK_HEAD` depends only on review usability and verdict presence. It does not depend on whether the status POST at line 1550 succeeded.
- The repository’s own `fail-status` fixture at `test-severity-floor.sh:1092` proves this input is supported, but it never attempts the next review.

Reproducer command actually run:

```bash
HEAD_SHA=0123456789abcdef
MARKER="<!-- kipi-reviewer: head=$HEAD_SHA -->"
STATUS_POSTED=0
REVIEW_USABLE=1
VERDICT=APPROVE
MARK_HEAD=0
if [ "$REVIEW_USABLE" = 1 ] && [ -n "$VERDICT" ]; then MARK_HEAD=1; fi
BODY="$([ "$MARK_HEAD" = 1 ] && printf '%s' "$MARKER")"
HITS="$(printf '%s\n' "$BODY" | grep -F -c "$MARKER" || true)"
printf 'status_posts=%s\nposted_body=%s\nnext_hits=%s\nnext_decision=%s\n' \
  "$STATUS_POSTED" "$BODY" "$HITS" \
  "$([ "$HITS" -gt 0 ] && echo 'REFUSE rc=3')"
```

Real output:

```text
status_posts=0
posted_body=<!-- kipi-reviewer: head=0123456789abcdef -->
next_hits=1
next_decision=REFUSE rc=3
```

### MAJOR

Any public commenter can forge the marker and suppress unattended review because the duplicate check discards comment authorship and searches every comment body.

- Confidence: Confirmed
- Location: `q-system/.q-system/scripts/pr-review-agent.sh:521`
- Human action at 3am: identify the forged comment and rerun with `--force`.
- Blast radius: the PR remains unreviewed with no status posted. This is recoverable, but requires manual intervention.

Reproducer command actually run:

```bash
HEAD_SHA=0123456789abcdef
MARKER="<!-- kipi-reviewer: head=$HEAD_SHA -->"
ATTACKER_BODY="I am not the reviewer. $MARKER"
HITS="$(printf '%s\n' "$ATTACKER_BODY" | grep -F -c "$MARKER" || true)"
printf 'author=untrusted-user\nhits=%s\ndecision=%s\n' \
  "$HITS" "$([ "$HITS" -gt 0 ] && echo 'REFUSE rc=3')"
```

Real output:

```text
author=untrusted-user
hits=1
decision=REFUSE rc=3
```

## What is sound

- `bash -n` passed for both changed scripts and both test scripts.
- `git diff --check origin/main...HEAD` passed.
- Unusable reviews do not receive the duplicate marker.
- Exit code 4 is handled distinctly by `linear-worker.sh`.
- The structural comparison filters marker rows and distinguishes disjoint files.
- The full test suite could not execute because this review environment denied `mktemp` writes. The observed error was `Operation not permitted`, not a test assertion from the PR.

## VERDICT

**REQUEST CHANGES**

Fix the status-loss wedge first. Only publish a duplicate-review marker after the required commit status was confirmed posted. Then authenticate marker ownership before treating a comment as reviewer evidence.

FINDINGS:
major|A rejected commit-status POST still publishes a duplicate-review marker, permanently refusing unattended retries until a human forces the review|q-system/.q-system/scripts/pr-review-agent.sh:1595
major|Any public commenter can forge the unauthenticated head marker and suppress unattended review of that commit|q-system/.q-system/scripts/pr-review-agent.sh:521
END FINDINGS
tokens used
85,806
## Findings

### MAJOR

A rejected commit-status POST still produces a duplicate-review marker, causing every later unattended run to refuse the same head while its required status remains absent.

- Confidence: Confirmed
- Location: `q-system/.q-system/scripts/pr-review-agent.sh:1595`
- Human action at 3am: rerun with `--force` after GitHub recovers, or manually restore the missing status.
- Root cause: `MARK_HEAD` depends only on review usability and verdict presence. It does not depend on whether the status POST at line 1550 succeeded.
- The repository’s own `fail-status` fixture at `test-severity-floor.sh:1092` proves this input is supported, but it never attempts the next review.

Reproducer command actually run:

```bash
HEAD_SHA=0123456789abcdef
MARKER="<!-- kipi-reviewer: head=$HEAD_SHA -->"
STATUS_POSTED=0
REVIEW_USABLE=1
VERDICT=APPROVE
MARK_HEAD=0
if [ "$REVIEW_USABLE" = 1 ] && [ -n "$VERDICT" ]; then MARK_HEAD=1; fi
BODY="$([ "$MARK_HEAD" = 1 ] && printf '%s' "$MARKER")"
HITS="$(printf '%s\n' "$BODY" | grep -F -c "$MARKER" || true)"
printf 'status_posts=%s\nposted_body=%s\nnext_hits=%s\nnext_decision=%s\n' \
  "$STATUS_POSTED" "$BODY" "$HITS" \
  "$([ "$HITS" -gt 0 ] && echo 'REFUSE rc=3')"
```

Real output:

```text
status_posts=0
posted_body=<!-- kipi-reviewer: head=0123456789abcdef -->
next_hits=1
next_decision=REFUSE rc=3
```

### MAJOR

Any public commenter can forge the marker and suppress unattended review because the duplicate check discards comment authorship and searches every comment body.

- Confidence: Confirmed
- Location: `q-system/.q-system/scripts/pr-review-agent.sh:521`
- Human action at 3am: identify the forged comment and rerun with `--force`.
- Blast radius: the PR remains unreviewed with no status posted. This is recoverable, but requires manual intervention.

Reproducer command actually run:

```bash
HEAD_SHA=0123456789abcdef
MARKER="<!-- kipi-reviewer: head=$HEAD_SHA -->"
ATTACKER_BODY="I am not the reviewer. $MARKER"
HITS="$(printf '%s\n' "$ATTACKER_BODY" | grep -F -c "$MARKER" || true)"
printf 'author=untrusted-user\nhits=%s\ndecision=%s\n' \
  "$HITS" "$([ "$HITS" -gt 0 ] && echo 'REFUSE rc=3')"
```

Real output:

```text
author=untrusted-user
hits=1
decision=REFUSE rc=3
```

## What is sound

- `bash -n` passed for both changed scripts and both test scripts.
- `git diff --check origin/main...HEAD` passed.
- Unusable reviews do not receive the duplicate marker.
- Exit code 4 is handled distinctly by `linear-worker.sh`.
- The structural comparison filters marker rows and distinguishes disjoint files.
- The full test suite could not execute because this review environment denied `mktemp` writes. The observed error was `Operation not permitted`, not a test assertion from the PR.

## VERDICT

**REQUEST CHANGES**

Fix the status-loss wedge first. Only publish a duplicate-review marker after the required commit status was confirmed posted. Then authenticate marker ownership before treating a comment as reviewer evidence.

FINDINGS:
major|A rejected commit-status POST still publishes a duplicate-review marker, permanently refusing unattended retries until a human forces the review|q-system/.q-system/scripts/pr-review-agent.sh:1595
major|Any public commenter can forge the unauthenticated head marker and suppress unattended review of that commit|q-system/.q-system/scripts/pr-review-agent.sh:521
END FINDINGS

PR #323, opened 2026-09-07
fix(destructive-op-deny): the fleet updater as a script file with any path prefix, a subshell, or a bypass-arming wrapper (sp-37c08fb1, sp-dc76e644)

assafkip:
Closes spillover **sp-37c08fb1** and **sp-dc76e644**: a destructive act wearing a different filename walks past the deny gate.

## What this PR touches

Four files, nothing else: the tested hook fixture, the new test suite, its capability-manifest entry, and the install proposal. The hook this repo ships is the fixture; the live copy on the founder's machine is written only through the sanctioned proposal path.

## The two ledger items

**sp-37c08fb1.** The fleet updater invoked as a script file. Measured on 2026-09-07 by driving the live hook: a relative path with no leading dot-slash, the bare basename on PATH, and an env-prefixed form were all allowed. The exact command that ran a real sync unchallenged at 06:08Z that day, a parenthesised subshell after an `&&` split, was allowed too, by the live hook and by the first cut of this PR.

**sp-dc76e644.** A wrapper script that assigns the founder bypass in its own text presents a command line carrying no destructive pattern. On 2026-09-06 that ran a real fleet sync on consulting and reverted a deliberate commit four minutes after it landed.

## The fix, both layers appended, deny-only

**Layer 1**, the script form at command position: any path prefix or none, a subshell or group opening the position, the transparent prefixes (sudo, command, nohup, nice, time, env, exec, setsid, eval, timeout with its duration), env assignments, and a quoted path. Reading the file is not running it: `cat`, `sed -n`, `wc -l` on the updater stay allowed. The dry-run flag alone in its own tool call still passes through the same preview predicate, inside a subshell too.

**Layer 2**, the wrapper scan: the target of a `bash <path>`, `sh <path>`, `./<path>` or `<path>` stage is read, bounded at 64 KB, comment lines dropped, and each remaining line is split into stages the way the command line is. A stage whose command is the updater, or whose command is an assignment of the bypass, refuses the invocation. A backup script's rsync is not the updater and is not matched. `bash -n FILE` is a syntax check and passes.

**Placement.** Both layers sit above the pre-existing `_fleet_preview` early exit. Round 1 measured the first cut below it: one dry-run stage anywhere in the command block set the preview flag and exited before either layer ran, so `X --dry-run && ( X --only Y )` was allowed, the exact hazard ASK-1118 names.

## Measured, red first

Every negative case was watched red against the fixture before the change that closes it, then green after:

- first cut, script forms: 8 failed of 40 against the pre-fix copy, then 40 passed
- subshell shape (round 2 of this branch): 4 failed, then 46 passed
- wrapper inside a subshell and the glued closing paren: 1 failed, then 48 passed
- round-2 fixes (preview in the same block, prefixes, quotes, the narrowed wrapper read): see the final commit message for the exact red and green counts

Population of the wrapper rule, driven with `bash <path>` on every tracked `.sh` in this repo and on every `.sh` under the founder's projects folder that carries a matchable token at all (counts only for the outside population): the first cut refused 20 of 212 here and 1933 of 5276 outside. The final commit message carries the re-measured numbers after the command-position read.

`fixture == live hook + proposal insert` holds byte for byte after every commit on this branch, so applying the proposal to the live hook reproduces the fixture.

## How the live hook gets installed, and by whom

No agent can write the live hook. The founder, one command, from a checkout that carries this branch:

```
bash q-system/.q-system/scripts/apply-claude-changes.sh q-system/output/claude-changes/sp-37c08fb1-fleet-updater-script-form.json
```

Until that runs, the fixture is the tested artifact and the live hook still allows every shape above. The anchor suite's drift test is red on the founder's machine for exactly that reason and skips on any machine without a live hook; a signal that runs where a machine can act on it is captured on the ledger, not built here.

## What is not verified here

- The live hook after installation. The identity above is the proof that the install reproduces the fixture; the installed hook has not been driven.
- Shapes outside the test set. A text-matching guard sees the spelling in front of it; the repo's own lesson on computed paths applies.


assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-323-20260907-142129.md` (12003 bytes).

```
FINDINGS:
major|A preview stage hits the _fleet_preview early exit at line 439, so both new layers never run and `./kipi-update.sh --dry-run && ( ./kipi-update.sh --only X )` is ALLOWED|q-system/.q-system/tests/fixtures/destructive-op-deny.reference.sh:439
major|Layer 2 refuses 19 of the repo's 212 tracked scripts, not the 6 the comment and the install proposal claim, including scripts that only print the words "kipi update" and the capability-token installer the design depends on|q-system/.q-system/tests/fixtures/destructive-op-deny.reference.sh:539
minor|The new command-position skip misses timeout/setsid/exec/eval prefixes and a quoted script path, all measured ALLOW with the real updater present|q-system/.q-system/tests/fixtures/destructive-op-deny.reference.sh:500
minor|The PR description documents pr-review-agent.sh changes that are not in this diff, including the "What is NOT verified here" list a founder reads before applying to ~/.claude|q-system/output/claude-changes/sp-37c08fb1-fleet-updater-script-form.json:1
minor|The only claimed install signal, the anchor drift test, fails locally and skips on any machine without a live hook, so no automated run can observe that the fix never installed|test_destructive_op_deny_anchor.py:477
END FINDINGS
```

--- reviewer output ---

## Review of PR #323

Read-only. Nothing in the repo was modified. All repros ran against a `cp` of the shipped fixture `q-system/.q-system/tests/fixtures/destructive-op-deny.reference.sh` into `$TMPDIR`, driven with the real payload shape (`tool_name`/`tool_input.command`/`cwd`), `HOME` under `mktemp`, bypass var unset.

---

### FINDING 1 (major): a preview stage anywhere in the command hits the old `exit 0` at line 439, and both new layers never run

`q-system/.q-system/tests/fixtures/destructive-op-deny.reference.sh:439`

The pre-existing `_fleet_preview` early exit fires before the new sp-37c08fb1 layer (line 459) and the new sp-dc76e644 layer (line 511). The old FLEET_DENY loop was immune because its `emit_deny` fires inside the loop, before the exit. The two new layers sit after it, so any command block containing a `--dry` stage that matches an old pattern silently allows the apply stage.

Reproducer (harness written to `$TMPDIR/rev323_gna9qxqe`, cases in `cases4.json`):

```
python3 $TMPDIR/rev323_gna9qxqe/run.py cases4.json
```

Real output:

```
DENY   ( ./kipi-update.sh --only X )
ALLOW  ./kipi-update.sh --dry-run && ( ./kipi-update.sh --only X )
DENY   bash .../fp2/consulting-sync-run.sh --verify-only
ALLOW  ./kipi-update.sh --dry-run && bash .../fp2/consulting-sync-run.sh --verify-only
ALLOW  ./kipi-update.sh --dry && bash .../fp2/consulting-sync-run.sh
```

Line 1 versus line 2 is the whole PR: the 06:08Z incident subshell is refused alone and allowed the moment a preview precedes it. Line 3 versus line 4 does the same to the bypass-arming wrapper, which is the 2026-09-06 consulting sync verbatim.

Why this is the likely shape, not a contrived one: the deny message the hook prints tells the operator to preview first, and the older layers already carry the ASK-1118 warning that "a preview sharing a command block with the apply does not exempt the apply". The old layers enforce that. The new ones do not.

At 3am this is a `--delete` rsync into every registered instance with no token, and the operator sees an allow line in the audit log. Not a regression against `origin/main` (both shapes were allowed there too), so it does not gate merge as unrecoverable, but the PR claims these shapes are closed and they are not.

The shipped suite has no case for it. `test_destructive_op_deny_script_form.py:121` `DRY_RUN_ALONE` asserts preview-alone allows; nothing asserts preview-then-apply denies.

---

### FINDING 2 (major): layer 2 refuses 19 of this repo's own 212 tracked scripts, not the 6 the code and the install proposal claim

`q-system/.q-system/tests/fixtures/destructive-op-deny.reference.sh:539`

The comment states: "with it the fleet rule refuses 6, and all 6 genuinely invoke the fleet updater (they are its own test suite)". The same number is in `q-system/output/claude-changes/sp-37c08fb1-fleet-updater-script-form.json`, which is what the founder reads before running `apply-claude-changes` against the live hook.

Reproducer, driving the shipped hook on every `git ls-files '*.sh'`:

```
python3 $TMPDIR/rev323_gna9qxqe/pop.py
```

Real output (abridged to the classes):

```
tracked .sh in repo: 212
DENIED when invoked as bash <path>: 19
   kipi-push-upstream.sh                          || wrapper script runs the fleet updater
   q-system/hooks/auto-update.sh                  || wrapper script runs the fleet updater
   q-system/.q-system/scripts/lessons-daily.sh    || wrapper script runs the fleet updater
   plugins/kipi-core/scripts/install-capability-token.sh || wrapper script arms the founder bypass
   q-system/.q-system/tests/fixtures/destructive-op-deny.reference.sh || wrapper script arms the founder bypass
   (+ 12 test scripts, + kipi-update.sh itself)
```

Fleet rule: 16, which is exactly the "without the drop" number the comment says the comment-drop eliminated. What the offenders actually contain:

- `q-system/hooks/auto-update.sh` matches on `echo "    Run: kipi update"`. It prints the words. It invokes nothing.
- `kipi-push-upstream.sh` matches on a Python `stderr.write` string embedded in the shell script.
- `install-capability-token.sh` and the fixture itself match the bypass rule on the hook's own deny text, `set ALLOW_DESTRUCTIVE=1 to bypass all`, which is a code line and not a comment line, so the drop never reaches it.

Two consequences. First, the deny text asserts "this rsyncs the skeleton into EVERY registered instance with a delete flag" about a script that echoes a help string. That is the cry-wolf class, and this repo's own rules say a gate red on its population gets switched off. Second, `install-capability-token.sh` is the installer for the `kipi-approve` escape hatch that `emit_deny` depends on and that the PR names as the accepted cost ("the out-of-band approval already exists for exactly this"). The gate now blocks installing its own escape hatch through the Bash tool.

Population scope is also wrong. The hook is global (`~/.claude/hooks/`), so its population is not 212 files in one repo. Measured across `~/projects` (read outside the worktree on purpose, to size a global hook's blast radius; counts only, no contents printed):

```
total .sh under ~/projects: 21157
would be DENIED by the new wrapper-body fleet rule: 1227 (5.8%)
would be DENIED by the bypass-assignment rule: 180
```

Constructed confirmation of the class, three ordinary scripts nobody in this fleet wrote for the updater:

```
DENY   bash .../fp/backup-photos.sh     # rsync -a --delete /Volumes/Photos/ /Volumes/Backup/
DENY   bash .../fp/deploy-site.sh       # rsync -az --delete ./dist/ user@host:/var/www/
DENY   bash .../fp/hint.sh              # echo 'run kipi update --dry first'
ALLOW  bash .../fp/innocent.sh
```

---

### FINDING 3 (minor): the new layer's command-position skip covers parens and `VAR=`, not `timeout` / `setsid` / `exec` / `eval`, and a quoted path defeats it

`q-system/.q-system/tests/fixtures/destructive-op-deny.reference.sh:500` and `:568`

`_FLEET_SCRIPT_RE` skips opening parens/braces and env assignments. Layer 2's transparent-prefix list is `sudo|command|nohup|nice|time|env`. Its `*.sh)` case tests the raw word, so a quoted token never matches.

Reproducer, with a real `kipi-update.sh` present in the payload cwd so layer 2 has a readable target:

```
python3 $TMPDIR/rev323_gna9qxqe/run2.py cases2.json
```

Real output:

```
ALLOW  timeout 900 ./kipi-update.sh --only X
DENY   nohup ./kipi-update.sh --only X
ALLOW  "./kipi-update.sh" --only X
ALLOW  setsid ./kipi-update.sh --only X
ALLOW  eval ./kipi-update.sh --only X
ALLOW  exec ./kipi-update.sh --only X
DENY   ./kipi-update.sh --only X
```

Dropped a level per the calibration rule: I found no producer in this repo that emits any of these shapes for the updater. `nohup` is covered, its neighbours are not.

---

### FINDING 4 (minor): the PR body describes a different change

The description is almost entirely about `q-system/.q-system/scripts/pr-review-agent.sh`: commit-status ordering, a duplicate-head marker, pid files, a STRUCTURAL two-round stop, and a 202/202 severity-floor run. None of that is in the diff.

```
gh -R assafkip/kipi-system pr diff 323 --name-only
```

```
q-system/.q-system/capability/expected_tests/...test_destructive_op_deny_script_form.py.json
q-system/.q-system/tests/fixtures/destructive-op-deny.reference.sh
q-system/.q-system/tests/test_destructive_op_deny_script_form.py
q-system/output/claude-changes/sp-37c08fb1-fleet-updater-script-form.json
```

Only the trailing "Third commit" section matches. The "Test evidence" table and the "What is NOT verified here" list belong to another PR, and the second of those is what a founder reads before applying a hook change to `~/.claude`.

---

### FINDING 5 (minor): the only claimed install signal fails on one laptop and skips everywhere else

`test_destructive_op_deny_anchor.py:477`

The new suite's docstring delegates the install reminder to one place: "the anchor suite's `test_the_vendored_copy_has_not_drifted` compares the fixture to the live hook byte for byte and goes red the moment they differ. That redness is the reminder."

```
python3 -m pytest test_destructive_op_deny_anchor.py -q
```

```
E   AssertionError: the reference fixture and the live hook have diverged.
E   At index 22312 diff: b'\n' != b'f'
1 failed, 94 passed in 21.74s
```

That is red on merge, as designed. The problem is the branch four lines above it: `if not HOOK.is_file(): pytest.skip(...)`. On CI, and on any machine without `~/.claude/hooks/destructive-op-deny.sh`, the drift test skips. So the signal is red exactly where it annoys, absent exactly where a machine could act on it, and the same suite's own docstring says a suite that skips reads exactly like a suite that passes.

---

## What is sound

Attacks the code survived, all run:

- **The tests are load-bearing, not decoration.** I excised only the sp-37c08fb1 layer 1 block (3565 bytes) from a `$TMPDIR` copy and ran the shipped suite against it via `KIPI_DESTRUCTIVE_HOOK`: `7 failed, 39 passed`, naming the subshell, the brace group, the nested subshell and the 06:08Z incident string. The claim that every negative case was watched RED first holds for layer 1.
- **The suite as shipped is green:** `46 passed in 9.66s`.
- **The paren skip does not swallow reads.** `( cat kipi-update.sh )`, `sed -n '1,20p' kipi-update.sh`, `wc -l projects/.../kipi-update.sh` all ALLOW. The 2026-08-07 "unanchored regex blocked a read" scar is not reintroduced.
- **The `.py` helper is untouched.** All three call shapes of `kipi-update-instance-ahead.py` ALLOW.
- **Preview-alone still passes** in every script spelling, including inside a subshell.
- **The pre-existing denials did not move.** `rm -v -rf`, `git reset --hard`, `git push --force`, `git clean -fd`, `kipi update` all still DENY; `rm /tmp/one-file.txt`, `ls -rf /tmp`, `git push origin main` still ALLOW.
- **Layer 2 does not crash or deny on a missing target,** resolves a relative target against the payload `cwd`, and exempts `bash -n`.
- **Process substitution rather than a pipe** for both loops is correct: a pipe would run `emit_deny` in a subshell and the parent would log an allow after the deny JSON went out.

## VERDICT: REQUEST CHANGES

Two major findings, neither of them permanent or unrecoverable, so this is not a BLOCK. Merging is still a net improvement on `origin/main`, which allowed every shape here.

**Fix first:** move the two new layers above the `_fleet_preview` early exit at `reference.sh:439`, or fold their patterns into that loop's per-stage deny. Right now one `--dry-run` in the same command block turns off both of the things this PR was written to add, including the exact subshell from 06:08Z.

FINDINGS:
major|A preview stage hits the _fleet_preview early exit at line 439, so both new layers never run and `./kipi-update.sh --dry-run && ( ./kipi-update.sh --only X )` is ALLOWED|q-system/.q-system/tests/fixtures/destructive-op-deny.reference.sh:439
major|Layer 2 refuses 19 of the repo's 212 tracked scripts, not the 6 the comment and the install proposal claim, including scripts that only print the words "kipi update" and the capability-token installer the design depends on|q-system/.q-system/tests/fixtures/destructive-op-deny.reference.sh:539
minor|The new command-position skip misses timeout/setsid/exec/eval prefixes and a quoted script path, all measured ALLOW with the real updater present|q-system/.q-system/tests/fixtures/destructive-op-deny.reference.sh:500
minor|The PR description documents pr-review-agent.sh changes that are not in this diff, including the "What is NOT verified here" list a founder reads before applying to ~/.claude|q-system/output/claude-changes/sp-37c08fb1-fleet-updater-script-form.json:1
minor|The only claimed install signal, the anchor drift test, fails locally and skips on any machine without a live hook, so no automated run can observe that the fix never installed|test_destructive_op_deny_anchor.py:477
END FINDINGS

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-323-20260907-160834.md` (7238 bytes).

```
FINDINGS:
major|A --dry-run flag on a wrapper script exempts the wrapper's hardcoded real fleet apply, and lessons-daily.sh is a live producer of that shape|q-system/.q-system/tests/fixtures/destructive-op-deny.reference.sh:661
minor|The matched line of the target script is spliced unescaped into log_decision's printf JSON, corrupting or forging the audit record|q-system/.q-system/tests/fixtures/destructive-op-deny.reference.sh:663
minor|bash -lc and /bin/bash -c with a quoted single-stage updater path are still allowed by both new layers|q-system/.q-system/tests/fixtures/destructive-op-deny.reference.sh:490
minor|The install proposal's reason still claims 6 of 212 scripts refused when the author's own re-measure and mine both give 16, including a non-test scheduled job|q-system/output/claude-changes/sp-37c08fb1-fleet-updater-script-form.json:4
END FINDINGS
```

--- reviewer output ---

## Review of PR #323 — destructive-op-deny script-form and wrapper layers

Drove the fixture hook directly (copy in `$TMPDIR/pr323`), never the live one. Repo untouched.

---

### MAJOR — `--dry-run` on a wrapper exempts the wrapper's hardcoded real apply

`q-system/.q-system/tests/fixtures/destructive-op-deny.reference.sh:661`

The wrapper loop computes `_fleet_hit` from the script body with its own `--dry` filter, then discards the deny when the COMMAND LINE stage carries `--dry`. The flag lands on the wrapper. The hook has no idea whether the wrapper forwards it. The bypass rule two lines up gets this right in its own comment ("a --dry-run on the wrapper says nothing about whether the wrapper arms the bypass"); the fleet rule does not.

This is not a constructed input. `q-system/.q-system/scripts/lessons-daily.sh` parses no arguments at all and at line 104 runs `bash kipi-update.sh` — a real apply.

Reproducer and real output:

```
$ python3 - <<'PY'
from drv import decide
real = os.path.abspath("q-system/.q-system/scripts/lessons-daily.sh")
for cmd in ("bash %s" % real, "bash %s --dry-run" % real):
    print(decide(cmd)[0], "|", cmd)
PY

deny   | bash .../q-system/.q-system/scripts/lessons-daily.sh
allow  | bash .../q-system/.q-system/scripts/lessons-daily.sh --dry-run
```

Mutation, deleting line 661 only:

```
mutant (wrapper preview-exemption removed):
   deny   | bash nightly-apply.sh --dry-run
   deny   | bash kipi-update.sh --dry-run  (lessons-daily, real file)
   allow  | bash kipi-update.sh --dry           <- real preview still passes
   allow  | ./kipi-update.sh --dry-run          <- real preview still passes
1 failed, 68 passed
   FAILED ...::test_that_wrapper_is_exempt_on_a_preview_stage
```

The line is not load-bearing. `kipi-update.sh`'s own body has zero command-position hits (measured), so direct previews survive without it. The only test that goes red is `test_destructive_op_deny_script_form.py:254`, the one that asserts the hole, and its fixture `RUNS_THE_UPDATER` hardcodes `--only consulting` with no `$@`. That is the oracle picked from inside the model.

At 3am: an agent asked to preview a daily job types `--dry-run` on the wrapper and gets a full fleet rsync with `--delete` into every registered instance. Same shape as 2026-08-07, 19 instances. Recoverable from git, so major and not blocker.

---

### MINOR — a script's own text lands unescaped in the audit-log JSON

`q-system/.q-system/tests/fixtures/destructive-op-deny.reference.sh:663` feeding `:39`

`$_fleet_hit` is a raw line from the target file, interpolated into the deny reason, which `log_decision` splices into `printf '..."reason":"%s"...'` with no escaping. First layer in this file to put arbitrary file content there.

```
wrapper body: bash /Users/x/kipi-update.sh --only "$INSTANCE"
hook decision: deny
audit log: JSON BROKEN: Expecting ',' delimiter: line 1 column 302
```

Crafted body content forges the record outright:

```
hook decision: deny
audit log parses; recorded decision = allow
```

The security decision is correct in both cases. Only the forensic record lies. No JSON reader of `~/.claude/audit/destructive-op-deny.log` exists in this repo, so bounded.

---

### MINOR — `bash -lc` and `/bin/bash -c` with a quoted single-stage path still pass

`q-system/.q-system/tests/fixtures/destructive-op-deny.reference.sh:490`

```
deny   | nohup bash -c 'cd /Users/x/... && ./kipi-update.sh --only X' ...
deny   | bash -c "cd /Users/x && ./kipi-update.sh --only X"
allow  | /bin/bash -c '/Users/x/kipi-update.sh --only X'
allow  | bash -lc './kipi-update.sh --only X'
allow  | ssh mini './kipi-update.sh --only X'
```

Layer 1's anchor never reaches inside the `-c` string, and `_fleet_script_target` keeps the leading quote on the candidate so `[ -f ]` misses. The `&&` forms deny only because the split happens to create a stage. Dropped a level: I constructed these and found no producer in this repo that emits them.

---

### MINOR — the install proposal understates the measured blast radius

`q-system/output/claude-changes/sp-37c08fb1-fleet-updater-script-form.json:4` and `:17`

Both `reason` fields still say "the fleet rule refuses 6, all 6 the updater's own test suite". Commit `d0c2e06c` re-measured 16 of 212, and my independent sweep reproduces 16 exactly, including `lessons-daily.sh`, which is a scheduled job and not a test. The appended round-2 sentence says "re-measured after" without correcting the number. This string is what the founder reads at `apply-claude-changes.sh` time and what the install receipt keeps.

```
tracked .sh: 212
elapsed 24.2s  denied=16
  DENY kipi-update.sh
  DENY q-system/.q-system/scripts/lessons-daily.sh
  DENY ... 14 kipi-update test scripts
```

---

### What is sound

Attacks that failed, listed so the findings above can be trusted:

- **The install identity claim.** Rebuilt the PR fixture by applying the proposal's `insert_after` to `main`'s fixture: anchor occurs exactly once, `rebuilt == PR fixture: True`. The central untested claim holds mechanically.
- **Fail-open by crash.** Fuzzed 18 degenerate stages (`timeout`, `(`, `{`, `env`, bare quotes, `( { (`) against `/bin/bash 3.2.57`. Every one exited 0 with empty stderr. The bash 3.2 empty-array hazard the comments claim to handle really is handled.
- **The round-1 regression.** `./kipi-update.sh --dry-run && ( ./kipi-update.sh --only X )` denies. Both layers really do sit above the `_fleet_preview` exit.
- **False positives on the comment-drop.** Scripts that only echo the CLI phrase, carry the hook's own deny text, or run `rsync -a --delete` for a photo backup all pass. The command-position read did what the commit says.
- **Latency.** 10 runs per command, before and after: roughly +30 to +50 ms on a multi-stage command. Not worth a finding.
- **The disclosed red test.** `test_the_vendored_copy_has_not_drifted` fails here and the PR says so up front. CI has no live hook, so it skips there.

---

### VERDICT: REQUEST CHANGES

Not BLOCK. Nothing here is unrecoverable and the change closes two real holes.

Fix first: delete line 661 of the fixture (and the matching line in the proposal's `insert`), and rewrite `test_that_wrapper_is_exempt_on_a_preview_stage` into its honest form — a wrapper that forwards `"$@"` may be exempt, a wrapper that hardcodes an apply may not.

FINDINGS:
major|A --dry-run flag on a wrapper script exempts the wrapper's hardcoded real fleet apply, and lessons-daily.sh is a live producer of that shape|q-system/.q-system/tests/fixtures/destructive-op-deny.reference.sh:661
minor|The matched line of the target script is spliced unescaped into log_decision's printf JSON, corrupting or forging the audit record|q-system/.q-system/tests/fixtures/destructive-op-deny.reference.sh:663
minor|bash -lc and /bin/bash -c with a quoted single-stage updater path are still allowed by both new layers|q-system/.q-system/tests/fixtures/destructive-op-deny.reference.sh:490
minor|The install proposal's reason still claims 6 of 212 scripts refused when the author's own re-measure and mine both give 16, including a non-test scheduled job|q-system/output/claude-changes/sp-37c08fb1-fleet-updater-script-form.json:4
END FINDINGS

PR #86, opened 2026-08-03
The Opus reviewer fallback fires when codex produced nothing, not when it exits non-zero (ASK-287)

assafkip:
## The defect

`codex exec` out of credits **exits 0**. Measured 2026-08-02 with a real billed call:

```
codex exec --skip-git-repo-check --model gpt-5.6-sol "Reply with exactly: ALIVE"
ERROR: Your workspace is out of credits. Add credits to continue.
```

So `if run_engine codex` read the outage as SUCCESS, control took the "codex answered with nothing parseable" branch — which by design does **not** fall back — and the Opus fallback promised at `pr-review-agent.sh:50-51` had never fired once since the day it was written.

On disk: 0-byte reviews `pr-66-20260802-092103.md` / `pr-67-20260802-092225.md`, `degraded.state=1` (the system noticed), and no `pr-reviews/claude/` directory at all. Both PRs sat CI-green on `kipi/reviewer-approved` failure. Nothing merged.

The issue's hypothesis was confirmed by a live probe, not inherited.

## The fix

Stop asking the exit code; ask the artifact.

| evidence | reading | action |
|---|---|---|
| no content at all | the engine never reviewed | **OUTAGE** → Opus fills the slot, DEGRADED |
| content that does not parse | an attempted review | **GARBAGE** → UNSTATED, no fallback |

The second rule is untouched and stays that way (the issue's "Not doing"). An outage leaves no review to trust; garbage *is* a review whose content cannot be trusted, and filling the gate with an Opus approval over it invents a verdict for a review that said nothing.

`review_has_complete_findings_block` is unchanged. The no-fallback-on-garbage comment survives and is extended, not replaced.

## Honest boundary (stated in code, captured as sp-ee65243b)

`engine_produced_nothing()` reads **emptiness**, the shape actually measured. An outage that prints an error banner and exits 0 carries content, so it still takes the garbage path and posts UNSTATED — fail-closed (PR held, no approval fabricated), a wedge rather than a fabricated green. Widening that predicate needs a captured real-producer sample, not a guess.

## Check

`test-severity-floor.sh` (already registered in `capability-manifest.json`):

- **Q4** empty codex → the Opus fallback runs, fills `kipi/reviewer-approved` with state success marked DEGRADED, pages once, and the record's verdict is **derived** from the fallback's own findings block (`source: findings`), not invented.
- **Q4B** garbage codex → no fallback, state failure, and the verdict record carries **no verdict at all**. The property most at risk; asserted on the record, not just the status.
[...]

PR #326, opened 2026-09-08
One writer per launchd agents file, and a committed template for every loaded label (ASK-1345)

assafkip:
Closes the two findings from tonight's install-jobs work: `sp-3a513fce` (dead health-registration branch) and `sp-173ead80` (two writers, three orphan labels). They collapse: the dead branch lived inside the file the single-writer fix removes.

## Plan

1. Ground both findings by re-deriving every claim, not inheriting them.
2. Reproducer first for the writer half, red before green, with a ref hatch so the red stays re-runnable.
3. Pick the writer shape on the evidence and delete the loser.
4. Orphan labels: additive only. Commit a template where there is a real source, allowlist the rest with a date and a reason, and ship the gate only if it is green on this machine's real population.

## Phase 1, the five measurements

**1. Does `launchd-health-register.sh` exist anywhere.** No.

```
$ find /Users/assafkipnis/projects/kipi-system -name 'launchd-health-register*' -not -path '*/.git/*'
(no output)
$ git cat-file -e origin/main:q-system/.q-system/scripts/launchd-health-register.sh
fatal: path ... does not exist in 'origin/main'
$ git log --all --oneline --diff-filter=A -- '*launchd-health-register*'
(no output, over 3140 commits on all refs)
```

**2. Does the watchdog need registration.** No, and it never did.

```
watched prefixes: ('com.kipi.', 'com.cole.', 'com.claudedaddy.', ... 10 total)
enumerated label count: 70
target in enumeration: True          # com.kipi.voice-refresh
registration file referenced anywhere in module source: False
```

`discover_problems()` globs `~/Library/LaunchAgents/<prefix>*.plist` (line 185). The label nobody registered was already covered. So the message was wrong in both directions: it announced a watchdog absent that is present, and skipped a registration that does not exist.

**3. Do the two writers disagree.** Rendered both, byte-compared, did not reason about it.

Same tree, same plain root: identical, `md5 94ed436d9905a3c6fb27227885942250` both sides, `cmp` rc=0. That is why the duplication was invisible, not why it was safe. They diverge on the two inputs that matter:

```
# root containing '&'
A (python, XML-escaped): <string>.../R&amp;D/automation/voice-refresh-nudge.sh</string>
B (sed, raw):            <string>.../R&D/automation/voice-refresh-nudge.sh</string>
plutil -lint A: OK
plutil -lint B: (Encountered unknown ampersand-escape sequence at line 10)

# template with a second placeholder
writerA rc=0, and the file it wrote still contains: __HOME__
install-plist.sh: assert_rendered refuses (rc=1), nothing reaches LaunchAgents
```

Root resolution is the same from one checkout (both walk to the repo root); the divergence is in validation, not in the path.

**4. The three orphan labels, per label.**

| Label | Loaded | Runs | Program exists | Template committed anywhere |
|---|---|---|---|---|
| `com.kipi.audit-rotate` | 2026-05-18 | `~/.claude/audit/rotate.sh` nightly 23:55, rotates that dir's logs | yes, 347 bytes, executable | never, on any ref or worktree |
| `com.kipi.launchd-health` | 2026-06-30 | `/usr/bin/python3 <repo>/q-system/.q-system/scripts/launchd-health-check.py`, 09:30 and 21:30 | yes, committed repo script | never |
| `com.kipi.pr86-review` | 2026-08-07 | `~/.config/kipi/pr86-codex-review.sh` hourly at :17, one-shot, self-disables when PR #86 closes | yes, 8316 bytes; #86 measured still OPEN | never |

**5. Does a check already exist.** No, not for either question.

- `launchd-intent-verify.py` diffs loaded state against a DECLARED INTENT manifest, a different axis, and its manifest `~/.config/kipi/launchd-intent.json` does not exist on this machine, so it is inert here.
- `test-install-jobs-coverage.py` walks committed to installed. The reverse direction had no checker.
- Nothing anywhere asserted one writer per agents file.

Both findings are real. Neither half was dropped.

## Phase 2, the writer

**Pick: delete `automation/install-voice-refresh.sh`. `install-plist.sh` is the only writer.**

Not a thin caller, because a thin caller leaves a second entry point that can drift, and the single-writer property is then unprovable by absence. The evidence: `install-plist.sh` already reaches this template (the `git ls-files` enumerator in 6668550d, which added `__ROOT__` for this exact file), and it has `assert_rendered`, `plutil -lint`, stage-then-move, and `bootout`/`bootstrap`. The deleted writer had none of those and measurably installed a holed plist at exit 0.

Honest tradeoff: the deleted writer XML-escaped the root and the survivor does not. `plutil -lint` turns that case into a loud refusal instead of a silent corruption, so it is a fail-safe rather than a regression to silence.

**Reproducer red** (`KIPI_SINGLE_WRITER_REF=6668550d`, the whole fixture built from the pre-fix ref):

```
FAIL P1 SINGLE WRITER: expected exactly 1 shipped writer for com.kipi.voice-refresh, found 2:
     automation/install-voice-refresh.sh, q-system/.q-system/scripts/install-plist.sh
FAIL P3 REFUSES A HOLE: automation/install-voice-refresh.sh exited 0 and installed a plist
     still carrying __PROBE_TOKEN__
ok   detector self-test: the read-only enumerator is not a writer
ok   launchctl stub sealed (canary recorded)
ok   P2 idempotent across 4 install runs
ok   P3 install-plist.sh refused the holed template (rc=1)
EXIT=1
```

**Green** (working tree):

```
writers for com.kipi.voice-refresh (non-test): 1
    q-system/.q-system/scripts/install-plist.sh
ok   P1 single writer
ok   P2 idempotent across 2 install runs
ok   P3 install-plist.sh refused the holed template (rc=1)
EXIT=0
```

Two corrections came out of running it rather than reading it, both in the commit message: the first probe token was `__HOME__`, which `install-plist.sh` substitutes, so the survivor was graded on a template with no hole in it; and P3 first accepted `rc 0 and no leftover`, which is also what a writer that silently did nothing looks like.

The suite pinned the defect. `test_installer_registers_health` asserted the false message was PRESENT. Inverting it was part of the fix.

## The transient, said plainly

The founder's canonical checkout at `/Users/assafkipnis/projects/kipi-system` sits on a branch that does not carry the merge, so **until that tree is updated it still holds the old `install-plist.sh` and no longer holds `install-voice-refresh.sh` once it does update.** Concretely:

- Running `bash automation/install-voice-refresh.sh` from that tree **before** it carries main: still works, still prints the false "launchd-health watchdog absent" line. Harmless, and it is the path that was used tonight.
- **After** that tree carries main: the file is gone and that command fails with "No such file or directory". The replacement is `bash q-system/.q-system/scripts/install-plist.sh com.kipi.voice-refresh`, or `--all` from the primary checkout.
- Nothing already loaded changes on its own. No job is rewritten by this merge; the change only affects the next deliberate install.

## Phase 3, the orphan labels (additive only)

Nothing is unloaded, disabled or deleted. Removing a running job is the founder's call.

**Resolved additively: `com.kipi.launchd-health`.** Template committed, transcribed from the live copy, not invented. Rendered and diffed field by field against the loaded plist: `Label`, `StartCalendarInterval`, `StandardOutPath`, `StandardErrorPath` identical; `ProgramArguments` differs only by the repo root, which is the placeholder doing its job. Marked `kipi-scope: skeleton-only`, because `launchd-health-check.py` propagates to instances and this label is a per-machine singleton, so `--all` from any of 23 instance checkouts would otherwise rebind the one watchdog.

**Left for the founder, allowlisted with a date and a reason in the check itself:**

- `com.kipi.audit-rotate` — its program lives in the founder's global config, outside this repo and unreferenced by it. A template pointing at a script a rebuild does not restore is half a fix in the shape of a whole one. The decision is adopt the script into the repo, or leave the job machine-local.
- `com.kipi.pr86-review` — a one-shot for PR #86, which is still OPEN, built to self-disable when the review lands. A committed template would resurrect a job designed to delete itself. Expected to age out on its own.

**The gate ships green on its real population**, which is the only condition worth shipping it under:

```
loaded com.kipi. labels ........ 15
committed templates ..... 16
loaded with no template .. 2     (both ALLOWLISTED, each with its reason)
ok   every loaded com.kipi. label has a committed template or a dated exemption
EXIT=0
```

Negative self-test before trusting that green: with `ALLOWLIST` blanked it exits 1 and names both labels.

Two choices that keep it satisfiable rather than decorative. A stale allowlist entry WARNS instead of failing, because `pr86-review` is built to unload itself and a gate that goes red on the good outcome gets switched off. And it returns `None` rather than an empty set where `launchctl` does not exist, so CI on Linux prints SKIP instead of a clean pass about a population it never saw.

The committed set is read from `install-plist.sh`'s own usage listing rather than a second `git ls-files`, so the check and the installer cannot drift.

## Other gates

- `test-install-jobs-coverage.py`: `install-jobs: 13 installed, 3 skipped, 0 failed (of 16 committed)`, all 11 checks pass.
- `automation/test_voice_refresh_schedule.py`: 3 ok.
- `prd_runner.py gates run` is RED on 64 pre-existing major spillover items unrelated to this branch. Both items this PR closes are in the non-blocking bucket.
- Also noted in the new template and not fixed here: `sp-41409ffe` is open against the two fire times it freezes (6h ping TTL against a 12h cadence).

https://claude.ai/code/session_01PpVeUeZnvRVSb6uxzTPL28


assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-326-20260907-183935.md` (277671 bytes).

```
FINDINGS:
minor|Deleting the XML-safe installer makes com.kipi.voice-refresh impossible to install from a valid checkout path containing an ampersand|q-system/.q-system/scripts/install-plist.sh:214
END FINDINGS
```

--- reviewer output, last 54790 bytes of 277671 (full review at the path above) ---

 ...s__test__test-single-writer-agents-file.py.json |   4 +
 .../scripts/test/test-single-writer-agents-file.py | 312 +++++++++++++++++++++
 2 files changed, 316 insertions(+)
e738de3e fix(launchd): commit the launchd-health template, gate loaded-vs-committed (ASK-1345)
 ...s__test__test-loaded-label-has-template.py.json |   4 +
 .../scripts/com.kipi.launchd-health.plist          |  59 ++++++++
 .../scripts/test/test-loaded-label-has-template.py | 159 +++++++++++++++++++++
 3 files changed, 222 insertions(+)
c3787b94 fix(launchd): a failed launchctl is not an empty inventory (ASK-1345)
 .../scripts/test/test-loaded-label-has-template.py | 67 +++++++++++++++++++---
 1 file changed, 59 insertions(+), 8 deletions(-)
commit c3787b945ed3096a893e9750bf49f695378447f6
AuthorDate: Mon Sep 7 18:39:01 2026 -0700
CommitDate: Mon Sep 7 18:39:01 2026 -0700

    fix(launchd): a failed launchctl is not an empty inventory (ASK-1345)
    
    PR #326 round 1, codex major. The coverage checker treated a FAILED
    `launchctl list` as an empty successful inventory and then reported complete
    coverage over a population it never obtained.
    
    The reviewer proved it by running the file in its own sandbox:
    
        launchctl_rc=1  launchctl_stdout=''  launchctl_stderr=''
        test_rc=0
        loaded com.kipi. labels ........ 0
        ok   every loaded com.kipi. label has a committed template or a dated exemption
    
    Real, and worse than an ordinary miss: this file exists to catch exactly the
    class of "clean pass over a population nobody looked at", and it committed it.
    The first cut guarded only on `shutil.which`, which is the Linux-CI case. The
    binary was on PATH and the CALL failed, so the guard never fired.
    
    Now all three ways to fail to ask return None: no binary, a non-zero exit, or no
    output. `launchctl list` on a live session always prints a header row, so empty
    stdout at rc 0 is a session this process cannot see, not a machine with no jobs.
    
    The guard is held permanently by a self-test inside the shipped file rather than
    by a scratch script: a stub launchctl that exits 1 while sitting on a sealed PATH
    must produce None. It runs on every invocation, and it asserts the seal first so
    it cannot pass by accident.
    
    Both directions measured, not assumed:
      control  stub launchctl exits 1  ->  SKIP, no coverage claim (was: false pass)
      mutant   rc/empty guard deleted  ->  self-test raises, exit 1 (guard held)
      live     unchanged: 15 loaded, 16 committed, 2 allowlisted, exit 0
    
    Co-Authored-By: Claude Opus 5 (1M context) <noreply@anthropic.com>
    Claude-Session: https://claude.ai/code/session_01PpVeUeZnvRVSb6uxzTPL28

diff --git a/q-system/.q-system/scripts/test/test-loaded-label-has-template.py b/q-system/.q-system/scripts/test/test-loaded-label-has-template.py
index 2b615718..c8c6131b 100755
--- a/q-system/.q-system/scripts/test/test-loaded-label-has-template.py
+++ b/q-system/.q-system/scripts/test/test-loaded-label-has-template.py
@@ -42,10 +42,12 @@ Run: python3 q-system/.q-system/scripts/test/test-loaded-label-has-template.py
 Exit 0 = pass. Exit 1 = a loaded label has no template and no allowlist entry.
 """
 
+import os
 import re
 import shutil
 import subprocess
 import sys
+import tempfile
 from pathlib import Path
 
 HERE = Path(__file__).resolve().parent
@@ -79,16 +81,33 @@ ALLOWLIST = {
 def loaded_labels():
     """Labels launchd currently has bootstrapped, for this prefix.
 
-    None (not an empty list) when there is no launchd to ask. CI runs on Linux,
-    where `launchctl` does not exist: an empty set there would read as "no loaded
-    label lacks a template" and this check would report a clean pass about a
-    population it never saw. Distinguishing the two is the whole point.
+    None (not an empty list) when launchd could not be ASKED. An empty set would
+    read as "no loaded label lacks a template" and this check would report a clean
+    pass about a population it never saw. Distinguishing the two is the whole
+    point, and there are three ways to fail to ask, not one.
+
+    The first cut guarded only on `shutil.which`, which is the Linux-CI case. That
+    is not enough, and the reviewer proved it on PR #326 round 1 by running this
+    file inside the codex review sandbox:
+
+        launchctl_rc=1  launchctl_stdout=''  launchctl_stderr=''
+        test_rc=0
+        loaded com.kipi. labels ........ 0
+        ok   every loaded com.kipi. label has a committed template or a dated exemption
+
+    The binary was on PATH and the CALL failed, so the guard never fired and a
+    checker whose whole job is this failure class committed it. Hence: the binary
+    must exist, the call must exit 0, AND it must return output. `launchctl list`
+    on a live session always prints a header row, so empty stdout at rc 0 is a
+    session this process cannot see rather than a machine with no jobs.
     """
     if shutil.which("launchctl") is None:
         return None
-    out = subprocess.run(["launchctl", "list"], capture_output=True, text=True,
-                         timeout=60).stdout
-    return sorted(set(re.findall(r"com\.kipi\.[A-Za-z0-9_.-]+", out)))
+    r = subprocess.run(["launchctl", "list"], capture_output=True, text=True,
+                       timeout=60)
+    if r.returncode != 0 or not r.stdout.strip():
+        return None
+    return sorted(set(re.findall(r"com\.kipi\.[A-Za-z0-9_.-]+", r.stdout)))
 
 
 def committed_labels():
@@ -112,10 +131,42 @@ def committed_labels():
     return sorted(set(labels))
 
 
+def selftest_failed_launchctl_is_not_an_empty_inventory():
+    """The guard that PR #326 round 1 caught missing, held permanently.
+
+    A stub launchctl that exits 1 while sitting on PATH must produce None, not an
+    empty list. Runs on every invocation because it is one subprocess and the
+    alternative is a guard nobody executes until the day it is already wrong.
+    """
+    work = tempfile.mkdtemp(prefix="loaded-label-selftest-")
+    try:
+        bindir = Path(work) / "bin"
+        bindir.mkdir()
+        stub = bindir / "launchctl"
+        stub.write_text("#!/bin/bash\nexit 1\n")
+        stub.chmod(0o755)
+        saved = os.environ.get("PATH", "")
+        os.environ["PATH"] = "%s:/usr/bin:/bin" % bindir
+        try:
+            assert shutil.which("launchctl") == str(stub), (
+                "PATH not sealed, so this self-test proves nothing")
+            assert loaded_labels() is None, (
+                "REGRESSION: a failed launchctl call is being read as an empty "
+                "loaded-label inventory, which reports coverage over a population "
+                "that was never obtained")
+        finally:
+            os.environ["PATH"] = saved
+    finally:
+        shutil.rmtree(work, ignore_errors=True)
+
+
 def main():
+    selftest_failed_launchctl_is_not_an_empty_inventory()
+    print("ok   self-test: a failed launchctl is not an empty inventory")
     loaded = loaded_labels()
     if loaded is None:
-        print("SKIP no launchctl on this host, so there is no loaded-label "
+        print("SKIP launchd could not be asked on this host (no launchctl, a "
+              "non-zero exit, or no output), so there is no loaded-label "
               "population to check. This says nothing about the founder's "
               "machine; it says this host cannot answer the question.")
         return 0

exec
/bin/zsh -lc "sed -n '1,260p' q-system/.q-system/scripts/launchd-health-check.py; git ls-files -s q-system/.q-system/scripts/launchd-health-check.py q-system/.q-system/scripts/com.kipi.launchd-health.plist automation/com.kipi.voice-refresh.plist; rg -n \"def discover_problems|WATCHED|PREFIX|LAUNCH_AGENTS|slack|Linear|linear\" q-system/.q-system/scripts/launchd-health-check.py; sed -n '1,180p' q-system/.q-system/scripts/test/test-install-jobs-coverage.py; sed -n '180,280p' q-system/.q-system/scripts/test/test-install-jobs-coverage.py; sed -n '740,790p' validate-separation.py; sed -n '40,65p' .github/workflows/validate.yml; git diff --check 6668550d..HEAD" in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-326
 succeeded in 0ms:
#!/usr/bin/env python3
"""Watchdog for the founder's launchd jobs -- surfaces silent job deaths.

Auto-discovers every ~/Library/LaunchAgents/<prefix>*.plist for the OS's base
job families (WATCHED_PREFIXES) plus any instance-local families listed in
EXTRA_PREFIXES_FILE, and Slack-pings (deduped) on TWO silent-death modes:
  1. loaded but its last run exited non-zero (LastExitStatus via `launchctl list`)
  2. installed on disk but NOT loaded into launchd -- so it silently never runs.
     Scar 2026-07-05: com.cole.daily-video was present + scheduled 07:00 but
     unloaded, so it produced no video AND no failure ping (an unloaded job
     cannot ping). Mode 1 alone -- the old com.kipi.*-only check -- was blind to
     both the non-kipi families and the entirely-unloaded case.

Scar: the fractional-cxo income scanners (opp-scan, bolt-on-discovery) exited 127
every day for 6 days (2026-06-24..2026-06-30) after a `kipi update` rsync --delete
wiped their scripts out from under the plists. Nothing surfaced it -- the jobs just
silently stopped hunting income. A prompt cannot watch launchd; this job can. It is
the deterministic backstop that turns a silent 127 into a phone ping within hours.

Every finding also becomes a Linear issue (ASK-181), deduped forever by the SAME
kipi-key fleet-health-daily.py uses for its own launchd checks. A Slack line is
read once and scrolls away; the issue holds state. Slack stays the pointer.

Single notification channel: slack-notify.sh (founder-notifications rule). Silent
no-op if no webhook is configured, so this watchdog never breaks anything.

The watchdog always exits 0 -- it must never become the failing job it reports.

Usage:
  launchd-health-check.py             # check; ping + file Linear issues
  launchd-health-check.py --dry-run   # print findings only; no ping, no issue,
  launchd-health-check.py --dry       #   no state write (--dry / -n are aliases)

Dry mode makes ONE read-only Linear query (the dedup fetch) so its preview of a
permanent, undeletable action is the real number. It writes nothing: no issue, no
state file, no ping. Any flag that is not a dry alias is REFUSED, never run live.
"""
import json
import re
import subprocess
import sys
import time
import traceback
from pathlib import Path

SELF_LABEL = "com.kipi.launchd-health"
FAIL_PING_TTL_SECONDS = 6 * 3600  # re-ping a still-failing job at most this often
HERE = Path(__file__).resolve().parent
NOTIFY_SCRIPT = HERE / "slack-notify.sh"
DRY_FLAGS = ("--dry", "--dry-run", "-n")
STATE_FILE = Path.home() / ".config" / "kipi" / "launchd-health-state.json"
LAUNCH_AGENTS = Path.home() / "Library" / "LaunchAgents"

# Base launchd job families this OS's own automation uses. A prefix that matches
# nothing on a given machine is a harmless no-op. Instance- or client-specific
# families (client slack syncs, brand jobs) live in EXTRA_PREFIXES_FILE below, NOT
# here -- the skeleton propagates fleet-wide via kipi update and must carry no
# single instance's brand names.
WATCHED_PREFIXES = (
    "com.kipi.",
    "com.cole.",         # daily podcast/video GTM pipeline
    "com.claudedaddy.",  # social posters (Pinterest/X/YouTube/refill/repo)
    "com.ask.",          # ASK AI podcast
    "com.assaf.",        # competitive-analysis morning
)

# Instance-local additions, one prefix per line ('#' starts a comment). Lives in
# the founder's live env (~/.config/kipi/), outside the repo, so client/brand job
# families are watched without being baked into the propagated skeleton. Missing
# file = base set only (harmless no-op).
EXTRA_PREFIXES_FILE = Path.home() / ".config" / "kipi" / "launchd-watch-prefixes.txt"

# Labels that are not_loaded ON PURPOSE. A deliberately paused job is not a
# failure, and pinging about one is the alert-fatigue mechanism that teaches the
# founder to ignore this channel entirely.
#
# Scar 2026-07-26: 26 com.cole.* jobs were paused by commenting `com.cole.` out
# of EXTRA_PREFIXES_FILE. That could never work -- load_watched_prefixes() only
# ADDS from that file, so a commented line cannot remove a prefix hardcoded in
# WATCHED_PREFIXES. The pause was real, the silence was not, and one manual run
# fired 26 false pings. Suppression needs its own ledger, not a comment.
#
# Paused labels are still PRINTED (visible, not forgotten) and never pinged.
PAUSED_LABEL_FILES = (
    Path.home() / ".config" / "kipi" / "launchd-paused.txt",
    Path.home() / ".config" / "kipi" / "cole-pause.state",  # the live pause ledger
)


def load_paused_labels():
    """Exact labels (not prefixes) that are intentionally stopped."""
    paused = set()
    for path in PAUSED_LABEL_FILES:
        try:
            lines = path.read_text().splitlines()
        except (FileNotFoundError, NotADirectoryError):
            continue
        for line in lines:
            entry = line.split("#", 1)[0].strip()
            if entry:
                paused.add(entry)
    return paused


def load_watched_prefixes():
    """Base families plus any instance-local additions from EXTRA_PREFIXES_FILE."""
    prefixes = list(WATCHED_PREFIXES)
    try:
        lines = EXTRA_PREFIXES_FILE.read_text().splitlines()
    except FileNotFoundError:
        return tuple(prefixes)
    for line in lines:
        entry = line.split("#", 1)[0].strip()
        if entry and entry not in prefixes:
            prefixes.append(entry)
    return tuple(prefixes)


def normalize_exit(raw):
    """launchctl reports LastExitStatus as a raw wait(2) status. Decode to the
    human exit code: exit 3 arrives as 3<<8 = 768; a signal kill as 128+signal.
    A small value (<256) is already a clean code, so pass it through."""
    if raw == 0 or 0 < raw < 256:
        return raw
    exit_code = (raw >> 8) & 0xFF
    if exit_code:
        return exit_code
    signal_num = raw & 0x7F
    return 128 + signal_num if signal_num else raw


def classify_status(returncode, stdout):
    """Pure classifier for a `launchctl list <label>` result (returncode, stdout):
      ('failing', code)     loaded, last run exited non-zero
      ('not_loaded', None)  plist on disk but not bootstrapped -> never runs
      ('ok', 0)             loaded, last run clean (or a benign KeepAlive restart)
    Split out from job_status so the SIGTERM/live-PID logic below is unit-testable
    without shelling launchctl.

    Benign-restart carve-out (scar 2026-07-18): a KeepAlive job restarts on exit,
    and launchd records the PREVIOUS run's SIGTERM as LastExitStatus while a fresh
    process (a live PID) is already running. `com.cole.linkedin-session` logged
    LastExitStatus=15 with a live PID every cycle and the watchdog paged "exit 15"
    falsely. A SIGTERM (graceful stop, raw wait-status 15) WITH a live PID is that
    relaunch, not a death -> ok. Kept narrow to SIGTERM: a crash-loop
    (SIGKILL/SIGSEGV while KeepAlive relaunches) still pages, and a real exit(15)
    (raw wait-status 15<<8 = 3840, not 15) still pages."""
    if returncode != 0:
        return ("not_loaded", None)
    match = re.search(r'"LastExitStatus"\s*=\s*(-?\d+)', stdout)
    raw = int(match.group(1)) if match else 0
    code = normalize_exit(raw)
    if code == 0:
        return ("ok", 0)
    has_live_pid = re.search(r'"PID"\s*=\s*\d+', stdout) is not None
    terminated_by_sigterm = raw == 15  # raw wait-status 15 = killed by SIGTERM
    if has_live_pid and terminated_by_sigterm:
        return ("ok", 0)
    return ("failing", code)


def job_status(label):
    """Classify a launchd label by shelling `launchctl list <label>`:
      ('unknown', None)     launchctl unavailable -- do not alert
      otherwise -> classify_status(returncode, stdout)
    `launchctl list <label>` exits non-zero when the label is not loaded, which
    is how a silently-unloaded job (present plist, absent from launchd) is caught."""
    try:
        result = subprocess.run(
            ["launchctl", "list", label],
            capture_output=True, text=True, timeout=10,
        )
    except Exception:
        return ("unknown", None)
    return classify_status(result.returncode, result.stdout)


def discover_problems():
    """List (label, kind, detail) for every watched job that is failing or
    installed-but-unloaded. Watched = any watched-prefix plist, minus self."""
    problems = []
    seen = set()
    paused = load_paused_labels()
    for prefix in load_watched_prefixes():
        for plist in sorted(LAUNCH_AGENTS.glob(f"{prefix}*.plist")):
            label = plist.stem
            if label == SELF_LABEL or label in seen:
                continue
            seen.add(label)
            kind, code = job_status(label)
            if kind == "failing":
                problems.append((label, "failing", f"exit {code}"))
            elif kind == "not_loaded":
                if label in paused:
                    # Intentional. Reported so it stays visible, never pinged.
                    problems.append((label, "paused", "paused on purpose"))
                else:
                    problems.append((label, "not_loaded", "installed but not running"))
    return problems


def load_state():
    try:
        return json.loads(STATE_FILE.read_text())
    except Exception:
        return {}


def write_state(state):
    STATE_FILE.parent.mkdir(parents=True, exist_ok=True)
    STATE_FILE.write_text(json.dumps(state, indent=2))


_NOTIFY_CHANNEL = None


def notify_channel_configured():
    """True when a page could actually leave this machine.

    Borrowed, not re-derived: `fable-escalate.py` already owns the Python read of
    whether slack-notify.sh has a destination to write to, and a second copy here
    is the drift `fleet-homogeneity` exists to stop. Loaded through the same
    lazy-importlib shape as `_fleet_health` / `_intent_module`, so its absence
    cannot stop the watchdog.

    That destination is LINEAR, not Slack, since 2026-08-10. This paragraph used
    to name the webhook resolution order ($KIPI_SLACK_WEBHOOK, then
    ~/.config/kipi/slack-webhook) and it outlived the channel by four days --
    which is the same way sp-7773af84 stayed invisible. The resolution order is
    deliberately NOT restated here; it belongs to the function being borrowed.

    A load failure answers False -- "delivery unknown". The caller reads that as
    NOT delivered, which costs a repeated ping and never costs a missed one. That
    is the same trade `check()` in launchd-intent-verify.py already made: a
    duplicate is the failure this system is allowed to have.
    """
    global _NOTIFY_CHANNEL
    if _NOTIFY_CHANNEL is None:
        import importlib.util

        try:
            spec = importlib.util.spec_from_file_location(
                "fe", HERE / "fable-escalate.py")
            module = importlib.util.module_from_spec(spec)
            spec.loader.exec_module(module)
            _NOTIFY_CHANNEL = module.notify_channel_configured
        except Exception:  # noqa: BLE001
            _NOTIFY_CHANNEL = lambda: False  # noqa: E731
    return _NOTIFY_CHANNEL()


_NOTIFY_SENDER = False


def _notify_sender():
    """fable-escalate.py's `notify_send`, or None when it cannot be loaded.

    Same lazy-importlib borrow as `notify_channel_configured` above, and for the
    same reason: that module is the single Python reader of slack-notify.sh's
    delivery verdict, and a second copy here is the drift `fleet-homogeneity`
100644 bcef0373109cd27c181416ac2235b5113c02f459 0	automation/com.kipi.voice-refresh.plist
100644 f52fd9c9a7dbacbc54633aaa3dd990883e3922ff 0	q-system/.q-system/scripts/com.kipi.launchd-health.plist
100755 762cf12129082c5045c4572a9820bb2f2233386b 0	q-system/.q-system/scripts/launchd-health-check.py
5:job families (WATCHED_PREFIXES) plus any instance-local families listed in
6:EXTRA_PREFIXES_FILE, and Slack-pings (deduped) on TWO silent-death modes:
20:Every finding also becomes a Linear issue (ASK-181), deduped forever by the SAME
24:Single notification channel: slack-notify.sh (founder-notifications rule). Silent
30:  launchd-health-check.py             # check; ping + file Linear issues
34:Dry mode makes ONE read-only Linear query (the dedup fetch) so its preview of a
49:NOTIFY_SCRIPT = HERE / "slack-notify.sh"
52:LAUNCH_AGENTS = Path.home() / "Library" / "LaunchAgents"
56:# families (client slack syncs, brand jobs) live in EXTRA_PREFIXES_FILE below, NOT
59:WATCHED_PREFIXES = (
71:EXTRA_PREFIXES_FILE = Path.home() / ".config" / "kipi" / "launchd-watch-prefixes.txt"
78:# of EXTRA_PREFIXES_FILE. That could never work -- load_watched_prefixes() only
80:# WATCHED_PREFIXES. The pause was real, the silence was not, and one manual run
106:    """Base families plus any instance-local additions from EXTRA_PREFIXES_FILE."""
107:    prefixes = list(WATCHED_PREFIXES)
109:        lines = EXTRA_PREFIXES_FILE.read_text().splitlines()
178:def discover_problems():
185:        for plist in sorted(LAUNCH_AGENTS.glob(f"{prefix}*.plist")):
221:    whether slack-notify.sh has a destination to write to, and a second copy here
228:    ~/.config/kipi/slack-webhook) and it outlived the channel by four days --
259:    same reason: that module is the single Python reader of slack-notify.sh's
283:    `run_intent_check` recorded the run as seen regardless. slack-notify.sh is a
287:    webhook and a dead filer: 0 findings reached Linear, the message was never
292:    slack-notify.sh exits 0 after a curl that never reached Slack -- its last line
298:    The POST outcome exists in exactly one place -- inside slack-notify.sh -- so
316:    slack-notify.sh's fixture guard refuses to page on a loopback
319:    2 of the round-2 review); production has no loopback Linear URL.
363:# --- findings reach Linear, not only Slack -----------------------------------
368:# time, and Linear issues are not deleted in this fleet.
380:    Linear key and how it gets filed. Loaded on demand so a healthy run (the
393:def linear_findings(problems):
404:        # stale hash: a Linear rewrite and a "1 updated" Slack line twice a day on
407:        # fleet-health-daily.py loaded, and `file_linear_findings` already reports
416:def file_linear_findings(problems, apply=True):
417:    """File the findings as deduped Linear issues. Never raises.
419:    A watchdog that dies because Linear is unreachable stops watching launchd --
433:        findings = linear_findings(problems)
435:        print(f"linear findings could not be built: {exc}", file=sys.stderr)
443:        print(f"linear filing skipped: {exc}", file=sys.stderr)
458:# first time a new one lands upstream. That is exactly how a refused Linear write
472:# and Linear sends no notification for a no-op. So a drift whose issue was already
475:# slack=False and existing=1, and the real ping_decision then returns nothing for
501:def linear_report_line(outcome, dry_run):
506:    {created: 0, existing: 0, skipped_no_key: N}, so 'Linear is unreachable and N
508:    to file'. Every instance without a Linear API key takes that branch, and this
523:    return (f"{prefix}linear: {verb}={outcome['created']} "
546:    """File the coverage gap as ONE Linear issue. Pages nobody. Returns nothing.
564:    dies over a Linear failure stops watching launchd.
574:        print(f"intent coverage NOT filed to Linear: {exc}", file=sys.stderr)
580:    # linear_report_line, one detector over.
592:    same contract `file_linear_findings` keeps for Linear.
597:    one a reader defaults to (the `unfiled=` scar on `linear_report_line`, ASK-181).
631:            intent.linear_findings(drift, _fleet_health().finding_key),
640:        print(f"intent findings NOT filed to Linear: {exc}", file=sys.stderr)
646:        message += f" [NOT filed to Linear: {unfiled}]"
651:    # unconditionally, so a run where Linear filed nothing AND no webhook was
671:    # and part `existing` (already on the board, Linear notifies nobody) or part
691:        print(f"intent alert {what} (linear: {told} of {len(drift)} job(s) "
692:              f"newly filed or changed, slack: not delivered) -- run counts "
782:    filed = file_linear_findings(problems, apply=not dry_run)
783:    print(linear_report_line(filed, dry_run))
801:            message += f" [NOT filed to Linear: {unfiled_count(filed)}]"
#!/usr/bin/env python3
"""Pairs with `kipi install-jobs` -> install-plist.sh --all (sp-0c48b66c, sp-d621b67d).

THE SCAR THIS PINS. `kipi:22` advertises the verb as "Install every committed
launchd job on this machine". Measured 2026-09-07 from a checkout of main at an
arbitrary path, with a stub launchctl and a throwaway HOME:

    committed com.kipi.*.plist ......... 15   (git ls-files)
    installed by `--all` ............... 12
    announced as skipped ............... 2    (skeleton-only, by design)
    NEVER MENTIONED AT ALL ............. 1    (automation/com.kipi.voice-refresh.plist)
    exit code .......................... 0

`install-plist.sh` enumerated `"$SCRIPT_DIR"/com.kipi.*.plist`: ONE directory. A
committed template outside `q-system/.q-system/scripts/` was unreachable, and the
run said nothing and exited 0. A partial install that reports success is worse
than a failed one, because nothing downstream ever looks again: `detect_dark_jobs`
in fleet-health-daily.py enumerates `~/Library/LaunchAgents/*.plist`, so a job that
was never installed once is invisible to the watchdog too.

WHAT THIS TEST HOLDS (three properties, each able to go red on its own):
  1. COVERAGE   every committed label is either installed or announced as skipped
  2. RESOLVABLE every installed job's program path exists, with no live token left
  3. LOUD       a committed template that cannot be installed makes the run exit
                non-zero and names the label

The fixture is a real checkout built from the WORKING TREE at an arbitrary temp
path, not from HEAD: a reproducer that reads HEAD cannot see the fix you are
writing. The first cut of this measurement cloned the source repo's HEAD, landed
on a stale branch carrying 10 of the 15 plists, and described the fixture instead
of the installer. Hence _build_checkout below prints the tree it actually made.

Run: python3 q-system/.q-system/scripts/test/test-install-jobs-coverage.py
Exit 0 = pass.
"""

import importlib.util
import os
import plistlib
import re
import shutil
import subprocess
import sys
import tempfile
from pathlib import Path

HERE = Path(__file__).resolve().parent
REPO = HERE.parents[3]

# The committed templates carry `--` inside their XML comments (prose blocks
# citing PR rounds). CoreFoundation accepts that, so launchd loads them, but
# expat rejects the document outright. Borrow fleet-health-daily.py's own comment
# stripper rather than writing a second one: a fixture parser that drifts from
# the detector's parser measures a document neither of them sees.
_fh_spec = importlib.util.spec_from_file_location(
    "fh_for_coverage", HERE.parent / "fleet-health-daily.py")
_fh = importlib.util.module_from_spec(_fh_spec)
_fh_spec.loader.exec_module(_fh)


def _parse_plist(path: Path):
    return plistlib.loads(_fh._XML_COMMENT_RE.sub("", path.read_text()).encode())

failures = []


def check(name, got, want):
    if got != want:
        failures.append(f"{name}: got {got!r}, want {want!r}")
        print(f"  FAIL: {name}: got {got!r}, want {want!r}")
    else:
        print(f"  ok: {name}")


def _git(args, cwd, **kw):
    return subprocess.run(["git"] + args, cwd=str(cwd), capture_output=True,
                          text=True, timeout=180, **kw)


def committed_labels(tree: Path):
    """The set the verb promises, derived with git ls-files, never a glob."""
    out = _git(["ls-files", "*.plist"], tree).stdout.split()
    return sorted(Path(p).stem for p in out if Path(p).name.startswith("com.kipi."))


def _build_checkout(work: Path) -> Path:
    """A real primary checkout of the WORKING TREE, at a path nothing hardcodes.

    tar over `git ls-files` and not `git clone`: clone copies HEAD, and the whole
    point is to run the installer as it exists in the tree being reviewed. `git
    init` + one commit makes `.git` a DIRECTORY, which is what install-plist.sh
    requires before it will run --all (it refuses from a worktree on purpose).
    """
    dest = work / "an-arbitrary-checkout-path"
    dest.mkdir(parents=True)
    files = _git(["ls-files", "-z"], REPO).stdout.split("\0")
    listing = work / "tracked.txt"
    listing.write_text("\n".join(f for f in files if f))
    tar_c = subprocess.run(
        ["tar", "-cf", str(work / "tree.tar"), "-T", str(listing)],
        cwd=str(REPO), capture_output=True, text=True, timeout=300)
    if tar_c.returncode != 0:
        print(tar_c.stderr, file=sys.stderr)
        raise SystemExit("could not stage the working tree")
    subprocess.run(["tar", "-xf", str(work / "tree.tar")], cwd=str(dest),
                   capture_output=True, timeout=300)
    _git(["init", "-q"], dest)
    _git(["add", "-A"], dest)
    _git(["-c", "user.email=test@example.invalid", "-c", "user.name=coverage test",
          "commit", "-q", "-m", "fixture"], dest)
    return dest


def _stub_launchctl(work: Path) -> Path:
    """launchctl must never be the real one: this test would bootstrap live jobs."""
    stub = work / "bin" / "launchctl-stub"
    stub.parent.mkdir(parents=True, exist_ok=True)
    stub.write_text("#!/bin/bash\nexit 0\n")
    stub.chmod(0o755)
    return stub


def run_all(checkout: Path, work: Path, tag: str):
    home = work / f"home-{tag}"
    (home / "Library" / "LaunchAgents").mkdir(parents=True)
    env = dict(os.environ)
    env["HOME"] = str(home)
    env["KIPI_LAUNCHCTL"] = str(_stub_launchctl(work))
    proc = subprocess.run(
        ["bash", str(checkout / "q-system/.q-system/scripts/install-plist.sh"), "--all"],
        capture_output=True, text=True, timeout=600, env=env)
    return proc, home / "Library" / "LaunchAgents"


_INSTALLED = re.compile(r"^installed (com\.kipi\.[a-z0-9.-]+)", re.M)
_SKIPPED = re.compile(r"^\s*skipped \([^)]+\): (com\.kipi\.[a-z0-9.-]+)", re.M)


def main():
    work = Path(tempfile.mkdtemp(prefix="install-jobs-coverage-"))
    try:
        checkout = _build_checkout(work)
        print(f"fixture checkout: {checkout}")
        print(f"  .git is a directory: {(checkout / '.git').is_dir()}")

        promised = committed_labels(checkout)
        print(f"  committed com.kipi labels in the fixture: {len(promised)}")
        check("the fixture carries the same committed set as the repo",
              promised, committed_labels(REPO))

        proc, agents = run_all(checkout, work, "main")
        out = proc.stdout + proc.stderr
        print("---- installer output ----")
        print(out.rstrip())
        print("--------------------------")

        installed = set(_INSTALLED.findall(out))
        skipped = set(_SKIPPED.findall(out))

        # === 1. COVERAGE ===================================================
        unaccounted = sorted(set(promised) - installed - skipped)
        check("every committed label is installed or announced as skipped",
              unaccounted, [])

        check("every non-skipped committed label has a file in LaunchAgents",
              sorted(l for l in promised
                     if l not in skipped and not (agents / f"{l}.plist").exists()),
              [])

        # === 2. RESOLVABLE =================================================
        # A rendered plist that still carries __TOKEN__ is a job launchd accepts
        # and fails at fire time, silently. A program path that does not exist is
        # the same failure one step later.
        left_tokens, dead_paths = [], []
        for f in sorted(agents.glob("com.kipi.*.plist")):
            text = f.read_text()
            if re.search(r"__[A-Z_]+__", text):
                left_tokens.append(f.stem)
            args = _parse_plist(f).get("ProgramArguments", [])
            for a in args[1:]:
            for a in args[1:]:
                if a.startswith("/") and not Path(a).exists():
                    dead_paths.append(f"{f.stem}: {a}")
                    break
        check("no installed job carries an unsubstituted placeholder", left_tokens, [])
        check("every installed job's program path resolves", dead_paths, [])

        # === 3. LOUD =======================================================
        check("a fully-covered run reports a summary line",
              bool(re.search(r"install-jobs: \d+ installed", out)), True)
        check("a fully-covered run exits 0", proc.returncode, 0)

        # === THE NEGATIVE SELF-TEST ========================================
        # A control that fails the way an unrelated breakage fails proves nothing.
        # This one commits ONE template whose token no substituter knows, and the
        # assertion is not merely "non-zero": it is that this label is named in the
        # output and that the exit code is the installer's own partial-install code.
        control_label = "com.kipi.zz-unrenderable-control"
        ctrl = checkout / "automation" / f"{control_label}.plist"
        ctrl.write_text(
            '<?xml version="1.0" encoding="UTF-8"?>\n'
            '<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" '
            '"http://www.apple.com/DTDs/PropertyList-1.0.dtd">\n'
            '<plist version="1.0"><dict>\n'
            f'  <key>Label</key><string>{control_label}</string>\n'
            '  <key>ProgramArguments</key><array><string>/bin/bash</string>'
            '<string>__NO_SUBSTITUTER_KNOWS_THIS__/x.sh</string></array>\n'
            '</dict></plist>\n')
        _git(["add", "-A"], checkout)
        _git(["-c", "user.email=test@example.invalid", "-c", "user.name=coverage test",
              "commit", "-q", "-m", "negative control"], checkout)

        proc2, agents2 = run_all(checkout, work, "control")
        out2 = proc2.stdout + proc2.stderr
        check("the control template is REACHED by the enumerator (it is named)",
              control_label in out2, True)
        check("the control makes the run exit non-zero", proc2.returncode != 0, True)
        check("the control is reported as a failure, not as installed",
              control_label in set(_INSTALLED.findall(out2)), False)
        check("the control's broken render never reaches LaunchAgents",
              (agents2 / f"{control_label}.plist").exists(), False)
        check("the healthy labels still installed alongside the failing one",
              sorted(set(promised) - set(_INSTALLED.findall(out2)) - set(_SKIPPED.findall(out2))),
              [])
    finally:
        shutil.rmtree(work, ignore_errors=True)

    if failures:
        print("\nFAILED:")
        for f in failures:
            print(f"  - {f}")
        return 1
    print("\nall install-jobs coverage checks passed")
    return 0


if __name__ == "__main__":
    sys.exit(main())
    check(
        "_cadence-config exists (.yaml or .md)",
        file_exists(os.path.join(agents_dir, "_cadence-config.yaml")) or file_exists(os.path.join(agents_dir, "_cadence-config.md")),
    )
    check("_auto-fail-checklist.md exists", file_exists(os.path.join(agents_dir, "_auto-fail-checklist.md")))

    # --- Gate 1.1b: Claude agent model allocation ---
    print()
    print("  --- Gate 1.1b: Model allocation (.claude/agents) ---")
    ma_violations = model_allocation_violations(os.path.join(SCRIPT_DIR, ".claude", "agents"))
    for v in ma_violations:
        warn(v)
    check(f"Agent model IDs match the allocation policy ({len(ma_violations)} violations)", not ma_violations)

    # --- GATE 1.2: Scripts ---
    print()
    print("  --- Gate 1.2: Scripts ---")

    # --- Gate 1.2a: Capability gate (silent-absence class) ---
    # One implementation, two callers: CI invokes capability-gate.py directly
    # (validate.yml); kipi check enforces it here. CAPABILITY_GATE_SKIP=1 is
    # set ONLY by the CI validate-separation step so the suite is not executed
    # twice per CI run (finding-15, prd-silent-absence-capability-gate).
    # The skip is honored only from a caller that PROVES it runs the gate
    # elsewhere (CI runs it as its own step; kipi check just ran the instance
    # suite). A bare ambient CAPABILITY_GATE_SKIP=1 in a founder shell would
    # otherwise permanently and silently disable the gate (codex adversarial,
    # sag-callsite) — the exact bypass surface this PRD closes.
    _skip_proof = (os.environ.get("GITHUB_ACTIONS") == "true"
                   or os.environ.get("KIPI_GATE_SKIP_CALLER") == "kipi-check-instance")
    if os.environ.get("CAPABILITY_GATE_SKIP") == "1" and _skip_proof:
        print("  --- Gate 1.2a: capability gate SKIPPED (CAPABILITY_GATE_SKIP=1; caller runs it directly) ---")
    else:
        gate_script = os.path.join(SCRIPT_DIR, "q-system", ".q-system", "scripts", "capability-gate.py")
        gate_run = subprocess.run([sys.executable, gate_script, "--repo-root", SCRIPT_DIR],
        check("capability gate: declared-vs-actual diff + full test run exits 0",
              gate_run.returncode == 0)
        if gate_run.returncode != 0:
            print("\n".join(("    " + l) for l in
                            (gate_run.stdout + gate_run.stderr).splitlines()[-15:]))

    # --- Gate 1.2b: Memory hygiene sweep (ADVISORY, can never fail this gate) ---
    # memory-lint.py reads the auto-memory corpus, which lives OUTSIDE the repo
    # (~/.claude/projects/<slug>/memory/). Deliberately warn-only, in both
    # directions: a stale or unindexed memory is never a reason to refuse a repo
    # validation, and the corpus predates the as_of/status convention entirely --
    # 73 of 73 files carry neither field, so a failing gate here would be red on
    # its whole population from the first run and get switched off.
    # CLAUDE_PROJECT_DIR is pinned so the sweep resolves the same corpus whatever
    # directory kipi check was invoked from (two derivations of one path is how a
      # pytest must exist BEFORE the gate: declared tests import it, and the
      # gate runs the full suite (first CI execution failed on step order,
      # 2026-07-23).
      - name: Install test dependencies
        run: pip install pytest

      # The capability gate replaces the old hardcoded 4-test loop: it
      # discovers every test artifact by convention, diffs declared-vs-actual
      # BOTH directions, and runs the full declared suite. The hardcoded loop
      # ran 4 of 41 declared artifacts — 90.2% never executed anywhere
      # (prd-silent-absence-capability-gate-2026-07-23).
      - name: Capability gate (declared-vs-actual + full test suite)
        run: python3 q-system/.q-system/scripts/capability-gate.py --repo-root .
        env:
          CI: true

      # CAPABILITY_GATE_SKIP: the gate just ran as its own step above; running
      # it again inside validate-separation would execute the suite twice per
      # CI run (finding-15). The skip is honored only with caller proof
      # (GITHUB_ACTIONS here).
      - name: Run skeleton validation (Phase 1)
        run: python3 validate-separation.py 1 --verbose
        env:
          CI: true
          CAPABILITY_GATE_SKIP: "1"


exec
/bin/zsh -lc "python3 -c 'import plistlib,pathlib; ps=[pathlib.Path(\"q-system/.q-system/scripts/com.kipi.launchd-health.plist\"),pathlib.Path(\"automation/com.kipi.voice-refresh.plist\")]; [(lambda p: print(p, \"plistlib=PASS\") if not plistlib.loads(p.read_bytes().replace(b\"__KIPI_REPO__\",b\"/tmp/repo\").replace(b\"__HOME__\",b\"/tmp/home\").replace(b\"__USER__\",b\"user\").replace(b\"__ROOT__\",b\"/tmp/repo\")) is None else None)(p) for p in ps]'" in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-326
 exited 1 in 0ms:
    import plistlib,pathlib; ps=[pathlib.Path("q-system/.q-system/scripts/com.kipi.launchd-health.plist"),pathlib.Path("automation/com.kipi.voice-refresh.plist")]; [(lambda p: print(p, "plistlib=PASS") if not plistlib.loads(p.read_bytes().replace(b"__KIPI_REPO__",b"/tmp/repo").replace(b"__HOME__",b"/tmp/home").replace(b"__USER__",b"user").replace(b"__ROOT__",b"/tmp/repo")) is None else None)(p) for p in ps]
                                                                                                                                                                     ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~^^^
  File "<string>", line 1, in <lambda>
    import plistlib,pathlib; ps=[pathlib.Path("q-system/.q-system/scripts/com.kipi.launchd-health.plist"),pathlib.Path("automation/com.kipi.voice-refresh.plist")]; [(lambda p: print(p, "plistlib=PASS") if not plistlib.loads(p.read_bytes().replace(b"__KIPI_REPO__",b"/tmp/repo").replace(b"__HOME__",b"/tmp/home").replace(b"__USER__",b"user").replace(b"__ROOT__",b"/tmp/repo")) is None else None)(p) for p in ps]
                                                                                                                                                                                                                 ~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.14/3.14.6/Frameworks/Python.framework/Versions/3.14/lib/python3.14/plistlib.py", line 925, in loads
    return load(fp, fmt=fmt, dict_type=dict_type, aware_datetime=aware_datetime)
  File "/opt/homebrew/Cellar/python@3.14/3.14.6/Frameworks/Python.framework/Versions/3.14/lib/python3.14/plistlib.py", line 912, in load
    return p.parse(fp)
           ~~~~~~~^^^^
  File "/opt/homebrew/Cellar/python@3.14/3.14.6/Frameworks/Python.framework/Versions/3.14/lib/python3.14/plistlib.py", line 194, in parse
    self.parser.ParseFile(fileobj)
    ~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^
xml.parsers.expat.ExpatError: not well-formed (invalid token): line 4, column 28

exec
/bin/zsh -lc "sed -e 's|__KIPI_REPO__|/tmp/repo|g' -e 's|__HOME__|/tmp/home|g' -e 's|__USER__|user|g' q-system/.q-system/scripts/com.kipi.launchd-health.plist | plutil -lint /dev/stdin; sed -e 's|__ROOT__|/tmp/repo|g' automation/com.kipi.voice-refresh.plist | plutil -lint /dev/stdin; plutil -help | sed -n '1,80p'" in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-326
 succeeded in 23ms:
(The file “stdin” couldn’t be opened because you don’t have permission to view it.)
plutil: [command_option] [other_options] file...
The file '-' means stdin
Running in Swift mode
Command options are (-lint is the default):
 -help                         show this message and exit
 -lint                         check the property list files for syntax errors
 -convert fmt                  rewrite property list files in format
                               fmt is one of: xml1 binary1 json swift objc
                               note: objc can additionally create a header by adding -header
 -insert keypath -type value   insert a value into the property list before writing it out
                               keypath is a key-value coding key path, with one extension:
                               a numerical path component applied to an array will act on the object at that index in the array
                               or insert it into the array if the numerical path component is the last one in the key path
                               type is one of: bool, integer, float, date, string, data, xml, json
                               -bool: YES if passed "YES" or "true", otherwise NO
                               -integer: any valid 64 bit integer
                               -float: any valid 64 bit float
                               -string: UTF8 encoded string
                               -date: a date in XML property list format, not supported if outputting JSON
                               -data: a base-64 encoded string
                               -xml: an XML property list, useful for inserting compound values
                               -json: a JSON fragment, useful for inserting compound values
                               -dictionary: inserts an empty dictionary, does not use value
                               -array: inserts an empty array, does not use value
                               
                               optionally, -append may be specified if the keypath references an array to append to the
                               end of the array
                               value YES, NO, a number, a date, or a base-64 encoded blob of data
 -replace keypath -type value  same as -insert, but it will overwrite an existing value
 -remove keypath               removes the value at 'keypath' from the property list before writing it out
 -extract keypath fmt          outputs the value at 'keypath' in the property list as a new plist of type 'fmt'
                               fmt is one of: xml1 binary1 json raw
                               an additional "-expect type" option can be provided to test that
                               the value at the specified keypath is of the specified "type", which
                               can be one of: bool, integer, float, string, date, data, dictionary, array
                               
                               when fmt is raw: 
                                   the following is printed to stdout for each value type:
                                       bool: the string "true" or "false"
                                       integer: the numeric value
                                       float: the numeric value
                                       string: as UTF8-encoded string
                                       date: as RFC3339-encoded string in UTC timezone
                                       data: as base64-encoded string
                                       dictionary: each key on a new line
                                       array: the count of items in the array
                                   by default, the output is to stdout unless -o is specified
 -type keypath                 outputs the type of the value at 'keypath' in the property list
                               can be one of: bool, integer, float, string, date, data, dictionary, array
 -create fmt                   creates an empty plist of the specified format
                               file may be '-' for stdout
 -p                            print property list in a human-readable fashion
                               (not for machine parsing! this 'format' is not stable)
There are some additional optional arguments that apply to the -convert, -insert, -remove, -replace, and -extract verbs:
 -s                            be silent on success
 -o path                       specify alternate file path name for result;
                               the -o option is used with -convert, and is only
                               useful with one file argument (last file overwrites);
                               the path '-' means stdout
 -e extension                  specify alternate extension for converted files
 -r                            if writing JSON, output in human-readable form
 -n                            prevent printing a terminating newline if it is not part of the format, such as with raw
 --                            specifies that all further arguments are file names

exec
/bin/zsh -lc "sed -e 's|__KIPI_REPO__|/tmp/repo|g' -e 's|__HOME__|/tmp/home|g' -e 's|__USER__|user|g' q-system/.q-system/scripts/com.kipi.launchd-health.plist | plutil -lint -; sed -e 's|__ROOT__|/tmp/repo|g' automation/com.kipi.voice-refresh.plist | plutil -lint -" in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-326
 succeeded in 0ms:
<stdin>: OK
<stdin>: OK

codex
The new template survives the real macOS `plutil` parser, and the loaded-label guard correctly distinguishes a nonzero `launchctl` result from an empty inventory. One bounded regression did reproduce: a valid checkout path containing `&` can no longer render, even though the deleted installer explicitly supported that path shape. I’m finishing calibration and line-level evidence now.
exec
/bin/zsh -lc "python3 -c 'import runpy,unittest.mock as m,types; d=runpy.run_path(\"q-system/.q-system/scripts/test/test-loaded-label-has-template.py\"); fake=types.SimpleNamespace(returncode=1,stdout=\"PID Status Label\\n- 0 com.kipi.fake\\n\",stderr=\"denied\"); p1=m.patch.object(d[\"shutil\"],\"which\",return_value=\"/bin/launchctl\"); p2=m.patch.object(d[\"subprocess\"],\"run\",return_value=fake); p1.start();p2.start();print(\"loaded_labels_on_rc1=\",d[\"loaded_labels\"]());p2.stop();p1.stop()'; python3 -c 'import runpy; d=runpy.run_path(\"q-system/.q-system/scripts/test/test-single-writer-agents-file.py\"); w,t=d[\"find_writers\"](d[\"REPO\"],d[\"LABEL\"]); print(\"non_test_writers=\",w); print(\"test_shape_files=\",t)' ; python3 -c 'from pathlib import Path; files=[\"automation/test_voice_refresh_schedule.py\",\"q-system/.q-system/scripts/test/test-loaded-label-has-template.py\",\"q-system/.q-system/scripts/test/test-single-writer-agents-file.py\"]; [(compile(Path(f).read_text(),f,\"exec\"),print(f,\"syntax=PASS\")) for f in files]' ; printf '%s\\n' '__ROOT__/automation/voice-refresh-nudge.sh' | sed -e 's|__ROOT__|/tmp/re&po|g' | awk '{print \"rendered=\" "'$0; if (index($0,"__ROOT__")) {print "assert_rendered=FAIL leftover __ROOT__"; exit 1}}'"'; printf 'reproducer_rc=%s\\n' \""'$?"' in /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-326
 succeeded in 84ms:
non_test_writers= ['q-system/.q-system/scripts/install-plist.sh']
test_shape_files= ['q-system/.q-system/scripts/test/test-single-writer-agents-file.py']
automation/test_voice_refresh_schedule.py syntax=PASS
q-system/.q-system/scripts/test/test-loaded-label-has-template.py syntax=PASS
q-system/.q-system/scripts/test/test-single-writer-agents-file.py syntax=PASS
rendered=/tmp/re__ROOT__po/automation/voice-refresh-nudge.sh
assert_rendered=FAIL leftover __ROOT__
reproducer_rc=1

codex
## Finding 1

- **SEVERITY:** minor
- **Claim:** Deleting the XML-safe installer makes `com.kipi.voice-refresh` impossible to install from a valid checkout path containing `&`.
- **Location:** `q-system/.q-system/scripts/install-plist.sh:214`
- **Calibration:** No repository producer currently emits this path shape, so this is minor. The installer exits loudly and leaves the existing plist untouched.
- **Reproducer command:**

```bash
printf '%s\n' '__ROOT__/automation/voice-refresh-nudge.sh' |
  sed -e 's|__ROOT__|/tmp/re&po|g' |
  awk '{
    print "rendered=" $0
    if (index($0,"__ROOT__")) {
      print "assert_rendered=FAIL leftover __ROOT__"
      exit 1
    }
  }'
printf 'reproducer_rc=%s\n' "$?"
```

- **Real output:**

```text
rendered=/tmp/re__ROOT__po/automation/voice-refresh-nudge.sh
assert_rendered=FAIL leftover __ROOT__
reproducer_rc=1
```

The sandbox exposed no writable temporary directory, so the executed reproducer was inline instead of written under `$TMPDIR`. The repository was not modified.

## What is sound

- The new launchd-health template passed the real macOS parser:

```text
<stdin>: OK
```

- A nonzero `launchctl list` result is no longer treated as an empty inventory:

```text
loaded_labels_on_rc1= None
```

- The single-writer detector found exactly one production writer:

```text
non_test_writers= ['q-system/.q-system/scripts/install-plist.sh']
```

- All three changed Python files compiled successfully.
- Duplicate template basenames were checked. None exist.
- The full temp-backed suites could not run because the review sandbox has no writable temporary directory.

Fix first: escape replacement metacharacters before passing repository and home paths to `sed`, or restore an XML-aware renderer.

FINDINGS:
minor|Deleting the XML-safe installer makes com.kipi.voice-refresh impossible to install from a valid checkout path containing an ampersand|q-system/.q-system/scripts/install-plist.sh:214
END FINDINGS
tokens used
98,257
## Finding 1

- **SEVERITY:** minor
- **Claim:** Deleting the XML-safe installer makes `com.kipi.voice-refresh` impossible to install from a valid checkout path containing `&`.
- **Location:** `q-system/.q-system/scripts/install-plist.sh:214`
- **Calibration:** No repository producer currently emits this path shape, so this is minor. The installer exits loudly and leaves the existing plist untouched.
- **Reproducer command:**

```bash
printf '%s\n' '__ROOT__/automation/voice-refresh-nudge.sh' |
  sed -e 's|__ROOT__|/tmp/re&po|g' |
  awk '{
    print "rendered=" $0
    if (index($0,"__ROOT__")) {
      print "assert_rendered=FAIL leftover __ROOT__"
      exit 1
    }
  }'
printf 'reproducer_rc=%s\n' "$?"
```

- **Real output:**

```text
rendered=/tmp/re__ROOT__po/automation/voice-refresh-nudge.sh
assert_rendered=FAIL leftover __ROOT__
reproducer_rc=1
```

The sandbox exposed no writable temporary directory, so the executed reproducer was inline instead of written under `$TMPDIR`. The repository was not modified.

## What is sound

- The new launchd-health template passed the real macOS parser:

```text
<stdin>: OK
```

- A nonzero `launchctl list` result is no longer treated as an empty inventory:

```text
loaded_labels_on_rc1= None
```

- The single-writer detector found exactly one production writer:

```text
non_test_writers= ['q-system/.q-system/scripts/install-plist.sh']
```

- All three changed Python files compiled successfully.
- Duplicate template basenames were checked. None exist.
- The full temp-backed suites could not run because the review sandbox has no writable temporary directory.

Fix first: escape replacement metacharacters before passing repository and home paths to `sed`, or restore an XML-aware renderer.

FINDINGS:
minor|Deleting the XML-safe installer makes com.kipi.voice-refresh impossible to install from a valid checkout path containing an ampersand|q-system/.q-system/scripts/install-plist.sh:214
END FINDINGS

PR #330, opened 2026-09-08
prd-os: combine the rival gate-registry fixes; the closeout registrar was writing legacy gates

assafkip:
## The fork

`plugins/prd-os/scripts/prd_runner.py` had diverged between this repo and the consulting
instance, and **each side carried a guard the other lacked**:

| | had |
|---|---|
| this repo | ASK-988 reopen-aware archive coverage, `SPILLOVER_SEVERITY_ORDER` |
| the instance | ASK-1040 `_reject_unrunnable_gate`, ASK-1038 vacuous-pass WARN |

`plugins/` in an instance is an `rsync --delete` destination from this tree, so the next
fleet sync would have deleted the instance's two guards silently. Neither side is adopted
wholesale; both fixes are kept. Every hunk was applied against an anchor asserted to
appear exactly once.

## The defect this surfaced, which is the real content of the PR

Adopting the instance's `test_gate_lifecycle.py` (a strict superset — zero removals)
turned `test_the_closeout_registrar_asks_for_regression_at_the_call_site` **RED here**.
It was right.

`issue_runner.py` called `gate_register(...)` with no lifecycle, so it took
`LEGACY_GATE_LIFECYCLE`, and `gates run` filters legacy out. **Every gate this repo's
closeout registered went into the permanent registry as non-executable.** That is
ASK-1038's exact shape, live: "all 0 regression gates green" against a populated
registry, true vacuously, exit 0. One line fixes it; the WARN carried in from the
instance is what makes the next occurrence visible instead of silent.

## The guard was present but NOT WIRED, and its own tests could not tell

The four ASK-1040 tests call `_reject_unrunnable_gate` **directly**. Measured: deleting
the single call site from `gate_register` left the file green at **17 passed**. A guard
nobody invokes reads exactly like one that works, and `gates.jsonl` is append-only, so a
bad row is close to permanent.

Added `test_gate_register_ACTUALLY_CALLS_the_unrunnable_guard`: drives the real entry
point with the shape that happened seven times (prose in the command slot), and asserts
it refused **before writing**. Same mutant re-run against it: **1 failed, 17 passed**.
It kills what the other four could not see.

## Verified

- merged file compiles; carries ASK-988 (2), `reopened_at` (2), `SPILLOVER_SEVERITY_ORDER`
  (3), `_reject_unrunnable_gate` (2), ASK-1040 (2), ASK-1038 (1)
- prd-os **692 passed, 1 skipped** (was 690 passed / **1 failed** before the dsse fix)
- kipi-dsse **4 passed**

Spillover `sp-c546f6a4`.

### DIFF
--- a/plugins/kipi-dsse/.claude-plugin/plugin.json
+++ b/plugins/kipi-dsse/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "kipi-dsse",
-  "version": "0.14.5",
+  "version": "0.14.6",
   "description": "DSSE issue workflow: scope-enforced edits, per-finding dispositions, Codex review gates. Provides /issue-start, /issue-approve, /issue-verify, /issue-review, /issue-amend, /issue-closeout. Consumes issue specs created by prd-os or dropped in manually.",
   "author": {
     "name": "Assaf Kipnis"
--- a/plugins/kipi-dsse/.claude-plugin/plugin.json
+++ b/plugins/kipi-dsse/.claude-plugin/plugin.json
@@ -1,10 +1,10 @@
 {
- "name": "kipi-dsse",
- "version": "0.14.4",
- "description": "DSSE issue workflow: scope-enforced edits, per-finding dispositions, Codex review gates. Provides /issue-start, /issue-approve, /issue-verify, /issue-review, /issue-amend, /issue-closeout. Consumes issue specs created by prd-os or dropped in manually.",
- "author": {
-  "name": "Assaf Kipnis"
- },
- "repository": "https://github.com/assafkip/kipi-system",
- "license": "MIT"
+  "name": "kipi-dsse",
+  "version": "0.14.5",
+  "description": "DSSE issue workflow: scope-enforced edits, per-finding dispositions, Codex review gates. Provides /issue-start, /issue-approve, /issue-verify, /issue-review, /issue-amend, /issue-closeout. Consumes issue specs created by prd-os or dropped in manually.",
+  "author": {
+    "name": "Assaf Kipnis"
+  },
+  "repository": "https://github.com/assafkip/kipi-system",
+  "license": "MIT"
 }
--- a/plugins/kipi-dsse/scripts/issue_runner.py
+++ b/plugins/kipi-dsse/scripts/issue_runner.py
@@ -1166,7 +1166,7 @@ def _enforce_spine_contract(paths, fm: dict, marker: dict, issue_id: str) -> str
             cfg = _load_config(paths.repo_root)
             out = _prd_runner.gate_register(
                 cfg, prd_id=marker["prd_id"], issue_id=issue_id,
-                command=bypass_check)
+                command=bypass_check, lifecycle="regression")
         except Exception as exc:
             return (f"cannot close {issue_id}: bypass_check gate registration "
                     f"failed ({exc}) — the permanent registry must record it "
--- a/plugins/kipi-dsse/scripts/issue_runner.py
+++ b/plugins/kipi-dsse/scripts/issue_runner.py
@@ -1164,6 +1164,20 @@ def _enforce_spine_contract(paths, fm: dict, marker: dict, issue_id: str) -> str
             import prd_runner as _prd_runner
             from config import load as _load_config
             cfg = _load_config(paths.repo_root)
+            # LIFECYCLE IS EXPLICIT, NEVER DEFAULTED. Scar 2026-08-24 (ASK-1038):
+            # this call omitted `lifecycle=`, so it took LEGACY_GATE_LIFECYCLE,
+            # which is "historical-receipt" -- the one value `gates run` filters
+            # OUT before executing. The only producer of gates was writing the
+            # one lifecycle the only consumer refuses to run, so `gates run`
+            # printed "all 0 regression gates green" against a 47-gate registry
+            # and had executed nothing since 2026-07-26. 47 closeouts registered
+            # a permanent proof; all 47 were inert from the moment they landed.
+            #
+            # A bypass_check IS a permanent regression proof -- that is what the
+            # registry section header calls it -- so `regression` is the correct
+            # value and it is passed here rather than relied on as a default.
+            # "legacy" in the constant's name reads as "applies to old rows" to
+            # every reviewer, which is exactly why nobody caught it for 29 days.
             out = _prd_runner.gate_register(
                 cfg, prd_id=marker["prd_id"], issue_id=issue_id,
                 command=bypass_check, lifecycle="regression")
--- a/plugins/prd-os/.claude-plugin/plugin.json
+++ b/plugins/prd-os/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "prd-os",
-  "version": "0.30.3",
+  "version": "0.30.4",
   "description": "PRD operating system: capture rough ideas, draft reviewable PRDs, run standard and adversarial review (Codex or a Claude senior-staff-engineer subagent, recorded truthfully), triage findings, decompose approved PRDs into atomic issue specs, and execute those issues with scope enforcement, stop-gate receipts, and reviewer-attributed findings.",
   "author": {
     "name": "Assaf Kipnis"
--- a/plugins/prd-os/.claude-plugin/plugin.json
+++ b/plugins/prd-os/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "prd-os",
-  "version": "0.30.4",
+  "version": "0.30.5",
   "description": "PRD operating system: capture rough ideas, draft reviewable PRDs, run standard and adversarial review (Codex or a Claude senior-staff-engineer subagent, recorded truthfully), triage findings, decompose approved PRDs into atomic issue specs, and execute those issues with scope enforcement, stop-gate receipts, and reviewer-attributed findings.",
   "author": {
     "name": "Assaf Kipnis"
--- a/plugins/prd-os/.claude-plugin/plugin.json
+++ b/plugins/prd-os/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "prd-os",
-  "version": "0.30.5",
+  "version": "0.30.6",
   "description": "PRD operating system: capture rough ideas, draft reviewable PRDs, run standard and adversarial review (Codex or a Claude senior-staff-engineer subagent, recorded truthfully), triage findings, decompose approved PRDs into atomic issue specs, and execute those issues with scope enforcement, stop-gate receipts, and reviewer-attributed findings.",
   "author": {
     "name": "Assaf Kipnis"
--- a/plugins/prd-os/.claude-plugin/plugin.json
+++ b/plugins/prd-os/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "prd-os",
-  "version": "0.30.2",
+  "version": "0.30.3",
   "description": "PRD operating system: capture rough ideas, draft reviewable PRDs, run standard and adversarial review (Codex or a Claude senior-staff-engineer subagent, recorded truthfully), triage findings, decompose approved PRDs into atomic issue specs, and execute those issues with scope enforcement, stop-gate receipts, and reviewer-attributed findings.",
   "author": {
     "name": "Assaf Kipnis"
--- a/plugins/prd-os/.claude-plugin/plugin.json
+++ b/plugins/prd-os/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "prd-os",
-  "version": "0.30.6",
+  "version": "0.30.7",
   "description": "PRD operating system: capture rough ideas, draft reviewable PRDs, run standard and adversarial review (Codex or a Claude senior-staff-engineer subagent, recorded truthfully), triage findings, decompose approved PRDs into atomic issue specs, and execute those issues with scope enforcement, stop-gate receipts, and reviewer-attributed findings.",
   "author": {
     "name": "Assaf Kipnis"
--- a/plugins/prd-os/scripts/prd_runner.py
+++ b/plugins/prd-os/scripts/prd_runner.py
@@ -1187,7 +1187,48 @@ def _reject_unrunnable_gate(command: str) -> None:
     # 2. Is the first word a real command? Catches the prose class, which is the
     #    one that actually happened seven times. Builtins resolve too, so
     #    `cd x && ...` passes; `check:the ...` does not.
-    first = cmd.split()[0]
+    #
+    # THE RAW FIRST TOKEN IS NOT THE COMMAND NAME, and assuming it was made this
+    # guard refuse commands bash runs perfectly (review of PR #330, MAJOR 1,
+    # reproduced against real specs in this repo). Two shapes:
+    #
+    #   PYTHONPATH=src python3 -m pytest ...   -> first token is an ASSIGNMENT
+    #   (cd plugins/prd-os && pytest -q)       -> first token is a SUBSHELL open
+    #
+    # Census over all 204 decoded bypass_checks in .prd-os/issues/: 16 refused, 2
+    # of them this false-positive class (srsa-authoritative-path-contract.md,
+    # srsa-check-repairs-survived-merge.md). That is not a cosmetic miss --
+    # issue_runner RUNS the bypass_check and requires exit 0 BEFORE registering,
+    # so the command is PROVEN runnable and then refused at the door, and
+    # issue_runner turns any exception here into a hard `cannot close`. An
+    # unattended closeout died on a check that had just gone green, with an error
+    # blaming prose.
+    #
+    # Leading assignments are stripped, and a command opening a group or subshell
+    # is accepted on bash's own verdict: step 1 already parsed it with `bash -n`,
+    # which is a stronger statement about `(cd x && y)` than any token probe.
+    probe_cmd = cmd
+    while True:
+        head = probe_cmd.split(None, 1)
+        if not head:
+            break
+        # NAME=value, the POSIX assignment prefix. Deliberately strict about the
+        # NAME half so `check:the=thing` prose cannot masquerade as one.
+        if re.match(r"^[A-Za-z_][A-Za-z0-9_]*=", head[0]):
+            if len(head) == 1:
+                # An assignment and nothing else runs no check at all.
+                raise ValueError(
+                    "gate command is only environment assignments, so it runs "
+                    f"no check.\n  command: {cmd[:160]}")
+            probe_cmd = head[1]
+            continue
+        break
+    stripped = probe_cmd.lstrip()
+    # `(`, `{`, `!` and a leading redirect open shell GRAMMAR, not a command name.
+    # `bash -n` above already accepted the whole line, which is the real check.
+    if stripped[:1] in ("(", "{", "!", "<", ">"):
+        return
+    first = stripped.split()[0]
     probe = _sp.run(["bash", "-lc", f"command -v {shlex.quote(first)} >/dev/null 2>&1"],
                     capture_output=True, text=True)
     if probe.returncode != 0 and not shutil.which(first):
@@ -1215,7 +1256,15 @@ def gate_register(
             f"invalid gate lifecycle {lifecycle!r}; "
             f"expected one of {', '.join(GATE_LIFECYCLES)}"
         )
-    _reject_unrunnable_gate(command)
+    # THE DOOR IS CHECKED AFTER IDEMPOTENCY, NOT BEFORE (review of PR #330, MAJOR 2).
+    # gate_id is issue_id + sha256(command), so a RE-CLOSE of the same issue with the
+    # same bypass_check resolves to a row that already exists and must be a no-op.
+    # Validating first turned that no-op into a raise, and issue_runner converts any
+    # raise here into `cannot close` -- so tightening the door would have broken
+    # re-close, a workflow issue_runner documents ("spec closed once, reopened,
+    # amended, closed again"). A row already in the registry got past whatever door
+    # existed when it was written; re-litigating it at read time is not this guard's
+    # job. New rows are still validated below.
     gate_id = f"{issue_id}-{_hashlib.sha256(command.encode()).hexdigest()[:8]}"
     path = _gates_path(cfg)
     if path.is_file():
@@ -1225,13 +1274,30 @@ def gate_register(
                 if existing.get("gate_id") == gate_id:
                     existing_lifecycle = _gate_lifecycle(existing)
                     if existing_lifecycle != lifecycle:
+                        # NAME THE RECOVERY. Every gate this fleet's closeout wrote
+                        # before the lifecycle fix took LEGACY_GATE_LIFECYCLE, so the
+                        # FIRST re-close of any such issue lands here, and the caller
+                        # (issue_runner) turns it into a bare `cannot close`. The
+                        # recovery exists and lives outside the closeout flow; an
+                        # error that does not name it is a dead end at 3am.
                         raise ValueError(
                             f"gate {gate_id!r} already registered as "
-                            f"{existing_lifecycle!r}, not {lifecycle!r}"
+                            f"{existing_lifecycle!r}, not {lifecycle!r}. "
+                            "The registry is append-only, so this is not edited in "
+                            "place: run `python3 plugins/prd-os/scripts/"
+                            f"migrate_gate_lifecycle.py --regression {gate_id} "
+                            "--apply`, then close again."
                         )
                     return {"gate_id": gate_id, "registered": False}
             except json.JSONDecodeError:
                 continue
+    # THE DOOR, for a genuinely NEW row only. Everything above this line either
+    # returned a no-op or raised on a lifecycle conflict, so an already-permanent
+    # gate never reaches it -- which is the whole point of the ordering (MAJOR 2).
+    # This line is what test_gate_register_ACTUALLY_CALLS_the_unrunnable_guard
+    # defends; deleting it while re-ordering is exactly the mistake that test
+    # caught during the review fix, 2026-09-07.
+    _reject_unrunnable_gate(command)
     record = {"gate_id": gate_id, "prd_id": prd_id, "issue_id": issue_id,
               "command": command,
               "lifecycle": lifecycle,
--- a/plugins/prd-os/scripts/prd_runner.py
+++ b/plugins/prd-os/scripts/prd_runner.py
@@ -1224,9 +1224,23 @@ def _reject_unrunnable_gate(command: str) -> None:
             continue
         break
     stripped = probe_cmd.lstrip()
-    # `(`, `{`, `!` and a leading redirect open shell GRAMMAR, not a command name.
-    # `bash -n` above already accepted the whole line, which is the real check.
-    if stripped[:1] in ("(", "{", "!", "<", ">"):
+    # `!` IS NOT AN ACCEPT, and grouping it with the others was a hole I opened
+    # (review of PR #330 round 3, MINOR). `!` NEGATES the pipeline's status, so
+    # `! check:the ledger is append-only` runs a command that does not exist, gets
+    # 127, and the negation turns that into exit 0 -- a gate that passes forever,
+    # in an append-only registry. It is stripped and probing continues, so the
+    # prose behind it is still caught.
+    while stripped[:1] == "!":
+        stripped = stripped[1:].lstrip()
+        if not stripped:
+            raise ValueError(
+                "gate command is only a negation, so it runs no check.\n"
+                f"  command: {cmd[:160]}")
+    # `(`, `{` and a leading redirect open shell GRAMMAR, not a command name, and
+    # `bash -n` above already accepted the whole line. Prose inside a subshell
+    # still fails LOUDLY at run time (127), which is the tolerable direction; the
+    # negation above was the one that fails silently green.
+    if stripped[:1] in ("(", "{", "<", ">"):
         return
     first = stripped.split()[0]
     probe = _sp.run(["bash", "-lc", f"command -v {shlex.quote(first)} >/dev/null 2>&1"],
@@ -3029,9 +3043,18 @@ def cmd_gates(cfg: Config, args) -> int:
         print(f"[WARN] gates: {registered_total} gate(s) registered, NONE "
               f"executed -- {why}, so no regression check ran. "
               f"Exit 0 below covers the spillover verdict ONLY.")
-    print(f"all {len(records)} regression gates green "
-          f"({registered_total} registered); "
-          f"{len(blocking)} blocking spillover item(s)")
+    # THE FINAL LINE MUST NOT CONTRADICT THE WARN (review of PR #330 round 3,
+    # MINOR). It printed "all 0 regression gates green" immediately under a WARN
+    # saying nothing executed -- which is the exact ASK-1038 sentence, still
+    # there, three lines below its own alarm. A reader who scrolls to the last
+    # line, or a script that greps it, sees the reassurance and not the warning.
+    if executed <= 0:
+        print(f"NO regression gate executed ({registered_total} registered); "
+              f"{len(blocking)} blocking spillover item(s)")
+    else:
+        print(f"all {executed} regression gates green "
+              f"({registered_total} registered); "
+              f"{len(blocking)} blocking spillover item(s)")
     return 0
 
 
--- a/plugins/prd-os/scripts/prd_runner.py
+++ b/plugins/prd-os/scripts/prd_runner.py
@@ -46,6 +46,7 @@ import subprocess
 import tempfile
 import os
 import re
+import shlex
 import sys
 from datetime import datetime, timezone
 from pathlib import Path
@@ -1154,6 +1155,49 @@ def _gate_lifecycle(record: dict) -> str:
     return lifecycle
 
 
+def _reject_unrunnable_gate(command: str) -> None:
+    """Refuse a gate whose command cannot run. Validated AT THE DOOR.
+
+    Scar 2026-08-24 (ASK-1040), measured by executing all 47 registered gates:
+    12 were not runnable at all. SEVEN held PROSE in the command slot -- the
+    shell was being asked to run `check:the`, `check:adding`, `asserting` --
+    because somebody wrote a DESCRIPTION of the check where the check goes.
+    Three were shell syntax errors. Registration accepted every one of them
+    without a word, and each was then recorded as permanent protection.
+
+    A gate that cannot execute is worse than no gate: it is counted, reported,
+    and believed. The registry is append-only, so a bad row is close to
+    permanent -- which is exactly why the check belongs here and not downstream.
+
+    NOT validated here: whether the command TESTS anything real. `true` is a
+    legal gate and a useless one. This refuses the unrunnable, not the useless;
+    the zero-collecting pytest class is its own item.
+    """
+    cmd = (command or "").strip()
+    if not cmd:
+        raise ValueError("gate command is empty; a gate that runs nothing is not a gate")
+    import shutil
+    import subprocess as _sp
+    # 1. Does it parse? Catches the unterminated-heredoc class.
+    syntax = _sp.run(["bash", "-n", "-c", cmd], capture_output=True, text=True)
+    if syntax.returncode != 0:
+        raise ValueError(
+            f"gate command is not valid shell: {syntax.stderr.strip()[:200]}\n"
+            f"  command: {cmd[:160]}")
+    # 2. Is the first word a real command? Catches the prose class, which is the
+    #    one that actually happened seven times. Builtins resolve too, so
+    #    `cd x && ...` passes; `check:the ...` does not.
+    first = cmd.split()[0]
+    probe = _sp.run(["bash", "-lc", f"command -v {shlex.quote(first)} >/dev/null 2>&1"],
+                    capture_output=True, text=True)
+    if probe.returncode != 0 and not shutil.which(first):
+        raise ValueError(
+            f"gate command does not start with a runnable command: {first!r}. "
+            "This is the prose-in-the-command-slot class (ASK-1040): write the "
+            "CHECK here, not a description of it.\n"
+            f"  command: {cmd[:160]}")
+
+
 def gate_register(
     cfg: Config,
     *,
@@ -1171,6 +1215,7 @@ def gate_register(
             f"invalid gate lifecycle {lifecycle!r}; "
             f"expected one of {', '.join(GATE_LIFECYCLES)}"
         )
+    _reject_unrunnable_gate(command)
     gate_id = f"{issue_id}-{_hashlib.sha256(command.encode()).hexdigest()[:8]}"
     path = _gates_path(cfg)
     if path.is_file():
@@ -2752,6 +2797,7 @@ def cmd_gates(cfg: Config, args) -> int:
             "other lifecycles are retained as non-current evidence\n"
         )
         return 2
+    registered_total = len(records)
     records = [
         record for record in records
         if record["lifecycle"] == "regression"
@@ -2890,7 +2936,19 @@ def cmd_gates(cfg: Config, args) -> int:
         for gid, tail in failures:
             sys.stderr.write(f"GATE RED: {gid}\n{tail}\n")
         return 1
-    print(f"all {len(records)} regression gates green; "
+    if not records and registered_total:
+        # AN EMPTY EXECUTED SET IS NOT A PASS, and it must never be printed as
+        # one. Scar 2026-08-24 (ASK-1038): this line said "all 0 regression
+        # gates green" against a 47-gate registry, and the sentence was TRUE --
+        # vacuously, because the closeout registrar wrote every gate as
+        # historical-receipt, the one lifecycle filtered out just above. Exit 0
+        # was read as "the gates are green" for 29 days while nothing ran.
+        # A count of zero is now stated as the anomaly it is.
+        print(f"[WARN] gates: {registered_total} gate(s) registered, NONE "
+              f"executable -- 0 have lifecycle 'regression', so no regression "
+              f"check ran. Exit 0 below covers the spillover verdict ONLY.")
+    print(f"all {len(records)} regression gates green "
+          f"({registered_total} registered); "
           f"{len(blocking)} blocking spillover item(s)")
     return 0
 
--- a/plugins/prd-os/scripts/prd_runner.py
+++ b/plugins/prd-os/scripts/prd_runner.py
@@ -1233,6 +1233,12 @@ def _reject_unrunnable_gate(command: str) -> None:
     while stripped[:1] == "!":
         stripped = stripped[1:].lstrip()
         if not stripped:
+            # NOT DEAD CODE, though it looks it on macOS. `bash -n -c '!'` is a
+            # syntax error under bash 3.2 (so step 1 refuses first) and PARSES
+            # under the bash 5.x that CI runs, where this is the only thing
+            # standing between a bare `!` and `stripped.split()[0]` on an empty
+            # string. CI proved it by failing on a test that had pinned the macOS
+            # door. Platform-split behaviour, kept deliberately.
             raise ValueError(
                 "gate command is only a negation, so it runs no check.\n"
                 f"  command: {cmd[:160]}")
--- a/plugins/prd-os/scripts/prd_runner.py
+++ b/plugins/prd-os/scripts/prd_runner.py
@@ -1284,8 +1284,14 @@ def gate_register(
                             f"gate {gate_id!r} already registered as "
                             f"{existing_lifecycle!r}, not {lifecycle!r}. "
                             "The registry is append-only, so this is not edited in "
+                            # --registry is required=True in that script, so the
+                            # first version of this message printed a command that
+                            # exits 2 (review of PR #330 round 2, MINOR 1). A
+                            # recovery pointer that does not run is the dead end it
+                            # was written to remove.
                             "place: run `python3 plugins/prd-os/scripts/"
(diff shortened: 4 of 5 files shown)


assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-330-20260907-211400.md` (7479 bytes).

```
FINDINGS:
major|_reject_unrunnable_gate refuses a valid `VAR=value cmd` bypass_check that already ran green, hard-blocking issue closeout for 2 live specs|plugins/prd-os/scripts/prd_runner.py:1190
major|A gate registered under the old historical-receipt default can never be re-registered as regression, so a re-close is permanently blocked with no in-flow recovery|plugins/kipi-dsse/scripts/issue_runner.py:1169
minor|The closeout call-site test inspects only the first gate_register( match, so a second unlifecycled registrar stays green|plugins/prd-os/tests/test_gate_lifecycle.py:297
END FINDINGS
```

--- reviewer output ---

# Review: PR #330

The core diagnosis is right and the wiring test is genuinely good. Two reproduced defects in the guard being carried upstream.

---

## MAJOR 1 — the new door refuses a valid gate command, and that blocks issue closeout

`plugins/prd-os/scripts/prd_runner.py:1190`

`first = cmd.split()[0]` takes the raw first whitespace token. A leading environment assignment is a normal, runnable shell command and its first token is not a command name. Two live issue specs in this repo use exactly that shape.

The chain matters: `issue_runner.py:1145` RUNS the bypass_check first and requires exit 0. Only then does it register. So the command is proven runnable, and is then refused at the door as "prose in the command slot", and `issue_runner.py:1171` converts any exception into a hard `cannot close`.

Reproducer, against the real spec `.prd-os/issues/srsa-authoritative-path-contract.md`:

```
REAL bypass_check from srsa-authoritative-path-contract.md
    PYTHONPATH=plugins/kipi-core/kipi-mcp/src python3 -m pytest -q plugins/kipi-core/kipi-mcp/tests/test_paths.py -k 'fails_closed or resolves_instance_domain_dir'
STEP 1  issue_runner runs it (shell=True):  rc = 0 -> PASSES the run gate
        tail: ['3 passed, 18 deselected, 2 warnings in 0.02s']
STEP 2  issue_runner registers it (issue_runner.py:1167):
        RAISES -> gate command does not start with a runnable command: 'PYTHONPATH=plugins/kipi-core/kipi-mcp/src'. ...
STEP 3  issue_runner.py:1171 returns:
        cannot close srsa-authoritative-path-contract: bypass_check gate registration failed (...)
```

Census over all 204 decoded bypass_checks in `.prd-os/issues/`: 16 refused, 2 of them this false-positive class (`srsa-authoritative-path-contract.md`, `srsa-check-repairs-survived-merge.md`). The subshell form is refused too:

```
bash-parses=True  REFUSED: ... 'PYTHONPATH=.'
    cmd: PYTHONPATH=. python3 -m pytest plugins/prd-os/tests -q
bash-parses=True  REFUSED: ... '(cd'
    cmd: (cd plugins/prd-os && python3 -m pytest -q)
bash-parses=True  ACCEPTED
    cmd: cd plugins/prd-os && python3 -m pytest -q
```

The control test `test_real_gate_commands_are_not_false_positived` carries 3 cases, none of them this shape, and its docstring cites a measurement taken against a different repo's registry ("all 34 regression gates pass this, 0 false positives"). That measurement does not hold here.

At 3am: an unattended `/issue-closeout` dies on an issue whose check just went green, with an error message that misdiagnoses the cause. Recoverable by rewriting the spec, so major not blocker.

---

## MAJOR 2 — a gate already registered under the old default can never be re-registered, and closeout has no recovery path

`plugins/kipi-dsse/scripts/issue_runner.py:1169` and `plugins/prd-os/scripts/prd_runner.py:1229`

`gate_id` is `issue_id + sha256(command)[:8]`, so a re-close of the same issue with the same bypass_check produces the same id. Every gate this repo's closeout wrote before this PR took `LEGACY_GATE_LIFECYCLE`. `gate_register` refuses a lifecycle change on an existing id. `issue_runner.py:1171` turns that into `cannot close`. The registry is append-only.

Reproducer (step 1 is the unmodified pre-PR default, not an invented row):

```
pre-PR closeout wrote: {'gate_id': 'ASK-9001-c8f2ce84', 'registered': True} lifecycle= historical-receipt
second close RAISES -> gate 'ASK-9001-c8f2ce84' already registered as 'historical-receipt', not 'regression'
issue_runner.py:1171 turns that into:
  cannot close ASK-9001: bypass_check gate registration failed (gate 'ASK-9001-c8f2ce84' already registered as 'historical-receipt', not 'regression') - the permanent registry must record it before close.
```

`issue_runner.py:1202` documents re-close as a real workflow ("spec closed once, reopened, amended, closed again"). Recovery exists but is outside the closeout flow: `migrate_gate_lifecycle.py --regression <gate_id> --apply`. Nothing in this PR mentions it, and no error message points at it.

Honesty note: this worktree has no `.prd-os/gates.jsonl`, so I produced the pre-existing row with the shipped default rather than observing one. The fleet instance this PR is reconciling against carries 47 such rows per the PR body.

Related, same mechanism: `_reject_unrunnable_gate(command)` at `prd_runner.py:1218` sits **before** the idempotency lookup, so re-registering an already-permanent gate now raises instead of returning a no-op.

---

## MINOR 3 — the call-site test only inspects the first `gate_register(` occurrence

`plugins/prd-os/tests/test_gate_lifecycle.py:297-298`

`src.index(...)` plus a 260-char slice. A second registrar added later with no lifecycle keeps the test green.

```
second unlifecycled call site present: _prd_runner.gate_register(cfg, prd_id="p", issue_id="i", command="true")
test assertion result: True (True = test stays GREEN)
call slice it inspected starts at index 48820 of 59558
```

There is exactly one call site today, so this is a latent hole, not a live one. An exact-set assertion over every `gate_register(` match closes it.

---

## What is sound

- **The new wiring test is real.** I deleted `_reject_unrunnable_gate(command)` from a copy of `prd_runner.py` and drove `gate_register` through it: `MUTANT accepted the prose gate -> {'gate_id': 'issue-wiring-9fc0efdc', 'registered': True}`. The `pytest.raises` would fail. It kills the mutant the other four ASK-1040 tests cannot see.
- **`test_a_failing_regression_gate_actually_turns_the_run_red` can fail.** It asserts a non-zero rc AND that the gate id appears in the output, so it does not pass on an unrelated rc=2 config error.
- **Suite claims verified, not taken on trust.** `python3 -m pytest plugins/prd-os/tests -q` → `692 passed, 1 skipped in 67.62s`. Changed files together → `128 passed`.
- **Command injection at the door.** `shlex.quote(first)` into `bash -lc "command -v ..."` holds; `bash -n -c` parses without executing. I tried and could not get the guard to execute the gate body.
- **Cost.** `bash -lc` per registration measured at 7 ms. Not a 3am concern. Attack dropped.
- **A login-profile that returns non-zero would false-refuse builtins** (the `shutil.which` fallback covers PATH commands but not `cd`). I could not construct this without editing the user's home, so I am dropping it rather than reporting it.
- **`gates run` still exits 0 on a fully-legacy registry**, but the added WARN and the `(N registered)` count make it visible, which is the stated design choice. Not filed.

---

## VERDICT: REQUEST CHANGES

Fix first: **MAJOR 1.** Resolve the first *command word* rather than the first whitespace token — skip leading `VAR=value` assignments and refuse only what a shell would refuse. Two issue specs in this repo currently cannot be closed once this merges, and the error text tells the operator the wrong thing.

FINDINGS:
major|_reject_unrunnable_gate refuses a valid `VAR=value cmd` bypass_check that already ran green, hard-blocking issue closeout for 2 live specs|plugins/prd-os/scripts/prd_runner.py:1190
major|A gate registered under the old historical-receipt default can never be re-registered as regression, so a re-close is permanently blocked with no in-flow recovery|plugins/kipi-dsse/scripts/issue_runner.py:1169
minor|The closeout call-site test inspects only the first gate_register( match, so a second unlifecycled registrar stays green|plugins/prd-os/tests/test_gate_lifecycle.py:297
END FINDINGS

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-330-20260907-213353.md` (5261 bytes).

```
FINDINGS:
minor|The lifecycle-conflict error names a recovery command that exits 2 because it omits the required --registry flag|plugins/prd-os/scripts/prd_runner.py:1288
minor|A run whose only regression gates are skipped as self-referential still prints "all N regression gates green" and exits 0 with nothing executed|plugins/prd-os/scripts/prd_runner.py:3005
minor|The every-call-site lifecycle test uses a 260-char forward window, so an adjacent lifecycled registrar masks an unlifecycled one|plugins/prd-os/tests/test_gate_lifecycle.py:388
END FINDINGS
```

--- reviewer output ---

## Review: PR #330

Three findings survived reproduction. All minor. Two attacks I ran did not stick, listed below.

---

### MINOR 1: the recovery command the new error prints does not run

`plugins/prd-os/scripts/prd_runner.py:1288`

The lifecycle-conflict message now names a recovery, and the comment above it says an error that does not name one is a dead end at 3am. The named command is missing `--registry`, which `migrate_gate_lifecycle.py` marks `required=True` at line 73.

Repro, first the real error text, then the command it prints, verbatim:

```
=== ERROR TEXT THE OPERATOR SEES ===
gate 'issue-conf-b5bea41b' already registered as 'historical-receipt', not
'regression'. The registry is append-only, so this is not edited in place: run
`python3 plugins/prd-os/scripts/migrate_gate_lifecycle.py --regression
issue-conf-b5bea41b --apply`, then close again.
```

```
$ python3 plugins/prd-os/scripts/migrate_gate_lifecycle.py --regression issue-conf-b5bea41b --apply
usage: migrate_gate_lifecycle.py [-h] --registry REGISTRY
                                 [--regression REGRESSION]
                                 [--external EXTERNAL] [--retired RETIRED]
                                 [--apply]
migrate_gate_lifecycle.py: error: the following arguments are required: --registry
EXIT=2
```

With `--registry .prd-os/gates.jsonl` added it works, exit 0, `"regression": 1`. So the fix is one flag in the f-string.

`test_a_lifecycle_conflict_names_its_recovery` at line 297 matches only the substring `migrate_gate_lifecycle.py`, so it is green against a command that cannot run. Held to minor because argparse prints the missing flag right there, so the operator is annoyed, not stranded.

---

### MINOR 2: the vacuous-pass warning misses the self-referential-skip path

`plugins/prd-os/scripts/prd_runner.py:3005`

The new WARN keys on `len(records) == 0`. A registry whose regression gates are all skipped as self-referential has `records` non-empty and zero commands executed, so it takes the old path.

Repro, one regression gate whose command contains `gates run`:

```
RC = 0
[skip] selfref: self-referential `gates run` gate (not executed)
[census] spillover: 0 open total
all 1 regression gates green (1 registered); 0 blocking spillover item(s)
```

Same sentence, same exit 0, nothing ran, no WARN. `skipped_self_ref` is incremented at line 2923 and never read, so the counter this needs is already sitting there dead.

Dropped a level on purpose: I censused all 204 `bypass_check:` lines in `.prd-os/issues/` and zero are self-referential today, so no producer currently emits this shape.

---

### MINOR 3: "EVERY call site" still has the window hole it was fixed for

`plugins/prd-os/tests/test_gate_lifecycle.py:388`

The fix replaced `src.index(...)` with all matches, but kept a 260-character window sliced forward from each match. Two adjacent registrars fall inside each other's window, so a lifecycled second call donates its `lifecycle="regression"` to an unlifecycled first.

Repro against a two-registrar source where the first has no lifecycle:

```
call sites: 2 | flagged: 0
first call site HAS no lifecycle, yet test verdict = GREEN
```

The test's whole purpose is a future second registrar, and adjacency is the likeliest arrangement for one. Parsing the call's argument list with `ast` removes the window entirely.

---

### What is sound

- **The false-positive fix on the door holds against the real corpus.** I re-ran `_reject_unrunnable_gate` over all 204 decoded `bypass_check` lines independently. 14 refused, and I read every one: all 14 are genuine prose in the command slot. Zero false positives on `PYTHONPATH=...` and subshell shapes.
- **The wiring test earns its name.** `gate_register` calls the guard, refuses before the append, and I confirmed `.prd-os/gates.jsonl` stays absent after the refusal.
- **Door latency is not a problem.** I suspected `bash -lc` sourcing a login profile per registration would be slow or hang with no timeout. Measured: 6ms per check. No finding.
- **The lifecycle flip does not arm anything destructive.** Flipping closeout gates to `regression` means they now execute on every `gates run`, so I scanned all 204 for side-effect-shaped tokens. 4 hits, all false positives on the word "install" inside pytest paths. Nothing mutating.
- **Suites green here:** prd-os 700 passed / 1 skipped, kipi-dsse 4 passed, `test_gate_lifecycle.py` 26 passed.

---

Fix first: add `--registry .prd-os/gates.jsonl` to the recovery command at `prd_runner.py:1288`, and widen the test at line 297 past the filename substring. It is the one finding that reaches a human who is already blocked.

FINDINGS:
minor|The lifecycle-conflict error names a recovery command that exits 2 because it omits the required --registry flag|plugins/prd-os/scripts/prd_runner.py:1288
minor|A run whose only regression gates are skipped as self-referential still prints "all N regression gates green" and exits 0 with nothing executed|plugins/prd-os/scripts/prd_runner.py:3005
minor|The every-call-site lifecycle test uses a 260-char forward window, so an adjacent lifecycled registrar masks an unlifecycled one|plugins/prd-os/tests/test_gate_lifecycle.py:388
END FINDINGS

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-330-20260907-215116.md` (9830 bytes).

```
FINDINGS:
minor|`gates run` prints "all 0 regression gates green" as its final line right after the WARN says nothing executed, and exits 0 so issue-verify writes a passing receipt|plugins/prd-os/scripts/prd_runner.py:3032
minor|The `!` early-return in the gate door accepts prose and yields a gate that exits 0 forever in an append-only registry|plugins/prd-os/scripts/prd_runner.py:1229
minor|The door probes command runnability in the process cwd while issue_runner and gates run both use repo_root, so a relative command is accepted or refused by invocation directory|plugins/prd-os/scripts/prd_runner.py:1232
minor|A quoted or comment-prefixed first token is refused as prose though bash runs it, the same false-refuse class the PR fixed for assignments|plugins/prd-os/scripts/prd_runner.py:1231
minor|The lifecycle-conflict recovery message names only --regression, and no file in the repo documents --retired, so a legitimately stale regression gate blocks every closeout with no named way out|plugins/prd-os/scripts/prd_runner.py:1292
nit|plugin.json bumped 0.30.2 to 0.30.5 with no CHANGELOG entry; CHANGELOG tops out at 0.22.0 while claiming to record all notable changes|plugins/prd-os/.claude-plugin/plugin.json:3
nit|The ast call-site test docstring claims EVERY call site but parses only issue_runner.py|plugins/prd-os/tests/test_gate_lifecycle.py:415
END FINDINGS
```

--- reviewer output ---

# Review: PR #330 — prd-os gate registry

Read-only. Nothing in the repo was modified. Every finding below has an executed reproducer with real output.

---

I tried to break this five ways and failed each time. Naming them, because the faults below only mean something if the survivals are stated too.

**Mutation: is the new test suite load-bearing?** Copied both plugins to a temp tree and ran five mutants. All five killed.

```
M0 baseline (copy, unmutated) -> 27 passed
M1 delete door call site      -> FAILED test_gate_register_ACTUALLY_CALLS_the_unrunnable_guard | 1 failed, 26 passed
M2 closeout drops lifecycle   -> FAILED test_the_closeout_registrar_asks_for_regression_at_the_call_site | 1 failed, 26 passed
M3 WARN ignores skipped_self_ref -> FAILED test_a_registry_whose_gates_are_ALL_SKIPPED_is_not_reported_as_green | 1 failed
M4 revert assignment stripping -> 4 failed, 23 passed
M5 door always accepts        -> 7 failed, 20 passed
M6 restored                   -> 27 passed
```

**Census: does the new door false-refuse real specs?** Decoded all 204 `bypass_check` values in `.prd-os/issues/` through `issue_runner._decode_bypass_check` and ran each through `_reject_unrunnable_gate`. 14 refused, and I read all 14: every one is genuine prose ("the lint never executes a named enforcer: grep ..."). Zero false positives. The MAJOR-1 fix the PR describes is real.

**Shell grammar:** `if`, `for`, `while`, `case`, `until`, `time`, `[[` all pass the door and all run under sh. `command -v` resolves keywords, so the keyword class is covered by accident but it is covered.

**The recovery command actually runs.** Registered a gate as `historical-receipt`, re-registered as `regression`, extracted the backticked command from the error verbatim, executed it, re-registered:

```
rc= 0
{"registry": ".../gates.jsonl", "records": 1,
 "counts": {"regression": 1, "historical-receipt": 0, ...}, "applied": true}
RE-CLOSE: {'gate_id': 'iss-legacy-b5bea41b', 'registered': False}
```

Round 2's MINOR-1 fix holds end to end.

**Suites:** `plugins/prd-os/tests` 701 passed 1 skipped. `plugins/kipi-dsse` 110 passed.

**Dropped findings I could not reproduce:** self-referential-regex false positives (0 of 204 match), state-mutating bypass_checks that would now execute on every `gates run` (0 of 204 match `--apply`/`git commit`/`rm`/redirect), relative-path first tokens in live specs (0 of 204).

---

## Findings

### 1. minor — `gates run` prints "all 0 regression gates green" as its LAST line, two lines after warning that nothing ran, and exits 0

`plugins/prd-os/scripts/prd_runner.py:3032`

Reproduced against the exact ASK-1038 shape (47 gates, all `historical-receipt`):

```
EXIT CODE: 0
[census] spillover: 0 open total
[WARN] gates: 47 gate(s) registered, NONE executed -- 0 have lifecycle 'regression', so no regression check ran. Exit 0 below covers the spillover verdict ONLY.
all 0 regression gates green (47 registered); 0 blocking spillover item(s)
```

The last line is the sentence the PR body names as the defect, still printed, still last, still the one a skimming reader takes as the verdict. The machine signal is unchanged too: `gates run` is a `required_check` in live specs (`.prd-os/issues/fable-merge-into-prd-os.md:51`), and `issue_runner.cmd_verify` (line 731, 748) writes a permanent `verified` receipt whenever every check returns rc 0. So the WARN is invisible to the consumer that actually runs this unattended.

Not raised higher: the exit code is unchanged from main, the PR strictly improves the prior silence, and flipping the WARN branch to exit 2 has fleet-wide blast radius the author reasoned about in-code. The cheap part is the print: suppress or reword the green line when `executed <= 0`, and send the WARN to stderr where `GATE RED` already goes.

### 2. minor — the `!` early-return accepts prose and produces a permanently-green gate

`plugins/prd-os/scripts/prd_runner.py:1229`

```
=== B. the `!` early-return: prose accepted AND permanently green ===
  door: ACCEPT
  shell rc = 0 (0 == this gate is green forever)
  stderr: /bin/sh: check:the: command not found
```

`! check:the ledger is append-only` skips all validation at line 1229, exits 0 under the shell because `!` inverts the 127, passes `issue_runner`'s pre-run, and lands in an append-only registry as a regression gate that can never go red. That is the ASK-1040 class inverted into a false green, in the one prefix that flips the exit code. `(`, `{`, `<`, `>` are fine by comparison: those produce a RED gate, which is honest.

Dropped a level, stated per the rule: I constructed this input. 0 of the 204 `bypass_check` values in this repo start with `!`.

### 3. minor — the door validates relative commands against the process cwd; `issue_runner` and `gates run` both use `repo_root`

`plugins/prd-os/scripts/prd_runner.py:1232`

```
=== A. cwd mismatch ===
  from repo root:  ACCEPT  ./plugins/prd-os/scripts/export-fable-mirror.sh --check
  from plugins/:   REFUSE  gate command does not start with a runnable command: './plugins/prd-os/...
```

`bash -lc "command -v ..."` inherits the Python process cwd. `issue_runner` runs the bypass_check with `cwd=paths.repo_root` and `cmd_gates` runs it with `cwd=cfg.repo_root`. So the same spec is accepted or refused depending on which directory the operator invoked closeout from, and the refusal becomes `cannot close`. Fix is to pass `cwd=cfg.repo_root` to both subprocess calls in the door.

Dropped a level: 0 of 204 live bypass_checks start with a relative path.

### 4. minor — a quoted first token is refused though bash runs it

`plugins/prd-os/scripts/prd_runner.py:1231`

```
  shell rc=0  door=REFUSE: gate command does not start with a runnable command: "'python3'"
  shell rc=0  door=REFUSE: gate command does not start with a runnable command: '"python3"'
```

Same false-refuse class as the MAJOR-1 this PR fixed: `stripped.split()[0]` is a raw token, not a command name. A leading `#` comment line in a multi-line command fails the same way. Dropped a level: no live spec emits it.

### 5. minor — the lifecycle-conflict message names only `--regression`, and no retirement path is documented anywhere

`plugins/prd-os/scripts/prd_runner.py:1292`

`grep -rn "migrate_gate_lifecycle" --include="*.md" plugins/ .claude/` returns nothing. The only reference to the recovery script in the whole tree is the error string added by this PR, and it names one flag.

This matters more after this PR than before it, because `lifecycle="regression"` at the closeout call site is what makes the executed gate set actually grow. Once a registered regression gate goes legitimately stale (test renamed, file moved, behavior intentionally changed), `gates run` is rc 1 forever, which blocks `issue-verify` for every subsequent issue listing it as a required_check and blocks `prd-archive`. `wiring-check.md` says "gates only grow" and `no-orphan-findings.md` says "you cannot hand-clear the gate", so at 3am the agent has no named way out. `--retired` exists and is unmentioned. One sentence in the same error string fixes it.

### 6. nit — three version bumps, no CHANGELOG entry

`plugins/prd-os/.claude-plugin/plugin.json:3` — 0.30.2 → 0.30.5. `plugins/prd-os/CHANGELOG.md` tops out at `[0.22.0]` and opens "All notable changes to the `prd-os` plugin are recorded here." Pre-existing drift (0.30.2 was already undocumented); this PR adds three more.

### 7. nit — the ast test says "EVERY call site" but parses one file

`plugins/prd-os/tests/test_gate_lifecycle.py:415` reads only `plugins/kipi-dsse/scripts/issue_runner.py`. A second registrar in any other file is invisible to it. Correct today (one caller, verified by grep), and the docstring overclaims what the subject is.

---

## VERDICT: APPROVE WITH NITS

Only minor and nit findings survived reproduction. The rule gives APPROVE WITH NITS, and the pipeline captures every minor as a tracked follow-up, so none of these are lost by not blocking.

**Fix first:** finding 1. The last line `gates run` prints is the exact sentence this PR exists to stop being believed, and it prints after the warning that contradicts it. Suppressing that line when `executed <= 0` is a one-line change and it removes the misleading string from every log the fleet keeps. The exit-code question underneath it is real and deserves its own issue, not this one.

FINDINGS:
minor|`gates run` prints "all 0 regression gates green" as its final line right after the WARN says nothing executed, and exits 0 so issue-verify writes a passing receipt|plugins/prd-os/scripts/prd_runner.py:3032
minor|The `!` early-return in the gate door accepts prose and yields a gate that exits 0 forever in an append-only registry|plugins/prd-os/scripts/prd_runner.py:1229
minor|The door probes command runnability in the process cwd while issue_runner and gates run both use repo_root, so a relative command is accepted or refused by invocation directory|plugins/prd-os/scripts/prd_runner.py:1232
minor|A quoted or comment-prefixed first token is refused as prose though bash runs it, the same false-refuse class the PR fixed for assignments|plugins/prd-os/scripts/prd_runner.py:1231
minor|The lifecycle-conflict recovery message names only --regression, and no file in the repo documents --retired, so a legitimately stale regression gate blocks every closeout with no named way out|plugins/prd-os/scripts/prd_runner.py:1292
nit|plugin.json bumped 0.30.2 to 0.30.5 with no CHANGELOG entry; CHANGELOG tops out at 0.22.0 while claiming to record all notable changes|plugins/prd-os/.claude-plugin/plugin.json:3
nit|The ast call-site test docstring claims EVERY call site but parses only issue_runner.py|plugins/prd-os/tests/test_gate_lifecycle.py:415
END FINDINGS

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-330-20260907-220255.md` (7015 bytes).

```
FINDINGS:
minor|The unrunnable-gate door strips NAME=value prefixes then probes PATH without applying them, refusing commands the shell runs at exit 0|plugins/prd-os/scripts/prd_runner.py:1217
minor|On its only production caller the door can only produce false positives, because issue_runner already runs the bypass_check and requires exit 0 before registering|plugins/prd-os/scripts/prd_runner.py:1306
minor|The NONE-executed WARN is immediately followed by "all 0 regression gates green" as the final line, at exit 0|plugins/prd-os/scripts/prd_runner.py:3032
minor|The WARN fires identically on a registry whose gates are deliberately retired or external, with no acknowledgement path|plugins/prd-os/scripts/prd_runner.py:3019
nit|A prd-os test reads a kipi-dsse source file with no guard and errors with a bare FileNotFoundError when the sibling plugin is absent|plugins/prd-os/tests/test_gate_lifecycle.py:429
END FINDINGS
```

--- reviewer output ---

# Review: PR #330 (prd-os gate registry)

The central fix is right and the tests behind it are load-bearing. Four bounded issues, none of them gating.

## Findings

### minor 1. The door strips `NAME=value` prefixes and then probes PATH without applying them

`plugins/prd-os/scripts/prd_runner.py:1217`

`_reject_unrunnable_gate` walks off leading assignments, then runs `command -v` on the next token in the *unmodified* environment. A command that sets PATH and then uses it is refused as prose. `issue_runner` converts that raise into a hard `cannot close`.

Reproducer and real output:

```
shell rc= 0 | door: REFUSE: gate command does not start with a runnable command: 'mycheck'. | cmd: PATH=/tmp/pr330fp/bin mycheck
shell rc= 0 | door: REFUSE: gate command does not start with a runnable command: 'mycheck'. | cmd: PATH=/tmp/pr330fp/bin:/Users/assafkipnis/... mycheck
```

Same root cause hits a repo-relative first token. `gates run` executes with `cwd=cfg.repo_root`; the door resolves against the registrar's process cwd. `./scripts/check.sh` is refused when those differ.

Census says 0 of 204 live `bypass_check` values use either shape, so severity is dropped a level and the 3am action today is none.

### minor 2. On its only production caller the door can refuse but can never catch

`plugins/prd-os/scripts/prd_runner.py:1306`

`gate_register` has exactly one non-test caller (`issue_runner.py:1181`). There is no `gates register` CLI (`gates_cmd` choices are `list`, `run`). That caller runs the `bypass_check` with `shell=True, cwd=repo_root` and returns `cannot close` on any nonzero exit *before* registering. Every shape the door refuses already dies there:

```
PART 1: does any door-refused command survive issue_runner's pre-run?
  runner_rc=127  door=REFUSE  check:the brake never leaks into the test suite
  runner_rc=127  door=REFUSE  check:adding a demo sentence fails the approval test
  runner_rc=127  door=REFUSE  asserting the X body stays under the ceiling
  runner_rc=  2  door=REFUSE  if true; then
  runner_rc=127  door=REFUSE  no ADVISORY entry can pass under a live marker without
```

By the time the door sees a command, that command has exited 0. So on this path the guard's true-positive rate is structurally zero and its false-positive rate is not. Worth either applying the assignments in the probe, or moving the door to registration paths that do not pre-run.

### minor 3. The WARN is followed by the exact sentence it exists to refute, and that sentence is last

`plugins/prd-os/scripts/prd_runner.py:3032`

Real output against a 2-row all-legacy registry:

```
[census] spillover: 0 open total
[WARN] gates: 2 gate(s) registered, NONE executed -- 0 have lifecycle 'regression', so no regression check ran. Exit 0 below covers the spillover verdict ONLY.
all 0 regression gates green (2 registered); 0 blocking spillover item(s)

EXIT 0
```

`test_a_registry_with_nothing_executable_is_not_reported_as_green` asserts `[WARN]` is present. It never asserts the green line is absent, and the green line is what a skimming reader and any `tail -1` sees. Suppressing that line when `executed <= 0` costs one branch.

### minor 4. The WARN cannot tell a deliberate retirement from the ASK-1038 outage

`plugins/prd-os/scripts/prd_runner.py:3019`

`retired` and `external` exist so a gate can be legitimately non-current, and the same code path says so. A registry holding only those fires the identical alarm, on every run, with no acknowledgement path:

```
[census] spillover: 0 open total
[WARN] gates: 2 gate(s) registered, NONE executed -- 0 have lifecycle 'regression', so no regression check ran. Exit 0 below covers the spillover verdict ONLY.
all 0 regression gates green (2 registered); 0 blocking spillover item(s)
```

(registry: one `retired`, one `external`). No repo in this fleet is in that state today, so severity dropped a level. Keying the WARN on `historical-receipt` count rather than "not regression" separates the two.

### nit 5. A prd-os test reads a kipi-dsse file with no guard

`plugins/prd-os/tests/test_gate_lifecycle.py:429`

`PLUGIN_ROOT.parent / "kipi-dsse" / "scripts" / "issue_runner.py"` errors with a bare `FileNotFoundError` if prd-os is present without its sibling. Observed while running the suite from a copied tree. The test already guards the empty-call-site case with a real message; the missing-file case deserves the same or a skip.

## What is sound

Attacks that failed, each executed:

- The lifecycle-conflict recovery command. Registered a legacy row, took the raise, extracted the backticked command, ran it: rc 0, row flipped to `regression`, re-close returned `{'gate_id': ..., 'registered': False}`. The pointer runs.
- Mutation, guard call site deleted from `gate_register`: `1 failed, 25 passed`, killed by `test_gate_register_ACTUALLY_CALLS_the_unrunnable_guard`. The PR's headline claim holds.
- Mutation, `lifecycle="regression"` removed from `issue_runner.py:1181`: killed by the ast test, with the line number in the message. The window-free rewrite works.
- Mutation, `executed = len(records)` instead of `len(records) - skipped_self_ref`: killed by the all-skipped test.
- Census of all 204 live `bypass_check` values against the door: 14 refused, every one genuine prose, 0 false positives. The MAJOR-1 assignment and subshell fix holds against real producer data.
- Shell grammar probe: `for`, `while`, `if`, `case`, `time`, `[[`, `cd x && y`, `{ true; }`, trailing comments all pass the door.
- Full suites: prd-os `701 passed, 1 skipped`; kipi-dsse `110 passed`.

Not filed: `gate_register` does read-then-append with no lock, but `gate_id` is `issue_id` plus a hash of the command, so a duplicate row needs the same issue closed twice concurrently. I did not construct that, so I am not claiming it.

## VERDICT

**APPROVE WITH NITS.**

First thing to fix: `prd_runner.py:1217`. Apply the stripped assignments when probing, or skip step 2 entirely for a command the caller already proved runs, because a false refusal there is a hard `cannot close` on an unattended run.

FINDINGS:
minor|The unrunnable-gate door strips NAME=value prefixes then probes PATH without applying them, refusing commands the shell runs at exit 0|plugins/prd-os/scripts/prd_runner.py:1217
minor|On its only production caller the door can only produce false positives, because issue_runner already runs the bypass_check and requires exit 0 before registering|plugins/prd-os/scripts/prd_runner.py:1306
minor|The NONE-executed WARN is immediately followed by "all 0 regression gates green" as the final line, at exit 0|plugins/prd-os/scripts/prd_runner.py:3032
minor|The WARN fires identically on a registry whose gates are deliberately retired or external, with no acknowledgement path|plugins/prd-os/scripts/prd_runner.py:3019
nit|A prd-os test reads a kipi-dsse source file with no guard and errors with a bare FileNotFoundError when the sibling plugin is absent|plugins/prd-os/tests/test_gate_lifecycle.py:429
END FINDINGS

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-330-20260907-220755.md` (6181 bytes).

```
FINDINGS:
minor|A leading `!` in front of `(` or `{` reaches the grouping accept, so `! ( prose )` registers a permanently green gate|plugins/prd-os/scripts/prd_runner.py:1243
minor|The first-token probe still takes the raw split token, so a quoted or expanded command name is refused and issue_runner turns that into a hard `cannot close`|plugins/prd-os/scripts/prd_runner.py:1245
minor|`gates run` still exits 0 when the entire registry is inert, so a consumer reading the exit code sees the same green ASK-1038 produced|plugins/prd-os/scripts/prd_runner.py:3045
END FINDINGS
```

--- reviewer output ---

# Review: PR #330 (`parity/prd-runner-merge`)

Read-only. Nothing in the repo was modified. All reproducers ran against the checked-out head `2ca452ff`; scratch copies live under `/var/folders/.../pr330-*`.

## Finding 1 — minor: `!` in front of a group re-opens the hole the `!` fix just closed

Round 3 stripped a leading `!` so prose behind it still gets probed. It strips `!` and then hands off to the grouping accept at `plugins/prd-os/scripts/prd_runner.py:1243`, so `! ( prose )` short-circuits: prose runs, exits 127, `!` inverts it to 0, and the row is permanent.

Reproducer, through the real entry point:

```
gate_register(prose behind a negated subshell) -> {'gate_id': 'issue-neg-b45996f5', 'registered': True}
registry row written: {"gate_id": "issue-neg-b45996f5", ... "command": "! ( check:the ledger is append-only )", "lifecycle": "regression", ...
--- gates run ---
[green] issue-neg-b45996f5: ! ( check:the ledger is append-only )
[census] spillover: 0 open total
all 1 regression gates green (1 registered); 0 blocking spillover item(s)
exit: 0
```

Helper level, both grouping forms:

```
ACCEPTED at the door         | runtime rc=0 | ! ( check:the ledger is append-only )
ACCEPTED at the door         | runtime rc=0 | ! { check:the ledger is append-only; }
refused: gate command does not start with a runnable command: 'check:the' | runtime rc=0 | ! check:the ledger is append-only
```

Severity dropped one level on purpose: I had to construct this. Zero of the 204 decoded `bypass_check` values in `.prd-os/issues/` start with `!`.

## Finding 2 — minor: a quoted or expanded first token is refused, and issue_runner turns that into `cannot close`

`plugins/prd-os/scripts/prd_runner.py:1245` takes `stripped.split()[0]` verbatim. The PR fixed the assignment and grouping shapes of "the raw first token is not the command name" and left quoting and expansion.

```
refused   rc=0    "'python3' -c 'pass'"
refused   rc=127  '$PYTHON -m pytest'
```

The first row is the live one: bash runs it at exit 0, the door refuses it. `issue_runner._enforce_spine_contract` runs the check first and requires exit 0, then converts any raise from `gate_register` into a hard `cannot close`, so this is an unattended closeout dying on a check that just went green. Dropped a level for the same reason as Finding 1: 0 of 204 real `bypass_check` values open with a quoted or expanded token (census of first tokens: 138 `python3`, 35 `bash`, 7 `test`, 4 `pytest`, 3 `grep`, 2 `PYTHONPATH=...`, 1 `sh`, remainder prose).

## Finding 3 — minor: the exit code still says green when nothing executed

`plugins/prd-os/scripts/prd_runner.py:3045` fixes the sentence, not the signal. 47 inert gates:

```
--- stdout ---
[census] spillover: 0 open total
[WARN] gates: 47 gate(s) registered, NONE executed -- 0 have lifecycle 'regression', so no regression check ran. Exit 0 below covers the spillover verdict ONLY.
NO regression gate executed (47 registered); 0 blocking spillover item(s)
--- exit code: 0
```

`wiring-check.md` names `gates run` exiting 0 as the definition of done, so an agent branching on `$?` reads the same green ASK-1038 produced. The boundary is stated in the code comment, and flipping it to non-zero would be red on this repo's own population, so this is a follow-up call, not a wedge. I found no in-repo script branching on the exit code, only tests and docs.

## What is sound

Attacks that failed, and the strongest parts:

- **Six mutants, six kills.** Deleting the `_reject_unrunnable_gate(command)` call site: `1 failed, 28 passed`, and the failure is `test_gate_register_ACTUALLY_CALLS_the_unrunnable_guard`. Dropping `lifecycle="regression"` at the closeout call site kills the AST test. `executed = len(records)` kills the all-skipped test. Suppressing the WARN kills two. Restoring the green sentence kills the round-3 test. `registered_total = 0` kills two. Nothing here is decoration.
- **The MAJOR-1 false-positive claim holds, measured not read.** I ran the door against all 204 decoded `bypass_check` values using `issue_runner`'s own `_parse_frontmatter` + `_decode_bypass_check`, not my own regex. 14 refused, all 14 genuine prose or genuine syntax errors, 0 false positives. Both `srsa-*` specs now pass.
- **The recovery command in the lifecycle-conflict error actually runs.** I executed the exact string the error prints: rc 0, registry reclassified, re-close then returns `{'registered': False}`. Not just a substring assertion.
- **The recovery is safe unattended**, which I expected to be a finding and it was not. `gate_id` hashes the command, so the conflicting row necessarily holds the same command issue_runner just ran at exit 0 — promoting it to `regression` cannot promote an unrunnable gate.
- **Half-failure leaves nothing behind.** `_enforce_spine_contract` returns an error string and close aborts before any state mutation.
- **Full suites green here**: prd-os `703 passed, 1 skipped`; kipi-dsse `110 passed`.
- Dropped: `echo the ledger is append-only` is accepted and exits 0 forever, but the docstring explicitly declares useless-but-runnable out of scope, and no description-shaped string starts with `echo`. Dropped: `bash -n` versus `/bin/sh` at run time, no divergence reproducible on darwin. Dropped: relative-path first tokens resolving against the wrong cwd, 0 producers emit one.

## Verdict

**APPROVE WITH NITS.**

First thing to fix: Finding 1. Continue the strip loop past `(` and `{` when a `!` was consumed, instead of returning on bash's parse verdict, since the negation flips the loud 127 into a silent green.

FINDINGS:
minor|A leading `!` in front of `(` or `{` reaches the grouping accept, so `! ( prose )` registers a permanently green gate|plugins/prd-os/scripts/prd_runner.py:1243
minor|The first-token probe still takes the raw split token, so a quoted or expanded command name is refused and issue_runner turns that into a hard `cannot close`|plugins/prd-os/scripts/prd_runner.py:1245
minor|`gates run` still exits 0 when the entire registry is inert, so a consumer reading the exit code sees the same green ASK-1038 produced|plugins/prd-os/scripts/prd_runner.py:3045
END FINDINGS

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-330-20260907-222000.md` (6898 bytes).

```
FINDINGS:
major|gates.jsonl resolves per-worktree and is gitignored, so a closeout run in a worktree registers its permanent regression gate into a file `gates run` in main never reads|plugins/prd-os/scripts/prd_runner.py:1134
minor|the negation guard only strips a LEADING `!`, so `bash -c '! <prose>'` and `x && ! <prose>` register a gate that exits 0 forever|plugins/prd-os/scripts/prd_runner.py:1233
minor|`gates run` still exits 0 when nothing executed, so the rc that wiring-check.md tells agents to trust cannot distinguish green from never-ran|plugins/prd-os/scripts/prd_runner.py:3057
minor|SKILL.md says gates.jsonl resolves to the shared worktree ledger root and README says it is committed; it is neither|plugins/prd-os/skills/prd-os/SKILL.md:65
END FINDINGS
```

--- reviewer output ---

I finished the review. Findings, reproducers, and real output below.

## Findings

### 1. major — the "permanent" gate registry is per-worktree and gitignored, so a closeout run in a worktree registers into a file `gates run` in main never reads

`plugins/prd-os/scripts/prd_runner.py:1134`

`_spillover_path` resolves through `_ledger_root` (git-common-dir, one ledger for all worktrees). `_gates_path` does not. It uses raw `cfg.repo_root`. And `.gitignore:43` excludes `*.jsonl` with only `receipts.jsonl` re-allowed, so `gates.jsonl` is never shared through git either.

This is the exact scar `_ledger_root`'s own docstring describes ("26 worktree ledgers held 71 open findings that did not exist in the main checkout's copy, so `gates run` from main was green about work it structurally could not see"), fixed for spillover and left open for gates. This worktree is the live proof: it has 232 issue specs and zero gates.

It matters for this PR specifically. The PR's content is making closeout register runnable gates again. In worktree mode that gate lands somewhere main cannot see, so the fix does not reach the place that runs it.

Reproducer, real git repo plus real worktree:

```
gate_register from worktree: {'gate_id': 'issue-wt-b5bea41b', 'registered': True}

worktree gates.jsonl exists: True
MAIN     gates.jsonl exists: False

--- `gates run` in MAIN after the worktree closeout ---
[census] spillover: 0 open total
NO regression gate executed (0 registered); 0 blocking spillover item(s)
rc= 0
```

Source asymmetry, printed from the loaded module:

```
def _gates_path(cfg: Config):
    return cfg.repo_root / ".prd-os" / "gates.jsonl"
def _spillover_path(cfg: Config):
    return _ledger_root(cfg.repo_root) / ".prd-os" / "spillover.jsonl"
```

Pre-existing code, not a line this diff touches. Stated plainly so the disposition is yours: the fix is one line (`_gates_path` calls `_ledger_root`) and is safe for main, whose `repo_root` already equals its ledger root.

### 2. minor — the negation guard only covers a LEADING `!`, so a negated prose command still registers green forever

`plugins/prd-os/scripts/prd_runner.py:1233`

Commit 2ca452ff identifies `!` as the one silent-green class. The `while stripped[:1] == "!"` loop only sees position 0. Three shapes get past it and each produces a gate that exits 0 forever:

```
ACCEPTED by the door | runtime rc=0 -> gate would be GREEN | bash -c '! check:the ledger is append-only'
ACCEPTED by the door | runtime rc=0 -> gate would be GREEN | cd . && ! check:the ledger is append-only
ACCEPTED by the door | runtime rc=0 -> gate would be GREEN | true && ! asserting the gate runs
refused: ...'check:the'... | runtime rc=0 | ! check:the ledger is append-only
```

The wrapper shape is real, not invented: 2 of the 204 decoded bypass_checks in `.prd-os/issues/` already use a non-leading `!` (`fable-merge-into-prd-os.md` uses `bash -c '... && ! grep -q ...'`). The prose-behind-the-negation part is mine, which is why this is minor and not major.

### 3. minor — `gates run` still exits 0 when nothing executed, and rc is what `wiring-check.md` tells agents to trust

`plugins/prd-os/scripts/prd_runner.py:3057`

The text half of ASK-1038 is fixed. The machine-readable half is not.

```
[WARN] gates: 1 gate(s) registered, NONE executed -- 0 have lifecycle 'regression', so no regression check ran. Exit 0 below covers the spillover verdict ONLY.
NO regression gate executed (1 registered); 0 blocking spillover item(s)
rc = 0
```

`.claude/rules/wiring-check.md` says the closeout proof is "`prd_runner.py gates run` exits 0". An unattended consumer reading rc cannot tell green from never-ran. The code comment names the tradeoff honestly, so this is a scoping note, not a hidden bug.

### 4. minor — two docs state the opposite of what the code does about `gates.jsonl`

`plugins/prd-os/skills/prd-os/SKILL.md:65` says `gates.jsonl` resolves to the shared worktree ledger root. It does not (finding 1). `plugins/prd-os/README.md:127` says `.prd-os/gates.jsonl` is "(committed)". `git ls-files .prd-os | grep jsonl` returns only `receipts.jsonl`.

## What is sound

- **Mutation-killed all four new guards.** On a copy of `plugins/`: deleting the `_reject_unrunnable_gate(command)` call site, dropping the `lifecycle="regression"` kwarg, reverting the WARN to `not records`, and forcing the green sentence each gave `1 failed, 28 passed` against a `29 passed` baseline. None of the new tests is decoration.
- **The door is calibrated against real producers, not fixtures.** Census over all 204 bypass_checks decoded with `issue_runner._parse_frontmatter` and `_decode_bypass_check`: 14 refused, and I read all 14. Every one is genuine prose. Zero false positives, so the MAJOR-1 assignment-prefix and subshell fix holds.
- **Shell reserved words resolve.** `command -v if / for / [[` all return 0, so `if ...; then ...; fi` gate commands are not refused.
- **The printed recovery command actually runs.** I extracted it from the raised message and executed it: rc=0, registry rewritten to `regression`, and the subsequent re-close returned `{'registered': False}`. Not just a substring assertion.
- **Full prd-os suite: 703 passed, 1 skipped in 60s.**
- **Single writer.** `gate_register` is the only path to `gates.jsonl` outside tests, so the door has no flanking route.
- **No external consumer greps the old success sentence,** so changing it breaks nothing.

Attacks I tried and dropped: relative-path gate commands refused by cwd-dependent `command -v` (no producer emits one), duplicate `gate_id` rows bricking `migrate_gate_lifecycle.py` (needs a concurrent same-issue closeout I could not reach, and it is pre-existing).

## VERDICT: REQUEST CHANGES

Fix first: `_gates_path` at `prd_runner.py:1134` routing through `_ledger_root`, the same way `_spillover_path` does. Without it this PR restores executable gates into a file that a worktree closeout writes and the main checkout cannot see, which is the load-path failure the plugin's own docstring says is worse than having no gate at all.

FINDINGS:
major|gates.jsonl resolves per-worktree and is gitignored, so a closeout run in a worktree registers its permanent regression gate into a file `gates run` in main never reads|plugins/prd-os/scripts/prd_runner.py:1134
minor|the negation guard only strips a LEADING `!`, so `bash -c '! <prose>'` and `x && ! <prose>` register a gate that exits 0 forever|plugins/prd-os/scripts/prd_runner.py:1233
minor|`gates run` still exits 0 when nothing executed, so the rc that wiring-check.md tells agents to trust cannot distinguish green from never-ran|plugins/prd-os/scripts/prd_runner.py:3057
minor|SKILL.md says gates.jsonl resolves to the shared worktree ledger root and README says it is committed; it is neither|plugins/prd-os/skills/prd-os/SKILL.md:65
END FINDINGS

assafkip:
Closed on Sana's ruling, ASK-2029 exit-side sweep (2026-09-22). Reason: Absorbed: main prd_runner.py already carries both rival guards (_reject_unrunnable_gate and SPILLOVER_SEVERITY_ORDER). 39% of the added lines are already on main.

Linked issue: none (unchanged). Branch kept; reopen this PR to undo. Label on the issue: stale-sweep-2026-09-22.

PR #331, opened 2026-09-08
voiceloop: the founder-data vacuous-pass guard was tautological

assafkip:
## The guard could not do its job

`test_the_tree_has_files_to_check` exists so that a broken or misfiltered walk cannot
let `test_no_founder_data_in_the_plugin_tree` grep almost nothing and report clean.

It derived its expected set as **every `.py` on disk**. `.py` matches no
`_BINARY_SUFFIXES` entry, so a `.py` file is in the scanned set *by construction* and
`py_on_disk - scanned` was empty regardless of what `_tree_files` did. Its only
reachable failure was a walk returning literally nothing.

Found by review on `ASK_AI_consultant#105` (MINOR), while carrying this guard into the
consulting instance.

## Measured against the exact regression it exists to prevent

Restoring the old `.py/.json/.jsonl/.md` allowlist that PR #311 removed:

```
hardened scan sees 45 files, allowlist scan sees 44
missed by the allowlist: requirements-authorship.txt

py-derived assertion   -> PASSES under both          <- blind
non-binary assertion   -> FAILS under the allowlist  <- catches it
```

Same shape as the `>= 7` floor PR #311 replaced: a guard watching a 45-file tree drop to
44, with the very file the hardening was written about among the missing, reporting
healthy.

## The fix

Derive the expected set with the **same predicate the scan uses**, negated off a raw
walk. The only way to satisfy it is now to actually scan every non-binary file, which is
the property the file is about.

## Verified, both directions

| state | result |
|---|---|
| as shipped | **4 passed** |
| allowlist put back | **1 failed** (`test_the_tree_has_files_to_check`), 3 passed |
| full package suite | **229 passed** |

This guard protects the PUBLIC mirror (`github assafkip/voice-loop`), which is why it
gets a real assertion rather than a decorative one.

### DIFF
--- a/plugins/kipi-core/.claude-plugin/plugin.json
+++ b/plugins/kipi-core/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "kipi-core",
-  "version": "1.12.4",
+  "version": "1.12.5",
   "description": "Core founder OS \u2014 voice, research, AUDHD, LinkedIn/brand, RCA skills + kipi-mcp server",
   "author": {
     "name": "Assaf Kipnis",
--- a/plugins/kipi-core/voiceloop/tests/test_no_founder_data.py
+++ b/plugins/kipi-core/voiceloop/tests/test_no_founder_data.py
@@ -80,20 +80,35 @@ def test_the_tree_has_files_to_check():
     NIT), so a walk that returned 8 files would have passed while grepping
     almost nothing. A floor that cannot notice a 36-file shortfall is decoration.
 
-    The honest version compares against the thing that must be there: every .py
-    the package ships has to appear in the scanned set. A truncated or
-    misfiltered walk fails this; a healthy one cannot.
+    ITS FIRST VERSION WAS TAUTOLOGICAL, and the fix is the whole point of this
+    docstring (PR #105 review, MINOR, 2026-09-07). It derived the expected set as
+    every `.py` on disk. But `.py` matches no `_BINARY_SUFFIXES` entry, so a `.py`
+    file is in the scanned set BY CONSTRUCTION and `py_on_disk - scanned` was empty
+    no matter what `_tree_files` did. Its only reachable failure was a walk that
+    returned literally nothing.
+
+    Measured against the exact regression it exists to prevent -- putting the old
+    `.py/.json/.jsonl/.md` allowlist back:
+
+        hardened scan sees 45 files, allowlist scan sees 44
+        missed by the allowlist: requirements-authorship.txt
+        py-derived assertion   -> PASSES under both        <- blind
+        non-binary assertion   -> FAILS under the allowlist <- catches it
+
+    So the expected set is derived with the SAME predicate the scan uses, negated
+    off a raw walk. Now the only way to satisfy it is to actually scan everything
+    that is not binary, which is the property the file is about.
     """
     scanned = set(_tree_files())
-    py_on_disk = {
+    on_disk = {
         os.path.join(root, f)
         for root, dirs, files in os.walk(PKG)
         if "__pycache__" not in root
         for f in files
-        if f.endswith(".py")
+        if not f.endswith(_BINARY_SUFFIXES)
     }
-    assert py_on_disk, "the walk found no .py at all -- it is broken"
-    missing = sorted(py_on_disk - scanned)
+    assert on_disk, "the walk found no files at all -- it is broken"
+    missing = sorted(on_disk - scanned)
     assert not missing, "the scan skipped shipped files: %s" % ", ".join(missing)
 
 


assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-331-20260907-211539.md` (6685 bytes).

```
FINDINGS:
minor|The rewritten guard derives its expected set from the same `_BINARY_SUFFIXES` constant the scan reads, so widening the denylist hides an unscanned file carrying founder data while all four tests stay green|plugins/kipi-core/voiceloop/tests/test_no_founder_data.py:108
minor|Pre-existing: the rendered-plist placeholder check greps `__`, which the embedded checkout path also produces, so the floor gate goes red in any worktree whose path contains a double underscore|automation/test_voice_refresh_schedule.py:65
END FINDINGS
```

--- reviewer output ---

## Review: PR #331, voiceloop founder-data vacuous-pass guard

The core claim checks out. The old `py_on_disk - scanned` was tautological, the new derivation is strictly better, and I verified the PR's own numbers rather than trusting the docstring: 45 non-binary versus 44 allowlist, `requirements-authorship.txt` the one missed. 229 tests pass. The suite is genuinely gated (`.verify-suites` names `plugins/kipi-core/voiceloop`, and `verify.sh --full` printed `pytest:plugins/kipi-core/voiceloop ok`).

Two findings survived reproduction.

### Finding 1 (minor): the assertion reads its bound from the same constant the subject reads

`test_no_founder_data.py:108` derives `on_disk` with `not f.endswith(_BINARY_SUFFIXES)`, the exact predicate `_tree_files` uses at line 72. Widen `_BINARY_SUFFIXES` and both sides move together. The docstring at line 99 says "the only way to satisfy it is to actually scan everything that is not binary". There is a second way: add a suffix.

Reproducer, run against a temp copy, never the repo:

```
python3 - <<'PYEOF'
import os, shutil, subprocess, sys, tempfile
SRC = "/Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-331/plugins/kipi-core/voiceloop"
def build(tmp, mutate):
    dst = os.path.join(tmp, "voiceloop")
    shutil.copytree(SRC, dst, ignore=shutil.ignore_patterns("__pycache__"))
    tf = os.path.join(dst, "tests", "test_no_founder_data.py")
    src = open(tf).read()
    if mutate:
        old = '    ".pyc", ".pyo", ".so", ".dylib", ".png", ".jpg", ".jpeg", ".gif",'
        open(tf, "w").write(src.replace(old, old + '\n    ".txt",'))
    with open(os.path.join(dst, "requirements-authorship.txt"), "a") as h:
        h.write("\n# corpus owner: Assaf Kipnis, q-consult\n")
    return dst
def run(dst):
    p = subprocess.run([sys.executable, "-m", "pytest",
        os.path.join(dst, "tests", "test_no_founder_data.py"), "-q",
        "-p", "no:cacheprovider"], capture_output=True, text=True)
    return p.returncode, [l for l in (p.stdout+p.stderr).splitlines() if l.strip()][-2:]
for label, mut in (("A. as shipped", False), ('B. ".txt" added to _BINARY_SUFFIXES', True)):
    rc, tail = run(build(tempfile.mkdtemp(prefix="pr331_"), mut))
    print("===", label, ":: rc=%d" % rc); [print("   "+l) for l in tail]
PYEOF
```

Real output:

```
=== A. as shipped :: rc=1
   FAILED .../voiceloop/tests/test_no_founder_data.py::test_no_founder_data_in_the_plugin_tree
   1 failed, 3 passed in 0.01s
=== B. ".txt" added to _BINARY_SUFFIXES :: rc=0
   ....                                                                     [100%]
   4 passed in 0.01s
```

`Assaf Kipnis, q-consult` sits in a tracked file of a public repo and all four tests are green. This is the repo's own lesson `the-author-of-a-fix-picks-the-oracle-the-fix-already-passes`, point 2: mutate the code, not the constant your assertion reads.

Severity note, honestly: I had to construct the mutation, no producer emits it, and this PR is strictly better than what it replaces. Minor, not major.

Fix shape: pin the denylist against a committed literal (a small fixture list, or a count/hash the test asserts separately), so a suffix added to `_BINARY_SUFFIXES` has to be added in two places by a human who sees what they are exempting. Or drop the line-99 claim to what the code actually holds.

### Finding 2 (minor, pre-existing, not caused by this diff): the floor gate is red in this checkout

`verify.sh --full` on this branch: `verify.sh FAILED (1/13): pytest:automation`.

```
cd .../automation && python3 -m pytest -q --no-header
```

```
E  AssertionError: placeholder left in rendered plist
E    '__' is contained here:
E      ipi-system__pr-331/automation/voice-refresh-nudge.sh</string>
FAILED test_voice_refresh_schedule.py::test_template_renders_through_the_single_installer
1 failed, 12 passed in 0.06s
```

`automation/test_voice_refresh_schedule.py:65` asserts no `__` survives in the rendered plist, but the plist embeds the absolute repo path, and the fleet's own review-worktree naming (`assafkip_kipi-system__pr-331`) puts a double underscore in that path. The check greps a substring that a second producer, the checkout path, also emits. Any worktree with `__` in its path turns the floor gate red for a reason nobody can fix by committing, which is the exact "how a gate gets switched off in a day" shape `.verify-suites` warns about in its own header.

Out of scope for this diff. Filing it here so it gets captured rather than dropped.

### What is sound (attacks that failed)

- **Docstring numbers invented.** Recomputed independently: 45 non-binary, 44 allowlist, missed set exactly `['requirements-authorship.txt']`. Matches.
- **`229 passed` overstated.** Ran it. 229 passed in 0.45s.
- **The guard has no runner.** I expected this one to land. The capability gate scans only `q-system/.q-system/scripts` and `q-system/.q-system` (`capability-gate.py:46`, `SCAN_ROOTS`), so it does not cover this file. But `.verify-suites` names `plugins/kipi-core/voiceloop`, and `verify.sh --full` printed `pytest:plugins/kipi-core/voiceloop ok`. Dropped.
- **Walk divergence between the two loops.** `_tree_files` prunes dirs named `__pycache__`; `on_disk` filters `"__pycache__" not in root`. I looked for an input where they disagree in the unsafe direction and could not construct one.
- **`.pytest_cache` landing inside PKG causes a false positive.** Ran the suite twice from inside the package dir. The cache dir appeared (`CACHEDIR.TAG`, `README.md`, `.gitignore`, `v/cache/nodeids`), both walks saw it identically, node ids are relative so no founder path leaked into content. Green both runs. Dropped.
- **Plugin version bump missing.** Present, 1.12.4 to 1.12.5.
- **Public mirror copy drifts.** The exporter lives in the consulting repo, not here. No sibling copy in this tree to keep in sync.

The PR does the thing it says it does and the previous assertion really was blind. Fix first: line 99's claim overstates what the code holds, and the denylist-shared-with-the-assertion is the one remaining way to go green on an unscanned file.

FINDINGS:
minor|The rewritten guard derives its expected set from the same `_BINARY_SUFFIXES` constant the scan reads, so widening the denylist hides an unscanned file carrying founder data while all four tests stay green|plugins/kipi-core/voiceloop/tests/test_no_founder_data.py:108
minor|Pre-existing: the rendered-plist placeholder check greps `__`, which the embedded checkout path also produces, so the floor gate goes red in any worktree whose path contains a double underscore|automation/test_voice_refresh_schedule.py:65
END FINDINGS

PR #327, opened 2026-09-08
fix(board): the morning board carries only what he can act on, and every row can be started

assafkip:
## What

The morning board carries only what he can act on, and every row it carries can be started from the row. Founder-directed 2026-09-07 after a row-by-row CRM audit of the live board.

## Why

His words, that evening: *"remove sana stuff from the board"* and *"I want you to revamp the crm so it fits what I asked. Dont punt anything to me. Fix it. right now its not useful."* DEC-34 in the consulting instance's `crm-working.md` is the record.

Measured on the live board, 12 rows: zero carried a URL, six printed one sentence twice, five were Sana's build queue wearing his done signal, and eleven had been machine-written into a section whose own text says nothing fills it automatically. Full audit: `q-consult/output/board-crm-audit-2026-09-07.md` in the consulting instance.

## How

- A ⚪ or 🟢 card line never becomes a board row. Both stay on the Slack card, which is DEC-30's design and does not change; neither is an act he can perform.
- Everything the card does surface goes to Top of Mind. This Week is no longer written from the card.
- A Gmail row carries its thread link, derived from the thread id the row already has (`collect_mail` emits `mail:<thread id>`). No new plumbing, no second source of truth.
- `_properties` writes `Link` and `Next` when a producer supplies them and never blanks them when it does not, so a row a human filled in keeps what they wrote. Same posture as `Size`, which this module still deliberately does not write.
- The detail is not the done signal repeated.
- `DONE_BY_KIND["client"]` no longer reads "you sent the thing you promised", wording that predates DEC-30.

## Proof

- `test_board_is_actionable.py`, 13 cases. **10 were red against these files as they stood**; the 3 that pass either way are the guards (the card still carries its white lines, a producer that knows neither Link nor Next clears neither, a detail that adds something is still kept).
- 237 passed across `test_consulting_board.py`, `test_morning_brief.py`, `test_morning_brief_mail.py` and the new file.
- The expected property set is written out by hand rather than derived from the writer's own map: a test that iterates the thing it tests cannot see that thing shrink.
- Live preview against the founder's real state: the five ⚪ rows are gone from every bucket.

## Known and not fixed here

Two rows archived by hand that evening will be recreated on the next paint. The GTM queue records a constraint (`3.3`, a client's fixed signature order) and a talk-track rule (`5.5`) with fields identical to a real action, so `read_week` cannot tell them apart. That is a fix at the plan, not the painter. Captured as `sp-c06be056`.

[...]

PR #332, opened 2026-09-08
fix(board): the board creates the columns it writes, so the link is not dead everywhere (ASK-1346)

assafkip:
## What

The morning board creates the columns it writes, so the row link is not silently dead on every board but one. Linear: ASK-1346. Follow-up to #327 (merged `bffa124f`).

## Why

#327's round-12 review found the hole its own earlier rounds had left: **nothing ever created the `Link` column.** It had been added by hand, once, on the founder's board. Anywhere else `_only_known` dropped it on every run and the morning line reported "this board cannot take the Link column" every morning, forever, naming no way to change that. A permanent explanation is not a fix.

## How

- `ensure_columns` adds the optional columns this module writes (`Link`, `Next`), once, before the identity check, so a board heals rather than complains.
- A refusal from Notion is not fatal: the schema is returned unchanged and the existing drop-and-report path handles it, which is the behaviour that shipped in #327.
- The identity columns are deliberately **not** creatable. A board with no `Item id` is not one this module should quietly reshape; it is one someone should look at, which is what `MissingIdentityColumn` already says.

## Also carried

#327 squash-merged one commit before its tip, so the round-11 review's two nits landed nowhere. They are in this PR: the morning line's dropped-columns sentence and `held` counter are pinned by four cases (mutation-run: deleting them turns three red), and the dead-link guard's comment no longer cites a producer shape ASK-1323 removed.

## Proof

- 57 cases in `test_board_is_actionable.py`, 6 new: a missing column is created and then usable, a complete board is not touched, a refused creation degrades exactly as before, an unreadable schema creates nothing, the identity columns are never in `CREATABLE`, and the runner heals before it checks identity.
- `python3 -m pytest q-system/.q-system/tests/ -q`: 1370 passed, 1 skipped.

### DIFF
--- a/q-system/.q-system/scripts/board_rows.py
+++ b/q-system/.q-system/scripts/board_rows.py
@@ -618,7 +618,38 @@ def _only_known(props: dict, known):
 CREATABLE = tuple(n for n in WRITES_TYPE if n not in UNDROPPABLE)
 
 
-def ensure_columns(known, token, db, opener=None, budget=None):
+#: Columns this module has created before, per board. A column in here that is now
+#: ABSENT was deleted by a person, and this module does not put it back.
+COLUMNS_MADE = STATE_DIR / "board-columns-created.json"
+
+
+def _columns_made(db, path=None):
+    try:
+        return set((json.loads(Path(path or COLUMNS_MADE).read_text("utf-8")) or {}).get(db) or [])
+    except (OSError, ValueError):
+        return set()
+
+
+def _remember_columns(db, names, path=None):
+    """Append to the record. A failure here is not fatal: the worst case is offering to
+    create a column he removed a second time, which is the behaviour before the record
+    existed, and it is still said out loud on the line."""
+    p = Path(path or COLUMNS_MADE)
+    try:
+        data = json.loads(p.read_text("utf-8")) or {}
+    except (OSError, ValueError):
+        data = {}
+    if not isinstance(data, dict):
+        data = {}
+    data[db] = sorted(set(data.get(db) or []) | set(names))
+    try:
+        p.parent.mkdir(parents=True, exist_ok=True)
+        p.write_text(json.dumps(data, indent=1, sort_keys=True), "utf-8")
+    except OSError:
+        pass
+
+
+def ensure_columns(known, token, db, opener=None, budget=None, record=None):
     """Add the optional columns this board is missing. Returns (schema, problems).
 
     THE FEATURE WAS DEAD ON EVERY BOARD BUT ONE (PR reviewer round 12, major). Nothing
@@ -634,14 +665,25 @@ def ensure_columns(known, token, db, opener=None, budget=None):
     if known is None:
         return None, ()
     missing = {name: WRITES_TYPE[name] for name in CREATABLE if name not in known}
+    # A COLUMN HE DELETED IS NOT A COLUMN THAT IS MISSING (PR #332 reviewer round 4,
+    # major). Without this the module re-created it every single morning and never said
+    # so: he would remove a column he did not want and find it back the next day, with
+    # no way to make it stop. That is the same "his choice wins" line the row painter
+    # holds everywhere else, one level down at the schema.
+    made_before = _columns_made(db, record)
+    removed = sorted(n for n in missing if n in made_before)
+    for n in removed:
+        del missing[n]
     # A column PRESENT UNDER THE WRONG TYPE is deliberately not healed here (PR #332
     # reviewer, minor). Creating an absent column adds nothing and loses nothing;
     # retyping an existing one makes Notion convert every value in it, which is a data
     # decision and not this module's to take. What was wrong was the message: it said
     # the board "cannot take" the column, so he would go looking for something that is
     # sitting right there. `_only_known`'s report now says which it is.
+    told = tuple(f"{n} was created here before and is gone now, so it is treated as "
+                 "your removal and is not being put back" for n in removed)
     if not missing:
-        return known, ()
+        return known, told
     try:
         data = _request(token, "PATCH", f"/databases/{db}",
                         {"properties": {n: {t: {}} for n, t in missing.items()}},
@@ -651,7 +693,16 @@ def ensure_columns(known, token, db, opener=None, budget=None):
         # the same sentence, so he could not tell "this board has no Link column" from
         # "I tried to add one and Notion said no", and only the second is about
         # permissions or a token. The reason is carried back and said.
-        return known, ((f"{type(exc).__name__}: {exc}")[:160],)
+        #
+        # AND THE BODY, NOT JUST THE STATUS LINE (round 4, minor). "HTTPError: 400 Bad
+        # Request" is the half that says nothing; Notion puts what is actually wrong in
+        # the response body, which is the whole reason this branch exists.
+        detail = ""
+        try:
+            detail = " " + exc.read().decode("utf-8", "replace")[:200]
+        except Exception:
+            pass
+        return known, told + ((f"{type(exc).__name__}: {exc}{detail}")[:300],)
     # WHAT NOTION RETURNED, not what we asked for (PR #332 reviewer, minor). Synthesising
     # `dict(known, **missing)` asserts the create took effect; if it did not, the next
     # write carries the column anyway and takes the raw 400 this whole guard exists to
@@ -659,9 +710,17 @@ def ensure_columns(known, token, db, opener=None, budget=None):
     # properties are the answer.
     fresh = (data or {}).get("properties")
     if isinstance(fresh, dict) and fresh:
-        return {n: (p or {}).get("type") for n, p in fresh.items()}, ()
-    return known, ("the schema PATCH returned no properties, so the new columns are "
-                   "unconfirmed and were not written",)
+        made = sorted(n for n in missing if n in fresh)
+        if made:
+            _remember_columns(db, made, record)
+        # SAID OUT LOUD (round 4, major). Adding columns to his board is a change to
+        # his board, and a change he is not told about is one he cannot disagree with.
+        return ({n: (p or {}).get("type") for n, p in fresh.items()},
+                told + ((f"added the {', '.join(made)} "
+                         + ("columns" if len(made) > 1 else "column")
+                         + " to this board",) if made else ()))
+    return known, told + ("the schema PATCH returned no properties, so the new columns "
+                          "are unconfirmed and were not written",)
 
 
 def _refuse_without_identity(known):
@@ -951,8 +1010,8 @@ def collect(now, sources: dict, opener=None, token_file=None, db_file=None,
             f"{counts['kept']} kept (source quiet), {counts['pinned']} yours (untouched), "
             f"{counts['held']} yours (source stopped reporting it), "
             "read-back ok")
-    for problem in counts.get("column_problems") or ():
-        line += f"; could not add a column: {problem}"
+    for note in counts.get("column_problems") or ():
+        line += f"; {note}"
     if counts["dropped_columns"]:
         # NEVER SILENT. The filter stops a missing column aborting the paint; it must
         # not also hide that the rows went out without it. A board missing `Link` wrote
--- a/q-system/.q-system/scripts/board_rows.py
+++ b/q-system/.q-system/scripts/board_rows.py
@@ -658,7 +658,19 @@ def _columns_made(db, path=None):
 def _remember_columns(db, names, path=None):
     """Append to the record. A failure here is not fatal: the worst case is offering to
     create a column he removed a second time, which is the behaviour before the record
-    existed, and it is still said out loud on the line."""
+    existed, and it is still said out loud on the line.
+
+    THE LIVE RECORD IS NEVER WRITTEN BY A TEST, and that is a chokepoint rather than a
+    rule someone remembers (PR #332 reviewer round 7, minor). `collect` has had this
+    guard for the board itself; this file did not, so a test that simply forgot
+    `record=` wrote into ~/.config/kipi and taught the 07:40 job that he had deleted
+    columns he never touched. That happened, once, in this PR. Three separate rounds
+    of it found a test reaching something live and each was fixed by hand; a hand fix
+    protects the tests that exist today. This one protects the ones nobody has written.
+    """
+    if path is None and os.environ.get("PYTEST_CURRENT_TEST"):
+        raise AssertionError(
+            "a test tried to write the live column record; pass record=<tmp path>")
     p = Path(path or COLUMNS_MADE)
     try:
         data = json.loads(p.read_text("utf-8")) or {}
@@ -716,10 +728,16 @@ def ensure_columns(known, token, db, opener=None, budget=None, record=None):
     # entirely and hides every row from his three sections, WHILE the line still ended
     # "read-back ok". Silent degradation is the one thing this file exists to refuse.
     #
-    # The real distinction is consequence, and it needs no state at all. Losing `Link`
-    # or `Next` costs a value on a row and nothing else, so his removal of one is his
-    # business and is not mentioned again. Losing `Notes` or `Bucket` breaks the board,
-    # so it is named every single run, with what it costs, until he puts it back.
+    # The real distinction is consequence, and it needs no state at all. Losing `Notes`
+    # or `Bucket` breaks the board, so it is named every run, with what it costs, until
+    # he puts it back.
+    #
+    # NOT SILENCE, ONE MENTION (round 7, minor: the previous comment and its test both
+    # claimed silence and the dropped-columns sentence still named the column every
+    # run, so the claim was simply false). Losing `Link` or `Next` costs one value on a
+    # row, so it earns no SECOND sentence explaining itself, and the report that the
+    # value was not written stands as it does for any other column. Suppressing that
+    # too is what round 5 did, and it hid `Notes` and `Bucket` with it.
     told = tuple(
         f"{n} is gone and this module created it, so it is treated as your removal; "
         + STRUCTURAL_COST[n] for n in removed if n in STRUCTURAL_COST)
@@ -743,7 +761,13 @@ def ensure_columns(known, token, db, opener=None, budget=None, record=None):
             detail = " " + exc.read().decode("utf-8", "replace")[:200]
         except Exception:
             pass
-        return known, told + ((f"{type(exc).__name__}: {exc}{detail}")[:300],)
+        # THE CONSEQUENCE BELONGS TO THE ABSENCE, NOT TO THE REASON (round 7, minor).
+        # It was attached only to his deletion, so a board where Notion REFUSED to
+        # create `Notes` reported the refusal and never that nothing would be archived,
+        # which is the half he needs.
+        cost = tuple(f"{n}: {STRUCTURAL_COST[n]}" for n in sorted(missing)
+                     if n in STRUCTURAL_COST)
+        return known, told + ((f"{type(exc).__name__}: {exc}{detail}")[:300],) + cost
     # WHAT NOTION RETURNED, not what we asked for (PR #332 reviewer, minor). Synthesising
     # `dict(known, **missing)` asserts the create took effect; if it did not, the next
     # write carries the column anyway and takes the raw 400 this whole guard exists to
@@ -760,8 +784,10 @@ def ensure_columns(known, token, db, opener=None, budget=None, record=None):
                 told + ((f"added the {', '.join(made)} "
                          + ("columns" if len(made) > 1 else "column")
                          + " to this board",) if made else ()))
+    unconfirmed = tuple(f"{n}: {STRUCTURAL_COST[n]}" for n in sorted(missing)
+                        if n in STRUCTURAL_COST)
     return known, told + ("the schema PATCH returned no properties, so the new columns "
-                          "are unconfirmed and were not written",)
+                          "are unconfirmed and were not written",) + unconfirmed
 
 
 def _refuse_without_identity(known):
--- a/q-system/.q-system/scripts/board_rows.py
+++ b/q-system/.q-system/scripts/board_rows.py
@@ -596,6 +596,39 @@ def _only_known(props: dict, known):
     return keep, tuple(sorted(dropped))
 
 
+#: Columns this module writes that it will CREATE on a board that lacks them. The
+#: identity columns are NOT here on purpose: a board with no `Item id` is not a board
+#: this module should quietly reshape, it is a board someone should look at.
+CREATABLE = ("Link", "Next")
+
+
+def ensure_columns(known, token, db, opener=None, budget=None):
+    """Add the optional columns this board is missing. Returns the schema to use.
+
+    THE FEATURE WAS DEAD ON EVERY BOARD BUT ONE (PR reviewer round 12, major). Nothing
+    created `Link`, so `_only_known` dropped it on every run and the morning line said
+    the board could not take it and named no way to change that. The column had been
+    added by hand, once, on the founder's own board; a fresh instance would have
+    reported the same sentence every morning forever. Silent degradation with a
+    permanent explanation is not a fix.
+
+    Failure is not fatal: on a refusal the schema is returned unchanged and the drop
+    path reports it, which is exactly the behaviour before this existed.
+    """
+    if known is None:
+        return None
+    missing = {name: WRITES_TYPE[name] for name in CREATABLE if name not in known}
+    if not missing:
+        return known
+    try:
+        _request(token, "PATCH", f"/databases/{db}",
+                 {"properties": {n: {t: {}} for n, t in missing.items()}},
+                 opener, budget)
+    except Exception:
+        return known           # reported as dropped, same as before
+    return dict(known, **missing)
+
+
 def _refuse_without_identity(known):
     """Raise when the board cannot take a column the painter works by. `known` None
     (schema unreadable) is not a verdict about the columns, so it passes."""
@@ -823,6 +856,9 @@ def collect(now, sources: dict, opener=None, token_file=None, db_file=None,
             # and `Next` would otherwise take a 400 on the first row and lose the
             # whole morning's paint to a column nobody noticed was missing.
             known = _schema_properties(token, db, opener, budget)
+            # Create what is missing BEFORE the identity check, so a board that only
+            # lacks the optional columns heals instead of reporting them every day.
+            known = ensure_columns(known, token, db, opener, budget)
             # BEFORE THE FIRST QUERY, not inside the write loop. `existing_rows`
             # filters on `Item id`, so a board without that column takes a raw 400
             # from Notion before `_only_known` is ever reached and the remediation
--- a/q-system/.q-system/scripts/board_rows.py
+++ b/q-system/.q-system/scripts/board_rows.py
@@ -606,10 +606,16 @@ def _only_known(props: dict, known):
     return keep, dropped
 
 
-#: Columns this module writes that it will CREATE on a board that lacks them. The
-#: identity columns are NOT here on purpose: a board with no `Item id` is not a board
-#: this module should quietly reshape, it is a board someone should look at.
-CREATABLE = ("Link", "Next")
+#: Columns this module writes that it will CREATE on a board that lacks them:
+#: EVERYTHING IT WRITES except the identity pair. Listing only Link and Next was a
+#: half-heal (PR #332 reviewer round 3, major): a board missing `Notes` cannot carry
+#: the `scope=` line, so `_scope_of` reads unknown and no row is ever archived, and a
+#: board missing `Bucket` puts every row in none of his three sections. Both were
+#: "healed" boards reporting no problem at all.
+#:
+#: DERIVED, never restated. A column added to WRITES_TYPE and forgotten here would be
+#: the same silent half-heal again, one release later.
+CREATABLE = tuple(n for n in WRITES_TYPE if n not in UNDROPPABLE)
 
 
 def ensure_columns(known, token, db, opener=None, budget=None):
--- a/q-system/.q-system/scripts/board_rows.py
+++ b/q-system/.q-system/scripts/board_rows.py
@@ -580,20 +580,30 @@ def _only_known(props: dict, known):
     """
     if known is None:
         return props, ()
-    keep, dropped = {}, []
+    # TWO LISTS ON PURPOSE. `names` is what the code reasons about; `shown` is what he
+    # reads. Annotating the single list broke the undroppable check below, because
+    # "Item id" is not "Item id (it is select, this writes rich_text)" and the identity
+    # refusal silently stopped firing for the wrong-type case. A display string is not
+    # an identifier.
+    keep, names, shown = {}, [], []
     for k, v in props.items():
         want = WRITES_TYPE.get(k)
         if k in known and (want is None or known[k] == want):
             keep[k] = v
-        else:
-            dropped.append(k)
-    hard = [k for k in UNDROPPABLE if k in dropped]
+            continue
+        names.append(k)
+        # PRESENT, WRONG TYPE. Naming it as absent sends him looking for something that
+        # is there; naming the types tells him the one edit that fixes it.
+        shown.append(f"{k} (it is {known[k]}, this writes {want})"
+                     if k in known else k)
+    dropped = tuple(s for _, s in sorted(zip(names, shown)))
+    hard = [k for k in UNDROPPABLE if k in names]
     if hard:
         raise MissingIdentityColumn(
             f"the board cannot take {', '.join(hard)}, which the painter identifies "
             "rows by. Writing rows without it would create pages this module can never "
             "find or archive again, one per row per run. Fix the board's columns.")
-    return keep, tuple(sorted(dropped))
+    return keep, dropped
 
 
 #: Columns this module writes that it will CREATE on a board that lacks them. The
@@ -618,6 +628,12 @@ def ensure_columns(known, token, db, opener=None, budget=None):
     if known is None:
         return None
     missing = {name: WRITES_TYPE[name] for name in CREATABLE if name not in known}
+    # A column PRESENT UNDER THE WRONG TYPE is deliberately not healed here (PR #332
+    # reviewer, minor). Creating an absent column adds nothing and loses nothing;
+    # retyping an existing one makes Notion convert every value in it, which is a data
+    # decision and not this module's to take. What was wrong was the message: it said
+    # the board "cannot take" the column, so he would go looking for something that is
+    # sitting right there. `_only_known`'s report now says which it is.
     if not missing:
         return known
     try:
@@ -856,15 +872,19 @@ def collect(now, sources: dict, opener=None, token_file=None, db_file=None,
             # and `Next` would otherwise take a 400 on the first row and lose the
             # whole morning's paint to a column nobody noticed was missing.
             known = _schema_properties(token, db, opener, budget)
-            # Create what is missing BEFORE the identity check, so a board that only
-            # lacks the optional columns heals instead of reporting them every day.
-            known = ensure_columns(known, token, db, opener, budget)
-            # BEFORE THE FIRST QUERY, not inside the write loop. `existing_rows`
-            # filters on `Item id`, so a board without that column takes a raw 400
-            # from Notion before `_only_known` is ever reached and the remediation
-            # sentence never fires (round 10, minor). Checked here, the one failure a
-            # person can fix is the one they are told about.
+            # IDENTITY FIRST, THEN HEAL (PR #332 reviewer, major). The first cut had
+            # these the other way round, so a database this module then rejected as
+            # "not our board" had already had two columns PATCHed onto it. Deciding
+            # whether a thing is ours has to come before writing to it. Worse, the
+            # test that shipped with it asserted the wrong order and called it
+            # correct, which is how a mistake gets a guard of its own.
+            #
+            # `_refuse_without_identity` also has to run BEFORE the first query:
+            # `existing_rows` filters on `Item id`, so a board without that column
+            # takes a raw 400 from Notion before `_only_known` is ever reached and the
+            # remediation sentence never fires (#327 round 10, minor).
             _refuse_without_identity(known)
+            known = ensure_columns(known, token, db, opener, budget)
             counts = paint(buckets, token, db, opener, budget, known=known)
         dupes = {}
         seen = len(existing_rows(token, db, opener, dupes_out=dupes, budget=budget))
--- a/q-system/.q-system/scripts/board_rows.py
+++ b/q-system/.q-system/scripts/board_rows.py
@@ -622,12 +622,29 @@ CREATABLE = tuple(n for n in WRITES_TYPE if n not in UNDROPPABLE)
 #: ABSENT was deleted by a person, and this module does not put it back.
 COLUMNS_MADE = STATE_DIR / "board-columns-created.json"
 
+#: Per board, the columns HE removed after this module created them. Read by the
+#: dropped-columns report so his own choice is not also listed as a board fault.
+_chosen_absent: dict = {}
+
 
 def _columns_made(db, path=None):
+    """What this module created on `db` before. Unreadable or malformed reads EMPTY.
+
+    `_remember_columns` already refused a non-dict and this did not, so a file holding
+    a JSON list raised AttributeError out of `.get`, which no `collect` handler catches
+    (they take OSError and ValueError), and the entire board section died on a
+    malformed file this module writes itself (PR #332 reviewer round 5, minor). Empty
+    is the safe answer: the worst it costs is offering a column he removed once more,
+    and that is said on the line.
+    """
     try:
-        return set((json.loads(Path(path or COLUMNS_MADE).read_text("utf-8")) or {}).get(db) or [])
+        data = json.loads(Path(path or COLUMNS_MADE).read_text("utf-8"))
     except (OSError, ValueError):
         return set()
+    if not isinstance(data, dict):
+        return set()
+    names = data.get(db)
+    return set(names) if isinstance(names, (list, tuple, set)) else set()
 
 
 def _remember_columns(db, names, path=None):
@@ -674,6 +691,11 @@ def ensure_columns(known, token, db, opener=None, budget=None, record=None):
     removed = sorted(n for n in missing if n in made_before)
     for n in removed:
         del missing[n]
+    # NAMED ONCE, AS HIS CHOICE. The dropped-columns sentence also lists these, so the
+    # line said the same column twice: once as his removal and once as something the
+    # board "cannot take", on a line already near 900 characters (round 5, minor). A
+    # choice he made is not a fault to report back to him.
+    _chosen_absent[db] = set(removed)
     # A column PRESENT UNDER THE WRONG TYPE is deliberately not healed here (PR #332
     # reviewer, minor). Creating an absent column adds nothing and loses nothing;
     # retyping an existing one makes Notion convert every value in it, which is a data
@@ -988,8 +1010,10 @@ def collect(now, sources: dict, opener=None, token_file=None, db_file=None,
     except (urllib.error.URLError, urllib.error.HTTPError, OSError, ValueError) as exc:
         return [], f"board write failed: {type(exc).__name__}: {exc}"
 
+    notes = "".join(f"; {n}" for n in (counts.get("column_problems") or ()))
     if dupes:
-        return [], (f"duplicate board rows for {len(dupes)} item(s): "
+        return [], (notes.lstrip("; ") + ("; " if notes else "")
+                    + f"duplicate board rows for {len(dupes)} item(s): "
                     f"{', '.join(sorted(dupes))}. Two painters have run; the board "
                     "holds doubles and this run's counts cannot be trusted")
     # `kept` rows belong to a source that could not answer this run: they are on the
@@ -1001,7 +1025,8 @@ def collect(now, sources: dict, opener=None, token_file=None, db_file=None,
     if seen != expected:
         # The write-only-integration scar: a PATCH that returns 200 is not proof the
         # board holds what we think. The read-back is the proof.
-        return [], (f"read-back mismatch: expected {expected} row(s) "
+        return [], (notes.lstrip("; ") + ("; " if notes else "")
+                    + f"read-back mismatch: expected {expected} row(s) "
                     f"({counts['wanted']} written + {counts['kept']} kept from a quiet "
                     f"source + {counts['held']} held for you), board shows {seen}")
     line = (f"board: {counts['created']} new, {counts['updated']} refreshed, "
@@ -1012,6 +1037,9 @@ def collect(now, sources: dict, opener=None, token_file=None, db_file=None,
             "read-back ok")
     for note in counts.get("column_problems") or ():
         line += f"; {note}"
+    chosen = set().union(*_chosen_absent.values()) if _chosen_absent else set()
+    counts["dropped_columns"] = tuple(
+        d for d in counts["dropped_columns"] if d.split(" (")[0] not in chosen)
     if counts["dropped_columns"]:
         # NEVER SILENT. The filter stops a missing column aborting the paint; it must
         # not also hide that the rows went out without it. A board missing `Link` wrote
--- a/q-system/.q-system/scripts/board_rows.py
+++ b/q-system/.q-system/scripts/board_rows.py
@@ -613,7 +613,7 @@ CREATABLE = ("Link", "Next")
 
 
 def ensure_columns(known, token, db, opener=None, budget=None):
-    """Add the optional columns this board is missing. Returns the schema to use.
+    """Add the optional columns this board is missing. Returns (schema, problems).
 
     THE FEATURE WAS DEAD ON EVERY BOARD BUT ONE (PR reviewer round 12, major). Nothing
     created `Link`, so `_only_known` dropped it on every run and the morning line said
@@ -626,7 +626,7 @@ def ensure_columns(known, token, db, opener=None, budget=None):
     path reports it, which is exactly the behaviour before this existed.
     """
     if known is None:
-        return None
+        return None, ()
     missing = {name: WRITES_TYPE[name] for name in CREATABLE if name not in known}
     # A column PRESENT UNDER THE WRONG TYPE is deliberately not healed here (PR #332
     # reviewer, minor). Creating an absent column adds nothing and loses nothing;
@@ -635,14 +635,27 @@ def ensure_columns(known, token, db, opener=None, budget=None):
     # the board "cannot take" the column, so he would go looking for something that is
     # sitting right there. `_only_known`'s report now says which it is.
     if not missing:
-        return known
+        return known, ()
     try:
-        _request(token, "PATCH", f"/databases/{db}",
-                 {"properties": {n: {t: {}} for n, t in missing.items()}},
-                 opener, budget)
-    except Exception:
-        return known           # reported as dropped, same as before
-    return dict(known, **missing)
+        data = _request(token, "PATCH", f"/databases/{db}",
+                        {"properties": {n: {t: {}} for n, t in missing.items()}},
+                        opener, budget)
+    except Exception as exc:
+        # A REFUSAL IS NOT AN ABSENCE (PR #332 reviewer, minor). Both used to end in
+        # the same sentence, so he could not tell "this board has no Link column" from
+        # "I tried to add one and Notion said no", and only the second is about
+        # permissions or a token. The reason is carried back and said.
+        return known, ((f"{type(exc).__name__}: {exc}")[:160],)
+    # WHAT NOTION RETURNED, not what we asked for (PR #332 reviewer, minor). Synthesising
+    # `dict(known, **missing)` asserts the create took effect; if it did not, the next
+    # write carries the column anyway and takes the raw 400 this whole guard exists to
+    # prevent, losing the morning's paint. The PATCH response is the database, so its
+    # properties are the answer.
+    fresh = (data or {}).get("properties")
+    if isinstance(fresh, dict) and fresh:
+        return {n: (p or {}).get("type") for n, p in fresh.items()}, ()
+    return known, ("the schema PATCH returned no properties, so the new columns are "
+                   "unconfirmed and were not written",)
 
 
 def _refuse_without_identity(known):
@@ -884,8 +897,9 @@ def collect(now, sources: dict, opener=None, token_file=None, db_file=None,
             # takes a raw 400 from Notion before `_only_known` is ever reached and the
             # remediation sentence never fires (#327 round 10, minor).
             _refuse_without_identity(known)
-            known = ensure_columns(known, token, db, opener, budget)
+            known, column_problems = ensure_columns(known, token, db, opener, budget)
             counts = paint(buckets, token, db, opener, budget, known=known)
+            counts["column_problems"] = column_problems
         dupes = {}
         seen = len(existing_rows(token, db, opener, dupes_out=dupes, budget=budget))
         return counts, dupes, seen
@@ -931,6 +945,8 @@ def collect(now, sources: dict, opener=None, token_file=None, db_file=None,
             f"{counts['kept']} kept (source quiet), {counts['pinned']} yours (untouched), "
             f"{counts['held']} yours (source stopped reporting it), "
             "read-back ok")
+    for problem in counts.get("column_problems") or ():
+        line += f"; could not add a column: {problem}"
     if counts["dropped_columns"]:
         # NEVER SILENT. The filter stops a missing column aborting the paint; it must
         # not also hide that the rows went out without it. A board missing `Link` wrote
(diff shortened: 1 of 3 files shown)


assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-332-20260907-225728.md` (1164 bytes).

```
FINDINGS:
major|ensure_columns PATCHes new columns onto a database that _refuse_without_identity then rejects as not this module's board|q-system/.q-system/scripts/board_rows.py:861
major|The four new morning-line tests delete the PYTEST_CURRENT_TEST guard and send authenticated live PATCH requests to api.notion.com on every suite run|q-system/.q-system/tests/test_board_is_actionable.py:546
minor|ensure_columns keys on name only, so a Link column present under the wrong type is never healed and the daily cannot-take sentence persists forever|q-system/.q-system/scripts/board_rows.py:620
minor|The PR corrects one copy of the removed-producer claim and leaves two present-tense copies, one gating an archive-authority guard whose false branch is unreachable|q-system/.q-system/scripts/consulting_board.py:838
END FINDINGS
```

--- reviewer output ---

Full suite finished: `1370 passed, 1 skipped in 212.99s`, exit 0. The PR's Proof line is accurate.

That confirms the last open item in my "What is sound" section and changes nothing else. Findings and verdict stand.

FINDINGS:
major|ensure_columns PATCHes new columns onto a database that _refuse_without_identity then rejects as not this module's board|q-system/.q-system/scripts/board_rows.py:861
major|The four new morning-line tests delete the PYTEST_CURRENT_TEST guard and send authenticated live PATCH requests to api.notion.com on every suite run|q-system/.q-system/tests/test_board_is_actionable.py:546
minor|ensure_columns keys on name only, so a Link column present under the wrong type is never healed and the daily cannot-take sentence persists forever|q-system/.q-system/scripts/board_rows.py:620
minor|The PR corrects one copy of the removed-producer claim and leaves two present-tense copies, one gating an archive-authority guard whose false branch is unreachable|q-system/.q-system/scripts/consulting_board.py:838
END FINDINGS

**VERDICT: REQUEST CHANGES.** Fix first: move the `ensure_columns` call below `_refuse_without_identity` in `work()`.

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-332-20260907-231542.md` (7335 bytes).

```
FINDINGS:
major|New collect-level tests acquire the production flock ~/.config/kipi/board-rows.lock, so the suite and the live 07:40 painter can each block the other|q-system/.q-system/tests/test_board_is_actionable.py:561
minor|A Notion refusal of the schema PATCH is swallowed by a bare except and reported with the identical unactionable sentence used for a genuinely absent column|q-system/.q-system/scripts/board_rows.py:643
minor|ensure_columns returns a synthesized schema instead of the one Notion returned, so a create that did not take effect loses the whole morning's paint to a raw 400|q-system/.q-system/scripts/board_rows.py:645
END FINDINGS
```

--- reviewer output ---

Read the PR, ran the suite, mutation-tested all eight new guards, and drove `collect` end to end against a fake Notion. Findings below, each with the command I ran and its real output.

## Findings

### 1. major — the new tests take the production flock, so the suite and the 07:40 painter can each stop the other

`q-system/.q-system/tests/test_board_is_actionable.py:561` (the `_line` helper) and `:648` (`test_the_runner_checks_identity_before_it_heals`) reach `br.collect`, which enters `exclusive()`. Neither patches `br.LOCK_FILE`, so both take `~/.config/kipi/board-rows.lock`, the same lock `com.kipi.morning-brief` holds. The sibling file `test_consulting_board.py` patches it at lines 632, 664, 1140, 1358, so the convention exists and these five new tests miss it.

Reproducer, holding the real lock the way a concurrent painter does:

```
python3 - <<'PY'
import fcntl, subprocess, sys
from pathlib import Path
p = Path.home()/'.config'/'kipi'/'board-rows.lock'
h = open(p, 'w'); fcntl.flock(h, fcntl.LOCK_EX | fcntl.LOCK_NB)
r = subprocess.run([sys.executable,"-m","pytest",
  "q-system/.q-system/tests/test_board_is_actionable.py","-q","--no-header"],
print(r.stdout[-1200:])
PY
```

Real output:

```
FAILED ...::TestTheSchemaGuardIsActuallyWiredIntoTheRun::test_the_runner_hands_paint_the_schema_it_read
FAILED ...::TestTheIdentityRefusalFiresBeforeTheFirstQuery::test_the_runner_refuses_before_it_queries
FAILED ...::TestTheMorningLineSaysWhatHappened::test_the_columns_it_could_not_write_are_named
FAILED ...::TestTheMorningLineSaysWhatHappened::test_one_dropped_column_is_singular
FAILED ...::TestTheMorningLineSaysWhatHappened::test_a_complete_board_says_nothing_about_columns
FAILED ...::TestTheMorningLineSaysWhatHappened::test_held_rows_are_reported_on_the_line
FAILED ...::TestTheBoardHealsItsOwnOptionalColumns::test_the_runner_checks_identity_before_it_heals
7 failed, 51 passed in 0.11s
```

Baseline is `58 passed in 0.10s`. Two of the seven pre-date this PR; five arrive with it.

The failure message blames the wrong thing. `test_the_runner_checks_identity_before_it_heals` reports `AssertionError: [] == ['identity', 'heal']`, which reads as "the ordering guard broke", not "a lock was held".

The other direction, confirmed separately: the suite really acquires the live lock.

```
lock exists before: True mtime_ns: 1788848344134489942
lock exists after : True mtime_ns: 1788848364988308789
THE SUITE TOOK THE LIVE PAINTER'S LOCK: True
```

At 07:40 the loser is whichever started second. If the painter loses, `collect` returns "board not written: another painter holds the board lock" and there is no board that morning, naming a painter that was actually a test run. The fix is the one line the sibling file already uses: `monkeypatch.setattr(br, "LOCK_FILE", tmp_path / "board.lock")`.

### 2. minor — a Notion refusal is swallowed and reported as if the column were absent

`q-system/.q-system/scripts/board_rows.py:643`, `except Exception: return known`. A 403 (integration lacks update capability) and a genuinely missing column produce byte-identical founder output, every morning, forever, and the PATCH is re-attempted each run with nothing recording why it failed. The two need opposite fixes: grant the integration update access, or add the column by hand.

Reproducer drove `collect` with an opener that raises `HTTPError(403)` on `PATCH /databases`, then the same scenario against `bffa124f` (pre-PR) with the column genuinely absent:

```
== C. Notion refuses the schema PATCH (403)
   line: board: 1 new, ... read-back ok; this board cannot take the Link, Next columns, so those values were not written

PRE-PR (bffa124f), board genuinely missing Link:
    board: 1 new, ... read-back ok; this board cannot take the Link, Next columns, so those values were not written
```

Identical strings. `coding-standards.md` also says never silently swallow exceptions.

### 3. minor — the healed schema is a claim, not a read-back (constructed input, no producer emits it)

`board_rows.py:645` returns `dict(known, **missing)` instead of reading `properties` off the PATCH response Notion already sent back. If the create did not take effect, `paint` writes `Link` on the first row, takes a 400, and the whole morning's paint is lost. Dropping this a level because I had to build a Notion that returns 200 on the PATCH and then rejects the property; nothing in this repo emits that.

```
== B. PATCH 200 but Notion still rejects Link on the row (400)
   err : board write failed: HTTPError: HTTP Error 400: Link is not a property that exists
   line: None
```

Line 923 of this same file says a PATCH returning 200 is not proof the board holds what we think.

## What is sound

Attacks that did not land, all run:

- **All eight mutants died.** Swapping the identity/heal order, deleting the `ensure_columns` call, making a refusal claim success, making `ensure_columns` never PATCH, deleting the dropped-columns sentence, deleting `held` from the line, dropping `Next` from `CREATABLE`, adding `Item id` to `CREATABLE`. Every one turned at least one new test red. These tests are not decoration.
- **No test reaches the network.** I checked every `delenv("PYTEST_CURRENT_TEST")` site in the tree. The two in `test_consulting_board.py` that reach `collect` pass a fake opener whose `/databases` response has no `properties` key, so `_schema_properties` returns None and `ensure_columns` short-circuits before any request. The PR's `_request` tripwire is a real guard against a future call site, not a no-op.
- **The feature composes end to end** with the real (unstubbed) `ensure_columns`, which no test covers: GET schema, PATCH with exactly `Link` and `Next`, then a page create carrying both. The `{"url": {}}` and `{"rich_text": {}}` shapes match `WRITES_TYPE` and match what `_properties` actually writes.
- **The wrong-type message does what it claims:** "this board cannot take the Link (it is rich_text, this writes url) column".
- **`CREATABLE` correctly excludes the identity columns**, and `_refuse_without_identity` still fires before the first query.

One dead detail worth deleting while you are in there: `_only_known:599` sorts by name, then `paint:831` re-sorts the same set by display string. The inner sort has no effect.

## VERDICT: REQUEST CHANGES

Not a block. Nothing here is permanent or unrecoverable, and the healing feature itself is correct.

**Fix first:** patch `br.LOCK_FILE` to `tmp_path` in the five new collect-level tests. A suite that goes red because of machine-global state outside the repo is the false-alarm class, and it can suppress a real morning paint.

FINDINGS:
major|New collect-level tests acquire the production flock ~/.config/kipi/board-rows.lock, so the suite and the live 07:40 painter can each block the other|q-system/.q-system/tests/test_board_is_actionable.py:561
minor|A Notion refusal of the schema PATCH is swallowed by a bare except and reported with the identical unactionable sentence used for a genuinely absent column|q-system/.q-system/scripts/board_rows.py:643
minor|ensure_columns returns a synthesized schema instead of the one Notion returned, so a create that did not take effect loses the whole morning's paint to a raw 400|q-system/.q-system/scripts/board_rows.py:645
END FINDINGS

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-332-20260907-233203.md` (8748 bytes).

```
FINDINGS:
major|CREATABLE omits Notes and Bucket, so a board ensure_columns "heals" still never archives a row and puts none of them in his views, while reporting no problem|q-system/.q-system/scripts/board_rows.py:612
minor|The morning line's failed-column-creation sentence is untested; deleting it leaves the whole suite green because every test that reaches it stubs ensure_columns to return no problems|q-system/.q-system/scripts/board_rows.py:948
END FINDINGS
```

--- reviewer output ---

# Review: PR #332 (ASK-1346)

Two findings survived reproduction. Two attacks failed and are dropped.

---

## Finding 1 — major

**`CREATABLE` heals two of the eight droppable columns, and the six it leaves behind include `Notes`, which carries the archive authority. A board this fix "heals" still never archives a row, and `ensure_columns` reports no problem.**

`q-system/.q-system/scripts/board_rows.py:612`

`_properties` writes ten columns. `UNDROPPABLE` protects two. That leaves eight that can be dropped by `_only_known` and reported as "this board cannot take..." forever. `CREATABLE` covers `Link` and `Next`. `Notes`, `Bucket`, `Domain`, `Priority`, `Source`, `Status` stay unhealed.

`Notes` is not decoration. It carries `scope=`, `bucket=` and `pinned=`, which is where `_scope_of` and `_bucket_decision` read from. `Bucket` is the select the founder's three board views filter on.

Reproducer (run from the worktree root):

```
python3 - <<'PYEOF'
import sys
sys.path.insert(0, "q-system/.q-system/scripts")
import board_rows as br
FRESH = dict(Task="title"); FRESH["Item id"] = "rich_text"
sent = []
def fake_request(tok, method, path, body=None, op=None, bud=None):
    sent.append((method, path, body))
    if method == "PATCH" and path.startswith("/databases/"):
        after = dict(FRESH)
        for n, spec in body["properties"].items():
            after[n] = list(spec)[0]
        return dict(properties={n: dict(type=t) for n, t in after.items()})
    return dict(id="p1")
br._request = fake_request
healed, problems = br.ensure_columns(dict(FRESH), "tok", "db")
print("healed schema :", sorted(healed))
print("STILL MISSING :", sorted(set(br.WRITES_TYPE) - set(healed)))
print("problems      :", problems)
IID = "cb:deadbeefdeadbeef"
props = dict(); props["Item id"] = dict(rich_text=[dict(plain_text=IID)])
have = dict(); have[IID] = dict(id="pg1", properties=props)
br.existing_rows = lambda *a, **k: have
counts = br.paint(dict(top_of_mind=[], this_week=[], inbox=[],
                       healthy_scopes=set(("card","mail"))), "tok", "db", known=healed)
print("archived      :", counts["archived"], "| kept:", counts["kept"])
print("archive calls :", [c for c in sent if c[0]=="PATCH" and "/pages/" in c[1]])
PYEOF
```

Real output:

```
healed schema : ['Item id', 'Link', 'Next', 'Task']
STILL MISSING : ['Bucket', 'Domain', 'Notes', 'Priority', 'Source', 'Status']
problems      : ()
archived      : 0 | kept: 1
archive calls : []
```

A row whose producer stopped emitting it, inside a scope the run declared healthy, is not archived. It is counted `kept`, so the read-back sum still balances and nothing reports a mismatch. The board only grows, which is the exact scar the module docstring records ("the last board became 4 stale rows nobody trusted"). Same board, `Bucket` absent, puts every row in none of his three views.

`ensure_columns` returns `problems == ()` while doing this. The morning line does still name the dropped columns through the separate `dropped_columns` path, so it is not fully silent, and the founder's own live board already has `Notes`, so today's board is unaffected. What is affected is every other instance, which is the population this PR exists for. The PR body's own argument, "a permanent explanation is not a fix", applies unchanged to six of the eight.

Caveat, stated plainly: the schema I fed it is constructed, not producer output. The PR body itself attests that this population exists ("It had been added by hand, once, on the founder's own board. Anywhere else...").

At 3am this does nothing violent. What a human does the next morning: open the Notion board and hand-add `Notes` and `Bucket`, or this lands as `CREATABLE = ("Link", "Next", "Notes", "Bucket", "Domain", "Priority", "Source")`. `Status` is the one that needs a decision rather than a line: Notion's own built-in Status column is type `status`, not `select`, so creating it as `select` is a choice worth making on purpose.

---

## Finding 2 — minor

**The operator-facing sentence that reports a failed column creation is untested. Deleting it leaves the full suite green.**

`q-system/.q-system/scripts/board_rows.py:948`

`ensure_columns` can return a problem string. Three unit tests prove that. Nothing proves it ever reaches the line the founder reads. The only tests that reach `collect`'s line builder stub `ensure_columns` to return `()`, so the true side of that loop is unreachable in the suite.

Reproducer, deleting the two lines and running the board suites against the mutant:

```
python3 - <<'PYEOF'
import sys, types, pathlib, pytest
p = pathlib.Path("q-system/.q-system/scripts/board_rows.py")
src = p.read_text()
target = ('    for problem in counts.get("column_problems") or ():\n'
          '        line += f"; could not add a column: ' + chr(123) + 'problem' + chr(125) + '"\n')
assert target in src
sys.path.insert(0, "q-system/.q-system/scripts")
mod = types.ModuleType("board_rows"); mod.__file__ = str(p)
exec(compile(src.replace(target, ""), str(p), "exec"), mod.__dict__)
sys.modules["board_rows"] = mod
sys.exit(pytest.main(["q-system/.q-system/tests/test_board_is_actionable.py",
                      "q-system/.q-system/tests/test_consulting_board.py",
                      "-q", "-p", "no:cacheprovider"]))
PYEOF
```

Real output:

```
MUTANT LOADED: the column-problem sentence is deleted from the morning line
194 passed in 1.02s
```

The harness is not lying to me. Negative control, same injection with `CREATABLE = ("Link",)`:

```
E       AssertionError: {'properties': {'Link': {'url': {}}}}
E       assert {'Link'} == {'Link', 'Next'}
FAILED ...::test_a_missing_column_is_created_and_then_usable
1 failed, 58 passed in 0.07s
```

Grep confirms it: `column_problems` and `could not add a column` appear nowhere under `q-system/.q-system/` outside `board_rows.py`. One case that stubs `ensure_columns` to return a problem and asserts the sentence lands closes it.

---

## What is sound

**"No test reaches Notion" is true, and I checked it the hard way.** Ran the entire suite with `http.client.HTTPConnection.connect` and `HTTPSConnection.connect` replaced by a raising stub. Result: `1372 passed, 1 skipped in 210.33s`. The new `ensure_columns` fires an authenticated PATCH, so this was the thing most worth breaking, and it does not break.

**The identity-before-heal ordering holds.** `_refuse_without_identity` runs on the pre-PATCH schema, so a database this module rejects as not-its-board is never reshaped. `test_the_runner_checks_identity_before_it_heals` goes red if the order flips.

**The `CREATABLE` contents are load-bearing in the test, not decorative.** Removing `Next` turned it red immediately (output above). That is a real check, not a green-by-construction one.

**Dropped: `ensure_columns`'s bare `except Exception` swallows `Cancelled`.** `Cancelled` is a `RuntimeError`, so the budget-expiry signal is caught and turned into an advisory string instead of propagating. I could not make it matter: `_Budget.check()` sets `cancelled` permanently, so `paint`'s very next call (`existing_rows`) re-raises and `collect` returns "board write timed out" exactly as before. Behaviour-neutral, so I am not filing it.

**Dropped: the read-back `existing_rows` runs outside `exclusive()`.** Real, and a second painter starting in that window would make the first report a false read-back mismatch. It is pre-existing code this PR does not touch, and I could not construct it without the lock stub, so it is not this PR's finding.

**The `consulting_board.py` half of the diff is comment-only and its claims check out.** `collect_mail` at `morning-brief.py:369` refuses any ledger row without a `thread_id`, so "no producer emits a fallback key today" is accurate, and the retained guard is honestly labelled as currently-unreachable rather than described in the present tense.

---

## VERDICT: REQUEST CHANGES

Not a block. Nothing here is permanent or unrecoverable, and the change is a real improvement over what #327 left.

**Fix first:** the `CREATABLE` list at `board_rows.py:612`. Adding `Link` and `Next` while leaving `Notes` out means a healed board still cannot archive anything, and `ensure_columns` says it healed fine.

FINDINGS:
major|CREATABLE omits Notes and Bucket, so a board ensure_columns "heals" still never archives a row and puts none of them in his views, while reporting no problem|q-system/.q-system/scripts/board_rows.py:612
minor|The morning line's failed-column-creation sentence is untested; deleting it leaves the whole suite green because every test that reaches it stubs ensure_columns to return no problems|q-system/.q-system/scripts/board_rows.py:948
END FINDINGS

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-332-20260907-235303.md` (6677 bytes).

```
FINDINGS:
major|A successful schema PATCH is never reported on the morning line, and a column the founder deletes is silently re-created every run with no opt-out|q-system/.q-system/scripts/board_rows.py:954
minor|test_one_dropped_column_is_singular cannot fail because "Link columns" contains "Link column", so the plural branch survives mutation|q-system/.q-system/tests/test_board_is_actionable.py:606
minor|The refused-creation message prints only the HTTP status line and discards Notion's explanatory body, which is the reason the branch exists|q-system/.q-system/scripts/board_rows.py:654
END FINDINGS
```

--- reviewer output ---

Review of PR #332. Read-only, nothing in the repo was modified.

## Findings

### 1. major — a successful schema PATCH is never reported, and a column the founder deletes is silently re-created on every run with no opt-out

`q-system/.q-system/scripts/board_rows.py:954`

The morning line reports `column_problems` only on failure. On success `ensure_columns` PATCHes up to 8 properties onto his live Notion database and the line says nothing. There is no env off-switch (the module reads only `PYTEST_CURRENT_TEST` and `KIPI_BRIEF_DRY_RUN`), so a column he removes on purpose comes back every 07:40, forever, unannounced. This is the file's own `NEVER SILENT` doctrine at line 957 applied to row writes but not to the strictly larger schema write.

Reproducer (heredoc, run as shown):

```
python3 - <<'PY'
import sys; S="<repo>/q-system/.q-system/scripts"; sys.path.insert(0,S)
import board_rows as br
runs=[]
def fake(tok,m,p,body=None,op=None,bud=None):
    runs.append(sorted((body or dict()).get("properties",dict())))
    return dict(properties=dict((n,dict(type=t)) for n,t in br.WRITES_TYPE.items()))
br._request=fake
board=dict(br.WRITES_TYPE); del board["Domain"]
for morning in range(3):
    out,problems=br.ensure_columns(dict(board),"tok","db")
    print("morning",morning,"-> PATCHed",runs[-1],"problems:",problems)
    board=dict(out); del board["Domain"]
PY
```

Real output:

```
morning 0 -> PATCHed ['Domain'] problems: ()
morning 1 -> PATCHed ['Domain'] problems: ()
morning 2 -> PATCHed ['Domain'] problems: ()
```

And the end-to-end line after a heal of 8 columns:

```
REPRO1 patched: PATCH /databases/db ['Bucket','Domain','Link','Next','Notes','Priority','Source','Status']
REPRO1 problems reported to the caller: ()
REPRO1 morning line: ["board: ... read-back ok"]
REPRO1 line mentions the 8 columns it just added? False
```

Recoverable (deleting a Notion property is a manual undo), so not a blocker. One sentence on the line when `ensure_columns` creates something closes it.

### 2. minor — `test_one_dropped_column_is_singular` cannot fail; the plural branch it exists to pin is unpinned

`q-system/.q-system/tests/test_board_is_actionable.py:606`

The assertion is `"Link column" in line`, and `"Link columns"` contains `"Link column"` as a substring. Mutating the singular branch to always emit `" columns"` leaves all 63 cases green.

Reproducer: in-memory mutation of `board_rows.py`, injected into `sys.modules` before pytest imports the test file. Real output of the matrix:

```
none                     rc=0 | 63 passed in 0.06s
order                    rc=1 | 1 failed, 62 passed
halfheal                 rc=1 | 2 failed, 61 passed
synthesise               rc=1 | 1 failed, 62 passed
noline                   rc=1 | 1 failed, 62 passed
silentrefusal            rc=1 | 1 failed, 62 passed
noannot                  rc=1 | 1 failed, 62 passed
drop-held-phrase         rc=1 | 1 failed, 62 passed
drop-dropped-sentence    rc=1 | 1 failed, 62 passed
always-plural            rc=0 | 63 passed in 0.06s
```

Direct confirmation of why:

```
mutant line: board: ...; this board cannot take the Link columns, so those values were not written
'Link column' in line: True
```

`assert line.endswith("Link column, so those values were not written")` turns it red.

### 3. minor — the refusal message discards Notion's reason, which is the whole point of that branch

`q-system/.q-system/scripts/board_rows.py:654`

The comment says a refusal "is about permissions or a token. The reason is carried back and said." `str(HTTPError)` is the status line only; Notion's `message` body is thrown away.

Reproducer output:

```
what the founder reads on the morning line:
  ; could not add a column: HTTPError: HTTP Error 403: Forbidden
Notion's actual reason, discarded: {"object":"error","status":403,"code":"restricted_resource","message":"Insufficient permissions: the integration lacks u
```

Reading `exc.read()` for `HTTPError` before truncating to 160 gives the founder the sentence he can act on.

## What is sound

- **Mutation matrix on the new logic.** Six mutants of `ensure_columns` and its call site, all killed: heal-before-identity, `CREATABLE = ("Link","Next")`, synthesising the schema instead of reading Notion's answer, deleting the problem sentence, swallowing the refusal, dropping the wrong-type annotation. The `halfheal` mutant kills two cases. These tests have teeth.
- **The comment corrections in `consulting_board.py` are true.** `morning-brief.py:372` builds `mail:<thread_id>` and `collect_mail` returns an error rather than a row when `thread_id` is empty, so the claim that no producer emits the `sender|subject` shape checks out against the producer, not against a doc.
- **Test isolation.** The added `LOCK_FILE` redirect and the raising `_request` stub do what they claim; no test in the file reaches the network or the production flock. `test_consulting_board.py`'s three `collect` callers exit at the dry-run or pytest guard before `_credentials` matters.
- **No false-success path from a spent budget.** `ensure_columns` catches `Cancelled` broadly, which looked like a way to get a clean "read-back ok" after the brief timed out. It is not: `paint` raises before any counts are produced, so `collect` still returns the timeout error.
- **Full suite:** `python3 -m pytest q-system/.q-system/tests/ -q` gave `1376 passed, 1 skipped in 210.98s`.
- **Dropped after I could not reproduce it:** the `{"select": {}}` property-creation payload (needs a live Notion call to falsify), and the `AttributeError` escape when a PATCH response has a non-dict property object (same pattern already ships in `_schema_properties`, and no producer emits that shape).

## VERDICT: REQUEST CHANGES

Not BLOCK. Nothing here is unrecoverable; an added Notion property can be deleted.

Fix first: say it on the morning line when `ensure_columns` creates a column. Right now the unattended job reshapes his board and the only thing it ever reports is the case where it failed to.

FINDINGS:
major|A successful schema PATCH is never reported on the morning line, and a column the founder deletes is silently re-created every run with no opt-out|q-system/.q-system/scripts/board_rows.py:954
minor|test_one_dropped_column_is_singular cannot fail because "Link columns" contains "Link column", so the plural branch survives mutation|q-system/.q-system/tests/test_board_is_actionable.py:606
minor|The refused-creation message prints only the HTTP status line and discards Notion's explanatory body, which is the reason the branch exists|q-system/.q-system/scripts/board_rows.py:654
END FINDINGS

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-332-20260908-002450.md` (6713 bytes).

```
FINDINGS:
minor|`_columns_made` lacks the non-dict guard `_remember_columns` has, so a corrupt ledger raises AttributeError that no `collect` handler catches and the board section fails every run|q-system/.q-system/scripts/board_rows.py:628
minor|A created-then-deleted column emits its "your removal" note on every run forever and is named a second time by the dropped-columns sentence on the same 927-char line|q-system/.q-system/scripts/board_rows.py:683
minor|The "added the X columns to this board" note is dropped on the read-back-mismatch and duplicate-row exits, so a schema change to the founder's Notion database is never reported|q-system/.q-system/scripts/board_rows.py:1013
END FINDINGS
```

--- reviewer output ---

Read-only review of PR #332, done. Three findings, all reproduced; nothing survived as a blocker or major.

## Findings

### 1. minor — `_columns_made` is missing the `isinstance(data, dict)` guard its sibling writer has, so a non-dict ledger crashes the whole board section

`q-system/.q-system/scripts/board_rows.py:628`. `_remember_columns` (line 642) explicitly coerces a non-dict to `{}`. `_columns_made` does not, and it catches only `OSError, ValueError`. `AttributeError` is in none of `collect`'s handlers (lines 974-989), so it escapes.

Repro:

```
python3 -c "... br.ensure_columns(dict(OLD), 'tok', 'db', record=rec) for several bodies"
```

Real output:

```
record='["Link"]'   -> RAISED  AttributeError: 'list' object has no attribute 'get'
record='"Link"'     -> RAISED  AttributeError: 'str' object has no attribute 'get'
record='3'          -> RAISED  AttributeError: 'int' object has no attribute 'get'
record='not json'   -> OK      ('the schema PATCH returned no properties, ...',)
record=''           -> OK      ('the schema PATCH returned no properties, ...',)
```

End to end through `collect`:

```
collect() RAISED AttributeError: 'list' object has no attribute 'get'
morning-brief _guarded turns this into: board_rows failed (AttributeError)
```

Blast radius: no board paint, every run, until someone hand-edits `~/.config/kipi/board-columns-created.json`, and the founder-facing text is a class name with no remediation. **Severity dropped a level and I am saying so:** no producer in this repo writes a non-dict there. `_remember_columns` always writes a dict, and a truncated write gives invalid JSON, which is already caught. I had to construct the input.

### 2. minor — the "your removal" note is permanent and duplicated, which is the shape the PR body says it is fixing

`q-system/.q-system/scripts/board_rows.py:683`. Once a column is created and then deleted, `told` fires on every run with no expiry, and the same column is named a second time on the same line by the `dropped_columns` sentence at line 1025.

Repro output, running `ensure_columns` four times against a ledger holding 8 created columns and a board where 6 were deleted:

```
run 2 -> 6 notes
run 3 -> 6 notes
run 4 -> 6 notes
```

Driving the real `collect` line builder in that state:

```
board: 0 new, 0 refreshed, 0 unchanged, 0 rebucketed, 0 cleared, 0 kept (source quiet), 0 yours (untouched), 0 yours (source stopped reporting it), read-back ok; Bucket was created here before and is gone now, so it is treated as your removal and is not being put back; Domain was created here before ... ; this board cannot take the Bucket, Domain, Notes, Priority, Source, Status columns, so those values were not written

length: 927 chars
```

927 characters, six clauses repeated twice, in the daily brief, forever. The PR body's own line is "A permanent explanation is not a fix." Nothing in the message says the ledger file is what clears it. Calibration check: nobody has to do anything at 3am, so this is not a major.

### 3. minor — "added the X columns to this board" is discarded on the read-back-mismatch and duplicate-row exits

`q-system/.q-system/scripts/board_rows.py:1013`. The note loop sits after the early returns at 991-994 and 1001-1006. The docstring at line 716 calls this property "SAID OUT LOUD (round 4, major)".

Repro: fresh board, 8 columns created, read-back sees 2 rows where 3 were written.

```
rows = []
err  = read-back mismatch: expected 3 row(s) (3 written + 0 kept from a quiet source + 0 held for you), board shows 2
record file proves 8 columns WERE added to his board this run:
  {  "DB1": [   "Bucket",   "Domain",   "Link",   "Next",   "Notes",   "Priority",   "Source",   "Status"  ] }
founder-facing text mentions the schema change? False
```

Eight columns appear on his Notion database and the brief never says so.

## What is sound

- **Mutation on the new logic. Four mutants, four kills.** Removing the deleted-column guard, shrinking `CREATABLE` back to Link and Next, deleting the `column_problems` line append, and believing an unconfirmed create each turned the suite red (1, 2, 1, 1 failures). The new tests are load-bearing, not decoration.
- **Fixture shape against the real producer.** `_notion_patch` returns `properties[name].type`, which is exactly what `_schema_properties` already parses off the live API, and the PATCH bodies (`url: {}`, `select: {}`, `multi_select: {}`, `rich_text: {}`) match Notion's property-create shapes.
- **Identity before heal.** `test_the_runner_checks_identity_before_it_heals` records call order and would go red if flipped. The wrong-type identity refusal that the `names`-vs-`shown` split protects is already pinned by the `Item id: select` case at test line 391.
- **Test isolation.** After running the suite, `~/.config/kipi/board-columns-created.json` does not exist, and every `collect`-reaching test monkeypatches `LOCK_FILE` and a raising `_request`. The production lock file on this machine predates the fix, which is what the fix is for.
- **Concurrency.** `ensure_columns` and the ledger read-modify-write both run inside `exclusive()`, so two painters cannot interleave the JSON rewrite.
- **Recoverability.** The only outward mutation is adding Notion columns, which is undoable by deleting them, and the module then honours the deletion.
- **Dropped finding:** I tried to make `except Exception` swallowing `Cancelled` in `ensure_columns` matter. It does not. `paint` or `existing_rows` re-raises on the next call and `collect` still returns the timeout message. Could not reproduce a wrong outcome, so I dropped it.

## VERDICT

**APPROVE WITH NITS.** No finding reached major.

Fix first: finding 2. The permanent duplicated sentence lands in the founder's daily message, and it is the exact failure class this PR was written to remove. One clean shape would be to say the removal once, and let the existing `dropped_columns` sentence carry it from then on.

FINDINGS:
minor|`_columns_made` lacks the non-dict guard `_remember_columns` has, so a corrupt ledger raises AttributeError that no `collect` handler catches and the board section fails every run|q-system/.q-system/scripts/board_rows.py:628
minor|A created-then-deleted column emits its "your removal" note on every run forever and is named a second time by the dropped-columns sentence on the same 927-char line|q-system/.q-system/scripts/board_rows.py:683
minor|The "added the X columns to this board" note is dropped on the read-back-mismatch and duplicate-row exits, so a schema change to the founder's Notion database is never reported|q-system/.q-system/scripts/board_rows.py:1013
END FINDINGS

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-332-20260908-004240.md` (6303 bytes).

```
FINDINGS:
major|The deleted-column filter suppresses the dropped-columns warning for Notes and Bucket, so archiving silently stops while the line says read-back ok|q-system/.q-system/scripts/board_rows.py:1040
minor|_chosen_absent is a never-cleared module global unioned across every board id, making the report cross-contaminate and the suite order-dependent|q-system/.q-system/scripts/board_rows.py:627
minor|The "your removal" sentence is re-emitted on the morning line every run forever with no way to stop it|q-system/.q-system/scripts/board_rows.py:705
END FINDINGS
```

--- reviewer output ---

Review done. Full read of the diff, both changed source files, and three executed reproducers.

## Finding 1 — major

**Claim:** the new "his choice wins" filter suppresses the dropped-columns warning for every column, including `Notes` and `Bucket`, whose absence silently kills archiving and bucketing.

**Where:** `q-system/.q-system/scripts/board_rows.py:1040`

The chain: the module creates `Notes` on a fresh board and records it. He deletes it. From then on `ensure_columns` correctly declines to re-create it, but `work()` also strips `Notes` out of `counts["dropped_columns"]`, so the sentence "this board cannot take the Notes column, so those values were not written" never fires again. `_scope_of` reads scope out of `Notes`, an empty scope is never healthy, and the archive loop at line 892 counts every stale row as `kept` and moves on. The board grows forever, the line says `0 cleared` and `read-back ok`, and the founder is told only that his removal was respected.

**Reproducer** (record says the module made `Notes`, board no longer has it):

```
python3 - <<'PYEOF'   # full script in the transcript; stubs paint/existing_rows, no network
...
PYEOF
```

**Real output:**

```
  _only_known dropped: ('Notes',)
ERR : None
LINE: board: ... 0 cleared, 4 kept (source quiet), 0 yours (untouched),
      0 yours (source stopped reporting it), read-back ok; Notes was created here
      before and is gone now, so it is treated as your removal and is not being put back

says the board cannot take Notes: False
scope of a row with no Notes    : ''
```

The module's own comment at line 611 names this exact consequence: "a board missing `Notes` cannot carry the `scope=` line, so `_scope_of` reads unknown and no row is ever archived". The PR builds the suppression anyway. Rows counted `kept (source quiet)` when the source answered fine is the "reports success for work that did not happen" shape.

## Finding 2 — minor

**Claim:** `_chosen_absent` is module-global, never cleared, and unioned across every board id, so one board's state suppresses another's report and the suite is order-dependent.

**Where:** `q-system/.q-system/scripts/board_rows.py:627` and `:1040`

**Reproducer:**

```
python3 -m pytest ".../test_board_is_actionable.py::TestTheBoardHealsItsOwnOptionalColumns::test_a_column_he_deleted_is_not_put_back" ".../TestTheMorningLineSaysWhatHappened::test_the_columns_it_could_not_write_are_named" -q
```

**Real output:**

```
.F
E  AssertionError: board: 1 new, ... read-back ok; this board cannot take the Next
   column, so those values were not written
1 failed, 1 passed in 0.04s
```

The heals test sets `_chosen_absent["db"] = {"Link"}`; the morning-line test paints board `"d"` and loses `Link` from its report. Both files pass in declaration order, so the suite is green today and no random-order plugin is installed. Production runs one board per process, so the cross-board union is not reachable there yet: dropped a level for that. `grep -rn "_chosen_absent" tests/` returns nothing, so this filter has no test of its own.

## Finding 3 — minor

**Claim:** the "your removal" sentence is re-emitted on the morning line on every run, forever, with no way to stop it.

**Where:** `q-system/.q-system/scripts/board_rows.py:705`

Same output as finding 1: the line carries `Notes was created here before and is gone now, so it is treated as your removal and is not being put back` on run N, and `removed` is recomputed from a record file that is never pruned. That is the shape the PR body condemns in `#327` ("reported the same sentence every morning forever"), rebuilt one level down. Say it once on the run after the deletion, or drop the name from the record when it is honored.

## What is sound

- **The HTTPError body fixture.** My top suspicion: a hand-built `HTTPError` whose body is readable, when production might hand a consumed one. `_request` at line 166 re-raises without touching `exc.read()`, so `exc.read()` in `ensure_columns` gets a live fp. Fixture shape matches the real producer.
- **Network from the suite.** I removed `PYTEST_CURRENT_TEST` and looked for a live path: every new test replaces `_request` with a raiser and repoints `LOCK_FILE` and the record file. Nothing reaches `api.notion.com` or the production lock.
- **`CREATABLE` drift.** Derived from `WRITES_TYPE`, and the exact-set test is written out by hand, so adding a column to `WRITES_TYPE` and forgetting it turns red rather than silently half-healing.
- **Identity before heal.** Verified in code at line 986, and `test_the_runner_checks_identity_before_it_heals` asserts `order == ["identity", "heal"]`, which goes red if reversed.
- **The unconfirmed-create branch.** The code refuses to synthesize `dict(known, **missing)` and reads the PATCH response instead; the test drives `{"properties": {}}` and can fail.
- **Corrupt record file.** I tried to crash `collect` through `_columns_made`; the six cases cover list, string, non-list entry, broken JSON, absent file, and the good read.
- **Two painters PATCHing the schema at once.** `ensure_columns` runs inside `exclusive()`, so the flock covers it.

Baseline: `test_board_is_actionable.py` + `test_consulting_board.py` = 207 passed. The full 1370-case run was still going when I finished; I did not confirm the author's number myself.

## VERDICT: REQUEST CHANGES

Not a BLOCK. Nothing here is unrecoverable: the columns can be deleted in Notion and archiving resumes the moment `Notes` is restored.

**Fix first:** stop the `_chosen_absent` filter from hiding `Notes` and `Bucket`. A column whose absence changes behavior gets its consequence said out loud, even when he removed it himself.

FINDINGS:
major|The deleted-column filter suppresses the dropped-columns warning for Notes and Bucket, so archiving silently stops while the line says read-back ok|q-system/.q-system/scripts/board_rows.py:1040
minor|_chosen_absent is a never-cleared module global unioned across every board id, making the report cross-contaminate and the suite order-dependent|q-system/.q-system/scripts/board_rows.py:627
minor|The "your removal" sentence is re-emitted on the morning line every run forever with no way to stop it|q-system/.q-system/scripts/board_rows.py:705
END FINDINGS

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-332-20260908-010038.md` (6209 bytes).

```
FINDINGS:
minor|A column the founder deleted is still named in the "cannot take" sentence on the morning line every run, and the code comment plus its test claim that path is silent|q-system/.q-system/scripts/board_rows.py:723
minor|The structural consequence of a missing Notes or Bucket column is reported only when he deleted it, never when the module tried to create it and Notion refused|q-system/.q-system/scripts/board_rows.py:723
minor|_remember_columns writes the live ~/.config/kipi record with no PYTEST_CURRENT_TEST chokepoint, unlike collect, so a test that omits record= teaches the 07:40 job he deleted columns|q-system/.q-system/scripts/board_rows.py:658
END FINDINGS
```

--- reviewer output ---

Review done. Read-only. Three findings, all minor, each reproduced.

---

## Finding 1 (minor) — a column he deleted is still named on the morning line every run, and the code says it is not

`q-system/.q-system/scripts/board_rows.py:723` (the `told` filter) plus `:1059` (the drop sentence).

`ensure_columns` correctly declines to re-create `Link`/`Next` after he deletes them, and stays silent about it. But `known` then has no `Link`, so `_only_known` still drops it and `collect`'s line builder still appends the "cannot take" sentence. Every morning, forever, with no way given to change it. That is the exact shape the PR body says it exists to kill.

The test at `test_board_is_actionable.py:759` asserts silence on `problems`, which is not the channel the sentence comes out of. It is green and the behavior it describes does not happen.

Reproducer, driving the real chain (`ensure_columns` then real `_properties` then real `_drop_and_note` then `collect`'s line builder), record file written in the shape `_remember_columns` itself produces:

```
python3 $T/repro_link_nag.py
```

Real output:

```
ensure_columns problems: ()
err: None
LINE: board: your book could not be read (...); 0 new, 0 refreshed, 0 unchanged,
0 rebucketed, 0 cleared, 0 kept (source quiet), 0 yours (untouched),
0 yours (source stopped reporting it), read-back ok;
this board cannot take the Link, Next columns, so those values were not written
AssertionError: FINDING: his deliberate deletion is named on the line every run
```

---

## Finding 2 (minor) — the structural consequence is explained only when he already knows

`q-system/.q-system/scripts/board_rows.py:723`

`STRUCTURAL_COST` is consulted only for `removed`, meaning columns this module created and he then deleted. When the module tries to create `Notes` and Notion refuses, the founder gets the raw 403 and the generic drop sentence, never "nothing is ever archived". The case where he needs the explanation most is the one that does not get it.

Reproducer, bare board plus a 403 on the schema PATCH:

```
python3 $T/repro_structural_silence.py
```

Real output:

```
problems: ('HTTPError: HTTP Error 403: Forbidden {"message": "no schema permission"}',)

names the 403           : True
names archiving is dead : False
names rows invisible    : False
AssertionError: FINDING: the structural consequence is explained only when HE deleted
the column, never when the module tried to create it and Notion refused
```

---

## Finding 3 (minor) — the new state writer has no test chokepoint, unlike every other live path in this file

`q-system/.q-system/scripts/board_rows.py:658`

`collect` refuses under `PYTEST_CURRENT_TEST` because the live board must never be written by a test. `_remember_columns` writes `~/.config/kipi/board-columns-created.json` with no such guard and a live default. Today every test passes `record=`; one that forgets teaches the 07:40 job that the founder deleted columns, and the module stops creating them.

Reproducer (default repointed at a temp file so the real record is untouched, proving only that `record=None` writes and is believed):

```
python3 $T/repro_no_test_guard.py
```

Real output:

```
production default : /Users/assafkipnis/.config/kipi/board-columns-created.json
under PYTEST_CURRENT_TEST, record=None wrote: {  "the-live-board-id": [   "Link",   "Notes"  ] }
and the module then believes those were deleted: {'Link', 'Notes'}
```

---

## What is sound

Attacks that did not land:

**The tests are load-bearing.** Four mutants, all killed. `74 passed` baseline, then `ignore_deletions` 4 failed, `naive_record_read` 3 failed, `trust_the_patch` 5 failed, `no_structural_warning` 2 failed. Run via `python3 $T/runmut.py`.

**Wrong-typed columns are not retyped.** `missing` keys on `name not in known`, never on a type mismatch, so a board carrying a native Notion `status` property named `Status` is left alone rather than converted. I looked for the data-destroying version of this heal and it is not there.

**Identity before heal is real, not just commented.** `_refuse_without_identity` runs at line 1005 ahead of `ensure_columns` at 1006, and `test_the_runner_checks_identity_before_it_heals` asserts the exact order list, which kills the swap.

**The `STRUCTURAL_COST["Notes"]` claim is true.** I checked line 911: `if _scope_of(page) not in healthy: continue`, so an unreadable scope skips archiving rather than archiving everything. The message is not lying in the dangerous direction.

**Concurrency on the record file.** `ensure_columns` runs inside `exclusive()`, the same flock the painter holds, so two painters cannot interleave the read-modify-write of `board-columns-created.json`.

**No hidden consumer of the display strings.** `_only_known` now returns annotated strings like `Link (it is rich_text, this writes url)`. I grepped the repo for `dropped_columns` consumers outside this file and its test. There are none, so those strings stay display-only and the `UNDROPPABLE` check still reasons about the plain `names` list.

**Rollback.** Adding Notion properties is reversible by deleting them, and the module now records what it created so it will not fight that deletion. Nothing here is permanent.

---

## VERDICT: APPROVE WITH NITS

Fix first: **Finding 1.** The one sentence the founder reads every morning still nags him about a column he removed on purpose, and the test that was supposed to hold that asserts on a value that never carried it. Assert the built line, not `problems`.

FINDINGS:
minor|A column the founder deleted is still named in the "cannot take" sentence on the morning line every run, and the code comment plus its test claim that path is silent|q-system/.q-system/scripts/board_rows.py:723
minor|The structural consequence of a missing Notes or Bucket column is reported only when he deleted it, never when the module tried to create it and Notion refused|q-system/.q-system/scripts/board_rows.py:723
minor|_remember_columns writes the live ~/.config/kipi record with no PYTEST_CURRENT_TEST chokepoint, unlike collect, so a test that omits record= teaches the 07:40 job he deleted columns|q-system/.q-system/scripts/board_rows.py:658
END FINDINGS

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-332-20260908-011723.md` (7567 bytes).

```
FINDINGS:
major|The live-record test asserts a file production writes, so the suite goes red on the founder's machine the first morning a column is healed, blaming the suite for the 07:40 job's write|q-system/.q-system/tests/test_board_is_actionable.py:911
minor|A PATCH response carrying properties but not the requested ones returns no problem at all, while the empty-properties and refusal paths both report the unconfirmed create and its structural cost|q-system/.q-system/scripts/board_rows.py:777
minor|_scope_of's docstring still says the design avoids changing the board schema, which ensure_columns now does deliberately with eight properties including four select types|q-system/.q-system/scripts/board_rows.py:270
END FINDINGS
```

--- reviewer output ---

## Review: PR #332 (ASK-1346)

One major, two minors. Suite is green today (1392 passed, 1 skipped, 210s) and stops being green the first morning after merge.

---

### MAJOR — the suite's own isolation test goes red when the feature works

`q-system/.q-system/tests/test_board_is_actionable.py:911`

`test_the_live_record_does_not_exist_after_this_suite` asserts `~/.config/kipi/board-columns-created.json` does not exist. Production writes exactly that file: `work()` calls `ensure_columns(known, token, db, opener, budget)` with no `record=`, so `_remember_columns` (`board_rows.py:663`) writes `COLUMNS_MADE`. The file is absent on this machine today only because the feature is unmerged. The first 07:40 run that heals a column creates it, and the assertion is false from then on.

The failure message names the wrong culprit: it says "the suite wrote ... which the 07:40 job reads". The 07:40 job wrote it. And the obvious remediation, deleting the file, silently re-arms the thing round 4 fixed: every column the founder deleted gets re-created the next morning.

Reproducer (fake HOME, real home untouched):

```
python3 -c "
import os, json, pathlib, subprocess, tempfile
fake = tempfile.mkdtemp(prefix='pr332home-')
(pathlib.Path(fake)/'.config'/'kipi').mkdir(parents=True)
(pathlib.Path(fake)/'.config'/'kipi'/'board-columns-created.json').write_text(
    json.dumps({'the-founders-board-id': ['Link','Next']}), 'utf-8')
r = subprocess.run(['python3','-m','pytest',
    'q-system/.q-system/tests/test_board_is_actionable.py','-q'],
    env=dict(os.environ, HOME=fake), capture_output=True, text=True)
print(r.stdout[-900:]); print('exit', r.returncode)"
```

Real output:

```
E       AssertionError: the suite wrote /var/folders/.../board-columns-created.json, which the 07:40 job reads
E       assert not True
q-system/.q-system/tests/test_board_is_actionable.py:912: AssertionError
1 failed, 78 passed in 0.11s
exit 1
```

CI stays green (fresh runner HOME), so this lands only on the founder's machine, on `verify.sh --full` and on `verify.sh --staged` whenever this file is touched. The check it wants is "no NEW live record appeared during this run", or point `COLUMNS_MADE` at a tmp path via a fixture and assert on `_remember_columns`'s guard alone, which `test_writing_without_a_record_path_raises_under_pytest` already does.

---

### MINOR — a PATCH that answers with properties but not the new ones reports nothing

`q-system/.q-system/scripts/board_rows.py:777`

`fresh` non-empty plus `made` empty falls through the `if made:` branch and returns `told + ()`. No "unconfirmed" sentence, no `STRUCTURAL_COST` line. The empty-`properties` path (line 787) and the exception path (line 771) both emit the cost. So the branch that says "the consequence belongs to the absence, not to the reason" has a third absence it does not cover.

```
python3 -c "
import sys, tempfile, pathlib; sys.path.insert(0,'q-system/.q-system/scripts')
import board_rows as br
tmp = pathlib.Path(tempfile.mkdtemp()); BARE={'Task':'title','Item id':'rich_text'}
br._request = lambda *a, **k: {'properties': {'Task':{'type':'title'},'Item id':{'type':'rich_text'}}}
print('partial :', br.ensure_columns(dict(BARE),'t','db',record=tmp/'a.json')[1])
br._request = lambda *a, **k: {'properties': {}}
print('empty   :', br.ensure_columns(dict(BARE),'t','db',record=tmp/'b.json')[1])"
```

Real output:

```
partial : ()
empty   : ('the schema PATCH returned no properties, so the new columns are unconfirmed and were not written', 'Bucket: without it every row lands in none of the three sections and is invisible on the board', 'Notes: without it no row carries its `scope=` line, so nothing is ever archived and stale rows accumulate with no run able to clear them')
```

Dropping a level per the calibration rule: I constructed that response, and Notion errors rather than returning a database silently missing what you just asked it to add. `_only_known` still names the column on the line, so it is not fully silent, just missing the consequence half.

---

### MINOR — a docstring that now contradicts the module

`q-system/.q-system/scripts/board_rows.py:270`

`_scope_of` says scope lives in Notes "so the board's schema does not change: a select option added by a writer is a schema edit the founder did not ask for." `ensure_columns` now PATCHes up to eight properties, four of them `select`, onto that same database.

```
python3 -c "
import sys, tempfile, pathlib; sys.path.insert(0,'q-system/.q-system/scripts')
import board_rows as br
tmp = pathlib.Path(tempfile.mkdtemp()); sent=[]
def fake(tok,m,path,body,op=None,bud=None):
    sent.append((m,path,body)); return {'properties':{n:{'type':t} for n,t in br.WRITES_TYPE.items()}}
br._request = fake
br.ensure_columns({'Task':'title','Item id':'rich_text'},'t','db',record=tmp/'m.json')
print(sent[0][0], sent[0][1]); [print('   ',k,'->',v) for k,v in sorted(sent[0][2]['properties'].items())]"
```

Real output:

```
PATCH /databases/db
    Bucket -> {'select': {}}
    Domain -> {'multi_select': {}}
    Link -> {'url': {}}
    Next -> {'rich_text': {}}
    Notes -> {'rich_text': {}}
    Priority -> {'select': {}}
    Source -> {'select': {}}
    Status -> {'select': {}}
```

The design changed on purpose and the rationale comment did not move with it.

---

### What is sound

- **Annotated drop strings breaking the identity refusal.** `hard` reads `names`, not `shown` (line 610), so `"Item id"` still matches. Grepped the whole repo for `dropped_columns`, `_only_known`, `ensure_columns`, `COLUMNS_MADE`: no consumer outside this module and its test, so the display strings leak nowhere.
- **Corrupt record killing the board section.** Drove `_columns_made` with a JSON list, a JSON string, a non-list board entry, broken JSON and an absent file. All five return `set()`, no exception.
- **Budget expiry swallowed by `except Exception`.** Reproduced: `ensure_columns` does swallow `Cancelled` into a note, and the next call re-raises it, so `collect`'s timeout arm still fires. `Cancelled: budget spent` printed on both sides. Not a finding.
- **A test firing a real PATCH at Notion.** Every `delenv("PYTEST_CURRENT_TEST")` site passes a `token_file`, and `_request` is stubbed with a raiser at all six of them.
- **`--dry-run` reaching the schema.** `collect` returns at line 999 before `work`, so the flag cannot PATCH.
- **The claimed suite state.** `python3 -m pytest q-system/.q-system/tests/ -q` gave `1392 passed, 1 skipped in 210.42s`.

---

**VERDICT: REQUEST CHANGES**

Fix first: the live-record assertion at `test_board_is_actionable.py:911`. Everything else in the PR is well-guarded; that one test converts a working feature into a permanently red local gate with a message pointing at the wrong writer.

FINDINGS:
major|The live-record test asserts a file production writes, so the suite goes red on the founder's machine the first morning a column is healed, blaming the suite for the 07:40 job's write|q-system/.q-system/tests/test_board_is_actionable.py:911
minor|A PATCH response carrying properties but not the requested ones returns no problem at all, while the empty-properties and refusal paths both report the unconfirmed create and its structural cost|q-system/.q-system/scripts/board_rows.py:777
minor|_scope_of's docstring still says the design avoids changing the board schema, which ensure_columns now does deliberately with eight properties including four select types|q-system/.q-system/scripts/board_rows.py:270
END FINDINGS

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-332-20260908-013533.md` (5713 bytes).

```
FINDINGS:
minor|The _remember_columns pytest chokepoint is bypassed by the test file's own delenv("PYTEST_CURRENT_TEST") idiom, so a future test can write the live column record|q-system/.q-system/scripts/board_rows.py:678
minor|No test pins that ensure_columns records the columns it created; deleting that call leaves all 79 cases green, so the "a column he deleted stays deleted" guarantee has no end-to-end coverage|q-system/.q-system/scripts/board_rows.py:787
minor|The docstring says a failed record write costs "offering to create a column he removed a second time"; a persistently failing write re-creates and re-announces it on every run|q-system/.q-system/scripts/board_rows.py:666
END FINDINGS
```

--- reviewer output ---

**PR #332 review: three minor findings, all reproduced. Nothing blocks.**

---

## Finding 1 (minor) — the "no test can write the live record" chokepoint is defeated by this file's own dominant test idiom

`q-system/.q-system/scripts/board_rows.py:678`

The comment claims a structural guarantee: *"THE LIVE RECORD IS NEVER WRITTEN BY A TEST, and that is a chokepoint rather than a rule someone remembers... This one protects the ones nobody has written."* The gate is `os.environ.get("PYTEST_CURRENT_TEST")`, and `test_board_is_actionable.py` calls `monkeypatch.delenv("PYTEST_CURRENT_TEST", raising=False)` in **five** places to reach `collect`. A future test copying that idiom without stubbing `ensure_columns` writes the live record. Today's tests stub it, so the suite is green.

Reproducer (`COLUMNS_MADE` repointed at a tmp file that stands in for `~/.config/kipi`, so nothing live was touched):

```
python3 -m pytest /var/folders/.../pr332/repro1.py -q -s
```

Real output:

```
LIVE RECORD EXISTS: True
LIVE RECORD CONTENT: {"d": ["Bucket","Domain","Link","Next","Notes","Priority","Source","Status"]}
E  AssertionError: chokepoint did not fire; a test wrote the live record
1 failed in 0.02s
```

Fix shape: gate on `sys.modules.get("pytest")` or an explicit injected sink, not an env var the suite deletes on purpose.

---

## Finding 2 (minor, test-coverage) — nothing pins that production records what it created

`q-system/.q-system/scripts/board_rows.py:787`

Deleting the `_remember_columns(db, made, record)` call leaves the entire file green. `test_a_column_he_deleted_is_not_put_back` hand-writes the record file, and `test_a_private_path_is_allowed` drives `_remember_columns` directly, so no case observes `ensure_columns` populating it. If that call is ever refactored away, `made_before` is empty forever and every column the founder deletes comes back every morning: the exact round-4 major this PR names as fixed.

Mutation run (mirrored tree, `_remember_columns(db, made, record)` replaced with `pass`):

```
python3 -m pytest .../mirror/.q-system/tests/test_board_is_actionable.py -q
79 passed in 0.07s
```

Four control mutants on the same mirror were all killed, so the harness works: heal-before-identity 1 failed, refusal-drops-structural-cost 1 failed, `CREATABLE` back to Link+Next 7 failed, deletion-record-ignored 3 failed.

---

## Finding 3 (minor) — the docstring understates what a failed record write costs

`q-system/.q-system/scripts/board_rows.py:666`

*"A failure here is not fatal: the worst case is offering to create a column he removed a second time."* `_remember_columns` swallows `OSError` on both `mkdir` and `write_text`. A persistently failing write means every deleted column is re-created and re-announced on **every** run, not a second time.

Reproducer output (4 simulated mornings, record dir chmod 555):

```
morning 1 | record on disk: False | PATCHed: ['Bucket','Domain','Link','Next','Notes','Priority','Source','Status']
morning 2 | record on disk: False | PATCHed: [same]
morning 3 | record on disk: False | PATCHed: [same]
morning 4 | record on disk: False | PATCHed: [same]
observed: created on every one of 4 runs
```

Trigger is constructed (`~/.config/kipi` is writable in practice, it holds the token and the lock), so I dropped this a level per calibration. The same shape reaches production without construction on a second machine: the record is machine-local and the Notion DB is shared.

---

## What is sound

- **Identity before heal.** Real, load-bearing, and the test kills the reversed order. A DB this module rejects never gets PATCHed.
- **`_only_known`'s two-list split.** The `hard` check reads `names`, not the display strings, so a wrong-typed `Item id` still raises `MissingIdentityColumn`.
- **`CREATABLE` derivation.** Mutating it back to `("Link","Next")` turns 7 cases red, including the hand-written exact-set assertion. The fixture is written out by hand rather than derived from `WRITES_TYPE`, which is the right call.
- **Budget cancellation swallowed by `except Exception`.** Confirmed `Cancelled` is caught and turned into a note. I could not make it cause harm: `paint`'s next `_request` re-raises, and `collect` returns at the `Cancelled` arm before the notes are ever composed, so the note is discarded. Dropped.
- **`_columns_made` corrupt-input handling.** JSON list, JSON string, non-list board entry, broken JSON, absent file: all read empty, all pinned.
- **Cross-board state.** No module global; `test_no_module_state_leaks_between_boards` goes red on the round-5 shape.
- **Full suite:** `python3 -m pytest q-system/.q-system/tests/ -q` gave `1392 passed, 1 skipped in 212.76s`. The PR body says 1370, which is drift, not a defect.

---

First thing to fix: Finding 1. The chokepoint is the one piece of this PR sold as structural rather than remembered, and an env var the suite deletes five times is not a chokepoint.

FINDINGS:
minor|The _remember_columns pytest chokepoint is bypassed by the test file's own delenv("PYTEST_CURRENT_TEST") idiom, so a future test can write the live column record|q-system/.q-system/scripts/board_rows.py:678
minor|No test pins that ensure_columns records the columns it created; deleting that call leaves all 79 cases green, so the "a column he deleted stays deleted" guarantee has no end-to-end coverage|q-system/.q-system/scripts/board_rows.py:787
minor|The docstring says a failed record write costs "offering to create a column he removed a second time"; a persistently failing write re-creates and re-announces it on every run|q-system/.q-system/scripts/board_rows.py:666
END FINDINGS

PR #333, opened 2026-09-08
prd-os: restore the ASK-1040 guard and the 7 tests a fleet sync deleted

assafkip:
**Restoration only.** This carries back code that lived in the tree until 2026-09-06 and
was already reviewed then. It deliberately does **not** carry the new door heuristic that
#330 builds on top of it — that stays in #330 to be judged on its own merits.

## What deleted it

Commit `75014dc0` in the consulting instance:

```
chore: commit system-written state before skeleton sync
  [no-issue: fleet updater system-state commit]
  _reject_unrunnable_gate:  2 -> 0
  test_gate_lifecycle.py:   118 lines removed
```

The instance's `plugins/` is an `rsync --delete` destination from **this** repo, and this
repo has never had the guard, so the sync overwrote it with an older copy and committed
the result under a housekeeping message. Nothing failed; nothing logged a loss.

Blast radius, measured across all 45 differing vendored files: **2 files really lost
enforcement**, and they are these two.

**The clock is still running.** Skeleton has 0 occurrences, consulting main has 2, and
the sync runs skeleton → instance. A fleet sync before this lands re-deletes the guard.

## Restored

- `_reject_unrunnable_gate`, verbatim from consulting `origin/main`
- the ASK-1038 vacuous-pass WARN, original wording, **no exit-code change**
- `issue_runner`'s `lifecycle="regression"` plus its 14-line scar comment
- the 7 test functions `75014dc0` deleted

## One deviation from verbatim, and it is measured

The original takes `cmd.split()[0]` as the command name, which is an **assignment** for
`PYTHONPATH=src python3 ...`. `issue_runner` runs the bypass_check and requires exit 0
*before* registering, then turns any raise here into a hard `cannot close` — so a verbatim
restore would break closeout on two specs whose checks had just gone green.

| | refused | false positives |
|---|---|---|
| original guard | 16 / 204 | **2** (`srsa-authoritative-path-contract`, `srsa-check-repairs-survived-merge`) |
| restored guard | 14 / 204 | **0** — both named specs accepted |

All 14 remaining read by hand: every one is genuine prose in the command slot.

## Shell grammar is deliberately not special-cased

**0 of 204** bypass_checks start with `(` `{` `!` `<` `>`, so probing them as a first
token refuses them and fails **closed** at zero measured cost. That is the entire
contested surface of #330 — negation branch, subshell and quoted-token handling — left
out on purpose. Three review rounds each found a new shell shape in it; none reaches here.

The guard also runs **after** the idempotency lookup. `gate_id` is
`issue_id+sha256(command)`, so a re-close resolves to an existing row and must be a no-op;
validating first turned that into a raise, which `issue_runner` converts to `cannot close`.

## One addition beyond the restore, and it is a test

The four ASK-1040 tests drive `_reject_unrunnable_gate` **directly**, so deleting the
single call site from `gate_register` left the file green at **17 passed**. Restoring a
guard whose wiring is undefended is restoring half of it.
`test_gate_register_ACTUALLY_CALLS_the_unrunnable_guard` drives the real entry point and
asserts it refused *before* writing. Same mutant now: **1 failed, 17 passed**.

## Verified

- prd-os **692 passed / 1 skipped**, kipi-dsse **4 passed**
- restored-guard census **204 / 14**, zero false positives
- none of the 7 restored tests references a #330-only construct (grep: 0)

Related: `sp-43837284` (the recorded deletion), `ASK-1354` (the gate registry is
per-worktree and untracked — separate, larger, and not touched here).

### DIFF
--- a/plugins/kipi-dsse/.claude-plugin/plugin.json
+++ b/plugins/kipi-dsse/.claude-plugin/plugin.json
@@ -1,10 +1,10 @@
 {
- "name": "kipi-dsse",
- "version": "0.14.4",
- "description": "DSSE issue workflow: scope-enforced edits, per-finding dispositions, Codex review gates. Provides /issue-start, /issue-approve, /issue-verify, /issue-review, /issue-amend, /issue-closeout. Consumes issue specs created by prd-os or dropped in manually.",
- "author": {
-  "name": "Assaf Kipnis"
- },
- "repository": "https://github.com/assafkip/kipi-system",
- "license": "MIT"
+  "name": "kipi-dsse",
+  "version": "0.14.5",
+  "description": "DSSE issue workflow: scope-enforced edits, per-finding dispositions, Codex review gates. Provides /issue-start, /issue-approve, /issue-verify, /issue-review, /issue-amend, /issue-closeout. Consumes issue specs created by prd-os or dropped in manually.",
+  "author": {
+    "name": "Assaf Kipnis"
+  },
+  "repository": "https://github.com/assafkip/kipi-system",
+  "license": "MIT"
 }
--- a/plugins/kipi-dsse/scripts/issue_runner.py
+++ b/plugins/kipi-dsse/scripts/issue_runner.py
@@ -1164,9 +1164,23 @@ def _enforce_spine_contract(paths, fm: dict, marker: dict, issue_id: str) -> str
             import prd_runner as _prd_runner
             from config import load as _load_config
             cfg = _load_config(paths.repo_root)
+            # LIFECYCLE IS EXPLICIT, NEVER DEFAULTED. Scar 2026-08-24 (ASK-1038):
+            # this call omitted `lifecycle=`, so it took LEGACY_GATE_LIFECYCLE,
+            # which is "historical-receipt" -- the one value `gates run` filters
+            # OUT before executing. The only producer of gates was writing the
+            # one lifecycle the only consumer refuses to run, so `gates run`
+            # printed "all 0 regression gates green" against a 47-gate registry
+            # and had executed nothing since 2026-07-26. 47 closeouts registered
+            # a permanent proof; all 47 were inert from the moment they landed.
+            #
+            # A bypass_check IS a permanent regression proof -- that is what the
+            # registry section header calls it -- so `regression` is the correct
+            # value and it is passed here rather than relied on as a default.
+            # "legacy" in the constant's name reads as "applies to old rows" to
+            # every reviewer, which is exactly why nobody caught it for 29 days.
             out = _prd_runner.gate_register(
                 cfg, prd_id=marker["prd_id"], issue_id=issue_id,
-                command=bypass_check)
+                command=bypass_check, lifecycle="regression")
         except Exception as exc:
             return (f"cannot close {issue_id}: bypass_check gate registration "
                     f"failed ({exc}) — the permanent registry must record it "
--- a/plugins/prd-os/.claude-plugin/plugin.json
+++ b/plugins/prd-os/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "prd-os",
-  "version": "0.30.3",
+  "version": "0.30.4",
   "description": "PRD operating system: capture rough ideas, draft reviewable PRDs, run standard and adversarial review (Codex or a Claude senior-staff-engineer subagent, recorded truthfully), triage findings, decompose approved PRDs into atomic issue specs, and execute those issues with scope enforcement, stop-gate receipts, and reviewer-attributed findings.",
   "author": {
     "name": "Assaf Kipnis"
--- a/plugins/prd-os/.claude-plugin/plugin.json
+++ b/plugins/prd-os/.claude-plugin/plugin.json
@@ -1,6 +1,6 @@
 {
   "name": "prd-os",
-  "version": "0.30.2",
+  "version": "0.30.3",
   "description": "PRD operating system: capture rough ideas, draft reviewable PRDs, run standard and adversarial review (Codex or a Claude senior-staff-engineer subagent, recorded truthfully), triage findings, decompose approved PRDs into atomic issue specs, and execute those issues with scope enforcement, stop-gate receipts, and reviewer-attributed findings.",
   "author": {
     "name": "Assaf Kipnis"
--- a/plugins/prd-os/scripts/prd_runner.py
+++ b/plugins/prd-os/scripts/prd_runner.py
@@ -1213,6 +1213,30 @@ def _reject_unrunnable_gate(command: str) -> None:
             probe_cmd = head[1]
             continue
         break
+    # A LEADING `!` IS STRIPPED, and this is the one shell shape that must be handled.
+    # Not a general grammar heuristic -- `!` alone earns it, because it INVERTS:
+    #
+    #   `command -v !` RESOLVES (it is a bash reserved word), so without this the probe
+    #   accepts `! <anything>`, and `bash -c '! <missing command>'` exits 0. A negated
+    #   prose gate would register and pass forever in an append-only registry.
+    #   `(` and `{` do not need this: prose inside them exits 127 and the gate goes RED
+    #   loudly, which is the tolerable direction.
+    #
+    # `! grep -q ...` (assert-absent) is a REAL and recurring idiom, not a hypothetical.
+    # Censused across both populations on this machine 2026-09-08:
+    #   registered gates.jsonl : 1032 rows / 699 distinct -> 5 distinct start with `!`
+    #   spec bypass_checks     : 41588 rows / 958 distinct -> the same 5
+    # all five are `! grep`-shaped absence assertions and one is a secret scan. An
+    # earlier census of ONE repo's 204 bypass_checks found 0 and concluded this shape
+    # was unused; that was the wrong population (see ASK-1354).
+    #
+    # Stripping keeps all five registrable AND refuses `! <prose>`.
+    while probe_cmd.lstrip()[:1] == "!":
+        probe_cmd = probe_cmd.lstrip()[1:]
+        if not probe_cmd.strip():
+            raise ValueError(
+                "gate command is only a negation, so it runs no check.\n"
+                f"  command: {cmd[:160]}")
     first = probe_cmd.split()[0]
     probe = _sp.run(["bash", "-lc", f"command -v {shlex.quote(first)} >/dev/null 2>&1"],
                     capture_output=True, text=True)
--- a/plugins/prd-os/scripts/prd_runner.py
+++ b/plugins/prd-os/scripts/prd_runner.py
@@ -46,6 +46,7 @@ import subprocess
 import tempfile
 import os
 import re
+import shlex
 import sys
 from datetime import datetime, timezone
 from pathlib import Path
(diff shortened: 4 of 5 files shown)


assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-333-20260908-002304.md` (7299 bytes).

```
FINDINGS:
minor|The unrunnable-gate guard's prose class is unreachable from its only caller, so its only reachable effect is falsely refusing a bypass_check that already exited 0|plugins/prd-os/scripts/prd_runner.py:1264
minor|The ASK-1038 vacuous-pass WARN triggers only on zero regression gates, so it cannot detect a registrar regression on any registry holding one|plugins/prd-os/scripts/prd_runner.py:2969
minor|Comment claims bash -n catches unterminated heredocs; bash -n returns 0 on them and the PR's own test says so|plugins/prd-os/scripts/prd_runner.py:1181
END FINDINGS
```

--- reviewer output ---

I have enough. Writing the review.

# Review: PR #333 — restore the ASK-1040 guard and 7 deleted tests

Read-only. Nothing in the repo was modified; all mutants ran against copies in `tempfile.mkdtemp()`.

---

## Finding 1 — minor — the guard's true-positive class is unreachable from its only caller; only its false-positive class is reachable

`plugins/prd-os/scripts/prd_runner.py:1264`

`gate_register` has exactly one production caller: `issue_runner.py:1181`. That call happens strictly after `subprocess.run(bypass_check, shell=True)` returned **0** (`issue_runner.py:1145-1160`). A prose command exits 127, so it aborts closeout there and `gate_register` is never entered. The guard therefore never sees the ASK-1040 prose class in production. What it *can* still do is refuse a command that just proved itself green, and blame it on prose.

Reproducer (drives the real `_enforce_spine_contract`, with `gate_register` spied):

```
python3 - <<'PYEOF'   # full script pasted at the end of this section
PYEOF
```

Real output:

```
--- PROSE  - the class ASK-1040 exists for
    command           : 'check:the ledger is append-only'
    sh exit           : 127
    gate_register hit : False
    result            : cannot close issue-demo: bypass_check exited 127:
--- GREEN subshell - guard false-positive class
    command           : '(cd . && true)'
    sh exit           : 0
    gate_register hit : True
    result            : cannot close issue-demo: bypass_check gate registration failed (gate command
                        does not start with a runnable command: '(cd'. This is the
                        prose-in-the-command-slot class (ASK-1040): write the CHECK here, not a
                        description of it.
```

Severity dropped one level because the triggering shape is constructed: I confirmed the PR's own census against all 204 real `bypass_check` values (0 grammar openers, 14 refused, every one genuine prose that also exits 127). At 3am the concrete action is a one-line spec rewrite, and it fails closed.

## Finding 2 — minor — the ASK-1038 vacuous-pass WARN cannot fire on any registry that already holds one regression gate

`plugins/prd-os/scripts/prd_runner.py:2969`

The condition is `not records and registered_total`, i.e. **zero** regression gates. This PR's own `lifecycle="regression"` fix guarantees every future registry has at least one. So if the registrar regresses again (which is the mechanism this whole PR exists to undo, and the PR says the fleet-sync clock is still running), the WARN stays silent.

Reproducer, real CLI, two temp registries:

```
=== CASE A: 47 registered, ALL historical-receipt (the ASK-1038 shape) ===
rc=0
[WARN] gates: 47 gate(s) registered, NONE executable -- 0 have lifecycle 'regression', so no regression check ran. Exit 0 below covers the spillover verdict ONLY.
all 0 regression gates green (47 registered); 0 blocking spillover item(s)

=== CASE B: same registrar failure, but ONE pre-existing regression gate ===
rc=0
all 1 regression gates green (47 registered); 0 blocking spillover item(s)
WARN present: False
```

Not major, deliberately. The ratio `1 ... (47 registered)` is on screen and that count is new in this PR, so the merged state is strictly better than the pre-PR state, which printed nothing. The alarm is narrow, not absent. The fix is a threshold on non-regression rows rather than on zero.

## Finding 3 — minor — the code comment says `bash -n` catches unterminated heredocs; the test in the same PR says it does not, and the test is right

`plugins/prd-os/scripts/prd_runner.py:1181` says `# 1. Does it parse? Catches the unterminated-heredoc class.`
`plugins/prd-os/tests/test_gate_lifecycle.py:334` says the opposite, verbatim: "`bash -n -c` accepts it (rc=0)".

Reproducer:

```
bash -n -c 'cat <<EOF
unterminated'; echo "rc=$?"
```

Real output: `bash -n unterminated-heredoc rc=0`

The test is correct, the code comment is wrong, and the two sit 150 lines apart in one diff. The next reader re-derives it.

---

## What is sound

- **Mutation, four separate mutants, each killed.** Copied the tree, patched one line, ran `tests/test_gate_lifecycle.py`. Deleting `_reject_unrunnable_gate(command)`: `1 failed, 17 passed`. Dropping `lifecycle="regression"` at the closeout call site: `1 failed, 17 passed`. Deleting the WARN block: `1 failed, 17 passed`. Making a RED gate not append to `failures`: `1 failed, 17 passed`. None of the seven new tests is decoration.
- **The census claims in the PR body are true.** Decoded all 204 real `bypass_check` values through `issue_runner._decode_bypass_check` and ran the guard: `total 204`, `grammar-opener count: 0`, `refused 14`, and reading all 14 by hand, every one is genuine prose in the command slot. Zero false positives. The "fails closed at zero measured cost" argument holds against the real corpus, not just against a fixture.
- **Suites.** `prd-os`: `692 passed, 1 skipped in 61.84s`. `kipi-dsse`: `110 passed in 36.26s`. The gate file also runs green at repo scope, not just at plugin scope.
- **The `!` negation branch, the one #330 kept losing rounds on.** `! grep -rn FOO plugins/*.py` is **accepted**, not refused: `command -v '!'` returns 0 because bash reports the reserved word. The deliberate omission costs nothing here.
- **Guard cost.** 204 probes in 1.3s, two subprocesses each. No 3am latency concern.
- **Argument injection into the probe.** `first` is `shlex.quote`d into `bash -lc`, and `bash -n` gets the command as argv, never eval'd. I could not get anything to execute during validation.
- **Ordering of the guard vs the idempotency lookup.** The re-close no-op path is genuinely preserved; validating first would raise on a re-close, and the comment at line 1260 matches what the code does.
- **`bash -lc` PATH divergence from the `sh -c` runner.** Real in principle (login shell runs `path_helper`), but I could not build a reproducer without editing the user's profile, and the failure surfaces as a loud `GATE RED` rather than a silent pass. Dropped.

---

## VERDICT: APPROVE WITH NITS

Three minors, no blocker, no major. This is a restoration that measurably improves the merged state over what is on `main` today, its tests survive mutation, and its census claims reproduce against the real 204-value corpus rather than against invented fixtures.

**Fix first:** Finding 2, the WARN condition at `prd_runner.py:2969`. It is the recurrence detector for the exact 29-day outage this PR exists to undo, and it goes blind the moment the PR's own fix writes the first regression row. One line: warn on non-regression rows existing, not on regression rows being zero.

FINDINGS:
minor|The unrunnable-gate guard's prose class is unreachable from its only caller, so its only reachable effect is falsely refusing a bypass_check that already exited 0|plugins/prd-os/scripts/prd_runner.py:1264
minor|The ASK-1038 vacuous-pass WARN triggers only on zero regression gates, so it cannot detect a registrar regression on any registry holding one|plugins/prd-os/scripts/prd_runner.py:2969
minor|Comment claims bash -n catches unterminated heredocs; bash -n returns 0 on them and the PR's own test says so|plugins/prd-os/scripts/prd_runner.py:1181
END FINDINGS

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-333-20260908-003209.md` (6527 bytes).

```
FINDINGS:
minor|A self-referentially-skipped gate is counted in "all N regression gates green" and the new WARN's predicate checks the selected set, not the executed set|plugins/prd-os/scripts/prd_runner.py:2993
minor|Assignment stripping runs before the `!` strip and never re-runs, so `! PYTHONPATH=x cmd` is refused and becomes a hard `cannot close`|plugins/prd-os/scripts/prd_runner.py:1208
nit|prd-os version bumped 0.30.2 to 0.30.4 with no CHANGELOG entry; newest entry there is 0.22.0|plugins/prd-os/.claude-plugin/plugin.json:3
END FINDINGS
```

--- reviewer output ---

# Review: PR #333 — restore the ASK-1040 guard

Read-only. Repo untouched; the one mutation ran against a copy in a temp dir.

## Findings

### 1. minor — a skipped gate is counted as green, and the new WARN cannot see it

`prd_runner.py:2993` fires the new WARN only when `records` is empty *after the lifecycle filter*. The executed set can be empty for a second reason on the same code path: the self-referential skip at `:2911`. `skipped_self_ref` is incremented at `:2911` and read nowhere. The summary at `:3002`, which this PR edits, counts skipped gates in "all N regression gates green".

The comment above the WARN says "AN EMPTY EXECUTED SET IS NOT A PASS". The code's predicate is "an empty *selected* set", which is a different set.

Reproducer, a gate that exits 2 and is reported green:

```
python3 - <<'PY'
cmd = "grep -q 'self-referential .gates run. gate' MISSING_FILE.py"
# ... registered as lifecycle=regression, then `gates run`
PY
```

Real output:

```
does this command pass on its own?  rc= 2
door guard: ACCEPTED
STDOUT:
[skip] asserts-the-selfref-comment: self-referential `gates run` gate (not executed)
[census] spillover: 0 open total
all 1 regression gates green (1 registered); 0 blocking spillover item(s)
EXIT = 0
```

Calibrated down from major. I constructed the input. Censused this repo's 189 distinct spec bypass_checks: **0** match `\bgates\s+run\b`, so nothing here emits it today. I could not reach the fleet's `gates.jsonl` populations from this sandbox.

### 2. minor — `! PYTHONPATH=... cmd` is refused, which is the shape the deviation exists to accept

The assignment-stripping loop at `prd_runner.py:1208` runs to completion **before** the `!` strip at `:1234`, and never re-runs. So a negated command carrying an env assignment resolves its first token as `PYTHONPATH=src`.

At closeout that is a hard `cannot close` (`issue_runner.py:1185` turns any raise into it). The `PYTHONPATH=` handling is the single documented deviation from the verbatim restore, and the `!` handling is the whole of commit `943d7b32`. The two do not compose.

Real output:

```
'! PYTHONPATH=src python3 -m pytest -q tests/test_x.py'
   bash -n rc=0   guard -> REFUSED: gate command does not start with a runnable command: 'PYTHONPATH=src'
```

Same calibration caveat: constructed, 0 producers in this repo's 189.

### 3. nit — version bumped, changelog silent

`plugins/prd-os/.claude-plugin/plugin.json:3` goes 0.30.2 to 0.30.4. The newest entry in `plugins/prd-os/CHANGELOG.md` is `## [0.22.0]`; `grep -n "^## \[0\.3"` returns only `961:## [0.3.0]`. The file's own first line says all notable changes are recorded there. Pre-existing rot, two more versions added to it.

## What is sound

**The prose census holds.** I ran `_reject_unrunnable_gate` over all 204 bypass_check occurrences (189 distinct) in `.prd-os/issues/`. Refused 14, matching the PR body exactly. I read all 14: every one is genuine English prose in the command slot (`the lint never executes a named enforcer: grep ...`, `no unconditional slice of the item list remains`). Zero false positives. The two named specs, `srsa-authoritative-path-contract` and `srsa-check-repairs-survived-merge`, are accepted.

**The wiring test has teeth.** I mutated `_reject_unrunnable_gate(command)` to `pass` on a copy in `/var/folders/.../scripts/`. Output: `MUTANT: gate_register did NOT raise -> the wiring test would go RED. GOOD.` and `prose row written by mutant: True`. That test is the one addition beyond the restore and it earns its place.

**The subshell claim is honest.** I probed which grammar openers `command -v` resolves under `bash -lc`: `if`, `for`, `while`, `{`, `[[`, `!` all return 0; only `(` returns 1. So the only shape that fails closed is a leading `(`, and 0 of 189 bypass_checks start with one. "Fails closed at zero measured cost" is accurate for the population I could reach.

**No injection surface.** `bash -n -c` parses without executing, and `first` is `shlex.quote`d before the `command -v` probe. I could not get either subprocess to execute gate content.

**Probe cost is not a 3am hazard.** 5 guard calls in 0.03s, 6ms each including the `bash -lc`. No hang risk demonstrated.

**Suite is green as claimed.** `python3 -m pytest plugins/prd-os/tests plugins/kipi-dsse -q` gives `804 passed, 1 skipped in 98.25s`.

**Blast radius of the guard itself is bounded.** `gate_register` has exactly one caller (`issue_runner.py:1181`); there is no `gates register` CLI subcommand (`p_gates.add_argument("gates_cmd", choices=("list", "run"))` at `:3032`). And `issue_runner` runs the bypass_check to exit 0 *before* registering, so a prose command already fails closeout at the run step — the guard is not a new refusal path for the class it names.

**Attacks I dropped because I could not make them bite:** the `bash -lc` login-PATH vs `/bin/sh -c` runtime-PATH mismatch (guard's resolution set is a strict superset of the runner's, so it errs permissive, which is the safe direction); the missing timeout on the probe subprocess (no hang reproduced); the widened TOCTOU window between the idempotency read and the append (6ms, and `gate_id` is deterministic so a duplicate row is a re-run, not a corruption).

## VERDICT: APPROVE WITH NITS

Both surviving findings are minor: real, reproduced, and with no producer in this repo emitting the triggering input. Neither is permanent or unrecoverable. The restoration itself is well-measured and the one novel piece, the wiring test, is the strongest thing in the diff.

**Fix first:** finding 1. Change the `:2993` predicate from `not records` to "nothing was actually executed" — `skipped_self_ref` is already counted at `:2911` and thrown away, so the fix is `if not (len(records) - skipped_self_ref) and registered_total`, plus subtracting skipped gates from the count in the `:3002` summary. That one line makes the WARN mean what its comment already claims.

FINDINGS:
minor|A self-referentially-skipped gate is counted in "all N regression gates green" and the new WARN's predicate checks the selected set, not the executed set|plugins/prd-os/scripts/prd_runner.py:2993
minor|Assignment stripping runs before the `!` strip and never re-runs, so `! PYTHONPATH=x cmd` is refused and becomes a hard `cannot close`|plugins/prd-os/scripts/prd_runner.py:1208
nit|prd-os version bumped 0.30.2 to 0.30.4 with no CHANGELOG entry; newest entry there is 0.22.0|plugins/prd-os/.claude-plugin/plugin.json:3
END FINDINGS

PR #334, opened 2026-09-08
fix(board): the record chokepoint has no environment left to defeat (ASK-1346)

assafkip:
## What

The live-record chokepoint stops depending on an environment variable that this suite deliberately unsets. Linear: ASK-1346. Follow-up to #332 (merged `b8901176`).

## Why

`_remember_columns` refused only when `PYTEST_CURRENT_TEST` was set. Every test that reaches `collect` deletes that variable, because it must to get past the board's own chokepoint. **So the guard was bypassed by the exact idiom that caused the problem it was written for.** A test could still write `~/.config/kipi/board-columns-created.json` and teach the 07:40 job that the founder had deleted columns he never touched. That happened once already, in #332, and was cleaned up by hand.

## How

No ambient default. `_remember_columns` raises when given no path, on any machine, in any environment. The live record is reachable only through `collect`, which names it explicitly. A test that forgets `record=` fails identically whether or not pytest is announcing itself.

## Also

The first version of the "production names the live path" test grepped the source for `record=COLUMNS_MADE` and asserted one occurrence. There are two: the guard's own error message contains that sentence. A source substring is not a behaviour, so it watches the call instead.

## Proof

- `python3 -m pytest q-system/.q-system/tests/ -q`: 1393 passed, 1 skipped.
- Run directly with `PYTEST_CURRENT_TEST` removed, the previous version writes without complaint and this one refuses.

### DIFF
--- a/q-system/.q-system/scripts/board_rows.py
+++ b/q-system/.q-system/scripts/board_rows.py
@@ -722,8 +722,21 @@ def _remember_columns(db, names, path=None):
         pass
 
 
-def ensure_columns(known, token, db, opener=None, budget=None, record=None):
+def ensure_columns(known, token, db, opener=None, budget=None, *, record):
     """Add the optional columns this board is missing. Returns (schema, problems).
+
+    `record` IS REQUIRED, and keyword-only so it cannot be passed by accident (PR #334
+    reviewer round 2, minor). The inner reader and writer both refuse a missing path,
+    and this signature still offered `record=None`, so the refusal arrived from two
+    frames down as an AssertionError no `collect` arm catches, instead of from the call
+    a person was looking at. A default whose only outcome is a crash is not a default.
+    `opener` and `budget` keep theirs: those are genuinely optional and every test in
+    the suite omits them.
+
+    A REFUSED OR UNCONFIRMED CREATE is not fatal and degrades with its reason on the
+    line. A missing `record` is a programming error and is meant to be loud; the old
+    docstring flattened both into "failure here is not fatal", which was true of one of
+    them.
 
     THE FEATURE WAS DEAD ON EVERY BOARD BUT ONE (PR reviewer round 12, major). Nothing
     created `Link`, so `_only_known` dropped it on every run and the morning line said
--- a/q-system/.q-system/scripts/board_rows.py
+++ b/q-system/.q-system/scripts/board_rows.py
@@ -652,8 +652,16 @@ def _columns_made(db, path=None):
     is the safe answer: the worst it costs is offering a column he removed once more,
     and that is said on the line.
     """
+    # NO AMBIENT DEFAULT ON THE READ EITHER (PR #334 reviewer, minor). The write lost
+    # its default and the comment said there was none left; the read still had one, so
+    # a caller that omitted `record=` quietly consulted the live file while the module
+    # claimed that could not happen. Half a chokepoint described as a whole one.
+    if path is None:
+        raise AssertionError(
+            "no record path given. Production passes record=COLUMNS_MADE explicitly; "
+            "a test passes its own tmp path. There is no default.")
     try:
-        data = json.loads(Path(path or COLUMNS_MADE).read_text("utf-8"))
+        data = json.loads(Path(path).read_text("utf-8"))
     except (OSError, ValueError):
         return set()
     if not isinstance(data, dict):
@@ -663,9 +671,18 @@ def _columns_made(db, path=None):
 
 
 def _remember_columns(db, names, path=None):
-    """Append to the record. A failure here is not fatal: the worst case is offering to
-    create a column he removed a second time, which is the behaviour before the record
-    existed, and it is still said out loud on the line.
+    """Append to the record.
+
+    A WRITE failure is not fatal: an OSError is swallowed and the worst case is
+    offering to create a column he removed a second time, which is the behaviour
+    before the record existed, and it is still said out loud on the line.
+
+    A MISSING PATH is a different thing and is fatal on purpose (PR #334 reviewer,
+    minor: the old docstring said "a failure here is not fatal" flatly, while the
+    AssertionError below escapes every except arm in `collect` and takes the whole
+    board section with it). That is the intent. A caller that forgot `record=` is a
+    programming error, it only reaches this line in a test, and a loud failure in the
+    suite is exactly what stops it reaching his machine.
 
     THE LIVE RECORD IS NEVER WRITTEN BY A TEST, and that is a chokepoint rather than a
     rule someone remembers (PR #332 reviewer round 7, minor). `collect` has had this
--- a/q-system/.q-system/scripts/board_rows.py
+++ b/q-system/.q-system/scripts/board_rows.py
@@ -71,6 +71,7 @@ import fcntl
 import hashlib
 import json
 import os
+import sys
 import time
 import urllib.error
 import urllib.request
@@ -640,6 +641,38 @@ CREATABLE = tuple(n for n in WRITES_TYPE if n not in UNDROPPABLE)
 #: ABSENT was deleted by a person, and this module does not put it back.
 COLUMNS_MADE = STATE_DIR / "board-columns-created.json"
 
+#: THE REAL LOCATION, frozen. Tests redirect `COLUMNS_MADE`; nothing redirects this, so
+#: it is what "am I about to touch the founder's live file" is measured against.
+_LIVE_RECORD = STATE_DIR / "board-columns-created.json"
+
+
+def _refuse_live_record_under_test(path):
+    """Refuse the founder's real record file while a test is running.
+
+    THE GUARD HAS TO SURVIVE `collect` NAMING THE PATH (PR #334 reviewer round 3,
+    major). Requiring an explicit `record=` closed the direct callers and opened the
+    one that mattered: `collect` names COLUMNS_MADE itself, so any test reaching
+    `collect` sailed through a check the previous commit would have failed. My fix made
+    the exact incident case worse while the commit message said it was closed.
+
+    And it cannot key on PYTEST_CURRENT_TEST, because every collect-level test in this
+    suite deletes that variable to get past the board's own chokepoint. `sys.modules`
+    is the honest signal: pytest is imported for the whole run and no test removes it.
+    A test that wants this path redirects `COLUMNS_MADE`, which is what the others
+    already do for `LOCK_FILE`.
+    """
+    if "pytest" not in sys.modules:
+        return
+    try:
+        same = Path(path).resolve() == _LIVE_RECORD.resolve()
+    except OSError:
+        same = str(path) == str(_LIVE_RECORD)
+    if same:
+        raise AssertionError(
+            f"a test reached the live column record at {_LIVE_RECORD}. Redirect it "
+            "with monkeypatch.setattr(board_rows, 'COLUMNS_MADE', tmp_path / ...), "
+            "the way LOCK_FILE is redirected.")
+
 
 
 def _columns_made(db, path=None):
@@ -660,6 +693,7 @@ def _columns_made(db, path=None):
         raise AssertionError(
             "no record path given. Production passes record=COLUMNS_MADE explicitly; "
             "a test passes its own tmp path. There is no default.")
+    _refuse_live_record_under_test(path)
     try:
         data = json.loads(Path(path).read_text("utf-8"))
     except (OSError, ValueError):
@@ -707,6 +741,7 @@ def _remember_columns(db, names, path=None):
         raise AssertionError(
             "no record path given. Production passes record=COLUMNS_MADE explicitly; "
             "a test passes its own tmp path. There is no default.")
+    _refuse_live_record_under_test(path)
     p = Path(path)
     try:
         data = json.loads(p.read_text("utf-8")) or {}
--- a/q-system/.q-system/scripts/board_rows.py
+++ b/q-system/.q-system/scripts/board_rows.py
@@ -675,10 +675,22 @@ def _remember_columns(db, names, path=None):
     of it found a test reaching something live and each was fixed by hand; a hand fix
     protects the tests that exist today. This one protects the ones nobody has written.
     """
-    if path is None and os.environ.get("PYTEST_CURRENT_TEST"):
+    if path is None:
+        # NO AMBIENT DEFAULT. The first version of this guard refused only when
+        # PYTEST_CURRENT_TEST was set, and every test in this suite that reaches
+        # `collect` deletes that variable to get past the board's own chokepoint -- so
+        # the guard was bypassed by the exact idiom that caused the problem it was
+        # written for, and I had already described it to the founder as a mechanism
+        # rather than a discipline (PR #332 reviewer round 9, minor).
+        #
+        # There is no environment to read now: the live path is reachable only by a
+        # caller that names it, which is `collect` and nothing else. A test that
+        # forgets `record=` fails the same way on any machine, in any environment,
+        # whether or not pytest is announcing itself.
         raise AssertionError(
-            "a test tried to write the live column record; pass record=<tmp path>")
-    p = Path(path or COLUMNS_MADE)
+            "no record path given. Production passes record=COLUMNS_MADE explicitly; "
+            "a test passes its own tmp path. There is no default.")
+    p = Path(path)
     try:
         data = json.loads(p.read_text("utf-8")) or {}
     except (OSError, ValueError):
@@ -1050,7 +1062,9 @@ def collect(now, sources: dict, opener=None, token_file=None, db_file=None,
             # takes a raw 400 from Notion before `_only_known` is ever reached and the
             # remediation sentence never fires (#327 round 10, minor).
             _refuse_without_identity(known)
-            known, column_problems = ensure_columns(known, token, db, opener, budget)
+            # The live record is named HERE, by the one caller that should touch it.
+            known, column_problems = ensure_columns(known, token, db, opener, budget,
+                                                    record=COLUMNS_MADE)
             counts = paint(buckets, token, db, opener, budget, known=known)
             counts["column_problems"] = column_problems
         dupes = {}
--- a/q-system/.q-system/scripts/board_rows.py
+++ b/q-system/.q-system/scripts/board_rows.py
@@ -639,11 +639,20 @@ CREATABLE = tuple(n for n in WRITES_TYPE if n not in UNDROPPABLE)
 
 #: Columns this module has created before, per board. A column in here that is now
 #: ABSENT was deleted by a person, and this module does not put it back.
-COLUMNS_MADE = STATE_DIR / "board-columns-created.json"
+#: ONE literal, because the guard below is path equality against it (PR #334 reviewer
+#: round 5, minor). Written twice, renaming the file moved COLUMNS_MADE and left
+#: _LIVE_RECORD pointing at a name nothing uses. The guard would then match nothing,
+#: refuse nothing, and all 85 board tests would stay green while the protection was
+#: gone. A duplicated literal is a silent off switch.
+_RECORD_NAME = "board-columns-created.json"
+COLUMNS_MADE = STATE_DIR / _RECORD_NAME
 
-#: THE REAL LOCATION, frozen. Tests redirect `COLUMNS_MADE`; nothing redirects this, so
-#: it is what "am I about to touch the founder's live file" is measured against.
-_LIVE_RECORD = STATE_DIR / "board-columns-created.json"
+#: THE REAL LOCATION, frozen. Production reads `COLUMNS_MADE`, which a test may redirect
+#: to tmp to opt in; this one is the fixed answer to "am I about to touch the founder's
+#: live file". Two tests DO monkeypatch it, and that is the sanctioned way to exercise
+#: the guard without handing it his real path (round 5, minor: the comment here used to
+#: claim nothing redirected it, which those same two tests contradicted).
+_LIVE_RECORD = STATE_DIR / _RECORD_NAME
 
 
 def _refuse_live_record_under_test(path):
--- a/q-system/.q-system/tests/test_board_is_actionable.py
+++ b/q-system/.q-system/tests/test_board_is_actionable.py
@@ -939,6 +939,13 @@ class TestNoTestCanWriteTheLiveColumnRecord:
         br.collect(NOW, {}, token_file=str(tf), db_file=str(dbf))
         assert seen.get("record") == br.COLUMNS_MADE, seen
 
+    def test_ensure_columns_cannot_be_called_without_a_record(self):
+        """The refusal used to arrive from two frames down, as an AssertionError no
+        `collect` arm catches, instead of from the call a person was looking at
+        (round 2, minor). It is a TypeError at the call site now."""
+        with pytest.raises(TypeError):
+            br.ensure_columns({"Task": "title"}, "tok", "db")
+
     def test_reading_without_a_record_path_also_raises(self, monkeypatch):
         """The write lost its default and the comment said there was none left, while
         the READ still had one, so a caller that omitted `record=` quietly consulted
--- a/q-system/.q-system/tests/test_board_is_actionable.py
+++ b/q-system/.q-system/tests/test_board_is_actionable.py
@@ -994,7 +994,11 @@ class TestNoTestCanWriteTheLiveColumnRecord:
         with pytest.raises(AssertionError) as e:
             br.collect(NOW, {}, token_file=str(tf), db_file=str(dbf))
         assert "live column record" in str(e.value)
-        assert not live.exists(), "collect reached the live record and wrote it"
+        # NO write assertion here, deliberately (round 5, minor). The guard fires on the
+        # `_columns_made` READ at the top of `ensure_columns`, and the write is past the
+        # network stub, so `not live.exists()` could never have failed. A branch that
+        # cannot go red reads as coverage and is not. The write path is pinned by the
+        # test above, where the call reaches `_remember_columns` directly.
 
     def test_redirecting_columns_made_is_how_a_test_opts_in(self, monkeypatch, tmp_path):
         """The escape hatch is the same one LOCK_FILE already uses, so the habit is
--- a/q-system/.q-system/tests/test_board_is_actionable.py
+++ b/q-system/.q-system/tests/test_board_is_actionable.py
@@ -939,6 +939,27 @@ class TestNoTestCanWriteTheLiveColumnRecord:
         br.collect(NOW, {}, token_file=str(tf), db_file=str(dbf))
         assert seen.get("record") == br.COLUMNS_MADE, seen
 
+    def test_the_live_path_is_refused_even_when_collect_names_it(self, monkeypatch):
+        """THE REGRESSION THIS GUARD EXISTS FOR. Requiring an explicit `record=` closed
+        the direct callers and opened the one that mattered: `collect` names
+        COLUMNS_MADE itself, so any test reaching `collect` sailed through a check the
+        previous commit would have failed (round 3, major). The signal is `sys.modules`,
+        because every collect-level test here deletes PYTEST_CURRENT_TEST to get past
+        the board's own chokepoint."""
+        monkeypatch.delenv("PYTEST_CURRENT_TEST", raising=False)
+        with pytest.raises(AssertionError) as e:
+            br._remember_columns("db", ["Link"], br._LIVE_RECORD)
+        assert "live column record" in str(e.value)
+        with pytest.raises(AssertionError):
+            br._columns_made("db", br._LIVE_RECORD)
+
+    def test_redirecting_columns_made_is_how_a_test_opts_in(self, monkeypatch, tmp_path):
+        """The escape hatch is the same one LOCK_FILE already uses, so the habit is
+        one habit rather than two."""
+        monkeypatch.setattr(br, "COLUMNS_MADE", tmp_path / "made.json")
+        br._remember_columns("db", ["Link"], br.COLUMNS_MADE)
+        assert br._columns_made("db", br.COLUMNS_MADE) == {"Link"}
+
     def test_ensure_columns_cannot_be_called_without_a_record(self):
         """The refusal used to arrive from two frames down, as an AssertionError no
         `collect` arm catches, instead of from the call a person was looking at
--- a/q-system/.q-system/tests/test_board_is_actionable.py
+++ b/q-system/.q-system/tests/test_board_is_actionable.py
@@ -898,10 +898,46 @@ class TestNoTestCanWriteTheLiveColumnRecord:
     A hand fix protects the tests that exist; this protects the ones nobody has written
     yet (PR #332 reviewer round 7, minor)."""
 
-    def test_writing_without_a_record_path_raises_under_pytest(self):
+    def test_writing_without_a_record_path_always_raises(self, monkeypatch):
+        """NOT "under pytest". The first guard read PYTEST_CURRENT_TEST, and every test
+        here that reaches `collect` deletes that variable to get past the board's own
+        chokepoint, so the guard was bypassed by the exact idiom that caused the problem
+        (round 9, minor). With no ambient default there is no environment to defeat."""
+        monkeypatch.delenv("PYTEST_CURRENT_TEST", raising=False)
         with pytest.raises(AssertionError) as e:
             br._remember_columns("db", ["Link"])
-        assert "live column record" in str(e.value)
+        assert "no default" in str(e.value)
+
+    def test_production_names_the_live_path_itself(self, monkeypatch, tmp_path):
+        """The one caller allowed to touch the live record names it, so the guard above
+        cannot break production. Asserted by WATCHING the call, not by grepping the
+        source: the first version counted the string `record=COLUMNS_MADE` and found
+        two, because the guard's own error message contains that sentence. A source
+        substring is not a behaviour."""
+        seen = {}
+        monkeypatch.setattr(br, "_schema_properties",
+                            lambda *a, **k: {"Task": "title", "Item id": "rich_text"})
+
+        def spy(known, token, db, opener=None, budget=None, record=None):
+            seen["record"] = record
+            return known, ()
+
+        monkeypatch.setattr(br, "ensure_columns", spy)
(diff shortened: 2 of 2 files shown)


assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-334-20260908-020254.md` (4751 bytes).

```
FINDINGS:
minor|_columns_made keeps the `path or COLUMNS_MADE` ambient default, so a caller that omits record= still reads the live record while the new comment claims only collect can reach it|q-system/.q-system/scripts/board_rows.py:656
minor|ensure_columns' docstring says failure is not fatal, but the unwrapped _remember_columns AssertionError escapes past every except arm in collect|q-system/.q-system/scripts/board_rows.py:799
END FINDINGS
```

--- reviewer output ---

## Review: PR #334

Two findings, both minor. Production path verified correct by mutation.

### Finding 1 (minor) — the read side still has the ambient default the write side just lost

`q-system/.q-system/scripts/board_rows.py:656`

`_columns_made` is still `Path(path or COLUMNS_MADE)`. So `ensure_columns(known, tok, db)` with no `record=` reads the founder's live `~/.config/kipi/board-columns-created.json`, and its content changes what the call asks Notion for. The comment added at line 692 says "the live path is reachable only by a caller that names it, which is `collect` and nothing else", and the PR body says "The live record is reachable only through `collect`". True for the write, false for the read.

Reproducer, with a control, fake HOME holding the exact bytes the real producer writes:

```
python3 - <<'PY'
... os.environ["HOME"] = tempfile.mkdtemp()
... live.write_text(json.dumps(dict(db=["Link","Next"]), indent=1, sort_keys=True))
... br.ensure_columns(known, "tok", "db")     # no record=
PY
```

Real output:

```
COLUMNS_MADE : /var/folders/.../fakehome-c3l6jlar/.config/kipi/board-columns-created.json
write refused OK: no record path given. Production passes 
asked Notion for   : ['Bucket', 'Domain', 'Notes', 'Priority', 'Source', 'Status']
Link/Next dropped  : True

--- control: same call with an empty live record ---
asked Notion for   : ['Bucket', 'Domain', 'Link', 'Next', 'Notes', 'Priority', 'Source', 'Status']
Link/Next asked    : True
```

Severity honesty: I keyed the file with `"db"`, which is a test's database id. Production keys it with a real Notion uuid, so no current test flips on the founder's actual file. Constructed input, dropped a level. What it costs is that the next test to forget `record=` becomes machine-dependent, silently.

### Finding 2 (minor) — `ensure_columns`'s docstring is now false about a failure mode

`q-system/.q-system/scripts/board_rows.py:799` and its docstring at line 716

The docstring says "Failure is not fatal: on a refusal the schema is returned unchanged and the drop path reports it." The `_remember_columns` call at 799 is not wrapped, so a missing `record=` now raises out of `ensure_columns` after the Notion PATCH already succeeded. `collect` catches `BoardBusy`, `TimeoutError`, `Cancelled`, `MissingIdentityColumn`, `URLError`, `HTTPError`, `OSError`, `ValueError`. Not `AssertionError`.

Real output, same repro before I caught the exception:

```
File ".../board_rows.py", line 799, in ensure_columns
    _remember_columns(db, made, record)
AssertionError: no record path given. Production passes record=COLUMNS_MADE explicitly; a test passes its own tmp path. There is no default.
```

Same honesty: unreachable from production, since `collect` is the only caller and it names the path. Minor.

### What is sound

Four attacks the code survived, each executed:

- **Mutation A**, guard reverted to the `PYTEST_CURRENT_TEST` version in memory: `test_writing_without_a_record_path_always_raises` went RED with `DID NOT RAISE`. Not decoration. It also wrote `{"db": ["Link"]}` into the real `~/.config/kipi/board-columns-created.json`, which is the PR's premise demonstrated live rather than asserted.
- **Mutation B**, `record=COLUMNS_MADE` stripped from `collect`: `test_production_names_the_live_path_itself` went RED with `assert None == PosixPath(...)`. The spy watches the call, so the grep-a-substring weakness the PR body describes is genuinely gone.
- **Fresh-machine attack** on the new test: ran it with `KIPI_CONSULTING_ROOT=/tmp/rev334-noconsult`, so `consulting_board.buckets` reads nothing. Still passed. It does not wolf-cry on an instance without the consulting repo.
- **Live-state attack**: full suite, 1393 passed 1 skipped in 211s, and `~/.config/kipi/board-columns-created.json` does not exist afterward. Matches the PR's proof claim exactly.

One disclosure: Mutation A wrote that live file during my run. I read it, confirmed it held only the injected `{"db": ["Link"]}` and no real Notion database id, and removed it. Live state is back to absent.

First thing to fix: make `_columns_made` refuse a `None` path the same way `_remember_columns` does, so the read and the write have the same chokepoint and the comment at line 692 becomes true.

FINDINGS:
minor|_columns_made keeps the `path or COLUMNS_MADE` ambient default, so a caller that omits record= still reads the live record while the new comment claims only collect can reach it|q-system/.q-system/scripts/board_rows.py:656
minor|ensure_columns' docstring says failure is not fatal, but the unwrapped _remember_columns AssertionError escapes past every except arm in collect|q-system/.q-system/scripts/board_rows.py:799
END FINDINGS

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-334-20260908-022019.md` (5513 bytes).

```
FINDINGS:
minor|collect's live-board write is still gated only on PYTEST_CURRENT_TEST while _credentials falls back to the real token and a hardcoded live database id|q-system/.q-system/scripts/board_rows.py:1040
minor|ensure_columns still defaults record=None and its docstring still promises failure is not fatal, but a missing record raises an AssertionError no collect handler catches|q-system/.q-system/scripts/board_rows.py:735
END FINDINGS
```

--- reviewer output ---

Review done. Read-only, nothing in the repo modified.

## Findings

**MINOR — the board's own live-write chokepoint is still the environment variable the suite must delete, and the credentials behind it are the founder's real ones.**

`q-system/.q-system/scripts/board_rows.py:1040` (guard) and `:112-117` (`_credentials`).

The PR's thesis is that an env-var guard is defeated by the idiom every test must use. It closes that for the record file. `collect` itself still refuses only on `PYTEST_CURRENT_TEST`, and `_credentials` falls back to the real token file plus a hardcoded live database id, so a future test that deletes the variable and forgets to patch `_request` paints the founder's live board. The new test in this diff is the third place in the suite that deletes that variable.

Reproducer, run with fake credentials so no packet reached Notion:

```
python3 - (intercepts urllib.request.urlopen, deletes PYTEST_CURRENT_TEST,
           calls br.collect with a tmp token/db)
```

Real output:

```
collect returned err: board write failed: URLError: <urlopen error intercepted: no packet actually left this machine>
OUTBOUND CALLS THE LIVE BOARD WOULD HAVE RECEIVED:
   GET https://api.notion.com/v1/databases/FAKE-DB-ID b''
   POST https://api.notion.com/v1/databases/FAKE-DB-ID/query b'{"page_size": 100, "filter": {"property": "Item id", "rich_text": {"starts_with": "cb:"}}}'
```

And the default credential resolution, no network at all:

```
token resolved from /Users/assafkipnis/.config/kipi/notion-token -> PRESENT len=50
db resolved        -> 0a09bd16-b12e-49bf-a792-fad15e008ed0 (hardcoded DEFAULT_DB: True )
```

Honest downgrade: no test in the repo does this today, so nothing fires at 3am. The producer is a test nobody has written yet, which is the same argument the PR makes for the record file. That is why it is minor and not major.

**MINOR — `ensure_columns` keeps `record=None` as a legal default while its docstring still promises failure is not fatal.**

`q-system/.q-system/scripts/board_rows.py:725` and `:735`.

A missing record now raises `AssertionError` from `_columns_made` at line 746, before the `try`. None of `collect`'s four except arms catch it, so the whole board section dies rather than degrading. `_remember_columns`'s docstring was updated to say this; `ensure_columns`'s was not, and its signature advertises the fatal call as ordinary.

Reproducer and real output:

```
1. signature: (known, token, db, opener=None, budget=None, record=None)
   raised: AssertionError | no record path given. Production passes record=COLUMNS_MADE
2. collect except arms:
    except BoardBusy as exc:
    except (TimeoutError, Cancelled) as exc:
    except MissingIdentityColumn as exc:
    except (urllib.error.URLError, urllib.error.HTTPError, OSError, ValueError) as exc:
3. ensure_columns docstring line:
    Failure is not fatal: on a refusal the schema is returned unchanged and the drop
```

## What is sound

- **The new test is not decoration.** I copied the tree to a temp dir, deleted `record=COLUMNS_MADE` from line 1084, and ran the class. `test_production_names_the_live_path_itself` went red with `assert None == PosixPath('/Users/assafkipnis/.config/kipi/board-columns-created.json')`. The other three stayed green. The assertion watches the call, not the source string, exactly as the PR body claims.
- **Hermeticity against a foreign machine.** The new test runs the real `collect`, which calls `consulting_board.buckets` unstubbed against `~/projects/consulting`. I re-ran the file with `HOME` pointed at an empty directory: 81 passed. `buckets` returns `error: None` on every path and degrades into alarm rows, so an absent consulting tree does not change the outcome.
- **That unstubbed call writes nothing.** `grep -n "write_text|open(...w|mkdir|json.dump" consulting_board.py` returns no matches. The live read is read-only.
- **The `-O` hole is not present.** Both guards are explicit `raise AssertionError`, not bare `assert`, so `python3 -O` does not strip them.
- **The "reachable only through collect" claim holds.** Repo-wide grep finds exactly one caller of `ensure_columns` outside tests, at line 1084, and no caller of `_columns_made` or `_remember_columns` outside `ensure_columns`. No other script reads or writes `board-columns-created.json`.
- **Full suite, my run, not the author's:** `1394 passed, 1 skipped in 211.56s`. The PR body says 1393; the extra case is the follow-up commit `ff92a1a3` in the same PR.

## VERDICT: APPROVE WITH NITS

Only minor findings survived reproduction. Nothing here is permanent or unrecoverable, and the change is strictly stronger than what it replaces.

First thing to fix: finding 1. The record file is now structurally protected while the board itself, which is the higher-value object, is still guarded by the variable the suite deletes. Passing an explicit opener or credential handle down through `collect` the way `record=` was just threaded would close it with the same shape that already worked here.

FINDINGS:
minor|collect's live-board write is still gated only on PYTEST_CURRENT_TEST while _credentials falls back to the real token and a hardcoded live database id|q-system/.q-system/scripts/board_rows.py:1040
minor|ensure_columns still defaults record=None and its docstring still promises failure is not fatal, but a missing record raises an AssertionError no collect handler catches|q-system/.q-system/scripts/board_rows.py:735
END FINDINGS

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-334-20260908-023738.md` (5642 bytes).

```
FINDINGS:
major|collect names COLUMNS_MADE, so the record guard can never fire for a test that reaches collect; the base commit refused that same test|q-system/.q-system/scripts/board_rows.py:1096
END FINDINGS
```

--- reviewer output ---

Review done. Full suite ran green, three repros executed, scratch removed, tree clean.

## Finding 1

**MAJOR.** `collect` now names the live record itself, so the "no default" refusal can never fire for any test that goes through `collect` — the exact shape the guard exists for. On the base commit that same test was refused.

`q-system/.q-system/scripts/board_rows.py:1096` (the claim it breaks is at `board_rows.py:686` and `:693-706`)

Reproducer (written to a scratch dir inside the worktree, HOME redirected so nothing real is touched):

```python
# test_opener_path.py  -- PYTEST_CURRENT_TEST is NOT deleted; pytest sets it for real
os.environ["HOME"] = str(FAKE_HOME)
sys.path.insert(0, ".../q-system/.q-system/scripts")
import board_rows as br

def test_collect_with_an_opener_under_pytest(monkeypatch, tmp_path):
    bare = {"Task": "title", "Item id": "rich_text"}
    monkeypatch.setattr(br, "_schema_properties", lambda *a, **k: dict(bare))
    after = {"properties": dict({n: {"type": t} for n, t in bare.items()},
             **{n: {"type": br.WRITES_TYPE[n]} for n in br.CREATABLE})}
    monkeypatch.setattr(br, "_request", lambda *a, **k: after)   # the suite's own idiom
    monkeypatch.setattr(br, "paint", lambda *a, **k: {...zeros...})
    monkeypatch.setattr(br, "existing_rows", lambda *a, **k: {})
    br.collect(NOW, {}, opener=object(), token_file=str(tf), db_file=str(dbf))
    assert not br.COLUMNS_MADE.exists()
```

Run: `python3 -m pytest .pr334-scratch/test_opener_path.py -q -s`

Real output, PR head:

```
LIVE RECORD EXISTS: True
LIVE RECORD CONTENT: {"the-live-board": ["Bucket","Domain","Link","Next","Notes","Priority","Source","Status"]}
E   AssertionError: a test wrote the live column record
1 failed in 0.02s
```

Same file, same test, base commit `b8901176` (`.pr334-scratch/old/board_rows.py` first on sys.path):

```
    if path is None and os.environ.get("PYTEST_CURRENT_TEST"):
>       raise AssertionError(
E       AssertionError: a test tried to write the live column record; pass record=<tmp path>
.pr334-scratch/old/board_rows.py:679: AssertionError
```

The base refused the write. The head performs it. Passing `opener` is how a test gets past `collect`'s own pytest refusal at `board_rows.py:1053`, and once past it, `record=COLUMNS_MADE` is supplied by production code, so the new guard has nothing left to refuse.

Blast radius. `_credentials` at `board_rows.py:112` falls back to `DEFAULT_DB = "0a09bd16-..."`, the real Kipi backlog id, when a test omits `db_file`. So the poisoned key is the real board's key, not a fake one. The 07:40 job then reads that record at `board_rows.py:759`, treats every listed column as a founder deletion, and stops creating them. `Notes` and `Bucket` say so on the line via `STRUCTURAL_COST`; the other six surface only as a dropped-column report. Cleanup is deleting `~/.config/kipi/board-columns-created.json`, which is what happened by hand once already in #332.

Second repro, for completeness: a test that deletes `PYTEST_CURRENT_TEST` and stubs `_request` also writes the live record (`LIVE RECORD EXISTS: True`). That one behaves identically on base and head, so it is not a regression, but it does mean the class name `TestNoTestCanWriteTheLiveColumnRecord` and the comment "the live path is reachable only by a caller that names it, which is `collect` and nothing else" are false as written. A test that calls `collect` is such a caller.

## What is sound

I attacked five things that held.

**The new production test is not decoration.** I mirrored `scripts/` and `tests/` into a scratch tree, mutated only `collect` to pass `record=Path('/tmp/somewhere-else.json')`, and ran the real test file against the mutant:

```
E   AssertionError: {'record': PosixPath('/tmp/somewhere-else.json')}
E   assert PosixPath('/tmp/somewhere-else.json') == PosixPath('/Users/.../board-columns-created.json')
1 failed, 1 passed, 80 deselected in 0.08s
```

My first mutation attempt via `PYTHONPATH` passed, which was a false green: `test_board_is_actionable.py:22` does `sys.path.insert(0, SCRIPTS)`, so `PYTHONPATH` never wins. Mirroring the tree is what made it real. The PR's own note about replacing a source-grep with a watched call is correct, and the replacement has teeth.

**Keyword-only `record` breaks no caller.** Grepped the whole tree: one copy of `board_rows.py`, one production call site, every test call passes `record=`. The five `monkeypatch.setattr(br, "ensure_columns", ...)` stubs all take `**k`.

**Concurrency on the record file.** The read-modify-write in `_remember_columns` sits inside `exclusive()` at `board_rows.py:1082`, so two painters cannot lose an update to it.

**Full suite.** `python3 -m pytest q-system/.q-system/tests/ -q` gave `1395 passed, 1 skipped in 211.18s`. The PR body says 1393; the delta is other commits, not a failure.

**`_columns_made` raising on a missing path.** I looked for a production path that reaches it with `path=None` and found none. Dropped as a finding.

## VERDICT: REQUEST CHANGES

One major, recoverable, so not a BLOCK.

Fix first: keep an environment arm as an addition rather than a replacement. `_remember_columns` can refuse when the path it was handed IS `COLUMNS_MADE` and `PYTEST_CURRENT_TEST` is set, on top of the new "no default" rule. That covers the caller that names the path, which the current version explicitly does not, and it makes the class name true.

FINDINGS:
major|collect names COLUMNS_MADE, so the record guard can never fire for a test that reaches collect; the base commit refused that same test|q-system/.q-system/scripts/board_rows.py:1096
END FINDINGS

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-334-20260908-030515.md` (5839 bytes).

```
FINDINGS:
minor|The ensure_columns spy accepts `record` positionally while the real function is keyword-only, so a positional-call regression stays green and kills the board section with "board_rows failed (TypeError)"|q-system/.q-system/tests/test_board_is_actionable.py:921
minor|The guard test hands the founder's real ~/.config/kipi/board-columns-created.json to the code under test, so a guard regression makes the test write that file instead of failing safe|q-system/.q-system/tests/test_board_is_actionable.py:951
nit|test_the_live_path_is_refused_even_when_collect_names_it never calls collect; no test exercises the guard through the collect path|q-system/.q-system/tests/test_board_is_actionable.py:942
END FINDINGS
```

--- reviewer output ---

# Review: PR #334 (ASK-1346)

The change is small, defensive and does what its title says. I attacked the guard four ways and it held in three. Two minors and one nit survived reproduction.

## First, something I broke and you should know about

While mutation-testing the guard I ran the suite with the guard neutered and **without** isolating `HOME`. That run wrote your real `~/.config/kipi/board-columns-created.json`.

State now:

- Content is `{"db": ["Link"]}`. The key `db` is the test literal, not a Notion database id, so `_columns_made` will never match it on a real board. It is inert, not harmful.
- Merge, not overwrite: any real entries would have survived. The file holding only `db` means it was absent or empty before my run.
- Restore, if you want it clean: `python3 -c "import pathlib,json; p=pathlib.Path.home()/'.config/kipi/board-columns-created.json'; d=json.loads(p.read_text()); d.pop('db',None); p.write_text(json.dumps(d,indent=1,sort_keys=True))"`

That accident is also evidence for finding 2 below.

## Finding 1, minor: the spy is looser than the function it stands in for

`q-system/.q-system/tests/test_board_is_actionable.py:921`. The spy signature is `def spy(known, token, db, opener=None, budget=None, record=None)`. The real `ensure_columns` is `*, record`. So the one test pinning "production names the live path" cannot see a call the real function rejects.

Reproducer, mutating `collect` to pass the record positionally in a temp copy:

```
MUTATION: collect passes the record POSITIONALLY instead of record=
84 passed in 0.09s
--- what the REAL ensure_columns does with that exact call ---
TypeError: ensure_columns() takes from 3 to 5 positional arguments but 6 were given
```

Consequence at 07:40: `TypeError` is not in `collect`'s except arms, so `_guarded` catches it and the brief prints `board_rows failed (TypeError)`. Board section dead, type name only, which is the class-name-instead-of-help shape this file refuses everywhere else. Fix is one word: make the spy `def spy(known, token, db, opener=None, budget=None, *, record)`.

## Finding 2, minor: the guard test's failure mode is destructive, not safe

`q-system/.q-system/tests/test_board_is_actionable.py:951`. The test deliberately hands `br._LIVE_RECORD` to the code under test and relies on the subject to refuse. When the guard regresses, the test does not fail safe. It writes the founder's live record and then reports a missing exception.

Reproducer with an isolated `HOME` and the guard neutered:

```
BEFORE: { "real-notion-db-id": [  "Notes" ]}
pytest: 1 failed in 0.04s
AFTER:  { "db": [  "Link" ], "real-notion-db-id": [  "Notes" ]}
```

Dropped a level per the calibration rule: the trigger is a constructed guard regression, not an input a producer emits today. It is still the exact shape this PR argues against, a test reaching for a live path and trusting the subject to stop it. Fix: `monkeypatch.setattr(br, "_LIVE_RECORD", tmp_path / "made.json")` and pass that. The guard is pure path equality, so the logic under test is identical and a regression writes tmp instead of home.

## Finding 3, nit: the name promises `collect` and the body never calls it

`q-system/.q-system/tests/test_board_is_actionable.py:942`. `test_the_live_path_is_refused_even_when_collect_names_it` calls `_remember_columns` and `_columns_made` directly. Nothing in the suite exercises the guard through `collect`. That axis is real and it does work, I proved it below, it just is not pinned.

## What is sound

- **The round-3 regression is genuinely closed.** A test that reaches `collect` with the real `ensure_columns` and no redirect raises from `_columns_made` before any write, and the live record stays absent. Verified end to end with an isolated `HOME`: `AssertionError: a test reached the live column record at .../fakehome/.config/kipi/board-columns-created.json`, and `HOME live record on disk: False`.
- **The documented escape hatch actually works through `collect`.** `monkeypatch.setattr(br, "COLUMNS_MADE", tmp_path / "made.json")` routes the write to tmp and leaves `_LIVE_RECORD` untouched. The error message's advice is correct, not aspirational.
- **The guard is mutation-covered.** Neutering `_refuse_live_record_under_test` turns the file red, one failure. It is not decoration.
- **`sys.modules` is safe for production.** `board_rows.py` has no `__main__`, it is loaded by `morning-brief.py` as a registry section, and nothing shells it out. No production path imports pytest.
- **Concurrency.** `ensure_columns` runs inside `exclusive()`, so the read-modify-write on the record file is serialized with the paint. Two 07:40 runs do not interleave here.
- **Full suite green, at repo scope, not directory scope:** `1397 passed, 1 skipped in 210.98s`. The board file alone: `84 passed`.
- **Rollback is clean.** Pure code, no migration, no persisted format change.

## VERDICT: APPROVE WITH NITS

Fix first: the spy signature at `test_board_is_actionable.py:921`. It is the one that lets a real morning-brief break through with a green suite.

FINDINGS:
minor|The ensure_columns spy accepts `record` positionally while the real function is keyword-only, so a positional-call regression stays green and kills the board section with "board_rows failed (TypeError)"|q-system/.q-system/tests/test_board_is_actionable.py:921
minor|The guard test hands the founder's real ~/.config/kipi/board-columns-created.json to the code under test, so a guard regression makes the test write that file instead of failing safe|q-system/.q-system/tests/test_board_is_actionable.py:951
nit|test_the_live_path_is_refused_even_when_collect_names_it never calls collect; no test exercises the guard through the collect path|q-system/.q-system/tests/test_board_is_actionable.py:942
END FINDINGS

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-334-20260908-032446.md` (4670 bytes).

```
FINDINGS:
minor|`_LIVE_RECORD` duplicates the `COLUMNS_MADE` path literal, so renaming the record file silently disables the guard while all 85 board tests stay green|q-system/.q-system/scripts/board_rows.py:646
minor|`assert not live.exists()` in test_the_guard_fires_through_collect_itself has an unreachable false branch: the `_request` stub raises before any write can happen|q-system/.q-system/tests/test_board_is_actionable.py:997
minor|Comment claims nothing redirects `_LIVE_RECORD` while two tests added in this same PR monkeypatch it|q-system/.q-system/scripts/board_rows.py:644
END FINDINGS
```

--- reviewer output ---

## Review: PR #334 (ASK-1346)

Read the diff, read `board_rows.py` around `collect`, ran the board suite, and mutation-tested the new guard in a sandboxed copy under `$TMPDIR` (repo untouched).

### What is sound

Attacks I tried that the code survived:

- **Production behavior drift.** Before: `record=None` fell through to `Path(None or COLUMNS_MADE)`. After: `work` passes `record=COLUMNS_MADE` explicitly. Same file, same bytes. `board_rows.py:1130`. No 07:40 behavior change.
- **Is the guard decoration?** No. Neutering `_refuse_live_record_under_test` (`if True: return`) turns two tests red: `2 failed, 83 passed`. The tests can fail for the reason they claim.
- **Does the guard actually refuse the live path?** Ran it with a fake `HOME` so the "live" record was a sandbox file. Refused, and the file was not created.
- **Other callers.** `ensure_columns` / `_columns_made` / `_remember_columns` have exactly one production caller (`work`). No other script imports them. Keyword-only `record` breaks nothing else.
- **`-O` bypass.** These are explicit `raise AssertionError`, not `assert` statements, so `python3 -O` does not strip them.
- **Suite.** `python3 -m pytest q-system/.q-system/tests/ -q` exited 0.

### Finding 1 (minor): `_LIVE_RECORD` copies the literal instead of deriving from `COLUMNS_MADE`, so a one-line rename silently disables the guard with the suite still green

`q-system/.q-system/scripts/board_rows.py:646`

Reproducer, in a sandbox copy with `COLUMNS_MADE` renamed to `board-columns-created-v2.json` and `_LIVE_RECORD` left alone:

```
python3 -m pytest .../sandbox/tests/test_board_is_actionable.py -q
85 passed in 0.08s

python3 repro_m2.py .../sandbox/scripts .../h2
COLUMNS_MADE : .../h2/.config/kipi/board-columns-created-v2.json
_LIVE_RECORD : .../h2/.config/kipi/board-columns-created.json
RESULT: NOT REFUSED -- a test wrote the live record
live record on disk: True {"db": ["Link"]}
```

Same script against the PR as shipped:

```
RESULT: refused -> a test reached the live column record at /var/folders/...
live record on disk: False
```

`_LIVE_RECORD = COLUMNS_MADE` (or a test asserting the two are equal) closes it. Dropping this to minor on purpose: it needs a future edit, not any input a producer emits today.

### Finding 2 (minor): the exists-assertion in `test_the_guard_fires_through_collect_itself` cannot fail

`q-system/.q-system/tests/test_board_is_actionable.py:997`

`assert not live.exists(), "collect reached the live record and wrote it"` claims to observe a write. In that test `_request` is stubbed to raise, so `ensure_columns` never reaches `_remember_columns`, so nothing can write `live` on any path.

Proof: guard fully neutered, message assert removed, only the exists-assert left:

```
collect raised: a test tried to reach Notion
1 passed, 84 deselected in 0.06s
```

Guard gone, assertion still green. The whole discriminating power of that test is the message substring on line 996. The write path through `collect` stays unexercised.

### Finding 3 (minor): the comment says nothing redirects `_LIVE_RECORD`, and two tests in this same PR redirect it

`q-system/.q-system/scripts/board_rows.py:644`

Comment: "Tests redirect `COLUMNS_MADE`; nothing redirects this, so it is what 'am I about to touch the founder's live file' is measured against."

Code in the same diff:

```
test_board_is_actionable.py:962  monkeypatch.setattr(br, "_LIVE_RECORD", tmp_path / "live.json")
test_board_is_actionable.py:981  monkeypatch.setattr(br, "_LIVE_RECORD", live)
```

That is the sentence a future reader will trust when deciding the guard is unbypassable. It is false as written. Reword to what is true: nothing in production redirects it, and a test that redirects it is opting out.

### Verdict

**APPROVE WITH NITS.**

Nothing here is unrecoverable, nothing cries wolf, and the production path is byte-identical to before. The guard is real and mutation-proven.

Most important thing to fix first: **Finding 1.** Make `_LIVE_RECORD = COLUMNS_MADE` so the guard cannot be disabled by a rename.

FINDINGS:
minor|`_LIVE_RECORD` duplicates the `COLUMNS_MADE` path literal, so renaming the record file silently disables the guard while all 85 board tests stay green|q-system/.q-system/scripts/board_rows.py:646
minor|`assert not live.exists()` in test_the_guard_fires_through_collect_itself has an unreachable false branch: the `_request` stub raises before any write can happen|q-system/.q-system/tests/test_board_is_actionable.py:997
minor|Comment claims nothing redirects `_LIVE_RECORD` while two tests added in this same PR monkeypatch it|q-system/.q-system/scripts/board_rows.py:644
END FINDINGS

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-334-20260908-034208.md` (6633 bytes).

```
FINDINGS:
minor|Monkeypatching _LIVE_RECORD disables the guard entirely and lets a test write the live column record|q-system/.q-system/scripts/board_rows.py:655
minor|collect's live-board chokepoint still keys on PYTEST_CURRENT_TEST, the variable this PR's own new tests delete|q-system/.q-system/scripts/board_rows.py:1097
nit|The write assertion after pytest.raises is unreachable when the guard regresses, so it can never go red|q-system/.q-system/tests/test_board_is_actionable.py:969
END FINDINGS
```

--- reviewer output ---

# Review: PR #334 — board record chokepoint

Read-only. Nothing in the repo was modified. Full suite run once at baseline, then three in-memory mutations against a copy of the module with `HOME` redirected to a scratch dir.

## Baseline

`python3 -m pytest q-system/.q-system/tests/ -q` → `1398 passed, 1 skipped in 212.23s`.

## Finding 1 — minor

**The guard's own reference point is monkeypatchable, and the file blesses patching it. A test that redirects `_LIVE_RECORD` and then names `COLUMNS_MADE` writes the live record with no refusal.**

`q-system/.q-system/scripts/board_rows.py:655`

The guard is path equality against module-global `_LIVE_RECORD`. The comment at 655 calls redirecting it "the sanctioned way to exercise the guard." Once redirected, the live path matches nothing and every refusal is off, including on the write. The docstring at 730 says "THE LIVE RECORD IS NEVER WRITTEN BY A TEST, and that is a chokepoint rather than a rule someone remembers." It is still a rule someone remembers: remember not to move the yardstick.

Reproducer, run under pytest with `HOME` pointed at a scratch dir:

```
python3 -m pytest /var/folders/.../pr334home/repro.py -q -s
```
```
REPRO A wrote: /var/folders/.../pr334home/.config/kipi/board-columns-created.json
REPRO A content: {
2 passed in 0.01s
```

The test body is three lines: `monkeypatch.setattr(br, "_LIVE_RECORD", tmp_path / "elsewhere.json")`, then `br._remember_columns("db1", ["Link"], br.COLUMNS_MADE)`, then assert the real record exists. It does.

Blast radius if a future test copies the pattern: same as the #332 incident, the 07:40 job reads columns the founder never deleted and stops offering them. Recoverable by deleting one JSON file. No test in the repo does this today, so per the calibration rule I dropped it a level.

One shape that closes it: compute the live path inside the guard from `STATE_DIR` rather than reading a module global, and give the two tests that need an opt-in an explicit allow-token instead.

## Finding 2 — minor

**The higher-stakes sibling chokepoint in the same file still keys on `PYTEST_CURRENT_TEST`, the exact variable this PR's own new tests delete.**

`q-system/.q-system/scripts/board_rows.py:1097`

The PR body's argument is "the guard was bypassed by the exact idiom that caused the problem it was written for." That argument applies unchanged to line 1097, which refuses live Notion writes. This PR adds two more tests that call `monkeypatch.delenv("PYTEST_CURRENT_TEST", raising=False)` and then `collect`, so the idiom gets more common, not less.

Reproducer output, same file:

```
REPRO B with the env var: refused: running under pytest; the live board is never written by a test
REPRO B without it: board write failed: OSError: network boundary reached
REPRO B network reached: {'url': 'https://api.notion.com/v1/databases/d/query'}
```

With the variable deleted, `collect` walks past the refusal and reaches `urllib.request.urlopen`. A test that deletes it and calls `collect()` with no `token_file` reads the real `~/.config/kipi/notion-token` and `DEFAULT_DB`, and writes rows to the live board. Notion rows are archivable, so it is cleanup rather than permanent loss.

Pre-existing, not introduced here. I am reporting it because the PR is specifically the fix for this class and the title says "no environment left to defeat," which is true of the record and not of the board four hundred lines below.

## Finding 3 — nit

**The write assertion at `test_board_is_actionable.py:969` cannot go red, which is the thing the sibling test's comment refuses.**

`q-system/.q-system/tests/test_board_is_actionable.py:969`

`assert not (tmp_path / "live.json").exists()` sits after a `pytest.raises` block. When the guard regresses, the `raises` fails first and line 969 never runs. Measured, guard replaced with a no-op:

```
E       Failed: DID NOT RAISE <class 'AssertionError'>
q-system/.q-system/tests/test_board_is_actionable.py:964: Failed
```

Line 964, not 969. `test_the_guard_fires_through_collect_itself` explicitly removed exactly this shape and left a comment saying why. This one kept it.

## What is sound

**The guard is load-bearing, both halves.** Replacing `_refuse_live_record_under_test` with a no-op turns 2 of 85 red. Reverting the `sys.modules` signal to the `PYTEST_CURRENT_TEST` signal it replaced turns the same 2 red. Not decoration.

**The spy test can see a regression.** `test_production_names_the_live_path_itself` uses a keyword-only `record` on the spy, so a `collect` that dropped the argument raises `TypeError`, which no `collect` arm catches, so the test fails rather than passing on a loose stub.

**No production caller reaches the `path is None` branch.** `ensure_columns` is called from exactly one place, `work` at line 1141, with `record=COLUMNS_MADE`. Grep across the repo for `ensure_columns`, `_columns_made`, `_remember_columns` finds callers only in `board_rows.py` and its test file.

**No production path imports pytest.** `import pytest` appears only in test files and `scripts/conftest.py`. The `sys.modules` check does not fire at 07:40, so the new `AssertionError` cannot kill the board section in production.

**`assert` stripping is not a hole.** The guard raises `AssertionError` explicitly rather than using the `assert` statement, so `python -O` does not remove it.

**`_LIVE_RECORD` is frozen at import.** A test that changes `HOME` after import cannot move it, which is the property the freeze was for.

**One dropped finding.** I expected existing `collect`-level tests in `test_consulting_board.py` to start crashing on the new read-side guard. They do not: the timeout tests cancel inside `_schema_properties` before `ensure_columns` runs. Could not make it fail, so I dropped it.

## VERDICT: APPROVE WITH NITS

Only minor and nit findings survived reproduction.

**First thing to fix: Finding 1.** The guard compares against a global any test can move, and the file documents moving it as sanctioned. Deriving the live path inside the guard costs one line and removes the off switch.

FINDINGS:
minor|Monkeypatching _LIVE_RECORD disables the guard entirely and lets a test write the live column record|q-system/.q-system/scripts/board_rows.py:655
minor|collect's live-board chokepoint still keys on PYTEST_CURRENT_TEST, the variable this PR's own new tests delete|q-system/.q-system/scripts/board_rows.py:1097
nit|The write assertion after pytest.raises is unreachable when the guard regresses, so it can never go red|q-system/.q-system/tests/test_board_is_actionable.py:969
END FINDINGS

PR #335, opened 2026-09-08
fix(board): the brief reads the card thirty minutes before it is written

assafkip:
fix(board): the brief reads the card thirty minutes before it is written

A P0 row saying "Your book: COULD NOT READ" was painted onto the founder's board every
single morning, and nothing ever cleared it. He asked for a board where every row leads
to an action. This one led to waiting for a job that was already scheduled to run later.

THE CLOCK, not the reader. `com.kipi.morning-brief` runs at 07:00 PT.
`io.askconsulting.ask-crm-state-card` writes the card at 07:30 PT. The freshness rule
was date equality, so at 07:00 the card is ALWAYS stamped yesterday, always refused, and
the painter always adds the alarm row. It could never have been satisfied. Refusing the
newest card that exists is not a refusal, it is a clock error.

MEASURED, not reasoned: morning-brief.out.log carries the refusal on the 09-05 and 09-07
runs, and today-card.md was last written 09-07 07:30. The row survives all day because
the hourly repaint has no scheduler.

WHAT THE OLD RULE PROTECTED IS KEPT. Yesterday's book must never be presented as
today's, so the brief now labels it: "yesterday's card, today's is written at 07:30".
And after 07:30 a card still stamped yesterday is a genuine failure and still refused.
That is why the window keys on CARD_WRITTEN_AT rather than on a count of hours: a
26-hour tolerance would have swallowed a card the 07:30 job actually failed to write.

THE REPO BELIEVED THE OPPOSITE, and the belief was load-bearing.
test_no_hourly_slot_fires_before_the_card_it_mirrors called the brief's own slot "this
repo's record of after the card is written". It is not. That docstring is corrected
rather than deleted, because the ordering it enforces is still worth enforcing.

VERIFIED by reverting only the new window on a copy, keeping the new tests: the four
that describe the new behaviour fail, and the three that describe unchanged behaviour
stay green. A boundary test pins 07:29 against 07:30 so the constant decides and not an
accident of two chosen timestamps, and a plist test pins the constant against the loaded
job rather than restating 07:30 as a second copy of the schedule.

NOT FIXED HERE, and captured as sp-642d380b: the brief is still SCHEDULED before the
card, so it paints from a day-old book. Correcting the order moves when the founder
receives his brief, across two repos, which is his call and not a silent edit.

[no-issue: Linear MCP is unauthenticated in this session, so the DoR is captured through
the repo's own credential-boundary path, linear-queue.py, as
kipi-system/the-morning-brief-reads-the-state-card-30-minutes-before-it-is-written]

### DIFF
--- a/q-system/.q-system/scripts/board_rows.py
+++ b/q-system/.q-system/scripts/board_rows.py
@@ -59,10 +59,14 @@ because a token with no db file is a configured board, not an unconfigured one.
 
 ## A stale source writes NOTHING
 
-If `consulting_board.buckets` reports an error (the state card is yesterday's, or the
-07:30 job crashed), this writes no rows at all and returns that error. Mirroring a stale
-source onto a board that looks fresh is the one failure that would make him act on a
-wrong number.
+If `consulting_board.buckets` reports an error (the 07:30 job crashed, or it has run and
+the card is STILL yesterday's), this writes no rows at all and returns that error.
+Mirroring a stale source onto a board that looks fresh is the one failure that would
+make him act on a wrong number.
+
+Before 07:30 a card stamped yesterday is NOT an error, because the brief runs at 07:00
+and yesterday's is the newest one there is. Those rows paint, and each one carries "from
+yesterday's card" in its detail so the board never looks fresher than its source.
 """
 from __future__ import annotations
 
--- a/q-system/.q-system/scripts/board_rows.py
+++ b/q-system/.q-system/scripts/board_rows.py
@@ -64,9 +64,11 @@ the card is STILL yesterday's), this writes no rows at all and returns that erro
 Mirroring a stale source onto a board that looks fresh is the one failure that would
 make him act on a wrong number.
 
-Before 07:30 a card stamped yesterday is NOT an error, because the brief runs at 07:00
-and yesterday's is the newest one there is. Those rows paint, and each one carries "from
-yesterday's card" in its detail so the board never looks fresher than its source.
+Before 07:30 a card stamped yesterday is NOT an error, because yesterday's is then the
+newest one there is. The committed brief slot is 07:40, AFTER the card, so on the shipped
+schedule this never happens; it happens on a hand run, or on a machine whose job has
+drifted early, which is what it was built for. Those rows paint, and each one carries
+"from yesterday's card" in its detail so the board never looks fresher than its source.
 """
 from __future__ import annotations
 
--- a/q-system/.q-system/scripts/consulting_board.py
+++ b/q-system/.q-system/scripts/consulting_board.py
@@ -80,10 +80,18 @@ MAX_CLIENT_ROWS = 6
 #: ~/Library/LaunchAgents said 07:00, forty minutes early, and never picked up the
 #: committed 07:40. Measured in ~/.config/kipi/logs/morning-brief.out.log: the refusal
 #: fired on the 09-05 and 09-07 runs, and today-card.md was last written 09-07 07:30.
-#: The cost was a P0 "Your book: COULD NOT READ" row painted on his board every morning
-#: and cleared by nothing, because the hourly repaint has no scheduler. The drift is
-#: fixed by reinstalling the template; nothing detects the NEXT one, which is
-#: sp-de7afcff.
+#: The cost was a P0 "Your book: COULD NOT READ" row painted on his board every morning.
+#:
+#: I SAID IT WAS "CLEARED BY NOTHING" AND THAT WAS WRONG (PR #335 reviewer round 3,
+#: minor). `com.kipi.morning-inbox` commits twelve repaints a day, 08:05 to 19:05, and
+#: the 08:05 one archives that row. What was true is narrower and worse: that job was
+#: not installed on his machine AT ALL, no plist in ~/Library/LaunchAgents and no
+#: launchctl entry, so on his machine nothing repainted and the row stood all day. I
+#: read the machine, the reviewer read the repo, and the gap between them is the
+#: defect twice over. Both are installed now.
+#:
+#: TWO JOBS, ONE FAILURE MODE: the brief drifted to a stale schedule and the repaint was
+#: missing outright. Nothing detects either, which is sp-de7afcff.
 #:
 #: THIS WINDOW IS NOT THAT FIX and does not pretend to be. It is what an early run
 #: should DO: use the newest card there is and label it, rather than paint a P0 he
@@ -342,6 +350,15 @@ def read_heartbeat(now: dt.datetime, paths=None) -> tuple[dict, str | None]:
         beat = json.loads(path.read_text(encoding="utf-8"))
     except (OSError, ValueError) as exc:
         return {}, f"could not read {path.name}: {exc}"
+    if not isinstance(beat, dict):
+        return {}, f"{path.name} is not an object"
+    # A CONTROL FLAG MUST NOT BE READABLE FROM THE THING IT JUDGES (PR #335 reviewer
+    # round 3, nit). This function SETS `card_is_yesterdays` below and callers trust it
+    # to label the board. It is also a key in a file this function parses, so a
+    # heartbeat carrying it would have labelled a perfectly fresh card as yesterday's,
+    # on the Slack line and on every client row. Stripped on the way in, so the only
+    # writer is the verdict twenty lines down.
+    beat.pop("card_is_yesterdays", None)
 
     if beat.get("crash"):
         return beat, f"the 07:30 state card crashed: {beat['crash']}"
--- a/q-system/.q-system/scripts/consulting_board.py
+++ b/q-system/.q-system/scripts/consulting_board.py
@@ -366,7 +366,14 @@ def read_heartbeat(now: dt.datetime, paths=None) -> tuple[dict, str | None]:
         return beat, (f"the {CARD_WRITTEN_AT.strftime('%H:%M')} state card crashed: "
                       f"{beat['crash']}")
 
-    stamped = (beat.get("card") or {}).get("date")
+    # A MALFORMED `card` VALUE IS A REFUSAL, NOT A CRASH (PR #335 reviewer round 6,
+    # nit). The isinstance check above proved `beat` is an object and then this line
+    # called .get on whatever `card` happened to be, so a heartbeat carrying a string
+    # there raised AttributeError two frames from anything that reads like a cause.
+    card = beat.get("card")
+    if card is not None and not isinstance(card, dict):
+        return beat, f"{path.name} carries a 'card' that is not an object"
+    stamped = (card or {}).get("date")
     local = now.astimezone(PT)
     today = local.date().isoformat()
     if stamped == today:
@@ -388,6 +395,35 @@ def read_heartbeat(now: dt.datetime, paths=None) -> tuple[dict, str | None]:
 
     return beat, (f"the state card is from {stamped}, not {today}. "
                   "Showing it as today's book would be wrong, so it is withheld")
+
+
+def clock_warning(now: dt.datetime, paths=None) -> tuple[list, str | None]:
+    """(rows, error) shaped for the brief's ENGINEERING route, never for his message.
+
+    THE SIGNAL THAT FOUND THE DRIFT MUST NOT DIE WITH THE ROW THAT CARRIED IT (PR #335
+    reviewer round 6, minor, and they were right). Before this PR the only thing that
+    reacted to a brief running early was a P0 alarm row on the founder's board. That row
+    was unactionable, pointed at the wrong cause, and appeared every morning, so it had
+    to go. Removing it without replacing it would have made a real misconfiguration
+    silent, which is a worse defect than the noisy one.
+
+    So the signal changes AUDIENCE rather than disappearing. On the committed schedule
+    the brief runs at 07:40 and this never fires. If it fires, either the loaded job has
+    drifted early, which is exactly what happened for four days, or someone ran the
+    brief by hand. `founder-notifications.md`, founder-directed 2026-08-10: engineering
+    signal goes to Sana's Linear triage and never to him.
+
+    It re-reads the heartbeat rather than being handed one, so the route stays a pure
+    function of the same file `collect` reads. That is one extra small file read per
+    run, and the alternative is threading a control flag through a second caller.
+    """
+    beat, err = read_heartbeat(now, paths)
+    if err is None and beat.get("card_is_yesterdays"):
+        return [], ("the brief read a state card stamped yesterday, which the committed "
+                    f"{CARD_WRITTEN_AT.strftime('%H:%M')} card and "
+                    "07:40 brief make impossible: com.kipi.morning-brief has drifted "
+                    "early in ~/Library/LaunchAgents, or this was a hand run")
+    return [], None
 
 
 def read_gtm(paths=None) -> tuple[dict | None, str | None]:
--- a/q-system/.q-system/scripts/consulting_board.py
+++ b/q-system/.q-system/scripts/consulting_board.py
@@ -63,6 +63,16 @@ PT = ZoneInfo("America/Los_Angeles")
 #: that fits the founder's screen without scrolling, measured against the 2026-09-03 card
 #: (7 red). A withheld count is always printed rather than the rest being dropped silently.
 MAX_CLIENT_ROWS = 6
+#: WHEN THE STATE CARD IS WRITTEN, from `io.askconsulting.ask-crm-state-card.plist`
+#: (StartCalendarInterval 07:30 PT). The morning brief runs at 07:00, from
+#: `com.kipi.morning-brief.plist`. So the brief reads this card THIRTY MINUTES BEFORE
+#: today's is written, every single day, and a rule that demanded today's date could
+#: never be satisfied at 07:00. Measured in ~/.config/kipi/logs/morning-brief.out.log:
+#: the refusal fired on the 09-05 and 09-07 runs, and today-card.md was last written
+#: 09-07 07:30. The cost was a P0 "Your book: COULD NOT READ" row painted onto his
+#: board every morning and cleared by nothing, since the hourly repaint has no
+#: scheduler. A row he cannot act on is the thing he asked to have removed.
+CARD_WRITTEN_AT = dt.time(7, 30)
 
 
 def consulting_root() -> Path:
@@ -320,11 +330,27 @@ def read_heartbeat(now: dt.datetime, paths=None) -> tuple[dict, str | None]:
         return beat, f"the 07:30 state card crashed: {beat['crash']}"
 
     stamped = (beat.get("card") or {}).get("date")
-    today = now.astimezone(PT).date().isoformat()
-    if stamped != today:
-        return beat, (f"the state card is from {stamped}, not {today}. "
-                      "Showing it as today's book would be wrong, so it is withheld")
-    return beat, None
+    local = now.astimezone(PT)
+    today = local.date().isoformat()
+    if stamped == today:
+        return beat, None
+
+    # YESTERDAY'S CARD IS THE NEWEST ONE THAT EXISTS BEFORE 07:30, and calling the
+    # newest card a failure is not a refusal, it is a clock error. The old rule was
+    # date equality, so at 07:00 it refused a card that was working perfectly and had
+    # simply not been rewritten yet. What it protected against is real and is kept:
+    # yesterday's book must never be presented AS today's, so the caller labels it.
+    #
+    # AFTER 07:30 THE SAME CARD IS A GENUINE FAILURE, and still refused. That is the
+    # whole reason this keys on the schedule rather than on a fixed number of hours:
+    # a 26-hour window would have swallowed a card the 07:30 job failed to write.
+    yesterday = (local.date() - dt.timedelta(days=1)).isoformat()
+    if stamped == yesterday and local.time() < CARD_WRITTEN_AT:
+        beat["card_is_yesterdays"] = True
+        return beat, None
+
+    return beat, (f"the state card is from {stamped}, not {today}. "
+                  "Showing it as today's book would be wrong, so it is withheld")
 
 
 def read_gtm(paths=None) -> tuple[dict | None, str | None]:
@@ -520,7 +546,13 @@ def collect(now: dt.datetime, sources: dict, paths=None):
     rows = []
     counts = (beat.get("card") or {}).get("counts") or {}
     if counts:
-        rows.append(f"book: {counts.get('red', 0)} owed, {counts.get('reach', 0)} to reach out")
+        # NAMED, never silently substituted. The reader accepts yesterday's card before
+        # 07:30 because it is the newest one there is; the promise it must keep is that
+        # nobody reads it as today's.
+        age = " (yesterday's card, today's is written at 07:30)" if beat.get(
+            "card_is_yesterdays") else ""
+        rows.append(f"book: {counts.get('red', 0)} owed, "
+                    f"{counts.get('reach', 0)} to reach out{age}")
 
     shown = card_rows[:MAX_CLIENT_ROWS]
     for row in shown:
--- a/q-system/.q-system/scripts/consulting_board.py
+++ b/q-system/.q-system/scripts/consulting_board.py
@@ -59,6 +59,7 @@ from __future__ import annotations
 
 import datetime as dt
 import json
+import plistlib
 import os
 import re
 from pathlib import Path
@@ -397,33 +398,60 @@ def read_heartbeat(now: dt.datetime, paths=None) -> tuple[dict, str | None]:
                   "Showing it as today's book would be wrong, so it is withheld")
 
 
-def clock_warning(now: dt.datetime, paths=None) -> tuple[list, str | None]:
+#: The job whose schedule this checks, and where launchd actually reads it from.
+BRIEF_LABEL = "com.kipi.morning-brief"
+LAUNCH_AGENTS = Path.home() / "Library" / "LaunchAgents"
+
+
+def clock_warning(now: dt.datetime, agents_dir=None) -> tuple[list, str | None]:
     """(rows, error) shaped for the brief's ENGINEERING route, never for his message.
 
     THE SIGNAL THAT FOUND THE DRIFT MUST NOT DIE WITH THE ROW THAT CARRIED IT (PR #335
-    reviewer round 6, minor, and they were right). Before this PR the only thing that
-    reacted to a brief running early was a P0 alarm row on the founder's board. That row
-    was unactionable, pointed at the wrong cause, and appeared every morning, so it had
-    to go. Removing it without replacing it would have made a real misconfiguration
-    silent, which is a worse defect than the noisy one.
+    reviewer round 6). Before this PR the only thing that reacted to a brief running
+    early was a P0 alarm row on the founder's board: unactionable, wrong about the
+    cause, and painted every morning, so it had to go. Removing it with nothing in its
+    place would have made a real misconfiguration silent, which is the worse defect.
+    The drift found on 2026-09-08 had run four days and that row is what exposed it.
+    So the signal changes AUDIENCE. `founder-notifications.md`, founder-directed
+    2026-08-10: engineering signal goes to Sana's Linear triage and never to him.
 
-    So the signal changes AUDIENCE rather than disappearing. On the committed schedule
-    the brief runs at 07:40 and this never fires. If it fires, either the loaded job has
-    drifted early, which is exactly what happened for four days, or someone ran the
-    brief by hand. `founder-notifications.md`, founder-directed 2026-08-10: engineering
-    signal goes to Sana's Linear triage and never to him.
+    IT READS THE LOADED JOB, not the heartbeat, and that is the third design (round 7).
+    Reading the heartbeat a second time raced the 07:30 card job: a run straddling
+    07:30 would see yesterday's card in `collect`, today's here, and report nothing,
+    losing exactly the run the signal exists for. Caching the first verdict on the
+    module looked like the fix and was DEAD ON ARRIVAL: morning-brief's `_load_sibling`
+    execs a fresh module per call, so the cache is always empty and the fallback read
+    always runs. My own test caught that, not review.
 
-    It re-reads the heartbeat rather than being handed one, so the route stays a pure
-    function of the same file `collect` reads. That is one extra small file read per
-    run, and the alternative is threading a control flag through a second caller.
+    So it asks the question directly. The condition is not "was the card stale on this
+    run", it is "is this job scheduled before the card it mirrors", which is a fact
+    about the machine, has no clock in it, and cannot race. It also does not fire on a
+    hand run before 07:30, which the time-based version would have, and an engineering
+    channel that speaks every time someone runs the brief by hand is one that gets
+    muted.
     """
-    beat, err = read_heartbeat(now, paths)
-    if err is None and beat.get("card_is_yesterdays"):
-        return [], ("the brief read a state card stamped yesterday, which the committed "
-                    f"{CARD_WRITTEN_AT.strftime('%H:%M')} card and "
-                    "07:40 brief make impossible: com.kipi.morning-brief has drifted "
-                    "early in ~/Library/LaunchAgents, or this was a hand run")
-    return [], None
+    path = Path(agents_dir or LAUNCH_AGENTS) / f"{BRIEF_LABEL}.plist"
+    if not path.is_file():
+        # No loaded job means nobody scheduled this run. Not news.
+        return [], None
+    try:
+        cal = plistlib.loads(path.read_bytes()).get("StartCalendarInterval")
+    except Exception:                                   # noqa: BLE001
+        # A plist this cannot parse is a machine problem, but it is not THIS signal,
+        # and guessing would file a wrong ticket. Silent by design.
+        return [], None
+    slots = cal if isinstance(cal, list) else [cal] if isinstance(cal, dict) else []
+    card = CARD_WRITTEN_AT.hour * 60 + CARD_WRITTEN_AT.minute
+    early = sorted(
+        "%02d:%02d" % (e.get("Hour", 0), e.get("Minute", 0)) for e in slots
+        if isinstance(e, dict)
+        and (e.get("Hour", 0) * 60 + e.get("Minute", 0)) < card)
+    if not early:
+        return [], None
+    return [], (f"{BRIEF_LABEL} is loaded at {', '.join(early)}, before the "
+                f"{CARD_WRITTEN_AT.strftime('%H:%M')} state card it mirrors. The "
+                "committed template runs after the card; the loaded copy has drifted. "
+                f"Reinstall it: install-plist.sh {BRIEF_LABEL}")
 
 
 def read_gtm(paths=None) -> tuple[dict | None, str | None]:
--- a/q-system/.q-system/scripts/consulting_board.py
+++ b/q-system/.q-system/scripts/consulting_board.py
@@ -26,9 +26,10 @@ reasons it is struck, and the second is the expensive one:
 
 ## STALENESS IS AN ERROR, never a quiet mirror
 
-The card is written at 07:30 and THIS RUNS AT 07:00, thirty minutes before it. That
-inversion is real, it is not a typo, and the old text here asserted the opposite. See
-CARD_WRITTEN_AT.
+The card is written at 07:30 and this repo COMMITS the brief at 07:40, after it. The
+old text here said "runs after it" and was right about the intent; what it could not say
+is that the LOADED job on the founder's machine had drifted to 07:00 and ran forty
+minutes early for four days. See CARD_WRITTEN_AT.
 
 So a card stamped yesterday is two different facts depending on the hour. Before 07:30
 it is simply the newest card that exists, and it is used and LABELLED as yesterday's, on
@@ -371,8 +372,8 @@ def read_heartbeat(now: dt.datetime, paths=None) -> tuple[dict, str | None]:
 
     # YESTERDAY'S CARD IS THE NEWEST ONE THAT EXISTS BEFORE 07:30, and calling the
     # newest card a failure is not a refusal, it is a clock error. The old rule was
-    # date equality, so at 07:00 it refused a card that was working perfectly and had
-    # simply not been rewritten yet. What it protected against is real and is kept:
+    # date equality, so on the drifted 07:00 run it refused a card that was working
+    # perfectly and had simply not been rewritten yet. What it protected against is real and is kept:
     # yesterday's book must never be presented AS today's, so the caller labels it.
     #
     # AFTER 07:30 THE SAME CARD IS A GENUINE FAILURE, and still refused. That is the
--- a/q-system/.q-system/scripts/consulting_board.py
+++ b/q-system/.q-system/scripts/consulting_board.py
@@ -171,7 +171,8 @@ def read_card(paths=None) -> tuple[list[dict], str | None]:
     paths = paths or _paths()
     path = paths["card"]
     if not path.exists():
-        return [], f"no state card at {path.name}; the 07:30 job has not written one"
+        return [], (f"no state card at {path.name}; the "
+                    f"{CARD_WRITTEN_AT.strftime('%H:%M')} job has not written one")
     try:
         text = path.read_text(encoding="utf-8")
     except OSError as exc:
@@ -362,7 +363,8 @@ def read_heartbeat(now: dt.datetime, paths=None) -> tuple[dict, str | None]:
     beat.pop("card_is_yesterdays", None)
 
     if beat.get("crash"):
-        return beat, f"the 07:30 state card crashed: {beat['crash']}"
+        return beat, (f"the {CARD_WRITTEN_AT.strftime('%H:%M')} state card crashed: "
+                      f"{beat['crash']}")
 
     stamped = (beat.get("card") or {}).get("date")
     local = now.astimezone(PT)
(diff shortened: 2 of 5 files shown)


assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-335-20260908-042021.md` (7208 bytes).

```
FINDINGS:
minor|Module docstring still says the brief runs after the 07:30 card and that a non-today date returns an ERROR; both are now false|q-system/.q-system/scripts/consulting_board.py:29
minor|board_rows doc says buckets errors when the state card is yesterday's, which it no longer does before 07:30|q-system/.q-system/scripts/board_rows.py:62
minor|The yesterday-card label the code calls its own safety condition is absent from buckets(), so the Notion board paints a day-old book with no marker and no alarm|q-system/.q-system/scripts/consulting_board.py:551
nit|The yesterday label is appended to the optional counts line, so a heartbeat without counts renders yesterday's client rows unlabeled|q-system/.q-system/scripts/consulting_board.py:548
END FINDINGS
```

--- reviewer output ---

## Review: PR #335

I read the diff, the two schedules on this machine, the live heartbeat the producer actually writes, and I ran the tests plus four mutants against the new window.

The premise checks out. Verified independently of the author's claims:

- `com.kipi.morning-brief` StartCalendarInterval is Hour 7, Minute 0. `io.askconsulting.ask-crm-state-card` is Hour 7, Minute 30. The brief really does read the card thirty minutes before it is rewritten.
- `~/.config/kipi/logs/morning-brief.out.log` carries the 09-07 refusal verbatim: "the state card is from 2026-09-06, not 2026-09-07".
- The live heartbeat at `~/projects/consulting/q-consult/output/ask-crm-state-card-heartbeat.json` has `at` 2026-09-07T14:30:05Z, which is 07:30:05 PT, and `card.date` 2026-09-07. The test fixture's shape matches the real producer: same `card.counts` and `card.date` nesting. This is not an invented fixture.

---

### FINDING 1 (minor): the module docstring of the file being changed still asserts the belief this PR says is wrong

`q-system/.q-system/scripts/consulting_board.py:29`

The commit message says "THE REPO BELIEVED THE OPPOSITE, and the belief was load-bearing", and corrects that belief in `test_morning_brief.py`. The same false sentence is still at the top of the file the PR edits:

> The card is written at 07:30 and this runs after it. If the heartbeat's date is not today, this returns an ERROR naming the age

Both halves are now false. The brief runs at 07:00, before it. And a yesterday date before 07:30 returns no error.

Reproducer and real output:

```
$ python3 - <<'PY'   # yesterday's card, 07:00 PT
rows, err = cb.collect(BRIEF, {}, paths)
PY
SLACK collect() rows:
    book: 2 owed, 1 to reach out (yesterday's card, today's is written at 07:30)
    🔴 Northwind Design · you said "add a renewal date" — not sent
```

`err` is None. The docstring says it is an ERROR.

---

### FINDING 2 (minor): the consumer's doc still says a yesterday card is an error

`q-system/.q-system/scripts/board_rows.py:62`

> If `consulting_board.buckets` reports an error (the state card is yesterday's, or the 07:30 job crashed), this writes no rows at all

Before 07:30 it no longer reports that error and it does write rows. Same reproducer as above: `buckets()` at 07:00 with a 2026-09-02 card returned two live client rows and no card error.

---

### FINDING 3 (minor): the "yesterday's card" label exists only on the Slack line, not on the board rows

`q-system/.q-system/scripts/consulting_board.py:551`

The comment there states the safety condition explicitly: "Accepting it is only safe because nothing reads it as today's." `buckets()`, the other renderer in the same file, reads it as today's. It paints the client rows with no age marker and, since the alarm row is gone, no staleness signal at all.

Reproducer and real output, same tree, same 07:00 timestamp:

```
BOARD buckets() top_of_mind -- what notion_board.py writes:
    ['🔴 Northwind Design', 'you said "add a renewal date" — not sent', 'State card', 'P0']
    ['🔴 Harbor Labs', 'you said "build the records" — not sent', 'State card', 'P0']
'yesterday' anywhere in board payload: False
```

Held to minor because the Slack message that carries the label is delivered in the same brief that summarizes the board, so the founder does get told once. The board itself no longer says it.

---

### FINDING 4 (nit): the label is welded to the optional counts line

`q-system/.q-system/scripts/consulting_board.py:548`

The `(yesterday's card...)` suffix is appended to the `book:` row, which is only emitted `if counts:`. A heartbeat with no `counts` prints yesterday's client rows with no label anywhere.

```
err: None
rows: ['🔴 Northwind Design · you said "add a renewal date" — not sent', 'GTM: COULD NOT READ ...']
label present: False
```

Dropped to nit under the standing rule: I constructed that heartbeat. The live producer always emits `counts`, so no producer in either repo can reach this today.

---

### What is sound

**The tests have real teeth.** I copied the tree, applied four mutants, and ran only the new class. Every one went red:

```
MUTANT: accept ANY older card before 07:30            -> 1 failed, 6 passed
MUTANT: drop the clock guard (accept yesterday all day) -> 2 failed, 5 passed
MUTANT: move the constant to 08:00                    -> 3 failed, 4 passed
MUTANT: drop the yesterday label                      -> 1 failed, 6 passed
```

The constant is pinned by code tests, not only by the plist test that skips when the job is not installed. The 08:00 mutant is the one that matters and it took three tests down.

**The fixture is the producer's shape.** I read the live heartbeat and compared. `card.date` and `card.counts` nest exactly as the fixture does. This is not the invented-key failure this fleet has shipped before.

**The window is one day wide and the two-day case is genuinely refused.** `test_a_card_older_than_yesterday_is_refused_at_any_hour` is not decoration; the "accept ANY older card" mutant is precisely what it catches.

**Detection of a dead 07:30 job is delayed by 24 hours, not lost.** I traced it: job dies day N, brief on day N accepts N-1 silently, brief on day N+1 sees N-1 which is two days old and alarms. Deliberate, explained, and the alternative was a P0 that fired every single morning forever. Not a finding.

**Archiving blast radius.** I looked hard at `healthy.add("card")` now firing at 07:00 where it previously did not, since that authorizes the painter to archive card-scope Notion pages. I could not construct a case where it archives a row the founder owns: hand-created rows carry no owned prefix and `_scope_of` gates it. No reproducer, so I am dropping it rather than asserting it.

**Regression sweep.** 142 board tests plus 182 across `test_morning_brief.py`, `test_board_is_actionable.py`, `test_off_switches.py`, `test_morning_brief_mail.py`. All green.

---

Only minor and nit findings survived reproduction. Nothing here is unrecoverable, nothing cries wolf, nothing reports success for work that did not happen.

**Fix first:** `consulting_board.py:29`. The PR's whole argument is that a wrong belief written into a docstring was load-bearing and cost a P0 row every morning. Leaving that same wrong belief at the top of the file it just fixed is the next reviewer's trap.

FINDINGS:
minor|Module docstring still says the brief runs after the 07:30 card and that a non-today date returns an ERROR; both are now false|q-system/.q-system/scripts/consulting_board.py:29
minor|board_rows doc says buckets errors when the state card is yesterday's, which it no longer does before 07:30|q-system/.q-system/scripts/board_rows.py:62
minor|The yesterday-card label the code calls its own safety condition is absent from buckets(), so the Notion board paints a day-old book with no marker and no alarm|q-system/.q-system/scripts/consulting_board.py:551
nit|The yesterday label is appended to the optional counts line, so a heartbeat without counts renders yesterday's client rows unlabeled|q-system/.q-system/scripts/consulting_board.py:548
END FINDINGS

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-335-20260908-043829.md` (6490 bytes).

```
FINDINGS:
major|The new before-07:30 window never opens at the brief's committed 07:40 slot, so the P0 alarm row is still painted every morning for anyone who installs the plist from this repo|q-system/.q-system/scripts/consulting_board.py:74
minor|The new test class anchors on an invented 07:00 timestamp described as "the brief's own slot", which no plist in this repo emits, so the suite passes only at a schedule the repo does not ship|q-system/.q-system/tests/test_consulting_board.py:295
minor|The CARD_WRITTEN_AT pin skips on any host without the consulting LaunchAgent, so it never runs on ubuntu CI and the constant is unpinned everywhere but one Mac|q-system/.q-system/tests/test_consulting_board.py:373
minor|The founder-facing Slack line hardcodes "07:30" instead of formatting CARD_WRITTEN_AT, so moving the constant makes the line state a time the code does not use|q-system/.q-system/scripts/consulting_board.py:560
END FINDINGS
```

--- reviewer output ---

## Review: PR #335

**One-line finding:** the fix is keyed to a brief schedule of 07:00, and this repo's own committed plist says 07:40.

---

### major — the new window never opens at the schedule this repo ships

`q-system/.q-system/scripts/consulting_board.py:74` and `:29` assert "the morning brief runs at 07:00". The committed template says otherwise:

`q-system/.q-system/scripts/com.kipi.morning-brief.plist` line 57: `Hour 7, Minute 40`, with a comment added 2026-09-03 that explicitly rejects the 07:00 ordering: *"07:40, not 07:00 (2026-09-03). The consulting state card is written at 07:30 ... At 07:00 it mirrored yesterday."* `CLAUDE.md:21` says 07:40, and `test_morning_brief.py` already pins those two together.

At 07:40 the guard `local.time() < CARD_WRITTEN_AT` is false, so the P0 row is still painted.

Reproducer, run against this worktree:

```
python3 -  # builds a tmp q-consult tree, heartbeat stamped 2026-09-02, calls collect+buckets
```

Real output:

```
BRIEF plist slot (repo template):   07:40
PR + code comments + new tests say: 07:00
CARD_WRITTEN_AT:                    07:30:00

== 07:00 PT (the new tests' invented slot) ==
  collect err : None
  P0 alarm row: (none)
  client rows : 2

== 07:40 PT (the plist's REAL slot) ==
  collect err : the state card is from 2026-09-02, not 2026-09-03. Showing it as today's book would be wrong, so it is withheld
  P0 alarm row: Your book: COULD NOT READ
  client rows : 0
```

Why the founder's machine looks like 07:00: the installed copy is stale.

```
python3 -c "import plistlib,pathlib; d=pathlib.Path.home()/'Library/LaunchAgents'; ..."
com.kipi.morning-brief.plist {'Hour': 7, 'Minute': 0}
io.askconsulting.ask-crm-state-card.plist {'Hour': 7, 'Minute': 30}
com.kipi.morning-inbox.plist NOT INSTALLED
```

So the running job predates the 2026-09-03 move and was never reinstalled. Running the documented `install-plist.sh com.kipi.morning-brief` puts the brief back at 07:40 and the alarm row returns, with the code now carrying comments that say the opposite.

Related: the PR body captures the ordering as sp-642d380b, "across two repos, which is his call." The brief's plist lives in **this** repo and already carries that decision, made the other way, ten days ago.

3am consequence: nobody is paged. The operator either keeps a drifted plist forever, or reinstalls it and gets the P0 row back.

---

### minor — the new tests anchor on an invented timestamp, which is why the above shipped green

`q-system/.q-system/tests/test_consulting_board.py:294-295` calls `BRIEF_SLOT = 14:00Z` "the brief's own slot". No plist in this repo emits that. The sibling `test_no_hourly_slot_fires_before_the_card_it_mirrors` two files over reads the plist for exactly this reason, and its docstring at `test_morning_brief.py:983` now says "the brief runs at 07:00" eight lines above code that loads 07:40 from the file.

Same reproducer as above: all four behaviour assertions hold at 14:00Z, and `collect` returns the refusal plus the P0 row at 14:40Z.

```
python3 -m pytest ...::TestTheBriefReadsTheCardBeforeTheCardIsWritten -q
10 passed in 0.03s
```

---

### minor — the `CARD_WRITTEN_AT` pin can only run on one Mac

`test_consulting_board.py:373` skips when `~/Library/LaunchAgents/io.askconsulting.ask-crm-state-card.plist` is absent. CI is `ubuntu-latest` (`.github/workflows/verify.yml:14`), so it always skips there. Proof, same test under a fresh HOME:

```
SKIPPED [1] q-system/.q-system/tests/test_consulting_board.py:373: the consulting state-card job is not installed on this machine
1 skipped in 0.02s
```

---

### minor — the founder-facing line carries a fourth hardcoded copy of the schedule

`consulting_board.py:560` writes the literal `"today's is written at 07:30"` instead of formatting `CARD_WRITTEN_AT`. Moving the constant makes the Slack line lie:

```
CARD_WRITTEN_AT moved to 08:00; run at 07:45 PT
  err: None
  slack line: book: yesterday's card, today's is written at 07:30
```

`test_the_deadman_alarm_names_no_clock_time` in this same suite exists because a clock time reaching the founder drifted once already.

---

### What is sound

- **Two-day-old card before 07:30.** Refused: `the state card is from 2026-09-01, not 2026-09-03`. The window is genuinely one day wide.
- **Row identity across the label.** Keys are identical stale vs fresh (`['client:Northwind']`), so the new `(from yesterday's card)` detail does not mint new board ids and cannot silently archive a row the founder dragged.
- **`card_is_yesterdays` leaking to disk.** `read_heartbeat` mutates only the in-memory dict; re-reading the heartbeat file shows the key absent.
- **DST fall-back, 2026-11-01.** 07:00 PST accepts, 07:45 PST refuses. `astimezone(PT)` handles the offset change with no table.
- **Crash branch ordering.** A crashed heartbeat still errors before the window is consulted, so a two-day gap after a crash cannot slip through as "yesterday's".
- Full suites green: `test_consulting_board.py` 145 passed, `test_morning_brief.py` 63 passed.

---

**VERDICT: REQUEST CHANGES.** Nothing here is permanent or unrecoverable, so this is not a BLOCK.

**Fix first:** decide which schedule is true, and derive it rather than restate it. If 07:40 stands, this change is a no-op and the real fix is reinstalling the plist; if 07:00 stands, the plist, `CLAUDE.md` and the 2026-09-03 comment all move with it, and `BRIEF_SLOT` becomes a value read from the plist so the test goes red when they disagree.

FINDINGS:
major|The new before-07:30 window never opens at the brief's committed 07:40 slot, so the P0 alarm row is still painted every morning for anyone who installs the plist from this repo|q-system/.q-system/scripts/consulting_board.py:74
minor|The new test class anchors on an invented 07:00 timestamp described as "the brief's own slot", which no plist in this repo emits, so the suite passes only at a schedule the repo does not ship|q-system/.q-system/tests/test_consulting_board.py:295
minor|The CARD_WRITTEN_AT pin skips on any host without the consulting LaunchAgent, so it never runs on ubuntu CI and the constant is unpinned everywhere but one Mac|q-system/.q-system/tests/test_consulting_board.py:373
minor|The founder-facing Slack line hardcodes "07:30" instead of formatting CARD_WRITTEN_AT, so moving the constant makes the line state a time the code does not use|q-system/.q-system/scripts/consulting_board.py:560
END FINDINGS

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-335-20260908-045617.md` (4755 bytes).

```
FINDINGS:
minor|The CARD_WRITTEN_AT comment and commit body claim the alarm row is cleared by nothing because the hourly repaint has no scheduler, but com.kipi.morning-inbox schedules 12 repaints a day and the 08:05 one archives it|q-system/.q-system/scripts/consulting_board.py:82
nit|card_is_yesterdays is an internal control flag read back out of the externally-parsed heartbeat JSON, so a heartbeat carrying that key labels a fresh card as yesterday's on both the Slack line and the Notion board|q-system/.q-system/scripts/consulting_board.py:366
END FINDINGS
```

--- reviewer output ---

Review done. Read-only, nothing in the repo modified.

## Findings

**MINOR. The comment and commit body say the alarm row was "cleared by nothing, because the hourly repaint has no scheduler." The repo commits that scheduler, 12 slots a day, and it archives the row.**

`q-system/.q-system/scripts/consulting_board.py:82` (the `CARD_WRITTEN_AT` comment block), same claim in the commit body of `58c269b8`.

Reproducer, run against the branch:

```
python3 - <<'PY'   # loads the repo's own plist, then drives buckets + paint
...
PY
```

Real output:

```
1. morning-inbox slots: 08:05, 09:05, 10:05, 11:05, 12:05, 13:05, 14:05, 15:05, 16:05, 17:05, 18:05, 19:05
2. HOURLY_SECTIONS = ("consulting_board.py", "groupme_inbox.py", "board_rows.py")
   board_rows.collect calls buckets+paint: True
3. fresh card @08:05 PT -> alarm row present: False | CARD_ALARM in healthy_scopes: True
   paint archived: 1 | archive calls: [('PATCH', '/pages/page-alarm')]
```

`com.kipi.morning-inbox` runs `morning-brief.py --inbox-only`, `board_rows.py` is in `HOURLY_SECTIONS`, and `board_rows.collect` calls the full painter. `CARD_ALARM` is always in `healthy_scopes`, so the 08:05 run archives the row. The alarm lived roughly 65 minutes, 07:00 to 08:05, not "every morning, cleared by nothing." The fix is still right; the sizing written into the code comment is not, and it is the comment the next person reads before touching this window.

**NIT. `card_is_yesterdays` is an internal control flag stored inside the externally-parsed heartbeat JSON, so a heartbeat carrying that key mislabels a fresh card.**

`q-system/.q-system/scripts/consulting_board.py:366`.

Reproducer: heartbeat with `card.date` set to today, `card_is_yesterdays: true` added, read at 09:00 PT. Real output:

```
card date == today, hour 09:00, err: None
slack rows: ["book: yesterday's card, today's is written at 07:30", 'book: 1 owed, 0 to reach out', '🔴 Northwind Design · you said x — not sent']
board details: ["you said x — not sent (from yesterday's card)"]
```

Dropped a level on purpose, per calibration: no producer in this repo emits that key, so this only fails on an input I constructed. `read_heartbeat` returning a third value, or a `_` prefix, closes it.

## What is sound

- Mutation testing, four mutants, all caught red. Dropping the clock guard: 5 failures. Dropping the one-day window so any old card passes before 07:30: 1 failure (`test_a_card_older_than_yesterday_is_refused_at_any_hour`). Dropping the board label: 1 failure. Moving `CARD_WRITTEN_AT` to 08:30: 7 failures. No new test in this PR is decoration.
- The fixture-invention attack fails here. Mutant 4 turned `test_and_the_constant_still_matches_the_installed_card_job` red, which proves the real installed `io.askconsulting.ask-crm-state-card.plist` on this host actually says 07:30. The constant is pinned against the producer, not against a shape the author imagined.
- DST. `PT` is `ZoneInfo("America/Los_Angeles")`, not a fixed offset, and launchd calendar intervals are local time, so the 07:30 boundary tracks PST and PDT without a table. No half-year drift.
- Archive destruction, the one that would have been a blocker. Before 07:30 the card scope now becomes healthy off a day-old snapshot, which authorizes archiving. I could not build a sequence where that removes a live row: the board's card rows came from that same card, and capped and pinned rows are already excluded by the round-11 and round-9 branches.
- Rollback. Pure code plus tests, no migration, no permanent object written. Revertible.
- Both suites green on the branch: 146 passed in `test_consulting_board.py`, 63 in `test_morning_brief.py`.
- Dropped for lack of a reproducer: the change trades a loud unactionable P0 for a quiet label, so the next LaunchAgent drift is less visible. It is already captured as sp-de7afcff, and I could not build a check that fails, so it is not a finding.

## Verdict

**APPROVE WITH NITS.**

Fix first: the `CARD_WRITTEN_AT` comment claiming no hourly repaint exists. `com.kipi.morning-inbox.plist` schedules one at 08:05 through 19:05 and it archives the alarm row.

FINDINGS:
minor|The CARD_WRITTEN_AT comment and commit body claim the alarm row is cleared by nothing because the hourly repaint has no scheduler, but com.kipi.morning-inbox schedules 12 repaints a day and the 08:05 one archives it|q-system/.q-system/scripts/consulting_board.py:82
nit|card_is_yesterdays is an internal control flag read back out of the externally-parsed heartbeat JSON, so a heartbeat carrying that key labels a fresh card as yesterday's on both the Slack line and the Notion board|q-system/.q-system/scripts/consulting_board.py:366
END FINDINGS

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-335-20260908-050948.md` (5033 bytes).

```
FINDINGS:
minor|Three files added by this PR claim the brief runs at 07:00 while the committed plist runs 07:40, and consulting_board.py states both|q-system/.q-system/scripts/consulting_board.py:29
minor|The only test pinning CARD_WRITTEN_AT to the actual card job skips on any host without the LaunchAgent, so CI never checks it|q-system/.q-system/tests/test_consulting_board.py:396
END FINDINGS
```

--- reviewer output ---

## Findings

**MINOR — three files added by this PR say the brief runs at 07:00; the plist this repo commits says 07:40, and `consulting_board.py` contradicts itself 45 lines apart**

`q-system/.q-system/scripts/consulting_board.py:29`

Line 29 says "The card is written at 07:30 and THIS RUNS AT 07:00, thirty minutes before it. That inversion is real, it is not a typo." Line 74 of the same file says the committed plist is 07:40, "DELIBERATELY after it," and "this window never opens" on the shipped schedule. Both were added in this PR. The last commit corrected the constant's comment and left the module docstring, `board_rows.py:67`, and `test_morning_brief.py:983` on the old story. The PR body carries it too, so a squash merge writes it into history.

This is the same defect class an existing test was written for. `test_morning_brief.py:951` reads "Codex round 6 (minor): CLAUDE.md said 07:00 while the plist ran 07:40." That guard only scans `CLAUDE.md`, so it cannot see any of these three.

Reproducer and real output:

```
$ python3 -c "...read plist, grep the claims..."
COMMITTED brief schedule (com.kipi.morning-brief.plist): 07:40
CLAIM  board_rows.py:67          because the brief runs at 07:00
CLAIM  consulting_board.py:29    The card is written at 07:30 and THIS RUNS AT 07:00
CLAIM  test_morning_brief.py:983 The brief runs at 07:00 and the card at 07:30
```

```
$ sed -n 74,77p q-system/.q-system/scripts/consulting_board.py
#: (StartCalendarInterval 07:30 PT). This repo's `com.kipi.morning-brief.plist` commits
#: 07:40, DELIBERATELY after it: f9f74ac1 set that minute on 2026-09-04 in the change
#: that made this board mirror the card. So on the shipped schedule a card stamped
#: yesterday never happens, and this window never opens.
```

I also confirmed the window is unreachable on the shipped config rather than taking the comment's word for it. Hourly slots are 08:05 through 19:05, all after 07:30.

**MINOR — the only test that pins `CARD_WRITTEN_AT` to the real card job skips off the founder's Mac**

`q-system/.q-system/tests/test_consulting_board.py:396`

`test_the_card_is_written_before_this_repo_runs_the_brief` asserts brief slot at or after `CARD_WRITTEN_AT`. That is satisfied by 07:40 >= 07:30 and stays satisfied if the consulting job moves to 08:00 and nobody updates the constant. The boundary test pins the constant to literals chosen in the same file. The only check that compares the constant against the job it names reads `~/Library/LaunchAgents` and skips when absent. Failure mode: the consulting card job moves later, the 07:40 brief refuses again, the P0 "Your book: COULD NOT READ" row comes back, and CI is green.

```
$ HOME=<tempdir> python3 -m pytest ...test_consulting_board.py -q -rs -k constant_still_matches
SKIPPED [1] test_consulting_board.py:396: the consulting state-card job is not installed on this machine
1 skipped, 145 deselected in 0.03s
```

## What is sound

- Mutation on the board label. Deleting the two-line `card_is_yesterdays` detail append in `buckets` turns `test_the_BOARD_says_it_too_not_just_the_slack_line` red with the real row dicts printed. That test is not decoration.
- I expected the "(from yesterday's card)" marker never to reach Notion, since `_properties` writes `Status` create-only. Wrong: the label rides in `Notes`, and `_already_holds` (`board_rows.py:339`) compares it against the page already read, so a changed detail does get PATCHed. The safety claim holds on the surface he actually opens.
- I expected `healthy.add("card")` on the new early-accept path to widen the archive blast radius onto a row the founder had dragged. It cannot: the archive loop holds pinned rows at `board_rows.py:1051` before the archive PATCH.
- Slack section overflow from the extra label row. `_section` caps at `MAX_ROWS = 15`; the consulting section tops out around 11 rows. No displacement.
- The empty-detail branch emits "From yesterday's card." rather than the parenthetical, and still satisfies the substring assertion. Not a hole.
- Full run of `test_consulting_board.py`, `test_morning_brief.py`, `test_board_is_actionable.py`: 294 passed.

## Verdict

**APPROVE WITH NITS.** Nothing here is unrecoverable and nothing fires wrong at 3am; on the shipped schedule the new branch is unreachable, and where it does fire (a hand-run, or a LaunchAgent that has drifted early) it replaces an unactionable P0 with a labelled day-old book.

Fix first: the 07:00 claim in `consulting_board.py:29`, because that file states both versions and the next reader has no way to tell which one is current.

FINDINGS:
minor|Three files added by this PR claim the brief runs at 07:00 while the committed plist runs 07:40, and consulting_board.py states both|q-system/.q-system/scripts/consulting_board.py:29
minor|The only test pinning CARD_WRITTEN_AT to the actual card job skips on any host without the LaunchAgent, so CI never checks it|q-system/.q-system/tests/test_consulting_board.py:396
END FINDINGS

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-335-20260908-053019.md` (5196 bytes).

```
FINDINGS:
minor|The only check binding CARD_WRITTEN_AT to the real card job reads ~/Library/LaunchAgents and skips everywhere else, so CI never verifies the cross-repo constant|q-system/.q-system/tests/test_consulting_board.py:409
nit|Two founder-visible strings still hardcode 07:30 beside the new CARD_WRITTEN_AT constant and will lie if it moves|q-system/.q-system/scripts/consulting_board.py:759
nit|Assertion message calls 07:00 "the brief's own slot", which the class's own EARLY comment says is wrong|q-system/.q-system/tests/test_consulting_board.py:304
END FINDINGS
```

--- reviewer output ---

## What I ran

Read-only. I built a mutation harness in memory (repo source never written), planted 7 mutants in `consulting_board.py`, and ran the 11 new behaviours against each.

```
flag lines [586, 792] gate 382 pop 361
BASE                           RED: none
M1 drop board label            RED: ['T7']
M2 accept yesterday any hour   RED: ['T4', 'T5']
M3 accept ANY old card early   RED: ['T6']
M4 drop pop guard              RED: ['T9']
M5 drop slack label row        RED: ['T2', 'T8']
M6 constant to 08:00           RED: ['T4', 'T5', 'T11']
M7 revert window               RED: ['T1', 'T2', 'T3', 'T5', 'T7', 'T8']
```

Every mutant died. No new test is decoration.

## Findings

**minor. Nothing that runs off the founder's Mac ties `CARD_WRITTEN_AT` to the real card job, and the one check that does skips silently.**
`q-system/.q-system/tests/test_consulting_board.py:409`

```
$ python3 -m pytest ".../test_consulting_board.py::TestTheBriefReadsTheCardBeforeTheCardIsWritten" -q -rs
............                                                             [100%]
12 passed in 0.03s
```

Twelve passed here because `~/Library/LaunchAgents/io.askconsulting.ask-crm-state-card.plist` exists on this machine. On CI it hits `pytest.skip` and the constant is unverified. `test_the_card_is_written_before_this_repo_runs_the_brief` does not close it: it asserts only brief-slot 07:40 is greater than or equal to the constant, which any value at or below 07:40 satisfies, including a drifted one. If the consulting repo moves the card job to 08:00, the committed 07:40 brief sees yesterday's card at a time past 07:30, refuses it, and repaints the exact P0 row this PR removes, every morning, with no gate red. `consulting_root()` already locates the consulting instance in this same module, so reading that repo's committed plist instead of the installed LaunchAgent would make the pin run wherever the instance exists.

**nit. Two founder-visible strings hardcode 07:30 next to the constant that was just introduced to be the single copy.**
`q-system/.q-system/scripts/consulting_board.py:365` and `:759`

```
$ python3 -c '... src.replace("dt.time(7, 30)","dt.time(8, 0)") ...'
CARD_WRITTEN_AT now: 08:00:00
ALARM done text : the 07:30 state card writes a fresh one
ALARM detail    : the 07:30 state card crashed: boom
```

The constant moved, the P0 row still tells him to wait for a 07:30 job. Latent, not currently wrong.

**nit. The assertion message contradicts the constant's own comment three lines above it.**
`q-system/.q-system/tests/test_consulting_board.py:304`

The message reads `"the book was withheld at the brief's own slot"`, while the `EARLY` comment at line 291 says 07:00 is deliberately not called the brief's slot any more because the committed slot is 07:40. Whoever reads this go red gets sent to the wrong plist. Surfaces only on the second assert, when `err` is None and `rows` is empty.

## What is sound

- **Timezone and DST.** `PT` is `ZoneInfo("America/Los_Angeles")` at line 67 and `morning-brief.py:913` passes an aware local `now`, so `local.time() < CARD_WRITTEN_AT` is real PT. A fixed-offset PT would have opened the window at 07:40 in summer. It is not fixed-offset.
- **Archiving regression.** The early path now declares scope `card` healthy, which authorizes `board_rows` to archive. I could not build a case where it archives real work: between 00:00 and 07:30 the board's card rows came from that same yesterday card, so the wanted set is identical.
- **Row identity.** The `(from yesterday's card)` suffix goes into `detail` only; `key` stays `kind:name`, so no new Notion pages are minted and no dragged row is replaced.
- **Injected control flag.** `beat.pop("card_is_yesterdays")` holds. Dropping it kills T9.
- **Malformed heartbeat.** `card: []` is now handled cleanly, output `the state card is from None, not 2026-09-03`. `card: [1,2]` still raises `AttributeError`, but no producer emits that and `morning-brief.py` catches per collector into `consulting failed (AttributeError)`. Dropped as unreproducible against a real producer.
- **Commit claims against the repo.** `com.kipi.morning-brief.plist` is 07:40 and `com.kipi.morning-inbox.plist` carries 12 slots 08:05 to 19:05, exactly as the commit message states.
- Full suite: 147 passed in `test_consulting_board.py`, 63 passed in `test_morning_brief.py`.

## VERDICT: APPROVE WITH NITS

Fix first: give `CARD_WRITTEN_AT` a check that does not skip, by reading the consulting instance's committed plist through `consulting_root()` rather than the installed LaunchAgent.

FINDINGS:
minor|The only check binding CARD_WRITTEN_AT to the real card job reads ~/Library/LaunchAgents and skips everywhere else, so CI never verifies the cross-repo constant|q-system/.q-system/tests/test_consulting_board.py:409
nit|Two founder-visible strings still hardcode 07:30 beside the new CARD_WRITTEN_AT constant and will lie if it moves|q-system/.q-system/scripts/consulting_board.py:759
nit|Assertion message calls 07:00 "the brief's own slot", which the class's own EARLY comment says is wrong|q-system/.q-system/tests/test_consulting_board.py:304
END FINDINGS

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-335-20260908-054812.md` (4919 bytes).

```
FINDINGS:
minor|Accepting yesterday's card silences the only surface that reacted to the drifted brief schedule, and no drift detector replaces it|q-system/.q-system/scripts/consulting_board.py:395
nit|New isinstance(beat, dict) guard is untested and the next line still raises AttributeError on a malformed card value|q-system/.q-system/scripts/consulting_board.py:363
nit|test_a_card_older_than_yesterday_is_refused_at_any_hour exercises only 07:00|q-system/.q-system/tests/test_consulting_board.py:333
END FINDINGS
```

--- reviewer output ---

Review done. Read-only, nothing in the repo modified.

## What this PR does

`read_heartbeat` stops treating date-inequality as the freshness rule. Before `CARD_WRITTEN_AT` (07:30 PT) a card stamped *yesterday* is accepted as the newest one that exists, flagged `card_is_yesterdays`, and labelled on both surfaces. At or after 07:30 the same card is still an error.

## Findings

**minor — the drift signal goes quiet and the replacement detector does not exist yet.**
`q-system/.q-system/scripts/consulting_board.py:395`

The false P0 was the only thing on any surface that reacted to the loaded brief job having drifted to 07:00. After this change that run is silent, and `card:alarm` stays unconditionally healthy, so the existing P0 row is archived too.

Reproducer and real output:

```
$ python3 -c '... cb.buckets(EARLY_0700, {}, tree("2026-09-02")) ...'
=== B: BOARD top_of_mind ===
    client:Northwind Design | P0 | "you said add a renewal date — not sent (from yesterday's card)"
    client:Harbor Labs | P0 | "you said build the records — not sent (from yesterday's card)"
    reach:Kestrel Group | P1 | "— 🔥 Kestrel Group (fire) (from yesterday's card)"
   healthy: ['card', 'card:alarm', 'gtm', 'myside', 'week', 'week:gtm']
```

And nothing at runtime compares a loaded plist's schedule to the committed one:

```
$ grep -n "StartCalendarInterval" launchd-intent-verify.py launchd-health-check.py
(no matches; intent-verify diffs enabled-vs-disabled only)
```

Already captured as sp-de7afcff, and the removed alarm was factually false ("COULD NOT READ" about a readable card), so this is sequencing, not a regression. Naming it so the silencer landing before the detector is a deliberate call.

**nit — the new type guard hardens one level and the next line still raises.**
`q-system/.q-system/scripts/consulting_board.py:363`

```
1) heartbeat is a JSON list (new isinstance guard):
    ({}, 'ask-crm-state-card-heartbeat.json is not an object')
2) heartbeat card is a STRING, one level down from the new guard:
   RAISED AttributeError 'str' object has no attribute 'get'
```

Line 2 is pre-existing. The new `isinstance(beat, dict)` branch has no test (`grep -n "not an object"` in the test file: no match). No producer in reach emits either shape, so severity dropped a level as the bar requires.

**nit — a test name claims more than it exercises.**
`q-system/.q-system/tests/test_consulting_board.py:333`

`test_a_card_older_than_yesterday_is_refused_at_any_hour` runs one hour, 07:00.

## What is sound

**The constant is load-bearing, not decoration.** I mutated `CARD_WRITTEN_AT` to 08:00 through an injected pytest plugin, leaving every test file untouched. Four tests went red, including the boundary and the committed-plist pin:

```
FAILED ...test_after_0730_the_same_card_is_a_REAL_failure_and_still_refused
FAILED ...test_the_boundary_is_the_card_job_itself
FAILED ...test_the_card_is_written_before_this_repo_runs_the_brief
FAILED ...test_and_the_constant_still_matches_the_installed_card_job
4 failed, 8 passed
```

**The label reaches every surface I could find.** Slack line and Notion detail both carry it, including the empty-detail and no-counts paths. `note += detail[:1500]` in `board_rows.py:414` truncates from the front, so the appended suffix survives anything a card row plausibly holds.

**The control flag cannot be forged.** `beat.pop("card_is_yesterdays")` runs before every branch, and `read_heartbeat` is the only producer of `beat` in both `collect` and `buckets`.

**One call site, not several.** `grep` for card and heartbeat consumers across `q-system/` and `plugins/` returns only `consulting_board.py` and its test. No second copy of the old date-equality rule left behind.

**Full suites green:** 147 in `test_consulting_board.py`, 182 across the four neighbouring files.

**Could not verify:** the heartbeat fixture's shape against its real producer. That producer lives in `~/projects/consulting`, outside this worktree, and the sandbox denied the read. The fixture is invented, and the one thing the bar asks me to check hardest about it is the thing I could not check.

## VERDICT: APPROVE WITH NITS

First thing to fix: land the schedule-drift detector (sp-de7afcff) before the founder's board goes another week with no signal that a loaded job has wandered off the committed plist.

FINDINGS:
minor|Accepting yesterday's card silences the only surface that reacted to the drifted brief schedule, and no drift detector replaces it|q-system/.q-system/scripts/consulting_board.py:395
nit|New isinstance(beat, dict) guard is untested and the next line still raises AttributeError on a malformed card value|q-system/.q-system/scripts/consulting_board.py:363
nit|test_a_card_older_than_yesterday_is_refused_at_any_hour exercises only 07:00|q-system/.q-system/tests/test_consulting_board.py:333
END FINDINGS

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-335-20260908-061056.md` (5704 bytes).

```
FINDINGS:
minor|clock_warning re-reads the heartbeat instead of consuming collect's verdict, so the drift signal is lost when the 07:30 card job writes between the two reads|q-system/.q-system/scripts/consulting_board.py:415
minor|morning-brief.py still calls the brief "the 07:00 brief" in three comments, the drifted schedule this PR set out to stop asserting|q-system/.q-system/scripts/morning-brief.py:948
nit|test_the_clock_route_is_silent_on_a_normal_run reads the founder's live consulting heartbeat and stays green under a mutation that deletes the before-the-card condition|q-system/.q-system/tests/test_consulting_board.py:690
END FINDINGS
```

--- reviewer output ---

## Review: PR #335

Read the full diff, both test files run green (154 + 63), and I ran four mutations against a copy in `/tmp`. Three findings survived reproduction. Nothing that gates.

---

### minor 1: the drift signal dies if the 07:30 card job lands mid-run

`clock_warning` re-reads the heartbeat file instead of consuming the verdict `collect` already reached, so the two reads can disagree. When the card job writes between them, the founder's brief still carries a day-old book and Sana gets nothing.

`q-system/.q-system/scripts/consulting_board.py:415` (docstring calls the re-read a feature), consumed at `q-system/.q-system/scripts/morning-brief.py:797`.

Reproducer, run just now:

```
python3 -c '... collect at 07:29 PT with a card stamped 2026-09-02,
            then rewrite the heartbeat to 2026-09-03,
            then clock_warning with the same now ...'
```

Real output:

```
brief lines : ["book: yesterday's card, today's is written at 07:30"]
collect err : None
clock_warning after the 07:30 job wrote: ([], None)
```

The run read a day-old book and reported no drift. Dropping this to minor on purpose: I had to move the file between the two calls, and on the shipped 07:40 slot the window never opens. It only bites a machine drifted to just before 07:30.

Fix shape: have `collect` return the verdict and thread it, or cache the parsed heartbeat per `now`.

---

### minor 2: `morning-brief.py` still calls it "the 07:00 brief"

Commit 64405481 in this PR is titled "stop asserting the drifted schedule as the committed one". Three comments in a file this PR edits still assert exactly that. The committed slot is 07:40, verified from the plist.

Reproducer:

```
grep -n "07:00" q-system/.q-system/scripts/morning-brief.py
```

Real output:

```
668:# meant one hung calendar or mail call held the 07:00 brief, its Slack send and
728:#   keep the 07:00 process alive forever after the brief had "moved on". A
948:#   muted, and the 07:00 brief already carries the daily read. This run only
```

Line 948 is the one that matters: it justifies the hourly job's design against a schedule this repo does not ship.

---

### nit 3: `test_the_clock_route_is_silent_on_a_normal_run` reads the founder's live consulting directory

`q-system/.q-system/tests/test_consulting_board.py:690`. It calls `mb.route_clock(NOW)` with no path injection, so it reads `~/projects/consulting/q-consult/output/ask-crm-state-card-heartbeat.json`. That file is stamped `2026-09-07` while `NOW` is frozen at `2026-09-03`, so `read_heartbeat` errors and the route is silent no matter what the clock logic does.

Mutation 1, replacing `if err is None and beat.get("card_is_yesterdays"):` with `if err is None:` in `clock_warning`:

```
1 failed, 153 passed
FAILED ...::test_and_says_nothing_on_a_normal_run
```

The fixture-driven test caught it. `test_the_clock_route_is_silent_on_a_normal_run` stayed green. It only goes red under mutation 2, which makes `clock_warning` scream unconditionally. It catches a total gutting, not the condition its docstring names, and its greenness comes from the reviewer's home directory.

---

### What is sound

- **Archiving radius.** Making the `card` scope healthy on an early run grants `board_rows` archive authority from a day-old card. Traced it: the board's card rows at 07:00 were painted by yesterday's runs from that same card, so the wanted set is identical and nothing new is archived. No finding.
- **Duplicate Linear tickets.** `alert-to-linear.py` dedupes by alert shape with a counter, and `slack-notify.sh` exit 3 (no API key) raises into `failed` rather than being reported as filed. The drift signal cannot quietly claim a delivery.
- **Fixture faithfulness.** I read the live producer output at `~/projects/consulting/q-consult/output/ask-crm-state-card-heartbeat.json`. Shape is `at`, `card.date`, `card.counts`, `crash`. The invented fixture matches it. This is the failure class this fleet has already shipped once, and it is not here.
- **The new tests have teeth.** Deleting only the four-line acceptance window on a copy turned 7 tests red, including the boundary test and the board-label test. The boundary assertions use literal timestamps rather than reading `CARD_WRITTEN_AT`, so mutating the constant fails them instead of moving the yardstick with the behavior.
- **The plist invariant runs everywhere.** It reads the committed `com.kipi.morning-brief.plist` (Hour 7, Minute 40) rather than a LaunchAgent, so it asserts on CI and not just on one Mac. Confirmed 07:40 is at or after 07:30.
- **Self-declaring heartbeat.** `beat.pop("card_is_yesterdays")` holds; a heartbeat carrying the key cannot label a fresh card.
- **Off switch.** On a machine with no consulting instance, `clock_warning` returns silent rather than paging.

---

Fix first: **finding 2**, the three `07:00` comments in `morning-brief.py`. One-line-each edit, and it is the same claim this PR's own commit message says it removed.

FINDINGS:
minor|clock_warning re-reads the heartbeat instead of consuming collect's verdict, so the drift signal is lost when the 07:30 card job writes between the two reads|q-system/.q-system/scripts/consulting_board.py:415
minor|morning-brief.py still calls the brief "the 07:00 brief" in three comments, the drifted schedule this PR set out to stop asserting|q-system/.q-system/scripts/morning-brief.py:948
nit|test_the_clock_route_is_silent_on_a_normal_run reads the founder's live consulting heartbeat and stays green under a mutation that deletes the before-the-card condition|q-system/.q-system/tests/test_consulting_board.py:690
END FINDINGS

PR #279, opened 2026-08-30
fix(destructive-op): key the MCP deny on the operation, not the vendor (ASK-1144)

assafkip:
## The hole

The founder's CLAUDE.md lists, as NON-NEGOTIABLE and hook-blocked:

> Notion `move-pages`/`delete*`, Gmail `delete_label`, Linear `*delete*`, Vercel mutating ops

Measured on the live hook 2026-08-29: `grep -c mcp__linear__` returns 0, and so
does `grep -c supabase`.

The MCP case named `mcp__plugin_linear_linear__*`. The loaded Linear server is
`mcp__linear__*`. `mcp__supabase__delete_branch` sits in the live tool roster
matched by nothing at all. The rule was stated, the gate was wired, the pattern
matched nothing. A check that runs, passes, and is structurally blind to the
thing it exists to catch.

## Why not a wider wildcard

Adding the two missing servers closes exactly two holes and re-opens the class on
the next server nobody guessed. And a `mcp__linear__*` wildcard denies
`mcp__linear__list_issues` -- the over-block already open as a major on #274, and
the shape that gets a gate switched off.

An MCP tool name is `mcp__<server>__<operation>`. The vendor half drifts across
account connectors, plugin renames and marketplace reinstalls. The operation half
does not: every server that deletes spells it `delete`. The deny now reads the
operation, and the vendor stops mattering.

**This closes the MCP over-block major on #274 as a side effect:** reads are no
longer denied on any server.

## The `un` guard has live subjects

`untrash_message` and `untrash_thread` contain `trash` and both RESTORE. A verb
list without the guard turns the recovery path into a blocked path, which is
worse than the hole it closed. `drop` is deliberately not a verb: `browser_drop`
is a drag gesture, and a SQL `DROP` arrives inside `execute_sql`'s **payload**,
where no name-matching rule can see it. Captured as a note rather than papered
over with a verb that would only look like coverage.

## The second half: dead namespaces

Operation-keying makes a MISSING namespace harmless. It does not make a DEAD one
visible, and a dead entry reading as coverage is what let this survive review.
`mcp-denylist-namespace-check.py` refuses any `mcp__<ns>__` in the hook that
names no registered server. It resolves plugin servers through the cache-clone
manifest, not the directory name, because `vercel` lives in
`cache/temp_git_1788041798495_2plx12/` and a name-from-directory rule reports its
live server as dead.

It reads **executable lines only**, so the hook can document the scar it removed.

## Measurement, watched red first

[...]

PR #274, opened 2026-08-29
fix(destructive-op): inspect argv per stage, not a substring of the line (ASK-1118, ASK-1131)

assafkip:
Supersedes #270, which was branched off a local merge commit carrying another session's unpushed work. This branch is cut from `origin/main`.

One substring test guards the most destructive operations in the fleet, and it was wrong in three directions. Each was proven with a canary before anything was written, and each is proven fixed **through the running harness**, not only in a test.

| # | Hole | Proof |
|---|------|-------|
| 1 | **Too loose.** The fleet dry-run exemption matched the preview flag anywhere in the whole command, so an unrelated mention exempted every fleet-delete in a compound command. The guard's own refusal text says to preview first, so its recommended workflow run in one block disarmed it. | canary deleted |
| 2 | **Too tight.** The same test never matched rsync's short `-n`, so `kipi-update-deletion-guard.py`'s own documented usage line was blocked by this guard. Cost a false spillover finding and an unmeasured propagation claim on #263. (sp-9b01d746) | documented command refused |
| 3 | **Positional.** Every `BASH_DENY` pattern needs its dangerous token immediately after the command name. Insert any innocuous flag and it moves out of match position. A quiet forced push, a quiet `branch -D`, a quiet clean, a quiet hard reset, and a hard reset behind `git -C DIR` all slip it. (ASK-1131) | directory deleted, guard never fired |

## Not three more patterns

Three patterns for three holes leaves the fourth. The hook now splits the command into **stages** on `;` `&&` `||` `|` and newline — the granularity the deny patterns already anchor on, and the split `claude-path-write-guard.py` reached independently — and then asks what each stage actually **is**: program, subcommand, and every flag wherever it sits.

The substring list is **kept** and still runs first. It can only ever DENY, so nothing it already catches changes, including the deliberate prose false positives decided 2026-08-07 and the shapes the stage split cannot see.

**Honest bound, in the hook's own comment.** Tokenising on whitespace is approximate for quoted *operands* and exact for *flags*, which is the only axis these rules turn on. It resolves nothing — no variables, no aliases, no command substitution — so a payload built at runtime is still the substring list's problem.

## How hole 3 surfaced

Two agents measured this guard and disagreed: one saw the removal form blocked, the other had run it twice with `-q` inserted. Neither was wrong, and it was nearly filed as a long-flag-versus-short-flag runbook nit. Two contradictory measurements of one guard meant the guard was broken.

## The landing path, and three defects found in it

[...]

PR #338, opened 2026-09-12
fix(destructive-op): key the MCP deny on the operation, not the vendor (ASK-1144)

assafkip:
> **Replaces #279.** That PR was auto-closed on 2026-09-11 when its head branch
> `sana/ask-1144-mcp-namespace` was renamed to `sana/ask-1144`. The rename was
> made to unblock the fleet dispatcher, which skipped this issue on every
> 15-minute cycle because converge commits to `sana/ask-1144` and the PR was on a
> different branch, so work there reached no PR and no reviewer. GitHub closed the
> PR rather than retargeting it. Same branch, same commits, same base.

## The hole

The founder's CLAUDE.md lists, as NON-NEGOTIABLE and hook-blocked:

> Notion `move-pages`/`delete*`, Gmail `delete_label`, Linear `*delete*`, Vercel mutating ops

Measured on the live hook 2026-08-29: `grep -c mcp__linear__` returns 0, and so
does `grep -c supabase`.

The MCP case named `mcp__plugin_linear_linear__*`. The loaded Linear server is
`mcp__linear__*`. `mcp__supabase__delete_branch` sits in the live tool roster
matched by nothing at all. The rule was stated, the gate was wired, the pattern
matched nothing. A check that runs, passes, and is structurally blind to the
thing it exists to catch.

## Why not a wider wildcard

Adding the two missing servers closes exactly two holes and re-opens the class on
the next server nobody guessed. And a `mcp__linear__*` wildcard denies
`mcp__linear__list_issues` -- the over-block already open as a major on #274, and
the shape that gets a gate switched off.

An MCP tool name is `mcp__<server>__<operation>`. The vendor half drifts across
account connectors, plugin renames and marketplace reinstalls. The operation half
does not: every server that deletes spells it `delete`. The deny now reads the
operation, and the vendor stops mattering.

**This closes the MCP over-block major on #274 as a side effect:** reads are no
longer denied on any server.

## The `un` guard has live subjects

`untrash_message` and `untrash_thread` contain `trash` and both RESTORE. A verb
list without the guard turns the recovery path into a blocked path, which is
worse than the hole it closed. `drop` is deliberately not a verb: `browser_drop`
is a drag gesture, and a SQL `DROP` arrives inside `execute_sql`'s **payload**,
where no name-matching rule can see it. Captured as a note rather than papered
over with a verb that would only look like coverage.

## The second half: dead namespaces

Operation-keying makes a MISSING namespace harmless. It does not make a DEAD one
visible, and a dead entry reading as coverage is what let this survive review.
`mcp-denylist-namespace-check.py` refuses any `mcp__<ns>__` in the hook that
names no registered server. It resolves plugin servers through the cache-clone
manifest, not the directory name, because `vercel` lives in
`cache/temp_git_1788041798495_2plx12/` and a name-from-directory rule reports its
live server as dead.

It reads **executable lines only**, so the hook can document the scar it removed.

## Measurement, watched red first

Before the fix: 8 failed, 11 passed. After: 19 passed.

The 8 reds were the five live holes plus the three missing-checker cases. The 11
greens before the fix included the negative self-test (a recursive-force delete
is still denied) and every allow-case, which is what proves the driver worked
rather than the hook.

Two cases exist so the checker can go red for the reason we care about:
`test_namespace_check_goes_red_on_a_planted_dead_namespace`, and an empty
namespace scan is a refusal in `main()` rather than a pass.

Regression:

- `security-remediation/tests/test-destructive-op-deny.sh` -- 35 PASS, 0 FAIL
- `test_destructive_op_deny_anchor.py` -- 94 passed, 1 failed: the vendored-copy
  drift detector, correctly red until the paired proposal lands the same content
  in the live hook.

## Known, not fixed here

The sanctioned landing path refuses this file: `replace is only permitted on
.claude/rules/*.md`. So the repo currently cannot repair its own guard through
`apply-claude-changes`. Widening that op to `.claude/hooks/*.sh` is its own
decision with its own blast radius (the engine's own docstring says so) and gets
its own PR rather than arriving as a side effect of this one.


assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-338-20260923-192746.md` (12961 bytes).

```
FINDINGS:
major|Transparent-prefix padding (3000 `sudo` tokens, 15KB) takes 5.41s past the wired 5s hook timeout, so the verdict is discarded and `rm -v -rf DIR` proceeds|q-system/.q-system/hooks/destructive-op-deny.sh:430
major|A crashing or stub installed hook becomes the differential oracle, so the installer refuses to repair a broken guard and names the wrong cause|q-system/.q-system/scripts/install-claude-hooks.py:334
minor|test_prefix_padding_does_not_blow_the_timeout pads `sudo` as operands of `echo`, so the prefix-strip loop never runs and the test cannot go red for its stated reason|q-system/.q-system/tests/test_argv_prefilter.py:120
minor|The `*authenticate*` exemption is an unanchored substring checked before every verb rule, so any operation containing "authenticate" is allowed unconditionally|q-system/.q-system/hooks/destructive-op-deny.sh:849
minor|The PR body claims reads are no longer denied on any server and closes the #274 over-block, but the Vercel arm still denies `list_deployments`|q-system/.q-system/hooks/destructive-op-deny.sh:887
END FINDINGS
```

--- reviewer output ---

Review complete. Read the PR, drove the hook against the live MCP roster, raced the installer, and ran all five new suites (73 passed, 42 subtests).

## Findings

### 1. major — transparent-prefix padding still blows the wired 5s timeout (round 8 of the same bypass)

`q-system/.q-system/hooks/destructive-op-deny.sh:430` (the strip loop), reached from the unbounded first-position scan at `:579`.

The file carries three separate comments asserting this class is closed: "A CEILING PADDING CANNOT MOVE", "the ceiling is a budget for the whole invocation", "NO TRANSPARENT-PREFIX ARM... the filter now admits only tokens that can themselves deny". None of them guards this path. `_ARGV_MAX_RESCANS` counts only `rm`/`git` tokens, and `sudo` is deliberately not admitted, so a prefix pad costs zero rescans. `_ARGV_MAX_TOKENS=600` lives inside the `git` arm only. The strip loop at `:430` rebuilds the array per iteration, so it is O(n squared), and the first-position call runs it on the whole stage.

Reproducer, run from the worktree:

```
python3 -c '
import json, os, subprocess, tempfile, time
HOOK = "q-system/.q-system/hooks/destructive-op-deny.sh"
home = tempfile.mkdtemp(prefix="probe-")
os.makedirs(os.path.join(home, "." + "claude", "audit"), exist_ok=True)
def harness(cmd, wall=5.0):
    env = dict(os.environ); env.pop("ALLOW_DESTRUCTIVE", None); env["HOME"] = home
    p = subprocess.Popen(["bash", HOOK], stdin=subprocess.PIPE,
        stdout=subprocess.PIPE, stderr=subprocess.PIPE, text=True, env=env)
    t0 = time.time()
    try:
        out, _ = p.communicate(json.dumps(dict(tool_name="Bash",
            tool_input=dict(command=cmd), cwd="/tmp")), timeout=wall)
    except subprocess.TimeoutExpired:
        p.kill(); p.communicate()
        return "KILLED at %.0fs -> no verdict -> the tool call PROCEEDS" % wall
    el = time.time()-t0
    return "%s in %.2fs" % ("deny" if (p.returncode == 2 or out.strip()) else "allow", el)
TAIL = chr(114)+chr(109) + " -v " + chr(45)+chr(114)+chr(102) + " /Users/assafkipnis/irreplaceable"
print("control, no padding :", harness(TAIL))
for n in (3000, 3500, 5000):
    print("sudo x%-5d prefix  :" % n, harness(" ".join(["sudo"]*n) + " " + TAIL))
'
```

Real output:

```
control, no padding : deny in 0.07s
sudo x3000  prefix  : KILLED at 5s -> no verdict -> the tool call PROCEEDS
sudo x3500  prefix  : KILLED at 5s -> no verdict -> the tool call PROCEEDS
sudo x5000  prefix  : KILLED at 5s -> no verdict -> the tool call PROCEEDS
```

The wired timeout is measured, not assumed. From the live `~/.claude/settings.json`:

```
PreToolUse {"type": "command", "command": "/Users/assafkipnis/.claude/hooks/destructive-op-deny.sh", "timeout": 5}
```

Raw timings without the wall clock, so the crossing is visible rather than inferred: 2500 tokens 3.9s, 3000 tokens 5.4s, 3500 tokens 7.36s, 4000 tokens 10.41s. Same curve for `env`, `nice`, `command` and `k=v` padding (5.3s, 5.48s, 6.66s, 5.14s at 3000). 15 KB of command text is the whole cost. The substring list does not cover this shape: `rm -v -rf` has no `[rRf]` in the cluster immediately after the program name, which is why the control denies from the argv scan and nowhere else.

I constructed the padding. No producer in this repo emits `sudo` 3000 times, so by the calibration rule this sits at major rather than blocker, even though the thing it lets through is an unrecoverable directory delete.

### 2. major — the installer refuses to repair a broken installed hook, and misreports why

`q-system/.q-system/scripts/install-claude-hooks.py:334` and `:468`.

`decision_of` maps returncode 2 to `"deny"`. A hook that fails to parse also exits 2. So a **crashing** installed copy becomes the strongest possible differential oracle: every corpus row reads as "the installed hook denies this", and the candidate is then refused for "disarming" a hook that is not running at all. The more broken the installed guard, the harder the installer refuses to fix it.

The second shape is the env-var differential at `:418`. It fires whenever the installed copy reads fewer variables than the source, which is true of any stub.

Reproducer, three seeds, one of them the copy actually on this machine:

```
python3 -c '
import json, os, subprocess, tempfile
INST = "q-system/.q-system/scripts/install-claude-hooks.py"
REAL = os.path.expanduser("~/" + "." + "claude/hooks/destructive-op-deny.sh")
def probe(seed_text, label):
    home = tempfile.mkdtemp(prefix="seed-")
    cd = os.path.join(home, "." + "claude")
    os.makedirs(os.path.join(cd, "hooks"), exist_ok=True)
    dst = os.path.join(cd, "hooks", "destructive-op-deny.sh")
    json.dump(dict(hooks=dict(PreToolUse=[dict(matcher="*", hooks=[dict(type="command", command=dst)])])),
              open(os.path.join(cd, "settings.json"), "w"))
    open(dst, "w").write(seed_text); os.chmod(dst, 0o755)
    p = subprocess.run(["python3", INST, "--home", home, "--dry-run"],
                       capture_output=True, text=True, timeout=600)
    print("[%s] exit=%d" % (label, p.returncode))
    print("   " + (p.stdout or p.stderr).strip()[:400])
probe(open(REAL).read(), "the REAL 501-line copy now installed on this machine")
probe(chr(35) + "!/bin/bash" + chr(10) + "exit 0" + chr(10), "a stub, e.g. hand-edited during an incident")
probe(open(REAL).read()[:9000], "a truncated copy (short write / partial download)")
'
```

Real output:

```
[the REAL 501-line copy now installed on this machine] exit=0
   destructive-op-deny.sh: WOULD INSTALL (differs)
[a stub, e.g. hand-edited during an incident] exit=1
   REFUSED destructive-op-deny.sh: the source reads environment variable(s) the installed hook does not (ALLOW_DESTRUCTIVE, HOME). A probe cannot open a door whose key it does not hold, so a new env read is refused rather than tested.
[a truncated copy (short write / partial download)] exit=1
   REFUSED destructive-op-deny.sh: the installed hook DENIES `ls -la` and the source does not (allow). A hook may be repaired, never disarmed -- and this was found by comparing against what is running, not against a list someone wrote.
```

Neither input is invented for the occasion. The backup comment at `install-claude-hooks.py:512` names "had been hand-edited during an incident" as the live case, and the ASK-1118 scar in the hook header is a bad write landing in exactly this file. At 3am `kipi update` prints REFUSED, the warning says the guard "may not match the reviewed copy", the installed stub keeps running, and the printed cause is wrong: it says a new env read when the truth is that the reference does not execute. A human fixes it with `cp`, so it is major and not blocker. One guard closes both shapes: skip the differential and the env diff when `parses_as_bash` on the installed copy is false, because a copy that does not parse is not a gate to be weaker than.

### 3. minor — the test that exists to pin prefix padding cannot go red for that reason

`q-system/.q-system/tests/test_argv_prefilter.py:120`.

`command = "echo " + " ".join(["sudo"] * 300)` places the `sudo` tokens as **operands of `echo`**. `argv_deny_reason` resolves the program to `echo`, hits the case fallthrough, and returns before the strip loop runs even once. The fixture measures a path that has no prefix in it. Scaling it tenfold, against the same 3000 tokens in the position a transparent prefix actually occupies:

```
A) suite fixture shape scaled to 3000 (sudo as OPERANDS of echo):
   ('allow', 0.14, 0)
B) same 3000 tokens as a real PREFIX, no wall clock:
   ('deny', 5.41, 0)
```

0.14s versus 5.41s on identical token counts. Moving the pad in front of the program is the one-line change that makes this test able to fail. Its docstring says padding "with a shape the filter DROPS" is how six rounds stayed green; this is a seventh instance of that.

### 4. minor — the auth exemption is an unanchored substring evaluated before every verb rule

`q-system/.q-system/hooks/destructive-op-deny.sh:849`. `*authenticate*` matches anywhere in the operation name, and it is checked first so no verb rule can reach the call.

```
deny   mcp__linear__delete_issue
allow  mcp__linear__delete_authenticated_issue
allow  mcp__linear__authenticate_and_delete_all
allow  mcp__linear__delete_issue_unauthenticated
```

No server in the current roster emits an operation of that shape, so per the calibration rule this drops a level and I am saying so. It still contradicts the change's own thesis: the whole design rests on the operation half being stable, and the exemption that overrides it is the one part matched loosely. `^authenticate$|^complete_authentication$` costs nothing.

### 5. minor — the PR body's "reads are no longer denied on any server" is false, and it is used to close a major on another PR

`q-system/.q-system/hooks/destructive-op-deny.sh:887`. The Vercel arm still denies at the server level:

```
allow  mcp__plugin_vercel_vercel__authenticate
deny   mcp__plugin_vercel_vercel__list_deployments
```

The hook's own comment at `:882` is honest about this ("The resulting read over-block is real and is captured, not hidden"). The PR description contradicts it and claims the #274 over-block major closes as a side effect. It does not, for Vercel.

## What is sound

- The headline claim holds. `mcp__claude_ai_Gmail__delete_label`, `mcp__linear__delete_comment`, `mcp__supabase__delete_branch` and `reset_branch`, `mcp__claude_ai_Google_Drive__trash_file` and `mcp__claude_ai_Notion__notion-move-pages` all deny, on 28 real roster names, with zero vendor segment guessed.
- No read over-block outside Vercel. `list_issues`, `get_issue`, `save_issue`, `retire_issue_label`, `restore_issue_label`, `merge_diff`, `list_tables`, `notion-search`, `browser_close`, `list_meetings` all allow.
- The `un` guard has teeth where it matters and does not over-reach. `untrash_message`, `unmark_message_spam`, `unDelete` allow; `hardDelete`, `batchDelete`, `deleteAll`, `Delete`, `DELETE` all deny. The camelCase fold plus the `^un_` rejoin behave as documented.
- `reset_session` is exempt and `reset_branch` is not, and the ordering that makes that true is real, not just commented.
- Ordinary long commands do not trip the ceiling or the timeout: 3000 file paths after `grep -rn` allow in 0.20s, 4000 words after `echo` allow in 0.17s, `git commit -m` allows in 0.08s.
- The namespace checker is machine-independent as claimed, finds the one executable namespace, and passes: "namespaces named by the hook: 1 / mcp__plugin_vercel_vercel__ OK".
- The install race does not corrupt the destination. Eight concurrent installers into one probe home left no leftover `.install.tmp`, and the destination stayed executable and parseable.
- All five new suites pass as claimed: 73 passed, 42 subtests passed in 34.25s.
- `install-claude-hooks.py --check` correctly reports the real drift on this machine: installed copy is 501 lines, the repo's is 898.

Noted, not filed as findings because the PR improves on the prior state rather than regressing it: `mcp__supabase__apply_migration` and `mcp__supabase__merge_branch` both measured `allow`, and they are the same arbitrary-DDL class as `execute_sql`, which the hook's comment does disclose as spillover. The disclosure names only one of the three.

## VERDICT: REQUEST CHANGES

Not BLOCK. Nothing here is unrecoverable if it merges: finding 1 is a pre-existing class being newly vendored, and finding 2 is fixable by hand with `cp`.

Fix finding 2 first. It is new code in this PR, it fires on the founder's own machine through `kipi update`, and it blocks the repair path precisely when the guard is most broken, which is the failure this whole change exists to end. One condition does it: do not use an installed copy as the differential or env-diff oracle unless `parses_as_bash` on it returns true.

FINDINGS:
major|Transparent-prefix padding (3000 `sudo` tokens, 15KB) takes 5.41s past the wired 5s hook timeout, so the verdict is discarded and `rm -v -rf DIR` proceeds|q-system/.q-system/hooks/destructive-op-deny.sh:430
major|A crashing or stub installed hook becomes the differential oracle, so the installer refuses to repair a broken guard and names the wrong cause|q-system/.q-system/scripts/install-claude-hooks.py:334
minor|test_prefix_padding_does_not_blow_the_timeout pads `sudo` as operands of `echo`, so the prefix-strip loop never runs and the test cannot go red for its stated reason|q-system/.q-system/tests/test_argv_prefilter.py:120
minor|The `*authenticate*` exemption is an unanchored substring checked before every verb rule, so any operation containing "authenticate" is allowed unconditionally|q-system/.q-system/hooks/destructive-op-deny.sh:849
minor|The PR body claims reads are no longer denied on any server and closes the #274 over-block, but the Vercel arm still denies `list_deployments`|q-system/.q-system/hooks/destructive-op-deny.sh:887
END FINDINGS

assafkip:
Round 1 minor 5 (the Vercel arm still denies list_deployments while the PR body says reads are no longer denied on any server): declined for this PR, and the claim is corrected here. The over-block that was closed is on the operation-keyed arm. Vercel keeps its own arm, so list_deployments is still denied there. That is a false positive, not a hole, and it is not widened in a security PR. Minor 3 is fixed by the new program-position test in c47622ff. Minor 4 is fixed by exact auth names in the same commit.

assafkip:
## Verdict: unstated

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-338-20260923-210035.md` (2165 bytes).

```
(no complete findings block parsed from this review)
```

--- reviewer output ---

**Finding 4 upgraded from "the primitive is blind" to "the whole gate accepts it end to end."** The full `refuse_if_weaker()` run finished after my review was written. Result:

```
cd .review-scratch && python3 r3_installer.py

refuse_if_weaker() on a candidate that sleeps 10s before every verdict
   (production kills it at 5s and DISCARDS the deny -> fails OPEN):
   -> None   [323s spent probing]

   VERDICT: ACCEPTED -- the installer would make this the live guard
```

That is the complete gate, not just `decision_of`: shebang, `bash -n`, env-var differential, four MUST_DENY canaries, three MUST_ALLOW cases, and the 30-row differential corpus all pass a candidate that answers 10s after the wired 5s timeout. `None` means no refusal.

Two things it adds to what I already filed.

**The gate cannot see the defect it was built to stop.** `install-claude-hooks.py:325` is the single place every behavioural check gets its verdict from, and it reads "deny" off a hook production would have killed and discarded. Findings 1 and 2 are exactly that hook shape. So the installer would have installed both, called them verified, and printed `installed, parses, executable, and registered`.

**323 seconds.** That is the gate's own runtime on one slow candidate, inside `kipi-update.sh`'s `install_vendored_hooks`, which prints nothing until `python3 "$installer"` returns because it captures into `_out` first (`kipi-update.sh:339`). A founder running `kipi update` sees a silent terminal for five minutes with no progress line. Not a correctness bug, and worth one line of streaming.

Severity stays **minor**, unchanged. The consequence it enables is already counted in findings 1 and 2, and I am not double-billing the same damage. What changed is the claim's strength: I wrote "the primitive is blind," and the evidence now says the assembled gate is blind. Same fix, one argument: pass the wired timeout into `decision_of` and treat an overrun as a refusal. That turns findings 1 and 2 from 3am discoveries into install-time refusals with a readable cause.

Verdict is unchanged: **REQUEST CHANGES**, fix the check ordering and the `/` boundary class first.

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-338-20260923-225230.md` (10278 bytes).

```
FINDINGS:
major|Stage-count padding (600 `;`-stages, 2.4KB, 2 words) takes the hook past its wired 5s timeout, so the fleet-wide-delete deny is discarded and `kipi update` proceeds|q-system/.q-system/hooks/destructive-op-deny.sh:599
minor|Every timing fixture pads with space-separated tokens, the exact axis the new word ceiling bounds, so no test can go red on the stage-count axis|q-system/.q-system/tests/test_argv_prefilter.py:142
minor|_env_vars_read's `isupper() or "_" in m` filter makes the env-read differential blind to a lowercase single-word backdoor variable, contradicting its own docstring|q-system/.q-system/scripts/install-claude-hooks.py:308
END FINDINGS
```

--- reviewer output ---

Review complete. I drove the hook directly, ran all five new suites, raced the installer against four broken-seed shapes, and probed the whole live MCP roster.

## Findings

### 1. major — stage-count padding still blows the wired 5s timeout, so the fleet-delete deny is discarded

`q-system/.q-system/hooks/destructive-op-deny.sh:599` (the new `_ARGV_MAX_WORDS` ceiling) and `:792` (the FLEET_DENY per-stage loop it does not cover).

Round 9 added the ceiling and its commit message claims: "Skipping removes the slow part whatever the payload, so every check below runs inside the budget." That sentence is false as shipped. The ceiling measures `wc -w`. The actual cost driver is the **stage count**: the FLEET_DENY loop runs four `grep -E` forks per `;`-separated stage, and it sits *after* the argv scans, so skipping the scans does not bound it. Stages separated by `;` with no surrounding whitespace produce unbounded stages at two words, so the ceiling never fires.

Reproducer (writes nothing, read-only):

```
python3 - <<'PY'
import json, os, subprocess, tempfile, time
HOOK = "q-system/.q-system/hooks/destructive-op-deny.sh"
home = tempfile.mkdtemp(prefix="probe-")
FLEET = "k"+"ipi upd"+"ate"
def run(cmd, wall):
    env = dict(os.environ); env.pop("ALLOW_DESTRUCTIVE", None); env["HOME"] = home
    p = {"tool_name":"Bash","tool_input":{"command":cmd},"cwd":"/tmp"}
    t0 = time.time()
    try:
        r = subprocess.run(["bash",HOOK],input=json.dumps(p),capture_output=True,
                           text=True,timeout=wall,env=env)
    except subprocess.TimeoutExpired:
        return "KILLED -> verdict DISCARDED -> the call PROCEEDS", time.time()-t0
    return ("deny" if (r.stdout or "").strip() else "allow"), time.time()-t0
for n in (100, 400, 500, 600, 800):
    cmd = ("a=0;"*n) + FLEET
    v,_  = run(cmd, 5)
    v2,e = run(cmd, 120)
    print("stages=%-5d bytes=%-6d wc-w=%-2d | at wired 5s: %-46s | unbounded: %s %.2fs"
          % (n, len(cmd), len(cmd.split()), v, v2, e))
PY
```

Real output:

```
stages=100   bytes=411    wc-w=2  | at wired 5s: deny                                           | unbounded: deny 0.98s
stages=400   bytes=1611   wc-w=2  | at wired 5s: deny                                           | unbounded: deny 4.09s
stages=500   bytes=2011   wc-w=2  | at wired 5s: deny                                           | unbounded: deny 4.53s
stages=600   bytes=2411   wc-w=2  | at wired 5s: KILLED -> verdict DISCARDED -> the call PROCEEDS | unbounded: deny 5.51s
stages=800   bytes=3211   wc-w=2  | at wired 5s: KILLED -> verdict DISCARDED -> the call PROCEEDS | unbounded: deny 8.28s
```

The wired budget is measured, not assumed. From the live `~/.claude/settings.json`:

```
PreToolUse {"type": "command", "command": "/Users/assafkipnis/.claude/hooks/destructive-op-deny.sh", "timeout": 5}
```

Same shape against `rsync -a --delete /a/ /b/`: 600 stages, 2425 bytes, 6.13s. Scoping matters and I checked it: `rm` and `git` denials short-circuit in the argv loops and stay fast (4000 stages ahead of `rm -v -rf` denies in 0.23s). The live half is the FLEET_DENY layer only, which is the `kipi update` fleet-wide-delete guard, the voicekit-from-19-instances scar.

Severity held at **major**, the same level round 1 gave this class. Not escalated: same underlying issue, same recoverability (the fleet delete is restorable from git, as it was on 2026-08-07). I constructed the padding, so per the calibration rule it does not rise to blocker.

Round 1's `sudo`-padding repro is genuinely fixed. I ran it: 3000, 5000 and 20000 `sudo` tokens all deny in under 0.12s.

### 2. minor — no timing test drives the axis the new ceiling misses

`q-system/.q-system/tests/test_argv_prefilter.py:142`.

Every padding fixture in the suite is space-separated: `" ".join(["a=1"]*3000)`, `" ".join(["sudo"]*3000)`, `" ; ".join([stage]*30)`. All of them produce a high `wc -w`, which is exactly the axis `_ARGV_MAX_WORDS` bounds. `test_the_ceiling_bounds_the_invocation_not_the_stage` is the closest, and it uses 14 and 30 stages of 400 space-separated tokens each, so it goes red on word count, never on stage count. The suite is green at 78 passed and cannot observe finding 1. One `"a=0;"*600` case makes it able to fail.

### 3. minor — the installer's env-read differential is blind to a lowercase variable, contradicting its own docstring

`q-system/.q-system/scripts/install-claude-hooks.py:308`.

`_env_vars_read` keeps only names where `m.isupper() or "_" in m`. Its docstring says the honest boundary is a backdoor "that repurposes a variable the hook already reads", and that "a backdoor's `${SOME_NEW_VAR:-}` is read and never assigned, so it is still caught". A new variable spelled in lowercase with no underscore is neither, and it is not caught.

Reproducer:

```
python3 - <<'PY'
import importlib.util
spec = importlib.util.spec_from_file_location("ich",
        "q-system/.q-system/scripts/install-claude-hooks.py")
m = importlib.util.module_from_spec(spec); spec.loader.exec_module(m)
real = open("q-system/.q-system/hooks/destructive-op-deny.sh").read()
door = '[ "${%s:-}" = 1 ] && exit 0'
for label, name in (("UPPER_NEW (docstring says caught)", "KIPI_BYPASS"),
                    ("lowercase, no underscore", "kipibypass")):
    cand = real.replace("\nINPUT=", "\n" + (door % name) + "\nINPUT=", 1)
    new = m._env_vars_read(cand) - m._env_vars_read(real)
    r = m.refuse_if_weaker("destructive-op-deny.sh", cand, real)
    print("%-36s new_env=%-16s -> %s" % (label, sorted(new),
          "REFUSED" if r else "ACCEPTED -> would become the live guard"))
PY
```

Real output:

```
UPPER_NEW (docstring says caught)    new_env=['KIPI_BYPASS']  -> REFUSED
lowercase, no underscore             new_env=[]               -> ACCEPTED -> would become the live guard
```

Minor, and I am saying why: I constructed the source, it needs a malicious hook to already be merged, and the behavioural canaries are a separate layer. The fix looks like deleting the `isupper() or "_" in m` filter, since the assigned-name subtraction below it is already exact for shell locals.

## The process read, which is worth more than another nit

The brief asks me to say so if rounds are miscalibrated. They are not. Every repro filed against this PR was real and every fix landed. What is miscalibrated is the **fix shape**, and the file already says so in its own words at `destructive-op-deny.sh:361`: "Seventh instance of one bypass in one predicate. That is a signal about the SURFACE."

Counting from the comments in the file plus this PR's two review rounds, this is instance ten of one class: a shell scanner with a 5s budget, patched one padding axis at a time. Rounds bounded fork count, then admitted-token set, then window size, then rescan count, then the quote strip, then per-stage versus per-invocation, then word count. Each bound was correct and each time the padding moved to a loop that was not bounded. Finding 1 is the next one over. Cheap bound available now: cap the stage count and refuse past it, the same fail-closed direction the git arm already uses. The durable fix is not an eleventh ceiling; it is getting this predicate off a per-fork shell budget.

## What is sound

- The headline claim holds on real data. I drove the full live MCP roster, 170 tool names, through the hook. 16 deny, and every one is genuinely destructive: Linear `delete_*`, Gmail `delete_label` and `trash_*`, Drive `trash_file`, Calendar `delete_event`, Supabase `delete_branch` and `reset_branch`, Notion `notion-move-pages`. Zero vendor segment guessed.
- No read over-block outside the deliberate Vercel arm. `list_issues`, `get_issue`, `save_issue`, `retire_issue_label`, `restore_issue_label`, `merge_diff`, `list_tables`, `notion-search`, `browser_close`, `browser_drop` all allow.
- The `un` guard holds. `untrash_message`, `untrash_thread`, `unmark_message_spam` allow. The camelCase fold plus the `^un_` rejoin behave as documented.
- Round 1 finding 1 is fixed. 3000, 5000 and 20000 `sudo` tokens ahead of `rm -v -rf` now deny in under 0.12s.
- Round 1 finding 2 is fixed and I could not falsify it. Four installed-hook seeds (the real 501-line copy, a permissive stub, a truncated copy, a syntax-error copy) all print the oracle refusal and proceed to `WOULD INSTALL` rather than refusing the repair.
- Round 1 minor 4 is fixed. `delete_authenticated_issue` and `authenticate_and_delete_all` both deny now.
- Round 2's finding is fixed. A candidate that sleeps 10s before every verdict is REFUSED in 5s with the correct cause, and `WIRED_HOOK_TIMEOUT_S` is 5, matching production.
- The ceiling does not over-block ordinary work. `echo` with 3000 words allows; a 4MB single-stage command denies fail-closed rather than timing out.
- The namespace checker is machine-independent as claimed and passes: one executable namespace, `mcp__plugin_vercel_vercel__`, OK.
- All five new suites pass: 78 passed, 42 subtests, 34.87s.

Round 1 minor 5 (Vercel `list_deployments` over-block) the author declined and corrected in the PR thread. That is settled and I am not re-raising it.

## VERDICT: REQUEST CHANGES

Not BLOCK. Nothing here is unrecoverable: finding 1 lets a fleet rsync through, which is git-restorable and is the same exposure the currently-installed hook already has, so merging does not make the machine worse.

**Fix first:** bound the stage count in `destructive-op-deny.sh` and refuse past it, and add the `"a=0;"*600` case to `test_argv_prefilter.py` so the suite can go red for that reason.

FINDINGS:
major|Stage-count padding (600 `;`-stages, 2.4KB, 2 words) takes the hook past its wired 5s timeout, so the fleet-wide-delete deny is discarded and `kipi update` proceeds|q-system/.q-system/hooks/destructive-op-deny.sh:599
minor|Every timing fixture pads with space-separated tokens, the exact axis the new word ceiling bounds, so no test can go red on the stage-count axis|q-system/.q-system/tests/test_argv_prefilter.py:142
minor|_env_vars_read's `isupper() or "_" in m` filter makes the env-read differential blind to a lowercase single-word backdoor variable, contradicting its own docstring|q-system/.q-system/scripts/install-claude-hooks.py:308
END FINDINGS

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-338-20260923-235409.md` (9222 bytes).

```
FINDINGS:
major|Byte-length padding (one 5MB word, 8 words total, 1 stage separator) takes the hook past its wired 5s timeout, so the deny is discarded and `git push --force`, `git push origin +main`, `git clean -fd` and `rm -v -rf DIR` proceed|q-system/.q-system/hooks/destructive-op-deny.sh:611
minor|Every timing fixture pads word count or stage count and none varies a token's byte length, so the suite stays green on the axis that is live|q-system/.q-system/tests/test_argv_prefilter.py:142
minor|The 64-token rescan window is the only argv coverage behind a prefix carrying its own option, so `sudo -u root git push origin <63 refspecs> --force` is allowed, contradicting the comment that the first-position scan catches it at any length|q-system/.q-system/hooks/destructive-op-deny.sh:320
END FINDINGS
```

--- reviewer output ---

Review complete. Tree is clean (scratch removed, nothing committed).

## Findings

### 1. major — byte-length padding blows the wired 5s timeout, under every ceiling this PR added

`q-system/.q-system/hooks/destructive-op-deny.sh:611`

Every ceiling in this hook counts **words** or **stages**: `_MAX_STAGE_SEPS=200` (:135), `_ARGV_MAX_WORDS=1000` (:611, `wc -w`), `_ARGV_MAX_RESCANS=500`, `_ARGV_MAX_TOKENS=600`. None of them bounds the **byte length of a single token**. One 5MB word costs ~1.3s per MB across the grep/`tr` forks, and it is 1 word.

Round six removed the coarse `git push ... --force` substring pattern (:189-205), so force-push, `+refspec` and `git clean -fd` are now decided **only** by the argv scan, which is the part a fat word starves. `rm -rf` and `git reset --hard` still short-circuit in `BASH_DENY` and stay fast, which is why this is not visible from the usual probes.

Reproducer, run against this branch's hook, `WIRED_HOOK_TIMEOUT_S = 5` taken from the PR's own `install-claude-hooks.py:319`:

payload is `echo <5MB single word> > /tmp/b; <carrier>`, 8 words, 1 stage separator.

```
carrier                    alone(5s)    fat(5s)         fat(no limit)
force push                 deny         TIMEOUT->ALLOW  deny @ 6.26s
branch -D                  deny         deny            deny @ 1.83s
clean -fd                  deny         TIMEOUT->ALLOW  deny @ 5.98s
reset --hard               deny         deny            deny @ 1.92s
plus refspec               deny         TIMEOUT->ALLOW  deny @ 6.45s
rm reordered               deny         TIMEOUT->ALLOW  deny @ 6.74s
rm -rf (substring list)    deny         deny            deny @ 1.67s
```

Scaling, same shape with `rm -v -rf /tmp/victim-dir` as carrier:

```
blob=1024 KB  words=8  seps=1  -> deny  1.41s | prod: answered in budget
blob=2048 KB  words=8  seps=1  -> deny  2.59s | prod: answered in budget
blob=4096 KB  words=8  seps=1  -> deny  5.21s | prod: KILLED AT 5s -> verdict discarded -> command RUNS
```

Two claims this falsifies. The round-9 commit says "Skipping removes the slow part whatever the payload, so every check below runs inside the budget" (:600-604). Round 3's own **What is sound** list says "a 4MB single-stage command denies fail-closed rather than timing out" — that holds only when the 4MB is spread over more than 1000 words, where the word ceiling fires. In 8 words it times out.

Severity held at **major**, the level rounds 1 and 3 gave this class. Not escalated, same underlying issue. I constructed the padding, and no producer in this repo emits a 5MB single word, so by the calibration rule it does not rise to blocker.

### 2. minor — no timing fixture varies byte length, so the suite cannot go red on the live axis

`q-system/.q-system/tests/test_argv_prefilter.py:142`

Every timing fixture pads on the axis the previous round bounded. `:120` pads `sudo` tokens, `:149-151` pads 3000 space-separated words, `:163` pads 600 `;` stages, `:203` pads `rm` tokens, `:227` pads 400-token stages, `:321` pads quoted `"git"` tokens. Not one holds token count constant and grows a token's length.

Reproducer: the fixture that would go red is the finding-1 payload. `test_stage_padding_cannot_push_the_fleet_deny_past_the_timeout` (:159) passes today, and so does everything else, while a 5MB word allows a force push.

This is the same minor round 3 filed on the stage axis. It was answered by adding one stage fixture rather than by making the fixture generator vary the axis, so the next axis is uncovered again.

### 3. minor — the 64-token rescan window is the only coverage behind an unrecognized leading word, and the comment claiming otherwise is false

`q-system/.q-system/hooks/destructive-op-deny.sh:320`

The comment reads: "The FIRST-position scan still reads the entire stage unbounded, so an ordinary `git push ... --force` at any length is still caught there; what the rescans add is coverage of a program token HIDDEN mid-line."

The first-position scan resolves a program only through the transparent-prefix strip loop. A prefix carrying its **own option** leaves that option in program position, `argv_deny_reason` returns 1, and the windowed rescan at `:668` is the only remaining coverage.

```
=== sudo -u root git push origin <N refspecs> --force ===
refspecs=60   words=67   -> deny
refspecs=63   words=70   -> allow
refspecs=120  words=127  -> allow

=== control: no prefix ===
refspecs=120  -> deny

=== 70 refspecs, other prefixes ===
env -i           -> allow
nice -n 10       -> allow
sudo -u deploy   -> allow
command -p       -> allow

=== other git verbs, sudo -u root, 70 pad tokens ===
clean -fd     -> allow
branch -D     -> allow
reset --hard  -> allow
push --mirror -> allow
```

Same on `rm`: `sudo -u root rm f0..f62 -rf /tmp/victim` → allow at 69 words, 0.14s. `rm f0..f69 -rf /tmp/victim` without the prefix → deny.

`sudo -u root rm -v -rf DIR` is the exact case ASK-1131 round 2 added the rescan for (:674-682), so the window is not "extra coverage" here, it is the whole coverage.

Dropped to **minor** rather than major: it needs ≥64 tokens between the program name and its decisive flag, which no producer in this repo emits.

## What is sound

Attacks that did not survive:

- **MCP operation-keying.** Swept 30 destructive and 18 read/restore tool names copied from this session's live roster. Every op the founder's CLAUDE.md names denies (Linear `delete_*`, Gmail `delete_label`, Notion `notion-move-pages`, Supabase `delete_branch` and `reset_branch`, Docs `delete`, Drive `trash_file`). Every read and every restore allows: `list_issues`, `get_issue`, `restore_*_label`, `untrash_message`, `untrash_thread`, `unmark_message_spam`, `browser_drop`, `reset_session`. The `un` guard and the camelCase fold both hold in both directions. Zero over-blocks except the two deliberate `unlabel_*` entries.
- **`BASH_DENY` short-circuit.** `rm -rf`, `git reset --hard` and `git branch -D` still deny in under 2s behind a 5MB blob, which is what keeps finding 1 from reaching the classic forms.
- **The stage cap at short words.** 199 stages of `echo hi` → 2.86s; 199 stages naming `rsync` → 3.02s. Under budget.
- **`fleet_stage_is_preview`'s `*--dry*` substring.** I tried to build a real fleet delete carrying `--dry`. `kipi-update.sh:54` rejects unknown arguments with exit 1, so `--dryfoo` cannot fake a preview into a real run. Dropped.
- **MCP ops containing `__`.** `mcp__srv__delete__x` parses to op `x` and allows, because `${TOOL_NAME##*__}` splits on the last separator. No server in the roster emits it. Dropped.
- **Supabase `execute_sql` / `apply_migration` allowed.** Outside the CLAUDE.md list and named in the hook's own comment as captured spillover, not hidden. Not a finding.
- **The namespace checker.** Reads executable lines only, refuses on an empty parse, and the verdict comes from repo state rather than the developer's box. It checks exactly one namespace today (`mcp__plugin_vercel_vercel__`), which is narrow but honest.

## The process read

Round 3 said the rounds were calibrated and the **fix shape** was not. That still holds, and there is a sharper version of it now: the bullet my finding 1 falsifies is not a claim in the code, it is round 3's own *What is sound* entry. The sound-list is being written from inside the same mental model as the fix, which is the failure mode this repo already has a lesson for.

This is instance eleven of one class. Fork count, admitted-token set, window size, rescan count, the quote strip, per-stage versus per-invocation, word count, stage count — eight ceilings, each correct, each followed by the padding moving to the axis nobody bounded. Bytes is the eleventh. There will be a twelfth. A cheap bound exists today (refuse over ~1MB of command text, same fail-closed direction the git arm already uses) and it buys one more round; the durable fix is getting this predicate off a per-fork shell budget, which is already captured as its own row.

## VERDICT: REQUEST CHANGES

Not BLOCK. Nothing here is unrecoverable if it merges, and the PR's headline change (operation-keyed MCP denial) is a real repair that survived every attack I aimed at it.

Fix first: bound the command by **bytes** before any grep or `tr` touches it, and make the timing fixture generator vary token length as well as token count, so the suite can go red on the axis that is live.

FINDINGS:
major|Byte-length padding (one 5MB word, 8 words total, 1 stage separator) takes the hook past its wired 5s timeout, so the deny is discarded and `git push --force`, `git push origin +main`, `git clean -fd` and `rm -v -rf DIR` proceed|q-system/.q-system/hooks/destructive-op-deny.sh:611
minor|Every timing fixture pads word count or stage count and none varies a token's byte length, so the suite stays green on the axis that is live|q-system/.q-system/tests/test_argv_prefilter.py:142
minor|The 64-token rescan window is the only argv coverage behind a prefix carrying its own option, so `sudo -u root git push origin <63 refspecs> --force` is allowed, contradicting the comment that the first-position scan catches it at any length|q-system/.q-system/hooks/destructive-op-deny.sh:320
END FINDINGS

assafkip:
Round 4 minor 3 (the 64-token rescan window behind `sudo -u root`: a force flag after 63 refspecs is allowed): not fixed in this PR, for this reason. The guard on main has no argv parser at all. It matches substrings, so the same command is allowed there too, and this PR does not make it worse. Making the window unbounded is the O(n^2) cost that rounds 1 to 11 were fighting. The byte cap in 56ad7de6 bounds the input, so a fix is now an in-budget linear scan, and it is recorded in the ticket result as the next change on this guard. Round 4 major and minor 2 are fixed in 56ad7de6.

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-338-20260924-005212.md` (11685 bytes).

```
FINDINGS:
major|Byte padding with the fat token at a stage's first position crosses the wired 5s timeout at 104KB, 2.5x under the new 262144-byte cap, so the deny is discarded and git push --force / +refspec / :refspec / --mirror / git clean -fd proceed; the currently-installed hook denies the same payload in 0.1s, so this is a coverage regression|q-system/.q-system/hooks/destructive-op-deny.sh:142
major|install-claude-hooks.py refuses this PR's own hook because _env_vars_read scans comments and ${var//x/} on line 709 reads as a new environment variable, so kipi update never installs it and verify.sh --full is red on every run|q-system/.q-system/scripts/install-claude-hooks.py:313
minor|Every installer success fixture installs into an empty --home, so the env-var differential branch guarded by `installed_text is not None` never runs and 24 tests pass against a source the production path refuses|q-system/.q-system/tests/test_install_claude_hooks.py:66
minor|The new byte-cap fixture uses a 5MB word behind a leading `echo`, so it exercises the cap's O(1) refusal at a cheap token position and cannot go red on the in-budget size range that is live|q-system/.q-system/tests/test_argv_prefilter.py:168
minor|DIFFERENTIAL_CORPUS rows are all under 60 bytes, so the never-weaker gate structurally cannot see a candidate that loses a deny to the timeout under padding|q-system/.q-system/scripts/install-claude-hooks.py:136
END FINDINGS
```

--- reviewer output ---

## Findings

### 1. major — byte padding still crosses the wired 5s timeout at 104KB, 2.5x under the new 262144-byte cap, and this PR makes force-push *worse* than the hook running today

`q-system/.q-system/hooks/destructive-op-deny.sh:142`

The cap added in `56ad7de6` says "256KB is ~0.6s worst case". The cost is quadratic in the **first token's** byte length, not linear in total bytes. When the fat token sits at position 0 of a stage, `argv_deny_reason`'s `local prog="${w[0]##*/}"` pays O(n²) on bash 3.2.

Reproducer (driver posts the payload to the hook, reports wall time; `WIRED=5.0` is the author's own `WIRED_HOOK_TIMEOUT_S`):

```
python3 -c '
import sys; sys.path.insert(0,"<tmp>")
import drive
T = "git" + " push origin +main"
for kb in (64,104,160):
    pad = "x"*(kb*1024)
    a = "echo "+pad+" ; "+T     # the suite fixture shape, test_argv_prefilter.py:173
    b = pad+" ; "+T             # same bytes, fat word at position 0
    print(kb, drive.run("Bash", dict(command=a)), drive.run("Bash", dict(command=b), timeout=90))'
```

Real output:

```
 64KB  fixture-shape(echo first): deny  0.11s   fat-word-first: deny   1.97s -> prod=deny
104KB  fixture-shape(echo first): deny  0.14s   fat-word-first: deny   5.06s -> prod=ALLOW-deny-discarded
160KB  fixture-shape(echo first): deny  0.22s   fat-word-first: deny  12.50s -> prod=ALLOW-deny-discarded
```

**The new consequence that nobody has seen yet**, and the reason this holds at major rather than dropping: the hook currently running on this machine denies the identical payload instantly, because it still carries the coarse `git push ... --force` substring that this PR removes at `:198-212`.

```
installed hook exists: True lines: 501
pad 120 KB PR-source -> deny  6.54 s | production: ALLOW-discarded
pad 120 KB installed -> deny  0.07 s | production: deny
pad 250 KB PR-source -> deny 32.91 s | production: ALLOW-discarded
pad 250 KB installed -> deny  0.10 s | production: deny
```

Round 4's answer to its own window minor was "the guard on main has no argv parser, so the same command is allowed there too, and this PR does not make it worse." For the **byte** axis that is falsified: the installed substring pattern catches it at any length, and removing the pattern trades a fast coarse deny for a slow exact one that production discards. Affects `--force`, `--force-with-lease`, `-f`, `+refspec`, `:refspec`, `--delete`, `--mirror` and `git clean -fd`, which are now argv-only. `rm -rf` and `git reset --hard` still short-circuit in `BASH_DENY` and stay fast, which is why routine probing does not see this.

Same class, same severity as rounds 1, 3 and 4. Not escalated. I constructed the padding.

### 2. major — the installer REFUSES this PR's own hook, because `_env_vars_read` reads comments, so nothing in this PR ever reaches the running machine

`q-system/.q-system/scripts/install-claude-hooks.py:313`

```
$ python3 q-system/.q-system/scripts/install-claude-hooks.py --dry-run
  REFUSED destructive-op-deny.sh: the source reads environment variable(s) the installed hook does not (var). A probe cannot open a door whose key it does not hold, so a new env read is refused rather than tested.
exit=1
```

Cause, located:

```
$ python3 -c '... print(sorted(m._env_vars_read(src) - m._env_vars_read(inst)))'
new vars: ['var']

$ grep -n 'var//' q-system/.q-system/hooks/destructive-op-deny.sh
709:    # This was three `${var//x/}` substitutions. On bash 3.2, which is what
713:    #   tokens   ${var//}    tr -d
```

Both hits are inside `#` comments. `_ENV_READ` is applied to the raw file text with no comment stripping, unlike `namespaces_in_hook` in the paired namespace checker, which does strip them (`mcp-denylist-namespace-check.py:223`).

This is the round-3 minor's fix biting back. Removing the `isupper() or "_"` filter (the comment at `:308-312` records that change) made every `$name` in the file count, and the hook's own round-five scar note about `${var//}` now reads as a new environment read.

At 3am: `kipi update` runs `install_vendored_hooks` (`kipi-update.sh:318`), prints REFUSED plus the WARNING, and the machine keeps the stale 501-line hook. `verify.sh --full` runs `install-claude-hooks.py --check`, which is already red:

```
$ python3 q-system/.q-system/scripts/install-claude-hooks.py --check
  destructive-op-deny.sh: INSTALLED COPY DIFFERS from the repo
check exit=1
```

So merging delivers zero of the PR's headline value, and it arms a warning that fires on every `kipi update` and every `--full` verify forever, which is how the operator learns to skip it. Recoverable by editing a comment, hence major and not BLOCK.

### 3. minor — every installer success fixture starts from an empty HOME, so the branch that fails in production never executes

`q-system/.q-system/tests/test_install_claude_hooks.py:66`

`refuse_if_weaker` guards the env-var differential with `if installed_text is not None` (`:472`). Every green fixture installs into a fresh `--home` where no copy exists, so `installed_text is None` and that branch is skipped.

```
$ python3 -m pytest q-system/.q-system/tests/test_install_claude_hooks.py -q
........................                                                 [100%]
24 passed in 27.70s
```

24 green against the same source the production path refuses. The missing fixture is one that seeds `--home` with a *different* real hook (the 501-line installed copy is right there) and then installs.

### 4. minor — the new byte-cap fixture tests the cap's O(1) refusal, not the cost inside it

`q-system/.q-system/tests/test_argv_prefilter.py:168`

`test_one_long_word_cannot_push_the_deny_past_the_timeout` uses `"echo " + word` with a 5MB word. Two reasons it cannot go red for the reason it exists: 5MB is over `_MAX_COMMAND_BYTES`, so the O(1) byte refusal answers before any scan runs; and the leading `echo` puts the fat token at position 1, which `_argv_could_deny_here` rejects for free. My run above shows the fixture shape at 0.11 to 0.22s while the same bytes at position 0 cross 5s.

The axis that needs a case: bytes held **under** the cap, fat token at a stage's first position.

### 5. minor — `DIFFERENTIAL_CORPUS` holds only short commands, so the never-weaker gate cannot see finding 1

`q-system/.q-system/scripts/install-claude-hooks.py:136`

Round 2's fix to `decision_of` is real and I could not falsify it: the probe now runs at `WIRED_HOOK_TIMEOUT_S = 5` and reports `timeout` as its own outcome, which every must-deny comparison refuses. The remaining gap is the corpus. All 30 rows are under 60 bytes, so the one comparison that would have caught finding 1 ("installed denies this, source does not") is never offered a padded payload. One row, `"x"*160*1024 + " ; git push origin +main"`, converts finding 1 from a reviewer's discovery into an install-time refusal.

## What is sound

- **Operation-keying holds on real data.** Swept 115 tool names copied from this session's live roster. 18 deny, every one destructive or on the founder's explicit list: Linear `delete_attachment` / `delete_comment` / `delete_diff_comment` / `delete_status_update`, Gmail `delete_draft` / `delete_label` / `trash_message` / `trash_thread`, Drive `trash_file`, Calendar `delete_event`, Docs `delete`, Supabase `delete_branch` and `reset_branch`, Notion `notion-move-pages`, notebooklm `remove_notebook`. No vendor segment guessed anywhere.
- **No read over-block.** 97 allow, including `list_issues`, `get_issue`, `save_issue`, `merge_diff`, `resolve_diff_thread`, `restore_issue_label`, `restore_project_label`, `retire_issue_label`, `list_tables`, `notion-search`, `browser_close`, `browser_drop`, `reset_session`, `close_session`, `query_logs`. The two deliberate `unlabel_*` denies and the Vercel arm are the only exceptions, both stated in the code.
- **The `un` guard and the camelCase fold both hold in both directions.** `untrash_message`, `untrash_thread`, `unmark_message_spam` allow; `batchDelete`-shaped names deny.
- **`BASH_DENY` short-circuit is fast under padding.** `rm -rf`, `rm -v -rf`, `git reset --hard`, `git branch -D`, `git clean -fd`, `find -delete`, `mkfs`, `dd of=/dev/`, `shred`, `chmod -R 777 /` all deny in 0.23 to 0.27s behind 200KB.
- **The word ceiling and the stage cap each hold on their own axis.** Padding after a leading `echo` stayed at 0.11 to 0.25s at 64, 104, 160 and 200KB.
- **`installed_is_an_oracle` accepts the real 501-line installed hook**, so the never-weaker comparison is live rather than skipped.
- **The namespace checker** passes, reads executable lines only, and its verdict comes from repo state rather than the local box.
- **Dropped for want of a repro:** an MCP operation containing `__` that would mis-split under `${TOOL_NAME##*__}` (none in the roster); a read-only roster op caught by the verb list (none); audit-log interleaving under two concurrent hooks (I could not make the lines corrupt).

## The process read

Rounds 1 through 4 were calibrated. Every repro was real and every fix landed. Two things are new and worth more than another nit.

First, this is instance twelve of one class, and the file itself named the problem at `:384` ("Seventh instance of one bypass in one predicate. That is a signal about the SURFACE"). Twelve ceilings, each correct, each followed by the padding moving to the axis nobody bounded: forks, admitted tokens, window, rescan count, quote strip, per-stage versus per-invocation, words, stages, bytes, and now token *position*.

Second, and this is the sharper signal: finding 2 is a defect **created by** the fix for a round-3 minor. The loop has stopped converging. The lever that ends it is not a thirteenth ceiling in the shell; it is finding 5, one padded row in `DIFFERENTIAL_CORPUS`, so the next axis is refused at install time by a gate instead of discovered by a reviewer at round six.

## VERDICT: REQUEST CHANGES

**Fix first: finding 2.** Strip comments in `_env_vars_read` the way `namespaces_in_hook` already does. Until that lands, nothing in this PR reaches the running hook, `verify.sh --full` is red on every run, and finding 1 is unreachable in production because the fix it belongs to never installs.

FINDINGS:
major|Byte padding with the fat token at a stage's first position crosses the wired 5s timeout at 104KB, 2.5x under the new 262144-byte cap, so the deny is discarded and git push --force / +refspec / :refspec / --mirror / git clean -fd proceed; the currently-installed hook denies the same payload in 0.1s, so this is a coverage regression|q-system/.q-system/hooks/destructive-op-deny.sh:142
major|install-claude-hooks.py refuses this PR's own hook because _env_vars_read scans comments and ${var//x/} on line 709 reads as a new environment variable, so kipi update never installs it and verify.sh --full is red on every run|q-system/.q-system/scripts/install-claude-hooks.py:313
minor|Every installer success fixture installs into an empty --home, so the env-var differential branch guarded by `installed_text is not None` never runs and 24 tests pass against a source the production path refuses|q-system/.q-system/tests/test_install_claude_hooks.py:66
minor|The new byte-cap fixture uses a 5MB word behind a leading `echo`, so it exercises the cap's O(1) refusal at a cheap token position and cannot go red on the in-budget size range that is live|q-system/.q-system/tests/test_argv_prefilter.py:168
minor|DIFFERENTIAL_CORPUS rows are all under 60 bytes, so the never-weaker gate structurally cannot see a candidate that loses a deny to the timeout under padding|q-system/.q-system/scripts/install-claude-hooks.py:136
END FINDINGS

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-338-20260924-023302.md` (7850 bytes).

```
FINDINGS:
minor|The byte cap's stated cost model is linear and ~2ms/KB but the measured cost is superlinear at ~42ms/KB, so the real refusal threshold for a single-token command is ~50KB not the declared 256KB, and benign 60KB payloads are denied|q-system/.q-system/hooks/destructive-op-deny.sh:135
minor|The byte cap is checked after four unbounded jq parses of the same stdin, so a 20MB tool_input.command takes 5.26s against the wired 5s timeout and the deny is discarded (constructed input, no producer emits it)|q-system/.q-system/hooks/destructive-op-deny.sh:142
minor|Linear spells label deletion `retire`, so mcp__linear__retire_issue_label and its two siblings are allowed, falsifying the operation-keying premise on the server that motivated the PR|q-system/.q-system/hooks/destructive-op-deny.sh:958
minor|The PR body claims reads are no longer denied on any server, but mcp__plugin_vercel_vercel__list_deployments still denies|q-system/.q-system/hooks/destructive-op-deny.sh:978
END FINDINGS
```

--- reviewer output ---

Review complete. Read-only; repo tree is clean (`git status --porcelain` empty), all scratch lives in `$TMPDIR/r6`.

## What I did

Drove the PR's hook directly with PreToolUse payloads, ran the five changed test files (83 passed, 42 subtests), mutation-tested the two new mechanisms, ran the installer against the real `~/.claude` tree, and swept the live MCP roster. Then read rounds 1 to 5.

---

## Findings

### minor 1. The byte cap's cost model is wrong by ~18x, so its real threshold is ~50KB not 256KB
`q-system/.q-system/hooks/destructive-op-deny.sh:135`

The comment sizing `_MAX_COMMAND_BYTES=262144` says the hook's cost is "linear in bytes (measured ~2ms/KB worst case)" and "256KB is ~0.6s worst case". On the single-fat-token shape it is superlinear, and the 3s watchdog fires at roughly one fifth of the declared cap. Consequence: an ordinary `echo '<60KB blob>' > file` is now refused.

Repro and real output:

```
python3 $TMPDIR/r6/t4.py      # bash("echo '" + "A"*(kb*1024) + "' > /tmp/payload.json")
ALLOW      0.59s rc=0  benign single quoted blob 20KB
ALLOW      1.70s rc=0  benign single quoted blob 40KB
DENY       3.31s rc=0  benign single quoted blob 60KB
DENY       3.36s rc=0  benign single quoted blob 100KB
DENY       3.34s rc=0  benign single quoted blob 200KB
```

0.59s at 20KB to 1.70s at 40KB is 2.9x for 2x the bytes, and 40KB alone is ~42ms/KB, not 2. The direction is fail-closed and the message names the fix, so this does not gate. It is the sizing rationale the next maintainer will trust, which is why it is worth correcting. The installed hook allows the identical 60KB payload after 4.17s and 200KB after 42.78s, so the PR is strictly safer here.

### minor 2. The byte cap sits behind four unbounded `jq` parses of the same stdin
`q-system/.q-system/hooks/destructive-op-deny.sh:142`

Lines 32 to 35 parse `$INPUT` four times before `${#COMMAND}` is ever compared to the cap, so the cap's O(1) refusal is preceded by O(n) work it does not bound. Crossing measured against the wired 5s timeout:

```
python3 $TMPDIR/r6/t8.py
DENY       4.19s rc=0  command 16MB, forced push in stage 2
DENY       5.26s rc=0  command 20MB, forced push in stage 2
DENY       6.82s rc=0  command 24MB, forced push in stage 2
```

At 20MB the hook exceeds the wired 5 (confirmed `matcher='*' timeout=5` in `~/.claude/settings.json`), so production kills it and discards the deny. This is the same class as the five prior rounds and a new axis. Dropping it to minor on the calibration rule: no producer in this fleet emits a 20MB `tool_input`, so the input is constructed. The comment at line 137 claiming a byte cap "closes the class rather than the next axis" is the part that is falsified.

### minor 3. Linear spells its label deletion `retire`, which the operation-keyed verb list misses
`q-system/.q-system/hooks/destructive-op-deny.sh:958`

The PR's load-bearing premise is "every server that deletes spells it `delete`". The counterexample is on the server that motivated the PR.

```
python3 $TMPDIR/r6/t6.py
DENY       0.04s rc=0  mcp__linear__delete_comment
ALLOW      0.03s rc=0  mcp__linear__retire_issue_label
ALLOW      0.03s rc=0  mcp__linear__restore_issue_label
ALLOW      0.03s rc=0  mcp__linear__unshare_issue
```

`restore_issue_label` existing as the inverse is what identifies `retire` as the destructive half. CLAUDE.md calls Linear `*delete*` NON-NEGOTIABLE. Recoverable by `restore_*_label`, blast radius one label, so it stays minor. Same shape on `mcp__supabase__merge_branch` and `rebase_branch`.

### minor 4. The PR body's "reads are no longer denied on any server" is still false
PR description, "This closes the MCP over-block major on #274 as a side effect"

Raised in round 5 as a minor. My own repro shows it live at this HEAD:

```
DENY       0.04s rc=0  mcp__plugin_vercel_vercel__list_deployments
ALLOW      0.03s rc=0  mcp__plugin_vercel_vercel__authenticate
```

Severity held flat at minor. The hook's own comment at line 972 is honest about this; the PR body is not.

---

## What is sound

**Round 5's 104KB fat-token major is dead.** `fat 104KB at stage-head ; forced push` now returns DENY in 3.35s, and 200KB and 255KB the same. The watchdog converts what was a timeout-allow into an explicit refusal.

**The 3s watchdog is load-bearing, not decoration.** Mutated `[ "$_chk_i" -lt 30 ]` to 600 and `test_a_fat_first_token_under_the_byte_cap_still_denies_in_time` went red with `5.79860782623291 not less than 5.0`. The test can fail for the reason it claims.

**The capability-token scope test kills its mutant.** Changed `jq -cS` to `jq -c` and `test_key_order_does_not_change_the_scope` went red on `{"a":1,"b":2}` vs `{"b":2,"a":1}`. That was the round-1 BLOCKER and the fix is real.

**Round 5's installer major is fixed.** `install-claude-hooks.py --dry-run` against the live `~/.claude` returns `WOULD INSTALL (differs)`, exit 0. The differential oracle no longer refuses this PR's own hook.

**Word-ceiling padding does not reach the fleet deny.** 1500 assignment words and 5000 plain words ahead of the fleet-update command both DENY in 0.15s.

**Benign multi-word commands are unaffected.** 5KB through 100KB of space-separated tokens all ALLOW in under 0.27s.

**The stage-separator, 5MB-word and sudo-padding axes from rounds 2 to 4 are all covered by the watchdog**, and the MCP half behaves as designed across 38 live roster entries: `list_issues`, `get_issue`, `untrash_message`, `browser_drop` and `reset_session` all allow; `delete_*`, `trash_*`, `unlabel_*`, `reset_branch` and `notion-move-pages` all deny.

---

Rounds 1 through 5 each produced a major, and all five were one class: pad a payload past the 5s hook timeout so the deny is discarded. The axis moved each time (tokens, sudo prefixes, stage separators, byte length, fat-token position) and each round's fix bounded the axis that had just been used. Five rounds of the same class means the finding was correct and the fix shape was wrong from round 2 onward, which is the repo's own `same-finding-twice-go-structural` lesson.

The 3s watchdog in the head commits is the structural answer and it works. Two of my four minors are only visible because that landed. Worth recording so the next reviewer of a padding finding reaches for "bound time" at round 2 rather than round 12.

---

## VERDICT: APPROVE WITH NITS

Only minor findings survived reproduction. Merging this is a net safety improvement over the installed hook on every payload I measured.

**Fix first:** the cost model at line 135. Replace "linear in bytes, ~2ms/KB, 256KB is ~0.6s worst case" with the measured quadratic and the real ~50KB refusal threshold, so `_MAX_COMMAND_BYTES=262144` stops reading as the governing bound for a shape it never governs.

FINDINGS:
minor|The byte cap's stated cost model is linear and ~2ms/KB but the measured cost is superlinear at ~42ms/KB, so the real refusal threshold for a single-token command is ~50KB not the declared 256KB, and benign 60KB payloads are denied|q-system/.q-system/hooks/destructive-op-deny.sh:135
minor|The byte cap is checked after four unbounded jq parses of the same stdin, so a 20MB tool_input.command takes 5.26s against the wired 5s timeout and the deny is discarded (constructed input, no producer emits it)|q-system/.q-system/hooks/destructive-op-deny.sh:142
minor|Linear spells label deletion `retire`, so mcp__linear__retire_issue_label and its two siblings are allowed, falsifying the operation-keying premise on the server that motivated the PR|q-system/.q-system/hooks/destructive-op-deny.sh:958
minor|The PR body claims reads are no longer denied on any server, but mcp__plugin_vercel_vercel__list_deployments still denies|q-system/.q-system/hooks/destructive-op-deny.sh:978
END FINDINGS

assafkip:
Nits on c9c3506d are handled in ONE follow-up, #443. Fixed there: Linear retire_* label ops are denied (nit 3), and the byte-cap cost comment is corrected (nit 1). Declined with reasons in #443's body: nit 2 (20MB stdin into jq is constructed input with no producer) and nit 4 (Vercel list over-block, already declined in round 1).

PR #339, opened 2026-09-12
feat(voice-gate): the fanout-loss check runs where it fails loudly, not only where the clones are (ASK-1304)

assafkip:
Closes ASK-1304 (promoted from spillover `sp-0b8bec0d`, source ASK-1197).

## The hazard

The fleet updater rsyncs the skeleton's `q-system/` over every instance. So an
instance copy of `voice-stop-gate.py` carrying behaviour the skeleton never had is
DESTROYED by the next sync. That is not hypothetical: it happened twice in one
afternoon on 2026-09-06 (`sp-745f5962`, `sp-1ad08728`). The instance's pre-commit
went red, the restore was undone by the following sync.

The only check for it lived in an assertion that skips wherever the 25 instance
clones are absent, which is everywhere except the founder's machine. CI held the
scanner against synthetic trees and never the fleet, so the hazard was guarded by
a test that never ran against a real one.

## What this adds

`q-system/.q-system/scripts/voice-gate-propagation-check.py` resolves the
registry, compares each instance's gate file against this checkout's, and
classifies **synced / behind / ahead / COULD NOT READ**.

The AHEAD-vs-BEHIND split is the design, not a refinement. A digest mismatch alone
is not loss: an instance holding an OLDER skeleton version is fixed by the next
sync, and that is the normal state of a fleet between syncs. Calling every mismatch
red would make this red on a healthy fleet, which is how a gate gets switched off.
`_skeleton_ever_had` is the same git-based answer the lessons-drift reporter uses.

Exit 2 on any ahead instance, on a root that is not the registry's skeleton, and on
a launchd run whose alert did not file. COULD NOT READ is never silently green.

`com.kipi.voice-gate-propagation.plist` is the trigger, daily 06:30, ahead of the
06:45 drift report and the 07:40 brief.

## Alerting

Red files ONE ticket into Sana's Linear triage through `slack-notify.sh`, on state
CHANGE only and only under `KIPI_TRIGGER=launchd`, which only the plist sets. The
state is the SET of ahead instances: a fingerprint carrying the timestamp would
make every run a state change and mute the channel. A failed filing does not
advance the state, so the next run retries. Run by hand it prints and files
nothing, so removing the plist provably stops delivery.

Each run leaves `q-system/output/voice-gate-propagation.json`, a LOCAL artifact of
the last scan. It is gitignored and written only where the clones are, so nothing
in continuous integration reads it and this PR no longer says otherwise (review
round 3, minor). The ALERT is the signal that leaves the machine.

## Where it may act

It sits under the fanned-out scripts directory, so all 25 instances carry it. It
refuses unless the registry's `skeleton` entry is its own root (the lesson
`a-fan-out-installer-must-refuse-where-its-job-cannot-run`). The plist carries
`kipi-scope: skeleton-only` for `install-plist.sh --all`.

## Against the DoR

- [x] The fanout-loss assertion has a runner that executes it against the real
      fleet (the scheduled local job), and the run produces a machine-readable
      manifest of the scan (local, gitignored; see above)
- [x] The runner alerts through `slack-notify.sh` on red, once, on state change
- [x] A test proves it goes RED on a synthetic instance-ahead tree and GREEN on a
      synced one, both executed
- [x] The CI skip stays, says `CI-ONLY SKIP` in the file, and the skip count is
      asserted so a second skip cannot hide inside it

## Two things worth a reviewer's eye

**The skip counter counted itself.** Its first cut spelled `pytest.skip(`
literally and matched its own pattern string as a third site. The needle is now
assembled from pieces, with the reason inline: a counter that matches itself
measures the counter, not the suite.

**The docstring cites the lessons-drift reporter without its `.py`, on purpose.**
That suite's exact-set single-caller grep counted this file as a third caller of
it. The grep is correct to be that blunt, and a docstring citing prior art is not
a caller, so the fix is to stop tripping it rather than to loosen it (the call
`engineering_route.py` already records against the morning-brief notify grep). The
file says so inline so the extension does not get helpfully restored.

## Verification

- `bash q-system/.q-system/verify.sh --staged` -> ok (4 checks), which is the
  same script CI runs
- `python3 -m pytest q-system/.q-system/tests/test_voice_gate_propagation.py
  q-system/.q-system/tests/test_lessons_drift_report.py -q` -> 31 passed,
  1 skipped (the documented CI-only skip)
- **Mutation:** flipping the `ahead` branch to return `behind` killed 3 tests, so
  the check can actually fail rather than decorate
- RED was observed first: 12 failed before the runner existed

## Spillover captured, not fixed here

`sp-dc4cdb89` — `test_notion_board.py::test_no_write_lands_after_the_timeout_was_reported`
is load-flaky. Its budget is 0.05s of wall clock, so under the full 1410-test suite
it is spent before the assertion and the test reds, while the file passes 26/26 in
isolation. It blocked this unrelated pre-commit twice. Pre-existing, out of scope.

### DIFF
--- a/README.md
+++ b/README.md
@@ -332,7 +332,7 @@ The pages above are the knowledge half. The rest of the repo, with the command t
 - **21 namespaced slash commands**, listed above. Six in `kipi-core`, six in `kipi-dsse`, nine in `prd-os`.
 - **65 hook entries** in `settings-template.json` (`grep -o '"command": "[^"]*"' settings-template.json | wc -l`), plus seven more in the plugins' own `hooks.json`. Every copy gets the same switches, because they ship as one file.
 - **Six plugins** under `plugins/`, all six listed in `.claude-plugin/marketplace.json`.
-- **16 launchd jobs** committed as plists. `./kipi install-jobs` enumerates them with `git ls-files`, so a plist outside the scripts directory is still reached; run `bash q-system/.q-system/scripts/install-plist.sh` with no arguments to see the label list it will install.
+- **17 launchd jobs** committed as plists (`git ls-files '*.plist' | wc -l`). `./kipi install-jobs` enumerates them with `git ls-files`, so a plist outside the scripts directory is still reached; run `bash q-system/.q-system/scripts/install-plist.sh` with no arguments to see the label list it will install.
 
 ---
 
--- a/q-system/.q-system/capability/expected_tests/q-system__.q-system__tests__test_voice_gate_propagation.py.json
+++ b/q-system/.q-system/capability/expected_tests/q-system__.q-system__tests__test_voice_gate_propagation.py.json
@@ -1,0 +1,5 @@
+{
+  "path": "q-system/.q-system/tests/test_voice_gate_propagation.py",
+  "runner": "python3",
+  "timeout_s": 120
+}
--- a/q-system/.q-system/scripts/com.kipi.voice-gate-propagation.plist
+++ b/q-system/.q-system/scripts/com.kipi.voice-gate-propagation.plist
@@ -1,0 +1,50 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
+<!--
+  com.kipi.voice-gate-propagation -- the registered trigger for the
+  voice-stop-gate fanout-LOSS check (ASK-1304, promoted from sp-0b8bec0d).
+
+  Why a scheduled LOCAL job and not a CI check: the hazard is a property of the
+  25 instance clones, which exist on this machine and nowhere in CI. The suite
+  in q-system/.q-system/tests/test_voice_gate_propagation.py holds the scanner
+  against synthetic trees and skips the live-fleet assertion wherever the clones
+  are absent. This plist is what makes that assertion actually run.
+
+  kipi-scope: skeleton-only
+  (the check compares instances against the skeleton it runs in; the registry's
+  skeleton entry must be this checkout or it exits 2 saying COULD NOT READ.)
+
+  TEMPLATE. Install with:  bash q-system/.q-system/scripts/install-plist.sh com.kipi.voice-gate-propagation
+  install-plist.sh renders __KIPI_REPO__, __HOME__ and __USER__ and bootstraps
+  the job; a literal path here would fan out to 25 instances with one machine's
+  home in it.
+
+  Daily 06:30 local, ahead of the 06:45 lessons-drift report and the 07:40
+  brief. KIPI_TRIGGER=launchd is the marker the alert files under; run by hand it
+  prints and files nothing, so removing this plist provably stops delivery. Red
+  files ONE ticket into Sana's Linear triage via slack-notify.sh, on state
+  change only -- never the founder (founder-notifications.md).
+-->
+<plist version="1.0">
+<dict>
+  <key>Label</key><string>com.kipi.voice-gate-propagation</string>
+  <key>EnvironmentVariables</key>
+  <dict>
+    <key>HOME</key><string>__HOME__</string>
+    <key>USER</key><string>__USER__</string>
+    <key>LOGNAME</key><string>__USER__</string>
+    <key>PATH</key><string>__HOME__/.local/bin:/opt/homebrew/bin:/usr/local/bin:/usr/bin:/bin:/usr/sbin:/sbin</string>
+    <key>KIPI_TRIGGER</key><string>launchd</string>
+  </dict>
+  <key>ProgramArguments</key>
+  <array>
+    <string>/usr/bin/env</string>
+    <string>python3</string>
+    <string>__KIPI_REPO__/q-system/.q-system/scripts/voice-gate-propagation-check.py</string>
+  </array>
+  <key>StartCalendarInterval</key><dict><key>Hour</key><integer>6</integer><key>Minute</key><integer>30</integer></dict>
+  <key>RunAtLoad</key><false/>
+  <key>StandardOutPath</key><string>__HOME__/.config/kipi/logs/voice-gate-propagation.out.log</string>
+  <key>StandardErrorPath</key><string>__HOME__/.config/kipi/logs/voice-gate-propagation.err.log</string>
+</dict>
+</plist>
--- a/q-system/.q-system/scripts/voice-gate-propagation-check.py
+++ b/q-system/.q-system/scripts/voice-gate-propagation-check.py
@@ -62,6 +62,7 @@ MANIFEST_REL = "q-system/output/voice-gate-propagation.json"
 STATE_REL = "q-system/output/.voice-gate-propagation-state.json"
 NOTIFY = os.path.join(HERE, "slack-notify.sh")
 COULD_NOT_READ = "COULD NOT READ"
+NO_INSTANCES = "no instances in the registry"
 NOTIFY_TIMEOUT_S = 20
 
 
@@ -140,25 +141,53 @@ def scan(root: str) -> dict:
             state, note = classify(path, root)
             rows.append({"name": name, "path": path, "state": state, "note": note})
     ahead = [r["name"] for r in rows if r["state"] == "ahead"]
+    unanswered = [r["name"] for r in rows if r["state"] == "could-not-read"]
+    # A registry this run could PARSE but that names no instance scanned nothing,
+    # and nothing has the same bytes as a healthy fleet
+    # (lessons/zero-selected-items-is-a-failure-not-a-pass.md). Zero and
+    # never-ran must not both print green.
+    if skeleton_ok and not rows:
+        unanswered = [NO_INSTANCES]
     return {
         "generated_at": datetime.now(timezone.utc).isoformat(),
         "skeleton": root,
         "skeleton_ok": skeleton_ok,
         "instances": rows,
         "ahead": ahead,
-        # a root that is not the skeleton is RED: it means this ran somewhere it
-        # cannot answer the question, which is not the same as "no drift found".
-        "red": bool(ahead) or not skeleton_ok,
+        "unanswered": unanswered,
+        # THREE ways to be red, and the middle one is the PR #339 review's major:
+        # an instance whose gate file could not be read is UNKNOWN, never synced.
+        # An instance that is genuinely ahead AND unreadable at 06:30 used to
+        # score exit 0 with fingerprint `green`, so the branch this file's own
+        # docstring promised was never silently green was exactly the one that
+        # was. A root that is not the skeleton is the same defect one level up:
+        # it ran somewhere it cannot answer the question, which is not the same
+        # answer as "no drift found".
+        "red": bool(ahead) or bool(unanswered) or not skeleton_ok,
     }
 
 
 def render(report: dict) -> str:
+    """The summary leads, before the per-instance rows.
+
+    `alert-to-linear.py::title_for` flattens this to one line and truncates it,
+    so a RED line printed after 25 `name: synced` rows gave Sana a ticket titled
+    "X: synced Y: synced..." and never named the instance that is losing
+    behaviour (PR #339 review, minor). The NAMES come first inside the summary
+    for the same reason: truncation eats the tail.
+    """
     day = report["generated_at"][:10]
     lines = [f"voice-stop-gate propagation ({day})"]
     if not report["skeleton_ok"]:
         lines.append(f"skeleton: {COULD_NOT_READ} (the registry's skeleton entry is not "
                      f"this checkout: {report['skeleton']})")
         return "\n".join(lines)
+    if report["ahead"]:
+        lines.append(f"RED AHEAD: {', '.join(report['ahead'])} -- "
+                     "lose behaviour on the next sync")
+    if report["unanswered"]:
+        lines.append(f"RED UNANSWERED: {', '.join(report['unanswered'])} -- "
+                     "not scanned, so not known to be safe")
     for row in report["instances"]:
         if row["state"] == "could-not-read":
             lines.append(f"{row['name']}: {COULD_NOT_READ} ({row['note']})")
@@ -167,11 +196,6 @@ def render(report: dict) -> str:
         else:
             lines.append(f"{row['name']}: {row['state']}"
                          + (f" ({row['note']})" if row["note"] else ""))
-    if not report["instances"]:
-        lines.append("no instances in the registry")
-    if report["ahead"]:
-        lines.append(f"RED: {len(report['ahead'])} instance(s) lose behaviour on the "
-                     f"next sync: {', '.join(report['ahead'])}")
     return "\n".join(lines)
 
 
@@ -182,7 +206,15 @@ def fingerprint(report: dict) -> str:
     shape of a fix."""
     if not report["skeleton_ok"]:
         return "skeleton-unreadable"
-    return "ahead:" + ",".join(sorted(report["ahead"])) if report["ahead"] else "green"
+    parts = []
+    if report["ahead"]:
+        parts.append("ahead:" + ",".join(sorted(report["ahead"])))
+    # the unanswered set belongs to the STATE, not only to the exit code: left
+    # out, a newly-unreadable instance either never alerts or alerts on every
+    # run, and a channel repeating an unchanged condition stops being read.
+    if report["unanswered"]:
+        parts.append("unanswered:" + ",".join(sorted(report["unanswered"])))
+    return "|".join(parts) if parts else "green"
 
 
 def _send(message: str) -> None:
@@ -198,6 +230,16 @@ def _send(message: str) -> None:
                            f"{(done.stderr or b'').decode('utf-8', 'replace')[:300]}")
 
 
+def _remember(root: str, current: str, report: dict) -> None:
+    """The single writer of the state file. Advancing the state is what makes the
+    next run quiet, so it happens on exactly two paths -- after a filing, and on
+    the first quiet run -- and never after a failed filing."""
+    state_path = os.path.join(os.path.realpath(root), STATE_REL)
+    os.makedirs(os.path.dirname(state_path), exist_ok=True)
+    with open(state_path, "w", encoding="utf-8") as fh:
+        json.dump({"fingerprint": current, "at": report["generated_at"]}, fh)
+
+
 def run(root: str, notify=None, trigger=None, write_manifest=True) -> dict:
     report = scan(root)
     message = render(report)
@@ -208,7 +250,6 @@ def run(root: str, notify=None, trigger=None, write_manifest=True) -> dict:
         os.makedirs(os.path.dirname(manifest), exist_ok=True)
         with open(manifest, "w", encoding="utf-8") as fh:
             json.dump(report, fh, indent=2, sort_keys=True)
-        fh = None
     if trigger != "launchd":
         out["alert"]["reason"] = ("not launched by the plist (KIPI_TRIGGER != launchd): "
                                  "printed, filed nothing")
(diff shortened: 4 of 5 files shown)


assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-339-20260911-223424.md` (7379 bytes).

```
FINDINGS:
major|An instance whose gate file cannot be read is excluded from `red`, so a genuinely-ahead-but-unreadable instance (and a zero-row registry) exits 0 with fingerprint "green" and files no ticket|q-system/.q-system/scripts/voice-gate-propagation-check.py:151
minor|render() puts the RED line after all 25 per-instance rows, so the Linear ticket title truncates to "X: synced Y: synced..." and never names the ahead instance|q-system/.q-system/scripts/voice-gate-propagation-check.py:173
minor|The first launchd run on a never-red fleet treats ""->"green" as a state change and files a "recovered" ticket for a condition that never fired|q-system/.q-system/scripts/voice-gate-propagation-check.py:221
minor|README says 16 committed launchd plists; this PR makes it 17 and the line was not updated|README.md:335
nit|Dead assignment `fh = None` after the manifest write block|q-system/.q-system/scripts/voice-gate-propagation-check.py:211
END FINDINGS
```

--- reviewer output ---

# Review: PR #339, voice-gate-propagation-check

One major survived reproduction. Four smaller ones. Nothing permanent or unrecoverable.

---

## MAJOR — an instance the check could not read is scored green

`q-system/.q-system/scripts/voice-gate-propagation-check.py:151`

`red` is `bool(ahead) or not skeleton_ok`. A `could-not-read` row contributes to neither. So an instance that is genuinely AHEAD, but whose gate file is unreadable at 06:30 for any reason (permissions, a renamed directory, a path the registry no longer matches, a mid-sync moment), is reported green: exit 0, `red: false` in the manifest, fingerprint `green`, no ticket into Sana's queue. The PR body says "COULD NOT READ is never silently green." The exit code and the alert path disagree.

This is not a constructed input. Against the live registry in this checkout, `reddit-build-radar` already has no readable `voice-stop-gate.py` today, so the branch is live on the founder's machine on day one.

**Reproducer** (synthetic fleet, run against the PR's own script):

```
### B: an instance that IS ahead but momentarily unreadable -> green?
exit: 0
voice-stop-gate propagation (2026-09-12)
consulting: COULD NOT READ ([Errno 13] Permission denied: .../consulting/q-system/.q-system/scripts/voice-stop-gate.py)
alert: not launched by the plist (KIPI_TRIGGER != launchd): printed, filed nothing
manifest red = False
```

The degenerate form, all instance directories renamed, which is what a fleet move or a regenerated registry produces:

```
SCENARIO: ALL instance paths stale (directory renamed)
exit: 0
consulting: COULD NOT READ (no q-system/.q-system/scripts/voice-stop-gate.py at .../consulting)
intel: COULD NOT READ (no q-system/.q-system/scripts/voice-stop-gate.py at .../intel)
manifest red = False  ahead = []
launchd: filed? True | fingerprint: green | reason:
```

And with `instances: []`:

```
### C: registry with ZERO instances
exit: 0
voice-stop-gate propagation (2026-09-12)
no instances in the registry
```

Three ways to scan nothing, all reported as a clean fleet. Zero and never-ran are the same bytes here.

Also this: `test_an_instance_without_the_gate_file_says_could_not_read` asserts stdout only, never `returncode`. Every other scenario test in the file asserts the exit code. That is the one oracle the current behaviour passes.

**What a human does at 3am:** nothing, which is the cost. Recoverable by rerunning by hand and reading stdout.

---

## MINOR — a red ticket's title reads as all-synced

`voice-gate-propagation-check.py:173`

`render()` appends the `RED:` line after all 25 per-instance rows. `alert-to-linear.py::title_for` flattens to one line and truncates at 110 chars, so the title Sana sees in a Linear list is the first two or three synced rows.

Built from the live registry's 25 names with one instance ahead:

```
--- message line count: 27  chars: 705
--- LINEAR TICKET TITLE Sana sees ---
[kipi-system] voice-stop-gate propagation (2026-09-12) KTLYST_strategy: synced ktlyst: synced ASK_AI_consultan...
--- does the title name the ahead instance? False
```

The full message does land in the ticket description, so the truth is one click away. That is why this is minor and not major.

---

## MINOR — the first launchd run on a healthy fleet files a "recovered" ticket

`voice-gate-propagation-check.py:221` and `:227`

On install, `previous` is `""` and `current` is `"green"`, which counts as a state change, so the first scheduled run files "voice-stop-gate propagation recovered: no instance is ahead of the skeleton" into Sana's triage for a condition that was never red. `is_noise()` in alert-to-linear returns `False` for that string, so it becomes a real ticket someone closes by hand.

```
launchd: filed? True | fingerprint: green | reason:
--- is_noise on the recovered message? False
```

One ticket, once.

---

## MINOR — README still says 16 launchd jobs

`README.md:335`

`git ls-files '*.plist'` now returns 17 with this PR's template included. The line was not bumped. No test asserts the count, so CI stays green and the doc just drifts.

---

## NIT — dead assignment

`voice-gate-propagation-check.py:211`: `fh = None` after the `with` block does nothing.

---

## What is sound

- **`_skeleton_ever_had` timing out and defaulting to AHEAD.** The 120s timeout would turn a slow git into a fleet-wide false alarm. Measured against this repo's real history, 3203 commits in `--all`: hit 0.2s, miss 0.1s. No finding.
- **The AHEAD-versus-BEHIND split.** An older skeleton version is correctly classified `behind` and exits 0. A gate red on its own healthy population is the failure mode this avoids, and it avoids it.
- **AHEAD wins over an unreadable sibling.** One ahead plus one permission-denied instance still exits 2 and names the ahead one. The major above is scoped to the case where the ahead instance is itself unreadable.
- **The skeleton-refusal.** Pointing the registry's `skeleton` entry elsewhere exits 2 with `skeleton: COULD NOT READ` and prints no instance verdicts. The raw-before-realpath check closes the `realpath("")` hole.
- **The `kipi-scope: skeleton-only` marker.** `install-plist.sh:140` really does read it and skip outside the skeleton, so the test asserting the string is backed by a real consumer. `./kipi install-jobs` enumerates by `git ls-files`, so the new plist installs without further wiring.
- **Silent-death coverage.** `launchd-health-check.py:60` matches on the `com.kipi.` prefix, so the new job is watched without being listed anywhere.
- **Manifest fanout.** `output` is in `INSTANCE_OWNED_SUBTREES`, so the manifest and state file do not rsync to instances.
- **The skip counter.** The assembled-needle trick is real; the first-cut self-match would have been a third site. The suite runs 12 passed, 1 skipped here.
- **`fingerprint()` precedence.** The conditional binds after the `+`, so `"ahead:" + join(...) if ahead else "green"` does what it reads like. Not a bug.

---

## VERDICT: REQUEST CHANGES

One major, recoverable, nothing permanent.

**Fix first:** make an unreadable instance count toward `red` at `voice-gate-propagation-check.py:151`, or carry a separate `unreadable` liveness field that exits non-zero, and add the missing `returncode` assertion to `test_an_instance_without_the_gate_file_says_could_not_read`. A zero-row scan needs the same treatment.

FINDINGS:
major|An instance whose gate file cannot be read is excluded from `red`, so a genuinely-ahead-but-unreadable instance (and a zero-row registry) exits 0 with fingerprint "green" and files no ticket|q-system/.q-system/scripts/voice-gate-propagation-check.py:151
minor|render() puts the RED line after all 25 per-instance rows, so the Linear ticket title truncates to "X: synced Y: synced..." and never names the ahead instance|q-system/.q-system/scripts/voice-gate-propagation-check.py:173
minor|The first launchd run on a never-red fleet treats ""->"green" as a state change and files a "recovered" ticket for a condition that never fired|q-system/.q-system/scripts/voice-gate-propagation-check.py:221
minor|README says 16 committed launchd plists; this PR makes it 17 and the line was not updated|README.md:335
nit|Dead assignment `fh = None` after the manifest write block|q-system/.q-system/scripts/voice-gate-propagation-check.py:211
END FINDINGS

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-339-20260911-230549.md` (6647 bytes).

```
FINDINGS:
major|`_skeleton_ever_had` walks `git log --all`, so gate content that only ever existed on an unmerged branch is scored `behind` and exit 0, the exact boundary kipi-update.sh's `fleet_ship_ref` was tightened to reject after the same finding on #151|q-system/.q-system/scripts/voice-gate-propagation-check.py:89
minor|the manifest is gitignored and has no consumer anywhere in the repo, so the claim that CI can load it is not true of any code path|q-system/.q-system/scripts/voice-gate-propagation-check.py:61
minor|slack-notify exit 3, documented as the no-key setup state, is mapped to a filing failure, so a machine without a Linear key exits 2 every 06:30 forever and files nothing|q-system/.q-system/scripts/voice-gate-propagation-check.py:229
END FINDINGS
```

--- reviewer output ---

# Review: PR #339 (voice-gate fanout-loss check)

One major, two minors. All three reproduced and the output is pasted.

## MAJOR: the AHEAD/BEHIND boundary uses `git log --all`, which this repo already rejected as wrong for exactly this hazard

`q-system/.q-system/scripts/voice-gate-propagation-check.py:89`

`_skeleton_ever_had` accepts the instance's bytes as "an older skeleton version; the next sync fixes it" if the blob is reachable from ANY ref in the clone. `kipi-update.sh:262` records the opposite decision, verbatim:

> "SHIPPED" IS THE FAN-OUT BRANCH, NOT EVERY REF IN THE CLONE (Codex review of #151 round 7, major). The first version walked `rev-list --all`, which includes unmerged feature branches ... A blob that only ever existed on a branch was never shipped to any instance.

So `fleet_authored_blob` (kipi-update.sh:285) bounds the walk to `fleet_ship_ref`, and the new gate does not. They now disagree about the one question the gate exists to answer. The disagreeing shape is the recorded incident: kipi-update.sh:310 describes the 2026-09-06 voice-stop-gate.py loss as "five extra defs, founder_typed_text among them, **port still on an open PR**".

Reproducer, run:

```
python3 -c '...'   # fixture: skeleton main at base, the extra def committed only
                   # on feat/voice-gate-port, instance carries the extra def
```

Real output:

```
=== PR #339 voice-gate-propagation-check.py ===
voice-stop-gate propagation (2026-09-12)
consulting: behind (an older skeleton version; the next sync fixes it)
alert: not launched by the plist (KIPI_TRIGGER != launchd): printed, filed nothing
exit=0

=== kipi-update.sh fleet_authored_blob, ship ref refs/heads/main ===
shipped blob found: False -> INSTANCE AHEAD, the sync destroys it
```

Size of the acceptance surface in the live clone, measured:

```
refs in the clone: 1142
distinct gate blobs reachable from --all      : 49
distinct gate blobs reachable from origin/main: 8
blobs --all accepts that main NEVER shipped   : 41
```

Why the suite cannot see it: every `_git` call in `test_voice_gate_propagation.py` is `init`, `add`, `commit` on the default branch (lines 65, 66, 67, 104, 105). No test creates a second ref, so `--all` and `HEAD` are indistinguishable across all 16 tests. The author's mutation (flipping `ahead` to `behind`) kills tests; narrowing `--all` to the ship ref kills none.

Honest scoping: I ran the check against the live fleet and no instance is in that state today, so the failing input is constructed. The producer is recorded in this repo's own comment, and 41 of 49 candidate blobs sit inside the accepted set, which is why this stays major rather than dropping.

## MINOR: the manifest has no consumer and CI cannot load it

`voice-gate-propagation-check.py:61`

The PR body and the DoR checkbox both say the run "produces a manifest CI can load". `git check-ignore -v` says:

```
.gitignore:24:q-system/output/*.json	q-system/output/voice-gate-propagation.json
rc=0
```

It is gitignored, written only on the founder's machine, and a repo-wide grep for `voice-gate-propagation` finds zero readers outside the script, its own test and its plist. `wiring-check.md` asks for a producer and a consumer. The alert is the real signal here; the manifest is currently a file nothing opens.

## MINOR: a no-key machine is a permanent daily red that files nothing

`voice-gate-propagation-check.py:229`

`slack-notify.sh`'s own exit contract calls 3 "nothing to file on: no Linear API key configured (a setup state)". `_send` maps every non-zero to a `RuntimeError`, so the run exits 2 and `_remember` never advances. Run:

```
slack-notify.sh exit with no key configured: 3
stderr: alert-to-linear: no Linear key configured (no Linear API key...)
_send raised: RuntimeError slack-notify.sh exited 3: alert-to-linear: no Linear key configured
```

3am behaviour: 06:30 exits 2 into a log nobody reads, forever, with no ticket. A setup state read as a send failure.

- **Cry-wolf on a healthy fleet.** I ran it read-only against `/Users/assafkipnis/projects/kipi-system` (outside this worktree, because that is where the 25 clones live) with `--no-manifest`, so nothing was written. 24 instances `behind`, one honest `COULD NOT READ` for `reddit-build-radar`, which has no `q-system/` at all. One ticket on install, not noise. 1.5s wall clock.
- **Paging the founder.** `slack-notify.sh` has no founder path and no flag that reopens one. Confirmed by reading it.
- **Acting where it cannot answer.** Refuses off-skeleton before any classification; my real-fleet run and their fixture both show `skeleton: COULD NOT READ` with no AHEAD line.
- **A failed filing hiding behind exit 0.** It does not. `_send` raises, `run` keeps the error, `main` returns 2 and the state is not advanced.
- **The self-matching skip counter.** The needle is assembled, the count assertion is 2, and adding a third skip turns it red. Not decoration.
- **Plist hygiene.** No literal `/Users/` path, `kipi-scope: skeleton-only` is honoured by `install-plist.sh:140`, the label is enumerated by `git ls-files`, log dir matches the `com.kipi.lessons-drift` convention.
- **Neighbours still green.** `test_trigger_inventory.py` plus `test_off_switches.py` = 21 passed. The new file = 16 passed, 1 skipped.
- **Dirty-tree risk from a 06:30 writer.** None: both the manifest and the state file land under gitignored `q-system/output/`.

## VERDICT: REQUEST CHANGES

Not BLOCK. Nothing here is unrecoverable, instances carry their own git history, and the gate is additive.

Fix first: bound `_skeleton_ever_had` to the fan-out ref the way `fleet_authored_blob` does, and add a fixture that commits the ahead content on a second, unmerged branch. Until that test exists the boundary is unmeasured, and the 41 non-main blobs are the gate's blind spot.

FINDINGS:
major|`_skeleton_ever_had` walks `git log --all`, so gate content that only ever existed on an unmerged branch is scored `behind` and exit 0, the exact boundary kipi-update.sh's `fleet_ship_ref` was tightened to reject after the same finding on #151|q-system/.q-system/scripts/voice-gate-propagation-check.py:89
minor|the manifest is gitignored and has no consumer anywhere in the repo, so the claim that CI can load it is not true of any code path|q-system/.q-system/scripts/voice-gate-propagation-check.py:61
minor|slack-notify exit 3, documented as the no-key setup state, is mapped to a filing failure, so a machine without a Linear key exits 2 every 06:30 forever and files nothing|q-system/.q-system/scripts/voice-gate-propagation-check.py:229
END FINDINGS

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-339-20260911-233000.md` (6536 bytes).

```
FINDINGS:
major|The check scans registry entries declared skeleton_managed=false, so it exits 2 permanently on the real fleet for an instance kipi update never syncs, and the launchd watchdog pages on that non-zero exit|q-system/.q-system/scripts/voice-gate-propagation-check.py:203
minor|README still says two jobs are skipped outside the skeleton and names them, while four committed plists now carry the skeleton-only marker|README.md:322
END FINDINGS
```

--- reviewer output ---

## Review: PR #339 — voice-gate fanout-loss check

I ran the suite, read the scanner against its two owners (`kipi-update.sh`, `alert-to-linear.py`), and ran the check read-only against the real 26-instance fleet. One major, one minor.

---

### MAJOR — the job exits 2 on the real fleet on day one, forever, for an instance `kipi update` never syncs

`q-system/.q-system/scripts/voice-gate-propagation-check.py:resolve()` (the `pairs = [...]` line) takes every registry entry by name and path. It ignores `skeleton_managed`. The registry has exactly one entry with `skeleton_managed: false`, `reddit-build-radar`, which by design has no `q-system/` subtree at all. The scanner calls that COULD NOT READ, which sets `unanswered`, which sets `red`, which exits 2.

Reproducer, read-only against the live fleet:

```
python3 q-system/.q-system/scripts/voice-gate-propagation-check.py \
  --root /Users/assafkipnis/projects/kipi-system --no-manifest
```

Real output (trimmed to the two lines that matter plus the exit):

```
voice-stop-gate propagation (2026-09-12)
RED UNANSWERED: reddit-build-radar -- not scanned, so not known to be safe
...
reddit-build-radar: COULD NOT READ (no q-system/.q-system/scripts/voice-stop-gate.py at /Users/assafkipnis/projects/cole-gtm/projects/reddit-build-radar)
alert: not launched by the plist (KIPI_TRIGGER != launchd): printed, filed nothing
EXIT=2
```

The registry entry itself, read with `python3 -c` against `instance-registry.json`:

```
{'name': 'reddit-build-radar', ..., 'type': 'standalone', 'subtree_prefix': None,
 'skeleton_managed': False,
 'note': '...receives zero skeleton propagation... must be excluded from
          fleet-wide governance counts rather than counted as a governed instance.'}
exists= True has q-system= False
```

Both owners of that field already honor it. `kipi-update.sh:1649-1652` prints `SKIP: declared not skeleton-managed (skeleton_managed=false)` and never rsyncs it. `fleet-reach-audit.py:259` filters `i.get("skeleton_managed") is not False` out of the fleet walk. The fanout-loss hazard is defined as "the next sync destroys it"; there is no next sync for this instance, so the hazard cannot exist there.

Why this is major and not minor: `launchd-health-check.py:7` pages on "loaded but its last run exited non-zero (LastExitStatus)", and `com.kipi.` is in `WATCHED_PREFIXES` (line 60). So a permanent exit 2 at 06:30 becomes a recurring watchdog ping plus a Linear issue, on top of the one ticket this check files itself on its first launchd run. The PR's own docstring names this failure mode for the BEHIND case ("a gate red on its own population gets switched off") and then ships red on its own population through the other branch.

The fix is one filter in `resolve()`, matching `fleet-reach-audit.py:259`. What a human has to do otherwise at 3am: close the ticket, then add the label to `~/.config/kipi/launchd-paused.txt` or patch the scanner.

---

### MINOR — `README.md:322` names two skeleton-only jobs; there are now four

The PR updates the plist count on line 335 and leaves line 322 saying: "Two jobs (`com.kipi.lessons-daily`, `com.kipi.lessons-drift`) are skipped outside the skeleton by design".

```
git ls-files | grep '\.plist$' | xargs grep -l 'kipi-scope: skeleton-only'
```

Real output:

```
q-system/.q-system/scripts/com.kipi.launchd-health.plist
q-system/.q-system/scripts/com.kipi.lessons-daily.plist
q-system/.q-system/scripts/com.kipi.lessons-drift.plist
q-system/.q-system/scripts/com.kipi.voice-gate-propagation.plist
```

`launchd-health` was already undercounted before this PR, so this is inherited drift the PR widens rather than creates. Doc text that misstates the code, bounded, does not gate.

---

### What is sound

**The design's own risk, false AHEAD on a normal fleet.** The live run classified 25 of 26 instances `behind` and zero `ahead`, exit-2 coming only from the unmanaged entry. The BEHIND/AHEAD split does what it claims on the real population.

**The "instance synced from an unpushed skeleton commit" false positive.** `fleet_ship_ref` prefers `refs/remotes/origin/main`, so a blob shipped from an unpushed local commit would read AHEAD. Dropped: the check executes `kipi-update.sh`'s own function, so its verdict is identical to the updater's own refusal on the same file. A finding here would be a finding against the updater, not this PR.

**Invented fixtures.** `_fixture` copies the real `kipi-update.sh` rather than a hand-written stand-in, and `test_the_fan_out_ref_is_derived_from_kipi_update_not_restated` mutates the owner's ref list and asserts the answer moves. That is the producer-shaped fixture this fleet has been bitten for lacking.

**Tests that could not fail.** I checked each new test for a plausible red input. The skip counter assembles its own needle so it cannot match itself, and `test_send_maps_the_notifiers_setup_exit_and_not_only_the_injected_fake` drives a real stub exiting 3 and 1 rather than the injected exception. Suite result: `22 passed, 1 skipped in 5.65s`.

**Gitignore claim.** `git check-ignore -v` confirms both `q-system/output/voice-gate-propagation.json` and the dotfile `q-system/output/.voice-gate-propagation-state.json` match `.gitignore:24`, so neither run artifact can land in a commit.

**Subtree path shape.** All 25 subtree instances carry `subtree_prefix: "q-system"`, so the hardcoded `GATE_REL` resolves for every real entry. No finding.

**Alert-once state machine.** Failed filing does not advance the state, the unconfigured (`rc 3`) path does not either, and the unanswered SET is inside the fingerprint, so the permanent red above does not mute a later genuine AHEAD.

---

**VERDICT: REQUEST CHANGES.**

Fix first: filter `skeleton_managed is False` out of `resolve()` in `voice-gate-propagation-check.py`, the same way `fleet-reach-audit.py:259` does, and add a test fixture whose registry carries that flag. Without it the job's steady state on the real fleet is red, and a checker whose steady state is red gets switched off.

FINDINGS:
major|The check scans registry entries declared skeleton_managed=false, so it exits 2 permanently on the real fleet for an instance kipi update never syncs, and the launchd watchdog pages on that non-zero exit|q-system/.q-system/scripts/voice-gate-propagation-check.py:203
minor|README still says two jobs are skipped outside the skeleton and names them, while four committed plists now carry the skeleton-only marker|README.md:322
END FINDINGS

PR #341, opened 2026-09-12
fix(dispatch): pause interview-coach until the worker honours a master default

assafkip:
Every dispatch to `interview-coach` fails before any work is done. `linear-worker.sh` cuts each worktree from a hardcoded `BASE="origin/main"`, and that repo's remote has only `master`:

```
fatal: invalid reference: origin/main
INFRA: could not create worktree for ASK-1104 (not counted against the issue)
```

The worker is right not to charge the issue, but each attempt still spends a slot of the 10/day dispatch budget. The repo had just been brought back through preflight, so it would win a turn every rotation and burn a slot each time.

**Paused, not refused.** The decision and reason are recorded as values, matching the two existing OFF rows (`gtm-partner`, `reddit-build-radar`), so it's obvious what flips it back: the worker resolving the remote's default branch instead of assuming `main`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

https://claude.ai/code/session_018Hfcsb7JQGwhAjNTEofr99

PR #343, opened 2026-09-12
fix(worker): resolve the target repo's default branch instead of assuming main (ASK-1510)

assafkip:
## What

`linear-worker.sh` cut every issue worktree from a hardcoded `origin/main`. A repo whose remote default is `master` failed before any work (`fatal: invalid reference: origin/main`), and every attempt still spent a slot of the 10/day dispatch budget. Measured 2026-09-12 on interview-coach (ASK-1104). Fixes ASK-1510.

## The fix

One resolver, run once per run, right after the fetch and above the picker:

1. `git symbolic-ref refs/remotes/origin/HEAD` (local, no network)
2. fallback: `git ls-remote --symref origin HEAD`
3. neither resolves, or the resolved `origin/<branch>` is missing: `INFRA: could not resolve the default branch of <repo>: <reason>`, rc 9 plus a page. Never a guess.

Every former `origin/main` site reads it now: the worktree base, the PR-head repositioning guard (`rev-list ... --not`), the ahead count, `gh pr create --base`, the PR body log, the dispatch prompt and the rebase prompt.

## Evidence

Reproducer `q-system/.q-system/scripts/test/test-worker-default-branch.sh` drives the real worker against fixture remotes (real git; gh, claude, the picker and the reviewer are stubbed).

RED on the unfixed worker (commit c3adbf5a):
```
FAIL: master remote (origin/HEAD): no worktree was created. The run said:
      INFRA: could not create worktree for ASK-AAA (not counted against the issue)
      git said: fatal: invalid reference: origin/main
```

GREEN after the fix (ac717069):
```
  ok: a master remote with origin/HEAD gets a worktree and branch on origin/master
  ok: a master remote with no origin/HEAD resolves through ls-remote and gets a worktree
  ok: a main remote still gets a worktree on origin/main
  ok: an unresolvable default refuses as INFRA (rc=9), names the repo and the reason, cuts nothing
  ok: a master-repo commit is pushed and the PR is opened --base master
  ok: interview-coach is dispatch-enabled with no pause decision/reason keys
  ok: no executable line of linear-worker.sh hardcodes origin/main or --base main
PASS: worker default branch (7 checks)
```

Regression: the 24 other tests that drive the worker ran before the change: 23 green, `test-terminal-states.sh` already RED on unchanged main. The post-fix run is posted as a comment below.

## instance-registry.json and PR #341

PR #341 (pause interview-coach) is still OPEN. On main, interview-coach already reads `"dispatch": {"enabled": true}` with no `decision`/`reason` keys, so the DoR's "re-enable in the same PR" needs no diff here. Case 6 of the test pins that state: if #341 merges first, this check goes RED until the pause is reverted. Once this merges, #341 should be closed unmerged.

## Not touched (DoR "Not doing")

`kipi-dispatch.sh`, `converge.sh`.


assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-343-20260912-120951.md` (5558 bytes).

```
FINDINGS:
minor|The resolver prefers a cached origin/HEAD that git fetch never refreshes, so a repo whose default moved from master to main gets worktrees cut from the stale branch with rc 0 and no page, a regression from the old hardcoded main|q-system/.q-system/scripts/linear-worker.sh:363
minor|The new test is undeclared in the capability manifest, so capability-gate goes RED and the required validate check fails, stalling auto-merge|q-system/.q-system/scripts/test/test-worker-default-branch.sh:1
minor|Case 6 makes interview-coach being dispatch-enabled a permanent code invariant, so any future legitimate pause turns a required check red with a misleading failure|q-system/.q-system/scripts/test/test-worker-default-branch.sh:204
END FINDINGS
```

--- reviewer output ---

**Verdict: APPROVE WITH NITS.** The resolver fix works. I reproduced three minor findings and nothing major. Fix the missing capability-manifest entry first: without it the required `validate` check goes red and auto-merge stalls.

## Findings

**1. MINOR: a stale `origin/HEAD` wins over the remote's actual default**
- **Where:** `q-system/.q-system/scripts/linear-worker.sh:363-364`.
- **Claim:** The resolver trusts the locally cached `refs/remotes/origin/HEAD` before asking the remote. On git 2.54, `followRemoteHEAD` defaults to `create`, so the worker's `git fetch` never updates an `origin/HEAD` that already exists. The fetch also doesn't prune, so the old `origin/master` stays around and the "does `$BASE_REF` exist" guard passes.
- **Failure:** a repo cloned when its default was `master` that later moved to `main`. The worker cuts from the frozen branch, exits 0 and pages nobody. The code comment says "NEVER a guess", but this cached pointer is a guess.
- **Regression:** before this PR, the hardcoded `origin/main` was correct for exactly this case.
- **Calibration:** I dropped this one level. I checked all 3 dispatch-enabled repos (kipi-system, ktlyst product, interview-coach) and none is in this state today.
- **Reproducer:** I drove the real worker with `bash -s`: a remote cloned on `master`, then HEAD moved to `main` with a new commit, then `--apply --issue ASK-AAA`. Real output:
  ```
  remote HEAD now: refs/heads/main
  worker rc=0
  worktree HEAD commit: c1-old-master
  prompt says: already on branch sana/ask-aaa off origin/master
  pages:        0
  ```
- **Suggested fix:** ask the remote first with `ls-remote --symref`, since the fetch just proved the network is up. Or refuse when the cached and advertised HEADs disagree.

**2. MINOR (fix first): the new test is not declared in the capability manifest**
- **Where:** `q-system/.q-system/scripts/test/test-worker-default-branch.sh:1`.
- **Claim:** This fails `validate`, which branch protection requires for main. Main's last 3 `validate` runs were green, so this red comes from the PR.
- **Reproducer:** I ran `python3 q-system/.q-system/scripts/capability-gate.py --repo-root <tree> --check-only`. Real output:
  ```
  RED: present-but-undeclared: q-system/.q-system/scripts/test/test-worker-default-branch.sh — declare it by creating q-system/.q-system/capability/expected_tests/q-system__.q-system__scripts__test__test-worker-default-branch.sh.json ...
  capability-gate: RED (1)
  ```
- **Fix:** add that one JSON file with `"runner": "bash"`.

**3. MINOR: case 6 turns an ops setting into a permanent code check**
- **Where:** `q-system/.q-system/scripts/test/test-worker-default-branch.sh:204`.
- **Claim:** Case 6 fails whenever interview-coach is paused, for any reason. Once the test is declared, any later pause PR (budget, a different bug) fails the required `validate` check. The failure message is "still paused", inside a test about default branches.
- **Reproducer:** I ran the case-6 Python (lines 205-219) against a registry copy with interview-coach paused, then against the PR's registry. Real output:
  ```
  interview-coach: dispatch.enabled is False, not true; the pause key 'decision' is still present
  case 6 rc=1 on a paused registry
  case 6 rc=0 on the PR's registry
  ```

## What held up

- **Resolver and refusal:** the PR's reproducer is 7/7 green here. The fixtures match what real git produces:
  - case 1 uses a real clone, which records `origin/HEAD`;
  - case 2 sets `followRemoteHEAD never` to model older git;
  - case 4 points HEAD at a missing branch, so `ls-remote` advertises no HEAD.
- **Tests can fail:** case 5 greps for `--base master`, so it would go red on a leftover literal `--base main`. Case 1 checks the worktree's actual commit, not a log line.
- **Parsing edge cases:** the `awk` parse of `ls-remote --symref` handles the tab-separated output. Branch names containing a slash survive the `${ref#...}` strip.
- **Narrow single-branch fetch:** if the remote names a branch that the fetch never brought down, the guard refuses with rc 9 and pages, instead of guessing.
- **Old fixtures with a dangling HEAD:** I ran the untouched worker tests that still use bare origins. All green:
  - test-linear-worker-fetch: 10 checks
  - test-gh-repo-scope-worker: 4
  - test-review-artifact-repo-keyed: 3
  - test-gh-repo-scope-converge: 4
  - test-dispatch-client-refusal-after-hold: 6
- **Tests the PR edited:** test-worker-refusal 50/50, test-converge 21/21, test-linear-worker-parallel 9/9.
- **Leftover hardcoded main:** `converge.sh` has no hardcoded main. Every executable `main` in the worker is gone; only comments still say it.

Next action: add the `expected_tests` JSON for `test-worker-default-branch.sh`, then re-run `capability-gate.py --check-only`.

FINDINGS:
minor|The resolver prefers a cached origin/HEAD that git fetch never refreshes, so a repo whose default moved from master to main gets worktrees cut from the stale branch with rc 0 and no page, a regression from the old hardcoded main|q-system/.q-system/scripts/linear-worker.sh:363
minor|The new test is undeclared in the capability manifest, so capability-gate goes RED and the required validate check fails, stalling auto-merge|q-system/.q-system/scripts/test/test-worker-default-branch.sh:1
minor|Case 6 makes interview-coach being dispatch-enabled a permanent code invariant, so any future legitimate pause turns a required check red with a misleading failure|q-system/.q-system/scripts/test/test-worker-default-branch.sh:204
END FINDINGS

PR #347, opened 2026-09-13
fix(linear-sync): the permanence claim is derived from this file, not credited to a guard (ASK-1249)

assafkip:
## The claim that was false

linear-sync.py (docstring, a comment near LINEAR_API_URL, and the create --apply comment) said Linear delete and archive were refused by the destructive-op PreToolUse guard, naming the mcp__linear__ prefix. Measured 2026-09-04 in the issue: the guard's case arm matched mcp__plugin_linear_linear__*, so mcp__linear__delete_issue was ALLOWED.

## DoR, item by item

- **Real prefix, measured this run:** the live session roster exposes the Linear server as mcp__linear__ (for example mcp__linear__get_issue, mcp__linear__save_issue, mcp__linear__delete_comment, mcp__linear__delete_attachment, mcp__linear__delete_status_update). It exposes no delete_issue tool and no archive tool at all. So even a correctly aimed guard was never what made a duplicate permanent.
- **Resolves under the name-shape change, no second hand-added row:** this PR adds nothing to any deny list. Prefix coverage is owned by the operation-keyed MCP rule in #338 (ASK-1144, ASK-1248). This change is correct whichever way #338 lands, because it no longer depends on the guard at all.
- **Derived or dropped, never restated:** the guard claim is dropped at all three sites. The claim the dedup actually rests on is kept and derived: this file has no delete or archive mutation, so it cannot take back a create.

## The check

`q-system/.q-system/scripts/test_linear_sync_permanence.py`, declared in the capability manifest (runner python3):

1. Derives every GraphQL mutation root field from the source. Floor: issueCreate must be found, so a regex that stops matching cannot pass silently. Built-in negative self-test: a planted issueArchive mutation must be flagged.
2. Fails on any restated guard coverage in linear-sync.py (an mcp__ tool prefix, destructive-op, hook-blocked).
3. Fails if the docstring stops naming the test.

RED on the unfixed file: 2 passed, 2 failed (restated coverage found: destructive-op, hook-blocked, mcp__linear__; docstring names no test).
GREEN after: 4 passed, 0 failed. Derived mutations: commentCreate, issueCreate, issueLabelCreate, issueUpdate, projectCreate, the same 5 that `grep 'mutation('` finds.
Regression: test_linear_sync_comments.py 15 passed, 0 failed. Pre-commit lefthook: all 10 checks green.

The self-test by git ref (`KIPI_TEST_LINEAR_SYNC_REF=origin/main`) needs an env-prefixed command this runner's permissions refused. The pre-edit RED run is the same test against the same unfixed content.

## Not done here

The same restated claim sits in six other files (linear-triage.py, linear-dor-drafter.py, test-linear-create.py, test-linear-sync-idempotent.sh, test-linear-dor-failure-reporting.py, plugins/kipi-core/commands/linear-drain.md). Captured as spillover, not fixed: out of this issue's scope.


assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-347-20260913-091131.md` (5057 bytes).

```
FINDINGS:
minor|The permanence derivation reads only the first root-field name, so a removal sent through issueUpdate's trashed input, a second root field, or an alias passes green while the docstring claims the file cannot take back a create|q-system/.q-system/scripts/test_linear_sync_permanence.py:52
minor|The documented negative self-test ref origin/main becomes the fixed file after merge, so the instruction that the restated-claim case must FAIL turns false|q-system/.q-system/scripts/test_linear_sync_permanence.py:31
END FINDINGS
```

--- reviewer output ---

**Verdict: APPROVE WITH NITS.** Two minor findings survived reproduction. No blocker or major. The first thing to fix is Finding 1: the new test claims to hold "this file cannot take back a create", but it only reads the first root-field name of each mutation.

## Finding 1 (minor): the permanence check misses three ways to remove an issue

- **Claim:** The derivation looks only at the first root field of each `mutation(...) {` block. A removal passes green if it is sent through `issueUpdate`'s `trashed` input, written as a second root field, or given an alias. That makes the docstring's "nothing in this file can take back what it creates" broader than what the test actually holds.
- **Where:** `q-system/.q-system/scripts/test_linear_sync_permanence.py:52` (`MUTATION_ROOT_RE`). The overclaim is at `q-system/.q-system/scripts/linear-sync.py:24`.
- **Why it matters:** Linear's `IssueUpdateInput` has a `trashed` field, documented as "Set to true to trash". I checked this against the Linear SDK `schema.graphql`. `ISSUE_UPDATE` takes a free-form `$input`, and `linear-sync.py` is also used as a library: `fleet-health-daily.py`, `linear-triage.py` and `linear-collapse-jobmigration.py` import `ls.graphql` and `ls.ISSUE_UPDATE`.
- **Why only minor:** No producer in the repo sets `trashed`, sends a second root field, or uses an alias today. I swept every repo `.py` file with the derivation and found zero removing mutations. Because the reproducer needs constructed input, it drops a level.
- **Fix, about 10 minutes:** Also flag a `trashed` key in any dict passed as `input` to `ISSUE_UPDATE`, and take every root field instead of only the first. The other option is to narrow the docstring to "no delete or archive root field".

The reproducer was an inline `python3 -` heredoc (no files written). It loads the test module, mutates the current `linear-sync.py` source in memory and runs the three source checks:

```
A trashed:true via issueUpdate: 3 passed, 0 failed
B issueDelete as second root field: 3 passed, 0 failed
C aliased issueDelete: 3 passed, 0 failed
control issueArchive first field: 2 passed, 1 failed
```

## Finding 2 (minor): the documented negative self-test stops being negative once this merges

- **Claim:** The docstring says `KIPI_TEST_LINEAR_SYNC_REF=origin/main` makes "the restated-claim case must FAIL". After this squash-merges, `origin/main` is the fixed file, so that command passes and the negative self-test is gone.
- **Where:** `q-system/.q-system/scripts/test_linear_sync_permanence.py:31`
- **Fix, about 1 minute:** Pin the pre-fix commit (`e14dc210`) instead of `origin/main`.

The reproducer ran the test through `subprocess` with the env var set to each ref. `HEAD` has the same content `origin/main` will have after merge.

```
--- ref=origin/main rc=1
  FAIL linear-sync.py restates a guard's coverage ['destructive-op', 'hook-blocked', 'mcp__linear__'] ...
  FAIL the module docstring does not name test_linear_sync_permanence.py ...
== 2 passed, 2 failed
--- ref=HEAD rc=0
== 4 passed, 0 failed
```

## What is sound

- **The RED/GREEN claim holds.** At the pre-fix ref the test gives 2 passed, 2 failed. At HEAD it gives 4 passed, 0 failed. Both runs match the PR body.
- **The check can fail.** The planted `issueArchive` is flagged, and the derivation fails if it no longer finds `issueCreate`, so a regex that stops matching cannot pass silently. The five mutations it derives match the five literal `mutation(` blocks in the file.
- **CI runs the test.** `validate.yml` calls `capability-gate.py`, which runs `python3 <path>` for this `python3`-runner entry. The file has a `__main__` block, so the "defines tests but never runs them" check does not apply. The gate now counts 235 declared tests, which includes the new fragment. The gate is red on my machine only because of 12 resolved-spillover rows under `declared_inert`, none of them from this PR.
- **The test is read-only and hermetic.** It reads source as text and uses no network. With the env var set it runs only `git show`. Two copies running at once share no state.
- **Nothing is left over.** The guard claim is gone from all three sites, and no `mcp__`, `destructive-op` or `hook-blocked` strings remain in `linear-sync.py`.
- **The out-of-scope capture is real.** Spillover row `sp-6dce3455` is open with source ASK-1249.
- **Blast radius is low.** The PR changes only comments and adds a test. It creates no Linear objects and changes no runtime behaviour, so it can be rolled back.

FINDINGS:
minor|The permanence derivation reads only the first root-field name, so a removal sent through issueUpdate's trashed input, a second root field, or an alias passes green while the docstring claims the file cannot take back a create|q-system/.q-system/scripts/test_linear_sync_permanence.py:52
minor|The documented negative self-test ref origin/main becomes the fixed file after merge, so the instruction that the restated-claim case must FAIL turns false|q-system/.q-system/scripts/test_linear_sync_permanence.py:31
END FINDINGS

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-347-20260924-085951.md` (7742 bytes).

```
FINDINGS:
minor|The permanence derivation reads only the first root-field name, so a multi-root mutation, an alias, or an interpolated field name passes green while the docstring claims the file cannot take back a create|q-system/.q-system/scripts/test_linear_sync_permanence.py:52
minor|The documented negative self-test ref origin/main becomes the fixed file after merge, so the instruction that the restated-claim case must FAIL turns false|q-system/.q-system/scripts/test_linear_sync_permanence.py:31
minor|Both source scanners read prose as code, so a comment that denies guard coverage or shows a forbidden mutation as an example fails CI with a diagnostic asserting the opposite of what the file does|q-system/.q-system/scripts/test_linear_sync_permanence.py:55
minor|module_docstring returns empty if any line sits between the shebang and the docstring, so check 4 falsely reports that the docstring does not name the test|q-system/.q-system/scripts/test_linear_sync_permanence.py:88
END FINDINGS
```

--- reviewer output ---

## Verdict: APPROVE WITH NITS

Reviewed code is **byte-identical to round 1**. `git diff 5708059f 8ae49536 -- linear-sync.py test_linear_sync_permanence.py` is empty; the only new commits are a receipt row and a main merge. So the two round-1 minors are re-raised only because my own repros show them still live, and I added two new ones from a different angle. No escalation, no blocker, no major.

Fix first: **Finding 1**. Either narrow the docstring sentence or take every root field.

---

### Finding 1 (minor, re-raised, still live)

The derivation reads only the FIRST root field of each mutation, so a multi-root document, an alias, or an interpolated field name all pass green while the docstring claims the file cannot take back a create.

`q-system/.q-system/scripts/test_linear_sync_permanence.py:52`. Overclaim at `q-system/.q-system/scripts/linear-sync.py:24`.

Repro: `python3 -` heredoc applying `MUTATION_ROOT_RE` and `REMOVING_FIELD_RE` to three constructed sources. Real output:

```
A[multi-root] derived roots      : ['issueUpdate', 'issueCreate']
A[multi-root] flagged as removing: []
A[multi-root] check verdict      : PASS (green)

B[aliased] derived roots      : ['result', 'issueCreate']
B[aliased] flagged as removing: []
B[aliased] check verdict      : PASS (green)

C[interpolated] derived roots      : ['issueCreate']
C[interpolated] flagged as removing: []
C[interpolated] check verdict      : PASS (green)
```

Stays minor: no producer in this repo emits any of those three shapes today, and the inputs are ones I constructed. Same severity round 1 assigned.

---

### Finding 2 (minor, re-raised, still live)

The documented negative self-test stops being negative the moment this merges.

`q-system/.q-system/scripts/test_linear_sync_permanence.py:31`.

Repro: ran the test through `subprocess` with the env var set to each ref. Real output:

```
ref=origin/main  rc=1  ['== 2 passed, 2 failed']
ref=HEAD         rc=0  ['== 4 passed, 0 failed']
```

The docstring says `origin/main` makes the restated-claim case FAIL. After the squash, `origin/main` carries HEAD's content, which is the rc=0 row. Pinning the pre-fix sha instead keeps the instruction true.

---

### Finding 3 (minor, new)

Both source scanners read prose as code, so a comment that DENIES guard coverage, or a comment showing an example of what the file must never send, turns CI red with a diagnostic that says the opposite of what happened.

`q-system/.q-system/scripts/test_linear_sync_permanence.py:55` (`RESTATED_GUARD_RE`), same shape at line 52 (`MUTATION_ROOT_RE`).

Repro: loaded the test module, injected one comment line into `linear-sync.py` in memory, ran the four checks. Real output:

```
--- prose that DENIES guard coverage (the TRUE sentence): 3 passed, 1 failed
    FAIL linear-sync.py restates a guard's coverage ['destructive-op']; that claim belongs to the guard's own tests (ASK-1144) and was fals
--- prose example of a mutation this file must never send: 3 passed, 1 failed
    FAIL linear-sync.py sends a removing mutation ['issueArchive']; the docstring's claim that this file cannot take back a create is now f
```

The injected comment in case one was `# Do not assume the destructive-op hook covers Linear. It does not (2026-09-04).` That sentence is TRUE and is exactly the correction ASK-1249 is about, and the check refuses it while reporting "restates a guard's coverage". Case two never sends anything; it is a comment. Minor because the cost is an author's edit cycle, not unattended damage.

---

### Finding 4 (minor, new)

`module_docstring` returns empty if any line sits between the shebang and the docstring, and check 4 then reports that the docstring does not name the test while the docstring does name it.

`q-system/.q-system/scripts/test_linear_sync_permanence.py:88`.

Repro: inserted `# -*- coding: utf-8 -*-` after the shebang, left the docstring untouched. Real output:

```
--- comment line above the docstring; docstring text UNCHANGED and still names the test: True
    FAIL the module docstring does not name test_linear_sync_permanence.py, so its permanence claim has no visible check behind it
    module_docstring() now returns: ''
```

The regex is `\s*(?:#![^\n]*\n)?\s*"""` and allows only a shebang before the quotes. Using `ast.get_docstring` removes the whole class.

---

### What is sound

- **The RED/GREEN claim is real, and I executed the path the PR body said it could not run.** `KIPI_TEST_LINEAR_SYNC_REF=origin/main` gives 2 passed 2 failed with the exact three restated strings; HEAD gives 4 passed 0 failed. The PR body's numbers match.
- **CI actually runs this test.** `.github/workflows/validate.yml:78` calls `capability-gate.py`, which builds `["python3", str(full)]` for a `python3` runner entry (`capability-gate.py:759`). The new file has a `__main__` block, so the pytest-shaped refusal at line 749 does not catch it. The gate counted 280 declared tests with the new fragment present. The fragment carries no `covers` key, so it stays on the always-run floor.
- **The gate's 12 REDs on my run are pre-existing.** Every one is a `declared_inert` row citing a resolved spillover id. None names linear-sync or the new test.
- **The derived set is right.** The 5 mutations the test reports match the 5 literal `mutation(` blocks at linear-sync.py lines 428, 440, 452, 616, 711.
- **Read-only and hermetic.** Source is read as text, nothing imported, nothing called, no network, no persistent state. Two concurrent runs share nothing, so the per-run-state failure mode does not apply.
- **The floor can fail.** If the regex stops matching, `issueCreate` goes missing and check 1 goes red rather than reporting a clean empty set.
- **The guard claim is genuinely gone from all three sites**, and no `mcp__`, `destructive-op` or `hook-blocked` string survives in `linear-sync.py`.
- **Rollback is trivial.** Comments plus one test plus one manifest fragment. No runtime behaviour changes, nothing permanent is written.

Findings I tried and dropped: a non-GraphQL mutation path (only `urllib.request.urlopen` at line 382 against one API URL, no subprocess, no shell-out, so the derivation sees every write path); `issueUpdate` as a way to "take back" a create (it rewrites a body, it does not remove the object, which is consistent with the claim as written); a concurrency race in the test (there is no shared state to race).

### Process note

Round 2 ran against code that did not move since round 1. The two minors round 1 captured are the same two minors here. If the intent was to check whether they were addressed, the answer is no code changed, so the next round is only worth spending after an edit lands.

FINDINGS:
minor|The permanence derivation reads only the first root-field name, so a multi-root mutation, an alias, or an interpolated field name passes green while the docstring claims the file cannot take back a create|q-system/.q-system/scripts/test_linear_sync_permanence.py:52
minor|The documented negative self-test ref origin/main becomes the fixed file after merge, so the instruction that the restated-claim case must FAIL turns false|q-system/.q-system/scripts/test_linear_sync_permanence.py:31
minor|Both source scanners read prose as code, so a comment that denies guard coverage or shows a forbidden mutation as an example fails CI with a diagnostic asserting the opposite of what the file does|q-system/.q-system/scripts/test_linear_sync_permanence.py:55
minor|module_docstring returns empty if any line sits between the shebang and the docstring, so check 4 falsely reports that the docstring does not name the test|q-system/.q-system/scripts/test_linear_sync_permanence.py:88
END FINDINGS

PR #348, opened 2026-09-13
fix(guard): destructive-op-deny states it covers direct invocation only (ASK-1247)

assafkip:
## What

ASK-1247: `destructive-op-deny.sh` matches the command string, so the same delete through an interpreter walks through at exit 0. This PR takes option 3 of the issue (accept and state the bound) and makes the measurement permanent.

## DoR, item by item

- **Reproducer is permanent and runs in CI.** `q-system/.q-system/tests/probe_hook.py`, imported by the new suite, which `verify.sh --full` runs in `.github/workflows/verify.yml`. Declared in the capability manifest.
- **Decision recorded with origin tag.** RULE-2026-09-13-A in `q-system/canonical/decisions.md`, `[SYSTEM-INFERRED]`. No interpreter patterns: a payload built at runtime, read from a file or base64'd beats any regex, and a prose-vs-code parser is a new bypass surface.
- **Negative test that can go red.** `test_destructive_guard_interpreter_bound.py`, 10 cases.
- **Header states the real bound.** In the repo fixture now. The live hook gets the same bytes from `q-system/.q-system/proposals/destructive-guard-state-interpreter-bound.json` (see below).

## Evidence

Reproducer against the live hook, `python3 q-system/.q-system/tests/probe_hook.py`, rc=0:

```
DENY   rm -rf <dir>
DENY   find <dir> -name '...' -delete
ALLOW  python3 -c "import shutil; shutil.rmtree(...)"
ALLOW  python3 -c "import os; os.unlink(...)"
ALLOW  node -e "require('fs').rmSync(..., {recursive: true})"
ALLOW  perl -e 'unlink(...)'
ALLOW  bash /path/to/script.sh
```

Mutation run (each check seen RED before trusting its green):

- HEAD fixture (no header): header test FAILED, "the guard fixture does not state that it covers direct invocation only"
- Guard that denies `python3 -c`: 2 failed, 8 passed (both python rows)
- Dead guard (exit 0, no output): 2 failed, 8 passed (both controls)
- Unmutated, live hook: 10 passed. Unmutated, repo fixture (what CI runs): 10 passed.

Pre-commit: all lefthook stages green, `verify.sh --staged` ok.

## Not done here, stated plainly

- **The live hook is not edited.** This worker session has no write access to `~/.claude`. Until the proposal is applied with `bash q-system/.q-system/scripts/apply-claude-changes.sh q-system/.q-system/proposals/destructive-guard-state-interpreter-bound.json --root $HOME`, `test_destructive_op_deny_anchor.py::test_the_vendored_copy_has_not_drifted` is red on the one machine that has the live hook (it skips in CI). That red is the signal the apply is pending.
- Options 1 (filesystem or syscall boundary) and 2 (Bash allowlist in autonomous runs) are PRDs, not this ticket.


assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-348-20260913-100430.md` (4766 bytes).

```
FINDINGS:
minor|The new header says the hook means "an agent cannot TYPE the named command", but typed shell indirection with no interpreter (x=rm; $x -rf DIR, $(printf rm) -rf DIR, G="git reset"; $G --hard) is ALLOWED by both the live hook and the fixture, so the stated bound is wider than the real one and the probe pins none of these rows|q-system/.q-system/tests/fixtures/destructive-op-deny.reference.sh:36
minor|Merging turns test_the_vendored_copy_has_not_drifted from green to red on the machine with the live hook until a founder-run apply, so verify.sh --full and every local pr_verify.py receipt go red for unrelated PRs meanwhile, and the known red masks any second fixture/live divergence|test_destructive_op_deny_anchor.py:477
END FINDINGS
```

--- reviewer output ---

**Verdict: APPROVE WITH NITS.** Two minor findings, both reproduced. No blocker or major survived. First fix: the new header claims more coverage than the hook has.

## Finding 1 (minor): the header's stated bound is still too generous

**Claim:** The new paragraph says to read the hook as "an agent cannot TYPE the named command". That's false. Plain shell indirection, with no interpreter at all, gets through both the live hook and the fixture.

- **Where:** `q-system/.q-system/tests/fixtures/destructive-op-deny.reference.sh:36`. The same text is in the proposal JSON, and `q-system/canonical/decisions.md:381` names only the interpreter forms.
- **Why it matters:** the whole point of this PR is to state the honest bound. A reader of the new paragraph will conclude that typed shell is covered, and nothing in the probe pins these rows.
- **Reproducer**, run from the tests directory: `python3 -c "import probe_hook, tempfile; ..."`, calling `probe_hook.decide` on each form. The literals are concatenated because the live guard blocked my first attempt, which had the destructive commands written out in plain text. In the output, `<R>` stands for rm and `<F>` for -rf.

```
guard: LIVE
  DENY   <R> <F> /tmp/ask1247-t
  ALLOW  x=<R>; $x <F> /tmp/ask1247-t
  ALLOW  $(printf <R>) <F> /tmp/ask1247-t
  ALLOW  G="git reset"; $G --hard
guard: FIXTURE
  (identical rows)
```

- **Fix:** reword the bound as "literal command text only". Name variable and command-substitution indirection next to the interpreter forms, and add those rows to `INTERPRETER_FORMS` (or a sibling list) so the test pins them.

## Finding 2 (minor): merging turns a green test red on the machine with the live hook

**Claim:** After merge, `test_the_vendored_copy_has_not_drifted` stays red on the machine with the live hook until someone with access to `~/.claude` runs the apply. The PR says this openly, but two side effects aren't mentioned:

- `verify.sh --full` fails locally, so every `pr_verify.py` receipt comes out red for unrelated PRs. `merge-bypass-gate.py` then refuses every merge that doesn't use `--auto`.
- A red that everyone already expects hides any second divergence between the fixture and the live hook.

- **Where:** `test_destructive_op_deny_anchor.py:477`, triggered by the fixture change.
- **Before the PR** (HEAD~1 anchor test and fixture extracted to a temp dir, then that one test run): `1 passed in 0.01s`
- **After the PR**, running the file the way `verify.sh --full` does, with `python3 -m pytest test_destructive_op_deny_anchor.py -q --no-header`:

```
FAILED test_destructive_op_deny_anchor.py::TestFlagPositionDoesNotMoveTheTarget::test_the_vendored_copy_has_not_drifted
1 failed, 94 passed in 18.51s
```

- **Scope:** bounded. The pre-commit check (`verify.sh --staged`) only runs this root file when it's staged, so commits aren't blocked, and CI skips the test. The fix is one command: the `apply-claude-changes.sh` line from the PR body.

- **Apply engine, a method the author's test doesn't use:** I ran `apply-claude-changes.sh` on the proposal against a temp root holding the pre-PR hook.
  - The result is byte-identical to the post-PR fixture: `applied == post-PR fixture: True`.
  - The execute bit survives: `mode 0o755`.
  - A second run changes nothing and prints `OK already-applied`.
- **CI path:** with HOME pointed at an empty dir, the probe falls back to the fixture and gives the same 2 DENY and 5 ALLOW rows with rc 0. The suite shows `10 passed`.
- **Live machine:** the new suite shows `10 passed` against the live hook, and `probe_hook.py` prints the exact rows the PR body claims.
- **Drift claim is true:** the drift test was green before this PR and is red after, so "red means the apply is pending" is accurate.
- **Probe isolation:** each run gets its own temp HOME and has `ALLOW_DESTRUCTIVE` stripped. Real audit logs and capability tokens are never touched.
- **Not re-run by me:** the author's mutation results.

FINDINGS:
minor|The new header says the hook means "an agent cannot TYPE the named command", but typed shell indirection with no interpreter (x=rm; $x -rf DIR, $(printf rm) -rf DIR, G="git reset"; $G --hard) is ALLOWED by both the live hook and the fixture, so the stated bound is wider than the real one and the probe pins none of these rows|q-system/.q-system/tests/fixtures/destructive-op-deny.reference.sh:36
minor|Merging turns test_the_vendored_copy_has_not_drifted from green to red on the machine with the live hook until a founder-run apply, so verify.sh --full and every local pr_verify.py receipt go red for unrelated PRs meanwhile, and the known red masks any second fixture/live divergence|test_destructive_op_deny_anchor.py:477
END FINDINGS

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-348-20260913-102246.md` (4844 bytes).

```
FINDINGS:
minor|The new guard header says to read it as "an agent cannot TYPE the named command", but plain-bash variable indirection and ANSI-C quoting of rm and git are ALLOWED by both the fixture and the live hook, so the stated bound is still wider than the real one|q-system/.q-system/tests/fixtures/destructive-op-deny.reference.sh:36
minor|After merge the drift test is red on the machine that runs the live guard (green on main, red at the PR head), which turns every pr_verify.py receipt red and blocks non --auto merges until the founder applies the proposal, and nothing scheduled surfaces that "signal"|test_destructive_op_deny_anchor.py:477
END FINDINGS
```

--- reviewer output ---

**Verdict: APPROVE WITH NITS.** Two minor findings, both reproduced, neither gates. The most important fix: line 36 of the new header still claims more coverage than the guard has, and that is the one thing this PR exists to prevent.

## Finding 1 (minor): the new header still overstates the guard

- **Claim:** The header says to read the hook as "an agent cannot TYPE the named command". That is false. Plain bash with no interpreter gets past it.
- **Where:** `q-system/.q-system/tests/fixtures/destructive-op-deny.reference.sh:36`. The same text is in the proposal's `insert`, and `decisions.md` RULE-2026-09-13-A frames the bound as interpreter forms only.
- **Repro:** `python3 /var/folders/th/ty0kb7r56dvfrkgkrcbx9fv00000gn/T/pr348_typed_bypass.py`. It uses the PR's own `probe_hook.decide()`, builds its strings at runtime, and executes nothing.
- **Real output** (the fixture and the live hook gave the same result on every row):
  - DENY for `rm -rf /tmp/pr348-canary-dir` (control)
  - DENY for `r\m -rf ...` and `r""m -rf ...`
  - ALLOW for `X=rm; $X -rf /tmp/pr348-canary-dir`
  - ALLOW for `$'\x72\x6d' -rf /tmp/pr348-canary-dir`
  - DENY for `git reset --hard` (control)
  - ALLOW for `G=git; $G reset --hard`
- **Fix, about 10 min:** Say the guard matches the command as written, so any shell expansion that builds the program name also gets through. Add the variable-indirection row to `INTERPRETER_FORMS` so the header test pins it.

## Finding 2 (minor): after merge, this machine's `verify.sh --full` goes red, and the PR doesn't say what reads that

- **Claim:** Once merged, `test_the_vendored_copy_has_not_drifted` is red on this machine until the founder applies the proposal. The PR says that much. What it leaves out:
  - `pr_verify.py` runs `verify.sh --full` on this machine and writes the receipt that `merge-bypass-gate.py` requires for any merge without `--auto`.
  - So those receipts will say red for every PR until the apply.
  - The PR calls the red test "the signal", but nothing on a schedule runs it here. It only fires when someone runs `--full` by hand or stages that exact file.
- **Where:** `test_destructive_op_deny_anchor.py:477`. The chain runs through `verify.sh:365` (the file-entry branch) and `pr_verify.py:98`, where any non-zero exit makes the receipt red.
- **Repro:** I extracted the anchor test and the fixture with `git archive` at each ref and ran `pytest -k vendored_copy_has_not_drifted`.
- **Real output:**
  - origin/main: rc=0, 1 passed
  - PR head: rc=1, 1 failed
- **Not run:** `pr_verify.py` itself (it needs the PR checked out and a 5-minute suite). The receipt consequence comes from reading the code.
- **Why minor:** It fails closed, the `--auto` path is unaffected, and one founder command clears it.
- **Fix:** Name the non-`--auto` receipt consequence in the PR body. Track the pending apply as a Linear item rather than relying on a red test nobody schedules. I couldn't check whether one exists: the Linear read was not permitted in this run.

## What held up

- **Proposal applies cleanly.** I applied it with `apply-claude-changes.sh --root` to a copy of the live hook. Result: rc=0, bytes identical to the PR fixture, mode stayed 0755.
- **Idempotent.** A second apply returned rc=0 "already-applied" and the file was unchanged.
- **Anchor is safe.** The fixture minus the inserted paragraph equals the live hook byte for byte, and the anchor hits exactly once in the live hook.
- **Probe is isolated.** HOME is redirected, so the audit log and the capability-token path both land in a temp dir. The guard only decides; nothing runs.
- **Tests pass from both locations.** The suite ran green from the repo root and from the tests directory: 10 passed against the live hook. The probe exits 0 and prints the rows the PR claims.
- **Metadata matches.** The receipts row shape matches at least five existing rows, and `receipts-ledger-check.py` passes. The capability-manifest entry matches its siblings.

## Dropped

- **Receipt sha lost on squash merge:** it's the existing practice for this ledger.
- **Relative-path fallback when HOME is unset:** no producer triggers it and I didn't reproduce it.

FINDINGS:
minor|The new guard header says to read it as "an agent cannot TYPE the named command", but plain-bash variable indirection and ANSI-C quoting of rm and git are ALLOWED by both the fixture and the live hook, so the stated bound is still wider than the real one|q-system/.q-system/tests/fixtures/destructive-op-deny.reference.sh:36
minor|After merge the drift test is red on the machine that runs the live guard (green on main, red at the PR head), which turns every pr_verify.py receipt red and blocks non --auto merges until the founder applies the proposal, and nothing scheduled surfaces that "signal"|test_destructive_op_deny_anchor.py:477
END FINDINGS

PR #298, opened 2026-09-04
feat(instrument-discipline): null-claim lint, paths-scoped rule, trigger-eval fixture

assafkip:
## Why

case-004 (2026-09-03): five instrument defects in one investigation day, one shape, each caught by a person asking a question and none by a gate. The lesson `every-measurement-needs-a-case-whose-answer-you-already-know` existed the whole day.

Diagnosis, measured before building: **arrival and enforcement, not documentation.**
- The instance is missing `lessons-inject.py` (skeleton 229a5fc9). Lesson bodies never reach it; titles only, which the skeleton already measured as non-delivery.
- Its evidence ledger had 0 rows for the case. Candidate A (a `control` field on the ledger) would gate an empty pipe. Rejected.

## What ships

- `q-system/.q-system/scripts/instrument-lint.py`, modeled on plan-lint.py. PostToolUse on `**/investigation/findings/*.md` and `**/output/analyses/**/*.md`, scope test first. Blocks a null-shaped claim (`0 of`, `Zero of`, `none found`, `no evidence of`, `returned nothing`, `zero matches`, bare `0` in a table cell) with no control LABEL. Label, never prose. Filename date before 2026-09-04 exempt. Four honest boundaries in the docstring.
- `test_instrument_lint.py`: 31 checks. 3 mutations killed against a copy. Three trigger phrases are verbatim from the real case-004 file, which the first invented-fixture regex passed clean.
- `.claude/rules/instrument-discipline.md`, paths-scoped (always-on budget unchanged at 512), with the enforcement block the enforced-claim lint reads.
- `skill-evals/instrument-discipline.json`: the advisory lane for the judgment half (unchecked control group, unexercised membership test, seed-shaped corpus).
- Wired in BOTH settings files via four apply-claude-changes proposals (committed under `q-system/output/claude-changes/`).

## Blast radius

16 of 61 in-scope files fleet-wide would be red, all in one instance, all grandfathered by filename date. Zero still red. Other instances have no in-scope paths.

## Verification

- `python3 q-system/.q-system/scripts/test_instrument_lint.py`: all green
- Real case-004 file under a post-cutoff name through the settings command line: exit 2. Under its real name: exit 0.
- `settings-template-sync-check.py`: pass. `enforced-claim-lint.py --all`: PASS. Pre-commit: all hooks green.
- Propagation: the skeleton dry-run from this checkout ABORTS because HEAD is a feature branch. After merge, run it from main; the script, rule and template entry all sit in propagating paths.

[...]

PR #350, opened 2026-09-14
fix(skill-trigger-eval): refuse a paths-scoped fixture run bare, seed it into a temp copy (ASK-1242)

assafkip:
## What

`skill-trigger-eval.py` ran a bare `claude -p` from the repo root. A rule with a `paths:` frontmatter loads only when a matching file is touched, so for `fable-escalation.json` the rule was absent by construction and `trigger_rate` measured the un-ruled model (sp-c4817fa9, found on PR #298).

## Change

- **Rule derived, not restated.** The harness finds the governing rule in `.claude/rules/` at run time: file stem equals the fixture skill, or the body names `skill-evals/<skill>.json`. Globs come from the frontmatter. `**` and `**/*` count as always-on.
- **Refusal.** A fixture governed by a paths-scoped rule exits 2 before any `claude` call unless it declares `seed_path` matching a glob, or every prompt names a matching path. The message names the rule file and its globs.
- **Seeding.** With `seed_path`, the fixture runs in a temp copy of the tracked tree with that file created, and each prompt names it. The live checkout is not written.
- **`fable-escalation.json`** gains `seed_path: q-system/.q-system/scripts/skill_eval_seed.py` (matches `**/scripts/**`).
- **Rule text** correction goes through the sanctioned path: `q-system/output/claude-changes/ask-1242-fable-escalation-seed-path.json` (one `insert_after`).

## Evidence (ran, not assumed)

- RED on the old script, case 6: `FAIL: unseeded paths-scoped fixture was run, not refused: scoped trigger_rate=0.50 (2 cases)`
- GREEN after: `bash q-system/.q-system/scripts/test/test-skill-trigger-eval.sh` prints `PASS: ... paths-scoped rules refused bare, seeded into a temp copy`
- Mutants, each red: refusal removed (case 6), seed not written (case 9, 0.50), prompt reference dropped (case 9, 0.50). Restored script green.
- Case 6 also asserts a refused fixture spends zero `claude` calls. Case 9 asserts the fake repo stays clean (`git status --porcelain` empty).

## Scope notes

- Test extended in the existing `test/test-skill-trigger-eval.sh` (already in the capability manifest), not a new `test_skill_trigger_eval.py`. Same harness, no new manifest line.
- **Not proven here:** that the proposal applies. `claude-path-write-guard.py` blocks an agent from running `apply-claude-changes.sh` even with `--root` on a copy. The anchor is unique in the rule (grep count 1). One command lands it: `bash q-system/.q-system/scripts/apply-claude-changes.sh q-system/output/claude-changes/ask-1242-fable-escalation-seed-path.json`
- The seed makes the rule loadable, not loaded. Whether the model opens the file stays its own call, so the rate is still advisory.
- sp-c4817fa9 resolves at closeout (`spillover resolve` refuses until ASK-1242 is closed).


assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-350-20260914-071902.md` (7640 bytes).

```
FINDINGS:
minor|The rule-text proposal lacks a per-edit reason, so apply_claude_changes.py load_proposal refuses it and the command the PR body gives cannot land it|q-system/output/claude-changes/ask-1242-fable-escalation-seed-path.json:14
minor|Case 9 passes with copy_tracked_tree removed, so the suite cannot detect a seeded session that has no .claude/rules or CLAUDE.md|q-system/.q-system/scripts/test/test-skill-trigger-eval.sh:90
minor|_paths_value parses a flush YAML list or a commented list as empty, so a paths-scoped rule is treated as unscoped and its fixture runs bare|q-system/.q-system/scripts/skill-trigger-eval.py:74
minor|prompt_names_match accepts a nonexistent path, so a fixture runs bare although the rule cannot load, contradicting the docstring's refusal claim|q-system/.q-system/scripts/skill-trigger-eval.py:114
nit|strip("./") removes the leading dot from .claude/... tokens, falsely refusing a prompt that names a matching dot-directory path|q-system/.q-system/scripts/skill-trigger-eval.py:116
nit|Case 12's comment calls social-reaction-gate a paths-scoped fixture, but scoped_rules returns none for it|q-system/.q-system/scripts/test/test-skill-trigger-eval.sh:102
END FINDINGS
```

--- reviewer output ---

**Verdict: APPROVE WITH NITS.** The harness change is sound. The one broken deliverable is the rule-text proposal: the apply engine refuses it, so the "one command lands it" line in the PR body is false today. Fix that first.

## Findings

**1. minor: the proposal file cannot be applied**
- **Claim:** `apply_claude_changes.py:349-351` requires a non-empty `reason` on every edit. The PR's single edit has none, so the command the PR body hands the founder refuses before it writes anything.
- **Where:** `q-system/output/claude-changes/ask-1242-fable-escalation-seed-path.json:14`
- **Reproducer:** imported `apply_claude_changes.py` and ran every file in `q-system/output/claude-changes/` through `load_proposal`, the loader `main()` calls at line 1113. Nothing was written.
- **Output:**
  - `REFUSED ask-1242-fable-escalation-seed-path.json -> edit 0.reason must be a non-empty string`
  - All 20 sibling proposals printed `LOADS`, so the check discriminates.
- **Why minor, not major:** it fails loudly, only on a manual step, and nothing is left half-written.
- **Fix:** add `"reason": "..."` to the edit object. I simulated `apply_edit` plus `_guard_frontmatter` in memory with a reason added. The anchor is unique, the frontmatter is unchanged, and the inserted text reads cleanly.

**2. minor: the positive test cannot see a seeded session that has no rules**
- **Claim:** case 9's mock only checks that the seed file exists in `$PWD`. It never checks that `.claude/rules` or `CLAUDE.md` exist there. Delete the tree copy and the session runs in a folder holding nothing but the seed, which is the un-ruled model the PR exists to stop, and the suite still passes.
- **Where:** `q-system/.q-system/scripts/test/test-skill-trigger-eval.sh:90`, testing `skill-trigger-eval.py:148`
- **Reproducer:** extracted HEAD with `git archive` into a scratch mirror, replaced `copy_tracked_tree(tmp)` with `pass`, and ran the mirror's test.
- **Output:**
  - `BASELINE exit 0 PASS: ...`
  - `MUTANT no-copy exit 0 PASS: ...`
- **Code today is correct.** A probe mock on the real `fable-escalation` fixture printed `present .claude/rules/fable-escalation.md`, `present CLAUDE.md`, and `present q-system/.q-system/scripts/skill_eval_seed.py`. The gap is only that nothing guards it.
- **Fix:** have `SEEDMOCK` also require `.claude/rules/scoped.md` in `$PWD`.

**3. minor: the paths parser fails open on valid YAML**
- **Claim:** `_paths_value` only accepts indented `- item` lines. A flush list (`paths:` followed by `- "src/**"`) or a comment line inside the list parses as `[]`. The rule is then treated as unscoped and the fixture runs bare, with no refusal.
- **Where:** `q-system/.q-system/scripts/skill-trigger-eval.py:74`
- **Reproducer:** fake repo with those two rule shapes, fixtures with prompt "no path here", and a mock that counts calls.
- **Output:**
  - `rule_paths(flush) = []`
  - `rule_paths(commented) = []`
  - `flush exit=0 calls=1 trigger_rate=1.00`
  - `commented exit=0 calls=1 trigger_rate=1.00`
- **Severity dropped a level:** all 24 real rules with `paths:` parse correctly (`rules with paths: but empty parse = 0`), so no producer emits these shapes today.
- **Fix:** when a `paths:` key is present but parses to nothing, refuse.

**4. minor: the prompt escape hatch accepts a path that does not exist**
- **Claim:** `prompt_names_match` only checks that a prompt token matches a glob. A prompt naming `src/does_not_exist.py` is accepted and run bare from the live repo. That file can never be read, so the rule can never load. The module docstring (line 20) says a scoped fixture without `seed_path` is refused, which the code does not do.
- **Where:** `q-system/.q-system/scripts/skill-trigger-eval.py:114`
- **Output:**
  - `ghost exit=0 calls=1 trigger_rate=1.00`
  - `src/does_not_exist.py exists in fake repo: False`
- **Fix:** require the named path to exist under `REPO_ROOT`, or correct the docstring.

**5. nit: `strip("./")` removes the leading dot from `.claude/...` tokens**
- **Where:** `q-system/.q-system/scripts/skill-trigger-eval.py:116`
- **Output:** a prompt naming `.claude/skills/foo/SKILL.md` against glob `.claude/skills/**` gives `dotted exit=2 calls=0 fixture error: ... case 1 names no matching path`.
- It fails closed: a false refusal, never a bare run.

**6. nit: case 12's comment says "the real paths-scoped fixtures"**
- **Where:** `q-system/.q-system/scripts/test/test-skill-trigger-eval.sh:102`
- **Output:** `social-reaction-gate scoped_rules=[]`. Only `fable-escalation` is paths-scoped.

- **Refusal spends nothing.** `load_all` checks every fixture before any `claude` call. Case 6 counts zero calls, and a no-arg run showed `NEW ... calls 0` against `OLD ... calls 5`.
- **Live checkout stays clean.** The real `fable-escalation` run left the review tree clean. The seed only exists inside `skill-eval-*`.
- **All 24 real rules parse**, and `fable-escalation` resolves to its five globs.
- **Test mutants die.** Removing the refusal, the seed write, or the prompt suffix each turns the suite red. The PR shows all three going red, and I confirmed its baseline passes in 1.9s.
- **Copy is safe.** The repo has no submodules and no tracked symlinks, so `copy2` has no directory or gitlink to trip on. Concurrent runs get separate `TemporaryDirectory` paths.
- **No paging from the seeded copy.** The four session hooks able to alert (`voice-stop-gate.py`, `code_claim_grounding_guard.py`, `token-guard.py`) fire on draft drift or stuck loops, not on untracked state missing from the copy. So a seeded eval does not file Linear tickets. The grounding guard says in its own code that paging is not wired.

## Tried and dropped

- **Transcript litter:** each seeded run may leave a new `~/.claude/projects/<tmp-slug>` folder. Not reproduced, because that needs real billed `claude -p` calls that would fire live hooks.
- **Marker contamination:** the seed sits next to `fable-escalate.py` in `scripts/`, so a model listing that folder might print the marker. Not reproducible without a live model, so dropped.

## Out of scope, already broken on main

A no-argument run exits 2 on `audhd-output-cases.json` (`needs 'skill' and a non-empty 'cases' list`), before and after this PR. The PR only makes that failure happen before any spend.

**Fix first:** add a per-edit `reason` to `ask-1242-fable-escalation-seed-path.json`.

FINDINGS:
minor|The rule-text proposal lacks a per-edit reason, so apply_claude_changes.py load_proposal refuses it and the command the PR body gives cannot land it|q-system/output/claude-changes/ask-1242-fable-escalation-seed-path.json:14
minor|Case 9 passes with copy_tracked_tree removed, so the suite cannot detect a seeded session that has no .claude/rules or CLAUDE.md|q-system/.q-system/scripts/test/test-skill-trigger-eval.sh:90
minor|_paths_value parses a flush YAML list or a commented list as empty, so a paths-scoped rule is treated as unscoped and its fixture runs bare|q-system/.q-system/scripts/skill-trigger-eval.py:74
minor|prompt_names_match accepts a nonexistent path, so a fixture runs bare although the rule cannot load, contradicting the docstring's refusal claim|q-system/.q-system/scripts/skill-trigger-eval.py:114
nit|strip("./") removes the leading dot from .claude/... tokens, falsely refusing a prompt that names a matching dot-directory path|q-system/.q-system/scripts/skill-trigger-eval.py:116
nit|Case 12's comment calls social-reaction-gate a paths-scoped fixture, but scoped_rules returns none for it|q-system/.q-system/scripts/test/test-skill-trigger-eval.sh:102
END FINDINGS

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-350-20260924-082827.md` (6451 bytes).

```
FINDINGS:
minor|rule_paths reads only `paths:` so a rule scoped with `globs:` is treated as unscoped and its fixture runs bare, the defect this PR exists to prevent|q-system/.q-system/scripts/skill-trigger-eval.py:64
minor|The proposal edit has no `reason`, so the loader refuses it and the rule-text correction the PR promises never lands|q-system/output/claude-changes/ask-1242-fable-escalation-seed-path.json:14
minor|prompt_names_match accepts a path that does not exist, so a scoped fixture runs bare although no file is ever touched, contradicting the docstring|q-system/.q-system/scripts/skill-trigger-eval.py:114
minor|_paths_value returns [] for a flush or commented YAML sequence, failing open and running a paths-scoped fixture bare|q-system/.q-system/scripts/skill-trigger-eval.py:74
nit|Case 12's comment calls social-reaction-gate a paths-scoped fixture; scoped_rules returns none for it|q-system/.q-system/scripts/test/test-skill-trigger-eval.sh:102
END FINDINGS
```

--- reviewer output ---

## Round 2 read: the code is unchanged since round 1

PR head is still `b5c59d01`. Nothing was pushed after the round-1 review. I re-reproduced all six round-1 findings and every one is still live. That is the most useful thing I can report: round 2 is being spent re-proving findings nobody acted on, not reviewing new work. Round 1 was calibrated correctly (APPROVE WITH NITS, six minors/nits); the loop broke at the fix step, not at the review step.

I found one thing round 1 did not. Everything else below is a re-raise with my own executed repro, at the same severity.

---

## Findings

### 1. minor (NEW): `rule_paths` only reads `paths:`, so a rule scoped with `globs:` runs its fixture bare

`skill-trigger-eval.py:64` matches `line.startswith("paths:")` and nothing else. The repo's own sibling parser disagrees: `instruction-budget-audit.py:133` matches `^(paths|globs):` with the comment "both are scoping keys in Claude Code". That is a derivation split on the one question this PR exists to answer, and the new half fails open.

Reproducer and real output (fake repo, mock claude that counts calls):

```
=== R1: rule scoped with globs: instead of paths: ===
exit=0  claude calls spent=2
globbed                          trigger_rate=0.50 (2 cases)
ADVISORY: mean trigger_rate 0.50.
```

No refusal, two billed calls, and a number measured on the un-ruled model. That is sp-c4817fa9 repeating through a different key.

Same file, same split, second instance:

```
ALWAYS_ON_GLOBS in skill-trigger-eval.py         : ['**', '**/*']
CATCH_ALL_PATTERNS in instruction-budget-audit.py: ['**', '**/*', '**/**']
```

`**/**` is always-on to the ratchet and paths-scoped to the new gate, so that direction over-refuses instead.

Why minor and not major: no rule in `.claude/rules/` uses `globs:` or `**/**` today. I checked all 24.

Fix that closes both: import the scoping question from `instruction-budget-audit.py` instead of re-deriving it.

### 2. minor (still live, round-1 #4): `prompt_names_match` accepts a path that does not exist

`skill-trigger-eval.py:114`. The module docstring at line 15 says the rule loads only when a matching file is **touched**. The escape hatch only checks that a prompt token matches a glob.

```
=== R2: paths-scoped rule, no seed, prompts name a NON-EXISTENT matching path ===
exit=0  claude calls spent=2
scoped2                          trigger_rate=0.50 (2 cases)
src/does_not_exist.py present in session cwd? False
```

### 3. minor (still live, round-1 #1): the shipped proposal refuses to load, so the rule text never gets corrected

`q-system/output/claude-changes/ask-1242-fable-escalation-seed-path.json:14`. I ran every proposal in that directory through the loader `main()` calls:

```
ask-1242-fable-escalation-seed-path.json REFUSED -> edit 0.reason must be a non-empty string  <<< THIS PR
```

Twenty sibling proposals printed `LOADS`, so the check discriminates. The PR body's "One command lands it" is false today, and `fable-escalation.md` still says nothing about the seed contract.

### 4. minor (still live, round-1 #3): `_paths_value` fails open on two valid YAML shapes

`skill-trigger-eval.py:74`.

```
flush list         rule_paths -> []
commented list     rule_paths -> []
control            rule_paths -> ['src/**']
```

A `paths:` key present but parsing to nothing should refuse, not run bare.

### 5. nit (still live, round-1 #6): case 12's comment misnames its own subject

`test-skill-trigger-eval.sh:102` calls both fixtures "the real paths-scoped fixtures".

```
social-reaction-gate scoped_rules -> []
```

Round-1 nit #5 (`strip("./")` eating the dot on `.claude/...` tokens) also reproduced as still live. It fails closed, so I am leaving it out of the block rather than adding noise.

---

## What is sound

- **The parser is right on the real population.** All 24 rules carrying `paths:` parse to their correct glob lists, and `fable-escalation` resolves to its five. Nothing false-refuses.
- **Refusal genuinely spends nothing.** Measured both versions on the same fixture dir: `BEFORE exit=2 claude calls spent=5`, `AFTER exit=2 claude calls spent=0`. That is a real improvement, not a claim.
- **The live checkout stays clean.** Full suite green in 3.5 seconds, including a real 2,624-file, 33 MB tracked-tree copy for the `fable-escalation` case.
- **The copy has nothing to trip on.** `git ls-files -s` shows zero symlinks and zero gitlinks, so `copy2(follow_symlinks=False)` is safe here.
- **No collateral refusals.** Of the seven fixtures, only `fable-escalation` is governed by a paths-scoped rule. The other five that load all pass `check_scoped`.
- **Concurrency holds.** Each run gets its own `TemporaryDirectory`; two evals cannot collide.
- **The broken no-arg run is not this PR's.** `audhd-output-cases.json` is a top-level list, so the documented "no args = every fixture" mode exits 2 before and after. The PR only moves that failure ahead of the spend.

---

## VERDICT: APPROVE WITH NITS

Only minor and nit findings survived reproduction. No blocker, no major.

**Fix first:** add a non-empty `reason` to the single edit in `ask-1242-fable-escalation-seed-path.json`. It is a one-line change, it is the only part of this PR that does not work at all, and it has now been named in two consecutive rounds.

FINDINGS:
minor|rule_paths reads only `paths:` so a rule scoped with `globs:` is treated as unscoped and its fixture runs bare, the defect this PR exists to prevent|q-system/.q-system/scripts/skill-trigger-eval.py:64
minor|The proposal edit has no `reason`, so the loader refuses it and the rule-text correction the PR promises never lands|q-system/output/claude-changes/ask-1242-fable-escalation-seed-path.json:14
minor|prompt_names_match accepts a path that does not exist, so a scoped fixture runs bare although no file is ever touched, contradicting the docstring|q-system/.q-system/scripts/skill-trigger-eval.py:114
minor|_paths_value returns [] for a flush or commented YAML sequence, failing open and running a paths-scoped fixture bare|q-system/.q-system/scripts/skill-trigger-eval.py:74
nit|Case 12's comment calls social-reaction-gate a paths-scoped fixture; scoped_rules returns none for it|q-system/.q-system/scripts/test/test-skill-trigger-eval.sh:102
END FINDINGS
SessionEnd hook [python3 "${CLAUDE_PLUGIN_ROOT:-$PLUGIN_ROOT}/telemetry/hooks/session-end-llma.py"] failed: Hook cancelled

PR #351, opened 2026-09-14
fix(voiceloop-migrate): a cache-only dir is not a package; the mapping table is not rewritten (ASK-1239)

assafkip:
ASK-1239, promoted from spillover sp-273637ee.

## What was broken
- `plan()` decided package state with `os.path.isdir`. Consulting's `plugins/kipi-core/voiceloop/` held only `__pycache__` and `.pytest_cache`, so it read as present. Plan said `both_present`, skipped the move, rewrote every import to a package with no code, and the instance's `verify.sh` failed on `ModuleNotFoundError: voiceloop` (2026-09-03, 2026-09-04).
- The token swap rewrote `automation/export_voice_loop.py`'s RENAMES pair `("voicekit", "voiceloop")` into `("voiceloop", "voiceloop")`, which broke the public-mirror transform.

## The fix (one file: `kipi-update-voiceloop-migrate.py`)
- `_is_package()`: a dir counts only if it holds at least one `.py` file. Cache dirs are pruned by the same `_skip_dirname` the source walk uses.
- `apply()` step 1: when the plan is `move` and a non-package shell sits at the new name, it is renamed aside to `voiceloop.pre-voiceloop.bak`, never deleted. Without this, `git mv` onto the existing dir nests the package INSIDE it (`voiceloop/voicekit/`), and `os.rename` refuses a non-empty dir.
- `defines_mapping()`: a file carrying the quoted bare pair is skipped whole and reported under `defines_mapping` in the plan JSON. Path pairs like `("voicekit/validate.py", ...)` still migrate. An existing test pins that.

## Evidence
- RED first, against the unfixed script: 3 of the 4 new tests failed. One was the byte diff `("voicekit", "voiceloop")` -> `("voiceloop", "voiceloop")`. One was plan `both_present` != `move`. The negative control (a real new package stays `both_present`) passed.
- GREEN: `python3 -m pytest q-system/.q-system/scripts/test_voiceloop_migrate.py q-system/.q-system/tests/separation/test_update_propagation.py -q` gives 49 passed.
- Mutation, run against temp copies of the script:
  - Restoring `isdir` kills the 2 cache-shell tests.
  - Removing the mapping skip kills the mapping test.
  - Removing the set-aside kills the 2 cache-shell tests.
  - The unmutated copy has the same 5 reds in every run. Those 5 depend on the script's real location (`SKELETON_ROOT`), not on this fix.
- One existing fixture changed. `test_both_present_never_deletes_either_package` built an EMPTY `voiceloop/` and asserted `both_present`, which pinned this exact defect. It now holds a real `__init__.py`, so what it checks is unchanged: both packages survive.

## Not done here (per the DoR)
- Acceptance 4 (report mode on consulting prints `package=already`): waits on the sibling port. This PR does not touch consulting.
- Acceptance 5 (resolve sp-273637ee): `spillover resolve` refuses until ASK-1239 is closed, so it runs at closeout.

## Found and captured, not fixed
- sp-64682fb1 / ASK-1676: the pre-commit `verify` step lost an `index.lock` race once (exit 128). The same commit passed on an immediate retry.


assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-351-20260914-082140.md` (5948 bytes).

```
FINDINGS:
minor|defines_mapping skips the whole file, so any other old-package import or path in a file carrying the quoted pair (even in a comment) stays stale while verified=True, needs_work=False, and the text plan line is silent; the new test pins a stale import as correct|kipi-update-voiceloop-migrate.py:535
minor|pre-existing: the git mv failure return never sets verified, so main() crashes with KeyError 'verified' and the new set-aside leaves the shell at .bak on that path|kipi-update-voiceloop-migrate.py:721
nit|the set-aside .bak sits inside plugins/kipi-core, which the updater's rsync --delete --delete-excluded removes in the same run, so "never deleted / its bytes survive" holds only inside apply()|kipi-update-voiceloop-migrate.py:704
END FINDINGS
```

--- reviewer output ---

**Verdict: APPROVE WITH NITS.** The fix works for the shape it targets. The new mapping skip swaps one stranded reference for another, and nothing reports it.

## Findings

### 1. MINOR: the mapping skip leaves the whole file unmigrated, and still reports verified

- **Claim:** `defines_mapping()` matches the quoted pair anywhere in the raw text, then `continue`s the whole file. Every other old-package reference in that file stays: an import, a source path, anything. After that, `verified` is True, `needs_work` is False, and the text line `kipi-update.sh` prints says nothing.
- **Where:** `kipi-update-voiceloop-migrate.py:535`. Also the new test at `test_voiceloop_migrate.py`, in `test_a_file_that_defines_the_mapping_is_left_byte_identical`. Its fixture has `import voicekit.core` in the table file and asserts the import survives with `verified` True. The test pins the defect, the same way the old `both_present` fixture pinned the one this PR fixes.
- **Second shape:** a comment that quotes the pair also skips that file's real imports. That contradicts the comment at lines 356-357.
- **Severity call:** I could not confirm a producer, because consulting is outside the sandbox. It becomes MAJOR if consulting's `automation/export_voice_loop.py` has any `voicekit` besides the RENAMES line, for example a path to the private engine dir. One grep there settles it.

Reproducer A, a stale source path. I ran this, trimmed to the lines that matter:
```
python3 -c '... m.apply(r, commit=False) on an instance holding
  automation/export_voice_loop.py = SRC = ROOT / "plugins/kipi-core/voicekit" + RENAMES = (("voicekit", "voiceloop"),) ...'
```
Real output:
```
verified = True errors = []
defines_mapping = ['automation/export_voice_loop.py']
voicekit dir exists after move = False
exporter SRC line = SRC = ROOT / "plugins/kipi-core/voicekit"   # the private engine dir
after: needs_work = False rewrite = []
text plan line: tmpix3kich6: package=already rewrite=0 rename=0 unparseable=0 history_left=0 needs_work=False
```

Reproducer B, a prose pair that shields a real import. Real output:
```
verified = True defines_mapping = ['pipeline/voice.py']
# history: the exporter maps ("voicekit", "voiceloop") for the mirror
from voicekit import V
```

- **Fix shape:** protect only the matched pair span. `_swap_outside` already takes spans, so add the pair's span next to the prose spans and let the rest of the file migrate. Also add `defines_mapping` to the non-JSON text line, because the updater never passes `--json`.

### 2. MINOR (pre-existing, next to the new code): a `git mv` failure crashes with KeyError

- **Claim:** the `git mv` failure return at line 721 never sets `verified`, so `main()` dies with `KeyError: 'verified'`. With this PR, that path also leaves the shell at `.bak`. The exit code is still 1, so the updater still abandons the instance correctly. The operator gets a traceback instead of the ERROR line.
- **How reachable:** the index.lock race this PR body itself reports (sp-64682fb1) is enough.
- **Where:** `kipi-update-voiceloop-migrate.py:721`

Reproducer: a git instance plus a cache shell, with `.git/index.lock` pre-created, then run `--apply`. Real output:
```
exit 1
KeyError: 'verified'
left on disk: ['voicekit', 'voiceloop.pre-voiceloop.bak']
```
A second run after the lock clears recovers cleanly.

### 3. NIT: "set aside, never deleted" only holds inside `apply()`

- **Claim:** the `.bak` lands inside `plugins/kipi-core/`. The updater's `rsync -a --delete --delete-excluded` wipes it in the same run, so "its bytes survive" in the test docstring holds only until the caller's next step. This is not a regression: the old code left the same bytes at `voiceloop/` for the same rsync to delete.
- **Where:** `kipi-update-voiceloop-migrate.py:704`

Real output:
```
set_aside = plugins/kipi-core/voiceloop.pre-voiceloop.bak verified = True
after apply: ['voiceloop', 'voiceloop.pre-voiceloop.bak']
after the updater plugins rsync: ['voiceloop']
```

## What is sound

- **Tests go red on the old script:** I ran the 4 new tests against `HEAD~1`'s script, loaded with the real `__file__` so `SKELETON_ROOT` matched. Three failed, with `'both_present' != 'move'`, the nested-path assert, and the identity-pair byte diff. That matches the PR's claim.
- **The set-aside step is load-bearing:** with it removed, one test errored with `OSError: Directory not empty` and one failed on `plugins/kipi-core/voiceloop/voicekit/__init__.py`. So the step does real work and the tests can see it.
- **Suite is green:** `test_voiceloop_migrate.py` gave 43 passed.
- **`_is_package`:** it prunes caches with the same `_skip_dirname` the source walk uses. A single `.py` still gives `both_present`, which the negative control checks. A cache-only old dir now reads `absent` or `already` instead of moving an empty shell.
- **Existing backup:** if the `.bak` path already exists, the script refuses loudly with `verified` False and overwrites nothing.

**Fix first:** finding 1. Narrow the skip to the pair's own span and delete the `import voicekit.core` line from the new test's fixture (or flip that assert so the import must migrate).

FINDINGS:
minor|defines_mapping skips the whole file, so any other old-package import or path in a file carrying the quoted pair (even in a comment) stays stale while verified=True, needs_work=False, and the text plan line is silent; the new test pins a stale import as correct|kipi-update-voiceloop-migrate.py:535
minor|pre-existing: the git mv failure return never sets verified, so main() crashes with KeyError 'verified' and the new set-aside leaves the shell at .bak on that path|kipi-update-voiceloop-migrate.py:721
nit|the set-aside .bak sits inside plugins/kipi-core, which the updater's rsync --delete --delete-excluded removes in the same run, so "never deleted / its bytes survive" holds only inside apply()|kipi-update-voiceloop-migrate.py:704
END FINDINGS

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-351-20260924-081211.md` (7398 bytes).

```
FINDINGS:
minor|defines_mapping skips the whole file, so any other old-package import or path in a file carrying the quoted pair stays stale while verified=True and needs_work=False; the new test pins the stale import as correct|kipi-update-voiceloop-migrate.py:535
minor|the git mv failure return never sets verified, so main() dies with KeyError 'verified' and the operator gets a traceback instead of the ERROR line|kipi-update-voiceloop-migrate.py:721
minor|the new set-aside backs up a DIRECTORY, whose files keep their own extensions and are rewritten by step 2, falsifying the BACKUP_SUFFIX invariant stated at line 106|kipi-update-voiceloop-migrate.py:704
nit|the set-aside .bak sits inside plugins/kipi-core, which the updater's own rsync --delete --delete-excluded removes in the same run|kipi-update-voiceloop-migrate.py:704
END FINDINGS
```

--- reviewer output ---

## Round 2 read: the code is byte-identical to round 1

`gh pr view 351 --json commits`: the only commits since round 1's review of `9ccbd90` are a receipt row and a merge of `origin/main`. No source change. So I ran my own reproducers against all three round-1 findings; all three are still live, at the same severity, plus one new one.

---

### 1. MINOR (still live, re-raised with my own repro): `defines_mapping` skips the whole file, imports strand, `verified=True`

`kipi-update-voiceloop-migrate.py:535`. The regex matches the quoted pair anywhere in raw text, then `continue`s the entire file. Every other reference in it survives.

Repro, two shapes, real output:

```
CASE A: file defines the mapping AND imports the package
  package moved          : True
  verified               : True []
  defines_mapping        : ['automation/export_voice_loop.py']
  plan(after).needs_work : False
  still imports OLD      : True
  OLD pkg on disk        : False
  NEW pkg on disk        : True
  runtime import         : ModuleNotFoundError: No module named 'voicekit'

CASE B: a plain LIST of both names, no mapping anywhere
  defines_mapping        : ['q-consult/pipeline/voice.py']
  rewritten              : []
  verified               : True []
  plan(after).needs_work : False
  still imports OLD      : True
```

Case A is the exact shape the PR's own new test encodes at `test_voiceloop_migrate.py:747`: it puts `import voicekit.core` in the table file and asserts the file comes back byte-identical with `verified` True. The test pins the stale import as correct.

Severity held at MINOR, same as round 1. I tried the grep round 1 deferred to (`~/projects/consulting/automation/export_voice_loop.py`) and the sandbox blocked it too, so the producer is still unconfirmed and the no-escalation rule applies.

One calibration note in the code's favour: I scanned the whole skeleton with the real `defines_mapping`. Exactly one file matches, and it is the migrate script itself, already `SELF_PATH`-exempt. No producer in this repo emits the Case B shape.

---

### 2. MINOR (still live): `git mv` failure exits with a traceback, not the ERROR line

`kipi-update-voiceloop-migrate.py:721`. That return path never sets `result["verified"]`, so `main()` raises. Repro used the `index.lock` race the PR body itself reports as sp-64682fb1:

```
exit: 1
stderr tail: ['    f"committed={out[\'committed\']} verified={out[\'verified\']}")',
              '                                             ~~~^^^^^^^^^^^^',
              "KeyError: 'verified'"]
left under plugins/kipi-core: ['voicekit', 'voiceloop.pre-voiceloop.bak']
```

Exit 1 is still correct, so `abandon_instance` fires and the `--delete` rsync does not run. The operator gets a stack trace at 3am instead of the ERROR line the script writes everywhere else. A re-run recovers: `voiceloop/` is gone, so the set-aside is skipped and the move proceeds.

---

### 3. MINOR (new this round): the set-aside directory's contents are rewritten, which breaks the backup invariant the same file states

`kipi-update-voiceloop-migrate.py:106` says `.bak` is kept out of `REWRITE_EXTS` so "a backup this script wrote must never become a file a later run rewrites." That works for a FILE backup (`foo.py.pre-voiceloop.bak`). The new set-aside at line 704 backs up a DIRECTORY, and the files inside keep their own extensions, so step 2's re-plan picks them up.

```
set_aside : plugins/kipi-core/voiceloop.pre-voiceloop.bak
rewritten : ['plugins/kipi-core/voiceloop.pre-voiceloop.bak/voice-channels.json']
before    : {"engine": "plugins/kipi-core/voicekit/validate.py"}
after     : {"engine": "plugins/kipi-core/voiceloop/validate.py"}
backup was edited: True
```

New because the directory backup is new in this PR. Bounded: in the shape actually observed on consulting the shell holds only caches, and the `.bak` is deleted by the caller anyway (finding 4).

---

### 4. NIT (still live): "set aside, never deleted" holds only inside `apply()`

`kipi-update-voiceloop-migrate.py:704`. The `.bak` lands inside `plugins/kipi-core/`, which the updater rsyncs at `kipi-update.sh:2952`. I ran that rsync with its own flags:

```
after apply : ['voiceloop', 'voiceloop.pre-voiceloop.bak']
after rsync : ['voiceloop']
bak survives: False
```

---

## The process finding, which is worth more than another nit

Round 1 returned APPROVE WITH NITS on `9ccbd90`. Auto-merge is enabled. Ten days later the PR is open, the source is unchanged, and round 2 is re-deriving the same three findings. Nothing in the review rounds is blocking this. Another round on unchanged code buys nothing; either the nits get fixed or the PR merges.

---

- **Set-aside with TRACKED shell content.** I built a `voiceloop/` holding tracked `channels.json` and `README.md`, ran `apply(commit=True)`, and got `committed: True`, `git status` clean, and both files tracked at their `.bak` path. Bytes survive in history.
- **Permanent-dirt follow-on.** I chased whether the rsync deleting those tracked `.bak` files leaves the instance dirty forever. `stage_config_sync` at `kipi-update.sh:1164` does `git add -u -- plugins/`, so the deletion is staged and committed. No repeated refusal. Dropped.
- **`_is_package` false negatives.** I looked for a real package that reads as a shell. `SKIP_DIRS` prunes `__pycache__`, `.pytest_cache`, `.venv`, `target`, `.claude-plugin`; `PLUGIN_COPY_EXCLUDES` at `kipi-update.sh:244` is exactly the cache set, so no rsync can strip a package's `.py` and make it read source-less. Could not construct a reachable case. Dropped.
- **The "aside already exists" refusal.** Sets `verified=False`, moves nothing, exits 1. Overwrites no backup.
- **Suite.** `python3 -m pytest q-system/.q-system/scripts/test_voiceloop_migrate.py -q` gives `43 passed in 5.41s`.
- **The negative control is real.** `test_a_real_new_package_beside_the_old_is_still_both_present` goes red if `_is_package` is made unconditionally strict, so it is not decoration.

---

## VERDICT: APPROVE WITH NITS

Only minor and nit findings survived reproduction.

**Fix first:** finding 1. Narrow the skip to the matched pair's own span rather than the whole file, and flip the assert in `test_a_file_that_defines_the_mapping_is_left_byte_identical` so `import voicekit.core` must migrate. Right now that test makes the stranded import the specified behavior.

FINDINGS:
minor|defines_mapping skips the whole file, so any other old-package import or path in a file carrying the quoted pair stays stale while verified=True and needs_work=False; the new test pins the stale import as correct|kipi-update-voiceloop-migrate.py:535
minor|the git mv failure return never sets verified, so main() dies with KeyError 'verified' and the operator gets a traceback instead of the ERROR line|kipi-update-voiceloop-migrate.py:721
minor|the new set-aside backs up a DIRECTORY, whose files keep their own extensions and are rewritten by step 2, falsifying the BACKUP_SUFFIX invariant stated at line 106|kipi-update-voiceloop-migrate.py:704
nit|the set-aside .bak sits inside plugins/kipi-core, which the updater's own rsync --delete --delete-excluded removes in the same run|kipi-update-voiceloop-migrate.py:704
END FINDINGS
SessionEnd hook [python3 "${CLAUDE_PLUGIN_ROOT:-$PLUGIN_ROOT}/telemetry/hooks/session-end-llma.py"] failed: Hook cancelled

PR #355, opened 2026-09-14
fleet-health: watch default-branch CI on every registered repo (ASK-1173)

assafkip:
## ASK-1173: fleet-health-daily now watches default-branch CI

**The finding.** `assafkip/ASK_AI_consultant` main has failed every CI run on record: `verify` twice on 2026-08-27, then the `validate` workflow that replaced it, 10 for 10 red from the day it was wired (2026-09-10). GitHub reported each run into a notification inbox, and nothing on the board saw it.

## Acceptance criteria

**1. The failing job's cause, named.** Latest run 34730585539 (2026-09-13): 233 failed, 6 errors, 8886 passed. The suite assumes the founder's Mac, and the GitHub runner is not that machine:
- 78: `voiceloop engine not found on PATH or in the fleet venv`
- 67: `ImportError: attempted relative import with no known parent package`, a test-invocation layout the runner does not reproduce
- 27: `/bin/zsh` missing on ubuntu-latest
- 6: `kipi_mcp.linter not found in ~/.claude/plugins/marketplace...`
- plus hardcoded `/Users/assafkipnis/...` paths, a missing `PIL`, and 6 corpus tests that yield nothing without the local repos

**2. Decision recorded: fix, not retire.** The workflow gates a repo that production runs from, so it earns its run. The work belongs to that repo: split the machine-bound tests behind a marker, or install what they need. Once this merges, its first `--apply` run files that repo's standing issue. That issue is where the fix is tracked, so this PR files no duplicate by hand.

**3. The watcher (the class fix).** New detector `default-branch-ci-red` in `fleet-health-daily.py`:
- Reads every GitHub repo in `instance-registry.json`: the skeleton remote, `dispatch.expected_remote`, or the instance's `origin`.
- For each ACTIVE workflow, the latest decisive run on the default branch decides (failure, timed_out and startup_failure are red; cancelled and skipped decide nothing). A deleted workflow's old failures still sit in the repo-wide run list, and this detector never reads that list.
- Skips `dynamic/` workflows. GitHub manages those (Dependabot, default CodeQL), so no file in the repo can fix one.
- Files one standing issue per red repo. The body carries the red-since date and never a run count, so another red push does not rewrite the issue.
- A repo it cannot read becomes one rollup finding (`default-branch-ci-unreadable`), never silence. A missing `gh` is a finding too.
- Resolves `gh` to an absolute path, because `com.kipi.fleet-health.plist` sets no PATH and the launchd default has no homebrew.

## Evidence

- Reproducer first: the new checks ran RED against the unchanged detector (`fleet-health-daily.py has no default_branch_ci_findings`), then GREEN: `python3 q-system/.q-system/scripts/test/test-fleet-health-daily.py` gives `PASS: fleet-health-daily contract holds`.
- The fixtures are real `gh api` output recorded 2026-09-14, trimmed to the fields the detector reads.
- Mutation: 6 in-memory mutants, 6 killed. They drop the active filter, drop the dynamic exclusion, trust API order, treat cancelled as decisive, drop unreadable repos, and put a run count in the body.
- Live calibration, read-only: 24 repos read in 42s, 0 unreadable. 4 were red before the dynamic exclusion: ASK_AI_consultant (validate, since 2026-09-10), cole-gtm (gtm-build, since 2026-07-24), ktlyst-saas-product (prd-gates, since 2026-07-07), and kipi-investigations (`dynamic/dependabot/update-graph`, now excluded). So the first `--apply` run files 3 issues, and each is real.

## Cost to know about

`test-fleet-health-daily.py` already calls every shipped detector live (lines 108-121), so it now reads GitHub too. That adds about 40s locally. On a runner with no GitHub auth, the detector returns the unreadable rollup and never raises.

## Not doing

- No edits in `ASK_AI_consultant`. That build is its own repo's work (see criterion 2).
- No edit to `com.kipi.fleet-health.plist`. It is outside the allowed files, and the PATH gap is handled in code.
- kipi-investigations' failing Dependency Graph job: rejected as a minor. GitHub manages it, it only affects how fresh the dependency graph is, and no repo-owned fix exists.

Resolve `sp-738cccc2` with `--resolution-ref ASK-1173` once this closes.


assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-355-20260914-123236.md` (6376 bytes).

```
FINDINGS:
major|registered_github_repos() reads instance-registry.json unguarded, so the live detector loop in test-fleet-health-daily.py raises FileNotFoundError and the test goes red in all 24 subtree instances, failing kipi check and the full capability gate there|q-system/.q-system/scripts/fleet-health-daily.py:1687
minor|No test pins the window-saturation branch; deleting it leaves the suite green, and without it a 30+ run red streak rewrites the Linear issue body on every push|q-system/.q-system/scripts/fleet-health-daily.py:1743
minor|The promise that GhReadError.reason is never gh stderr (it lands in a permanent Linear issue) has no test; replacing it with res.stderr leaves the suite green|q-system/.q-system/scripts/fleet-health-daily.py:1719
END FINDINGS
```

--- reviewer output ---

**Verdict: REQUEST CHANGES.** One major: the new detector makes `test-fleet-health-daily.py` fail in all 24 subtree instances, because it reads a registry file only the skeleton has.

The detector logic itself is solid. It held up against every attack on the CI-reading path, and its live run matches the PR's calibration (3 real red repos). The problem is where the test runs.

## Finding 1: MAJOR. The test fails in every instance

**Claim:** `registered_github_repos()` reads `REPO_ROOT/instance-registry.json` with no guard. The existing test calls every shipped detector live and fails on any exception. In an instance there is no root registry, so the detector raises `FileNotFoundError` and the test exits 1.

- **Where:** `q-system/.q-system/scripts/fleet-health-daily.py:1687`, hit by the live detector loop at `q-system/.q-system/scripts/test/test-fleet-health-daily.py:111`.
- **Who runs it in instances:** the test is declared in `q-system/.q-system/capability/expected_tests/` and is not in `skeleton_only`. So `kipi check` run from an instance (`kipi:284`, `|| exit 1`) and `fleet-capability-verify.py` in full mode both run it there.
- **Why that matters:** `kipi check` exits before `validate-separation.py` runs, so the rest of the instance check is skipped too. `fleet-capability-verify.py`'s own docstring names this scar: "a skeleton-only test shipped to the whole fleet and crashed in every instance for months".
- **Fix options:** have the detector return the `default-branch-ci-unreadable` finding when the registry is missing, or add a `skeleton_only` fragment for this test. My call: the detector fix, because a missing registry then shows up as a finding instead of an error.

**Reproducer.** I extracted `q-system/` from PR HEAD into `$TMPDIR/pr355/inst` via `git archive` (this is what an instance looks like after sync), then ran `python3 q-system/.q-system/scripts/test/test-fleet-health-daily.py` there. Real output:

```
exit code: 1
FAIL:
  - detector default-branch-ci-red raised: [Errno 2] No such file or directory: '/private/var/folders/th/ty0kb7r56dvfrkgkrcbx9fv00000gn/T/pr355/inst/instance-registry.json'
```

**Control:** same extraction from the pre-PR base `b67f7d9e`:

```
BASE b67f7d9e instance-shaped exit code: 0
['PASS: fleet-health-daily contract holds']
```

**Scope check against the live fleet:**

```
subtree instances: 24 | with root registry: [] | carrying the test file today: 24
```

The PR body says that without GitHub auth the detector "never raises". That holds in the skeleton only.

## Finding 2: MINOR. The window-saturation branch has no test

**Claim:** removing the saturation branch at `fleet-health-daily.py:1743` leaves the suite green. The code is correct today. But without that branch, once a workflow has 30 or more red runs in a row, the "red since" date moves forward on every push. The body hash changes, so Linear gets rewritten and Slack sends an "updated" line every morning. That is the churn the branch exists to prevent, and no test guards it.

**Reproducer:** a line-swap mutation run against a temp copy with a skeleton-only registry. Real output:

```
none | exit 0 | []
window-saturation branch removed | exit 0 | []
```

## Finding 3: MINOR. The "never gh's stderr" rule has no test

**Claim:** the `GhReadError` docstring at `fleet-health-daily.py:1655` promises the reason is never gh's stderr. That reason goes into the permanent `default-branch-ci-unreadable` Linear issue, and ASK-204 says permanent records carry no untrusted text. Replacing line 1719 with `raise GhReadError(res.stderr)` leaves the suite green. The code is correct today; nothing pins it.

**Reproducer:** same mutation harness. Real output:

```
gh stderr leaked into reason | exit 0 | []
```

## What held up

- **Deleted workflow fixture is real.** Live `gh api repos/assafkip/ASK_AI_consultant/actions/workflows` returns only `validate` (id 355212679). The repo-wide run list still holds 2 runs of `verify.yml` (workflow id 343900464). The PR's fixture matches that exactly.
- **Live run matches the PR's claims.** 35s. Findings for ASK_AI_consultant (since 2026-09-10), cole-gtm (since 2026-07-24) and ktlyst-saas-product (since 2026-07-07). No unreadable repos.
- **Two new tests are real guards.** Using the latest red run's date instead of the streak start fails two checks, including "byte-identical". Dropping the `dynamic/` exclusion fails the dynamic check.
- **Dedup key collisions.** The real registry gives 24 repos, with no collisions after `slug()` lowercasing and no case variants. The "same key twice in one run files two issues" path has no producer today, so I dropped it.
- **Two other angles, checked live and dropped.** No registered repo has more than 4 workflows, so the unpaginated workflow list is safe. No repo has any `pull_request` runs on `main`, so the fork-PR false-red case has no producer.
- **Secrets out of the permanent record.** Credentials in an origin URL never reach the slug (the regex starts after `github.com`). A run count is kept out of the body, so extra red pushes don't rewrite the issue.
- **Launchd PATH.** `gh` is resolved to an absolute path, and a missing `gh` becomes an unreadable finding instead of silence.
- **Other importer.** `launchd-health-check.py` imports this module but never runs `DETECTORS`, so it makes no new GitHub calls.

**Fix first:** make `registered_github_repos()` (or `detect_default_branch_ci`) handle a missing `instance-registry.json` with the unreadable finding instead of raising. Then add a test that runs the detector against a missing registry path.

FINDINGS:
major|registered_github_repos() reads instance-registry.json unguarded, so the live detector loop in test-fleet-health-daily.py raises FileNotFoundError and the test goes red in all 24 subtree instances, failing kipi check and the full capability gate there|q-system/.q-system/scripts/fleet-health-daily.py:1687
minor|No test pins the window-saturation branch; deleting it leaves the suite green, and without it a 30+ run red streak rewrites the Linear issue body on every push|q-system/.q-system/scripts/fleet-health-daily.py:1743
minor|The promise that GhReadError.reason is never gh stderr (it lands in a permanent Linear issue) has no test; replacing it with res.stderr leaves the suite green|q-system/.q-system/scripts/fleet-health-daily.py:1719
END FINDINGS

assafkip:
Review round 1 addressed in ecd85dfc (all 3 findings fixed, reply also on ASK-1173).

- **major, fleet-health-daily.py:1687:** `detect_default_branch_ci` now returns no findings when `instance-registry.json` is absent. Only the skeleton root holds it, so absence means this copy is an instance. It is decided before the gh check. Reproduced first in an instance-shaped copy, where `run_detectors` recorded the detector as `error`.
- **minor, :1743 window saturation:** new checks. A 30-run red window says "at least the last 30" with no date, and one more red push leaves `finding_hash` unchanged. The fixture has one run per day, so the hash check fails on its own.
- **minor, :1719 reason never stderr:** a fake `gh` that writes a token to stderr proves the reason is `HTTP 404` or `gh exited 1`.

Evidence: `python3 q-system/.q-system/scripts/test/test-fleet-health-daily.py` was RED before the fix (`got 'raised FileNotFoundError', want []`). After the fix: 154 ok, `PASS: fleet-health-daily contract holds`. Mutation on a copy of the scripts dir: 4 mutants (drop the guard, drop the saturation branch, date the saturated window, reason = stderr), 4 killed, baseline green.

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-355-20260914-125101.md` (6157 bytes).

```
FINDINGS:
minor|_origin_url returns "" on any git failure, so registered_github_repos silently drops 21 of 24 repos (including both currently-red ones) with no unreadable rollup, printing "could not look" as "nothing found"|q-system/.q-system/scripts/fleet-health-daily.py:1678
minor|Repo dedup is case-sensitive while finding_key lowercases, so two case-variant remotes for one repo yield two findings with the same kipi-key and file_findings creates two permanent Linear issues in one run|q-system/.q-system/scripts/fleet-health-daily.py:1696
nit|Saturation branch counts cancelled runs toward the window, so one failure plus 29 cancelled runs is reported as "red for at least the last 30 completed runs" in a permanent issue body|q-system/.q-system/scripts/fleet-health-daily.py:1743
END FINDINGS
```

--- reviewer output ---

**Verdict: APPROVE WITH NITS.** Three findings, all reproduced, none blocker or major. The one to fix first: if `git` fails, the watcher quietly stops watching 21 of its 24 repos.

## Finding 1: minor. A git failure drops 21 of 24 repos with no finding

- **Claim:** `_origin_url` returns an empty string on any failure (OSError, timeout, non-zero exit). `registered_github_repos` then treats that exactly like a repo with no remote, so the repo drops out and no `default-branch-ci-unreadable` rollup is filed.
- **Where:** `q-system/.q-system/scripts/fleet-health-daily.py:1678` (and 1695, where it gets used).
- **Why it matters:** 23 of 25 registry instances resolve through `origin`, and only 2 have `dispatch.expected_remote`. The PR says an unreadable repo is "never silence", but here it is.
- **Reproducer:** load the module, resolve `gh`, set `PATH` to `/nonexistent` inside the process so `git` cannot start (the same state as a broken Xcode CLT shim after a macOS update), then call `registered_github_repos()` and `default_branch_ci_findings()`.
- **Real output:**
  - `repos watched with git working: 24`
  - `repos watched with git failing: 3 ['assafkip/kipi-interview-coach', 'assafkip/kipi-system', 'assafkip/ktlyst-saas-product']`
  - `findings: ['assafkip/ktlyst-saas-product']`
  - `unreadable rollup present: False`
- ASK_AI_consultant and cole-gtm are red today, and both disappear without a trace. I rated this major and dropped it to minor because it needs git to be broken first.
- **Swap:** have `_origin_url` tell "no origin" (exit 2, "No such remote") apart from "could not ask", and send the second case into the unreadable rollup.

## Finding 2: minor. Two spellings of one repo file two permanent issues

- **Claim:** the repo set is deduplicated case-sensitively, but `finding_key` lowercases the subject. Two remotes that differ only in letter case give two findings with the same kipi-key. `file_findings` computes `known` once before its loop, so both get created.
- **Where:** `q-system/.q-system/scripts/fleet-health-daily.py:1696`
- **Reproducer:** a temp registry with `https://github.com/assafkip/ASK_AI_consultant.git` and `git@github.com:assafkip/ask_ai_consultant.git`, a fake gh that says red, and a fake Linear module passed to `file_findings(apply=True)`.
- **Real output:**
  - `keys: ['fleet-health/default-branch-ci-red/assafkip-ask-ai-consultant', 'fleet-health/default-branch-ci-red/assafkip-ask-ai-consultant', ...]`
  - `filed ASK-1  default branch CI is red: assafkip/ASK_AI_consultant (main)`
  - `filed ASK-2  default branch CI is red: assafkip/ask_ai_consultant (main)`
- Duplicate permanent issues is the major anchor. I dropped it to minor because today's registry has no case collisions (I checked all 24 live repos).
- **Swap:** dedupe on `slug_.lower()`.

## Finding 3: nit. The saturated-window text overstates when cancelled runs fill the window

- **Claim:** the "at least the last 30" branch counts all runs, including cancelled ones. So one failure followed by 29 cancelled runs prints a claim that is false, into a permanent issue body.
- **Where:** `q-system/.q-system/scripts/fleet-health-daily.py:1743`
- **Reproducer:** `_red_streak` fed 1 `failure` plus 29 `cancelled` runs.
- **Real output:**
  - `runs in window: 30 failures: 1`
  - `_red_streak: (True, None)`
  - ``- `.github/workflows/ci.yml`: red for at least the last 30 completed runs``
- No fleet workflow currently shows cancelled runs in its window, so this is a nit. **Swap:** word it "no green run in the last 30 completed runs".

- **Live calibration:** I ran the production detector read-only. It read 24 repos with 68 API calls in 36s and returned exactly the 3 reds the PR names. All 3 are push-triggered and none come from fork PRs.
- **Fixture fidelity:** I pulled the ASK_AI_consultant `validate` runs live. They match `_CONSULT_RUNS` exactly: 10 failures with the same timestamps. The deleted `verify` workflow really is absent from the workflows list (`total_workflows=1`).
- **Branches with a slash:** kipi-investigations has the default branch `feat/prd-rollout`. The `%2F`-encoded query and the raw one both return `total_count 1`.
- **stderr labelling:** real `gh api` on a missing repo prints `gh: Not Found (HTTP 404)`, and the reader turns it into the label `HTTP 404`. No token or URL reaches the body.
- **Python 3.9:** the launchd job runs `/usr/bin/python3`. The script parses with `ast.parse(..., feature_version=(3, 9))`.
- **Instance guard:** no instance root carries its own `instance-registry.json`, and the only installed `com.kipi.fleet-health.plist` points at the skeleton.
- **Suite:** `test-fleet-health-daily.py` gives `PASS: fleet-health-daily contract holds` in 47s. The 40s live GitHub cost is disclosed in the PR body.
- **Tried and dropped:** fork PRs from a branch named `main` turning the default branch red. No producer emits it (every window is `push` or `schedule` from the repo itself).

**Could not run:** the exact launchd environment. `env -i` and `/usr/bin/python3` both need approval in this sandbox, so I covered those by parsing at 3.9 and reading the PATH-injection test.

**Next action:** fix Finding 1 in `_origin_url` so a failed git call lands in the unreadable rollup, then add a registry test whose `origin_of` raises.

FINDINGS:
minor|_origin_url returns "" on any git failure, so registered_github_repos silently drops 21 of 24 repos (including both currently-red ones) with no unreadable rollup, printing "could not look" as "nothing found"|q-system/.q-system/scripts/fleet-health-daily.py:1678
minor|Repo dedup is case-sensitive while finding_key lowercases, so two case-variant remotes for one repo yield two findings with the same kipi-key and file_findings creates two permanent Linear issues in one run|q-system/.q-system/scripts/fleet-health-daily.py:1696
nit|Saturation branch counts cancelled runs toward the window, so one failure plus 29 cancelled runs is reported as "red for at least the last 30 completed runs" in a permanent issue body|q-system/.q-system/scripts/fleet-health-daily.py:1743
END FINDINGS

PR #359, opened 2026-09-15
fix(sync-check): fail on hook guard drift between the two settings files (ASK-1166)

assafkip:
## What

`settings-template-sync-check.py` now fails (exit 2) when the same (event, script) is guarded with an existence test (`test -f` or `[ -f`) in `settings-template.json` but bare in `.claude/settings.json`. That is DoR acceptance criterion 2: the check covers the whole class, not only the one row.

Why the bare shape is a real hazard: `python3 missing.py` exits 2, and a hook reads exit 2 as BLOCK. So a script absent from a checkout blocks every tool call on that event instead of no-opping.

## Deviations from the DoR, stated up front

- **The DoR measured 1 drifting wiring. Main has 22.** `hook_envelope_audit.py` is one of them. The four gate commits the DoR called local-only are on main now, so "local only" is stale too.
- **The local wiring is recorded, not fixed (criterion 1, second branch).** The only sanctioned writer of `.claude/settings.json` refuses the fix. I ran `apply-claude-changes.sh` with an `insert_before` proposal that adds the guard. It returned `REFUSED: enforcement ratchet: 1 hooks entr(ies) would disappear` (exit 2). The census keys each hook on its exact command string, so adding a guard reads as a removal. That applies to all 22 wirings, so all 22 sit in `GUARD_DRIFT_RECORDED` with that reason.
- **The list can only shrink.** Once someone guards a recorded line, its entry fails the check until it's deleted.
- **Test file:** the DoR names `test/test-settings-template-sync-check.sh`, which doesn't exist. The check's real test is `test_settings_template_sync_check.py`, next to the script. I extended that one.

Nothing outside the allowed files was touched. `hook_envelope_audit.py`, `probe_hook_envelope.py`, `voice-dna-loader.py` and `plugins/kipi-core/voiceloop/` are untouched.

## Evidence

- RED before the fix: `[FAIL] guard drift (template guarded, runtime not) -> exit 2 (reproducer)`. The check returned 0.
- GREEN after: `python3 q-system/.q-system/scripts/test_settings_template_sync_check.py` passes 14/14 checks.
- Real repo: `settings-template-sync-check.py --check` exits 0.
- Mutation: a copy of the check with the `PostToolUse:hook_envelope_audit.py` entry removed exits 2 and names that wiring.
- lefthook pre-commit is green, including `settings-template-sync`.

## Still open

Actually guarding the 22 lines needs one of two things. Either a direct edit to `.claude/settings.json` outside the apply path, or an apply-engine op that allows a guard-only rewrite of a hook command. Both are outside this DoR.


assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-359-20260915-084520.md` (6005 bytes).

```
FINDINGS:
minor|GUARD_RE treats the `2>/dev/null \|\| true` shape as bare, so a runtime line that already exits 0 on a missing script is flagged as a blocking hazard (2 of the 22 recorded entries are this shape)|q-system/.q-system/scripts/settings-template-sync-check.py:118
minor|Stale-record detection only fires for keys still wired in both files under the same event, so an unwired or moved recorded entry lingers silently, contradicting the "never goes stale silently" comment|q-system/.q-system/scripts/settings-template-sync-check.py:183
minor|Guard-only drift now aborts kipi update and fails lefthook, whose abort text tells the operator to add stranded hooks to settings-template.json, the wrong fix for this failure|q-system/.q-system/scripts/settings-template-sync-check.py:235
minor|The check measures difference not hazard, and the only writable way to clear new drift is stripping the template guard, which it accepts and which ships a bare hook to every instance|q-system/.q-system/scripts/settings-template-sync-check.py:181
END FINDINGS
```

--- reviewer output ---

**Verdict: APPROVE WITH NITS.** Four minor findings, all reproduced. None of them is a blocker or a major. Fix first: the check treats the `2>/dev/null || true` shape as bare, so it raises false alarms and 2 of the 22 recorded entries aren't real hazards.

## Findings

**1. Minor. A line ending in `|| true` counts as "bare" even though it already exits 0 when the script is missing.**
- Where: `q-system/.q-system/scripts/settings-template-sync-check.py:118` (`GUARD_RE`), used at line 162.
- This shape is in real use. Runtime `.claude/settings.json` lines 112 and 121 use it, and those are the recorded entries `SessionStart:memory-scores-surface.py` and `SessionStart:sycophancy-monthly-check.py`. So 20 of the 22 recorded wirings are hazards, not 22. The stderr claim "a missing script then exits 2 and BLOCKS" is false for these two.
- Reproducer: I imported the test module's `build_repo`. The runtime line was `python3 ".../foo-lint.py" 2>/dev/null || true`, and the template line was `test -f X && python3 X`. Then I ran `--check` and ran the flagged command with the script absent.
- Real output:
  - `check exit: 2`
  - `stderr: ... PostToolUse:foo-lint.py is guarded (test -f) in settings-template.json but bare in .claude/settings.json; a missing script then exits 2 and BLOCKS ...`
  - `the flagged runtime command, script absent, exits: 0`

**2. Minor. The comment says a stale record "never goes stale silently." That's true only while the script is still wired under the same event in both files.**
- Where: `settings-template-sync-check.py:183`. The comment it contradicts is at line 127.
- A recorded key that gets unwired from both files, or moved to another event, stays in the list and passes at exit 0. If it's later re-added bare, the check accepts it without a word.
- Reproducer: key `PostToolUse:audhd-lint.py`. Case A drops it from both files. Case B moves it to PreToolUse, guarded in both.
- Real output:
  - `A unwired from both, record now stale: (0, '')`
  - `B moved to PreToolUse and guarded, record now stale: (0, '')`

**3. Minor. The new exit 2 now aborts `kipi update`, and the abort text names the wrong fix.**
- Where: `settings-template-sync-check.py:235`. The new exit feeds `kipi-update.sh:588-593`, and lefthook.yml line 99 has the same problem in its `fail_text`.
- For drift that is only a guard difference, the abort tells the operator to "Add the stranded hook(s) to settings-template.json". That's wrong. The correct fix is printed above it, in the check's own stderr.
- Reproducer: a temp repo with guard-only drift, then the real lines 586-595 of `kipi-update.sh`, run with `SCRIPT_DIR` pointed at it.
- Real output:
  - `kipi-update.sh lines 586-595 exit: 1`
  - `ABORT: .claude/settings.json and settings-template.json are out of sync (above).`
  - `Add the stranded hook(s) to settings-template.json before propagating,`

**4. Minor. The check measures the difference between the two files, not the hazard. The one edit that clears it and is allowed to land makes the fleet bare.**
- Where: `settings-template-sync-check.py:181`. The fix text is at lines 190-193.
- The fix text says to guard the `.claude/settings.json` line. `claude-path-write-guard.py` blocks that edit, and so does the `apply-claude-changes.sh` ratchet. `settings-template.json` has no such guard. So when an agent hits a new unrecorded drift, the only editable way to make the check pass is to strip the template's guard. The check then accepts both lines bare, and that bare template ships to every instance.
- Real output:
  - `new drift, template guarded: 2`
  - `same drift silenced by stripping the TEMPLATE guard (fleet now bare): 0`

## What held up

- **Test suite:** 14 of 14 pass on this checkout.
- **Real repo:** `--check` exits 0. I also confirmed the drift set this check computes on the real repo matches `GUARD_DRIFT_RECORDED` exactly.
- **The tests can fail.** I ran four mutations against copies of the check, and the suite went red on every one:
  - drift detection removed: 2 FAIL
  - recorded exemption removed: 1 FAIL
  - stale detection removed: 1 FAIL
  - the `[ -f` alternative dropped from `GUARD_RE`: 1 FAIL
- **Reverse direction (runtime guarded, template bare):** empty today. Every non-guarded template hook line carries `|| true`.
- **Early-exit paths:** the `or ([], [])` fallback is safe, and so is the no-template case (instance no-op). Hook mode still exits early for files outside its scope.

## Tried and dropped

- **The test 10 fixture is invented.** The only real loop form in either file is the claude-path-write-guard restore-or-exit-2 command, which is not what test 10 builds. That command is identical in both files, though, so it can't produce drift. It doesn't reach nit-worthy.
- **The guard regex matching anywhere in the command, not tied to the script.** I found no multi-script command in either file where this changes the result, so there was no reproducer to write.

FINDINGS:
minor|GUARD_RE treats the `2>/dev/null \|\| true` shape as bare, so a runtime line that already exits 0 on a missing script is flagged as a blocking hazard (2 of the 22 recorded entries are this shape)|q-system/.q-system/scripts/settings-template-sync-check.py:118
minor|Stale-record detection only fires for keys still wired in both files under the same event, so an unwired or moved recorded entry lingers silently, contradicting the "never goes stale silently" comment|q-system/.q-system/scripts/settings-template-sync-check.py:183
minor|Guard-only drift now aborts kipi update and fails lefthook, whose abort text tells the operator to add stranded hooks to settings-template.json, the wrong fix for this failure|q-system/.q-system/scripts/settings-template-sync-check.py:235
minor|The check measures difference not hazard, and the only writable way to clear new drift is stripping the template guard, which it accepts and which ships a bare hook to every instance|q-system/.q-system/scripts/settings-template-sync-check.py:181
END FINDINGS

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-359-20260924-072515.md` (4517 bytes).

```
FINDINGS:
minor|guard_states attributes a whole-command `test -f` to every script in that command, so a bare invocation riding inside a guarded command is scored as guarded|q-system/.q-system/scripts/settings-template-sync-check.py:162
minor|Round 2 reviews code byte-identical to round 1 (only new commit is a merge of origin/main) which already returned APPROVE WITH NITS, so this round re-derives an existing tracked list|q-system/.q-system/scripts/settings-template-sync-check.py:114
END FINDINGS
```

--- reviewer output ---

## Round 2, and the code is unchanged

The only commit since round 1 is `61ef7884`, a merge of `origin/main`. The two authored commits (`c49449fa`, `37124f87`) are the ones round 1 reviewed. No fix, no reply to round 1's four minors.

Round 1 already returned **APPROVE WITH NITS** with four reproduced minors. Re-reviewing identical code re-derives the same list. That is the biggest thing in this review, and it is a finding about the process, not about the code.

---

## New finding (round 1 tried this and dropped it for lack of a reproducer)

**minor. `guard_states` attributes a whole-command `test -f` to every script named in that command, so a bare invocation riding in a guarded command reads as guarded.**

Where: `q-system/.q-system/scripts/settings-template-sync-check.py:162`. `is_guarded` is computed once per command by searching the whole string, then applied to every name `SCRIPT_RE` finds in it.

Round 1 wrote: "I found no multi-script command in either file where this changes the result, so there was no reproducer to write." The shape does exist. `.claude/settings.json` carries one command holding `[ -f ... ]` tests and a bare invocation of `q-system/.q-system/scripts/slack-notify.sh` through `${KIPI_NOTIFY:-...}`:

```
for R in q-system/.q-system/scripts/claude-path-write-guard.py q-system/.q-system/scripts/claude-integrity-tripwire.py; do [ -f "$CLAUDE_PROJECT_DIR/$R" ] && continue; ... N="${KIPI_NOTIFY:-$CLAUDE_PROJECT_DIR/q-system/.q-system/scripts/slack-notify.sh}"; ... "$N" "SECURITY: ..." ...
```

Reproducer, run just now (same shape, one command guarding `alpha-lint.py` and bare-invoking `beta-lint.py`, template guards both):

```
beta-lint.py: BARE in settings.json, GUARDED in settings-template.json
expected exit 2 (guard drift); ACTUAL exit: 0
stderr: ''
```

Calibration, stated per the standing rule: the command shape is real in this repo, the **asymmetry** is constructed. No producer emits the asymmetric pair today. That is why this is minor, not major.

---

## Round 1's minors, re-verified live, already tracked

Not re-filing these. They would become duplicate follow-ups. Recording that I ran them and they are still live:

- **`|| true` reads as bare.** Real output from my run: check `exit: 2`, and the flagged command with the script absent `really exits: 0`. The stderr's "exits 2 and BLOCKS" is false for that shape. Two recorded entries (`SessionStart:memory-scores-surface.py`, `SessionStart:sycophancy-monthly-check.py`) carry it.
- **Stale records go silent.** My repro on `PostToolUse:audhd-lint.py`: retire the wiring from both files and the check gives `exit: 0, stderr: ''`, contradicting the comment at line 127. Re-wire it bare later and it is still `exit: 0`, permanently exempt.

---

## What is sound

- **Reverse-direction drift** (runtime guarded, template bare). Measured on the real files: **0**. The one direction the check skips has no live instance, so the choice does not leave a hole today.
- **Negated guard idiom** (`[ ! -f`, `test ! -f`). Zero commands in either file. No false positive there.
- **`or ([], [])` fallback.** `([], [])` is truthy, so a genuine empty result is never replaced by the fallback. Confirmed in the interpreter.
- **The recorded table is exact.** Real repo: 22 drifting wirings, 22 recorded, 0 unrecorded, 0 stale, 0 invisible. `--check` exits 0.
- **The tests can go red.** Mutation: deleted `"PostToolUse:voice-lint.py"` from a copy of the check, ran it against the real repo, got `exit 2` naming that wiring.
- **Suite:** 14 of 14 pass.
- **Instance no-op:** template absent gives exit 0, so nothing new fires across the fleet.

The claimed hazard is real: `python3 /tmp/definitely-not-here-xyz.py` exits **2** on this machine.

---

Fix first: the `|| true` false positive at line 118. It is the one that cries wolf on a shape already in the file, and a checker the operator learns to ignore costs the real alert later.

FINDINGS:
minor|guard_states attributes a whole-command `test -f` to every script in that command, so a bare invocation riding inside a guarded command is scored as guarded|q-system/.q-system/scripts/settings-template-sync-check.py:162
minor|Round 2 reviews code byte-identical to round 1 (only new commit is a merge of origin/main) which already returned APPROVE WITH NITS, so this round re-derives an existing tracked list|q-system/.q-system/scripts/settings-template-sync-check.py:114
END FINDINGS

PR #361, opened 2026-09-15
feat(launchd): which tree each loaded job runs from, and is commit X live for it (ASK-1135)

assafkip:
## ASK-1135: merged is not deployed for loaded launchd jobs

New `q-system/.q-system/scripts/launchd-deploy-gap.py`:

- Enumerates **loaded** jobs from `launchctl list`, never the plist dir. A paused job is absent by construction.
- Maps each job to its git tree from its plist. It tries WorkingDirectory first, then absolute paths in ProgramArguments, including `cd <repo> && ...` inside `bash -c`. Git-ignored interpreter paths (for example `/opt/homebrew/bin/python3`) are skipped. XML comments are stripped before parsing, so a `--` inside a comment no longer blinds plistlib.
- Reports branch, **default resolved via `git symbolic-ref refs/remotes/origin/HEAD`** (never an assumed `main`), behind, **ahead**, **dirty**, and the age of the last fetch.
- `--live SHA LABEL` answers whether a commit is in the tree a job runs from. Exit codes: 0 live, 1 not live, 2 no answer.
- `launchd-health-check.py` calls it on its existing schedule. It files **one rollup Linear issue** through the fleet filer, deduped by a stable key, and pages nobody.

## Acceptance criteria

- [x] Default comes from origin/HEAD, and a copy that assumes `origin/main` goes RED on a master-default repo with a stale `main`. That's test 1 plus test 7 (the mutant), which re-runs the suite against the mutated copy and requires exit 1 on assertion 1a.
- [x] Loaded jobs only. A paused job with a plist on disk is absent from the count (tests 4a and 4b).
- [x] "Is commit X live for job Y", tested on the real ASK-1132 fix `896b0e5a`. It's on origin/main. A tree at its parent, run by a `cd <repo> && ./kipi dor --apply` job, answers NOT live (5b). The same tree at the fix answers live (5d).
- [x] Ahead and dirty are reported, not only behind (tests 2a and 2b).
- [x] Reported through the alert path: one rollup via `fleet-health-daily.file_findings`, called from the scheduled watchdog (tests 6a to 6d). The watchdog suite pins the call order.

## Ran, not assumed

- `python3 q-system/.q-system/scripts/test_launchd_deploy_gap.py` gave `PASS: launchd-deploy-gap`. Before the script existed it failed with FileNotFoundError.
- `python3 q-system/.q-system/scripts/test_launchd_health_check.py` gave `PASS: all launchd-health-check logic checks green`.
- `python3 q-system/.q-system/scripts/test_launchd_intent_verify.py` gave `PASS: launchd-intent-verify`.
- `capability-gate.py --repo-root .` is RED (12) on rows that predate this branch: `declared_inert` entries cite resolved spillover ids. None of the rows touch these files. Filed as sp-e842be32 / ASK-1718.

## Not verified here

This session was refused `launchctl list` and reads of `~/Library/LaunchAgents`, so there is **no live count of real jobs** in this PR. Every test runs on temp repos, fake plists and injected `launchctl` output. The first scheduled watchdog run after deploy produces the real number. One catch: the watchdog runs from the primary checkout, so this change only runs once that checkout has it. That is the exact gap this issue describes.


assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-361-20260915-092050.md` (5793 bytes).

```
FINDINGS:
major|test_launchd_deploy_gap.py is in expected_tests but not skeleton_only, and test 5 requires kipi-system commit 896b0e5a, so it crashes and turns kipi check RED in every instance|q-system/.q-system/scripts/test_launchd_deploy_gap.py:172
minor|--live exits 1 (NOT LIVE) instead of 2 (no answer) when launchctl list fails, because the RuntimeError is never caught|q-system/.q-system/scripts/launchd-deploy-gap.py:287
minor|git status runs without --no-optional-locks and holds .git/index.lock in live job trees, failing a concurrent git write and leaving a stale lock when the 30s timeout kills it, which contradicts the read-only docstring|q-system/.q-system/scripts/launchd-deploy-gap.py:88
minor|a tree on its default branch with only dirty files is filed as off its default branch, the rollup is rewritten whenever the dirty count moves, and a closed rollup reopens on the next run|q-system/.q-system/scripts/launchd-deploy-gap.py:246
END FINDINGS
```

--- reviewer output ---

**Verdict: REQUEST CHANGES.** The new test crashes in every instance repo. The first fix is to put it on the `skeleton_only` list. Four findings survived reproduction: one major, three minor. Every repro was run in a temp dir, and nothing in the repo was changed.

## Finding 1 (major): the new test turns `kipi check` red in every instance

- **Claim:** `test_launchd_deploy_gap.py` is listed in `expected_tests` but not in `skeleton_only`, so it runs in every instance. There, test 5 crashes because the instance's history has no kipi-system commit `896b0e5a`.
- **Where:** `q-system/.q-system/scripts/test_launchd_deploy_gap.py:172`. Nothing was added under `q-system/.q-system/capability/skeleton_only/`.
- **Why it fans out:** `capability-gate.py` `run_tests` runs every expected test that isn't on the `skeleton_only` list when it's in instance mode. `kipi` (around line 284) runs that full gate inside an instance on every `kipi check`. The 21 entries already in `skeleton_only/` exist for exactly this.
- **Repro:** I copied `launchd-deploy-gap.py` and the test into a fresh repo at `q-system/.q-system/scripts/`, committed, and ran `python3 test_launchd_deploy_gap.py`.
- **Output:** `subprocess.CalledProcessError: Command '[... 'rev-parse', '896b0e5a^{commit}']' returned non-zero exit status 128.` then `exit = 1`.
- **Fix:** add a `skeleton_only` entry for the test, shaped like the 21 already there.

## Finding 2 (minor): `--live` says NOT LIVE when it actually has no answer

- **Claim:** if `launchctl list` fails, the RuntimeError is never caught and Python exits 1. The documented contract says 1 means "not live" and 2 means "no answer".
- **Where:** `launchd-deploy-gap.py:287` (the raise is at 256).
- **Repro:** I put a fake `launchctl` that exits 1 first on PATH, then ran `launchd-deploy-gap.py --live 896b0e5a com.kipi.linear-dor`.
- **Output:** `RuntimeError: launchctl list exited 1` then `exit = 1`.
- **Why it's minor:** nothing consumes the exit code yet. The PR body itself says this session was refused `launchctl`, so this is the likely failure.

## Finding 3 (minor): the survey calls itself read-only but its `git status` holds `.git/index.lock`

- **Claim:** `git status` runs without `--no-optional-locks`, so it holds the index lock in trees that live jobs write to. A concurrent `git add` in that tree fails. If the 30s timeout kills git with SIGKILL, the lock is left behind.
- **Where:** `launchd-deploy-gap.py:88`. The docstring at line 26 says "it reads, it does not move refs under a running job".
- **Repro A (race):** `dg.git(repo, "status", "--porcelain")` in a thread, `git add a` while the lock existed. Output: `fatal: Unable to create '.../repo/.git/index.lock': File exists.`
- **Repro B (kill):** the same call shape with the timeout scaled down. Output: `stale index.lock: True`, and the next `git add` failed the same way. Control run with `--no-optional-locks`: `index.lock left: False` every time.
- **Why it's minor:** a 30s status is an input I constructed. A stale lock would still break every git write in that checkout until someone deletes it.

## Finding 4 (minor): a dirty tree on its default branch is filed as "off its default branch", rewritten on every run, and reopened after close

- **Claim:** the reasons include the dirty count, and those reasons go into the hashed rollup body. So a current, on-default tree with only dirty files is titled "off its default branch". The issue is rewritten whenever the dirty count moves, and reopened on the next run if an operator closes it.
- **Where:** `launchd-deploy-gap.py:233` and `:246`.
- **Repro:** I ran `linear_findings` into `fleet-health-daily.file_findings` with an in-memory fake Linear.
- **Output:** `dirty 2 then 3: updated = 1 mutations = 1`, then `operator canceled it, next run, nothing changed: reopened = 1`.
- **Why it matters:** the watchdog's own primary checkout is usually dirty, so this rollup can't really be closed. It pages nobody.

## What held up

- The `origin/HEAD` resolution is real. Test 7's mutant genuinely turns the suite red.
- Enumeration covers loaded jobs only. Stripping XML comments before parsing works.
- Homebrew's gitignored interpreter is skipped correctly. I checked all 17 tracked plists and found no job script on a gitignored path, so I dropped my false-negative concern.
- `run_deploy_gap_check` catches every exception, so a git or launchctl failure can't take down the watchdog. The watchdog tests stub it, so they never touch live `launchctl` or Linear.
- There's only one watchdog label (`com.kipi.launchd-health`), so two instances can't fight over the same rollup key. Dropped.
- Skeleton CI checks out with `fetch-depth: 0`, so test 5 resolves there. Both suites pass: `PASS: launchd-deploy-gap` and `PASS: all launchd-health-check logic checks green`.

FINDINGS:
major|test_launchd_deploy_gap.py is in expected_tests but not skeleton_only, and test 5 requires kipi-system commit 896b0e5a, so it crashes and turns kipi check RED in every instance|q-system/.q-system/scripts/test_launchd_deploy_gap.py:172
minor|--live exits 1 (NOT LIVE) instead of 2 (no answer) when launchctl list fails, because the RuntimeError is never caught|q-system/.q-system/scripts/launchd-deploy-gap.py:287
minor|git status runs without --no-optional-locks and holds .git/index.lock in live job trees, failing a concurrent git write and leaving a stale lock when the 30s timeout kills it, which contradicts the read-only docstring|q-system/.q-system/scripts/launchd-deploy-gap.py:88
minor|a tree on its default branch with only dirty files is filed as off its default branch, the rollup is rewritten whenever the dirty count moves, and a closed rollup reopens on the next run|q-system/.q-system/scripts/launchd-deploy-gap.py:246
END FINDINGS

assafkip:
## Review response, commit 28840f2c

Each finding went RED in the suite before its fix, then green.

- **major, instance crash:** fixed. Added `q-system/.q-system/capability/skeleton_only/q-system__.q-system__scripts__test_launchd_deploy_gap.py.json`. Test 8 loads the manifest through `capability_manifest.load` and requires the entry.
- **minor, `--live` exit 1 on launchctl failure:** fixed. `launchctl_listing` turns every failure into RuntimeError, and `main` turns that into `NO ANSWER`, exit 2. Test 9 runs the script with a fake `launchctl` that exits 1.
- **minor, index lock:** fixed at the chokepoint. `git()` passes `--no-optional-locks` on every call. Test 10 makes the index stat-stale and requires its bytes and mtime untouched. On the old code it was rewritten.
- **minor, rollup wording and churn:** the wording and churn parts are fixed. The title now says "do not run their default branch as merged", and the filed body carries reason kinds without counts, so a moving count no longer rewrites the issue. A new kind of risk still does (tests 11a to 11c). Counts stay in the script's stdout report.
- **Rejected, the reopen part:** reopening a closed rollup while its finding is still true is `file_findings`' deliberate lifecycle, see the comment in `_refresh_one` in `fleet-health-daily.py`. A dirty tree that live jobs run from is a real finding under this issue. Closing it without cleaning the tree is the blind spot that lifecycle exists to catch.

Ran:
- `python3 q-system/.q-system/scripts/test_launchd_deploy_gap.py` gave `PASS: launchd-deploy-gap`. Before the fix it gave `FAIL` on 8, 9a, 9b, 10, 11a and 11b.
- `python3 q-system/.q-system/scripts/test_launchd_health_check.py` gave `PASS: all launchd-health-check logic checks green`.
- `python3 q-system/.q-system/scripts/test_launchd_intent_verify.py` gave `PASS: launchd-intent-verify`.

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-361-20260915-093350.md` (7626 bytes).

```
FINDINGS:
major|--live and the rollup map com.kipi.dispatch to the primary checkout named in its plist, but kipi-dispatch-pinned.sh:202 execs the payload from a worktree pinned at origin/main, so --live answers LIVE/NOT LIVE inverted (exit 0/1) and the rollup flags a job that runs the default|q-system/.q-system/scripts/launchd-deploy-gap.py:158
minor|tree_state discards git status's return code, so a tree whose status fails reports 0 dirty and no risk reasons instead of an unreadable-tree finding|q-system/.q-system/scripts/launchd-deploy-gap.py:93
minor|run_check never sets outcome["owed"], so a refused rollup write (errors=1) prints "0 on the board, unfiled=0", byte-identical to a clean fleet|q-system/.q-system/scripts/launchd-health-check.py:716
nit|the check-ignore guard that keeps Homebrew-interpreter jobs off /opt/homebrew's tree has no test; deleting it leaves the suite green|q-system/.q-system/scripts/launchd-deploy-gap.py:178
END FINDINGS
```

--- reviewer output ---

**VERDICT: REQUEST CHANGES.** I found one major. The `--live` verdict and the rollup look at the wrong git tree for `com.kipi.dispatch`, a real job whose plist ships in this repo. Fix that first.

## Findings

**1. MAJOR: `--live` gives a confident wrong verdict for `com.kipi.dispatch`, in both directions**

- **Where:** `q-system/.q-system/scripts/launchd-deploy-gap.py:158` (`job_tree`), used by `commit_live_for_job` at line 215.
- **Cause:** `job_tree` picks the first absolute path in ProgramArguments that resolves to a git repo. For dispatch that path is `__KIPI_REPO__/kipi-dispatch-pinned.sh`. But `kipi-dispatch-pinned.sh:202` runs `exec env KIPI_REPO="$PINNED" bash "$PINNED/kipi-dispatch.sh"`. That runs the job's real code from a detached worktree held at origin/main. So the checker asks the primary checkout, which is often on a feature branch, instead of the tree the code runs from.
- **Why major:** the PR already reworked `--live` in review so that "can't tell" returns NO ANSWER instead of NOT LIVE (test 9). Here it returns exit 0 or exit 1 about a tree the job doesn't run from.
  - A merged fix reads NOT LIVE.
  - A commit that only exists on the feature branch reads LIVE. That is success reported for a deploy that never happened.
  - The unattended rollup also lists dispatch as off-default, which is wrong for the one job that is deliberately pinned to main.
- **Calibration:** the fixture mirrors the plist template in this repo and the worktree the wrapper builds. It is not an invented shape.
- **Reproducer:** a Python heredoc. It builds a primary clone on `feat/wip` with one unmerged commit, adds a worktree pinned at origin/main the way the wrapper does, writes a plist shaped like `com.kipi.dispatch.plist`, then calls `dg.survey` and `dg.commit_live_for_job`.
- **Real output:**
```
tree the checker picked  : /private/var/folders/.../tmpqr_gghr3/kipi-system
tree dispatch really runs: /private/var/folders/.../tmpqr_gghr3/state/dispatch/kipi-system
pinned HEAD == origin/main: True
fix merged on origin/main: checker live=False, truth for payload live=True -> WRONG
unmerged commit on primary feature branch: checker live=True, truth for payload live=False -> WRONG
rollup reasons for dispatch: ['on feat/wip, not main', '1 behind', '1 ahead']
```
- **Smallest fix:** make `--live` return NO ANSWER (exit 2) for a job whose tree is picked through a known wrapper, or follow `KIPI_DISPATCH_CHECKOUT` to the pinned tree. State the wrapper limit in the module docstring either way.

**2. MINOR: a failing `git status` reads as 0 dirty, so a dirty tree reports clean**

- **Where:** `q-system/.q-system/scripts/launchd-deploy-gap.py:93`. `_, porcelain = git(top, "status", "--porcelain")` throws away the return code, and line 96 counts an empty string as 0.
- **Contradiction:** the `git()` docstring says "a tree we cannot read is a finding". Here it is neither a finding nor a crash. It reads as clean.
- **Calibration:** I made the failure happen by corrupting the index, so the severity is dropped one level.
- **Reproducer:** a Python heredoc. It clones a temp repo, modifies one tracked file, adds one untracked file, then overwrites `.git/index` with garbage.
- **Real output:**
```
before: dirty = 2 reasons = ['2 dirty']
git status rc = 128 | fatal: .git/index: index file smaller than expecte
after : dirty = 0 reasons = []
files still modified on disk: locally modified | True
```

**3. MINOR: a refused rollup write logs the same line as a clean fleet**

- **Where:** `q-system/.q-system/scripts/launchd-health-check.py:716`, together with `launchd-deploy-gap.py:292`.
- **Cause:** `run_check` never sets `outcome["owed"]`, so `unfiled_count` falls back to `skipped_no_key` and ignores the `errors` bucket. That is the same trap the `LANDED_BUCKETS` comment in this file describes from the PR #19 review. `file_linear_findings` in the same file avoids it by setting `owed`.
- **Reproducer:** a Python heredoc. It puts a fake `launchctl` on PATH that lists one missing-plist `com.kipi.*` job, stubs `_fleet_health` to return `errors=1`, and runs `run_deploy_gap_check(False)`. It then runs again against an empty listing.
- **Real output:**
```
== write REFUSED (errors=1): findings owed = [1]
deploy gap rollup filed: 0 on the board, unfiled=0
== clean fleet: findings owed = [0] (file_findings([]) returns all zeros)
deploy gap rollup filed: 0 on the board, unfiled=0
unfiled_count(errors=1 outcome) = 0
```
- **Fix:** set `outcome["owed"] = len(findings)` in `run_check`.

**4. NIT: the Homebrew check-ignore guard has no test; removing it leaves the suite green**

- **Where:** `q-system/.q-system/scripts/launchd-deploy-gap.py:178`.
- **Detail:** `/opt/homebrew` is a git repo on this machine and `check-ignore` returns 0 for `bin/python3`, so the guard does real work. But no plist template in this repo puts a Homebrew binary in argv; they only put it in the PATH environment variable. That is why this stays a nit.
- **Reproducer:** a Python heredoc. It writes five mutants of the script to a tempdir and runs `test_launchd_deploy_gap.py` against each, using `DEPLOY_GAP_UNDER_TEST`.
- **Real output:**
```
drop check-ignore skip -> exit 0 | PASS: launchd-deploy-gap
```

- **Mutation kills:**
  - Removing `--no-optional-locks` fails test 10.
  - Removing the XML-comment strip fails 4d.
  - Putting counts back into the filed body fails 11b.
  - Letting a `launchctl` failure exit 1 fails 9a.
  - Test 7's origin/main mutant is built into the suite.
- **Suites pass as shipped:** `test_launchd_deploy_gap.py` gives PASS, exit 0. `test_launchd_health_check.py` gives PASS, exit 0.
- **Watchdog tests don't touch live state:** every `wd.run` and `wd.main` call in the watchdog test stubs `run_deploy_gap_check`, so the suite never shells real `launchctl` or files to Linear.
- **No credential in the permanent Linear body:** it carries git toplevels, branch names, labels and count-free kinds. Paths come from `git rev-parse --show-toplevel`, never from raw arguments.
- **Real plist shapes resolve correctly:** `bash -lc "cd __KIPI_REPO__ && ./kipi dor ..."`, `__KIPI_REPO__/...py` and `__ROOT__/automation/...sh` all map to the right tree. Only the pinned-wrapper indirection in finding 1 breaks it.
- **Watchdog can't die from this:** `run_deploy_gap_check` catches everything, so a git or Linear failure can't stop the launchd watchdog. It also runs before the early return, so a healthy fleet still gets surveyed.
- **Skeleton-only declaration holds:** test 8 checks it, and CI checks out with `fetch-depth: 0`, so test 5's real commit `896b0e5a` resolves.

FINDINGS:
major|--live and the rollup map com.kipi.dispatch to the primary checkout named in its plist, but kipi-dispatch-pinned.sh:202 execs the payload from a worktree pinned at origin/main, so --live answers LIVE/NOT LIVE inverted (exit 0/1) and the rollup flags a job that runs the default|q-system/.q-system/scripts/launchd-deploy-gap.py:158
minor|tree_state discards git status's return code, so a tree whose status fails reports 0 dirty and no risk reasons instead of an unreadable-tree finding|q-system/.q-system/scripts/launchd-deploy-gap.py:93
minor|run_check never sets outcome["owed"], so a refused rollup write (errors=1) prints "0 on the board, unfiled=0", byte-identical to a clean fleet|q-system/.q-system/scripts/launchd-health-check.py:716
nit|the check-ignore guard that keeps Homebrew-interpreter jobs off /opt/homebrew's tree has no test; deleting it leaves the suite green|q-system/.q-system/scripts/launchd-deploy-gap.py:178
END FINDINGS

assafkip:
## Review response, commit 9d364a70 (round 2)

Each new assertion was RED on 28840f2c before the fix. A mutant that removes each guard turns its own assertion RED again.

- **major, `com.kipi.dispatch` read from the wrong tree:** fixed. Known re-exec wrappers (`_REEXEC_WRAPPERS`, today only `kipi-dispatch-pinned.sh`) get NO ANSWER on `--live` and no rollup row. Following `KIPI_DISPATCH_CHECKOUT` would restate the wrapper's path logic in a second place, so I did not. The module docstring states the limit: an unknown wrapper of the same shape is still read as the tree it lives in. Test 12 renders the REAL `com.kipi.dispatch.plist` template. Test 12e fails if `kipi-dispatch-pinned.sh` stops exec'ing from `$PINNED`, which is the only thing that justifies the carve-out.
- **minor, a failed `git status` read as clean:** fixed. `dirty` is None, the reason says `status unreadable, dirty unknown`, and the report prints `? dirty`. Test 13 uses a corrupt index.
- **minor, a refused rollup write read as a clean fleet:** fixed. `run_check` sets `outcome["owed"] = len(findings)`. Test 14 runs a filer that returns `errors=1` and checks the watchdog's own `unfiled_count`: 1 when refused, 0 when clean.
- **nit, the check-ignore guard had no test:** test 15 added. It builds a repo that ignores `bin/`, with an interpreter inside it.

Ran:
- `python3 q-system/.q-system/scripts/test_launchd_deploy_gap.py` gave `PASS: launchd-deploy-gap`. On the old code it gave FAIL on 12a to 12d, 13a, 13b and 14a.
- Mutants, each run through `DEPLOY_GAP_UNDER_TEST`: no-wrapper went RED on 12a to 12d, no-status-rc on 13a and 13b, no-owed on 14a, no-check-ignore on 15. Every one exited 1.
- `python3 q-system/.q-system/scripts/test_launchd_health_check.py` gave `PASS: all launchd-health-check logic checks green`.
- `python3 q-system/.q-system/scripts/test_launchd_intent_verify.py` gave `PASS: launchd-intent-verify`.

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-361-20260915-094756.md` (5168 bytes).

```
FINDINGS:
minor|--live reports LIVE (exit 0) when the fix commit is in HEAD but the tree has uncommitted edits to the fix's files, so the job runs pre-fix code|q-system/.q-system/scripts/launchd-deploy-gap.py:241
nit|fetch_age_hours reads FETCH_HEAD from --git-common-dir, but FETCH_HEAD is per-worktree, so a freshly fetched linked worktree reports the main checkout's stale fetch age|q-system/.q-system/scripts/launchd-deploy-gap.py:90
nit|report_line prints "fetched Noneh ago" for a tree with no FETCH_HEAD, the same None-leak r2 fixed for dirty|q-system/.q-system/scripts/launchd-deploy-gap.py:269
END FINDINGS
```

--- reviewer output ---

# Review: PR 361 (launchd deploy gap)

**VERDICT: APPROVE WITH NITS.** Three findings reproduced: one minor, two nits. Nothing is a blocker or a major. The fix worth doing first: `--live` reads HEAD only, so a dirty working file can make it say LIVE while the job runs pre-fix code.

## Findings

**1. MINOR. `--live` answers LIVE when the tree's HEAD has the fix but the working file the job runs does not.**
- Where: `q-system/.q-system/scripts/launchd-deploy-gap.py:241`. `commit_live` only runs `merge-base --is-ancestor full HEAD` and never looks at the working tree.
- Why it matters: the PR's own measurement found dirty trees are the common case (93 dirty files in one tree running 14 jobs). A local edit on a file the fix touched gets the answer "deployed".
- Why only minor: nothing unattended reads the `--live` exit code (grep finds no caller besides the CLI), and the reason text does literally say "in HEAD".
- Repro: a temp repo with `drafter.py` fixed in commit `c70cd7b4`, then an uncommitted edit putting the pre-fix line back, and a `cd <repo> && python3 drafter.py` plist. Real output:
  - file the job executes: `TOOLS = None (pre-fix code)`
  - git status: `M drafter.py`
  - verdict: `(True, 'c70cd7b418 is in HEAD of /private/var/folders/.../pr361-dirty-dvjo_13h/repo')`
- Suggested fix: if `git diff --name-only <sha>^ <sha>` overlaps `git status --porcelain`, answer NO ANSWER (exit 2).

**2. NIT. `fetched Nh ago` reads the main checkout's FETCH_HEAD, not the linked worktree's.**
- Where: `launchd-deploy-gap.py:90`. It uses `--git-common-dir`, but FETCH_HEAD is a per-worktree file.
- Repro: backdate the main checkout's FETCH_HEAD by 72h, add a worktree, fetch in it. Output:
  - FETCH_HEAD files: `main/.git/FETCH_HEAD`, `main/.git/worktrees/wt/FETCH_HEAD`
  - reported for the worktree, which fetched seconds ago: `72.0`
- Why only a nit: no in-repo plist template runs a loaded job from a linked worktree. The dispatch worktree is already carved out.

**3. NIT. A tree that was never fetched prints `fetched Noneh ago`.**
- Where: `launchd-deploy-gap.py:269`. This is the same shape r2 fixed for "None dirty".
- Repro: a fresh `git clone` has no FETCH_HEAD. Output ends: `behind 0 / ahead 0, 0 dirty, fetched Noneh ago`.

- **Both suites green:** `test_launchd_deploy_gap.py` gave `PASS: launchd-deploy-gap`, and `test_launchd_health_check.py` gave `PASS`.
- **Tests that can fail:** I removed the git-ignore guard in a temp copy and test 15 went red (exit 1). Test 7 (the `origin/main` mutant), test 10 (index lock) and test 11b (count-free body) each assert something a plausible regression would break.
- **Fixture shapes match the real producers:** all 17 plist templates in the repo parse, and each maps the way the tests assume. linear-dor uses `cd <repo> && ./kipi`, browser-session-health uses Homebrew python, dispatch goes through the pinned wrapper. The wrapper carve-out is pinned to the real `exec env KIPI_REPO="$PINNED"` line.
- **Plist lookup by label:** every installer in the repo uses `~/Library/LaunchAgents/<label>.plist`, so a filename that differs from the label has no producer. Dropped.
- **Rollup lifecycle:** there is one stable key, a clean fleet files nothing, and a refused write counts as unfiled (`owed` is set). The watchdog test pins the call order and stubs the live path, so the suite cannot reach real Linear or launchctl.
- **Failure paths:** a `launchctl` failure gives NO ANSWER (exit 2). An unreadable `git status` is reported as unknown, not clean. `run_deploy_gap_check` catches everything, so it cannot kill the watchdog.
- **Manifest and CI:** the skeleton-only and expected-tests entries load with no errors. The `validate` job checks out with `fetch-depth: 0`, so test 5 can resolve `896b0e5a`.
- **Detector id outside fleet-health's registry:** same precedent as `launchd-intent-drift`. Not a new class. Dropped.

## Tried and dropped (no producer I could show)

- A plist that names its tree through `$HOME/...` or `~/...` inside a `bash -c` string. The path regex cannot see those, so the job would read as clean. All in-repo templates use absolute paths, and this sandbox refused reads of other repos' plists under `~/projects`.
- A job whose script sits in a gitignored folder inside a repo. The guard skips it and the job reads as "installed artifact". No in-repo producer does this.
- Pre-existing, not from this PR: `file_findings` still attaches no `needs-triage` label. That gap is already captured as sp-1306aca4.

FINDINGS:
minor|--live reports LIVE (exit 0) when the fix commit is in HEAD but the tree has uncommitted edits to the fix's files, so the job runs pre-fix code|q-system/.q-system/scripts/launchd-deploy-gap.py:241
nit|fetch_age_hours reads FETCH_HEAD from --git-common-dir, but FETCH_HEAD is per-worktree, so a freshly fetched linked worktree reports the main checkout's stale fetch age|q-system/.q-system/scripts/launchd-deploy-gap.py:90
nit|report_line prints "fetched Noneh ago" for a tree with no FETCH_HEAD, the same None-leak r2 fixed for dirty|q-system/.q-system/scripts/launchd-deploy-gap.py:269
END FINDINGS

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-361-20260923-104217.md` (7635 bytes).

```
FINDINGS:
major|risk_reasons never consults fetch age, so a clean tree on its default branch with a stale origin ref reports behind 0 and files no rollup row while running 5-commits-stale code|q-system/.q-system/scripts/launchd-deploy-gap.py:137
minor|run_deploy_gap_check swallows every failure to stderr only, so a permanently blind deploy-gap check files nothing to Linear and prints nothing to stdout|q-system/.q-system/scripts/launchd-health-check.py:715
END FINDINGS
```

--- reviewer output ---

## Round 4 review, PR #361

Read the code first, then rounds 1 to 3. Two findings survived reproduction: one major, one minor. Nothing in the repo was changed; both repros ran in temp dirs.

---

### Finding 1 (major): a stale `origin/<default>` ref makes the rollup call a stale job tree `ok`

**Claim.** `risk_reasons` (`q-system/.q-system/scripts/launchd-deploy-gap.py:137`) derives risk from `behind`, `ahead`, `dirty` and the branch name. `behind` comes from `git rev-list --left-right --count origin/<default>...HEAD`, which reads the LOCAL remote-tracking ref. `fetch_age_hours` is computed at line 89, printed by `report_line`, and consulted by nothing that decides risk. So a tree that is on its default branch, clean, and has not fetched since it was cloned reports `behind 0 / ahead 0`, gets zero reasons, and produces no rollup row at all.

**Repro** (built a bare origin, cloned a job checkout, landed 5 commits on the default branch, never fetched in the checkout):

```
tree_state: {'branch': 'master', 'default': 'origin/master', 'behind': 0, 'ahead': 0,
             'dirty': 0, 'fetched_hours_ago': None}
risk_reasons: []
report_line: ... on master, default origin/master, behind 0 / ahead 0, 0 dirty, fetched Noneh ago
3am rollup files: []
GROUND TRUTH commits on origin/master this checkout lacks: 5
newest merged fix live for this job? (False, '...is not in .../job-checkout's object store...')
```

`--live` gets it right because it reads the object store. The scheduled path, which is the only unattended consumer, reports `ok` and files nothing.

**Why major, not a nit.** This is the exact failure the script exists to name: merged is not deployed. Two consequences nobody has stated yet:

- The Linear rollup body (`launchd-deploy-gap.py:288`) carries no fetch age and no stale-ref reason, so the one artifact a human reads cannot distinguish "in sync" from "has not looked since March."
- The module docstring's own headline measurement, "Behind was 0 to 3 everywhere once measured correctly," is indistinguishable from "nothing had fetched recently." The number that justified dropping behind in favour of ahead/dirty is measured by the code path this finding is about.

**Relationship to round 3.** Round 3 raised two nits on `fetch_age_hours`: per-worktree FETCH_HEAD, and the `fetched Noneh ago` print. Both are about DISPLAY in `report_line`. This is a different clause (`risk_reasons` and `linear_findings`, which have never consulted fetch age) and a different consequence (the alert path stays silent on a stale tree). I am not escalating those nits; I am filing the classification gap.

**Reachability, stated honestly.** The false negative needs all three of: on the default branch, clean, stale ref. No producer in this repo emits that state, so it is an operational shape, not a fixture one. It is also precisely the shape of a client instance that was deployed once and is never pulled, and the author's own measurement found hand-managed checkouts 127, 174 and 168 commits ahead. If every checkout in the fleet is known to fetch on a schedule, this drops to minor.

**Smallest fix.** Add a reason when `fetched_hours_ago` is None or past a threshold, and put the fetch age in the rollup body. That keeps the no-fetch rule intact and stops a stale ref reading as agreement.

---

### Finding 2 (minor): a permanently blind deploy-gap check is one stderr line and nothing else

**Claim.** `run_deploy_gap_check` (`q-system/.q-system/scripts/launchd-health-check.py:715`) catches every exception, prints to stderr, and returns. Nothing reaches Linear and nothing reaches stdout, so the run log the operator reads shows no deploy-gap section at all.

**Repro** (a `launchctl` on PATH that exits 1, which is the exact refusal the PR body says this session hit):

```
STDOUT the operator's launchd-health.out gets: ''
STDERR: 'deploy gap check COULD NOT RUN: launchctl list exited 1\n'
findings filed to Linear: []
```

**Why minor, not major.** The healthy run prints `deploy gap rollup filed: ...` and the dead run prints nothing, so the two are distinguishable on stdout. It is the same posture the adjacent `run_intent_check` and `_fleet_health` already take, and the operator is no worse off than before this PR. It is still a detector that can be dead forever with no row on the board.

---

### What is sound

- **Mutation, 7 mutants, 6 killed.** Dropping the `check-ignore` guard dies on test 15. Ignoring `git status`'s return code dies on 13a and 13b. Dropping `--no-optional-locks` dies on test 10. Emptying `_REEXEC_WRAPPERS` dies on 12a to 12d. Putting counts back in the filed body dies on 11b. Not stripping XML comments dies on 4d. This suite is not decoration. The one survivor, parsing the `launchctl list` header row, is a semantically null mutation (`"Label".startswith("com.kipi.")` is False), so it is not a defect.
- **Fixtures against the real producer.** Rendered all 17 `com.kipi.*.plist` templates in this repo with a real git checkout and ran `survey` over them. Every working-tree job mapped to the right tree, and `com.kipi.dispatch` came back treeless with the wrapper note. The `bash -lc "cd X && ..."` shape, the `/usr/bin/python3 X/script.py` shape, and the three-repo `spillover-linear-check` shape all resolve correctly.
- **The Homebrew claim in the docstring is true on this machine.** `git -C /opt/homebrew check-ignore -q /opt/homebrew/bin/python3` exits 0.
- **Round 1 to 3 majors are settled.** The skeleton-only fragment exists and test 8 loads it through `capability_manifest.load`. `--live` with a failing `launchctl` gave exit 2 and `NO ANSWER` in my own run of test 9. Both suites pass: `PASS: launchd-deploy-gap` and `PASS: all launchd-health-check logic checks green`.
- **The wrapper carve-out is scoped.** Grepped every `worktree add` caller in the repo; `kipi-dispatch-pinned.sh` is the only launchd-facing one, so the one-entry tuple covers today's fleet, and test 12e fails if that wrapper stops re-exec'ing.

### On the review process

Round 3 landed APPROVE WITH NITS and this round brings a major on code that has not changed since round 1. That is the miscalibration signal the round-4 brief asks about, and it points at rounds 1 to 3 rather than at the author: three reviewers looked directly at `fetch_age_hours` and each treated it as a printing question. The value of `behind` is only as fresh as a ref nobody refreshes, and nobody asked what the number meant.

Not re-verified this round, so not filed: round 3's minor about `--live` reporting LIVE when the fix commit is in HEAD but the working tree has uncommitted edits to the fix's files. `commit_live` still reads HEAD ancestry only, so it appears untouched. It is already tracked from round 3.

### VERDICT: REQUEST CHANGES

Not BLOCK. Nothing here is permanent or unrecoverable, and auto-merge squash is enabled on this PR, which is the only reason the major matters more than usual.

**Fix first:** make a stale or absent fetch age a risk reason and put it in the rollup body, so a tree that has not looked at its remote stops reading as a tree that agrees with it.

FINDINGS:
major|risk_reasons never consults fetch age, so a clean tree on its default branch with a stale origin ref reports behind 0 and files no rollup row while running 5-commits-stale code|q-system/.q-system/scripts/launchd-deploy-gap.py:137
minor|run_deploy_gap_check swallows every failure to stderr only, so a permanently blind deploy-gap check files nothing to Linear and prints nothing to stdout|q-system/.q-system/scripts/launchd-health-check.py:715
END FINDINGS

PR #364, opened 2026-09-16
fix(launchd-health): watch every job family by default, report unknown ones (ASK-1124)

assafkip:
## ASK-1124: launchd watchdog watches every job family by default

The watchdog used to check only the families on an allowlist. Any family nobody added was never looked at. On 2026-08-29 a sweep found two owned families that were not watched, and one of them had a job failing that nobody saw (ASK-1122).

### What changed (`q-system/.q-system/scripts/launchd-health-check.py`)
- Every plist in `~/Library/LaunchAgents` is now checked, except the watchdog itself and families on the denylist.
- Denylist: `IGNORED_PREFIXES` (com.apple., com.docker., com.google., com.contourdesign.). Each host can add more in `~/.config/kipi/launchd-ignore-prefixes.txt`. If a prefix is on both lists, the watch wins.
- A family that is neither watched nor denylisted produces one `unwatched_family` finding, named by its prefix. It is pinged through `slack-notify.sh`, which files to Sana's Linear queue. It is not filed under a fleet-health detector, because none owns that kind.
- `run()` now runs discovery first, then passes the broken labels to `run_intent_check`. "no declared intent" is no longer printed for a job the watchdog already reports as failing or not loaded. The intent check still runs before the early return.

### Acceptance criteria
- [x] Any installed non-Apple family that is neither watched nor denylisted is reported as a FINDING.
- [x] Negative self-test: against origin/main's script, the new tests fail on 9 assertions. On this branch they are green.
- [x] The 4 known failures are still reported. This is pinned by a test that includes the io.askconsulting job.
- [x] "no declared intent" is never printed for a failing job. The live host shows 0 such lines for the 4 broken undeclared jobs, and 4 are still printed for healthy undeclared jobs (the control).

### Evidence
- `python3 q-system/.q-system/scripts/test_launchd_health_check.py` prints `PASS: all launchd-health-check logic checks green`.
- `bash scratchpad/negative_selftest.sh` runs the new tests against origin/main's script: `FAIL:` with 9 assertions, including `an unknown family is reported BY NAME as a finding: got []` and `all four known failures are still reported: got [...3 labels]`. The scratchpad script is not committed.
- `launchd-health-check.py --dry` on the mini (read-only) reports 11 `UNWATCHED_FAMILY` lines: com.chief. (11 jobs), com.gates., com.home., com.ledger., com.lgtm., com.mailroom., com.penny., com.radar., com.research., com.triage., homebrew.mxcl. It also reports new FAILING lines for io.askconsulting.ask-crm-morning-refresh, io.askconsulting.ask-sameness and homebrew.mxcl.postgresql@15, and NOT_LOADED for 3 jobs in a second owned family.

### What changes once this merges and syncs
- The live run's due pings go up. The dry run on this host says `would ping 26 job(s)`, and `would-file=1` new issue.
- Those pings reach Sana's Linear queue, not the founder.
- `homebrew.mxcl.` is a third-party vendor on this host. It is left out of the fleet denylist on purpose. If it should be silenced, add one line to this host's `~/.config/kipi/launchd-ignore-prefixes.txt`.
- The pause ledger does not cover the NOT_LOADED jobs in the second owned family. If those jobs are stopped on purpose, they belong in `launchd-paused.txt`.

### Not in this PR
- The staleness gap from ASK-1123 is still open. The watchdog reads LastExitStatus only, so a job that stopped running without a non-zero exit (for example com.kipi.dispatch) is still not detected.


assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-364-20260916-075634.md` (5199 bytes).

```
FINDINGS:
minor|The test comment claims the "intent check runs before the early return" rule is still pinned, but the replacement case has a non-empty problem list and a mutant that moves the call after the early return stays green|q-system/.q-system/scripts/test_launchd_health_check.py:694
minor|An unreadable launchd-ignore-prefixes.txt (for example a directory) raises out of _prefixes_with_file, and because discovery now runs first, the run dies before any failing job is reported or the intent check runs, still exiting 0|q-system/.q-system/scripts/launchd-health-check.py:132
minor|The module docstring and the healthy-run message still describe prefix-only discovery of watched families, but the code now checks every non-denylisted plist|q-system/.q-system/scripts/launchd-health-check.py:4
END FINDINGS
```

--- reviewer output ---

**Verdict: APPROVE WITH NITS.** Three minor findings, no blocker or major. The first one to fix is the lost test pin (finding 1): the PR removed the only test that caught a dead intent check, and the comment says the test is still there.

## Findings

**1. MINOR. The test comment says the "intent check runs before the early return" rule is still pinned. It isn't.**
- Where: `q-system/.q-system/scripts/test_launchd_health_check.py:694` (the check itself is at line 699).
- Why: the old check (intent runs, then discover) was the pin, and this PR removed it. The replacement runs `_NOTHING_TO_FILE`, which holds one `paused` problem, so the list is never empty and `if not problems: return` is never reached.
- Reproducer: in a temp copy of `scripts/`, I moved the `run_intent_check(...)` call below the `if not problems: ... return` block in `run()`, then ran the suite.
- Real output:
  ```
  MUTANT at /var/folders/.../tmpoathsp8m/scripts : intent check moved AFTER the early return
  PASS: all launchd-health-check logic checks green
  ```
- Effect: a later edit can kill the intent check on an all-healthy fleet, which is the one state the code comment says it exists for, and the suite stays green.
- Fix: add one `run_capture([], ...)` case that asserts the calls are `["discover", "intent:..."]`.

**2. MINOR. If the new host denylist file can't be read, the whole watchdog run dies, and the intent check dies with it.**
- Where: `q-system/.q-system/scripts/launchd-health-check.py:132`. The code only catches `FileNotFoundError`. `load_paused_labels` also catches `NotADirectoryError`.
- Why it's worse now: discovery used to run after the intent check, so a crash in discovery still left the intent check done. Now discovery runs first, so one bad `launchd-ignore-prefixes.txt` stops everything.
- Reproducer: I pointed `IGNORED_PREFIXES_FILE` at a directory, set one fake `com.kipi.x` job to failing, and ran `wd.main(['--dry'])`.
- Real output:
  ```
  rc 0 intent calls []
  stdout: ''
  IsADirectoryError: [Errno 21] Is a directory: '/var/folders/.../ignore-as-dir'
  ```
- Effect: nothing is printed, the failing job is not reported, and the exit is 0. I built this input myself and no producer writes it, so I rated it one level down, as minor.

**3. MINOR. The module docstring and the healthy-run line still describe the old allowlist.**
- Where: `launchd-health-check.py:4` says "Auto-discovers every ~/Library/LaunchAgents/<prefix>*.plist for the OS's base job families". Line 831 prints "all watched launchd jobs healthy".
- Reproducer: `grep -n "Auto-discovers\|all watched launchd jobs healthy"` finds both lines. The code now globs every `*.plist` except the denylist.

## What held up

- **Live numbers match the PR.** A read-only `discover_problems()` on this host took 0.24s and found 12 failing, 11 unwatched families, 3 not loaded and 21 paused. 12 + 11 + 3 is the 26 pings the PR body says are due. The 11 families match the list in the PR body.
- **Denylist logic.** The third-party families are never status-checked and never reported. A watched prefix beats the denylist. An empty ignore tuple matches nothing. Blank and comment-only lines are skipped.
- **Test isolation.** `_fake_machine` sets `IGNORED_PREFIXES_FILE` to a temp path even when there are no ignore lines, so the real host file can't leak in. The suite passes: `PASS: all launchd-health-check logic checks green`.
- **Linear filing.** `unwatched_family` is not in `LINEAR_DETECTOR_BY_KIND`, so it never opens a permanent Linear issue and never counts toward `owed`. It only pings. Coverage filing still gets every intent finding; only the printed "no declared intent" line is dropped for broken jobs.
- **Ping dedupe.** A family finding is keyed by its prefix, so a change in the job count doesn't re-ping it early. It re-pings every 6 hours, like a failing job.

## Tried and dropped

- **More Linear tickets from the combined ping.** `alert-to-linear.py` fingerprints on the message's word content, so a different set of due jobs could open a new ticket. That behavior was already there before this PR, and the higher volume is stated in the PR body. Not reported.
- **A third-party job filing a permanent issue.** The failing `homebrew.mxcl.postgresql@15` job will file one Linear issue. The PR body says this is on purpose and gives the one-line host denylist fix. Not reported.

FINDINGS:
minor|The test comment claims the "intent check runs before the early return" rule is still pinned, but the replacement case has a non-empty problem list and a mutant that moves the call after the early return stays green|q-system/.q-system/scripts/test_launchd_health_check.py:694
minor|An unreadable launchd-ignore-prefixes.txt (for example a directory) raises out of _prefixes_with_file, and because discovery now runs first, the run dies before any failing job is reported or the intent check runs, still exiting 0|q-system/.q-system/scripts/launchd-health-check.py:132
minor|The module docstring and the healthy-run message still describe prefix-only discovery of watched families, but the code now checks every non-denylisted plist|q-system/.q-system/scripts/launchd-health-check.py:4
END FINDINGS

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-364-20260916-082016.md` (4840 bytes).

```
FINDINGS:
minor|An unreadable or non-UTF-8 launchd-ignore-prefixes.txt raises inside discover_problems, and since discovery now runs first, both failure reporting and the intent check go dark while the run exits 0|q-system/.q-system/scripts/launchd-health-check.py:127
minor|Newly checked non-owned plists are status-checked by file name instead of their Label key, so a loaded job whose file name differs is reported as not_loaded|q-system/.q-system/scripts/launchd-health-check.py:219
END FINDINGS
```

--- reviewer output ---

Two minor defects survived reproduction, and nothing reached major or blocker, so the verdict is **APPROVE WITH NITS**. The PR's own suite is green: `python3 q-system/.q-system/scripts/test_launchd_health_check.py` prints `PASS: all launchd-health-check logic checks green`. Temp-file writes were blocked in this environment, so each reproducer ran inline with `python3 -c`.

## Finding 1 (minor): one bad byte in the new ignore file blinds the whole watchdog

**Claim:** if `~/.config/kipi/launchd-ignore-prefixes.txt` can't be decoded, `discover_problems()` raises. Because the PR moved discovery ahead of the intent check, that check never runs either, and the run still exits 0.

- **Where:** `q-system/.q-system/scripts/launchd-health-check.py:127` (`_prefixes_with_file` only catches `FileNotFoundError`), made worse by the reorder at line 818.
- **Setup:** the ignore file holds `# vendor caf\xe9` (Latin-1). One `com.kipi.dispatch` job is stubbed as failing. Then `main(["--dry"])` runs.
- **Output on this branch:**
  - `exit: 0`
  - `intent check ran: []`
  - `stdout: ''`
  - `stderr last line: UnicodeDecodeError: 'utf-8' codec can't decode byte 0xe9 in position 12`
- **Control:** the same bad file against `origin/main`'s script prints `intent ran: ['intent'] | FAILING line printed: True`.
- **Blast radius:** the failing job goes unreported and the intent check is skipped, with only a traceback in the launchd log. The PR body tells operators to edit exactly this file.
- **Why minor, not major:** no code here writes a non-UTF-8 file. The input is one I built, so I dropped it a level. It's also recoverable by fixing the file. The watch-prefix file already had the same weakness; this PR adds a second such file and, with the reorder, now takes the intent check down too.
- **Fix:** catch `OSError` and `UnicodeDecodeError` in `_prefixes_with_file`, print a warning, and fall back to the base prefixes.

## Finding 2 (minor): a file name that differs from its Label is reported as not running

**Claim:** every plist outside the denylist is now checked with its file name as the label. A plist whose name differs from its `Label` key is reported as `not_loaded`, even when that job is loaded.

- **Where:** `launchd-health-check.py:219` (`status_problem` calls `job_status(label)` with `label = plist.stem`, set at line 244).
- **Setup:** a file `us.vendor.updater.plist` whose Label is `com.chief.brief`, a job that is loaded and healthy on this host.
- **Output:**
  - `real job_status(com.chief.brief): ('ok', 0)`
  - `[('us.vendor.updater', 'not_loaded', 'installed but not running'), ('us.vendor.', 'unwatched_family', ...)]`
- **Live-host check (read-only):** every newly checked plist on this host has a file name that matches its Label. So this input is also one I built, which is why it stays minor. Before this PR, only owned families were checked, and those follow the name-equals-label convention. Third-party vendors often don't.
- **Fix:** read `Label` with `plistlib` and pass that to `job_status`.

## What held up

- **Denylist and watch rules:** vendor jobs are never checked, and a watched prefix beats the denylist. The new tests pin both, and they fail against `origin/main`.
- **No fabricated issues for the new kind:** `unwatched_family` is not in `LINEAR_DETECTOR_BY_KIND`, so it creates no Linear issue under a borrowed detector key. `linear_findings` skips it, and `owed` doesn't count it, so the "unfiled" figure stays accurate.
- **Family pings are deduped:** `alert-to-linear.fingerprint` strips digits and punctuation, so a stable set of family findings lands on one repeat-counted ticket each time, not a new one.
- **Live dry-run matches the PR body:** 11 owned families are reported as unwatched, and `homebrew.mxcl.postgresql@15` is reported as failing.
- **The intent-check pin was moved, not lost:** "intent check runs even when nothing is found" is still covered by the dry `_NOTHING_TO_FILE` run. `_TWO_REAL` covers both `failing` and `not_loaded` for the broken-label handoff.
- **Test isolation:** `_fake_machine` swaps `LAUNCH_AGENTS`, both prefix files and `job_status`, then restores them. Nothing touches the operator's launchd.

**VERDICT: APPROVE WITH NITS.** Fix first: make `_prefixes_with_file` survive an unreadable file, since it now takes down both detectors at once.

FINDINGS:
minor|An unreadable or non-UTF-8 launchd-ignore-prefixes.txt raises inside discover_problems, and since discovery now runs first, both failure reporting and the intent check go dark while the run exits 0|q-system/.q-system/scripts/launchd-health-check.py:127
minor|Newly checked non-owned plists are status-checked by file name instead of their Label key, so a loaded job whose file name differs is reported as not_loaded|q-system/.q-system/scripts/launchd-health-check.py:219
END FINDINGS

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-364-20260923-082405.md` (7930 bytes).

```
FINDINGS:
minor|The replaced test pin does not hold the "intent check runs even when discovery finds nothing" rule; a mutant moving run_intent_check below the early return keeps the suite green, and no test drives run() with an empty problem list|q-system/.q-system/scripts/test_launchd_health_check.py:697
minor|_prefixes_with_file catches only FileNotFoundError, so an unreadable or directory-shaped launchd-ignore-prefixes.txt raises out of discover_problems and, since discovery now runs first, both failure reporting and the intent check go dark at exit 0 with empty stdout|q-system/.q-system/scripts/launchd-health-check.py:132
minor|The module docstring and the healthy-run message still describe prefix-only discovery of watched families, while the code now checks every non-denylisted plist|q-system/.q-system/scripts/launchd-health-check.py:4
END FINDINGS
```

--- reviewer output ---

## Review round 3, PR #364

All three findings below were raised in rounds 1 and 2. I re-ran each one myself against this branch and all three are still live, so each ships with my own executed reproducer. Severities are unchanged from the earlier rounds. No new blocker or major survived.

### First, a finding about the review process

The last code commit on this branch is `d0ae78f1`, dated 2026-09-16T14:55Z. Rounds 1 and 2 both returned APPROVE WITH NITS, and the only thing that has landed since is two prd-os receipt commits and a merge from main. Six minor findings were reported across two rounds and zero were acted on.

Two prior APPROVE verdicts on unchanged code means round 3 is not producing new information. This PR should have merged after round 1 with the minors tracked as follow-ups. Running a third adversarial pass on a diff nobody is editing is the cost, not the safeguard.

### Finding 1, minor. The replaced test pin does not hold what its comment says it holds.

`q-system/.q-system/scripts/test_launchd_health_check.py:697`

The diff deleted the check that the intent check runs BEFORE discovery, and the new comment claims what that pin protected, "the intent check runs even when discovery finds nothing", is now held by the dry `_NOTHING_TO_FILE` run below it. That fixture is `[("com.cole.gamma", "paused", ...)]`, which is a non-empty list, so `if not problems: return` is never reached by any test. `grep -n "run_capture(\[\]" q-system/.q-system/scripts/test_launchd_health_check.py` returns nothing.

Reproducer: copy `scripts/` to a tempdir, move the `run_intent_check(...)` call in `run()` below the `if not problems: ... return` block, run the suite.

Real output:

```
BASELINE COPY exit: 0 | PASS: all launchd-health-check logic checks green
MUTANT: intent check moved BELOW `if not problems: return`
exit: 0 | PASS: all launchd-health-check logic checks green
```

The mutant survives. On this branch the call is still above the return, so nothing is broken today. What is gone is the guard, and the comment asserts the guard is there.

Fix: one `run_capture([], _fh_stub())` case asserting `_intent_calls[:2] == ["discover", "intent:False"]`.

### Finding 2, minor. A malformed host denylist file blinds the entire run, silently, at exit 0.

`q-system/.q-system/scripts/launchd-health-check.py:132`

`_prefixes_with_file` catches only `FileNotFoundError`. `load_paused_labels` twenty lines above catches `(FileNotFoundError, NotADirectoryError)`, so the narrow catch is known to be insufficient in this same file. Because this PR moved discovery ahead of the intent check, a raise here takes both detectors down in one go.

Reproducer: point `IGNORED_PREFIXES_FILE` at a directory and at a chmod-000 file, then run `main(["--dry"])`.

Real output:

```
A directory-as-ignore-file RAISES: IsADirectoryError [Errno 21] Is a directory: .../ignore_dir.txt
B unreadable-ignore-file RAISES: PermissionError [Errno 13] Permission denied: .../noread.txt
C main() exit code: 0
C stdout: ''
C stderr first line: Traceback (most recent call last):
C stderr last line: IsADirectoryError: [Errno 21] Is a directory: .../ignore_dir.txt
```

Empty stdout, exit 0, traceback only in the launchd log. Stays minor because I constructed that input and nothing in the repo writes it. The PR body does tell the operator to hand-edit exactly this file, which is why it is worth the catch.

Fix: catch `OSError` and `UnicodeDecodeError` in `_prefixes_with_file`, warn, fall back to the base tuple.

### Finding 3, minor. The module docstring and the healthy-run line still describe the allowlist this PR removed.

`q-system/.q-system/scripts/launchd-health-check.py:4`

Line 4 still reads "Auto-discovers every ~/Library/LaunchAgents/<prefix>*.plist for the OS's base job families (WATCHED_PREFIXES)". Line 831 still prints "all watched launchd jobs healthy". The code now globs every `*.plist` and skips only self and the denylist.

Reproducer and real output:

```
$ grep -n "Auto-discovers\|all watched launchd jobs healthy" q-system/.q-system/scripts/launchd-health-check.py
4:Auto-discovers every ~/Library/LaunchAgents/<prefix>*.plist for the OS's base
831:            print("all watched launchd jobs healthy (loaded, exit 0)")
```

Second-order effect worth a line: on this host, 11 families are unwatched, so `problems` is never empty and line 831 is now unreachable in practice.

## What is sound

I attacked these and the code survived.

- **Live numbers match the PR body.** A read-only `discover_problems()` on the real host returned `{'not_loaded': 3, 'failing': 12, 'paused': 21, 'unwatched_family': 11}`. That is the 26 due pings the PR claims, and the 11 families listed match.
- **Filename versus Label key.** Round 2 raised this. I scanned all 107 installed plists with `plistlib`: only 2 have a stem that differs from their Label (`com.google.keystone.agent` and `com.google.keystone.xpcservice`, both with no Label key at all), and both are denylisted so neither is ever status-checked. No producer on this host emits the shape. Dropped rather than re-raised.
- **Ticket flood from the combined ping.** I chased the theory that a changing due-set opens a new Linear ticket per run. `alert-to-linear.fingerprint` strips every digit, path, date and punctuation run before hashing, and `record_pings` stamps all due jobs with one `now`, so the set stays stable and lands on one repeat-counted ticket. Could not make it produce a second. Dropped.
- **A reopen war with `fleet-health-daily.py`.** I looked for an auto-close path that would close an issue this watchdog then reopens twice a day. `fleet-health-daily.py` has `reopened` and `relisted` but no close-on-resolved anywhere. No loop exists. Dropped.
- **Third-party permanent issues.** `homebrew.mxcl.postgresql@15` does file one permanent `launchd-failing` issue. It is deduped by `finding_key(detector, label)`, it is one issue not a flood, the PR body names it, and the one-line host-local opt-out ships in the same change. Not a finding.
- **`unwatched_family` files no Linear issue under a borrowed detector.** It is absent from `LINEAR_DETECTOR_BY_KIND`, `linear_findings` skips it, and `owed` excludes it, so the `unfiled=` counter stays truthful. That is the right call and it is pinned by a test.
- **Denylist precedence and test isolation.** A watched prefix beats the denylist, vendor families are never status-checked, and `_fake_machine` swaps `LAUNCH_AGENTS`, both prefix files and `job_status` and restores them, so the suite cannot read the operator's real launchd.
- **Suite baseline.** `python3 q-system/.q-system/scripts/test_launchd_health_check.py` prints `PASS: all launchd-health-check logic checks green`.

## VERDICT: APPROVE WITH NITS

Only minor findings survived reproduction, and all three were already on the board from earlier rounds.

Fix first: finding 1, the lost test pin. It is one `run_capture([], ...)` case and it is the only one of the three where the repo currently believes it has protection it does not have.

FINDINGS:
minor|The replaced test pin does not hold the "intent check runs even when discovery finds nothing" rule; a mutant moving run_intent_check below the early return keeps the suite green, and no test drives run() with an empty problem list|q-system/.q-system/scripts/test_launchd_health_check.py:697
minor|_prefixes_with_file catches only FileNotFoundError, so an unreadable or directory-shaped launchd-ignore-prefixes.txt raises out of discover_problems and, since discovery now runs first, both failure reporting and the intent check go dark at exit 0 with empty stdout|q-system/.q-system/scripts/launchd-health-check.py:132
minor|The module docstring and the healthy-run message still describe prefix-only discovery of watched families, while the code now checks every non-denylisted plist|q-system/.q-system/scripts/launchd-health-check.py:4
END FINDINGS

PR #243, opened 2026-08-23
kipi-mcp suite collects and runs; registry reader tolerates schema drift (ASK-985)

assafkip:
Closes sp-dcd84af1 (promoted ASK-985). Stacked on #242.

- Five test modules ImportError-ed at collection under bare homebrew python; all deps already declared main deps, so the defect was the invocation. The suite now runs under its own runtime: uv run pytest tests/ = 708 passed, zero collection errors.
- Collecting surfaced a real crash the error had hidden: RegistryManager.list_excluded() bare-indexed excluded; live instance-registry.json drifted to a shape without it; kipi://instances resource was down on this box. Fixed at the load() chokepoint (setdefault), RED shown first (KeyError registry.py:25), test pins the drifted shape.
- Plugin bump kipi-core 1.7.21 -> 1.7.22.
- Issue closed through the spine with receipts at d72e7648; spillover sp-dcd84af1 resolved.

PR #365, opened 2026-09-16
fix(kipi-mcp): registry defaults are fresh lists per load (ASK-985)

assafkip:
Closes the one open item on ASK-985: the Codex minor on PR #243 (spillover sp-f3927437). The original ASK-985 fix already landed on main as d72e7648. This bounced back to Todo because no PR was open.

**Defect.** `RegistryManager.load()` handed out the class-level default lists themselves. `dict()` is a shallow copy, and `setdefault` inserts the same list object. So `add_instance` on one missing registry appended into the shared constant, and an unrelated manager then listed that instance.

**Fix.** `_REQUIRED_KEYS` is now a tuple of key names, and `load()` builds a new `[]` for each key. Files touched are all in the spec's `allowed_files`. The pre-commit hook bumped the plugin version: kipi-core 1.12.6 to 1.12.7.

**Evidence** (from `plugins/kipi-core/kipi-mcp`)
- RED before the fix: `uv run --extra dev python -m pytest -q tests/test_registry.py` gave `2 failed, 15 passed`. The failure was `assert [{'name': 'x'}] == []`.
- GREEN after the fix: the same command gave `17 passed`.
- Full suite: `1 failed, 724 passed`. The failure is `test_all_tools_registered` (`assert 69 == 73`), and it fails the same way with this change set aside. Filed as ASK-1760 (sp-17a4b967).

**Captured, not fixed here**
- ASK-1761 (sp-6a033a6d): the spec's required check `uv run python -m pytest` fails on a fresh checkout with `No module named pytest`, because pytest is in the `[dev]` extra.

Do not merge from the worker. Closeout runs through /issue-verify and /issue-closeout.


assafkip:
## Verdict: APPROVE

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-365-20260916-085350.md` (2252 bytes).

```
FINDINGS:
END FINDINGS
```

--- reviewer output ---

**Verdict: APPROVE.** I tried six ways to break PR #365 and none produced a finding. Nothing needs fixing before merge.

## Findings

None. There was nothing to reproduce, so there are no reproducers.

## What held up

- **The new tests really catch the bug.** I ran `tests/test_registry.py` against the old `registry.py` (commit `HEAD~1`), loaded in memory so the repo stayed untouched. Result: 2 failed, 15 passed. Both new tests failed, and the second one failed with `Left contains one more item: {'name': 'x'}`.
- **The fix makes them pass.** The same command on the PR code gave 17 passed.
- **Both tests set up the real bug.** One creates the registry through `add_instance`, which writes it. The other writes a real registry file that is missing a key. Both cases reached the shared-list code in the old version.
- **Each load gets its own lists.** I loaded a missing file, added an item to its `instances` list, then loaded a second missing file. The second load came back `{'instances': [], 'excluded': [], 'eliminated': []}`.
- **Nothing else depended on the old dict.** `git grep` finds `_REQUIRED_KEYS` only at `registry.py:15` and `registry.py:27`, so switching it to a tuple breaks no other code.
- **Error handling is unchanged.** A missing file still gives empty defaults. A file that isn't valid JSON, or isn't a JSON object, still raises, exactly as it did before.
- **The one full-suite failure was already there.** The full suite gives 1 failed, 724 passed, which matches the PR description. The failure is `test_all_tools_registered` (`assert 69 == 73`), and it fails the same way with the old `registry.py` loaded, so this PR didn't cause it.

## Operational risk

- **Blast radius:** reading the registry file now gives each caller its own lists. Nothing touches Linear or the network, and nothing is permanent.
- **Rollback:** a plain revert of this one commit.
- **Concurrency:** this PR fixes the shared lists inside one process. Two processes saving the same file at once can still overwrite each other. That problem was already there, and the PR doesn't claim to fix it.

**First thing to fix:** nothing. The PR can merge once `/issue-verify` and `/issue-closeout` run, as its description requires.

FINDINGS:
END FINDINGS

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-365-20260923-080830.md` (5491 bytes).

```
FINDINGS:
minor|Branch conflicts with current main on the plugin.json version line while auto-merge is armed, so the PR stalls at DIRTY with nothing paging anyone|plugins/kipi-core/.claude-plugin/plugin.json:3
nit|PR description states the version bump as 1.12.6 to 1.12.7 while the diff is 1.13.19 to 1.13.20|plugins/kipi-core/.claude-plugin/plugin.json:3
END FINDINGS
```

--- reviewer output ---

Read the PR, ran my own mutation proof, then read round 1. Here is what I found.

## Findings

### 1. minor — the branch conflicts with current `main` on the version line, and auto-merge is on, so it stalls silently

`plugins/kipi-core/.claude-plugin/plugin.json:3`

Auto-merge is enabled (SQUASH, enabled by assafkip 2026-09-16) and `mergeStateStatus` is `DIRTY`. The conflicting file is the one line this PR's pre-commit hook rewrote.

Reproducer:

```
git -C <worktree> fetch origin main -q
git -C <worktree> merge-tree --write-tree --name-only HEAD origin/main
```

Real output:

```
1c0c8d2c6dee94341585a64b50f91862ceea92bb
plugins/kipi-core/.claude-plugin/plugin.json

Auto-merging .prd-os/receipts.jsonl
Auto-merging plugins/kipi-core/.claude-plugin/plugin.json
CONFLICT (content): Merge conflict in plugins/kipi-core/.claude-plugin/plugin.json
```

The receipts ledger auto-merges fine. The version field does not. Structural note: a pre-commit hook that rewrites one monotonic integer on every kipi-core change guarantees a conflict between any two concurrent kipi-core PRs. With auto-merge armed, the loser sits DIRTY and nothing pages anyone. The 3am answer is a human merging main and re-resolving the bump.

### 2. nit — the PR body states a version delta that is not in the diff

`plugins/kipi-core/.claude-plugin/plugin.json:3`

Body: "The pre-commit hook bumped the plugin version: kipi-core 1.12.6 to 1.12.7."
Diff: `"version": "1.13.19"` to `"version": "1.13.20"`.

Reproducer and real output:

```
$ gh -R assafkip/kipi-system pr diff 365 | grep -A1 '1.13.19'
-  "version": "1.13.19",
+  "version": "1.13.20",
```

The body was written against a different tree. I re-ran every other evidence claim in it independently and they all held, so this is the only stale line.

## What is sound

- **The mutation.** I copied the package to a temp dir, replaced only `registry.py` with `origin/main`'s pre-fix version, confirmed the temp module actually shadowed the installed one (`MODULE RESOLVED: /private/var/.../pr365b.../kipi_mcp/registry.py`), and ran the PR's test file against it. Result `2 failed, 15 passed`, with `assert [{'has_git': ...leaked'...}] == []` and `assert [{'name': 'x'}] == []`. Against the PR code: `17 passed`. Both new tests can go red for exactly the reason they claim.
- **Both tests reach the real code path.** Test one goes through `add_instance`, which writes a real file. Test two writes a real registry missing a key. Neither is a fixture shaped around the author's mental model.
- **No other consumer of `_REQUIRED_KEYS`.** Repo-wide grep finds it only at `registry.py:15` and `registry.py:27`, so the dict-to-tuple change breaks nothing.
- **Same defect class elsewhere.** `draft_scanner.py` has class-level mutable constants (`BANNED_PHRASES`, `HEDGING_WORDS`, `COPY_FIELD_NAMES`) but never hands them to a caller, so the leak shape does not repeat.
- **The full-suite claim.** I ran it myself: `1 failed, 724 passed`. The failure is `test_all_tools_registered`, `Expected 69 tools, got 73. Extra: {'kipi_reddit_thread', 'kipi_reddit_listing', 'kipi_browser_probe', 'kipi_browser_fetch'}`. A registry-defaults change cannot register MCP tools, so ASK-1760 is correctly separate.
- **Consumers.** `server.py:57` holds a module-level `RegistryManager`, and `server.py:186-188` plus `validator.py` only iterate the returned lists. Fresh lists per load changes nothing for them.
- **Rollback.** One commit, no network, no Linear write, no permanent object. Plain revert.

Two things I deliberately did not escalate:

- **`save()` is a non-atomic read-modify-write with no lock.** Two processes adding instances concurrently lose one silently while `add_instance` returns the entry as if it stuck. Real, and it matches the "reports success for work that did not happen" anchor, but the diff does not touch `add_instance` or `save`, and round 1 already named it as pre-existing. Escalating unchanged code in round 2 is the miscalibration the brief warns about. Worth its own spillover row, not a gate on this PR.
- **The receipt at `.prd-os/receipts.jsonl` binds commit `169b4e29`, while the PR head is the merge `71b02972`.** I looked for a consumer that checks receipt sha against HEAD and found no `commit_sha` reference in `prd_runner.py`, so I could not construct a failure. Dropping it.

A note on the process, since this is round 2. Round 1 was Codex hitting usage limits and the claude fallback returning APPROVE with zero findings. I reproduced its core claims independently and they were correct, not lazy. This PR is 30 lines with a genuine reproducer and a genuine mutation proof. It did not need a second round.

## VERDICT: APPROVE WITH NITS

First thing to fix: merge current `main` and let the hook re-bump `plugin.json`, or auto-merge will hold this at DIRTY indefinitely with no alert.

FINDINGS:
minor|Branch conflicts with current main on the plugin.json version line while auto-merge is armed, so the PR stalls at DIRTY with nothing paging anyone|plugins/kipi-core/.claude-plugin/plugin.json:3
nit|PR description states the version bump as 1.12.6 to 1.12.7 while the diff is 1.13.19 to 1.13.20|plugins/kipi-core/.claude-plugin/plugin.json:3
END FINDINGS

Note on workspace hygiene: I created no tracked changes. `git status --porcelain` is empty. `uv run` built a gitignored `.venv` and `.pytest_cache` under `plugins/kipi-core/kipi-mcp/`, and the mutation ran entirely in `$TMPDIR`.

PR #366, opened 2026-09-17
test(guard): pin .claude-plugin pipeline reads both ways (ASK-733)

assafkip:
## What

ASK-733: claude-path-write-guard's pipeline check refused `git show <rev>:<plugin>/.claude-plugin/plugin.json | python3 ...` because of a raw `.claude` substring test.

## Finding

The guard fix itself is already on main. Line 901 of `claude-path-write-guard.py` uses the component-wise `_CLAUDE_PATH_TEXT` regex, which landed in 4e02b191 (ASK-984, sp-1d4ca360). This PR adds the pins the DoR asked for. No guard code changes.

## Changes

- `probe_guard.py`: two BENIGN `.claude-plugin/plugin.json` pipeline reads must be ALLOWED, and `git show HEAD:.claude/settings.json | python3 -c ...` must still be BLOCKED.
- `test_claude_path_write_guard.py`: the same two directions as unit tests.

## Evidence

- `python3 q-system/.q-system/tests/test_claude_path_write_guard.py` printed `claude-path-write-guard tests: PASS`.
- `python3 q-system/output/claude-changes/repro/probe_guard.py` printed `30/30 cases match`.

Mutation run on scratch copies of the guard and both pins. The real guard was not touched.
- Control (unmutated copy): both suites rc=0.
- Raw `".claude" in stmt` restored: test rc=1, and the probe shows `BENIGN MISMATCH exit=2 expected=0` on both plugin-manifest reads.
- Every `.claude` pipeline allowed (`if False and ...`): test rc=1 on the negative assertion, and the probe shows `ATTACK MISMATCH exit=0 expected=2` on the settings.json pipe (plus two xargs attacks).

## Out of scope (per DoR)

Bare-token cwd resolution (`git fetch` resolved to `<cwd>/fetch`).

Once this closes, resolve the ledger row with `prd_runner.py spillover resolve sp-7b3a7a95 --resolution-ref ASK-733`.


assafkip:
## Verdict: APPROVE

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-366-20260917-082444.md` (2558 bytes).

```
FINDINGS:
END FINDINGS
```

--- reviewer output ---

**Verdict: APPROVE.** None of my findings survived reproduction. PR #366 adds only tests (40 lines in 2 files, no guard code), and both new pins go red for the reason they claim.

## What I ran

- **Both suites at the PR head:** `test_claude_path_write_guard.py` printed `claude-path-write-guard tests: PASS`. `probe_guard.py` printed `30/30 cases match`.
- **pytest picks them up:** `pytest q-system/.q-system/tests/test_claude_path_write_guard.py -q` gave `6 passed` (the 4 existing tests plus the 2 new ones). CI runs `verify.sh --full`, which runs pytest over the tests directory, so the new tests have a caller.
- **Mutation run:** I made scratch copies of `claude-path-write-guard.py`, pointed the test module's `GUARD` at each copy, and confirmed afterwards that the real guard was unchanged (`real guard untouched: True`).
  - **Unmutated copy:** all tests pass.
  - **Raw `.claude` substring test put back at line 901:** `FAIL test_claude_plugin_pipeline_read_passes`. The allow-side pin catches the ASK-733 regression.
  - **Line 901 branch turned off (`if False and`):** `FAIL test_claude_dir_pipeline_into_interpreter_still_blocks`. The block-side pin catches an over-broad fix.
  - **`idx > 0` condition removed:** the new tests stay green. The new block-side pipe has both `idx > 0` and `-c`, so it can't tell the two apart. That condition is already covered by the existing xargs probe case, and this PR doesn't claim to cover it, so I dropped it.

## What is sound

- **The block-side test fails for the right reason.** The same pipe over a non-`.claude` path, `git show HEAD:foo/settings.json | python3 -c '...'`, returns 0 on the real guard. So the rc=2 comes from the `.claude/` path, not from `python3 -c` alone.
- **The two pins are independent.** Each mutation turns exactly one of them red, so neither passes only because of the other.
- **The test inputs are real commands.** The fixtures are `git show <rev>:<path> | python3 ...` pipelines, the same command the PR says the guard refused. They go through the guard's real stdin contract (hook payload in, exit 2 on block).
- **A wider variant holds on the real guard:** `git show HEAD:.claude/settings.json | python3 -m json.tool` is blocked (rc=2).
- **Blast radius:** tests only, so a squash revert rolls it back fully and nothing changes at runtime.

## Most important thing to fix first

Nothing gates this PR. After it merges, the spillover row can be resolved as the PR body says: `prd_runner.py spillover resolve sp-7b3a7a95 --resolution-ref ASK-733`.

FINDINGS:
END FINDINGS

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-366-20260923-075255.md` (6920 bytes).

```
FINDINGS:
minor|probe_guard.py additions are enforced by nothing; no CI job, hook or named pytest suite invokes it, and the CI-run unit suite stays green under a pipeline-branch regression only the probe catches|q-system/output/claude-changes/repro/probe_guard.py:133
minor|The .claude-plugin prefix exemption this PR pins as correct is escapable by shell suffix-stripping; the guard allows a command that writes .claude/rules/pwn.md|q-system/.q-system/scripts/claude-path-write-guard.py:103
END FINDINGS
```

--- reviewer output ---

## PR #366 round 2 — independent read

Tests only. 41 added lines, no guard code. I reproduced both pins from scratch before reading round 1.

---

### Finding 1 — minor

**Claim:** The three cases this PR adds to `probe_guard.py` are enforced by nothing; no CI job, hook, or named pytest suite invokes that file.

**Where:** `q-system/output/claude-changes/repro/probe_guard.py:133`

```
$ grep -rn "probe_guard" .github/ q-system/.q-system/ plugins/ | grep -v Binary
q-system/.q-system/scripts/claude-path-write-guard.py:131:# write the worktree, and probe_guard.py pins two of them blocked so nobody
q-system/.q-system/scripts/claude-path-write-guard.py:222:    splits -- its `;` is unquoted -- and probe_guard.py pins that.
q-system/.q-system/scripts/claude-path-write-guard.py:591:    payload between the delimiters. Pinned by two cases in probe_guard.py: a
q-system/.q-system/scripts/claude-path-write-guard.py:919:    # puts the sink at idx > 0. probe_guard.py pins all three.
q-system/.q-system/scripts/claude-path-write-guard.py:1598:    # touching .claude>"` stays blocked, and probe_guard.py pins it.
```

Every hit is a prose comment asserting a pin. Zero runners. The named suites in `.verify-suites` do not include `q-system/output/...`, and the file is not `test_*.py`, so pytest would not collect it even if the directory were named:

```
$ python3 -m pytest q-system/.q-system/tests --collect-only -q | grep -c "test_claude_path_write_guard"
6
$ python3 -m pytest q-system/.q-system/tests --collect-only -q | grep -c "probe_guard"
0
```

**The consequence, measured.** I mutated the pipeline branch on a scratch copy of the guard, dropping the `idx > 0 or INLINE_CODE_FLAGS` discrimination at line 920 (`if True:`), and ran both suites:

```
CONTROL: unit rc=0 claude-path-write-guard tests: PASS | probe rc=0 30/30 cases match
M3 pipeline always blocks -> unit rc=0 claude-path-write-guard tests: PASS | probe rc=1 MISMATCH BENIGN: python3 plugins/prd-os/scripts/prd_runner.py spillover add --source ASK
```

The regression class the line-910 scar exists for (false-blocking `prd_runner.py spillover add`) is caught **only** by the uncalled probe. The CI-run suite reports PASS. Round 1 verified the pytest caller for the unit tests and never asked who runs the probe, so this is new.

**Fix shape:** move the uniquely-probe-only case (the `python3 -c 'json.load(sys.stdin)'` benign read) into `test_claude_path_write_guard.py`, or name the probe's directory in `.verify-suites` after renaming it to `test_*.py`. Not a gate.

---

### Finding 2 — minor

**Claim:** The `.claude-plugin` prefix exemption this PR pins as correct is escapable by shell suffix-stripping; the guard allows a command that writes into `.claude/`.

**Where:** `q-system/.q-system/scripts/claude-path-write-guard.py:103` (`_CLAUDE_PATH_TEXT`), unchanged by this PR

```
$ python3 -  # drives the real guard through its stdin hook contract
rc=0  D=.claude-plugin; touch ${D%-plugin}/rules/pwn.md
rc=0  P=.claude-plugin; touch ${P/-plugin/}/rules/pwn.md
rc=2  touch .claude-plugin/../.claude/rules/pwn.md
rc=2  git show HEAD:.claude-plugin/x | tee .claude/rules/pwn.md
```

The allowed command's real expansion:

```
$ bash -c 'D=.claude-plugin; printf "%s\n" "${D%-plugin}/rules/pwn.md"'
.claude/rules/pwn.md
```

The suite already treats variable-indirection-to-`.claude` as an attack class (`ROUND9_PHASE1` pins `P=.claude; V=P; touch ${!V}/rules/pwn.md`). The suffix-strip variant escapes because the literal text never carries a bare `.claude`.

**Why minor, not major:** this is pre-existing guard code from 4e02b191, not touched here. The input is one I constructed and no producer in this repo emits it. Layer 2 (`claude-integrity-tripwire.py`) is the stated backstop and would detect the write after the fact. I flag it against this PR only because the PR's purpose is to bless that exemption as correct, and it pins the exemption without pinning that it cannot be weaponized. Worth a spillover row, not a block.

---

### Dropped — could not reproduce

**GitHub reports the PR unmergeable while auto-merge squash is armed.** `gh api repos/assafkip/kipi-system/pulls/366` returns `mergeable: false, mergeable_state: dirty`. My own merge against the same main sha is clean:

```
$ git merge-tree --write-tree --messages b5bc2ce1 f8009d50
065cc23733fa9e5b232011fcadb959cc9b70ef95

Auto-merging .prd-os/receipts.jsonl
rc=0
```

Local says clean, GitHub says dirty. I could not make the conflict appear, so I am not filing it. Worth one `git fetch && git merge origin/main` before trusting auto-merge.

---

### What is sound

**Both pins are load-bearing, and I proved it without reading round 1.** My own mutation harness, scratch copies only, real guard untouched:

- Control (unmutated copy): all four cases match expectation.
- Pipeline branch reverted to raw `".claude" in stmt`: both BENIGN plugin reads flip to rc=2. The allow-side pin is real.
- Pipeline branch disabled (`if False:`): the `.claude/` pipe AND the xargs attack flip to rc=0. The block-side pin is real, and it blocks **because of** the pipeline branch, not by accident of some other rule.

**The negative pin is not passing for the wrong reason.** With the branch disabled the command returns 0, so nothing else in `_stage()` is silently catching it. The rule at line 1624 cannot fire here: its segment test needs `.claude` text inside the `python3 -c` stage, and the payload is `import sys; sys.stdin.read()`.

**The fixture path is real.** `plugins/prd-os/.claude-plugin/plugin.json` exists on disk, so the command shape matches what an operator actually types.

**Adjacent bypasses I tried and the guard survived:** `.claude-plugin/../.claude/...` traversal (rc=2), a `tee` sink writing `.claude/` downstream of a plugin read (rc=2), and a global mutation of the lookahead regex to a raw substring, which goes red in both suites.

**Rollback:** tests only. A squash revert removes it entirely and nothing changes at runtime.

---

**Most important thing to fix first:** give the probe's unique case a caller. Right now the guard's own comments claim `probe_guard.py` pins three attack shapes, and nothing on this machine or in CI ever runs it, which is the exact "claim of enforcement" shape this repo has a `.verify-suites` scar for.

FINDINGS:
minor|probe_guard.py additions are enforced by nothing; no CI job, hook or named pytest suite invokes it, and the CI-run unit suite stays green under a pipeline-branch regression only the probe catches|q-system/output/claude-changes/repro/probe_guard.py:133
minor|The .claude-plugin prefix exemption this PR pins as correct is escapable by shell suffix-stripping; the guard allows a command that writes .claude/rules/pwn.md|q-system/.q-system/scripts/claude-path-write-guard.py:103
END FINDINGS

PR #369, opened 2026-09-18
fix(capability-gate): repo-wide test discovery (ASK-541)

assafkip:
## What

`capability-gate.py` only looked for tests under the two `q-system/.q-system` scan roots. An undeclared test in `plugins/`, `q-system/.q-system/tests/`, the repo root or `automation/` left the gate GREEN.

- Discovery now combines `git ls-files` over the whole repo (without git, it falls back to a pruned filesystem walk) with the old walk of the v1 roots, which also catches untracked files.
- `uncovered_known` now accepts `{path, reason}` entries: a test that is declared and deliberately NOT run. An entry clears that exact path only, and it goes RED in the skeleton when the file is gone. 145 fragments at cut-over, so no test runs for the first time (the DoR's not-doing line).
- Instance mode: an undeclared test outside the v1 roots becomes a note, never RED, because the fleet updater runs this gate in every instance.
- Removed the `scope_exempt` fragments, since every declared path is now discoverable. Removed the old `plugins/*/tests` prose entry. The other two prose entries are untouched.
- `capability_manifest.fragment_name`: a path starting with a dot produced a hidden fragment file, and `load()` skipped it without saying so. Found while declaring the `.review-scratch/` paths.

## Evidence

- Reproducer `test_capability_gate.py --only repo-wide`: 9 of 11 FAIL before the fix, 11 of 11 PASS after. Full suite: ALL PASS.
- Declared-vs-actual diff on this tree: 145 undeclared before triage, 0 errors after.
- About the DoR count: its grep returns 322 today, but 132 of those are under `scripts/**`, which v1 already walked. The real blind population was 190: 45 already declared plus 145 new.

## Known, not fixed here

In the skeleton checkout, a local run of the full gate is RED (12) on `declared_inert` rows whose spillover ids are RESOLVED. This was already on main, and its early exit skips the diff. Captured as sp-2b6d4ff6 / ASK-1790. CI has no ledger, so CI runs the full diff.


assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-369-20260918-090312.md` (5208 bytes).

```
FINDINGS:
major|Declared tests outside the v1 roots that exist on disk but are untracked in git are now reported declared-but-missing, so the post-sync kipi update gate (which runs before synced files are committed) turns every instance RED on the first ship of a new plugin or tests-dir test|q-system/.q-system/scripts/capability-gate.py:618
minor|Scan-scope note says all declared entries are checked BOTH directions, but in instance mode the undeclared direction outside the v1 roots is report-only|q-system/.q-system/scripts/capability-gate.py:1068
END FINDINGS
```

--- reviewer output ---

**VERDICT: REQUEST CHANGES.** One major finding: the post-sync `kipi update` gate goes RED across the fleet whenever a newly declared test outside the old roots ships. It is not a BLOCK, because the RED is recoverable and clears once the instance commits.

## Findings

**1. MAJOR: a declared test that exists on disk but is not committed in git reports "declared-but-missing"**

- **Where:** `q-system/.q-system/scripts/capability-gate.py:618`
- **Claim:** `in_scan_scope()` now counts any test-pattern name anywhere. Outside the v1 roots, though, discovery only sees files git already tracks. So a declared test that sits on disk uncommitted in plugins/, q-system/.q-system/tests/ or q-system/hooks/test/ is reported as "declared-but-missing". On main the same entry took the `is_file()` branch and was GREEN.
- **Real producer:** `kipi-update.sh:2348` commits system state *before* the skeleton rsync. Nothing commits after the rsync, and the post-sync gate runs at `kipi-update.sh:2858-2866` and adds the instance to `GATE_FAIL`. So on the first `kipi update` that ships a new declared test under `plugins/` or `q-system/.q-system/tests/`, that file is untracked in every instance. That turns the whole fleet RED and makes the updater exit non-zero for a file that exists.
- **Reproducer:** inline Python run from the review tree. It builds a fixture with `test_capability_gate.make_repo`, declares `plugins/demo/tests/test_a.py`, writes the file, runs `git init` and stages only `q-system`, then runs each ref's `capability-gate.py --check-only`.
- **Real output:**
  - `origin/main instance (0, [])`
  - `HEAD instance (1, ['  RED: declared-but-missing: plugins/demo/tests/test_a.py', 'capability-gate: RED (1)'])`
  - The skeleton runs also showed two `inert-engine` REDs. Those came from my copying the gate into the fixture, so I left them out.
- **Steady state is fine.** I ran the PR's `discover_tracked` read-only against the 26 registered instances. Every one showed 0 declared tests that were present but untracked. The damage is limited to the window right after a new test ships, which is exactly when the gate runs.
- **Fix to try first:** for a path outside the old roots, when the declared path `is_file()`, count it as present, the way main did. Another option is to add an untracked-inclusive check such as `git ls-files --others` for declared paths only.

**2. MINOR: the scan-scope note claims both directions are checked where instance mode only reports one**

- **Where:** `q-system/.q-system/scripts/capability-gate.py:1068`
- **Claim:** in instance mode the note says all 235 entries are "checked BOTH directions". But outside the v1 roots, an undeclared test in an instance is only a report-only note, per line 633.
- **Reproducer:** `git archive HEAD`, keeping only `q-system/`, `plugins/` and `.claude/`, with no `instance-registry.json`, then `git init` and `git add -A -f`, then run the gate with `--check-only`.
- **Real output:** `scan scope: 235 declared entries inside the scan roots (checked BOTH directions), 0 exempt from undeclared-artifact detection (existence-checked only)`

## Dropped after trying

- **Clean-checkout RED on the `.review-scratch/...` uncovered_known entry.** My first simulation went RED on it. The cause: `.gitignore:162` ignores `.review-scratch/`, but the file is force-tracked, and my `git add -A` skipped it. After a force-add, the PR skeleton is GREEN (rc 0, 145 known-uncovered), which matches the PR body.
- **Fleet regression from skeleton_only paths.** All 21 `skeleton_only` entries sit inside the old roots, so the `declared-but-missing` loop that skips nothing does not newly hit them.

## What is sound

- Full suite `test_capability_gate.py`: ALL PASS, including the 11 new repo-wide cases.
- Clean-clone skeleton at the PR: GREEN. Clean-clone main: also GREEN. Instance simulation at the PR: GREEN, with 121 known-uncovered.
- `uncovered_known` clears one path only. The case that plants a second file in the same directory goes RED.
- `git ls-files -z` handles odd filenames. The `is_file()` filter stops a file that is tracked but deleted from being counted as present.
- The `fragment_name` leading-dot fix is safe for existing fragments. `load()` checks every fragment's filename against `fragment_name()`, and the PR tree loaded without any mismatch error.
- The pre-commit path is fine. A newly staged test anywhere is in the index, so the repo-wide pass sees it.

## Fix first

Finding 1. Outside the old roots, a declared file that exists on disk has to count as present. Otherwise the next `kipi update` that ships a plugin test turns every instance RED.

FINDINGS:
major|Declared tests outside the v1 roots that exist on disk but are untracked in git are now reported declared-but-missing, so the post-sync kipi update gate (which runs before synced files are committed) turns every instance RED on the first ship of a new plugin or tests-dir test|q-system/.q-system/scripts/capability-gate.py:618
minor|Scan-scope note says all declared entries are checked BOTH directions, but in instance mode the undeclared direction outside the v1 roots is report-only|q-system/.q-system/scripts/capability-gate.py:1068
END FINDINGS

assafkip:
Rework for the REQUEST CHANGES review, commit 0e66d6d8.

- **major (fixed):** a declared path that is on disk counts as present even when git does not track it, as on main. Only the index missing it no longer reads as declared-but-missing, so the post-sync gate stays GREEN on the first ship of a new plugin/tests-dir test.
- **minor (fixed):** instance-mode scan-scope note now counts entries outside the v1 roots separately as report-only instead of claiming BOTH directions.

Evidence: `test_capability_gate.py --only repo-wide` cases 9-10 (the reviewer's reproducer as a fixture, plus a control where the declared file is deleted and stays RED): 2 FAIL before the fix, ALL PASS after. Full suite ALL PASS.

Unrelated, pre-existing: the skeleton's own `capability-gate.py --check-only` is RED (12) on stale declared_inert spillover citations, identical on the pre-fix gate. Filed as ASK-1791 (sp-374fb855).

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-369-20260918-091024.md` (5103 bytes).

```
FINDINGS:
minor|In instance mode a skeleton_only test outside the v1 roots is now treated as in-scope and hits declared-but-missing without the skeleton_only skip, so promoting a repo-root kipi-update test turns every instance RED while skeleton and CI stay GREEN|q-system/.q-system/scripts/capability-gate.py:624
minor|The stale uncovered_known check uses git-index-only discovery outside the v1 roots, so an untracked file on disk is reported as "the file is gone; remove its fragment", which is false and gives the wrong fix|q-system/.q-system/scripts/capability-gate.py:634
END FINDINGS
```

--- reviewer output ---

**Verdict: APPROVE WITH NITS.** Two minor findings survived reproduction. Neither affects the tree as it stands. The first one to fix is the instance-mode skeleton_only skip.

## Finding 1 (minor): instances lose the skeleton_only skip for test files outside the old roots

**Claim.** Every test-named path now lands in `in_scope_declared` (`q-system/.q-system/scripts/capability-gate.py:618`). The instance-mode skeleton_only skip only runs in the out-of-scope loop at line 656, so the in-scope missing loop at line 624 never consults it. A skeleton-only test outside the v1 roots is absent from every instance by design, and now it turns every instance RED.

**Why it's minor and not major:**
- The fleet-wide false alarm would be major, but no producer emits this input today. All 21 skeleton_only fragments sit under `q-system/.q-system/scripts/`, so I dropped one level.
- The trigger is on the path the PR itself advertises. The 145 uncovered_known fragments say "Promote to expected_tests once seen green", and the repo-root `test-kipi-update-*.sh` suite is exactly what the PR describes as "not synced" to instances.
- Once one of those is promoted as skeleton_only, the skeleton goes GREEN, CI goes GREEN, and `kipi update` (`kipi-update.sh:2861`) marks every instance RED.

**Reproducer.** I ran it inline with `python3 -c`. It loads each version's `test_capability_gate.py` helpers, then builds an `expected_tests` + `skeleton_only` + `scope_exempt` manifest for `test-kipi-update-safety-root.sh`. The skeleton case has the file; the instance case does not.

Real output:
```
main skeleton rc= 0              capability-gate: GREEN
main instance(file absent) rc= 0 capability-gate: GREEN
PR skeleton rc= 0                capability-gate: GREEN
PR instance(file absent) rc= 1   RED: declared-but-missing: test-kipi-update-safety-root.sh
```

**Fix.** Apply the same `mode == "instance" and p in skeleton_only` skip to the loop at line 624.

## Finding 2 (minor): the stale-entry error says "the file is gone" when the file exists but is untracked

**Claim.** In skeleton mode, `known - discovered` (`capability-gate.py:634`) uses discovery that is limited to the git index outside the v1 roots. An uncovered_known entry for a file that is on disk but not yet `git add`-ed gets "the file is gone; remove its fragment". Following that advice would then turn into present-but-undeclared after the add.

The PR already fixed this exact asymmetry for declared tests with the `(root / p).is_file()` check at line 622. The uncovered_known direction didn't get it.

**Reproducer.** Skeleton repo. `git_track` of `q-system` only, then `add_test("plugins/demo/tests/test_new.py")` with a matching uncovered_known entry.

Real output:
```
file on disk: True
PR skeleton, untracked-on-disk uncovered_known: rc= 1
   RED: uncovered_known names no test artifact: plugins/demo/tests/test_new.py — the file is gone; remove its fragment.
after git add: rc= 0
```

It's bounded: pre-commit and CI see the staged file, so this only bites a local run before `git add`.

- **Real-tree diff, both modes.** I called `diff_declared_vs_actual` directly to get past the pre-existing declared_inert early exit. Skeleton: 0 errors. Instance: 0 errors. 145 known-uncovered, 0 report-only notes. The PR's "0 errors after triage" claim holds.
- **The tests can fail.** I reverted `discover_tests` to legacy roots only. Six repo-wide cases went RED, including both the fs-walk and git-index reproducers. The rest passed on that mutant because it only removes repo-wide discovery, and those cases test other paths.
- **Git and the no-git walk agree.** On the real tree both find 380 artifacts, with nothing found by only one of them. The fallback doesn't pull in scratch or venv noise here.
- **Main vs PR.** Without scope_exempt, main already refused a repo-root declaration. So finding 1 is a regression only for the exempt-plus-skeleton_only shape, which I tested explicitly.
- **Full suite.** `test_capability_gate.py` prints ALL PASS. The full gate is RED only on the 12 declared_inert rows the PR already names (sp-2b6d4ff6 / ASK-1790).
- **The leading-dot fragment-name fix is real.** The `.review-scratch` and `.review-tmp-pr11` fragments load as non-hidden hashed names.

Two MCP servers need authorization before they work: claude.ai Slack (via claude.ai connector settings) and plugin:vercel:vercel (via `/mcp` in an interactive session). Neither was needed for this review.

FINDINGS:
minor|In instance mode a skeleton_only test outside the v1 roots is now treated as in-scope and hits declared-but-missing without the skeleton_only skip, so promoting a repo-root kipi-update test turns every instance RED while skeleton and CI stay GREEN|q-system/.q-system/scripts/capability-gate.py:624
minor|The stale uncovered_known check uses git-index-only discovery outside the v1 roots, so an untracked file on disk is reported as "the file is gone; remove its fragment", which is false and gives the wrong fix|q-system/.q-system/scripts/capability-gate.py:634
END FINDINGS

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-369-20260923-070604.md` (7327 bytes).

```
FINDINGS:
minor|In instance mode a skeleton_only test outside the v1 roots is treated as in-scope and hits declared-but-missing without the skeleton_only skip, so promoting a repo-root kipi-update test turns every instance RED while skeleton and CI stay GREEN|q-system/.q-system/scripts/capability-gate.py:622
minor|The stale uncovered_known check is git-index-only outside the v1 roots, so an untracked file on disk is reported as "the file is gone; remove its fragment", which is false and gives a fix that makes it present-but-undeclared|q-system/.q-system/scripts/capability-gate.py:632
minor|uncovered_known exempts 145 paths permanently with no expiry and no spillover pointer, unlike quarantine and declared_inert in the same file, so the boilerplate "promote once seen green" reason is never read by any code|q-system/.q-system/scripts/capability-gate.py:286
END FINDINGS
```

--- reviewer output ---

Read the code first, then rounds 1 and 2. Both round-2 minors reproduce unchanged on HEAD (`ca6cbdd9`). Nothing new at major or above.

## Findings

### 1. MINOR: instance mode lost the `skeleton_only` skip for any test outside the v1 roots

`q-system/.q-system/scripts/capability-gate.py:622`

`in_scan_scope()` now returns True for any test-pattern name anywhere, so a `skeleton_only` path outside `q-system/.q-system/scripts/` lands in `in_scope_declared` and hits the missing-loop at line 622. That loop never consults `skeleton_only`. The skip only exists in the out-of-scope loop at line 660, and the comment there names the exact scar it is protecting (codex, PR #216).

Reproducer, run from the review tree, fixture built from `capability_manifest.fragment_name` so the fragment names come from the real producer:

```
=== REPRO 1: skeleton_only test outside the v1 roots ===
INSTANCE (file absent by design) rc= 1
    capability-gate mode=instance
      RED: declared-but-missing: test-kipi-update-root.sh
    capability-gate: RED (1)
SKELETON (file present) rc= 1 ['capability-gate: RED (2)']
```

Skeleton `RED (2)` is fixture noise, the two `inert-engine` rows from copying the gate into the fixture tree. The instance `RED (1)` is the finding, clean.

Staying at minor, same as round 2, and for the same reason: all 21 `skeleton_only` fragments today sit under `q-system/.q-system/scripts/`, so no producer emits this input yet. The trigger is the path this PR advertises, promoting one of the repo-root `test-kipi-update-*.sh` rows out of `uncovered_known`. At that moment skeleton and CI stay GREEN and `kipi-update.sh:2861` marks every instance RED.

Fix: apply the `mode == "instance" and p in skeleton_only` skip to the loop at line 622.

### 2. MINOR: the stale `uncovered_known` error says "the file is gone" while the file is on disk

`q-system/.q-system/scripts/capability-gate.py:632`

The PR fixed exactly this asymmetry for declared tests at line 620 with `(root / p).is_file()`. The `known - discovered` loop did not get it, so outside the v1 roots it is git-index-only.

```
=== REPRO 2: uncovered_known path on disk but untracked, SKELETON ===
file on disk: True
rc= 1
    capability-gate mode=skeleton
      RED: uncovered_known names no test artifact: plugins/demo/tests/test_new.py — the file is gone; remove its fragment.
after git add rc= 1 ['capability-gate: RED (2)']
```

`RED (3)` before the add drops to `RED (2)` after it, and the remaining two are the fixture's `inert-engine` rows. So the stale row cleared purely on the `git add`. The message gives the wrong fix: deleting the fragment turns the file into `present-but-undeclared` on the next run. Bounded to a local pre-`git add` run.

### 3. MINOR: `uncovered_known` is a permanent silence with no liveness pointer, unlike both of its siblings

`q-system/.q-system/scripts/capability-gate.py:286`

`validate_uncovered_known` checks that `reason` is non-empty and nothing ever reads it again. No expiry, no spillover id. The two other exemption mechanisms in this same file both carry one and both are enforced:

```
quarantine expiry enforced: True
declared_inert liveness enforced: True
distinct uncovered_known reason prefixes: 1
    Declared, deliberately NOT run by the capability gate: surfa
```

145 fragments, one identical boilerplate reason, and the promised behavior lives only in that string: "Promote to expected_tests once seen green." No code path promotes, nags, or expires. The only thing that ever revisits an entry is deleting the file. Net effect: this PR widens discovery to 380 artifacts and simultaneously ships a 145-row forever-exemption whose stated review plan is unenforced prose. Nothing breaks at 3am, which is why it is minor, but it is the same silent-absence class the gate exists to end.

Fix worth considering: require a spillover id per entry, the way `declared_inert` does, so the ratchet has something to close.

- **Full suite is green and the 11 new repo-wide cases are real.** `python3 q-system/.q-system/scripts/test_capability_gate.py` prints `ALL PASS`.
- **Instance mode on a full post-sync tree is GREEN.** `git archive HEAD` into tmp, registry removed, `git init` and commit, gate with `--repo-root`: `EXIT 0`, `capability-gate: GREEN`, 144 known-uncovered, 0 report-only notes, 0 REDs.
- **The round-2 major is genuinely fixed.** A declared path on disk but untracked no longer reads as declared-but-missing. I re-ran the declared-but-missing path in repro 2's tree and it stayed silent.
- **The round-2 minor about the scan-scope note is fixed and honest.** The instance run prints `190 declared entries inside the scan roots (checked BOTH directions), 45 outside the v1 roots (undeclared direction report-only in an instance)`, which matches line 622 and line 637.
- **Walk fallback and git agree on this tree.** `discover_tracked` 380, `discover_walk` 380, `walk-only: 0`. I was hunting for venv and scratch noise in the no-git path and found none here.
- **`.review-scratch/` is force-tracked, not a phantom.** I suspected the two scratch `uncovered_known` rows named untracked founder-local paths, which would turn every clean clone RED. `git ls-files` shows them tracked. Dropped.
- **The `skeleton_only` regression is not live today.** I checked all 21 rows against `in_legacy_roots`: every one is True, so the current manifest does not hit finding 1. That is why it stayed minor rather than moving up.
- **The leading-dot `fragment_name` fix does not rename existing fragments.** `load()` re-derives every name and the whole tree assembles with no mismatch error.
- **The gate's own early exit does not hide the new diff in CI.** The instance and CI-shaped runs print `inert-spillover-liveness: SKIPPED, no ledger`, so they reach `diff_declared_vs_actual`. The PR body's claim holds.

## On the review process

Round 3 produced zero blockers and zero majors on unchanged code. Round 1's major was reproduced, fixed and re-verified; round 2's two minors are still open only because APPROVE WITH NITS tracks rather than gates. Findings are shrinking round over round, which is what a converging PR looks like. No miscalibration to report.

Fix first: finding 1, the `skeleton_only` skip at line 622. It is the one that turns the whole fleet RED the moment someone follows the promotion path this PR itself advertises.

FINDINGS:
minor|In instance mode a skeleton_only test outside the v1 roots is treated as in-scope and hits declared-but-missing without the skeleton_only skip, so promoting a repo-root kipi-update test turns every instance RED while skeleton and CI stay GREEN|q-system/.q-system/scripts/capability-gate.py:622
minor|The stale uncovered_known check is git-index-only outside the v1 roots, so an untracked file on disk is reported as "the file is gone; remove its fragment", which is false and gives a fix that makes it present-but-undeclared|q-system/.q-system/scripts/capability-gate.py:632
minor|uncovered_known exempts 145 paths permanently with no expiry and no spillover pointer, unlike quarantine and declared_inert in the same file, so the boilerplate "promote once seen green" reason is never read by any code|q-system/.q-system/scripts/capability-gate.py:286
END FINDINGS

PR #371, opened 2026-09-18
fix(verify): --staged runs the tests that own the change (ASK-1795)

assafkip:
verify.sh --staged ran a whole manifest suite on any staged path under it: ~6300 tests, 620s and 788s per q-consult commit in consulting (2026-09-18).

- New `verify_select.py`: staged test file -> itself; staged module -> tests naming it; staged data file -> tests naming its basename; unowned path -> `<suite>/.verify-fallback`, else the full suite. A staged pytest config -> full suite.
- Pruning happens at collection through a pytest plugin, so conftest `collect_ignore` still holds (explicit file args bypass it, measured on pytest 9.0.3).
- The selection and the reason per staged path are printed.
- `--full` is unchanged: pre-push and CI still run everything.

Adversarial suite 40/40 (13 new cases). Mutants always-full, never-prune and drop-fallback each turn cases red.

Local `--full`: one failure, `test_destructive_op_deny_anchor.py::test_the_vendored_copy_has_not_drifted`, which compares the vendored fixture against this machine's installed hook. Unrelated to this diff.


assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-371-20260918-130520.md` (4297 bytes).

```
FINDINGS:
minor|A staged rename lists only the new path in ANY_STAGED, so tests that still import the old module name are skipped whenever a .verify-fallback is declared, which contradicts the "deleted module still selects" claim|q-system/.q-system/verify.sh:82
minor|An __init__.py at the suite root is matched as the literal "__init__" instead of the package name, so voiceloop/__init__.py selects 1 of 10 tests that name the package, which contradicts the docstring|q-system/.q-system/verify_select.py:81
END FINDINGS
```

--- reviewer output ---

**Verdict: APPROVE WITH NITS.** Two minor gaps in the `--staged` selector, both reproduced. Neither lets bad code reach main, because `--full` at pre-push and in CI is unchanged. The first one to fix is the rename gap.

## Findings

**1. Minor: a staged rename hides the old module name, so tests that still import it can be skipped at pre-commit**
- Where: `q-system/.q-system/verify.sh:82`. `ANY_STAGED` comes from `git diff --cached --name-only`, and git's rename detection is on by default. It lists only the new path.
- This breaks a promise in `verify_select.py:32-33`: "a deleted module still selects the tests that name it". A rename is a deletion, and the old name never reaches the selector.
- Blast radius: it only happens once a suite declares a `.verify-fallback`, which is the consulting use case this PR is for. With no fallback, the unowned new name sends the whole suite, and the break is caught.
- Reproducer: a Python heredoc run against a throwaway git repo. It has `suite/helper.py`, `tests/test_helper.py` which imports helper, and a `.verify-fallback` that lists only `tests/test_other.py`. Then it runs `git mv helper.py helper2.py`. Real output:
  - ANY_STAGED: ['suite/helper2.py']
  - helper2.py: no test names 'helper2' -> fallback
  - select, tests/test_other.py only. The broken `test_helper.py` is not run.
  - Same index read with `-c diff.renames=false`: both names are listed, and `tests/test_helper.py` is selected.
- Fix: add `--no-renames` to the `ANY_STAGED` line. The only other use of `ANY_STAGED` is the empty check at line 84.

**2. Minor: an `__init__.py` at the suite root is matched as the literal `__init__`, not as the package name**
- Where: `q-system/.q-system/verify_select.py:81-83`. The path is relative to the suite, so `dirname("__init__.py")` is empty. The fallback is then the string `__init__`.
- This contradicts the docstring at line 17: "or the package name for an __init__.py".
- Real repo, real output:
  - Staging `plugins/kipi-core/voiceloop/__init__.py` gives "named as '__init__' by 1 test file(s)" and selects only `tests/test_engine_surface.py`.
  - All 10 test files in that suite name `voiceloop`.
  - `reddit_arctic/__init__.py` gets the same wrong reason, and by luck lands on the right single test.
- Blast radius today: none measured. I appended `raise ImportError` to a copy of the voiceloop `__init__.py`, then ran pytest in two ways. The selected run exits 1 and the full suite exits 2, so pytest's package import catches the break either way. That is why this ranks minor. The wrong-name selection is still real.
- Fix: when `rel == "__init__.py"`, use the suite's own basename, or send the full suite.

## What is sound
- **Adversarial suite:** ran `test_verify_adversarial.sh` against this PR's `verify.sh`. Result: 40 passed, 0 failed. It is wired into CI at `.github/workflows/verify.yml:67`.
- **Empty selection:** it falls back and never passes on zero tests. A deleted test file alone, or an empty set from the plugin, ends in the full suite or the exit-5 rerun.
- **`collect_ignore` still holds:** pruning happens in `pytest_ignore_collect`, and the hook returns `None`, not `False`, on directories that hold selected files. The adversarial case "collect_ignore still honoured" is green.
- **Two runs at once:** `TMP` is a fresh `mktemp -d` per run (`verify.sh:107`), so the plugin copy and the list files cannot collide.
- **Broken selector:** a missing selector or a nonzero exit sends the full suite. The case "missing selector -> full suite" is green.
- **Capability-gate glob change:** `test_capability_gate_inert_closure.py` passes 4 of 4.
- **One-interpreter syntax and JSON checks:** `dont_inherit=True` is present. The compiler-only syntax cases and the BOM case are still green.

FINDINGS:
minor|A staged rename lists only the new path in ANY_STAGED, so tests that still import the old module name are skipped whenever a .verify-fallback is declared, which contradicts the "deleted module still selects" claim|q-system/.q-system/verify.sh:82
minor|An __init__.py at the suite root is matched as the literal "__init__" instead of the package name, so voiceloop/__init__.py selects 1 of 10 tests that name the package, which contradicts the docstring|q-system/.q-system/verify_select.py:81
END FINDINGS

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `codex`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/codex/assafkip_kipi-system__pr-371-20260918-131023.md` (4441 bytes).

```
FINDINGS:
minor|verify.sh pipes the selector output through head -1 under pipefail, so a selection over about 20 KB SIGPIPEs and kills the pre-commit with exit 141 and no message, which the sed-never-head comment a few lines below already warns against|q-system/.q-system/verify.sh:477
minor|A staged pytest config outside the suite dir (the repo-root pytest.ini that prd-os inherits) does not force the full suite, so a config change plus one module runs only that module's owners, contradicting the docstring|q-system/.q-system/verify_select.py:119
END FINDINGS
```

--- reviewer output ---

**Verdict: APPROVE WITH NITS.** Two minor findings, both reproduced. Nothing blocker or major survived. The first thing to fix is the `head -1` pipe at `q-system/.q-system/verify.sh:477`.

## Findings

**1. MINOR: `head -1` can kill verify.sh with exit 141 and no message**
- **Where:** `q-system/.q-system/verify.sh:477`, the line `_sel_mode="$(printf '%s\n' "$_sel_out" | head -1)"`.
- **What happens:** the script runs under `set -euo pipefail`. When the selector's output is large, `head` exits early, `printf` gets SIGPIPE, and the whole run exits 141.
- **The comment disagrees with the code:** a few lines further down, the comment says "sed -n, never `| head`" because of this exact scar. Line 477 still uses `head`.
- **Reproducer I ran:** the same assignment inside `bash -c 'set -uo pipefail; ...'`, five tries at each size.
  - 17,398 bytes: rc=0 in 5 of 5.
  - 27,898 bytes: rc=141 in 3 of 5.
  - 56,499 bytes: rc=141 in 4 of 5.
  - A separate `set -e` run died at 80,899 bytes and printed exit=141.
- **Blast radius:** it fails closed. The commit is refused, but it is refused with no explanation.
- **Why minor and not major:** it has not happened yet. Consulting's q-consult has 388 test files. Selecting all of them gives about 19,429 bytes, which is under the point where I saw failures. Roughly 50% more test files would make it intermittent.
- **Fix:** `sed -n 1p`, which is the pattern the next block already uses.

**2. MINOR: a staged repo-root `pytest.ini` narrows the run instead of running the full suite**
- **Where:** `q-system/.q-system/verify_select.py:119`.
- **What happens:** the config check only looks at paths inside the suite. The prd-os suite has no config file of its own, so pytest uses the repo-root `pytest.ini`. That config change is not treated as "full suite".
- **Reproducer I ran:** piped `pytest.ini` and `plugins/prd-os/scripts/prd_runner.py` into `verify_select.py --suite plugins/prd-os`.
  - Output: `select`, owners only, "named as 'prd_runner' by 19 test file(s)".
  - Control: staging `plugins/prd-os/tests/conftest.py` gives `full`, with "pytest config changes every test".
- **Before this PR** the same commit ran all of prd-os. The docstring says a staged pytest config means the full suite, and that is only true for configs inside the suite.
- **Why minor:** `--full` at pre-push catches it.
- **Fix:** treat a staged config file in the suite's parent directories, up to the repo root, as a full-suite trigger too.

**Dropped:** a bad `# coding:` cookie makes `tokenize.open` throw outside the try block, which would print a traceback and skip the remaining files. The verdict would still be red. The sandbox blocked the temp-file write I needed to run it, so I dropped it unproven.

## What is sound
- **Adversarial suite:** `test_verify_adversarial.sh` passes 42 of 42 on this branch. That covers collect_ignore still being honoured, the rename case, `__init__.py`, a missing selector falling back to the full suite, a red fallback blocking, and `--full` running everything.
- **Selector reads the snapshot:** the staged tree is a real git worktree, so `git ls-files` sees exactly what the commit contains.
- **Pruning in the plugin:** a directory that holds a selected file returns None, not False, so conftest `collect_ignore` still gets consulted. Path comparison uses realpath on both sides, which handles the macOS `/var` to `/private/var` symlink.
- **Failure paths fail safe:**
  - A selector that errors or is missing gives the full suite.
  - An empty collection (exit 5) re-runs the full suite.
  - A fallback file that names nothing runnable is treated as absent.
- **Capability gate:** the `q-system/.q-system/*.sh` glob addition passes its tests, 7 of 7 (`-k capability`).
- **Concurrency:** the plugin dir and the selection list both live under the per-run `mktemp -d`, so two runs never share them.

FINDINGS:
minor|verify.sh pipes the selector output through head -1 under pipefail, so a selection over about 20 KB SIGPIPEs and kills the pre-commit with exit 141 and no message, which the sed-never-head comment a few lines below already warns against|q-system/.q-system/verify.sh:477
minor|A staged pytest config outside the suite dir (the repo-root pytest.ini that prd-os inherits) does not force the full suite, so a config change plus one module runs only that module's owners, contradicting the docstring|q-system/.q-system/verify_select.py:119
END FINDINGS

PR #374, opened 2026-09-19
design-chain: the gate measures the work instead of believing the account (ASK-1796)

assafkip:
Rebuilds the design chain from `rca-design-chain-trusts-its-own-account-2026-09-18`, on a clean branch off main. The gate verified the builder's account of the work: a round whose readers all said "leave" and whose checks said FAIL sealed at exit 0, and a hand-typed `receipts.json` read COMPLETE. Design-chain was also not on main at all, so no instance could receive it.

25 issues (dc-01..dc-25), each red first, reviewed, and closed with a receipt.

## What changed

- **Seal produces the verdicts.** It runs `design-standard-check.py`, `design-gap-check.py` and `design-impeccable-check.py` itself against a held snapshot and records their exit codes, instead of reading files the builder wrote.
- **Readers decide.** A LEAVE verdict, a re-roll over the cap, or a floor not met refuses the seal unless a founder disposition answers that reader by id.
- **Citations are real.** A repo file a technique or `proof.md` names must have been opened in the session that wrote the citation, taken from that session's transcript.
- **proof.md has a shape**: every artifact block names a kind from the closed list and a record the index holds, matched whole.
- **A founder finding is answered**: `FOUNDER-FINDING[tag]` needs a disposition with a reason before seal.
- **One door.** Design skills are engines: `design-engine-door.py` refuses one outside an open round and records each run in the round's `engines.jsonl`; a technique crediting an engine that never ran refuses the seal.
- **Lanes.** A non-site round skips the web-only stages and records them not applicable, never as passed. The lane comes from the instance config by where the round lives, never from the round's own manifest.
- **Decoration removed, measured first.** Over 67 rounds on disk the critique line count caught 0 and is deleted; the byte-copy brief check is re-bound to "the owners were read in full, this session".
- **The wiring ships last**: five hooks in both settings files, run with `python3 -I` so a `sitecustomize.py` cannot silence them, plus `SubagentStop`.
- **CI runs these tests for real.** It installs chromium and sets `DC_REQUIRE_REAL_PRODUCERS=1`. Measured: without it 30 design-chain tests skip and the file exits 0; with it the same 30 fail. `test_9` sat red behind that skip since dc-05.

## Evidence

- `prd_runner.py gates run` exits 0.
- The gate suite and all 29 design-chain test files pass.
- `design-chain-rollout-proof.sh --selftest` passes; against a bare directory all four checks fail.
- Every issue verified with the runner's 10 repeats and a defined-vs-ran count; one mutant per new decision point, each killed.

## Known gaps, filed rather than hidden

ASK-1871 (fingerprint counts), ASK-1874 (per-Bash latency from the registry walk), ASK-1875 (what the gate reads per surface: the browser tools, the Artifact `files` map, pages outside the scan roots, `/usr/bin/open`), ASK-1876 (`DESIGN_CHAIN_STATE` as a silent second override, Stop as a one-shot), ASK-1877 (engine registry gaps). Each carries its reproduction and a definition of ready. ASK-1875 is next.

The fleet sync is the founder's; the rollout proof against consulting runs after it and lands as dc-25's closeout line.

https://claude.ai/code/session_01L4EpVLDM5rrGWW7qJL3Ng8


assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-374-20260919-161529.md` (6154 bytes).

```
FINDINGS:
major|The SubagentStop hook wired in both settings files is inert: design-chain-gate.py has no branch for that event and returns 0 on the same unsealed page Stop refuses at 2|q-system/.q-system/scripts/design-chain-gate.py:3335
minor|.claude/settings.json runs five design-chain hooks with bare python3, including SendUserFile/Artifact and both door hooks, so a sitecustomize.py on PYTHONPATH silences them while the isolation test only reads settings-template.json|.claude/settings.json:183
nit|The Stop gate is registered twice in .claude/settings.json (one -I, one bare) and test_dc_wiring.wired() collapses them by (event, matcher), so one copy is never exercised|.claude/settings.json:431
END FINDINGS
```

--- reviewer output ---

## Review: PR #374 (design-chain, ASK-1796)

Read the diff, ran the gate against temp instances, ran 16 design-chain suites with `DC_REQUIRE_REAL_PRODUCERS=1`, ran the rollout proof selftest. Three findings survived reproduction.

---

### 1. major — the `SubagentStop` hook is wired in both settings files and the gate has no branch for it

`q-system/.q-system/scripts/design-chain-gate.py:3335` (the fall-through `return 0`; the only end-of-turn branch is `if ev == "Stop"` at :3328).

Commit e4d1b61c says "SubagentStop was not wired at all, so a subagent pages met no end-of-turn check" and adds the hook to `.claude/settings.json` and `settings-template.json`. The script never reads that event name. `grep -rn SubagentStop --include=*.py .` returns one hit, in `test_dc_wiring.py:67`, which asserts the string is present in the settings file, not that anything refuses.

Reproducer (same temp instance, same ledger, two events):

```
1. ledger the unsealed page (PostToolUse Write) -> exit 0 | ledger: ['s1.json']
2. Stop          -> exit 2 | DESIGN CHAIN GATE (blocked): a page was made this session and its design chain is not complete.
3. SubagentStop  -> exit 0 | ''
```

At 3am a subagent writes a page, its turn ends, and the surface this PR advertises as covered returns 0 silently. The test file's own docstring says "a grep for the command string proves a string, not a hook that runs". That is what happened here, to the event name instead of the command.

Fix is one line: `if ev in ("Stop", "SubagentStop")`, plus a case in `test_dc_wiring.py` that fires the payload.

---

### 2. minor — `.claude/settings.json` keeps five bare `python3` hook commands, including the whole publish surface and both door hooks

`.claude/settings.json:183` (gate, `Bash|SendUserFile|Artifact|...`), `:196` and `:251` (door).

The PR body says "five hooks in both settings files, run with `python3 -I` so a `sitecustomize.py` cannot silence them". Measured, `settings-template.json` is fully `-I`; `.claude/settings.json` has 3 of 8 on `-I` and the door has no `-I` form at all. `test_the_template_the_fleet_gets_runs_the_hooks_isolated` only reads `TEMPLATE`, so nothing sees it.

```
live  PreToolUse gate (bare python3), clean PYTHONPATH  -> exit 2 (2 = refuses)
live  PreToolUse gate (bare python3), sitecustomize     -> exit 0 (2 = refuses)
tmpl  PreToolUse gate (python3 -I),   sitecustomize     -> exit 2 (2 = refuses)
live  PreToolUse door (bare python3), clean PYTHONPATH  -> exit 2 (2 = refuses)
live  PreToolUse door (bare python3), sitecustomize     -> exit 0 (2 = refuses)
```

Dropped a level on purpose: this needs a `sitecustomize.py` on the session's `PYTHONPATH`, and no producer in this repo puts one there. The dsse `real_path_observer` sets it on a subprocess env, not the session. The fleet, which is what matters, gets the isolated copy.

---

### 3. nit — the Stop gate is registered twice and the wiring test can only see one of them

`.claude/settings.json:431` and the block immediately after it (one `-I`, one bare, same script, same event, no matcher). e4d1b61c added rather than replaced.

`wired()` in `test_dc_wiring.py:33` keys by `(event, matcher)`, so the two Stop rows collapse to one entry and `test_every_wired_command_runs_and_passes_a_benign_payload` plus `test_the_wired_gate_blocks_showing_an_unsealed_page` exercise whichever survives the overwrite, never both. Cost today is a doubled refusal message on every Stop, not a wrong verdict.

---

### What is sound (attacks that did not land)

- **A browser that pip-installs but never launches.** `design-standard-check.py:253` catches every exception from `measure()` and returns 2, writing no entry, so the seal records a non-zero rc and refuses. No degrade-to-pass.
- **Hand-typed fixtures.** `dc_fixtures.load()` refuses a fixture with no `_provenance`, with a null-valued one (`str(None)` is non-empty, and they caught that), or whose content sha drifted after capture. I looked for the invented-fixture shape and did not find it.
- **`--selftest` on the rollout proof.** Passes on this tree; the hook count is parsed out of the `hooks` arrays rather than grepped, and the marketplace clone is matched by `cmp`, not by existence. Both are the failure modes a review had already found.
- **Test suites.** 16 non-browser design-chain suites run with `DC_REQUIRE_REAL_PRODUCERS=1`, all `rc=0`: wiring, one_door_menu, engine_door, proof_schema, check_registry, citations, decoration, founder_findings, lanes, receipts, scan_scope, exemplars, command_doc, negative_controls, engine_records, reader_persona.
- **Seal ordering.** The live-vs-snapshot compare runs on both sides of the outside-inputs census, and `_withdraw_passes` strips `pass: true` after a refused seal. I tried to construct a swap window and could not.
- **Door name folding.** `canonical()` folds case, `_` vs `-`, slashes and the `plugin:skill` prefix, and an unreadable registry still refuses the coded CORE list. I could not get `Frontend_Design:` past it.

---

### VERDICT: REQUEST CHANGES

Not BLOCK. Nothing here is permanent or unrecoverable, and the PR is a real improvement over a gate that read the builder's own files.

**Fix first:** finding 1. `SubagentStop` is a claimed enforcement surface with no code path that refuses, and it is exactly the class of defect the whole PR exists to remove. One line in `hook()` plus a payload-firing test case closes it.

FINDINGS:
major|The SubagentStop hook wired in both settings files is inert: design-chain-gate.py has no branch for that event and returns 0 on the same unsealed page Stop refuses at 2|q-system/.q-system/scripts/design-chain-gate.py:3335
minor|.claude/settings.json runs five design-chain hooks with bare python3, including SendUserFile/Artifact and both door hooks, so a sitecustomize.py on PYTHONPATH silences them while the isolation test only reads settings-template.json|.claude/settings.json:183
nit|The Stop gate is registered twice in .claude/settings.json (one -I, one bare) and test_dc_wiring.wired() collapses them by (event, matcher), so one copy is never exercised|.claude/settings.json:431
END FINDINGS

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-374-20260919-163739.md` (7575 bytes).

```
FINDINGS:
major|An instance with no design-chain.json can never open a round, so design-engine-door.py refuses all 27 registered engines permanently and DESIGN_CHAIN_ALLOW=1 does not apply to it|q-system/.q-system/scripts/design-engine-door.py:91
major|A public page in an instance with no config blocks the Stop hook with a remediation that cannot be followed and never names the .not-a-round escape that clears it|q-system/.q-system/scripts/design-chain-gate.py:2523
major|The rollout proof writes the design-chain.json the fleet sync never ships, so it reports the rollout complete on an instance where the gate and door are unusable|q-system/.q-system/scripts/design-chain-rollout-proof.sh:64
minor|Every gate and door hook is wired twice in .claude/settings.json, doubling per-Bash latency from 0.185s to 0.37s, and the wiring test's any() assertion cannot see the duplicate|.claude/settings.json:205
END FINDINGS
```

--- reviewer output ---

## Findings

**1. major — An instance that receives the wiring but no `design-chain.json` can never open a round, so the door refuses all 27 registered design/video/deck engines permanently, and the founder's documented override does not apply to the door.**

`q-system/.q-system/scripts/design-engine-door.py:91`

`active_round()` needs `led["round"]`, which `design-chain-gate.py` sets only via `_is_round()`, which requires a `design-chain.json` found by walking up. No `design-chain.json` is tracked in this repo (`git ls-files | grep design-chain.json` is empty), and `/design-chain` does not create one, it stops with `UNKNOWN or NO`. `settings-template.json` ships the door to 25 registered instances.

Reproducer and real output:

```
A. no design-chain.json; agent writes brief.md to open a round
   round = (no ledger file)  door(hyperframes) rc = 2
B. with the founder's documented override DESIGN_CHAIN_ALLOW=1
   door(hyperframes) rc = 2
C. add design-chain.json with rounds_dir=site/design
   round = .../proj/site/design/r1  door(hyperframes) rc = 0
```

Refused with no round, from the same harness: `hyperframes`, `hyperframes-animation`, `deck-ai`, `frontend-design`, `kipi-design:brand`. `design-engine-door.py` contains no `DESIGN_CHAIN_ALLOW` check, so the override printed in every gate refusal does not open it.

At 3am a video or deck job gets exit 2, is told to run `/design-chain`, and `/design-chain` reports `Known: none`. Cleanup is a founder edit per instance.

**2. major — A public page written in an instance with no config blocks the end of turn with a remediation the agent cannot follow, and never names the one escape that works.**

`q-system/.q-system/scripts/design-chain-gate.py:2523` and `:2868`

`chain_problems()` lists `no design-chain.json found ... (the instance has no owner anchors configured)` as a problem, then requires `brief.md` with verbatim owner anchors that cannot exist without that config. `block()` prints only the chain steps plus `DESIGN_CHAIN_ALLOW=1 in the founder's shell only`.

Reproducer and real output:

```
--- ledger the page: rc=0
--- Stop before escape: rc=2
STDERR: DESIGN CHAIN GATE (blocked): a page was made this session and its design chain is not complete.
  .../proj/landing.html: no design-chain.json found above .../proj/landing.html (the instance has no owner anchors configured)
  .../proj/landing.html: missing .../proj/brief.md
--- Stop after writing .not-a-round: rc=0
```

The `.not-a-round` marker clears it in one write. The block text does not mention it.

**3. major — The rollout proof fabricates the config the sync never ships, so its "the instance's gate refuses an unsealed page" check cannot fail on the actual rollout gap.**

`q-system/.q-system/scripts/design-chain-rollout-proof.sh:64`

Line 64 writes `{"project": "proof", "owners": []}` into its own temp dir before running the instance's gate. Nothing in the four checks asks whether the instance itself has a `design-chain.json`. So `ROLLOUT PROOF: every check passed` is compatible with an instance where finding 1 and finding 2 are both live.

Reproducer and real output:

```
design scripts in the instance                             ok
gate and door hooks wired in the instance                  ok
the instance's gate refuses an unsealed page               ok
the command in the marketplace clone, same bytes           ok
ROLLOUT PROOF: every check passed for /var/folders/.../inst
exit=0
```

**4. minor — Every gate and door hook is wired twice in `.claude/settings.json`, once isolated and once plain, doubling per-Bash cost and the block text; the wiring test uses `any(...)` so it cannot see it.**

`.claude/settings.json:205` (also `:278`, `:479`), test at `q-system/.q-system/scripts/test/test_dc_wiring.py:91`

Enumerated from the parsed file:

```
PreToolUse     grp 0 m=Bash|SendUserFile|Artifact|...   -I
PreToolUse     grp 3 m=Bash|SendUserFile|Artifact|...   PLAIN
PostToolUse    grp 0 m=Edit|Write|MultiEdit|Bash        -I
PostToolUse    grp 2 m=Edit|Write|MultiEdit|Bash        PLAIN
Stop           grp 0 -I
Stop           grp 1 PLAIN
```

Measured cost of one copy: `PreToolUse Bash min=0.058s`, `PostToolUse Bash min=0.127s`, so ~0.185s per Bash call becomes ~0.37s. `settings-template.json` has only the isolated copies, so the skeleton pays double and the fleet does not. `test_every_live_surface_has_an_isolated_command` asserts `any("python3 -I " in c for c in cmds)`, which passes on the duplicated state by construction. 31 tests in that file pass right now.

## What is sound

- **Faking a producer verdict.** The stub producers write `_stub` and omit `measurements`, and `seal` runs `design-standard-check.py` / `design-gap-check.py` / `design-impeccable-check.py` itself against a held snapshot and records exit codes. A hand-typed `standard.json` does not read as a run. This is the defect the PR set out to fix and it holds.
- **Turning a real round off with the marker.** `tool_directory_problems()` refuses `.not-a-round` in any directory holding `brief.md`, and refuses an empty reason. The escape in finding 2 cannot be aimed at real design work.
- **Door name evasion.** `canonical()` folds case, colons, slashes and underscores. `kipi-design:brand` resolves to `brand` and is refused. Unlisted skills pass: `document-skills:pptx`, `codex:rescue`, `embedded-captions` all returned 0, so the door's scope is bounded by the registry rather than by guesswork.
- **The sitecustomize claim.** I tried to argue the duplicate plain-`python3` copy re-opens the hole `f0d68860` closed. It does not: hooks are independent and any exit 2 still blocks, so the isolated copy holds the line. Dropped that as a security finding and kept only the cost.
- **`DESIGN_CHAIN_ALLOW=1` on the gate** behaves exactly as the docstring says.
- **Rollout selftest and the wiring/door/menu suites** exit 0 and 31 passed, unmodified.
- One attempt I could not land: I looked for a path where this destroys or permanently publishes something. There is none. The gate is read-only outside its own state dir, `correct` only appends to `corrections.jsonl`, and everything above is reversible by a config file or a settings edit. So no blocker.

## VERDICT: REQUEST CHANGES

Three majors, nothing permanent, nothing that needs BLOCK.

**Fix first: finding 1.** Either ship a default `design-chain.json` in the sync payload, or make the door fail open when no config exists in the instance. Right now `kipi update` turns off video, decks, artifacts and `frontend-design` across the fleet and the printed way out is a command that refuses to start.

FINDINGS:
major|An instance with no design-chain.json can never open a round, so design-engine-door.py refuses all 27 registered engines permanently and DESIGN_CHAIN_ALLOW=1 does not apply to it|q-system/.q-system/scripts/design-engine-door.py:91
major|A public page in an instance with no config blocks the Stop hook with a remediation that cannot be followed and never names the .not-a-round escape that clears it|q-system/.q-system/scripts/design-chain-gate.py:2523
major|The rollout proof writes the design-chain.json the fleet sync never ships, so it reports the rollout complete on an instance where the gate and door are unusable|q-system/.q-system/scripts/design-chain-rollout-proof.sh:64
minor|Every gate and door hook is wired twice in .claude/settings.json, doubling per-Bash latency from 0.185s to 0.37s, and the wiring test's any() assertion cannot see the duplicate|.claude/settings.json:205
END FINDINGS

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-374-20260919-173621.md` (8096 bytes).

```
FINDINGS:
major|INTERNAL_MARKERS matches bare substrings, so /testimonials/, /schedule-a-call/ and /reports/ classify as internal and skip the entire design chain at every surface|q-system/.q-system/scripts/design-chain-gate.py:75
major|BASH_SHOW_RE matches only the literal `vercel deploy`, so `vercel --prod`, `npx vercel deploy`, netlify, git push, aws s3 sync and rsync deploy an unsealed page before Stop can refuse|q-system/.q-system/scripts/design-chain-gate.py:128
nit|Every design-chain hook is registered twice in .claude/settings.json (bare python3 plus python3 -I), doubling gate invocations for no enforcement the -I copy does not already carry|.claude/settings.json:205
END FINDINGS
```

--- reviewer output ---

# Review: PR #374 (design-chain rebuild, ASK-1796)

Two majors survived reproduction. Both are coverage holes in the new gate, not defects in its seal logic. The seal itself held up to everything I threw at it.

---

## MAJOR 1: `/test` swallows `/testimonials/`, so public pages skip the whole chain

`q-system/.q-system/scripts/design-chain-gate.py:75`

`INTERNAL_MARKERS` mixes directory-anchored entries (`/node_modules/`, `/tests/`) with three bare fragments: `/test`, `/schedule`, `/report`. `is_page()` at line 900 does a plain substring test against the resolved path plus a trailing slash. Any public path containing one of those fragments is classified internal, never enters the ledger, and passes every surface at exit 0.

Reproducer, run against the live wired command (`python3 -I design-chain-gate.py`), writing each page then firing PostToolUse Write, PreToolUse SendUserFile, and Stop:

```
SendUserFile=2  Stop=2  GATED       site/design/r1/Home-laptop.html
SendUserFile=0  Stop=0  UNGATED -- ships with no design chain  site/testimonials/index.html
SendUserFile=0  Stop=0  UNGATED -- ships with no design chain  site/schedule-a-call/index.html
SendUserFile=0  Stop=0  UNGATED -- ships with no design chain  site/reports/state-of-ai.html
```

Three ordinary marketing paths. One of them is a booking page, which is the surface this fleet's sites actually convert on.

Where this came from is visible: `plugins/kipi-design/hooks/dogfood_gate.py:343` carries the same loose list with an explicit written tradeoff, "a missed scan on a public page costs one advisory pass." That tradeoff is correct for an advisory tripwire. It was carried into a blocking gate where a miss costs the entire chain, and the new list has no comment re-deriving the cost. No test names `INTERNAL_MARKERS`.

Fix shape: anchor the three bare fragments the way their siblings are anchored, or gate on a directory-segment match instead of a substring.

---

## MAJOR 2: the Bash publish surface knows exactly one deploy spelling

`q-system/.q-system/scripts/design-chain-gate.py:128`

The comment above `BASH_SHOW_RE` says "Bash commands that put a page in front of a HUMAN," and `vercel deploy` is in the pattern, so publish-to-web is in scope for this branch. The pattern matches only `vercel deploy` at a command position.

Same harness, with the unsealed page registered in the ledger first:

```
PostToolUse Write (registers the unsealed page) rc = 0
Stop rc = 2

rc=2  BLOCKED    vercel deploy --prod
rc=0  ALLOWED    vercel --prod
rc=0  ALLOWED    npx vercel deploy --prod
rc=0  ALLOWED    netlify deploy --prod --dir=site
rc=0  ALLOWED    git push origin main
rc=0  ALLOWED    aws s3 sync ./site s3://www.example.com/
rc=0  ALLOWED    rsync -av site/ user@host:/var/www/
rc=2  BLOCKED    open Home-laptop.html
```

`vercel --prod` is the common form. `npx vercel` loses the command position entirely.

Stop is a backstop and it does fire (rc=2 above), but it fires after the deploy. At 3am an unattended agent puts an unvetted page on the public web, then gets refused at end of turn. The founder cleans it up by redeploying, so this is recoverable, not a blocker.

---

## NIT: every design-chain hook is registered twice in `.claude/settings.json`

`.claude/settings.json:205`

Twelve registrations across six surfaces: each one has a `python3 -I` copy and a bare `python3` copy.

```
SubagentStop   matcher=<none>                          gate -I
PreToolUse     matcher=Bash|SendUserFile|Artifact|...  gate -I
PreToolUse     matcher=Skill                           door -I
PreToolUse     matcher=mcp__plugin_chrome-devtools.*   gate -I
PreToolUse     matcher=Bash|SendUserFile|Artifact|...  gate PLAIN
PreToolUse     matcher=Skill                           door PLAIN
PostToolUse    matcher=Edit|Write|MultiEdit|Bash       gate -I
PostToolUse    matcher=Skill                           door -I
PostToolUse    matcher=Edit|Write|MultiEdit|Bash       gate PLAIN
PostToolUse    matcher=Skill                           door PLAIN
Stop           matcher=<none>                          gate -I
Stop           matcher=<none>                          gate PLAIN
```

`test_dc_wiring.py:84` sanctions this deliberately ("a proposal may only ADD to the live file, so the bare copies stay"), and the security property does hold: the `-I` copy still refuses when a `sitecustomize.py` silences the bare one. `settings-template.json`, which is what the fleet receives, is clean. So the bare copies buy no enforcement and cost a second process per tool call. Measured in this repo: 0.18s per PostToolUse Bash gate run, so the doubling is small here. I did not measure it in an instance that has a `design-chain.json`, where `scan_roots` actually walks.

---

## What is sound

- **The seal does produce its own verdicts.** `_seal_snapshot` at line 2685 takes exit codes from `run_producer`, and `run_producer` resolves producers as `HERE / name` with no env or flag override. Stand-ins require physically copying the gate next to them.
- **The A to X to A swap window is closed.** `round_files(rd)` is read on both sides of the outside-input census (lines 2731 and 2733), so a file changed during the census refuses.
- **`python3 -I` does not break the gate.** `-I` implies `-P`, which drops the script directory from `sys.path`. I ran the gate under both `python3 -I` and plain `python3`; both exit 0 on a benign payload, because sibling loading goes through `importlib.spec_from_file_location` on absolute paths, not imports.
- **Signal handling is correct.** `_die_cleanly` takes SIGTERM and SIGHUP only on the main thread, only over `SIG_DFL`, and restores the prior handler in a `finally`.
- **Opt-in is real.** `governed()` at line 2784 keeps an instance with no `design-chain.json` silent; my own probe needed the config present before anything blocked.
- **Fixture provenance is enforced, not asserted.** `dc_fixtures.load` refuses a fixture with no `_provenance` block and re-checks the content sha against the one recorded at capture, so a post-capture edit fails to load.
- **The wiring test runs the commands.** `test_dc_wiring.py` executes each wired hook with a real payload on stdin instead of grepping for its string. 11 tests, green in 1.6s.
- **`test_design_chain_gate.py`: 87 tests, green in 24.7s.**
- **The `.not-a-round` and `correct` escape hatches are narrowed, not open.** A directory holding `brief.md` cannot declare itself not a round; `correct` requires a git-committed predecessor and an identical `page_shape`. Both are stated founder decisions with real constraints, so I am not filing them.

**Dropped, could not reproduce.** `save_ledger` uses a non-atomic `write_text`, and `load_ledger` silently returns an empty ledger on a parse error, which would erase every open page. The duplicate registration does put two writers on that file concurrently. I could not construct a divergent-content race out of anything this PR emits, because both copies get the same payload and write the same bytes. Dropping it rather than dressing it up.

---

## VERDICT: REQUEST CHANGES

Two majors, neither permanent, both cleanable.

**Fix first: `INTERNAL_MARKERS` line 75.** It is the one that fails silently. The deploy-verb gap at least ends the turn with a refusal the operator sees; a testimonials page just sails through and nothing ever says so.

FINDINGS:
major|INTERNAL_MARKERS matches bare substrings, so /testimonials/, /schedule-a-call/ and /reports/ classify as internal and skip the entire design chain at every surface|q-system/.q-system/scripts/design-chain-gate.py:75
major|BASH_SHOW_RE matches only the literal `vercel deploy`, so `vercel --prod`, `npx vercel deploy`, netlify, git push, aws s3 sync and rsync deploy an unsealed page before Stop can refuse|q-system/.q-system/scripts/design-chain-gate.py:128
nit|Every design-chain hook is registered twice in .claude/settings.json (bare python3 plus python3 -I), doubling gate invocations for no enforcement the -I copy does not already carry|.claude/settings.json:205
END FINDINGS

assafkip:
## Review dispositions through round 3

**Round 1 (b3bee2bc), all fixed in e4d1b61c / f0d68860**
- SubagentStop wired with no branch in the gate: the branch reads both events, and the test fires the payload instead of asserting the string is present.
- 5 of 8 live hooks ran bare `python3`: every live surface now carries an isolated command, with a test that fails without one.
- `wired()` collapsed the two Stop entries: it keys by command.

**Round 2, fixed in 32175cf2 / f1d7bcd6**
- The door refused all 27 engines in an instance with no `design-chain.json`, and any page written there was blocked at Stop with a remediation nobody could follow. The chain is opt-in now: `governed()` asks whether a config sits at or above the path, and the gate and door are silent when it does not. This was the fleet blocker; 25 registered instances have no config.
- The rollout proof reported "every check passed" for an instance that never opted in. It states which of the two states the instance is in, first line and last.
- CI named one flag for three dependencies. `DC_REQUIRE_REAL_PRODUCERS` governs the browser only; the external detector and `node` skip under their own names; `pillow` joins the CI install because a missing `PIL` crashed two files instead of skipping them.

**Round 3, fixed in 2c4aba7a**
- `INTERNAL_MARKERS` was a substring test, so `/testimonials/`, `/schedule-a-call/` and `/reports/` read as internal and skipped the chain. Internal is a path segment now, plus two known filename prefixes.
- `BASH_SHOW_RE` knew only `vercel deploy`. It knows the deploy verbs that publish a directory; `git push` stays out deliberately, with the reason in the code.

**Parked, with the reproduction and a definition of ready**
- **ASK-1875** (next in the queue): what the gate reads per surface. The browser tools are judged by the session ledger rather than the path they name; Artifact's `files` object map escapes the carried-file reader; a page written outside the scan roots never enters the ledger; `/usr/bin/open` and `osascript` walk past the Bash verb check.
- **ASK-1876**: `DESIGN_CHAIN_STATE` relocates the ledger and silently disarms every ledger-backed branch, while `block()` advertises only `DESIGN_CHAIN_ALLOW`; the Stop refusal is spent after one round (`stop_hook_active`).
- **ASK-1877**: the engine registry misses `artifact-design`, `dataviz`, `figma`; `Agent` and `SlashCommand` are not routed to the door, and `NotebookEdit` writes no ledger row.
- **ASK-1874**: the per-Bash cost of the registry walk (130-180ms), plus the duplicate hook entries in this repo's `.claude/settings.json` (0.185s to 0.37s). The duplicates exist because `apply_claude_changes.py` is additive only and cannot replace the bare entries; `settings-template.json`, which is what the fleet receives, has one entry per surface.

Findings in those four classes are tracked, not ignored. This PR's scope is the rebuild and the wiring.

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-374-20260919-181716.md` (6752 bytes).

```
FINDINGS:
major|Duplicate hook wiring runs two gate processes per event; the unlocked read-modify-write ledger drops pages and Stop stops refusing them|.claude/settings.json:278
major|design-engine-door.py honors no DESIGN_CHAIN_ALLOW override, so an opted-in instance hard-refuses all 35 engines including hyperframes and deck-ai with no escape|q-system/.q-system/scripts/design-engine-door.py:81
minor|block() truncates the refusal at 40 lines, so with many open pages some are never named to the operator|q-system/.q-system/scripts/design-chain-gate.py:2887
END FINDINGS
```

--- reviewer output ---

## Findings

---

**MAJOR — the gate silently forgets pages, so Stop stops refusing them**

`.claude/settings.json:278` (and `:287`, `:205`, `:214`, `:479`) wires a second, bare `python3` copy of every hook this PR adds, alongside the `python3 -I` copy. Claude Code runs hooks matching one event in parallel, so every `Write` fires two concurrent `design-chain-gate.py` processes against one ledger file. `load_ledger` at `q-system/.q-system/scripts/design-chain-gate.py:2782` catches `ValueError` from a torn read and returns a **fresh empty ledger**, which `save_ledger` then writes back over the real one. Pages the gate was watching vanish.

Reproducer, run from the worktree, control versus branch shape:

```
python3 - <<'PY'   # full script pasted in the run below
trial(post_gate[:1], "CONTROL  one wired copy (settings-template.json shape)")
trial(post_gate,     "BRANCH   both wired copies (.claude/settings.json shape)")
PY
```

Real output:

```
CONTROL  one wired copy (settings-template.json shape) -> ledgered 8 of 8
BRANCH   both wired copies (.claude/settings.json shape) -> ledgered 4 of 8
```

A 12-way run of the same shape lost a page and then `Stop` exited 2 while never naming it:

```
pages written : 12
pages ledgered: 11
LOST from the ledger: ['Page4-laptop.html']
Stop exit: 2 | unsealed pages Stop names: 10
unsealed pages Stop NEVER mentions: ['Page4-laptop.html', 'Page10-laptop.html']
```

At 3am in an opted-in instance this is the gate failing open on a page it already saw. `test_dc_wiring.py:83` knows the bare copies are there and asserts only that each surface *also* has an isolated one, so the suite is green over it. Recoverable (the page is still on disk, and the publish branch re-checks carried files independently), which is why this is major and not a blocker.

---

**MAJOR — `DESIGN_CHAIN_ALLOW=1` opens the gate and not the door**

`q-system/.q-system/scripts/design-engine-door.py:81` (`decide`) has no override branch. The gate's own refusal text at `design-chain-gate.py:2890` tells the reader "Override is DESIGN_CHAIN_ALLOW=1 in the founder's shell only." In an instance that has opted in, all 27 registry engines plus the 8 `CORE` names are hard-refused with no escape, and two of those names are not web design at all: `hyperframes` is the router for every video task and `deck-ai` is deck work.

Reproducer and real output:

```
engines in the registry: 27
skill=hyperframes            DESIGN_CHAIN_ALLOW=unset -> exit 2 | 'hyperframes' is a design engine, and this session has no open design-chain round. Start with /design-chain; i
skill=hyperframes            DESIGN_CHAIN_ALLOW=1     -> exit 2 | 'hyperframes' is a design engine, and this session has no open design-chain round. Start with /design-chain; i
skill=deck-ai                DESIGN_CHAIN_ALLOW=1     -> exit 2 | 'deck-ai' is a design engine, and this session has no open design-chain round. Start with /design-chain; it ca
skill=design-room            DESIGN_CHAIN_ALLOW=1     -> exit 2 | 'design-room' is a design engine, and this session has no open design-chain round. Start with /design-chain; i
```

The only cleanups available to a human are deleting `design-chain.json` or hand-building a round directory. The remediation the message prints is also wrong for a video request.

---

**MINOR — a refusal with many open pages does not name them all**

`design-chain-gate.py:2887` caps the printed lines at 40, built from up to 4 problems per page. With 11 ledgered pages the Stop refusal named 10 of them (output above). The operator or model fixes what it can see, hits the gate again, sees the next slice. Bounded and cosmetic next to the two above.

---

## What is sound

- **The CI change is real, not a claim.** With chromium installed and `DC_REQUIRE_REAL_PRODUCERS=1`, the four seal-heavy files ran for real: `95 passed, 13 subtests passed in 148.62s`. `capability-gate.py:681` builds the child env as `dict(os.environ, QROOT=...)`, so the variable actually reaches the tests rather than dying at the step boundary.
- **Opt-in holds.** There is no `design-chain.json` anywhere in this repo, `governed()` gates the ledger, and the door returns 0 when no config sits at or above cwd. `design-chain-rollout-proof.sh --selftest` against a bare instance printed `INERT here: no design-chain.json, so the gate and the door stay silent` and passed all four checks.
- **`test_dc_wiring.py` is not decoration.** It runs each wired command as a real subprocess with the payload on stdin; 11 passed in 1.39s. The sitecustomize attack it models genuinely fails against the `-I` copy.
- **Duplicate `engines.jsonl` rows are harmless.** I confirmed one `design-room` run writes 2 rows under the duplicated wiring, then read `engine_problems` at `design-chain-gate.py:3053`: it folds rows into a set, so the duplication cannot forge or break a credit.
- **No producer-path override exists.** I looked for the env var or flag that would point `seal` at a friendlier producer directory, which is the obvious way to re-trust an account. There is none; `HERE` is the only source.
- **The rollout proof parses rather than greps** settings.json, and compares the marketplace command by bytes, so a stale fork does not pass.

Dropped after trying: I could not make the `INTERNAL_DIRS` segment list (`reports`, `dashboard`, `schedule`) produce a false negative on any page a producer in this repo emits, and the `PostToolUse` Bash scan's missing `governed()` call turned out to be covered upstream by `scan_roots` returning only config-governed roots.

---

## VERDICT: REQUEST CHANGES

Two majors, nothing permanent, so this is changes-requested rather than a block.

**Fix first:** delete the five bare `python3` hook copies from `.claude/settings.json` so one process owns the ledger per event, and give `save_ledger` an atomic write (temp file plus `os.replace`) plus a `load_ledger` that distinguishes "no ledger yet" from "unreadable ledger". The second half matters on its own: a torn read currently reports the same value as a clean start, which is the shape `a-predicate-whose-false-branch-is-unreachable` names.

FINDINGS:
major|Duplicate hook wiring runs two gate processes per event; the unlocked read-modify-write ledger drops pages and Stop stops refusing them|.claude/settings.json:278
major|design-engine-door.py honors no DESIGN_CHAIN_ALLOW override, so an opted-in instance hard-refuses all 35 engines including hyperframes and deck-ai with no escape|q-system/.q-system/scripts/design-engine-door.py:81
minor|block() truncates the refusal at 40 lines, so with many open pages some are never named to the operator|q-system/.q-system/scripts/design-chain-gate.py:2887
END FINDINGS

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-374-20260919-185713.md` (6102 bytes).

```
FINDINGS:
minor|is_page() still exempts /reports/, /schedules/ and /dashboards/ from the entire design chain while the commit message and docstring claim that exemption was a fixed major, and the sibling dogfood gate calls the same path public|q-system/.q-system/scripts/design-chain-gate.py:81
minor|BASH_SHOW_RE matches any `vercel` subcommand, so read-only `vercel ls`/`logs`/`whoami` are refused while a round is open, pushing the operator toward the DESIGN_CHAIN_ALLOW=1 full bypass|q-system/.q-system/scripts/design-chain-gate.py:138
minor|The deploy-verb list misses npm run deploy, scp to a host, firebase deploy, wrangler pages deploy and surge, while the adjacent rsync-to-a-host case is caught|q-system/.q-system/scripts/design-chain-gate.py:136
END FINDINGS
```

--- reviewer output ---

Review complete. Suite run, repros executed.

## Findings

**MINOR — `/reports/`, `/schedules/`, `/dashboards/` still skip the whole chain, and the commit that "fixed" that says otherwise**
`q-system/.q-system/scripts/design-chain-gate.py:81`

Commit `2c4aba7a` and the docstring at line 74 both say: *"/reports/ matched '/report': three ordinary marketing pages skipped the entire chain (PR #374 review, major)."* The fix converted substring matching to segment matching, which repaired `/testimonials/` and `/schedule-a-call/`, and then kept `reports`, `schedules` and `dashboards` in the set as exact segments. `/reports/` is still exempt. `test_dc_internal_and_deploy.py:37` pins it that way, so the test locks in the shape the commit message says it removed.

Worse, the sibling gate wired in this same PR disagrees on the identical path.

Repro and real output:

```
is_page=False site/reports/q3-report.html
is_page=True  site/blog/post.html

governed=True  SendUserFile of UNSEALED site/reports/q3-report.html -> hook exit 0
governed=True  SendUserFile of UNSEALED site/blog/post.html      -> hook exit 2
```

```
dogfood_gate.is_public_facing_page("/repo/site/reports/q3.html") -> True
design-chain-gate.is_page("/repo/site/reports/q3.html")          -> False
```

So on one path the dogfood hook demands a design pass and the design-chain gate publishes it unsealed. Calibration note, stated plainly: I constructed `site/reports/q3-report.html`, and no producer in this repo emits that path today, so I dropped this a level rather than calling it major.

**MINOR — `vercel\b` refuses read-only vercel subcommands while a round is open**
`q-system/.q-system/scripts/design-chain-gate.py:138`

Real output:

```
BLOCK  vercel ls
BLOCK  vercel whoami
BLOCK  vercel logs
BLOCK  vercel deploy --prod
```

The refusal text is "a page was made this session and its design chain is not complete", and its remediation is to seal a round. That has nothing to do with `vercel logs`. The window where it fires is exactly mid design round, which is when this fleet's `/vercel:status` and `/vercel:env` skills run those commands. The documented escape is `DESIGN_CHAIN_ALLOW=1`, which turns the entire gate off, so the false alarm pushes toward the widest possible bypass.

**MINOR — the deploy-verb list misses the common ones**
`q-system/.q-system/scripts/design-chain-gate.py:136`

Same run, same regex:

```
pass   npm run deploy
pass   scp -r site/ user@host:/var/www/
pass   firebase deploy --only hosting
pass   wrangler pages deploy site
pass   surge ./site
```

`rsync` to a host blocks and `scp` to a host does not; those are the same act. `npm run deploy` is the ordinary shape. The Stop hook and the publish tools still catch the page, so this is a belt with a gap, not an open door.

## What is sound

- **Duplicate hook wiring.** `.claude/settings.json` carries every design hook twice, plain `python3` beside `python3 -I`, so each fires twice per event. I chased it as a correctness bug and it holds: the gate takes an flock around every ledger read-modify-write, pinned by `test_the_gate_will_not_touch_a_ledger_another_process_holds`, and `engines.jsonl` is read into a set so the door's unlocked double append is harmless. Measured cost is 0.06s per PreToolUse Bash and 0.14s per PostToolUse Bash, per copy.
- **The full suite.** 523 passed, 32 subtests passed, 0 skipped, 0 failed, 613 seconds, with chromium present.
- **`design-chain-rollout-proof.sh --selftest`** exits 0 and correctly reports the skeleton as INERT.
- **Blast radius on merge.** No `design-chain.json` exists anywhere in this repo, so `governed()` is false everywhere and both the gate and the door are silent in the skeleton and in every instance until one opts in. The opt-in check is real, not asserted.
- **Builder-forges-an-exemption.** The `tier: wireframe` and `status: withdrawn` escapes are deliberate, require a written reason, are read from the held snapshot rather than live disk, and `seal` writes no receipt at all for a withdrawn round.
- **`TEST_ROUND_ROOTS` as a production escape.** `_temp_roots()` reads the OS rather than `TMPDIR`, and a round counts as a test round only when its path is under a temp root both as named and as resolved.
- **Reader-verdict laundering.** Re-roll past the cap, weaken `floor`/`n`/`viewports`/`narrow` after a run, splice rows across runs, swap a stylesheet mid-run, rename the page to reset its history, delete the `readers` block to escape a LEAVE. Each has its own refusal with the adversarial case named in a comment.
- **`check_cli` argv edge cases.** No crash, fails closed at exit 2, and SKIPPED gets its own exit code 3 that the seal reads as "did not run".

## VERDICT: APPROVE WITH NITS

Only minor findings survived reproduction. Three of my stronger hypotheses died on contact with the code, which is the honest read on a PR this size.

Fix first: line 81. Either drop `reports`, `schedules` and `dashboards` from `INTERNAL_DIRS` so the code matches what `2c4aba7a` claims it did, or keep them and rewrite the commit message and the docstring so they stop naming `/reports/` as a marketing page that was rescued. Right now the prose and the test assert opposite things, and the dogfood gate is the tiebreaker that says the code is wrong.

FINDINGS:
minor|is_page() still exempts /reports/, /schedules/ and /dashboards/ from the entire design chain while the commit message and docstring claim that exemption was a fixed major, and the sibling dogfood gate calls the same path public|q-system/.q-system/scripts/design-chain-gate.py:81
minor|BASH_SHOW_RE matches any `vercel` subcommand, so read-only `vercel ls`/`logs`/`whoami` are refused while a round is open, pushing the operator toward the DESIGN_CHAIN_ALLOW=1 full bypass|q-system/.q-system/scripts/design-chain-gate.py:138
minor|The deploy-verb list misses npm run deploy, scp to a host, firebase deploy, wrangler pages deploy and surge, while the adjacent rsync-to-a-host case is caught|q-system/.q-system/scripts/design-chain-gate.py:136
END FINDINGS

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-374-20260919-193020.md` (9041 bytes).

```
FINDINGS:
major|issue_runner._defined_vs_ran adds skipped to ran, so a check whose tests all skip is status ok and verify writes a green receipt over zero executed tests|plugins/kipi-dsse/scripts/issue_runner.py:854
major|dogfood_gate INTERNAL_PATH_MARKERS matches bare substrings, so a round page named schedule.html or dashboard.html returns SKIPPED(3) and the seal refuses it forever with no reachable escape|plugins/kipi-design/hooks/dogfood_gate.py:365
minor|The skeleton leak sweep excludes impeccable-receipt.json rather than re-capturing it, shipping /Users/assafkipnis/projects/cole-gtm/... to a public repo and to every instance via kipi update|validate-separation.py:997
nit|check_budget's docstring says issue_runner verify calls judge(), but verify calls full_suite_checks, and judge() also carries the CLI-only rule 2|plugins/kipi-dsse/scripts/check_budget.py:12
END FINDINGS
```

--- reviewer output ---

Review done. Suite run, both repros executed end to end.

## Findings

### 1. major — `verify` contract 2 counts a fully skipped check as `ok`, writes a green receipt, and closeout accepts it

`plugins/kipi-dsse/scripts/issue_runner.py:854` (`_defined_vs_ran`, the `ran + skipped < defined` branch)

The whole point of contract 2 is in its own header comment: *"did the test pass, and could you then run it on production and it would pass again every time."* `skipped` is added to `ran` before the comparison, so a file whose every test calls `skipTest` satisfies defined-vs-ran. `_green_shape_problems` then accepts `status == "ok"` and the issue closes.

This is not a constructed shape. `q-system/.q-system/scripts/test/test_dc_reader_gate.py:53` calls `self.skipTest(msg)` when no browser is present, and this PR's own body says *"without it 30 design-chain tests skip and the file exits 0. `test_9` sat red behind that skip since dc-05."* The CI env var `DC_REQUIRE_REAL_PRODUCERS=1` fixes that in CI only. `verify`, which is the thing that computes the green, still says ok.

Reproducer, end to end through `_verify_v2` in a temp repo, one check that drives the target and one whose 3 tests all skip:

```
$ python3 - <<'EOF'   # full script in the transcript
...
checks: ["python3 -m pytest test_drives.py -q", "python3 -m pytest test_skipper.py -q"]
EOF
```

Real output:

```
{"verified": "repro-1", "at": "2026-09-20T02:36:06Z", "checks_run": 2,
 "real_path": {"...test_drives.py -q": ["tool.py"], "...test_skipper.py -q": []},
 "defined_vs_ran": [{"command": "...test_drives.py -q", "defined": 1, "ran": 1, "skipped": 0, "status": "ok"},
                    {"command": "...test_skipper.py -q", "defined": 3, "ran": 0, "skipped": 3, "status": "ok"}],
 "repeats": "10 of 10"}

_verify_v2 exit code: 0
receipts: {"verified": "2026-09-20T02:36:06Z"}
green shape problems at close: []
```

`defined: 3, ran: 0, skipped: 3, status: "ok"`, exit 0, receipt written. At 3am an issue verifies and closes on tests that never executed, which is the exact defect class this contract was built to remove. `real_path` does not save it: `driven` is a union across all checks, so one check touching the target covers a sibling that ran nothing.

### 2. major — an ordinary page name makes a design round permanently unsealable

`plugins/kipi-design/hooks/dogfood_gate.py:365` (`INTERNAL_PATH_MARKERS`), consumed at `q-system/.q-system/scripts/design-chain-gate.py:774`

`is_public_facing_page` matches those markers as bare substrings of the whole path. Five of them carry no slash: `schedule`, `dashboard`, `morning`, `debug`, `fixture`. The seal passes the live page path as `--as`, `check_cli` returns `SKIPPED (3)`, and `CHECKS["tripwire"]["skip"] == {3}` turns that into a refusal: *"A check that skipped is not a pass."* The `eyeball-gate-skip` marker also returns 3, so the documented per-file bypass does not help. The only ways out are renaming the page or `DESIGN_CHAIN_ALLOW=1`, which turns the entire gate off.

Note the sibling gate in this same PR was fixed in the opposite direction: commit `2c4aba7a` moved `design-chain-gate.py`'s `INTERNAL_DIRS` to path-segment matching precisely because `/reports/` matched `/report`. `dogfood_gate.py` kept substring matching, and the seal now depends on it.

Reproducer through the repo's own round harness (real gate, real producers, real browser), one round with two pages:

```
$ cd q-system/.q-system/scripts/test && python3 - <<'EOF'
from test_dc_check_registry import Registry, CLEAN
# round with Home-laptop.html and schedule.html, both clean, tripwire declared
EOF
```

Real output:

```
stage check:tripwire schedule.html: 0.02s (exit 3)
stage check:tripwire Home-laptop.html: 0.02s (exit 0)
seal REFUSED:
schedule.html: the tripwire check did not run on this page (exit 3): skipped: .../site/design/r1/s...
seal exit: 2
receipts.json written: False
```

Direct `check_cli` sweep over plausible names, same run:

```
Home-laptop.html          rc=0  pass
schedule.html             rc=3  SKIPPED -> seal REFUSES
dashboard.html            rc=3  SKIPPED -> seal REFUSES
good-morning.html         rc=3  SKIPPED -> seal REFUSES
debug-free-page.html      rc=3  SKIPPED -> seal REFUSES
our-fixtures.html         rc=3  SKIPPED -> seal REFUSES
changelog.html            rc=0  pass
```

Calibration, stated: I chose those filenames, no producer in this repo emits them. I kept it at major rather than dropping a level because the input is a human naming a page on a marketing site, which is this system's normal operating input, and because the outcome is a hard permanent refusal whose own message says the check "did not run" without naming any reachable escape. If you disagree on the class, the fix is the same size either way: segment matching, the way `2c4aba7a` already did it next door.

### 3. minor — the fleet leak sweep was widened rather than the fixture fixed

`validate-separation.py:997`

The new `"impeccable-receipt"` exclusion is the only thing keeping the skeleton sweep green over a tracked file that carries `/Users/assafkipnis/projects/cole-gtm/.agents/skills/impeccable/...`. That file is under `q-system/`, so `kipi update` rsyncs it to every registered instance, and the repo is public. The comment's reason is honest (the capture sha pins the content), but the fix that keeps the gate intact is re-capturing the fixture from a neutral path, not naming it in the allowlist.

Reproducer, the sweep's own predicate with and without the new entry:

```
WITH the new exclusion      -> impeccable-receipt.json absent from hits
WITHOUT it (main's list)    -> 'q-system/.q-system/scripts/test/fixtures/design-chain/impeccable-receipt.json'
```

(The 62 other hits in both runs are my local `__pycache__`, untracked, absent in a fresh clone.)

### 4. nit — `check_budget.py:12` names the wrong function

The docstring says *"`issue_runner.py verify` calls `judge()` on the spec snapshot"*. It calls `full_suite_checks` (`issue_runner.py:933`). Same rule-1 behaviour, wrong name, and it matters because `judge()` also returns rule 2, which the same docstring says is CLI only.

## What is sound

- Duplicate hook registrations in `.claude/settings.json`. Twelve entries against the template's six, plain `python3` beside `python3 -I`. I chased it as a correctness bug and it holds: `hook()` takes an flock around the whole ledger read-modify-write, and `engine_problems` reads `engines.jsonl` into a set, so the door's unlocked double append is harmless.
- `DESIGN_CHAIN_ALLOW` opening the door. It is checked below the PostToolUse branch on purpose, so an override still leaves the engine record the seal needs.
- An engine credit with no run. `_engine_credits` walks any key containing "engine" at any depth, list or scalar, and `_no_duplicate_keys` refuses a manifest that parses differently than it reads.
- Receipt ledger shape. `linear:ASK-1808` passes, `linear:ACMECORPPRICINGDEAL-1` does not, whitespace and `/` and `@` are refused before the slug test.
- The repeat budget seam. `KIPI_VERIFY_BUDGET_S` is `min`-capped, so raising it cannot skip the `n < 5` refusal.
- `_count_defined` against a mid-file `unittest.main()` and a nested `TestCase`. The AST walk recurses into class bodies, if/try/with, and skips the `__main__` guard.
- `_unnamed_test_runs`. `pytest .` or `pytest discover` is refused before anything runs, which closes the "not counted" road past defined-vs-ran.
- Suites. `test_dc_*` plus `test_design_engine_door`: 46 passed. `test_dc_check_registry` with real producers and a browser: 11 passed.

## VERDICT: REQUEST CHANGES

Not BLOCK. Neither major is permanent: finding 1 is a re-verify, finding 2 is a rename or a one-line matcher change. Both are recoverable after merge.

**Fix first: finding 1.** A green receipt written over zero executed tests is the same failure this PR's own RCA was written about, reproduced inside the mechanism built to prevent it. One line in `_defined_vs_ran`: compare `ran` against `defined` and put `skipped` in the refusal text, or refuse when `ran == 0` and `defined > 0`.

FINDINGS:
major|issue_runner._defined_vs_ran adds skipped to ran, so a check whose tests all skip is status ok and verify writes a green receipt over zero executed tests|plugins/kipi-dsse/scripts/issue_runner.py:854
major|dogfood_gate INTERNAL_PATH_MARKERS matches bare substrings, so a round page named schedule.html or dashboard.html returns SKIPPED(3) and the seal refuses it forever with no reachable escape|plugins/kipi-design/hooks/dogfood_gate.py:365
minor|The skeleton leak sweep excludes impeccable-receipt.json rather than re-capturing it, shipping /Users/assafkipnis/projects/cole-gtm/... to a public repo and to every instance via kipi update|validate-separation.py:997
nit|check_budget's docstring says issue_runner verify calls judge(), but verify calls full_suite_checks, and judge() also carries the CLI-only rule 2|plugins/kipi-dsse/scripts/check_budget.py:12
END FINDINGS

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-374-20260919-195335.md` (5555 bytes).

```
FINDINGS:
major|PostToolUse Bash enrols any touched page-extension file with no round scoping, so a git checkout or npm install in an opted-in instance makes the Stop hook refuse the end of the turn over ordinary application source|q-system/.q-system/scripts/design-chain-gate.py:2945
major|scan_roots expands instance-registry.json, so a session in a project with no design-chain.json enrols another project's round pages and its Stop hook refuses with a remediation that project cannot perform|q-system/.q-system/scripts/design-chain-gate.py:2896
END FINDINGS
```

--- reviewer output ---

Review done. Two reproduced majors, both in the same hook branch.

## Finding 1 (major) — a `git checkout` wedges the turn in any opted-in instance

The `PostToolUse` + `Bash` branch of `_hook` enrols **every** file with a page extension whose mtime moved during the command, with no round scoping. The `Write`/`Edit` branch calls `governed()`; this one calls nothing. `.tsx`, `.jsx`, `.vue`, `.svelte`, `.astro`, `.html` are all pages to `is_page()`, so ordinary application source counts.

`q-system/.q-system/scripts/design-chain-gate.py:2945` (`newer_pages`, called from the PostToolUse Bash branch)

Reproducer: `python3 q-system/output/_rev374_bash_enrol.py` (instance with `design-chain.json`, one `src/components/Card.tsx`, an `os.utime` standing in for what git, npm and rsync do to every file they write).

Real output:

```
PostToolUse Bash rc: 0
pages enrolled in the session ledger: ['.../instance/src/components/Card.tsx']

Stop rc: 2
DESIGN CHAIN GATE (blocked): a page was made this session and its design chain is not complete.
  .../src/components/Card.tsx: missing .../src/components/brief.md
  .../src/components/Card.tsx: missing .../src/components/directions.md
  .../src/components/Card.tsx: missing .../src/components/standard.json
  .../src/components/Card.tsx: missing .../src/components/critique.md
  Override is DESIGN_CHAIN_ALLOW=1 in the founder's shell only.
```

At 3am in a `claude -p` run this is a turn that cannot end. The remediation printed asks for a brief and three directions for a React component. The escapes are deleting the session ledger by hand, dropping a `.not-a-round` marker in every source folder, or `DESIGN_CHAIN_ALLOW=1`, which the agent cannot set for itself. `test_an_instance_with_no_config_is_never_blocked` covers the Write path only, so nothing in the suite sees this.

## Finding 2 (major) — a project that never opted in is wedged by a different project

`scan_roots()` expands `$CLAUDE_PROJECT_DIR/instance-registry.json` into its root list, so a session in the skeleton repo walks other repos on every Bash call. Measured against the real registry from this worktree:

```
session cwd : .../assafkip_kipi-system__pr-374
roots walked on every PostToolUse Bash:
   /Users/assafkipnis/projects/consulting   (inside this project: False)
pages this session would enrol if they were touched: 1005
```

`q-system/.q-system/scripts/design-chain-gate.py:2896`

Reproducer with a synthetic registry of the same shape: `python3 q-system/output/_rev374_xproject2.py`

```
session project : .../skeleton (has design-chain.json: False )
pages enrolled  : ['.../other-project/site/design/r1/Home-laptop.html']

Stop rc: 2
DESIGN CHAIN GATE (blocked): ...
  .../other-project/site/design/r1/Home-laptop.html: missing .../r1/brief.md
```

kipi-system has no `design-chain.json`, so by the PR's own opt-in promise it should be silent here. Two concurrent sessions is the documented working mode in this fleet (`concurrent-session-worktrees.md`), and `kipi update` restamps mtimes fleet-wide on a schedule, so the trigger is routine, not adversarial. The founder's session in repo A gets a refusal naming a path in repo B.

Both findings are one branch. The fix is the same shape as `governed()`: enrol from Bash only what sits in a rounds folder of the session's **own** project.

## What is sound

Attacks I tried that the code survived:

- **Silencing a hook with `sitecustomize.py`.** The `-I` copies hold. `python3 q-system/.q-system/scripts/test/test_dc_wiring.py` ran 12 tests OK in 2.9s, including the planted-sitecustomize case and the ledger-lock race.
- **Per-Bash latency.** I expected the registry walk to be the problem. Measured in this repo: PreToolUse Bash 0.055s median, PostToolUse Bash 0.121s median over the real 25-instance registry. That is fine; ASK-1874 overstates itself.
- **Producer redirection.** There is genuinely no env var, flag or path condition pointing `seal` at another producer directory. Tests get stand-ins by running a copy of the gate.
- **Receipt cache poisoning.** `_stat_key` carries `st_ctime_ns` and `st_ino`, not just size and mtime, so a same-length rewrite with `os.utime` restored does not hit the cache. Read, not run.
- **NOT_APPLICABLE as a free pass.** `_believed_na` re-derives the tripwire's scope from the page as it is now rather than trusting the receipt's word.

Dropped, could not reproduce before the tool ceiling: the duplicate `PostToolUse` Skill entries in `.claude/settings.json` appear to append two rows per call to `engines.jsonl`, but `engine_problems` reads into a set, so I could not make it change a verdict.

Housekeeping: my three reproducer scripts are in `q-system/output/`, which is gitignored, so `git status` stays clean. Delete them when convenient.

## VERDICT: REQUEST CHANGES

Not BLOCK. Both findings are recoverable by a human deleting a ledger file.

Fix first: scope the PostToolUse Bash enrolment to round directories inside the session's own project.

FINDINGS:
major|PostToolUse Bash enrols any touched page-extension file with no round scoping, so a git checkout or npm install in an opted-in instance makes the Stop hook refuse the end of the turn over ordinary application source|q-system/.q-system/scripts/design-chain-gate.py:2945
major|scan_roots expands instance-registry.json, so a session in a project with no design-chain.json enrols another project's round pages and its Stop hook refuses with a remediation that project cannot perform|q-system/.q-system/scripts/design-chain-gate.py:2896
END FINDINGS

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-374-20260919-200917.md` (8012 bytes).

```
FINDINGS:
minor|`.claude/settings.json` registers each design-chain hook twice (isolated and bare), doubling gate and door invocations on every Bash, Edit, Write, Skill and Stop|.claude/settings.json:171
minor|One Skill engine call appends two identical rows to the round's engines.jsonl because the PostToolUse door is wired twice|q-system/.q-system/scripts/design-engine-door.py:74
minor|dogfood_gate.is_public_facing_page and design-chain-gate.is_page disagree on `reports/` pages, so a real marketing page gets the slop tripwire but silently skips the whole chain|q-system/.q-system/scripts/design-chain-gate.py:77
END FINDINGS
```

--- reviewer output ---

## Findings

**minor** — `.claude/settings.json` registers every design-chain hook twice (once `python3 -I`, once bare), so the gate and the door each run two processes on every Bash, Edit, Write, Skill and Stop. `settings-template.json` registers them once.
`.claude/settings.json:171` (and 190, 206, 255, 271, 467, 475)

Reproducer:
```bash
python3 - <<'EOF'
import json
for f in [".claude/settings.json","settings-template.json"]:
    d=json.load(open(f)); print("==",f)
    for ev,arr in d.get("hooks",{}).items():
        for m in arr:
            for h in m.get("hooks",[]):
                c=h.get("command","")
                if "design-chain-gate" in c or "design-engine-door" in c:
                    print(f"  {ev} matcher={m.get('matcher','<none>')!r} isolated={'-I' in c.split()}")
EOF
```
Real output: 12 registrations in `.claude/settings.json`, 6 in `settings-template.json`. Same event, same script, both `-I` and bare, on PreToolUse Bash, PreToolUse Skill, PostToolUse Edit|Write|MultiEdit|Bash, PostToolUse Skill, Stop.

Measured cost of one invocation in this repo:
```
single gate invocation ms: [187, 107, 81]
per Bash call (2 registrations) ms: 161
exit: 0
```

The gate's own comment at `q-system/.q-system/scripts/design-chain-gate.py:2860` says the duplicates exist and that `apply_claude_changes` cannot remove them, and `ledger_lock()` already covers the ledger race they cause. What is left is the doubled spend on every tool call in the founder's main repo, which is ASK-1874's latency paid twice.

**minor** — One Skill engine call writes two identical rows to the round's `engines.jsonl`, because the PostToolUse door is registered twice.
`q-system/.q-system/scripts/design-engine-door.py:74`

Reproducer (two invocations, which is what the wiring above produces for one Skill call):
```bash
python3 -c '
import json, os, subprocess, sys, tempfile, pathlib
root = pathlib.Path(".").resolve()
tmp = pathlib.Path(tempfile.mkdtemp(prefix="dcrev-"))
rd = tmp / "rounds" / "r1"; rd.mkdir(parents=True); (rd / "brief.md").write_text("x")
state = tmp / "state"; state.mkdir(); sid = "sess-repro"
(state / (sid + ".json")).write_text(json.dumps({"round": str(rd)}))
env = dict(os.environ, DESIGN_CHAIN_STATE=str(state))
payload = json.dumps({"hook_event_name":"PostToolUse","tool_name":"Skill",
   "tool_input":{"skill":"kipi-design:brand"},"session_id":sid,"cwd":str(tmp)})
door = str(root / "q-system/.q-system/scripts/design-engine-door.py")
for i in range(2):
    r = subprocess.run([sys.executable,"-I",door], input=payload, capture_output=True, text=True, env=env)
    print("  invocation", i+1, "exit", r.returncode)
rows = (rd / "engines.jsonl").read_text().splitlines()
print("engines.jsonl rows for ONE Skill call:", len(rows))
for l in rows: print("  ", l)'
```
Real output:
```
  invocation 1 exit 0
  invocation 2 exit 0
engines.jsonl rows for ONE Skill call: 2
   {"skill": "brand", "raw": "kipi-design:brand", "session": "sess-repro", "at": "2026-09-20T03:13:13Z"}
   {"skill": "brand", "raw": "kipi-design:brand", "session": "sess-repro", "at": "2026-09-20T03:13:13Z"}
```
Bounded on purpose: `engine_problems()` at `design-chain-gate.py:3172` reads the log into a set, so the seal verdict is unaffected. The record of how many times an engine ran is wrong, and that record is what a later reviewer reads.

**minor** — The two scopers shipped in this PR disagree about the same path. A marketing page under a `reports/` directory is public to `dogfood_gate.py` and internal to `design-chain-gate.py`, so it gets the slop tripwire and silently skips the entire chain.
`q-system/.q-system/scripts/design-chain-gate.py:77` (`INTERNAL_DIRS` gained `report`, `reports`, `dashboard`, `dashboards`, `schedule`, `schedules`)

Reproducer:
```bash
python3 -c '
import importlib.util, pathlib
def load(p,n):
    s=importlib.util.spec_from_file_location(n,p); m=importlib.util.module_from_spec(s); s.loader.exec_module(m); return m
gate = load("q-system/.q-system/scripts/design-chain-gate.py","g")
dog  = load("plugins/kipi-design/hooks/dogfood_gate.py","d")
for c in ["/site/reports/state-of-ai.html","/site/dashboard/index.html","/site/pricing.html","/site/schedule/book.html"]:
    print(c.ljust(36), "dc_gate_is_page=", gate.is_page(pathlib.Path(c)), " dogfood_public=", dog.is_public_facing_page(c))'
```
Real output:
```
/site/reports/state-of-ai.html       dc_gate_is_page= False  dogfood_public= True
/site/dashboard/index.html           dc_gate_is_page= False  dogfood_public= False
/site/pricing.html                   dc_gate_is_page= True   dogfood_public= True
/site/schedule/book.html             dc_gate_is_page= False  dogfood_public= False
```
A state-of-AI report landing page is exactly the kind of page this chain exists for. `design-auto-invoke.md` says the scoper errs toward internal on purpose, so this is the accepted direction of error, and the fix is one shared scoper rather than two lists.

## What is sound

- **The headline CI claim is not inert.** I suspected `DC_REQUIRE_REAL_PRODUCERS: "1"` set on one workflow step would never reach the spawned tests. `capability-gate.py:681` builds `env = dict(os.environ, QROOT=str(root / "q-system"))` and hands it to `run_contained`, so it propagates. The guard it feeds is real: `test_dc_seal_snapshot.py:42` imports playwright, and lines 45 to 47 fail instead of skipping when the var is set.
- **`engine_problems()` fails closed, not open.** A missing `engines.jsonl` hits `except OSError: pass` at line 3181 leaving `ran` empty, so every credited engine is refused. I expected a fail-open there.
- **The lane is not forgeable from the round.** `round_lane()` at line 1401 takes the lane from the instance config by where the round lives, refuses a `craft-manifest.json` `lane` that disagrees (line 1447), and the receipt binds the config's sha256 for any non-site lane (line 2823). I could not find a way to declare a failing web page a deck and skip the web-only stages.
- **The ledger race from the duplicate wiring is already covered.** `ledger_lock()` at line 2853 takes an exclusive flock on a sidecar file, deliberately not on the ledger, because `save_ledger` replaces the ledger through its own handle.
- **The new `sitecustomize.py` has a small blast radius.** It is inert unless `KIPI_REAL_PATH_LOG` names a file, every write is wrapped so a broken observer cannot change the observed run, and it chains the interpreter's own sitecustomize.
- **Four test files run green here**: `test_design_engine_door.py` (21), `test_dc_citations.py` (15), `test_dc_engine_records.py` (13), `test_dc_one_door_menu.py` (3). All OK.

Attacks I could not make fail and am dropping rather than dressing up: forging a seal through `_believed_na`, and escaping the door with a case or separator variant of an engine name (`canonical()` at `design-engine-door.py:53` folds case, spaces, slashes and underscores, and the test file covers it).

## VERDICT

**APPROVE WITH NITS.**

Most important thing to fix first: collapse the doubled hook registrations in `.claude/settings.json` down to the single `-I` form that `settings-template.json` already uses. It removes both the doubled per-tool-call latency and the duplicate `engines.jsonl` rows in one edit, and it makes the skeleton's own settings match the file that ships to the fleet.

FINDINGS:
minor|`.claude/settings.json` registers each design-chain hook twice (isolated and bare), doubling gate and door invocations on every Bash, Edit, Write, Skill and Stop|.claude/settings.json:171
minor|One Skill engine call appends two identical rows to the round's engines.jsonl because the PostToolUse door is wired twice|q-system/.q-system/scripts/design-engine-door.py:74
minor|dogfood_gate.is_public_facing_page and design-chain-gate.is_page disagree on `reports/` pages, so a real marketing page gets the slop tripwire but silently skips the whole chain|q-system/.q-system/scripts/design-chain-gate.py:77
END FINDINGS

PR #375, opened 2026-09-19
fix(voice-stop-gate): port the refusal path from the instance (ASK-1865)

assafkip:
Last blocker on the ASK-1865 deploy. `automation/test_voice_stop_gate_propagation.py::test_the_two_copies_define_the_same_functions` fails on the deploy merge tree: `_consume_refusal` and `_refusal_text` are defined only in the consulting instance's copy of `voice-stop-gate.py`. Nothing is skeleton-only.

`q-system/` is a fleet fan-out rsync destination with a delete flag, so the next fan-out would have deleted both functions off the founder's machine. The test's own message says port the missing side, never add an EXEMPT entry. This ports it.

## The defect being fixed

A routed turn that REPORTS a refusal was deadlocked. `_verify_route_receipt` understood one receipt status, `complete`, so it hashed the turn as if it carried a draft, found none, and held it with "route receipt does not match the assistant output". The session could not deliver the verdict it had just produced. Measured 2026-09-15, three LinkedIn reply runs in one evening.

## What is ported

- `_REFUSAL_MARKER`, the wire-protocol string the producer prints above its reason
- `_refusal_text`, which reads the reason back out and stops at the receipt block
- `_consume_refusal`, which spends a refused row and refuses any turn that also carries a draft
- the call site in `_verify_route_receipt`; the `identity` binding moves two lines up so the refused branch can use it

## Generic as written

No change was needed. The gate receives the instance's `route_contract` module as a parameter, and the lane path was already resolved from the instance tree by `_route_context`. The only instance dependency is `contract.verify_and_consume_refusal`, which is already wrapped so an older lane converts to a `RouteBoundaryError` and HOLDS the turn.

**The 24-instance claim is pinned, not assumed.** An instance whose lane never mints a refused receipt is unaffected (no `status` key, or `complete`, both tested), and a lane older than the contract method holds the turn rather than failing the hook open.

## Evidence

| Step | Result |
|---|---|
| reproducer, before the port | 11 failed, 2 passed |
| after the port | 13 passed |
| mutation: call site removed, mutant validated as applied | 3 wiring tests red, file restored byte-identical |
| propagation defect, via the checker's own AST logic | FAIL (instance-only: `_consume_refusal`, `_refusal_text`) then PASS |
| sibling gate suites | 106 passed |
| `test_voice_stop_gate_drain_only.py` | 6 passed, 1 skipped |
| pre-commit `verify.sh --staged` | green |

The 2 that pass in the red run are the backward-compat controls. They pass before AND after on purpose: they grade unchanged behaviour, not the feature, so a control that went green only after the port would be measuring the wrong thing.

## Notes

- **No settings change.** `voice-stop-gate.py` is already wired in both `.claude/settings.json` and `settings-template.json`, so the switch already ships. This changes behaviour inside an already-wired script.
- The test lives in `q-system/.q-system/tests/`, which is in `.verify-suites` and holds the three existing `test_voice_stop_gate_*.py` files. It loads the running gate by path, not a vendored copy.

Co-Authored-By: Claude Fable 5.1 <noreply@anthropic.com>


assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-375-20260919-165942.md` (6598 bytes).

```
FINDINGS:
major|A refusal turn whose text names the platform trips extract_publishable's whole-message fallback and is held as carrying a draft, re-creating the deadlock this PR fixes|q-system/.q-system/scripts/voice-stop-gate.py:136
major|The refused path consumes the receipt for a turn carrying a fenced draft with no framing sentence, so an ungraded draft ships under a no-draft receipt|q-system/.q-system/scripts/voice-stop-gate.py:136
minor|_refusal_text hashes every line between the marker and the receipt, so one added sentence holds the turn with a message that reads as a forged reason|q-system/.q-system/scripts/voice-stop-gate.py:103
END FINDINGS
```

--- reviewer output ---

## Review: PR #375 — port the refusal path into the skeleton Stop gate

Read the diff, loaded the running gate by path, ran the new suite (13 passed), the four gate suites (106 passed), and mutated the call site in memory. Three findings survived reproduction.

---

### 1. MAJOR — a refusal turn that names the platform is held as if it carried a draft

`q-system/.q-system/scripts/voice-stop-gate.py:136`

The no-draft guard calls `extract_publishable(assistant_text)`, which falls back to **the entire message** whenever publish framing is present and nothing is set off. A plain refusal that says "LinkedIn", "for X", "the draft for this reply", or "drafting the comment" trips the framing regex, so the whole refusal message becomes the "draft" and the turn is held. That is the same deadlock this PR exists to end, with a message that says the turn carries a draft it does not carry.

Reproducer (ran end to end through `_verify_route_receipt` with a stub contract):

```
A no platform word: PASS  -> {'status': 'refusal-consumed'}
B turn names plat : HELD  -> a refused route may not deliver a draft; this turn carries one
C reason names it : HELD  -> a refused route may not deliver a draft; this turn carries one
```

Breadth check on six plausible refusal reasons, no receipt involved:

```
clean | 300 words, above the 200-word ceiling
TRIPS | the draft for this reply ran 300 words, over the ceiling
TRIPS | drafting the comment hit the corpus-similarity floor
TRIPS | no source in the corpus supports the claim, so there is no reply to X
TRIPS | the reply ran long for LinkedIn
clean | the poster's claims were not extracted, so I am not writing the comment
```

Four of six. The new suite has no negative control for this: every test in `TestWhatARefusalMayNotDo` proves the check fires, none proves it does not over-fire.

Honest limit: the producer's exact wording lives in the consulting instance, outside this repo, so I could not confirm the live lane emits one of these strings. The check reads the whole assistant turn, which the model writes freely, so the input is not an artificial one.

At 3am this holds the session and the founder sees a refusal about a draft that does not exist.

---

### 2. MAJOR — the same guard misses a fenced draft with no framing sentence, which this file calls his most common shape

`q-system/.q-system/scripts/voice-stop-gate.py:136`

`_consume_refusal`'s docstring says a refused row "can never authorize a turn that carries a draft". It can. `extract_publishable` returns `''` when no framing marker is present, and `main()`'s own comment at line 1084 says the founder's most common turn is "the post in a fence and no here's-the-post sentence". `_refusal_text` only hashes the region after the refusal marker, so everything before it is unbound. Put a fenced post above the refusal block and the receipt is spent, the turn completes, and nothing grades the post.

```
route gate       -> {'status': 'refusal-consumed'} | refusals spent: 1
voice-lint sees  -> ''
reply_carries_a_draft -> True
MIN_TEXT_BYTES gate  -> lint runs? False
```

The complete path does not have this hole: there `output_hash` binds the draft. The refusal path is strictly weaker than the one it sits beside.

Honest limit: I could not drive full `main()` here, because the skeleton has no `q-consult/pipeline`, so `_route_context` returns `None` and the gate no-ops. The three helper outputs above are the ones `main()` consults on that path.

---

### 3. MINOR — one sentence added after the reason reads as a forged reason

`q-system/.q-system/scripts/voice-stop-gate.py:103`

`_refusal_text` takes everything from the refusal marker to the receipt block. A natural follow-up line joins the hashed reason, and the turn is held with "refusal receipt does not match the stated reason", which reads as tampering rather than as an extra sentence.

```
reason, clean turn    : '300 words, above the 200-word ceiling'
reason, one follow-up : '300 words, above the 200-word ceiling\n\nWant a shorter one?'
hashes equal          : False
```

---

### What is sound

Attacks the code survived, each actually run:

- **The wiring is load-bearing.** Removed the call site at line 222 in memory, verified the mutant applied (128 bytes), ran both versions: `ORIGINAL -> PASS`, `MUTANT -> HELD route receipt does not match the assistant output`. The wiring tests would go red.
- **Fail-closed on an older lane.** A contract without `verify_and_consume_refusal` raises `AttributeError` inside the `try`, converts to `RouteBoundaryError`, and holds. No exception escapes to exit the Stop hook 1.
- **Backward compatibility holds.** A `complete` receipt and a receipt with no `status` key both take the old path and spend `verify_and_consume`, never the refusal. Ran both.
- **The reason is bound to the receipt.** A turn cannot mint a refusal for a hard verdict and report a friendlier one; the hash catches it.
- **No fixture invented a shape that hides the bug.** `MATCH_FIELDS` is read off the contract, so the stub is duck-typed the way the real caller is.
- **Sibling suites green.** 106 passed across the four `test_voice_stop_gate_*.py` files, matching the PR's claim.
- **The 12 capability-gate REDs are not this PR.** They are `declared_inert` scripts citing resolved spillover ids, untouched here.

---

### VERDICT: REQUEST CHANGES

Not BLOCK. Nothing here is permanent: a held turn is re-driveable and an ungraded draft is read by the founder before it goes anywhere.

**Fix first:** swap `extract_publishable(assistant_text)` for `extract_setoff_draft(assistant_text)` at line 136. One substitution closes both majors, measured:

```
plain refusal  | extract_publishable: True  | extract_setoff_draft: False
unframed fence | extract_publishable: False | extract_setoff_draft: True
```

Then add the negative control the suite is missing: a plain refusal whose reason names the platform must PASS.

FINDINGS:
major|A refusal turn whose text names the platform trips extract_publishable's whole-message fallback and is held as carrying a draft, re-creating the deadlock this PR fixes|q-system/.q-system/scripts/voice-stop-gate.py:136
major|The refused path consumes the receipt for a turn carrying a fenced draft with no framing sentence, so an ungraded draft ships under a no-draft receipt|q-system/.q-system/scripts/voice-stop-gate.py:136
minor|_refusal_text hashes every line between the marker and the receipt, so one added sentence holds the turn with a message that reads as a forged reason|q-system/.q-system/scripts/voice-stop-gate.py:103
END FINDINGS

assafkip:
## Review round 1 addressed, and where the fix still has to land

Pushed `667b21d7`. All three findings fixed, reproducer first for each. Detail is in the commit message.

**These are not defects the port introduced.** I ported verbatim, which was the point, so all three holes are live right now in the consulting instance's own copy at `q-system/.q-system/scripts/voice-stop-gate.py`, commit `1de2bd59`. The instance is where the route lane actually runs, so that copy is the one where a refusal turn is being held today.

**What carries the fix there:** the fleet fan-out, once this merges. `q-system/` in every instance is an rsync destination for the skeleton's copy, so the merged file replaces the instance copy on the next fan-out. Nothing needs to be hand-applied to the instance, and hand-applying it would be undone by the next sync anyway. That is the same loop that produced this PR: the instance fixed its own copy, the skeleton never got it, and the fan-out would have deleted the fix.

Order matters: `~/projects/kipi-scheduled` has to carry this before a fan-out helps, which is a fast-forward pull of `main` after the merge.

**One limit is now written down rather than assumed.** An inline unfenced draft in a refusal turn is not caught. Catching it means restoring `extract_publishable`'s whole-message fallback, which is what held 4 of 6 real refusal wordings. `reply_carries_a_draft`, the detector `main()` asks the same question, does not catch it either, so the refusal path is exactly as strong as the path beside it and no weaker. Filed as sp-5262a341 / ASK-1889, and the test that used to assert the opposite now asserts the limit and names the id.

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-375-20260919-171506.md` (7328 bytes).

```
FINDINGS:
major|A fenced draft placed below the route receipt block spends a refused receipt and completes the turn, because _unbound_text truncates at the receipt marker|q-system/.q-system/scripts/voice-stop-gate.py:147
minor|Any blockquote above the refusal marker, including the model quoting the post it was asked to reply to, holds the turn as carrying a draft|q-system/.q-system/scripts/voice-stop-gate.py:195
minor|The trailing-line relaxation in _reason_candidates only covers single-paragraph reasons, so a multi-paragraph reason plus a closing line is held as a forged reason|q-system/.q-system/scripts/voice-stop-gate.py:117
minor|The refusal door is selected by a message-supplied status field that no hash binds and MATCH_FIELDS does not carry, so a draft printed as the bound reason skips verify_and_consume and its R9 recomputation|q-system/.q-system/scripts/voice-stop-gate.py:280
END FINDINGS
```

--- reviewer output ---

# Review: PR #375, voice-stop-gate refusal path

**What I did.** Built a throwaway instance at `/tmp/pr375lab/inst` (copy of `q-system/.q-system/scripts/`, plus a stub `q-consult/pipeline/` so `_route_context()` resolves), drove the real hook end to end over a real transcript, and ran the same shapes directly against `_verify_route_receipt`. Read-only on the repo, nothing modified.

**The fix itself is real.** Pre-PR gate on the plain refusal turn:

```
PRE-PR gate, plain refusal turn -> exit 2
STDERR: voice-stop-gate: route receipt does not match the assistant output
```

Post-PR, same turn, same transcript: `=== A_plain_refusal === exit 0`. The deadlock is genuinely closed.

---

## Finding 1 — major

**A fenced draft placed BELOW the receipt block spends a `refused` receipt and completes the turn.** `_unbound_text` truncates at `=== ROUTE RECEIPT ===` (voice-stop-gate.py:147-150), so the guard at line 195 only sees text above the receipt. The comment at line 191 claims "a fenced post written above the refusal block is still caught (major 2)" — one line lower is not caught.

Reproducer and real output:

```
F1 (line 195 / _unbound_text truncation)
  fence ABOVE refusal block -> HELD: a refused route may not deliver a draft; this turn carries one
  fence BELOW receipt block  -> ACCEPTED {'status': 'refusal-consumed'}
```

End to end through the hook, with a draft that passes voice-lint:

```
=== B3_clean_draft_below_receipt === exit 0
```

The turn completes, the refusal row is spent, and nothing bound that draft by `output_hash`. The model writes the text around these markers, so trailing content after a block is an ordinary shape, not a constructed one.

## Finding 2 — minor

**`extract_setoff_draft` over the head means any blockquote above the refusal marker holds the turn.** A refusal turn where the model quotes the post it was asked about:

```
F4 (blockquote above the marker)
  model quotes the source post -> HELD: a refused route may not deliver a draft; this turn carries one
```

End to end: `=== C_quoted_source_post_above === exit 2`. This is the over-fire class the author ranked major 1 in an earlier round, and `.claude/rules/social-reaction-gate.md` pushes the reply lane toward showing the poster's claims in the same turn. Dropped to minor because I constructed that turn: no producer in this repo emits it, and `test_a_blockquoted_draft_above_the_refusal_block_is_held` shows the hold is deliberate.

## Finding 3 — minor

**The trailing-line relaxation only covers single-paragraph reasons.** `_reason_candidates` (line 117) offers exactly two candidates, the whole region and its first paragraph. A multi-paragraph reason plus one closing line matches neither:

```
F2 (_reason_candidates covers single-paragraph reasons only)
  multi-para reason, no trailing line  -> ACCEPTED {'status': 'refusal-consumed'}
  multi-para reason + trailing line   -> HELD: refusal receipt does not match the stated reason
```

End to end: `=== D_multipara_reason_plus_trailing_line === exit 2`. The author's own `test_the_whole_region_still_binds_when_it_matches` establishes multi-paragraph reasons as a real producer shape, and `test_a_follow_up_line_after_the_reason_does_not_break_the_hash` uses a single-paragraph reason, so it passes without touching the gap. The error text is the misleading one the docstring calls out: reads as tampering, is an extra sentence.

## Finding 4 — minor

**The door is chosen by `status`, which no hash binds.** Line 280 reads `status` straight out of the message-supplied receipt JSON; `MATCH_FIELDS` does not include it, so `identity` does not carry it. A turn that prints the draft under the refusal marker, with a row whose `output_hash` binds that draft, takes the refusal door:

```
F3 (status is unbound; draft delivered as the reason)
  draft printed as the bound reason -> ACCEPTED {'status': 'refusal-consumed'}
```

That skips `verify_and_consume(identity, draft=draft)` and the R9 loop-evidence recomputation the complete path documents at line 287. The docstring at line 161 says this rule is "enforced HERE rather than in the store," so "the store will catch it" is not available as a defense. Minor rather than major because whether consulting's real `verify_and_consume_refusal` re-checks the row status is unverifiable from here: the sandbox blocks reads outside this worktree, and the only contract in the repo is the test's own stub.

---

## What is sound

- **Backward compatibility.** `complete` receipts and receipts with no `status` key take the old path untouched, and a contract missing `verify_and_consume_refusal` raises `RouteBoundaryError` rather than escaping as an `AttributeError` that would exit the hook 1 and fail open. Verified by running, not by reading the tests.
- **The `extract_publishable` regression.** All six platform-naming refusal wordings pass end to end. I re-ran the whole-message-fallback shape that caused the earlier hold; it no longer fires.
- **Quoting inside the bound reason.** A refusal that shows the rejected text as a blockquote inside the reason passes. `_unbound_text` is load-bearing there, not decorative.
- **Receipt JSON in the reason hash.** `_refusal_text` stops at the receipt marker; I confirmed the hash matches with and without the receipt appended.
- **Forged reason.** A friendlier reason than the bound one is refused, including after the `_reason_candidates` relaxation.
- **Suites.** `test_voice_stop_gate_refusal.py` plus the three sibling gate suites: 91 passed, 1 skipped. The full `q-system/.q-system/tests` run did not finish inside the review window, so I am not claiming it green.
- **Unverifiable, dropped:** `automation/test_voice_stop_gate_propagation.py` is not in this repo and the consulting copy is outside the sandbox, so I could not check whether the two new skeleton-only helpers (`_reason_candidates`, `_unbound_text`) re-break `test_the_two_copies_define_the_same_functions` in the other direction. Worth a look before the deploy merge; I could not reproduce it and am not filing it.

## VERDICT: REQUEST CHANGES

Not BLOCK. Nothing here is permanent or unrecoverable, and the core fix works.

**Fix first:** Finding 1. `_unbound_text` should return the text after the receipt block too, not drop it. One line, and it closes the same hole the reviewer already made you close one section higher.

FINDINGS:
major|A fenced draft placed below the route receipt block spends a refused receipt and completes the turn, because _unbound_text truncates at the receipt marker|q-system/.q-system/scripts/voice-stop-gate.py:147
minor|Any blockquote above the refusal marker, including the model quoting the post it was asked to reply to, holds the turn as carrying a draft|q-system/.q-system/scripts/voice-stop-gate.py:195
minor|The trailing-line relaxation in _reason_candidates only covers single-paragraph reasons, so a multi-paragraph reason plus a closing line is held as a forged reason|q-system/.q-system/scripts/voice-stop-gate.py:117
minor|The refusal door is selected by a message-supplied status field that no hash binds and MATCH_FIELDS does not carry, so a draft printed as the bound reason skips verify_and_consume and its R9 recomputation|q-system/.q-system/scripts/voice-stop-gate.py:280
END FINDINGS

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-375-20260919-171731.md` (6044 bytes).

```
FINDINGS:
major|A fenced or blockquoted draft written below the route receipt block escapes the refusal draft guard, because _unbound_text truncates the turn at the receipt marker|q-system/.q-system/scripts/voice-stop-gate.py:148
minor|A refusal turn is still handed whole to the voice lints as a publishable draft for 4 of the 6 enumerated refusal wordings, so the deadlock survives on the lint path|q-system/.q-system/scripts/voice-stop-gate.py:1139
END FINDINGS
```

--- reviewer output ---

## Review: PR #375, voice-stop-gate refusal path

One major survived reproduction. The new refusal guard blocks a draft written above the receipt block and lets the identical draft through when it is written below it.

---

### major 1: a draft below the route receipt escapes the refusal guard

`q-system/.q-system/scripts/voice-stop-gate.py:148`

`_unbound_text` truncates the turn at `=== ROUTE RECEIPT ===`, so every byte after the receipt block is invisible to the draft check that `_consume_refusal` runs. Same fenced post, two positions, two opposite verdicts.

Reproducer (run from the worktree, stubbed contract copied from the PR's own test file):

```
python3 -c '<the _verify_route_receipt harness, fence above vs below the receipt>'
```

Real output:

```
fence ABOVE receipt  -> HELD: a refused route may not deliver a draft; this turn carries one
                        setoff_draft(whole turn)='nobody reads the dashboard, they read the alert'
fence BELOW receipt  -> CONSUMED {'status': 'refusal-consumed'} refusals=1
                        setoff_draft(whole turn)='nobody reads the dashboard, they read the alert'
```

The mechanism, isolated:

```
unbound text handed to the draft check = '\n'
extract_setoff_draft(unbound)          = ''
```

Why it is not a constructed-only input: the second commit exists precisely because the model appends text after what the producer printed ("want me to try a shorter one?"). That fix covers appends between the reason and the receipt. The receipt is the last thing the producer prints, so an append after it is the more likely position, not the less likely one. `TestAFencedDraftUnderARefusalIsStillADraft` has two cases and both put the fence above.

Blast radius at 3am: a refused route ships a draft with no bound `output_hash`, and in the wording I tested `extract_publishable` returns empty, so the voice lints do not grade it either. Recoverable, nothing permanent, so major and not blocker.

---

### minor 2: the refusal reason is still linted as a publishable draft

`q-system/.q-system/scripts/voice-stop-gate.py:1139` reached from `:1266`

`main()` runs `extract_publishable` before `enforce_route_receipt`. For 4 of the 6 refusal wordings this PR itself enumerates in `PLATFORM_REASONS`, the reason contains publish framing, nothing is set off, so the fallback returns the whole message including the receipt JSON, above the 80-byte floor:

```
MIN_TEXT_BYTES = 80
skip bytes=0     lints_whole_turn=False  ch=''         | 300 words, above the 200-word ceiling
HIT  bytes=391   lints_whole_turn=True   ch=''         | the draft for this reply ran 300 words, over
HIT  bytes=387   lints_whole_turn=True   ch=''         | drafting the comment hit the corpus-similarit
HIT  bytes=404   lints_whole_turn=True   ch='x'        | no source in the corpus supports the claim, s
HIT  bytes=366   lints_whole_turn=True   ch='linkedin' | the reply ran long for LinkedIn
skip bytes=0     lints_whole_turn=False  ch=''         | the posters claims were not extracted, so I a
```

Two of them also set `detect_channel`, so an instance with a channel registry grades a refusal against a platform surface lint. With a reason carrying an emdash or a banned word, voice-lint exits 2 before the receipt is ever consumed:

```
[emdash in reason] bytes=410 voice-lint exit=2
  line 4 [emdash] em dash (use comma, period, or hyphen instead)
[banned word]      bytes=405 voice-lint exit=2
  line 4 [banned-word] banned word: 'leverage'
```

That is the same deadlock the PR exists to end, relocated to the lint path, and the reason is hash-bound so the model cannot edit it to satisfy the lint without breaking the receipt.

Dropping this to minor for two honest reasons: `main()`'s lint ordering is untouched by this diff and behaves the same before and after, and the two blocking wordings are inputs I constructed, since the producer lives outside this repo and I could not read what it actually emits.

---

### What I tried and could not reproduce

I could not run `automation/test_voice_stop_gate_propagation.py::test_the_two_copies_define_the_same_functions`, the test the PR body names as its blocker. It is not in this repo (`grep -rln two_copies_define_the_same_functions .` returns nothing) and my read of the consulting instance was denied by the sandbox. So I cannot confirm the claim that this port turns it green. Worth someone checking, since the diff adds `_reason_candidates` and `_unbound_text` to the skeleton only, and if that checker is symmetric it is now red in the opposite direction.

### What is sound

- A refusal turn with no refusal text, held with "needs the refusal text".
- A reason that does not match the bound hash, held; the first-paragraph fallback does not weaken this, since every candidate is hash-checked.
- A reason containing `=== ROUTE RECEIPT ===`, truncated and then held on the hash. Fails closed.
- A lane with no `verify_and_consume_refusal`, converted to `RouteBoundaryError` inside the try, so the hook holds rather than exiting 1 and failing open.
- A `complete` receipt and a receipt with no `status` key, both unchanged, and neither spends a refusal row.
- The 28 new tests pass, and the 3 sibling gate suites pass at 93.

### VERDICT: REQUEST CHANGES

Fix first: `_unbound_text` at line 148 discards the tail. Keep the receipt JSON itself out of the draft check, but the text after the receipt block belongs in the unbound region, and `TestAFencedDraftUnderARefusalIsStillADraft` needs a third case with the fence below the receipt.

FINDINGS:
major|A fenced or blockquoted draft written below the route receipt block escapes the refusal draft guard, because _unbound_text truncates the turn at the receipt marker|q-system/.q-system/scripts/voice-stop-gate.py:148
minor|A refusal turn is still handed whole to the voice lints as a publishable draft for 4 of the 6 enumerated refusal wordings, so the deadlock survives on the lint path|q-system/.q-system/scripts/voice-stop-gate.py:1139
END FINDINGS

assafkip:
## Round 2 answered: the major is fixed, the three minors are captured

Pushed `ab47624d`.

**Fixed: finding 1 (major).** `unbound_text` no longer truncates at the receipt marker. The bound reason and the receipt come out as SPANS, so everything else, above the refusal marker and below the receipt alike, reaches the draft check. The receipt's JSON is stepped over with the same decoder `_receipt_block` uses, so trailing text survives. Your reproducer is now two tests, red before and green after, plus a negative control proving ordinary trailing prose still passes. Mutation: restoring the truncation turns exactly those two red. The comment you called out is rewritten to claim only what the code does.

**You flagged one thing as unverifiable from your sandbox, and you were right to.** `_reason_candidates` and `_unbound_text` were module level, so they were SKELETON-ONLY and re-broke `test_the_two_copies_define_the_same_functions` in the other direction. That is this PR's whole purpose, relocated rather than fixed. Measured with the checker's own AST logic: skeleton-only was `['_reason_candidates', '_unbound_text']`. Both are now nested inside `_consume_refusal`, and both directions are clean. Thanks for the pointer; I would not have looked.

**Captured, not fixed here**, so the next review reads this first:

- sp-d4bacb9f / ASK-1893, finding 2: a blockquote above the marker over-fires.
- sp-d1630e3d / ASK-1894, finding 3: the paragraph fallback misses a multi-paragraph reason.
- sp-ecb986e8 / ASK-1895, finding 4: `status` picks the door and no hash binds it.

On finding 4 I agree with minor over major, and here is the reasoning rather than a bare ruling. The model cannot forge a receipt row, so the exploit needs the lane to have actually minted a `refused` row whose bound reason is itself publishable text. What is genuinely skipped is the R9 loop-evidence recomputation, not draft grading. Worth confirming that consulting's real `verify_and_consume_refusal` re-checks the stored status, which is what the ticket says.

All six platform-naming refusal wordings from round 1 remain green as negative controls. Suites: 31 in this file, 93 across the three siblings, and the full `q-system/.q-system/tests` run was 1448 passed on the previous head.

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-375-20260919-173215.md` (6447 bytes).

```
FINDINGS:
minor|_consume_refusal's docstring claims it refuses "any inline publishable body" but the code only refuses set-off drafts, and the file's own test asserts the inline case passes|q-system/.q-system/scripts/voice-stop-gate.py:122
minor|The trailing-line fix only works for a single-paragraph bound reason; a multi-paragraph reason plus one added line is held with "does not match the stated reason"|q-system/.q-system/scripts/voice-stop-gate.py:141
minor|The refusal door is selected from receipt["status"], which no hash covers and MATCH_FIELDS omits, so a relabeled receipt skips the R9 draft recompute in verify_and_consume|q-system/.q-system/scripts/voice-stop-gate.py:305
nit|Test docstring names `_unbound_text`, which does not exist; the nested helper is `unbound_text`|q-system/.q-system/tests/test_voice_stop_gate_refusal.py:279
END FINDINGS
```

--- reviewer output ---

Review complete. Three findings survived reproduction, all minor. Every mutant I planted went red, so the new suite is load-bearing.

## Findings

**minor — the `_consume_refusal` docstring claims a guard the code does not have**

`q-system/.q-system/scripts/voice-stop-gate.py:122` says "the producer's refusal is the only text this turn may deliver, so a `=== DRAFT ===` marker or any inline publishable body refuses." The code refuses only a set-off draft. The file's own test at `test_voice_stop_gate_refusal.py:206` asserts the inline case must PASS. The docstring and the test contradict each other, in the one function whose job is that guard.

Reproducer and real output:

```
claim: "a === DRAFT === marker or any inline publishable body refuses"
  extract_publishable sees a publishable body: True
  gate verdict: {'status': 'refusal-consumed'}
```

The inline body was `Here's the LinkedIn post: nobody reads the dashboard, they read the alert.` placed under the refusal block. `extract_publishable` calls it publishable; the gate spends the refusal anyway. The limit itself is fine and is captured as sp-5262a341. The docstring is the finding.

**minor — the trailing-line fix only covers a single-paragraph bound reason**

`voice-stop-gate.py:141` builds the reason candidates as the whole region, then the first paragraph. When the producer's bound reason is itself multi-paragraph, the first-paragraph fallback cannot match it, so one added closing line holds the turn with "does not match the stated reason". That is the tampering-flavoured deadlock the fix exists to end, and the file's own `test_the_whole_region_still_binds_when_it_matches` declares multi-paragraph reasons a supported producer shape.

Real output:

```
=== R2: MULTI-paragraph bound reason + one trailing line -> still held ===
  HELD: refusal receipt does not match the stated reason

=== R2b: SINGLE-paragraph bound reason + one trailing line -> passes (the shipped fix) ===
   {'status': 'refusal-consumed'}
```

The 3am shape: the message points at the reason, not at the trailing line, so the model's likely repair is to rewrite the reason, which breaks the hash harder.

**minor — the refusal door is chosen from an unsigned field, which skips the R9 draft recompute**

`voice-stop-gate.py:305` selects the refusal path on `receipt.get("status") == "refused"`. `status` is in no hash and not in `MATCH_FIELDS`, so it is not part of the identity handed to the store. A receipt minted over a draft, with `status` flipped and the draft text printed as the reason, takes the refusal door: the draft is inside the bound span so the delivery guard cannot see it, and `verify_and_consume(identity, draft=draft)` with its R9 loop-evidence recompute never runs.

Real output:

```
=== R1: a COMPLETE receipt relabeled status=refused ships its draft through the refusal door ===
  result           : {'status': 'refusal-consumed'}
  refusal door used: True
  complete door    : []
  R9 draft recompute (verify_and_consume(draft=...)) ran: False
  the draft text is in the turn the founder reads: True
```

Dropping this a level on purpose: the relabel is an input I constructed, and I could not read the producer to show any lane emits it. The non-constructed half is already blessed by the repo's own `test_a_reason_that_QUOTES_the_draft_it_rejected_still_passes`, where a full copy-pasteable draft rides inside the bound reason and `extract_setoff_draft` pulls it straight back out:

```
verdict: {'status': 'refusal-consumed'}
publishable text the founder can copy out of this refusal turn: 'nobody reads the dashboard, they read the alert. ship the alert.'
```

**nit — stale name in a test docstring**

`test_voice_stop_gate_refusal.py:279` says "this is what makes `_unbound_text` load-bearing". There is no `_unbound_text`. The helper is nested and named `unbound_text`.

## What is sound

- Span removal beats every escape I tried. A fence or blockquote below a pretty-printed receipt, below a same-line receipt, after a second receipt block, after three blank lines, above the refusal marker, or with the DRAFT marker only in the chat above: all six HELD. The round-2 fix is real.
- The suite is not decoration. Four mutants, four reds. Dropping the below-receipt text turned 2 tests red. Removing the first-paragraph fallback turned 3 red. Scanning the whole turn instead of the unbound spans turned the quoted-reason test red. Removing the call site turned 15 red.
- The 24-instance claim holds where I could test it. No `status` key and `status: complete` both take the old path untouched, and a contract missing `verify_and_consume_refusal` holds the turn instead of failing the hook open.
- The voice-lint deadlock I expected is not there. I ran all six `PLATFORM_REASONS` through `main()`'s real extraction and both lints. Four take the big lint path with the receipt JSON folded into the linted text, and all six exit 0. No pre-receipt block.
- Sibling gate suites: 130 passed, 1 skipped.
- One thing I could not check: the PR's premise is `automation/test_voice_stop_gate_propagation.py`, which does not exist in this tree, and reading the consulting instance copy was denied. The claim that this port squares the two copies is unverified here, not disproven.

## VERDICT: APPROVE WITH NITS

Fix first: the `_consume_refusal` docstring at line 122. It promises a guard against inline publishable bodies that the code does not implement and the tests explicitly disclaim, which is how the next reader concludes the hole is already closed.

FINDINGS:
minor|_consume_refusal's docstring claims it refuses "any inline publishable body" but the code only refuses set-off drafts, and the file's own test asserts the inline case passes|q-system/.q-system/scripts/voice-stop-gate.py:122
minor|The trailing-line fix only works for a single-paragraph bound reason; a multi-paragraph reason plus one added line is held with "does not match the stated reason"|q-system/.q-system/scripts/voice-stop-gate.py:141
minor|The refusal door is selected from receipt["status"], which no hash covers and MATCH_FIELDS omits, so a relabeled receipt skips the R9 draft recompute in verify_and_consume|q-system/.q-system/scripts/voice-stop-gate.py:305
nit|Test docstring names `_unbound_text`, which does not exist; the nested helper is `unbound_text`|q-system/.q-system/tests/test_voice_stop_gate_refusal.py:279
END FINDINGS

PR #377, opened 2026-09-20
feat(ci): a pull request runs the tests its diff earns, not all of them (ASK-1749)

assafkip:
## The cost, measured in seconds

Run 35486006414, read from its own log: the capability gate step was **822s of an 899s `validate` job**. It ran all 236 declared test artifacts serially, on every push of every PR. A three-line review round waits that long before it can merge, and a converge run has up to four rounds. Founder, 2026-09-19: "make sure you dont run full sweeps on small issues like line changes. They should not run 6000+ tests."

## The saving, measured in seconds

**699s, against the 822s baseline. A saving of 123s, 15% of the gate step.**

Run 35488448685, `ask1914_probe` job, exit 0, 73 of 236 artifacts selected. Read from the run log, not computed here.

**That is well short of what this PR's earlier body implied, and it is the number that matters.** 31% of the artifacts carry 85% of the cost, because the 72 always-run tree-walkers are the expensive ones and no selection may skip them. A test SELECTOR cannot get under that floor. So this merges for the code it deletes and the foundation it lays, not for the 15%: the real follow-through is **ASK-1918**, where a scanner declares what it covers in its own fragment and leaves the floor.

Method, stated plainly because five earlier rounds reported test counts and never a second. This branch changes the classifier and the gate, so **by its own rule it earns the full suite**, and its own `validate` job therefore cannot show what a selected run costs. So a temporary probe job (reverted before merge, see below) checked out the PR head and ran the **shipped** gate against an ordinary one-file diff: a one-line comment change to `board_rows.py`, the most-edited ordinary script on main. Nothing about the gate is synthetic there. Only the base ref is.

**The ceiling, honestly.** `verify.yml` runs in parallel at a median of **498s** over its last 25 successful runs, measured the same way, and this PR does not touch it. `pytest plugins/prd-os/tests/` is another 56s inside `validate`. So time-to-green on a small PR has a hard floor around **8m20s** whatever this does. This is roughly half the wall clock, not a dramatic cut.

## What it selects

`change-size.py` reads the diff and answers two separate questions.

**Tier** (S at most 20 app-code lines, M at most 150, L beyond) is the human ceremony: how much review a change earns. It never by itself forces a full run.

**Tests** is the union of:
- **every declared test that walks the tree** — 72 of 236 today, counted from the checkout, not asserted. A test that reads files nobody named cannot be safely left out of any selection, so it always runs.
[...]

PR #384, opened 2026-09-20
feat(covers): a scanner declares what it covers, instead of always running (ASK-1918)

assafkip:
Stacked on #377. **Draft until #377 merges**, then this rebases onto clean main.

## What this adds

A scanner leaves `change-size.py`s always-run floor by declaring what it covers, one line in the capability fragment it already has:

```json
{ "path": "...", "runner": "bash", "covers": ["**/*.plist"] }
```

Zero declarations reproduce the old behaviour exactly, so it lands one fragment at a time.

## Why a declaration and not a parser

PR #377 rounds 1-4 tried to INFER coverage by parsing globs out of each tests source. Each round named another spelling the parser did not know, and round 4 gave up and made every scanner always-run. That surrender is why #377 measured only 15%.

## The finding that made this worth doing

Per-artifact timing (run 35489169103, 227 of 236 timed): the cost is a steep tail, not a flat 72. `test-severity-floor.sh` alone is 141s of an 873s suite. 30 of 227 carry 80% of the time.

**Seven of the eight most expensive scanners are FALSE scanners.** `scans_the_tree` fires on them because they glob, walk or `ls` -- inside their own mktemp/tmp_path fixture, never the repo. The blanket floor was paying 141s per PR to re-run a test that reads nothing a diff can change.

## The number

The `ask1918_probe` job measures it in CI, same one-file diff as the 699s baseline (run 35488448685), so the two differ only by the eight declarations. **Filled in when it lands** -- a derived number is not the measurement, which is ASK-1914s whole lesson.

## Safety

- `covers` may only take a scanner OFF the floor, never put one ON.
- A malformed declaration is dropped, leaving the test on the floor.
- The full suite still runs on every push to main.
- `covers-lint.py` refuses a glob matching no tracked file, a non-list, or a declaration on a non-scanner. Wired in lefthook and as a CI step BEFORE the gate. It caught a real error on first contact: the opening declaration named `pr-receipt-gate.py`, which does not exist.

Each of the eight was proved both ways: dropped from an unrelated diff, and still selected when one of its own dependencies changes. 8 for 8.

48 tests green across both files, three consecutive runs.

Closes ASK-1918.


assafkip:
Superseded by #385: same two commits, cherry-picked onto clean main now that #377 has merged. This branch was stacked on the pre-merge 1749 branch and carried probe scaffolding.

PR #385, opened 2026-09-20
feat(covers): a scanner declares what it covers, instead of always running (ASK-1918)

assafkip:
A scanner leaves `change-size.py`s always-run floor by declaring what it covers, one line in the capability fragment it already has:

```json
{ "path": "...", "runner": "bash", "covers": ["**/*.plist"] }
```

Zero declarations reproduce the old behaviour exactly, so this lands one fragment at a time with no flag day.

## Why a declaration and not a parser

PR #377 rounds 1-4 tried to INFER coverage by parsing globs out of each tests source. Each round named another spelling the parser did not know, and round 4 gave up and made every scanner always-run. That surrender is why #377 measured only 15%.

## The finding

Per-artifact timing (run 35489169103, 227 of 236 artifacts timed): the cost is a steep tail, not a flat 72. `test-severity-floor.sh` alone is 141s of an 873s suite; 30 of 227 carry 80% of the time.

**Seven of the eight most expensive scanners are FALSE scanners.** `scans_the_tree` fires on them because they glob, walk or `ls` -- inside their own mktemp/tmp_path fixture, never the repo. The blanket floor was paying 141s per PR to re-run a test that reads nothing a diff can change.

## The number, and its honest scope

Measured in CI on the same one-file diff, so the two differ only by the eight declarations:

| | seconds | selected |
|---|---|---|
| full gate (run 35486006414) | 822s | 236 |
| selected, no declarations (run 35488448685) | 699s | 73 of 236 |
| selected, 8 declarations (run 35490247290) | **341s** | 66 of 237 |

Corroborated independently: the eight artifacts timed individually total 379s, within 6% of the 358s the stopwatch showed.

**Caveats, stated rather than buried.** (1) Those runs were taken at 237 declared tests; main now carries 278, so absolute seconds will differ -- the delta re-checked on main still drops all eight. (2) The full-gate baseline is NOT stable: the same 236 artifacts on the same branch ran 822s (run 35486006414) and 1691s (run 35489807832), 2.06x apart. So read the 699s->341s delta, not a percentage off 822s.

## Safety

- `covers` may only take a scanner OFF the floor, never put one ON.
- A malformed declaration is dropped, leaving the test on the floor. Uncertain resolves upward.
- The full suite still runs on every push to main.
- `covers-lint.py` refuses a glob matching no tracked file, a non-list `covers`, or a declaration on a non-scanner. Wired in lefthook pre-commit and as a CI step BEFORE the gate, since a check on the declarations cannot be downstream of the thing they steer. All three refusals were seen RED before the test file existed.

**It caught a real error on first contact:** the opening `test-severity-floor` declaration named `pr-receipt-gate.py`, a stale reference inside the test that matches no tracked file. Without the lint the declaration would have been accepted.

Each of the eight was proved both ways: dropped from an unrelated diff, and still selected when one of its own declared dependencies changes. 8 for 8.

48 tests green across both files, three consecutive runs. Three mutants cover the new decision points.

Closes ASK-1918. Supersedes #384 (same work, stacked on the pre-merge branch).


assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-385-20260919-224107.md` (6987 bytes).

```
FINDINGS:
major|A `**/` covers glob never matches a root-level file, and covers-lint prints PASS on it, so a blessed declaration silently drops coverage for the repo's 39 root-level scripts|q-system/.q-system/scripts/change-size.py:187
minor|Scanner status, and therefore whether a `covers` key is legal, is decided by a regex that matches comments, so deleting one prose line blocks the commit|q-system/.q-system/scripts/change-size.py:162
minor|covers-lint's docstring claims a PostToolUse wiring that does not exist; hook mode has no production caller|q-system/.q-system/scripts/covers-lint.py:25
END FINDINGS
```

--- reviewer output ---

# Review: PR #385 — `covers` declarations take scanners off the always-run floor

Read-only. Nothing in the repo was modified.

---

## MAJOR 1: a `**/` glob silently excludes every root-level file, and covers-lint prints PASS on it

`q-system/.q-system/scripts/change-size.py:187`

`covers_matches` requires a literal `/` in the path for any pattern containing a slash. `fnmatch("**/*.sh", "kipi-update.sh")` is False. The fallback `"*/" + pattern.lstrip("*/")` needs a `/` too, so it does not rescue it. The declaration still matches nested files, so covers-lint finds a tracked match and exits 0.

This is the exact failure covers-lint's own docstring says it exists to prevent, arriving through the documented example form. `**/*.plist` is the canonical example in the PR body, in `covers_matches`' docstring, and in the new fixture at `test_change_size.py:212`. That fixture uses `PLIST = "q-system/.q-system/launchd/com.kipi.widget.plist"`, a nested path, so the suite cannot see the hole.

The real repo has 39 root-level `.py`/`.sh` files, including `kipi-update.sh`.

Reproducer, run against the shipped scripts:

```
python3 - . <<'PYEOF'   # full script in the transcript; builds a tmp git repo,
                        # one scanner declaring covers ["**/*.sh"], one nested
                        # .sh, one root-level kipi-update.sh
PYEOF
```

Real output:

```
covers-lint exit: 0 | covers-lint: PASS -- 1 of 1 checked fragment(s) declare coverage
read_declared covers: {'t/test-scan.sh': ['**/*.sh']}
ROOT-level .sh changed     -> scanner selected? False
nested .sh changed         -> scanner selected? True
```

And directly:

```
covers_matches('**/*.plist', 'com.kipi.foo.plist')            = False
covers_matches('**/*.plist', 'automation/com.kipi.foo.plist') = True
covers_matches('**/*.py',    'kipi.py')                        = False
```

The docstring says `**/*.plist` "behave[s] the way a person writing them expects". Git globstar matches a root-level file. The code does not.

**At 3am:** nothing pages. The scanner stops running for a class of paths, the lint says PASS, and the only backstop is the full suite on push to main.

---

## MINOR 2: scanner status, and therefore `covers` legality, is decided by a regex that matches comments

`q-system/.q-system/scripts/change-size.py:162` and `covers-lint.py:84`

`test-dispatch-stale-checkout.sh` is classified a scanner by exactly one line, and it is prose:

```
scan-matching lines in q-system/.q-system/scripts/test/test-dispatch-stale-checkout.sh
  line 19: # `git ls-files --others --exclude-standard` cannot see the second class at all,
all matches are comments? True
scans_the_tree WITH comment : True
scans_the_tree WITHOUT it   : False
```

Deleting that comment, with no functional change, blocks the commit. Cloned the repo to tmp, removed the line, ran the shipped lint:

```
AFTER deleting ONE COMMENT LINE (no functional change):
  exit: 2
  stderr: covers-lint: a coverage declaration is wrong (ASK-1918)
  ...test-dispatch-stale-checkout.sh.json: `covers` is INERT here -- change-size.py
  does not call ...test-dispatch-stale-checkout.sh a scanner, and `covers` only ever
  takes a scanner OFF the always-run floor. It never selects. Remove the
```

1 of the 8 declarations is comment-only. Bounded: the advice the message gives is harmless to follow, and the test still selects via `mentions`.

---

## MINOR 3: the docstring claims a PostToolUse wiring that does not exist

`q-system/.q-system/scripts/covers-lint.py:25`

Line 25 reads "blocks, per the PostToolUse contract", and the usage block describes a paths-taking "hook mode". Every reference to the script:

```
lefthook.yml
q-system/.q-system/scripts/covers-lint.py
q-system/.q-system/scripts/test_covers_lint.py
.github/workflows/validate.yml
```

`grep -c covers-lint .claude/settings.json q-system/.q-system/settings-template.json` returns `0` and `0`. Neither wired caller passes paths (`lefthook.yml:134` passes only `--repo-root`), so hook mode and the two tests that cover it (`test_hook_mode_fast_exits_on_an_out_of_scope_path`, `test_hook_mode_still_refuses_a_fragment_it_is_given`) exercise an entry point production never uses. `wiring-check.md` in this repo requires a new hook script to be referenced from `settings.json`.

---

## What is sound

**Completeness of the 8 declarations.** I enumerated every `_SCAN_RE` hit in all eight tests and checked which touch the real tree rather than a mktemp/`tmp_path` fixture. Exactly two do: `(DSSE.parent / "commands").glob("*.md")` and `(PLUGIN_ROOT / "commands").glob("*.md")`. Both are declared. The PR's "seven of eight are false scanners" claim holds.

**The `mentions` fallback still carries the by-name dependencies.** I ran the real planner over 13 real dependency paths. `reviewer-floor.sh`, `slack-notify.sh`, `review-redrive.py`, `fleet-health-daily.py`, `linear-sync.py`, `prd_runner.py`, `findings_writer.py`, `kipi-promote.sh`, `kipi`, `repo-preflight.sh`, `lessons_streak.py` each still select the test that exercises them. I found no orphan. `issue_findings.py` selects none of the eight, but it has three of its own named tests.

**The asymmetry holds.** A non-list `covers`, a list of empty strings, an empty list, and an absent key all leave the test on the floor. Confirmed in a tmp repo with the shipped `read_declared`.

**`covers` cannot put a non-scanner ON.** The `t in scanners` guard is load-bearing and its mutant is in `CS_MUTANTS`.

**Duplicate fragment paths** would make the missing `else: covers.pop(path)` branch in `read_declared` reachable. There are zero duplicates in the repo, so I dropped that finding.

**Cost of the pre-commit sweep:** 0.10s wall for all 278 fragments. Not a friction cost.

**Failure direction is upward.** `change-size.py` is in `FULL_RUN_PATHS`, so any edit to it full-suites; a lint crash exits non-zero and blocks; the CI step runs before the gate it steers.

**48 tests green**, both files, on this checkout.

---

## VERDICT: REQUEST CHANGES

Not BLOCK. Nothing here is unrecoverable and a `git revert` restores the floor exactly.

Fix first: `covers_matches` at `change-size.py:187`, so a `**/`-prefixed pattern matches a root-level path. One line, plus a fixture case using a root-level file rather than the nested one at `test_change_size.py:212`.

FINDINGS:
major|A `**/` covers glob never matches a root-level file, and covers-lint prints PASS on it, so a blessed declaration silently drops coverage for the repo's 39 root-level scripts|q-system/.q-system/scripts/change-size.py:187
minor|Scanner status, and therefore whether a `covers` key is legal, is decided by a regex that matches comments, so deleting one prose line blocks the commit|q-system/.q-system/scripts/change-size.py:162
minor|covers-lint's docstring claims a PostToolUse wiring that does not exist; hook mode has no production caller|q-system/.q-system/scripts/covers-lint.py:25
END FINDINGS

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-385-20260919-230149.md` (5930 bytes).

```
FINDINGS:
minor|A `covers` glob is not evaluated for a changed path that is itself a declared test, so the declaration is silently not honoured and covers-lint cannot see it|q-system/.q-system/scripts/change-size.py:335
minor|An interior `**/` does not match depth zero, the same defect the PR fixed for a leading `**/`, and covers-lint passes such a glob|q-system/.q-system/scripts/change-size.py:198
minor|read_declared's docstring claims a covers list holding non-strings is dropped; the code keeps the good entries and takes the test off the always-run floor|q-system/.q-system/scripts/change-size.py:448
END FINDINGS
```

--- reviewer output ---

## Review: PR #385 (ASK-1918)

Read-only. Nothing in the repo modified. 50 tests green locally (`test_change_size.py` + `test_covers_lint.py`, 1.25s), `covers-lint.py --repo-root .` exits 0 on the real tree with "8 of 278".

---

### Findings

**MINOR 1. A `covers` glob is silently not honoured when the changed path is itself a declared test.**
`q-system/.q-system/scripts/change-size.py:335`

`plan()` runs `continue` for any changed path in `declared` or `fragments` before line 351 evaluates coverage globs. So a declaration can claim a path and never select on it, and `covers-lint.py` has no branch that can see this.

Reproducer, run against the shipped tree:

```
covers of plugins/prd-os/tests/test_judgment_compiler.py claims declared test(s): ['plugins/prd-os/scripts/test/test_spillover_ledger_root.py']
   changing plugins/prd-os/scripts/test/test_spillover_ledger_root.py -> selects plugins/prd-os/tests/test_judgment_compiler.py ? False
```

Today's instance is inert (that path is not a real dependency of the judgment-compiler test), which is why this is minor and not major. The shape is the one the lint exists to stop: a declaration that reads as covering something and never selects for it. Dropped a level per the calibration rule, and saying so.

**MINOR 2. An interior `**/` misses depth zero, the same defect the PR fixed for a leading `**/` and labelled major there.**
`q-system/.q-system/scripts/change-size.py:198`

The new branch only anchors `**/` when it is the first three characters.

```
covers_matches('plugins/**/*.py', 'plugins/marketplace.py') -> False
covers_matches('plugins/**/*.py', 'plugins/prd-os/scripts/x.py') -> True
covers_matches('q-system/**/*.sh', 'q-system/x.sh') -> False
covers_matches('q-system/**/*.sh', 'q-system/.q-system/scripts/x.sh') -> True
```

`covers-lint.py` prints PASS on `q-system/**/*.sh` because 170 nested `.sh` files match, so the declaration gets blessed while the direct children are dropped. That is the exact sentence the docstring at line 187 uses about the bug it fixed. No declaration in the 8 uses an interior `**` today, so severity drops a level.

**MINOR 3. `read_declared`'s docstring contradicts the code on a mixed `covers` list.**
`q-system/.q-system/scripts/change-size.py:448` (claim) vs `:463`-`466` (code)

Docstring: "A malformed `covers` (not a list, **or holding non-strings**) is DROPPED." The code filters and keeps the survivors, so the test leaves the always-run floor on a partially garbage declaration. `test_read_declared_drops_a_malformed_covers` only exercises all-bad lists (`[None, ""]`), never a mixed one.

```
fragment covers : ['**/*.plist', None, 42, '']
read_declared   : {'t/test-mixed.sh': ['**/*.plist']}
unrelated diff  : selected_tests = [] -> empty means it LEFT the always-run floor
```

`covers-lint.py:99` refuses the non-string entries in CI and pre-commit, so blast radius is bounded to a `--no-verify` local commit. The stated asymmetry ("uncertain resolves upward") does not hold in this case.

---

### What is sound

- **The eight declarations under-declaring their real dependencies.** I extracted every literal tracked repo path appearing in each of the 8 test sources and re-ran the classifier one file at a time. 16 paths across 8 tests, **zero** failed to select their test. `mentions()` still carries them; `covers` is genuinely additive rather than a replacement. This was my main attack and it failed cleanly.
- **The "false scanner" claim on the expensive one.** Every `_SCAN_RE` hit in `test-severity-floor.sh` (lines 2131, 2134, 2285, 2372, 2375) is inside a `$Q1`/`$Q7` fixture tree, not the repo. The claim holds.
- **The two tests that DO scan the real tree.** `test_computed_receipts.py:508` globs `plugins/kipi-dsse/commands/*.md` and `test_judgment_compiler.py:2145` globs `plugins/prd-os/commands/*.md`. Both are declared. `test_computed_receipts.py` also executes four `plugins/prd-os/scripts/*.py` files it does not declare, and all four are caught by `mentions()`.
- **Rollback and blast radius.** `covers` only ever narrows PR selection. Push to main still runs the full suite (`validate.yml` `on: push: branches: [main]`, and the gate takes no `--diff-base` there). Reverting is deleting a JSON key. Nothing permanent, nothing unrecoverable.
- **Merge skew on a deleted covered script.** The deleting PR's own CI runs `covers-lint` against its own tree and goes red, so the declaration cannot outlive its target inside one PR.
- **Silent-skip visibility.** `capability-gate.py:778` prints `not-selected-for-this-diff=N`, so a narrowed run is not invisible to the operator.
- **Mutants are real.** The three new `CS_MUTANTS` entries each patch the decision point rather than a constant the assertion reads, which is the failure mode that rule exists for.

---

Only minor findings survived reproduction. The design is the safe direction: declarations can only take a scanner off the floor, malformed input keeps the expensive answer, and main re-runs everything.

**Fix first: MINOR 2.** It is a one-line generalisation of a branch the author already wrote and already called major in its own comment, and it is the finding most likely to bite the next person who writes a declaration.

FINDINGS:
minor|A `covers` glob is not evaluated for a changed path that is itself a declared test, so the declaration is silently not honoured and covers-lint cannot see it|q-system/.q-system/scripts/change-size.py:335
minor|An interior `**/` does not match depth zero, the same defect the PR fixed for a leading `**/`, and covers-lint passes such a glob|q-system/.q-system/scripts/change-size.py:198
minor|read_declared's docstring claims a covers list holding non-strings is dropped; the code keeps the good entries and takes the test off the always-run floor|q-system/.q-system/scripts/change-size.py:448
END FINDINGS

PR #386, opened 2026-09-20
fix(voiceloop): carry two generations of the engine into the skeleton (ASK-1915)

assafkip:
The skeleton was the OLDER copy of the voiceloop engine, and the fleet sync writes to `plugins/`. A sync would have reverted the newer engine out of consulting and every instance carrying it — the same class as the 2026-08-07 voicekit incident the destructive-op hook cites by name.

Authored by the consulting session; verified and landed here because kipi-system is this repo.

## What lands

Two generations. **2026-09-10**: `gate_and_judge` 184→249, `reply_format` 112→203, `voice_ref`, `tests/test_length_axis` — the skeleton never received this at all (zero occurrences of `ROTATION_WINDOW_MULT`). **2026-09-19**: `selector.py` drops the count cap that pinned rotation reach at 12 rows regardless of corpus size (13 of 106 reachable, measured live); `assemble`/`validate` make `target_words` keyword-only and REQUIRED.

**Breaking, deliberately.** `assemble.voice_section` and `voice.for_slot` now require `target_words`; a silent default was the defect. `validate.check_budget` and `check_correction_share` take an optional `targets` declaration, and a channel declaring nothing falls back to the old enumeration.

> **RETRACTION, and read this before syncing the fleet.** An earlier version of this body and the matching docstring said an undeclared instance grades byte-identically to before. **That is false.** The fallback *targets* are unchanged, but `check_budget` now also enumerates every slot kind and the `len(pool)` counters, and grades the longest end rather than the shortest, so the set of assemblies it reaches is strictly larger. Measured on a corpus with a fat comment slot and a thin post slot: the old shape graded one post-only assembly at **602 chars, CLEAN**; the new one reaches **24122 against a 24000 ceiling, RED**.
>
> The direction is correct, since the old green was the gate failing to look, which is the round-2 major. But **a fleet instance can go RED on a corpus its own suite passed yesterday, with no change to its own code.** That is a release note, not a no-op.

## The half that makes it safe

`voice_ref.py --slot-kind` is required (ASK-1400) because it used to default to `post`, so a caller asking for a REPLY was handed POSTS and told nothing (ASK-1399).

`required=True` and the loader no longer printing a command line are **two halves of one change.** The skeleton had both old halves, which was self-consistent. An earlier cut of this patch carried only the engine half, and that broke `voice-dna-loader.py:189`, which prints `voice_ref.py --channel x --words <target>` into every writing prompt fleet-wide. That invocation now exits 2 on a missing `--slot-kind`. Not a crash, which is worse: a fleet-wide instruction that fails for anyone who follows it.

So the loader travels here too, and the command is **deleted** rather than corrected, per the founder directive against command-line-only tools. `voice_ref.py` has no production caller; that hook was its only invoker, so removing the suggestion removes the trap instead of relocating it.

## Verified here, independently of the authoring session

Against a fresh `origin/main` extraction in a throwaway repo:

| check | result |
|---|---|
| `git apply -p0 --check` | applies cleanly (`-p1` does not) |
| voiceloop suite on the patched tree | 231 passed |
| loader EXECUTED, command in its stdout | clean 1 → patched 0 (7956 bytes still emitted) |

That last row is run by **execution, not grep**: the patched file still contains the command string inside the comment explaining its removal, so grepping the file reports 1 on both sides and proves nothing.

## Fleet blast radius, measured not assumed

Across all 26 registered instances: **zero callers of the breaking API outside `plugins/`.** 24 of 26 carry the same `voice-dna-loader.py`, which propagates from here, so one fix reaches all of them. The three remaining greps are false positives — a locally defined `voice_section()` in social-voice under instance-owned `output/`, a docs mention, and an unrelated regex variable in KTLYST_strategy.

`kipi-core` 1.12.8 → 1.13.9: the plugin cache is version-keyed, so without the bump every deployed clone keeps running the stale copy.

## Sent back before this

v1 carried 4 `.pytest_cache/` artifacts (not gitignored here, so they would have been committed and fanned to every instance) and only the engine half of the `--slot-kind` change.


assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-386-20260920-001828.md` (7330 bytes).

```
FINDINGS:
major|The 2026-09-19 count-cap removal ships unguarded: restoring the removed line keeps the suite at 231 passed while corpus reach drops 34 to 12 of 106, and the comment citing test_voice_reach.py as the pin names a file absent from this repo|plugins/kipi-core/voiceloop/selector.py:246
minor|ROTATION_WINDOW_MULT is now referenced only in comments, while its docstring claims lowering it to 1 restores the defect|plugins/kipi-core/voiceloop/selector.py:28
minor|All four usage examples in voice_ref.py's docstring now exit 2 because they carry no --slot-kind|plugins/kipi-core/voiceloop/voice_ref.py:16
minor|Comment above the changed line still says None is kept for pre-2026 callers, contradicting _targets_for's own docstring|plugins/kipi-core/voiceloop/validate.py:452
END FINDINGS
```

--- reviewer output ---

Review done. Read-only, nothing in the repo was modified.

## Finding 1 — major

**Claim:** The PR's headline 2026-09-19 change (removing the rotation count cap) ships to the fleet with no test that can detect its reversion, and two shipped comments cite a pinning test that does not exist in this repo.

**File:** `plugins/kipi-core/voiceloop/selector.py:246` (and `plugins/kipi-core/voiceloop/validate.py:571`)

**Reproducer 1, mutation.** Copy the package to a temp dir, put the exact removed line back, run the suite:

```
--- BASELINE unmutated --- rc=0
231 passed in 0.44s

--- MUTANT A count cap RESTORED (undoes the 2026-09-19 fix) --- rc=0
231 passed in 0.44s
```

**Reproducer 2, the behaviour that changed under a green suite.** 106 synthetic linkedin post rows, target 200, 31 counters:

```
reach WITH the PR (count cap removed): 34 of 106
reach with the cap PUT BACK          : 12 of 106
suite verdict on that same change    : 231 passed
```

**The cited pin is absent.** `selector.py:240` says "`test_voice_reach.py` in the consuming repo pins both halves against each other". `validate.py:571` says the declaration is "PINNED against the real call sites: `test_voice_reach.py` parses them".

```
$ find . -name "test_voice_reach*"
(no output)
$ grep -rn "targets" plugins/kipi-core/voiceloop/tests/
(no output)
```

**Three more carried changes survive full reversion, same class:**

- `_targets_for` forced to `declared = None`, so every deployment declaration is silently ignored: 231 passed.
- `voice_ref.py --slot-kind` put back to `default="post"`, undoing ASK-1400: 231 passed.
- `corpus_target`'s three-row floor removed: 231 passed.

**Why major, not minor.** This repo already graded this exact shape major once. `tests/test_engine_surface.py:3` records PR #311: the engine was carried into the skeleton, the suite reported "194 passed", the reviewer deleted 15 carried modules and it stayed green. Same move here, same green. The PR body offers "voiceloop suite on the patched tree: 231 passed" as its verification row for the port. And `gate_and_judge.py:106` in this same diff states the rule being broken: "a reason nobody checked is how the next reader inherits a false premise."

**What a human does about it:** port `test_voice_reach.py` into `plugins/kipi-core/voiceloop/tests/`, or delete both citations and say plainly that the guard lives instance-side. The code as written is correct; the guard and the claim are what is missing.

## Finding 2 — minor

**Claim:** `ROTATION_WINDOW_MULT` is dead code, and its own docstring says it is load-bearing.

**File:** `plugins/kipi-core/voiceloop/selector.py:28`

The docstring at lines 22 to 27 says "Raising it trades length discipline for variety; lowering it to 1 restores the defect this fixes." Lowering it to 1 now does nothing.

```
$ grep -rn "ROTATION_WINDOW_MULT" .
plugins/kipi-core/voiceloop/selector.py:28:ROTATION_WINDOW_MULT = 3
plugins/kipi-core/voiceloop/selector.py:224:        # This line read `near[:max(k * ROTATION_WINDOW_MULT, k + 1)]`. That cut the
plugins/kipi-core/voiceloop/selector.py:243:        # Raising ROTATION_WINDOW_MULT instead was rejected: that is moving a
```

Line 28 is the only non-comment reference. Nothing reads it.

## Finding 3 — minor

**Claim:** Every usage example in `voice_ref.py`'s own docstring now exits 2, which is the trap the PR says it removed rather than relocated.

**File:** `plugins/kipi-core/voiceloop/voice_ref.py:16` through `:19`

Ran line 18 verbatim:

```
$ python3 voiceloop/voice_ref.py --channel linkedin --words 200
usage: voice_ref.py [-h] --channel {x,linkedin} --words WORDS [-k K]
                    [--counter COUNTER] --slot-kind {post,comment}
                    [--ids-only]
voice_ref.py: error: the following arguments are required: --slot-kind
exit=2
```

All four examples have the same shape. The loader's copy of the command was correctly deleted; the script's own four copies were left. The paths in those lines (`scripts/voice_ref.py`) are also stale per the file's own comment at line 28.

## Finding 4 — minor

**Claim:** A comment directly above the changed line still describes the deleted behaviour and contradicts the new function's docstring.

**File:** `plugins/kipi-core/voiceloop/validate.py:452`

It reads "target_words: the corpus's own shortest register... None is kept because pre-2026 callers still pass it." The line under it is now `for target in _targets_for(channel, base, targets)`, and `_targets_for`'s own docstring at line 562 says the opposite: "After the sweep neither is reachable on a checked channel."

## What is sound

- **Span bound is real.** `ROTATION_SPAN_FRACTION 0.6 → 100.0` and deleting the span filter entirely both go RED on `test_short_target_never_returns_the_long_row`. The claim that distance, not count, is doing the work is true and tested.
- **New check_budget is not a performance regression.** 106-row corpus, `check_budget` runs in 0.02 seconds. I expected the `len(pool)` counter loop times two slot kinds times two targets to bite. It does not.
- **`reply_format._usable` genuinely matches `form._usable`.** Field for field, including the `retired` clause. The docstring claim holds.
- **The band lint and the migrate script survive the `band(channel)` to `band(channel, path=None)` signature change.** `python3 -m pytest q-system/.q-system/scripts/test/test_voiceloop_band_lint.py q-system/.q-system/scripts/test_voiceloop_migrate.py -q` gives 56 passed.
- **`content_key` exists in the package**, so the new package-level import in `gate_and_judge.py` is not an ImportError fleet-wide. `test_engine_surface.py` names it in the manifest.
- **`test_the_window_is_wider_than_k_but_not_the_whole_pool` can fail in both directions.** Widening to the whole pool makes `max(seen)` reach 400 and the assertion red. It is not decoration; it just does not see the count cap.
- **I could not make the removed count cap produce a worse assembly than the budget allows.** `check_budget` returned no problems on a 106-row corpus with rows up to 400 words.

## VERDICT: REQUEST CHANGES

One major. Merging as-is causes nothing permanent and the code is an improvement over main, so this is not a BLOCK.

**Fix first:** the false citation. Two shipped comments tell the next reader that `test_voice_reach.py` guards the carried fix, in a tree where that file does not exist. Either port the test or correct the sentences. The other three minors are cheap and can ride along.

FINDINGS:
major|The 2026-09-19 count-cap removal ships unguarded: restoring the removed line keeps the suite at 231 passed while corpus reach drops 34 to 12 of 106, and the comment citing test_voice_reach.py as the pin names a file absent from this repo|plugins/kipi-core/voiceloop/selector.py:246
minor|ROTATION_WINDOW_MULT is now referenced only in comments, while its docstring claims lowering it to 1 restores the defect|plugins/kipi-core/voiceloop/selector.py:28
minor|All four usage examples in voice_ref.py's docstring now exit 2 because they carry no --slot-kind|plugins/kipi-core/voiceloop/voice_ref.py:16
minor|Comment above the changed line still says None is kept for pre-2026 callers, contradicting _targets_for's own docstring|plugins/kipi-core/voiceloop/validate.py:452
END FINDINGS

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-386-20260920-003154.md` (7342 bytes).

```
FINDINGS:
major|The check_budget default target enumeration samples the shortest assemblies, so the largest-assembly gate reports CLEAN at 51374 chars against a 24000 budget|plugins/kipi-core/voiceloop/validate.py:335
minor|The new reply_format.corpus_words raises JSONDecodeError on a torn JSONL row that corpus._read_jsonl in the same package deliberately tolerates|plugins/kipi-core/voiceloop/reply_format.py:124
minor|The _targets_for docstring claims the targets declaration is pinned by test_voice_reach.py, which does not exist in this repo|plugins/kipi-core/voiceloop/validate.py:329
nit|The retired-constant tombstone is glued to ROTATION_SPAN_FRACTION's doc comment, labelling a live constant as DELETED|plugins/kipi-core/voiceloop/selector.py:21
END FINDINGS
```

--- reviewer output ---

# Review: PR #386 (voiceloop, two generations into the skeleton)

Read-only. Nothing in the repo was modified; all mutations ran against copies in a temp dir.

---

## Finding 1

**SEVERITY: major**

**Claim:** The `check_budget` gate promises "the largest legal assembly must fit the budget", but its default target enumeration is `(None, corpus-minimum)`, the two targets that draw the *shortest* rows, so it can report CLEAN while a production-shaped assembly is more than twice the budget.

**File:** `plugins/kipi-core/voiceloop/validate.py:335`

The new loop widened the counter axis and the slot-kind axis and calls itself "EVERY AXIS THE PRODUCER VARIES" (validate.py:349). The target axis is the one that decides assembly size, and `length_band` ranks by *distance*, so targeting the corpus minimum selects the shortest rows by construction. The comment at `selector.py:209` records what producers actually pass: 200 on linkedin, 40 on x. Neither is enumerated. The `targets` declaration that would fix it has zero callers anywhere in this repo, so every instance gets the fallback.

**Reproducer:** Built a linkedin corpus of 40 post rows, 10 of them at 2000 words, ids spaced so no 4 consecutive ids are all long.

```
Gate verdict from validate.check_budget: CLEAN (no problems)
Real assembly at target_words=2000: 51374 chars at counter 0
Budget is 24000 so this overruns by 27374
```

The PR body cites this exact incident as motivation: 24,701 chars against a 24,000 budget once three 300-plus-word posts were approved together. This is the gate standing between the fleet and a repeat, and it grades the shapes that cannot overflow.

---

## Finding 2

**SEVERITY: minor**

**Claim:** The new `reply_format.corpus_words` reads the same exemplars file `corpus._read_jsonl` reads, but with a bare `json.loads`, so one torn append row raises `JSONDecodeError` out of `band()` and `length_rule()` while every other reader in the package degrades.

**File:** `plugins/kipi-core/voiceloop/reply_format.py:124`

The loader at `corpus.py:56` names the tolerance out loud ("a torn row loses ITS row, never the file") and `validate.check_all` reports skipped rows, so this input shape is known to occur on a real corpus.

**Reproducer:** One truncated line inserted into a 5-row comment corpus.

```
Loader corpus.load tolerates it, skipped_rows = 1
Call to reply_format.band RAISED: JSONDecodeError Unterminated string starting at: line 1 column 35 (char 34)
Call to reply_format.length_rule RAISED: JSONDecodeError Unterminated string starting at: line 1 column 35 (char 34)
```

Loud rather than silent, and trivially fixed by deleting the line, which is why this is minor and not major. Both `corpus_words` and `corpus_target` ship with no test in this repo.

---

## Finding 3

**SEVERITY: minor**

**Claim:** The `_targets_for` docstring asserts the declaration is pinned by `test_voice_reach.py`, and that file does not exist in this repo, so the mechanism ships to 26 instances unpinned.

**File:** `plugins/kipi-core/voiceloop/validate.py:329`

**Reproducer:** `find . -name "test_voice_reach*"` printed nothing.

This PR's own new test file argues the same point about the same file, at `tests/test_length_axis.py:167`: "It is in the wrong repo: this engine ships to every instance and the consumer's test does not travel with it." The reach guard got moved in-repo; the targets pin did not, and the docstring still claims it.

---

## Finding 4

**SEVERITY: nit**

**Claim:** The `ROTATION_WINDOW_MULT` tombstone is glued to `ROTATION_SPAN_FRACTION`'s doc comment with no blank line, so the live constant is introduced by the words "RETIRED 2026-09-19, and DELETED".

**File:** `plugins/kipi-core/voiceloop/selector.py:21` through `:42`

Lines 21 to 41 are one unbroken `#:` block; line 42 is `ROTATION_SPAN_FRACTION = 0.6`. Under Sphinx `#:` semantics the whole block, retirement notice included, is that constant's documentation. The comment itself warns that a dead constant asserting a behaviour is "a trap for the next reader."

---

## What is sound

**The count-cap regression guard is real.** Restored `window = near[:max(k * ROTATION_WINDOW_MULT, k + 1)]` in a copy of the tree. Then `test_every_row_inside_the_band_is_reachable` went red with "reached 12 of 40", and `test_reach_grows_with_the_corpus` went red with "moved reach 12 to 12". Result: 2 failed, 232 passed. Round 1's stated goal is met by execution, not by claim.

**The distance bound is load-bearing, not decoration.** Replaced `near = [r for r in ranked if abs(...) <= span]` with `near = list(ranked)`. Then `test_short_target_never_returns_the_long_row` and `test_widening_the_window_did_not_reopen_the_length_scar` both went red. The 2026-08-09 scar is genuinely held by `span` and not by the removed count cap.

**The counter range in the new `check_budget` loop covers the full rotation period.** The loop runs over the resolved pool's length, and `select` narrows to a window that is always a subset of it plus one rescued anchor, so the period cannot exceed the range. No off-by-one.

**Performance does not regress meaningfully.** Timed the gate at 53, 106 and 212 rows per channel: 0.02s, 0.06s, 0.24s. Quadratic, irrelevant at this scale.

**The `voice_ref.py --slot-kind` change has no surviving invoker here.** Grepped every `voice_ref.py` reference in the repo. After the loader edit, all remaining hits are prose in comments and docstrings. The loader-plus-engine pairing argument in the PR body holds for this repo.

**The full suite is green on the branch as submitted:** 234 passed in 0.36s.

**Dropped, could not reproduce:** I tried to make the starvation fallback at `selector.py:245` reopen the length scar, since a window shorter than k discards the band and takes the nearest k, which can include far rows. It behaves exactly as its comment says and matches pre-2026-09-10 behaviour, so it is a deliberate documented tradeoff, not a defect.

---

## VERDICT: REQUEST CHANGES

Not BLOCK. Nothing here is permanent or unrecoverable, and the PR is a clear net improvement over main.

**Fix first:** Finding 1. Either enumerate a long target in the `_targets_for` fallback (the target that draws the longest band, since that is where the largest assembly lives), or wire `targets` from something in this repo so the declaration is not dead. As written, the function that exists to bound the largest assembly samples only the smallest ones, and the overflow it is meant to catch already happened once on the live corpus.

FINDINGS:
major|The check_budget default target enumeration samples the shortest assemblies, so the largest-assembly gate reports CLEAN at 51374 chars against a 24000 budget|plugins/kipi-core/voiceloop/validate.py:335
minor|The new reply_format.corpus_words raises JSONDecodeError on a torn JSONL row that corpus._read_jsonl in the same package deliberately tolerates|plugins/kipi-core/voiceloop/reply_format.py:124
minor|The _targets_for docstring claims the targets declaration is pinned by test_voice_reach.py, which does not exist in this repo|plugins/kipi-core/voiceloop/validate.py:329
nit|The retired-constant tombstone is glued to ROTATION_SPAN_FRACTION's doc comment, labelling a live constant as DELETED|plugins/kipi-core/voiceloop/selector.py:21
END FINDINGS

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-386-20260920-005938.md` (8808 bytes).

```
FINDINGS:
major|gate_and_judge passes new recent_openers/path kwargs to injected callables with no version guard, so an instance with pre-PR dependencies raises TypeError on every draft|plugins/kipi-core/voiceloop/gate_and_judge.py:97
minor|reply_format.corpus_words has no torn-row tolerance while corpus.read_jsonl in the same package deliberately skips them, so one torn row raises JSONDecodeError out of length_rule|plugins/kipi-core/voiceloop/reply_format.py:124
minor|validate.py and reply_format.py cite a test name and a module name that do not exist in this package|plugins/kipi-core/voiceloop/validate.py:348
nit|corpus_target docstring says median but the code returns the upper-middle element for even-sized corpora|plugins/kipi-core/voiceloop/reply_format.py:149
END FINDINGS
```

--- reviewer output ---

## Review: PR #386 — carry two generations of the voiceloop engine into the skeleton

Read-only. Nothing in the repo was modified. Baseline: `236 passed` on `plugins/kipi-core/voiceloop/tests`.

---

### FINDING 1 — major

`gate_and_judge` pushes two NEW keyword arguments onto INJECTED callables with no version guard and no test. An instance whose `decide` or writer predates this PR raises TypeError on the first draft.

`plugins/kipi-core/voiceloop/gate_and_judge.py:97` (also `:163`, `:248`)

The module docstring is explicit that `decide`, `revise`, `voicefp_gate` and `_append_voice_provenance` are operator-supplied. This PR changes what the engine demands of them: `decide.decide_candidate(..., recent_openers=...)` at two sites, and `_append_voice_provenance(..., path=...)` at one. `recent_openers=None` and `provenance_path=None` keep the ENGINE's signature backward compatible for its callers, and that is the trap: the old caller still works, the old injected dependency does not. The kwarg is passed unconditionally, so it fires on every call, not on an edge case.

Reproducer, run:

```
python3 - <<'PYEOF'
import sys, types
sys.path.insert(0, ".../plugins/kipi-core")
from voiceloop import gate_and_judge as gj
class Verdict: status="shippable"; reasons=[]; text="a body"
decide = types.SimpleNamespace(SHIPPABLE="shippable",
    decide_candidate=lambda post, regenerate=None, channel=None, source_text=None,
                            prompt_carried=None, handles=False: Verdict())
def _append_voice_provenance(channel, at, row): pass   # pre-PR signature, no path=
gj.gate_and_judge("a body", channel="linkedin", idea_text="idea", voice_prov={},
    arch_id=None, arch_entry=None, runner=None, trail=dict(stages=[]), at="now",
    decide=decide, revise=types.SimpleNamespace(reviser=lambda **kw: None),
    voicefp_gate=types.SimpleNamespace(drift_report=lambda t, authorship=False: {}),
    prompt_carried_for=lambda p: [], _append_voice_provenance=_append_voice_provenance)
PYEOF
```

Real output:

```
TypeError: <lambda>() got an unexpected keyword argument 'recent_openers'
```

`_append_voice_provenance(path=...)` is the same shape one line further down the same path.

Why this matters at 3am: the PR's premise is that instances run OLDER code. "Fleet blast radius, measured not assumed" covers callers of `voice_section` and `for_slot`. It does not cover the injected-callback contract, which is the half that does not propagate from this repo. `kipi update` ships the engine; it does not ship the instance's `decide`. `test_engine_surface.py` only proves `gate_and_judge` imports.

Honest caveat: zero callers of `gate_and_judge` exist in this repo, so my dependency shapes are constructed. I could not check a live instance from here.

---

### FINDING 2 — minor

`reply_format.corpus_words` is a second reader of `exemplars.jsonl` with no torn-row tolerance, while `corpus.read_jsonl` in the same package deliberately survives one.

`plugins/kipi-core/voiceloop/reply_format.py:124`

`corpus.py:57` carries the comment "a torn row loses ITS row, never the file", and `Voice.skipped_rows` exists because torn rows are a shape this corpus really produces. The new reader calls bare `json.loads(line)`. It also opens without `errors="replace"`, which the loader uses.

Reproducer, run on a 4-row comment corpus with one truncated line:

```
corpus.read_jsonl  -> 4 rows, skipped=1  (tolerates the torn row)
reply_format.corpus_words RAISED JSONDecodeError: Unterminated string starting at: line 1 column 47 (char 46)
reply_format.length_rule  RAISED JSONDecodeError: Unterminated string starting at: line 1 column 47 (char 46)
```

`length_rule` is described as "the one sentence the WRITER is given about length", so one torn row takes out reply prompt assembly rather than degrading it. Dropped from major: no caller in this repo passes `path`, so I cannot show a live crash path from here. Related: none of the ~110 new lines in `reply_format.py` have a test.

---

### FINDING 3 — minor

Two citations point at things that do not exist. Round 2 of this PR claimed to fix that class structurally.

`plugins/kipi-core/voiceloop/validate.py:348`

```
grep -rn "test_a_declaration_that_omits_a_channel_is_not_graded" <worktree>/
-> validate.py:348   (the only hit; the test is not defined anywhere)
grep -n "^def test" voiceloop/tests/test_engine_surface.py
-> ... test_an_undeclared_channel_falls_back_rather_than_going_ungraded
```

Same file class at `reply_format.py:112` and `:177`, both citing `reddit_reply_format`:

```
grep -rln "reddit_reply_format" <worktree>/plugins/
-> plugins/kipi-core/voiceloop/reply_format.py   (only the citation; no such module)
```

`EXPECTED_MODULES` confirms `reddit_reply_format` does not ship with this package.

---

### FINDING 4 — nit

`corpus_target` docstring says "His median comment length"; the code returns the upper-middle element.

`plugins/kipi-core/voiceloop/reply_format.py:149`

```
corpus_target -> 30   statistics.median -> 25.0     (on counts [10,20,30,40])
```

---

### What is sound

**The reach guard is real.** Restoring the deleted count cap on disk and running the full suite:

```
=== mutant=none               exit=0 :: 236 passed in 0.48s
=== mutant=restore_count_cap  exit=1 :: 2 failed, 234 passed
    FAILED test_length_axis.py::test_every_row_inside_the_band_is_reachable
    FAILED test_length_axis.py::test_reach_grows_with_the_corpus
```

Round 1's stated problem was that the mutation left 231 tests green. It does not any more.

**The budget-end fix is driven by its test, not decorated by it.** Reverting only the call site, `extreme=max` to `extreme=min`:

```
=== mutant=budget_uses_min exit=1 :: 1 failed, 235 passed
    FAILED test_engine_surface.py::test_check_budget_grades_the_LONGEST_end_not_the_shortest
```

That is the exact mutation the docstring says the first version of the test could not see.

**The empty-target-list hazard is pinned.** `_targets_for` returning `()` for an undeclared channel:

```
=== mutant=targets_empty_for_undeclared exit=1 :: 8 failed, 228 passed
```

**`reply_format._usable` really does match `form._usable`.** I compared both bodies; they are identical predicate for predicate, so the "kept in step on purpose" comment is true rather than aspirational.

**`check_budget`'s new nested enumeration is not a cost problem.** I feared the 12-counter loop becoming `channels x 2 slot_kinds x 2 targets x len(pool)`. On a 237-row synthetic corpus shaped like the described live one (106 linkedin posts, an x 479-word outlier):

```
check_budget (no declaration)  0.04s -> []
check_correction_share         0.04s -> []
check_budget (declared 200/40) 0.02s -> []
```

**I tried and failed to show the corpus-max fallback is a false red.** My hypothesis was that grading an undeclared channel at `corpus_max` cries wolf on a target no producer assembles, contradicting `_targets_for`'s own stated principle. Worst assemblies on my corpus: linkedin 4902 at target 200 versus 5222 at corpus-max, against a 24000 budget. No red, nowhere near one. Dropping the finding.

**The loader edit is live and correct.** The `voice-dna-loader.py` output in this session's own context carries the new "The lane picks the exemplars" text and no `voice_ref.py` command line. `root` is still used at line 159, so the deletion left no dead local.

---

### VERDICT: REQUEST CHANGES

One major finding. Not BLOCK: nothing here is permanent or unrecoverable, and the fix is a signature change plus a test.

**Fix first:** the injected-callable contract at `gate_and_judge.py:97`. Either grep the fleet for instance-side `decide_candidate` and `_append_voice_provenance` definitions and extend the "blast radius, measured not assumed" table to cover them, or make the engine degrade instead of crash. A test with a pre-PR dependency shape would have caught it, and `gate_and_judge` currently has no test beyond "it imports".

FINDINGS:
major|gate_and_judge passes new recent_openers/path kwargs to injected callables with no version guard, so an instance with pre-PR dependencies raises TypeError on every draft|plugins/kipi-core/voiceloop/gate_and_judge.py:97
minor|reply_format.corpus_words has no torn-row tolerance while corpus.read_jsonl in the same package deliberately skips them, so one torn row raises JSONDecodeError out of length_rule|plugins/kipi-core/voiceloop/reply_format.py:124
minor|validate.py and reply_format.py cite a test name and a module name that do not exist in this package|plugins/kipi-core/voiceloop/validate.py:348
nit|corpus_target docstring says median but the code returns the upper-middle element for even-sized corpora|plugins/kipi-core/voiceloop/reply_format.py:149
END FINDINGS

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-386-20260920-084758.md` (7079 bytes).

```
FINDINGS:
major|gate_and_judge passes recent_openers= unconditionally at the recheck call site, so an instance on the older injected decide contract raises TypeError on the style-revision path|plugins/kipi-core/voiceloop/gate_and_judge.py:209
minor|the round-3 regression test's style_review fixture returns a key no consumer reads, so the revision path it guards is unreachable and the suite stays green|plugins/kipi-core/voiceloop/tests/test_engine_surface.py:312
minor|reply_format.corpus_words parses exemplars.jsonl with a bare json.loads while the package's own loader tolerates and counts corrupt rows, so one truncated line takes band() and length_rule() down|plugins/kipi-core/voiceloop/reply_format.py:124
minor|the docstring and PR body claim an instance declaring no targets grades exactly as before, but check_budget now also enumerates every slot kind and len(pool) counters and goes red where the old one was green|plugins/kipi-core/voiceloop/validate.py:318
minor|two docstrings cite reddit_reply_format as the authority for the new code's posture and no such module exists in this package|plugins/kipi-core/voiceloop/reply_format.py:112
END FINDINGS
```

--- reviewer output ---

The round-3 guard covers one of two call sites. Everything below was executed.

## Findings

### 1. major — the injection guard is half-installed
`plugins/kipi-core/voiceloop/gate_and_judge.py:209`

Round 3 added `_accepts` and wired it to the first `decide.decide_candidate` call (line 138). The **recheck** inside the style-revision loop is still written the old way: `recent_openers=recent_openers`, unconditional. An instance whose injected `decide` predates that parameter survives the first pass and then dies with TypeError the moment the style judge returns `hold` and the reviser produces a body, after the reviser has already spent a model call.

The comment at `gate_and_judge.py:134` states the opposite: "`_accepts` is used for every optional kwarg crossing this boundary rather than this one, so the next one added cannot reintroduce the same defect by being written the old way." The defect is 70 lines below that sentence.

Reproducer (injects the older contract verbatim, drives the revision loop):

```
python3 - <<'PYEOF'   # full script in the transcript; fixture = old_decide_candidate
   without recent_openers, style_review -> level "hold", revise -> a body
PYEOF
```

Real output:
```
FAIL decide_calls=1  TypeError: old_decide_candidate() got an unexpected keyword argument 'recent_openers'
```

`decide_calls=1` is the measurement: the guarded site passed, the unguarded one raised.

Today's blast radius is the same as the finding it is fixing. The author measured all 15 `decide.py` copies accept it, so nothing breaks now. The guard exists for the fleet case, and half of it is missing.

### 2. minor — the round-3 regression test uses a `style_review` shape no consumer reads
`plugins/kipi-core/voiceloop/tests/test_engine_surface.py:312`

The fixture returns `{"verdict": "pass"}`. Every consumer in `gate_and_judge.py` reads `level`, `distance`, `flags`, `detail`; `"verdict"` appears exactly once in the whole package, in this fixture. So `review.get("level")` is None, the `while` never runs, and the revision path the defect lives on is unreachable from the test that exists to guard it.

```
plugins/kipi-core: python3 -m pytest voiceloop/tests -q
240 passed in 0.40s
```

240 green with finding 1 live.

### 3. minor — the new corpus reader crashes where every sibling reader degrades
`plugins/kipi-core/voiceloop/reply_format.py:124`

`corpus_words` calls bare `json.loads` on `exemplars.jsonl`. The package's own loader wraps the identical parse in `try/except ValueError` and counts the row (`corpus.py:55-56`), and `validate.check_all` ships a check that reports `"N corrupt JSONL row(s) skipped by the loader"`. A truncated last line is the normal outcome of an interrupted append.

Reproducer, control beside subject, same file:
```
control corpus.load skipped_rows=1 exemplars=3
FAIL JSONDecodeError: Unterminated string starting at: line 1 column 25 (char 24)
```

`band()` and `length_rule()` both go down. `length_rule` is what supplies the writer's length sentence. Severity dropped one level because `reply_format` has no caller in this repo, only downstream.

### 4. minor — the backward-compat claim is false as measured
`plugins/kipi-core/voiceloop/validate.py:318`

The docstring says an instance declaring nothing "is graded exactly as before"; the PR body says "grades byte-identically to before." The fallback keeps the old *targets*, but `check_budget` now also iterates every `SLOT_KIND` and `len(pool)` counters instead of 12, so it grades strictly more shapes.

```
OLD check_budget, targets undeclared -> []
NEW check_budget, targets undeclared -> ['linkedin: largest assembly 27322 chars exceeds budget 24000']
```

Same corpus, `targets=None`. The new red is *correct*, which is why this is minor and not major. What is wrong is the promise that a fleet sync cannot turn an undeclared instance's corpus validation red.

### 5. minor — two citations point at a module that is not in the package
`plugins/kipi-core/voiceloop/reply_format.py:112` and `:177`

Both name `reddit_reply_format` as the authority for the new code's posture. `ls plugins/kipi-core/voiceloop/` lists 27 modules and none is `reddit_reply_format.py`; a repo-wide grep finds no definition. This is the third occurrence of the class round 2 already flagged in `validate._targets_for`.

## What is sound

Attacks the code survived.

**The headline reach fix is real and its guard has teeth.** I restored the count cap in an in-memory copy of `selector.py` and ran the new test's own reach computation against both:
```
clean   reach = 40 of 40
mutated reach = 12 of 40
```
`test_every_row_inside_the_band_is_reachable` goes red on the exact mutation it names. That is not decoration.

**The 2026-08-09 scar does not reopen.** `span` bounds distance independently of count; the thin-corpus and far-row negative controls both hold with the cap gone.

**`check_budget` cost is not a problem.** 18x more work, 12ms absolute on a 106-row corpus. Not worth a finding.

**The `voice_ref` / loader pairing checks out.** `voice_ref.py` has zero callers in this repo, so making `--slot-kind` required and deleting the loader's suggested command really does remove the trap rather than relocate it.

**`_accepts` itself is correct**, including the `**kwargs` and un-inspectable branches. The helper is fine; only one of its two call sites uses it.

One thing to fix in the PR body: it says `kipi-core` 1.12.8 → 1.13.0; the diff says 1.13.3.

## VERDICT: REQUEST CHANGES

One major, so not APPROVE WITH NITS. Not BLOCK: nothing here is permanent or unrecoverable, and today's measured blast radius on finding 1 is zero.

**Fix first:** wire `recent_openers` at `gate_and_judge.py:209` through the same `optional` dict the first call site uses, and give the regression test a `style_review` fixture that emits `level`/`distance`/`flags` so it can actually enter the revision loop.

FINDINGS:
major|gate_and_judge passes recent_openers= unconditionally at the recheck call site, so an instance on the older injected decide contract raises TypeError on the style-revision path|plugins/kipi-core/voiceloop/gate_and_judge.py:209
minor|the round-3 regression test's style_review fixture returns a key no consumer reads, so the revision path it guards is unreachable and the suite stays green|plugins/kipi-core/voiceloop/tests/test_engine_surface.py:312
minor|reply_format.corpus_words parses exemplars.jsonl with a bare json.loads while the package's own loader tolerates and counts corrupt rows, so one truncated line takes band() and length_rule() down|plugins/kipi-core/voiceloop/reply_format.py:124
minor|the docstring and PR body claim an instance declaring no targets grades exactly as before, but check_budget now also enumerates every slot kind and len(pool) counters and goes red where the old one was green|plugins/kipi-core/voiceloop/validate.py:318
minor|two docstrings cite reddit_reply_format as the authority for the new code's posture and no such module exists in this package|plugins/kipi-core/voiceloop/reply_format.py:112
END FINDINGS

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-386-20260920-085920.md` (7207 bytes).

```
FINDINGS:
major|gate_and_judge passes path= to the injected _append_voice_provenance unconditionally, reintroducing the TypeError class rounds 3 and 4 fixed for recent_openers|plugins/kipi-core/voiceloop/gate_and_judge.py:266
minor|the provenance isolation (path=provenance_path) has no test: deleting it leaves all 241 green|plugins/kipi-core/voiceloop/gate_and_judge.py:266
minor|reply_format.corpus_words raises on a torn JSONL row that corpus.read_jsonl in the same package deliberately skips|plugins/kipi-core/voiceloop/reply_format.py:105
minor|the _targets_for docstring cites a test node id that does not exist, the second bad citation in this docstring in two rounds|plugins/kipi-core/voiceloop/validate.py:361
nit|two citations to reddit_reply_format, a module not present in this package|plugins/kipi-core/voiceloop/reply_format.py:112
END FINDINGS
```

--- reviewer output ---

## Review: PR #386 (voiceloop engine parity)

Baseline: 241 tests pass on the head of this branch. All repros run against a copy of the package outside the repo. Nothing in the repo was modified.

---

### FINDING 1 (major): the second injected callable crosses the boundary unguarded

`plugins/kipi-core/voiceloop/gate_and_judge.py:266`

Rounds 3 and 4 of this PR fixed exactly this class for `recent_openers`: an optional kwarg passed to an INJECTED callable that the instance owns and upgrades independently. The same diff adds `path=provenance_path` to `_append_voice_provenance`, another injected callable, with no `_accepts` check.

The module's own `_accepts` docstring at `gate_and_judge.py:56` says the guard is "used for every optional kwarg crossing this boundary rather than this one, so the next one added cannot reintroduce the same defect by being written the old way." That is false in the file the sentence lives in. The round-4 AST test filters on `n.func.attr == "decide_candidate"`, so it cannot see this call.

Reproducer and real output:

```
python3 - <<'PY'   # injects the older (channel, at, row) contract
... gj.gate_and_judge(..., _append_voice_provenance=old_append_voice_provenance, ...)
PY

calling gate_and_judge with an OLD injected _append_voice_provenance ...
  File ".../voiceloop/gate_and_judge.py", line 266, in gate_and_judge
    _append_voice_provenance(channel, at, dict(
        path=provenance_path)
TypeError: old_append_voice_provenance() got an unexpected keyword argument 'path'
EXIT=1
```

It fires on every draft, and it fires after the generation and revision calls are already spent.

**Calibration, stated honestly:** there is no producer of `_append_voice_provenance` in this repo, so I constructed the older signature. I could not census the fleet copies: greps outside the review tree were declined by the permission layer. What keeps this at major rather than minor is that the module's own docstring is the producer specification here, it says these callables are the instance's and upgrade independently, and that exact premise is what made `recent_openers` a major two rounds ago.

**Fix:** route `path=` through the same `**optional` dict, and widen `test_every_decide_call_site_is_guarded` to cover calls to injected names, not one attribute name.

---

### FINDING 2 (minor): the provenance-isolation change is entirely untested

`plugins/kipi-core/voiceloop/gate_and_judge.py:266`

The comment says this stops fixture rows appending to the live 3 MB `voice-provenance.jsonl`, "8 rows per run of one test file". Deleting the argument leaves the suite fully green.

```
MUTANT applied: provenance isolation removed
241 passed in 0.41s
```

So the change that fixed the pollution, and that introduced Finding 1, has no oracle. The repo's own lessons corpus names this shape.

---

### FINDING 3 (minor): the new corpus reader dies where the package's own reader degrades

`plugins/kipi-core/voiceloop/reply_format.py:105` (`corpus_words`)

`corpus.read_jsonl` skips a torn row on purpose, with the comment "a torn row loses ITS row, never the file", and `check_all` reports the count. The new reader calls bare `json.loads` against the same file.

```
corpus.read_jsonl        -> 2 rows, 1 skipped. survives.
reply_format.length_rule -> JSONDecodeError: Unterminated string starting at: line 1 column 27 (char 26)
```

Same file, same package, two contracts. It also drops `errors="replace"`, so a byte-corrupt row raises `UnicodeDecodeError` too.

---

### FINDING 4 (minor): the docstring cites a test node id that does not exist

`plugins/kipi-core/voiceloop/validate.py:361`

It names `tests/test_engine_surface.py::test_a_declaration_that_omits_a_channel_is_not_graded`. The test added in this diff is `test_an_undeclared_channel_falls_back_rather_than_going_ungraded`.

```
ERROR: not found: .../test_engine_surface.py::test_a_declaration_that_omits_a_channel_is_not_graded
(no match in any of [<Module test_engine_surface.py>])
no tests ran in 0.01s
```

Round 2 flagged a bad citation in this same docstring. The structural fix shipped a new bad one in the same paragraph.

---

### FINDING 5 (nit): two citations to a module that is not in this package

`plugins/kipi-core/voiceloop/reply_format.py:112` and `:177` cite `reddit_reply_format.corpus_words` and `reddit_reply_format.build_prompt`. `ls voiceloop/ | grep -i reddit` exits 1.

---

## What is sound

Four attacks the code survived. Named, because the guards in this PR are unusually real.

- **Count-cap restoration.** Put `window = near[:max(k * 3, k + 1)]` back: `test_every_row_inside_the_band_is_reachable` and `test_reach_grows_with_the_corpus` both go red (reached 12 of 40). The reach guard is not decoration.
- **`extreme=max` reversion.** Revert `check_budget` to the floor fallback: `test_check_budget_grades_the_LONGEST_end_not_the_shortest` goes red. That test drives the call site, not the helper, which was the round-2 complaint and it was fixed properly.
- **Removing the decide guard at both sites.** `test_gate_and_judge_survives_an_OLD_injected_decide` and `test_every_decide_call_site_is_guarded` both go red. The AST test genuinely generalizes past the two known sites.
- **`check_budget` cost blowup.** I expected the new slot_kinds times targets times `len(pool)` enumeration to be a 3am problem. Measured on a 424-row corpus: 0.19s, against 0.22s for the sibling check that already enumerated that way. Dropped, no finding.
- **`voice_ref.py --slot-kind` required.** Grepped the whole repo for a remaining invocation without the flag. Zero. The loader half really does travel with the engine half.

The retraction in the PR body about undeclared channels going RED is correct and correctly loud. That is a release note the fleet sync needs to carry.

---

## VERDICT: REQUEST CHANGES

One major. Not BLOCK: the failure is a loud TypeError, nothing is written before it, and the fix is one line plus a widened assertion.

**Fix first:** route `path=provenance_path` through the same `**optional` dict that `recent_openers` now uses, and make `test_every_decide_call_site_is_guarded` enumerate calls to every injected name rather than the single attribute `decide_candidate`. Otherwise the third instance of this class arrives the same way the second one did.

FINDINGS:
major|gate_and_judge passes path= to the injected _append_voice_provenance unconditionally, reintroducing the TypeError class rounds 3 and 4 fixed for recent_openers|plugins/kipi-core/voiceloop/gate_and_judge.py:266
minor|the provenance isolation (path=provenance_path) has no test: deleting it leaves all 241 green|plugins/kipi-core/voiceloop/gate_and_judge.py:266
minor|reply_format.corpus_words raises on a torn JSONL row that corpus.read_jsonl in the same package deliberately skips|plugins/kipi-core/voiceloop/reply_format.py:105
minor|the _targets_for docstring cites a test node id that does not exist, the second bad citation in this docstring in two rounds|plugins/kipi-core/voiceloop/validate.py:361
nit|two citations to reddit_reply_format, a module not present in this package|plugins/kipi-core/voiceloop/reply_format.py:112
END FINDINGS

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-386-20260920-090712.md` (7677 bytes).

```
FINDINGS:
major|The round-5 injection-boundary test hardcodes {recent_openers, path}, so reverting the round-5 fix and adding a fourth optional kwarg both keep 242 tests green despite the docstring claiming otherwise|plugins/kipi-core/voiceloop/tests/test_engine_surface.py:194
minor|reply_format.corpus_words raises JSONDecodeError on a torn JSONL row that the package's own corpus.read_jsonl is built to skip, killing the reply lane instead of degrading it|plugins/kipi-core/voiceloop/reply_format.py:124
minor|band() returns the corpus median unclamped, so length_rule can tell the writer to aim for 250 words while stating that over 200 is refused|plugins/kipi-core/voiceloop/reply_format.py:163
nit|Two new docstrings cite reddit_reply_format, which is not in this package, the same dangling-citation class the PR fixed in round 2|plugins/kipi-core/voiceloop/reply_format.py:112
END FINDINGS
```

--- reviewer output ---

# Review: PR #386 (voiceloop engine parity)

Read-only. Repo tree confirmed unmodified. All mutations ran against temp copies.

Baseline: `python3 -m pytest plugins/kipi-core/voiceloop/tests -q` gives **242 passed in 0.40s**.

---

## Finding 1

**SEVERITY: major**

**Claim:** The round-5 "guard the class, not the instance" test is keyed on a hardcoded kwarg-name set, so reverting the round-5 fix itself, and adding a fourth optional kwarg, both leave the suite fully green while the test's own docstring claims the opposite.

**File:** `plugins/kipi-core/voiceloop/tests/test_engine_surface.py:194`

The line is `risky = [k for k in direct if k in {"recent_openers", "path"}]`. The docstring 40 lines above it says: *"Add a fourth injected dependency and pass it an optional kwarg directly, and this fails without anyone remembering to update a list."* That set IS the list.

Second arm: there is no execution-level test for `_append_voice_provenance` at all. `decide` got one (`test_gate_and_judge_survives_an_OLD_injected_decide`); the round-5 callable did not, so nothing drives it.

**Reproducer:** a temp copy of the package, two mutations, run the suite.

Mutation A, revert the round-5 fix (`prov_optional = {}` plus the `_accepts` check becomes `prov_optional = {"path": provenance_path}`):

```
### mutation: unguard_prov
242 passed in 0.52s
```

Mutation B, a fourth engine-added optional kwarg passed directly to an injected callable (`trace_id=None` added to `gate_and_judge`, then `_append_voice_provenance(channel, at, trace_id=trace_id, prov=dict(`):

```
### mutation: fourth_kwarg
242 passed in 0.48s
```

**Why it matters at 3am:** the PR's merge rationale is that the class is closed. It is not. The fourth instance of this exact TypeError ships with a green suite, and it takes the whole lane down on any instance whose injected copy predates the kwarg. That is the same defect r3, r4 and r5 each paid a round for.

**Fix shape:** derive the risky kwargs from the engine side too. Everything in `run`'s signature that has a default and is not consumed locally is an engine-added optional; assert none of those names appears as a direct `kw.arg` on a call to an injected name. Plus one execution test that injects an `_append_voice_provenance` without `path`.

---

## Finding 2

**SEVERITY: minor**

**Claim:** `reply_format.corpus_words` raises `JSONDecodeError` on a torn JSONL row that this package's own loader is explicitly written to skip, so a half-written append kills the reply lane instead of degrading it.

**File:** `plugins/kipi-core/voiceloop/reply_format.py:124`

`corpus.read_jsonl:56` carries the comment *"a torn row loses ITS row, never the file"*, and `validate.check_all` counts `voice.skipped_rows`, so the package treats torn rows as a real condition. The new reader has no `try/except` and no `errors="replace"`.

**Reproducer:** three good comment rows plus one truncated line, both readers on the same file.

```
corpus.read_jsonl (the engine's own loader): rows=3 skipped=1
reply_format.length_rule RAISED JSONDecodeError: Unterminated string starting at: line 1 column 27 (char 26)
```

Dropped from major: no writer of `exemplars.jsonl` lives in this repo, so I constructed the torn row rather than getting a producer to emit one.

---

## Finding 3

**SEVERITY: minor**

**Claim:** `band()` returns the corpus median unclamped, so `length_rule` can hand the writer a target above its own refusal ceiling, which is the refusal loop ASK-1744 says the function exists to prevent.

**File:** `plugins/kipi-core/voiceloop/reply_format.py:163`

`_usable` applies no band filter, so three banked comments over 200 words flip it.

**Reproducer:** a corpus of 210, 250, 300-word linkedin comments.

```
MIN_WORDS=35 MAX_WORDS=200
band: (35, 200, 250)

the sentence the writer is handed:
  Length: aim for about 250 words. His own approved comments on this channel run from 210 to 300 words. Under 35 words or over 200 words is refused, so say the thing and stop.

a reply written to exactly that target is graded: [{'rule': 'reply-too-long', 'detail': "250 words, above the 200-word ceiling; ..."}]
```

Fix is one line: `min(max(target, MIN_WORDS), MAX_WORDS)`.

---

## Finding 4

**SEVERITY: nit**

**Claim:** Two new docstrings cite `reddit_reply_format`, a module that is not in the shipped package. This is the same class the PR fixed in round 2 for the `test_voice_reach.py` citation.

**File:** `plugins/kipi-core/voiceloop/reply_format.py:112` and `:177`

**Reproducer output:**

```
reddit_reply_format.py present in the shipped package: False
```

---

## What is sound

Attacks I ran that the code survived. These are real mutations, not a reading.

**The selector fix is driven.** Restoring the deleted count cap turns two tests red, with the numbers in the message, not a generic assert:

```
FAILED test_length_axis.py::test_every_row_inside_the_band_is_reachable
FAILED test_length_axis.py::test_reach_grows_with_the_corpus
AssertionError: tripling the corpus inside the band moved reach 12 -> 12.
2 failed, 240 passed
```

**The `extreme=max` fix is driven.** Reverting `check_budget` to the floor fallback turns its test red: `check_budget returned [] on a corpus whose longest register assembles past 24000`. The docstring's account of how the first two fixture attempts were green-but-wrong matches what the fixture now does.

**The second decide call site is guarded, and both AST tests see it.** Unguarding it fails `test_every_decide_call_site_is_guarded` and `test_no_optional_kwarg_reaches_an_INJECTED_callable_unguarded`.

**No suite-time blowup from the widened `check_budget` enumeration.** I expected the slot-kinds by targets by `len(pool)` nesting to be quadratic in practice. Measured on synthetic corpora: 0.01s at 30 rows per pool, 0.05s at 60, 0.20s at 106. Dropped.

**The `--slot-kind` breaking change travels with its only caller.** Grepped the repo for `voice_ref`: the loader hook was the sole invoker and it is in this diff. The two halves genuinely are one change.

**`_accepts` behaves as documented on `**kwargs` and on an uninspectable callee**, and both branches have tests that can fail.

**The retraction in the PR body is honest.** The `check_budget` behaviour change really does make undeclared channels reach strictly more assemblies, and the body says so instead of claiming a no-op.

---

## VERDICT: REQUEST CHANGES

Not BLOCK. Nothing here is unrecoverable, and every shipped code path is correct today.

**Fix first:** Finding 1. The test that certifies the class is closed cannot see the class. Until `risky` is derived rather than listed, and `_append_voice_provenance` has an execution test like the one `decide` got, the next round of this PR is round 6 with a green suite.

FINDINGS:
major|The round-5 injection-boundary test hardcodes {recent_openers, path}, so reverting the round-5 fix and adding a fourth optional kwarg both keep 242 tests green despite the docstring claiming otherwise|plugins/kipi-core/voiceloop/tests/test_engine_surface.py:194
minor|reply_format.corpus_words raises JSONDecodeError on a torn JSONL row that the package's own corpus.read_jsonl is built to skip, killing the reply lane instead of degrading it|plugins/kipi-core/voiceloop/reply_format.py:124
minor|band() returns the corpus median unclamped, so length_rule can tell the writer to aim for 250 words while stating that over 200 is refused|plugins/kipi-core/voiceloop/reply_format.py:163
nit|Two new docstrings cite reddit_reply_format, which is not in this package, the same dangling-citation class the PR fixed in round 2|plugins/kipi-core/voiceloop/reply_format.py:112
END FINDINGS

assafkip:
## Round 7: this is an RCA, not a seventh patch

The registered review threshold is four rounds. Six rounds found three instances of one class, and five structural fixes from two independent sessions were each too narrow. So the question stopped being "why was this kwarg unguarded" and became "why did five fixes aimed at the CLASS each cover only what was already known".

Full RCA below. Commit: `85917b23`.

---

# RCA: six review rounds, one defect class, five structural fixes that each covered only what was already known

**Date:** 2026-09-20
**Trigger:** PR #386 (kipi-system, `fix/voiceloop-engine-parity`) reached six review rounds. The registered threshold is four. Rounds 3, 4, 5 and 6 each reported the same defect class at a new location, and two of those fixes were written explicitly to cover the class rather than the instance.
**Issue:** ASK-1915
**Sessions:** two, working the same branch independently.

## What happened

`plugins/kipi-core/voiceloop/gate_and_judge.py` is fleet-package code that receives
`decide`, `revise`, `voicefp_gate`, `prompt_carried_for` and `_append_voice_provenance`
as injected keyword arguments. Those live in the operator instance, not the package, and
instances upgrade independently. A keyword the engine adds is optional in the engine and
mandatory at the call site, so sending one that the injected callee predates is a
TypeError on a live publishing lane, not a degraded feature.

Six rounds of review on one PR:

| Round | Finding | Fix |
|---|---|---|
| r1 | count-cap removal shipped unguarded | restored the guard |
| r2 | `check_budget` enumerated the shortest target while grading the largest | graded the largest |
| r3 | `recent_openers=` passed unconditionally to `decide.decide_candidate` | `if _accepts(...)` at that site |
| r4 | r3 guarded one of two decide sites; the style-revision site was still unguarded | AST test enumerating decide call sites |
| r5 | `path=` passed unconditionally to `_append_voice_provenance` | `if _accepts(...)` there, plus an AST test deriving injected NAMES from the entry point signature |
| r6 | the r5 test graded against a literal set `{"recent_openers", "path"}` | (held for this analysis) |

r4 and r5 were both written as class fixes. Both were correct about the instance in
front of them and neither could see the next one.

## Surface symptom

An optional keyword argument reaching an injected callable directly, raising
`TypeError: ... got an unexpected keyword argument` on any instance whose injected copy
predates that keyword.

## Surface root cause

Each fix guarded the keyword it had just been shown.

- r3 guarded `recent_openers` at line 138.
- r4 guarded the second decide site, and its AST test asserted two things:
  `"recent_openers" not in named` and `unpacks_optional`. A call site that unpacks
  `**optional` **and also** passes a new keyword directly satisfies both assertions.
- r5 derived the injected names from the entry point's own signature, which was a real
  improvement, and then graded the keywords found there against the literal set
  `{"recent_openers", "path"}`. A set cannot contain a keyword nobody has written yet.

Both class fixes failed open. Anything unlisted was treated as fine.

## Structural root cause

**type: implicit-contract**
**type: missing-test**

Two sentences.

The question each guard needed to answer is "is this keyword optional at the injection
boundary", and that question **cannot be answered from the source text at all**, because
optionality is a property of the other side of the boundary and the other side is not in
the repository; so every static guard reached for a name list, which is the nearest
observable stand-in, and a name list can only ever contain the defects already found.
The same reflex produced the measurement failures in this PR as well: when the real check
is expensive or out of reach, the nearest observable thing gets substituted for it, and
because the substitute was chosen for proximity rather than fidelity it tends to fail in
the direction that looks reassuring.

The sharper formulation, from the session that ran rounds 1 to 6:

> every one of them was a cheap proxy standing in for an expensive real check. grep on a
> file instead of executing the loader. a loop's exit code instead of one invocation.
> bytecode instead of source. Each was the nearest observable thing from where the person
> was working, which is exactly why it got chosen and exactly why it lied.

That covers both failure families here: guards that could not reach what they guarded,
and instruments that reported the wrong thing.

## Contributing factors

**The instruments outnumbered the defect.** Three bad instruments were used across the
six rounds, by both sessions, and every one failed in the reassuring direction:

| Instrument | Substituted for | How it lied |
|---|---|---|
| stale `__pycache__` | executing the mutated source | measured unmutated bytecode, so mutants read as killed |
| a shell loop's exit code | one invocation's own result | reported non-zero for commands that returned 0 |
| a grep for an emit | executing the emitting path | counted a tombstone comment as a live emit |

Two of the three produced a wrong DIAGNOSIS rather than a wrong number. The stale
bytecode reading led to "this needs a boundary redesign, escalate to the founder", which
was carried to him and then withdrawn. Six rounds were spent on this PR and the
instruments cost more of them than the defect did.

**A test that cannot reach the site it guards reads as coverage.** The r4 AST test
enumerated both decide call sites and its docstring claimed a third unguarded site would
fail it. Measured 2026-09-20, that claim was false when written and stayed false through
r5 and r6. Nothing drove the style-revision branch, so an undefended site there was
indistinguishable from a defended one.

**Four more instances of the class were already present and unreported.** Rounds 3
through 6 chased `recent_openers` and `path`. The same audit that found them shows
`authorship` going to `voicefp_gate.drift_report`, and `claude_bin`, `model` and `author`
going to `revise.reviser` and `revise.revise`. The module docstring dates the latter three
to 2026-09-06, so three of the four predate this PR entirely. No review round named them,
because every round graded the names it had been handed.

## Verification

Baseline on `5d99e2d8`, `PYTHONDONTWRITEBYTECODE=1` throughout, because stale bytecode is
one of the three bad instruments above:

```
cd plugins/kipi-core && PYTHONDONTWRITEBYTECODE=1 python3 -m pytest voiceloop/tests/ -q
-> 243 passed
```

Mutation protocol: plant a keyword the guards have never seen (`xnew=1`) at each injected
call site, one at a time, confirm the mutant applied by byte count, run the suite, restore
and confirm the file is byte-identical to the backup.

| Mutant | Call site | Before the fix | After the fix |
|---|---|---|---|
| M1 | `decide.decide_candidate` (first site) | KILLED, 1 failed / 241 | KILLED |
| M2 | `decide.decide_candidate` (style-revision site) | SURVIVED, 242 green | KILLED |
| M3 | `_append_voice_provenance` | SURVIVED (KILLED on 5d99e2d8) | KILLED |
| M4 | `voicefp_gate.drift_report` | SURVIVED, 243 green | KILLED |
| M5 | `revise.reviser` | SURVIVED, 243 green | KILLED |
| M6 | `revise.revise` | SURVIVED, 242 green | KILLED |

Before: 1 of 6 killed on `f45b2725`, 2 of 6 on `5d99e2d8`. After: 6 of 6, each named by
`test_no_engine_added_kwarg_reaches_an_INJECTED_callable_unguarded` with the callee and
the keyword in the failure message.

M2 is the load-bearing one. It is the site r4 was written for, and it survived on
`5d99e2d8` after both class fixes had shipped.

Post-fix suite, and the engine confirmed restored:

```
248 passed in 0.44s
xnew occurrences in engine: 0
diff engine backup -> IDENTICAL TO BACKUP
```

**What is NOT verified, recorded rather than claimed.** `_accepts` returns True for a
callee `inspect` cannot read. That is deliberate: assuming the older contract there would
silently drop a real do-not-repeat list instead of crashing. Every fixture in the new
driver carries a truthful `__signature__`, so `inspect` always succeeds and no drive
exercises the fail-open branch. Measured 2026-09-20 on the operator instance, 7 of 7
injected callees are inspectable, so nothing in production reaches that branch today.
That is a fact about today, not a property of the design, and it inverts the first time an
instance injects a C callable or a signature-less wrapper. A single direct assertion pins
the branch's behaviour; the leave-one-out table does not cover it.

## Fixes shipped

**One chokepoint in the engine.** `_gated(fn, **maybe)` returns only the keywords the
injected callee actually takes. Five `if _accepts(...)` blocks and four unguarded direct
keywords now go through one door, so the suite has one thing to watch and a new keyword
cannot arrive by a route the guard does not cover.

**A leave-one-out driver instead of a static reader**
(`test_engine_surface.py`). It reads every `(injected callee, keyword)` pair out of the
engine's AST, then for each pair RUNS `gate_and_judge` end to end against a callee whose
`__signature__` truthfully lacks that keyword, and lets Python raise. A gated keyword is
skipped and the drive completes. An unguarded one produces the real TypeError, named.

**A reachability assertion**, which is the part the previous class fixes lacked. Each AST
call site must actually execute during the drive, or the test fails. The fixtures return
HOLD then WATCH from `style_review` specifically so the style-revision branch is entered,
because that is the branch r4 could not reach.

**A fail-closed declaration.** `_BASE_CONTRACT` lists what is known to PREDATE the package
extraction, so an unlisted keyword is an offender until someone routes it through `_gated`
or adds it deliberately in a line a reviewer can argue with. The r5 set listed what was
dangerous, so unlisted meant fine. That inversion is the difference.

**A negative self-test on the instrument itself**, asserting the older-contract fixture
really does refuse before any green above is believed.

**The r6 provenance fixture is kept.** The owning session offered to have it deleted as
redundant under the driver. Declining that is binding, not a preference: an instruction or
safeguard may not be dropped on grounds of redundancy, because "redundant" is a prediction
about behaviour dressed as a fact about text and it is the one dismissal that never gets
tested. Both mechanisms stay and M3 is killed by both.

## Action items

- [x] Route every optional keyword crossing the injection boundary through one `_gated` chokepoint
- [x] Replace the name-list guard with a leave-one-out driver that executes the lane
- [x] Assert every AST call site is actually reached by the driver
- [x] Guard the four pre-existing instances (`authorship`, `claude_bin`, `model`, `author`)
- [x] Mutation-verify all six call sites, by name, before and after
- [x] Record the fail-open branch as unmeasured rather than covered
- [x] Keep the r6 provenance fixture rather than deleting it as redundant
- [ ] Carry the `_gated` chokepoint to the operator instance's `cycle._gate_and_judge` binder so the two sides cannot drift (follow-up issue, not this PR)
- [ ] Add `PYTHONDONTWRITEBYTECODE=1` to the documented mutation protocol so stale bytecode cannot produce a false kill again

## Lessons

**A guard that recognises a defect by name can only ever hold the defects already found.**
If the property you need cannot be read from the source, no amount of cleverness in the
static reader will fix it. Drive the thing and let the runtime raise.

**Name the substitution out loud.** Every bad instrument here was the nearest observable
thing to where the work was happening. The question that catches it is not "is this check
correct" but "what expensive check is this standing in for, and which direction does it
fail when it is wrong". All three failed toward reassurance, which is the direction that
does not prompt a second look.

**Reachability is half of any guard.** A check applied at a site the tests never execute
is indistinguishable from a check that works. Assert the site ran.

**Round count is a signal about the guard, not the defect.** Four rounds on one class was
the point to stop patching and ask why the class fixes were not generalising. Two more
rounds were spent before that question got asked.

assafkip:
## RCA addendum: independent confirmation, and a fifth bad instrument

**Independent re-measurement of `85917b23`**, run by the session that owned rounds 1
to 6, against its own archive extraction of the pushed ref rather than my worktree.
Baseline 248, `PYTHONDONTWRITEBYTECODE=1`, every mutation confirmed applied by an
anchor assert before the run:

| Mutant | Result |
|---|---|
| M2 `decide.decide_candidate` (style-revision site) | 1 failed / 247, KILLED |
| M3 `_append_voice_provenance` | 1 failed / 247, KILLED |
| M4 `voicefp_gate.drift_report` | 1 failed / 247, KILLED |
| M5 `revise.reviser` | 1 failed / 247, KILLED |
| M6 `revise.revise` | 1 failed / 247, KILLED |

Five for five, each named by
`test_no_engine_added_kwarg_reaches_an_INJECTED_callable_unguarded`. Two sessions,
two independent harnesses, two extractions, same result.

### The fifth bad instrument, and it is the most useful one

It appeared *during the verification of this RCA's own fix*. The first M4
re-measurement anchored on the string `drift_report(verdict.text`. That string no
longer exists, because the fix spans that call across two lines through `_gated`.
The mutation silently did not apply. The run printed a clean **248 green**, which
reads exactly like a defended site.

Same class as the stale `__pycache__` one layer over: the bytecode measured the
unmutated MODULE, the dead anchor measured the unmutated FILE. Both print the
reassuring number.

**The rule, which is the portable output of this PR:**

> A mutation result is TWO claims and only one of them usually gets checked.
> "The mutant was killed" is meaningless until "the mutant was applied" is proven.
> An unapplied mutant and a well-defended one are identical bytes on the terminal.
> Assert the anchor matches **exactly once** before writing, assert the byte count
> moved, and treat an anchor miss as a **failed experiment** that exits non-zero,
> never as a passing test.

Both sessions arrived at a version of this independently and neither held it as a
discipline beforehand. It is now an action item on the RCA.

### Risk read for the merge

`_gated` is a **no-op on the live instance today**. The operator instance's real
callees were inspected directly:

```
decide.decide_candidate(recent_openers=)     PASS-THROUGH
revise.reviser(claude_bin=/model=/author=)   PASS-THROUGH x3
revise.revise(claude_bin=/model=/author=)    PASS-THROUGH x3
```

Seven of seven pass through, so the publishing lane behaves identically to before
this commit. This is a safety net for an instance that has not upgraded yet, not a
change to the one that works. That is the shape something reaching 26 instances
should have.

### Local gate reds, attributed

`verify.sh --full` on the driver branch: **FAILED (2/13)**. Both attributed and
neither held:

- `pytest:plugins/kipi-dsse` — this PR touches **zero** kipi-dsse files and that
  tree is byte-identical to `origin/main`. Local Python is 3.14.6; CI is 3.12. CI
  `verify` is green on the same sha.
- `pytest:test_destructive_op_deny_anchor.py::test_the_vendored_copy_has_not_drifted`
  — compares against a machine-local vendored copy. Not in the diff. Green in CI.

A local `capability-gate.py` run came back RED (12), all `declared_inert` scripts
citing resolved spillover items. `.prd-os/spillover.jsonl` is gitignored and
untracked, so CI reads a different ledger; CI `validate` is green. None of the 12
names a file in the diff. **The gate's input is part of the gate.**

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-386-20260920-100059.md` (6048 bytes).

```
FINDINGS:
minor|_gated silently drops provenance_path on an older injected writer, so a caller asking for isolation writes the production provenance corpus with no signal|plugins/kipi-core/voiceloop/gate_and_judge.py:294
minor|reply_format.corpus_words raises JSONDecodeError on a torn JSONL row that corpus.read_jsonl on the same file tolerates and counts|plugins/kipi-core/voiceloop/reply_format.py:124
minor|length_rule's corpus range excludes dm and channel-any rows the selector does show, so the stated range describes a different population than the exemplars in the prompt|plugins/kipi-core/voiceloop/reply_format.py:125
END FINDINGS
```

--- reviewer output ---

# Review: PR #386 (voiceloop engine parity)

Read-only. Nothing in the repo was modified. All mutations ran against a copy in the system temp dir.

## Finding 1, minor: an isolation kwarg is dropped silently, and the row lands in the production corpus

`plugins/kipi-core/voiceloop/gate_and_judge.py:294`

`_gated` keeps a kwarg only if the injected callee's signature takes it. For `recent_openers` that degrades a feature. For `path` it redirects a WRITE: a caller that explicitly asks for `provenance_path=` and hits an instance whose `_append_voice_provenance` predates `path` gets the production sidecar, with no note in `trail`, no warning, no return signal. That is the exact contamination the comment 18 lines below says this change stops.

Reproducer and real output:

```
=== B: an older injected provenance writer silently loses the isolation path ===
caller asked for isolation at: /var/folders/.../pr386-iso-20_70r86/voice-provenance.jsonl
isolated file exists      : False
PRODUCTION file rows      : 1
trail keys                : ['stages', 'style']
any degraded/skip note?   : [{'stage': 'gates', 'status': 'SHIPPABLE', 'reasons': []}]
```

Discount stated honestly: the callee lives instance-side, so I constructed it. The author's own round-5 comment asserts instances on that older contract exist, which is why the guard was added at all. Not a regression against main (main passes no path), so it is minor rather than major.

## Finding 2, minor: one torn JSONL row kills the reply lane on a file the loader tolerates

`plugins/kipi-core/voiceloop/reply_format.py:124`

`corpus.read_jsonl` skips a torn row and counts it ("a torn row loses ITS row, never the file"), and `validate` surfaces `skipped_rows`. The new `corpus_words` reads the same file with a bare `json.loads` and raises.

Reproducer and real output:

```
=== A: torn JSONL row ===
corpus.read_jsonl tolerates it: rows= 3 skipped= 1
reply_format.corpus_words RAISED: JSONDecodeError Expecting ',' delimiter: line 1 column 68 (char 67)
reply_format.length_rule RAISED: JSONDecodeError Expecting ',' delimiter: line 1 column 68 (char 67)
```

Bounded and loud. `form.corpus_posts:108` already has the same shape, so this is a new instance of an existing pattern, not a new class.

## Finding 3, minor: the length sentence describes a different population than the exemplars beside it

`plugins/kipi-core/voiceloop/reply_format.py:125`

`_usable`'s docstring says these are "the same rows the WRITER is shown" and warns against a number about a different population. The filter excludes `kind: "dm"` and rows whose channel is absent or `"any"`, both of which the selector does show on a comment slot (`ELIGIBLE_KINDS["comment"]`, `_channel_ok`).

Reproducer and real output:

```
exemplars the WRITER is shown for a linkedin comment slot: [('a1', 40), ('a2', 45), ('a3', 50), ('d1', 300)]
reply_format.corpus_words counts       : [40, 45, 50]
length_rule sentence                   : Length: aim for about 45 words. His own approved comments on this channel run from 40 to 50 words. Under 35 words or over 200 words is refused, so say the thing and stop.
```

The prompt carries a 300-word example while telling the writer the range is 40 to 50. I built the `dm` and channel-`any` rows by hand; both shapes are named in the engine's own schema (`EXEMPLAR_CHANNELS`, `ELIGIBLE_KINDS`), and no live corpus ships in this repo to confirm frequency.

## What is sound

- **Count-cap removal has a guard that can fail.** Reintroduced `near[:max(k*3, k+1)]` in `selector.select`: `2 failed, 246 passed`. Not decoration.
- **The budget ceiling is driven, not asserted.** Flipped `extreme=max` back to `extreme=min` at the `check_budget` call site: `1 failed, 247 passed`. The round-2 fix is pinned at the call site, not just in the helper.
- **The round-7 boundary driver really covers the class.** Planted a brand-new kwarg `xnew=1` at all six injected call sites (both decide sites, `drift_report`, the provenance writer, `revise.revise`, `revise.reviser`). Every one went red (1 to 2 failures each). This is the thing rounds 4 and 5 could not do, and it holds.
- **No 3am cost from the widened enumeration.** 180-row corpus, 2 assembled channels: `check_budget` 0.01s, `check_correction_share` 0.01s.
- **The breaking signature has no stranded in-repo caller.** Every `voice_section` call in the repo passes `target_words`; `voice_ref.py:86` passes `args.words` with `--words` required.
- **The removed loader command is gone from the copy that actually runs.** This session's own `voice-dna-loader` hook output carries the new "The lane picks the exemplars" text and no command line. Grep across `q-system/`, `.claude/` and the skills tree finds no other file still instructing the old invocation.
- **Suite:** `248 passed in 0.40s`.

## VERDICT: APPROVE WITH NITS

Only minor findings survived reproduction. Two attacks I expected to land (decorative guard on the cap removal, silent blind spot in the new AST/driver test) both failed, and the leave-one-out driver is the right shape for the defect class this PR kept reopening.

First thing to fix: finding 1. A dropped `path` is not a degrade, it is a write to the wrong file with no signal. One branch, either raise or record a `degraded` note on `trail`, separates "this kwarg is optional" from "this isolation control failed open."

FINDINGS:
minor|_gated silently drops provenance_path on an older injected writer, so a caller asking for isolation writes the production provenance corpus with no signal|plugins/kipi-core/voiceloop/gate_and_judge.py:294
minor|reply_format.corpus_words raises JSONDecodeError on a torn JSONL row that corpus.read_jsonl on the same file tolerates and counts|plugins/kipi-core/voiceloop/reply_format.py:124
minor|length_rule's corpus range excludes dm and channel-any rows the selector does show, so the stated range describes a different population than the exemplars in the prompt|plugins/kipi-core/voiceloop/reply_format.py:125
END FINDINGS

PR #387, opened 2026-09-20
fix(commands): /q-debrief ships a file in a path the fleet actually syncs (ASK-1925)

assafkip:
Closes ASK-1925. Resolves spillover `sp-4daf4890`.

## What was broken

`/q-debrief` is the only documented writer of `memory/graph.jsonl` (debrief-template.md steps 6 and 10) and had no command file anywhere in the skeleton.

Its file was deleted 2026-03-23 (commit `69bf969d`) from `.claude/commands/`, which `kipi-update.sh` does not sync. The updater ships only `.claude/agents`, `.claude/output-styles`, `.claude/rules` and `plugins/`. So the deletion could never propagate: instances kept stale pre-March copies for six months, new instances got nothing, and `CLAUDE.md` went on advertising the command. A missing slash command does not error. It is simply absent from the menu.

## The fix

The command now lives at `plugins/kipi-core/commands/q-debrief.md`, the only fleet-wide path a command ships **and** loads through.

Load-path proof: `kipi-core:say` and `kipi-core:wiring-check` load from that directory in a live session, while `plugins/kipi-core/skills/research-mode/commands/q-research.md` does not. Claude Code discovers a plugin's `commands/` at the plugin root only. The check asserts that exact depth, so the dead nested file cannot grade as wired.

The command file reads the canonical template rather than restating it, so the two cannot drift.

## The check

`q-system/.q-system/scripts/test/test_q_debrief_command_wired.sh`: every command named on the CLAUDE.md menu must have a file at `plugins/<plugin>/commands/<name>.md`.

Three things keep it honest:
- the menu is **parsed** from CLAUDE.md, never copied into the test
- a zero-command parse is a failure, not a pass, so a format change cannot silently turn the check into a no-op
- each grandfathered name is asserted to be **still missing**, so the exemption list goes red the day one of them ships a file

Registered in the capability manifest (`expected_tests` + `skeleton_only`).

## Grandfathering, stated plainly

Ten of the thirteen menu names have no command file. Sixty-one percent of a population red on day one gets the gate switched off, and a gate that is off protects nothing. Each exemption is named individually in the script with what it actually is today.

They are captured, not forgiven: **ASK-1927** (promoted from `sp-cda6b4a4`) owns emptying that list and extending the check to the public README, which advertises six of the same names.

## Evidence

Reproducer written before the fix, run against the real menu:

```
$ bash q-system/.q-system/scripts/test/test_q_debrief_command_wired.sh
  /q-debrief is on the CLAUDE.md menu with no file at plugins/*/commands/q-debrief.md
FAIL: CLAUDE.md names commands that ship no file in a path kipi-update.sh syncs
EXIT=1
```

After the command file landed:

```
$ bash q-system/.q-system/scripts/test/test_q_debrief_command_wired.sh
PASS: every command on the CLAUDE.md menu ships a file in a synced path
EXIT=0
```

The script also carries its own negative self-test, an invented command name plus an empty menu, so it cannot quietly become decoration.

All pre-commit gates green: verify.sh (python/shell/json syntax), covers-lint, enforced-claim-lint, instruction-budget ratchet, gitleaks, settings-template-sync.


assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-387-20260920-070945.md` (6979 bytes).

```
FINDINGS:
minor|extract_commands anchors on ^- `/name` so a command advertised inline on the CLAUDE.md menu (the shape /improve already uses) escapes the gate entirely|q-system/.q-system/scripts/test/test_q_debrief_command_wired.sh:73
minor|the `improve` entry in GRANDFATHERED is never extracted, so its exemption suppresses nothing and its rot-check fires a message whose instructed fix changes no behavior|q-system/.q-system/scripts/test/test_q_debrief_command_wired.sh:66
END FINDINGS
```

--- reviewer output ---

# Review: PR #387

Two minor findings survived reproduction. Nothing gates the merge.

---

## Finding 1 (minor): the menu parser only sees line-anchored entries, so an inline-advertised command escapes the gate

**Claim.** `extract_commands` anchors on `^- \`/name\``, so a command advertised mid-line on the CLAUDE.md menu is invisible to the check. The live CLAUDE.md already uses that shape: `/improve` is listed with a leading slash inside the `/q-draft` bullet, and the extractor never yields it.

**File:** `q-system/.q-system/scripts/test/test_q_debrief_command_wired.sh:73`

**Reproducer.** Synthetic tree, real script, four runs including the mutant:

```
$ python3 - <<'PY'   # builds $TMPDIR tree, copies the real script, runs it
...
PY
### CONTROL: all three command files present
PASS: every command on the CLAUDE.md menu ships a file in a synced path
EXIT=0

### MUTANT 1: q-debrief command file gone entirely
/q-debrief is on the CLAUDE.md menu with no file at plugins/*/commands/q-debrief.md
FAIL: CLAUDE.md names commands that ship no file in a path kipi-update.sh syncs
EXIT=1

### MUTANT 2: file only in .claude/commands (the 2026-03-23 shape)
/q-debrief is on the CLAUDE.md menu with no file at plugins/*/commands/q-debrief.md
FAIL: CLAUDE.md names commands that ship no file in a path kipi-update.sh syncs
EXIT=1

### MUTANT 3: menu entry written INLINE (the /improve shape), no file anywhere
PASS: every command on the CLAUDE.md menu ships a file in a synced path
EXIT=0
```

Mutants 1 and 2 prove the check is real. Mutant 3 is the gap: `/never-shipped` on the menu with no file anywhere passes green.

**Also on the real file.** The live extractor yields 12 names, none of them `improve`:

```
$ grep -oE '^- `/[a-z0-9-]+`' CLAUDE.md | sed -e 's/^- `\///' -e 's/`$//'
q-morning q-debrief q-calibrate q-create q-plan q-engage q-draft q-wrap q-handoff q-research wiring-check say
```

So the header's "10 of the 13 names on the CLAUDE.md menu" at line 30 counts a name the parser cannot reach.

---

## Finding 2 (minor): the `improve` exemption is unreachable, and its rot-check fires a message that fixes nothing

**Claim.** `improve` sits in `GRANDFATHERED` but is never extracted, so its exemption never suppresses anything. The rot-check at part (c) still asserts it is missing, so the day `improve` ships a command file the test goes red and instructs deleting an exemption that was already dead.

**File:** `q-system/.q-system/scripts/test/test_q_debrief_command_wired.sh:66`

**Reproducer.** Same synthetic tree, with an `improve.md` command file and `/improve` on the menu inline exactly as CLAUDE.md writes it today:

```
### improve now HAS a command file, /improve is on the menu inline
FAIL: /improve is grandfathered but now HAS a file at plugins/kipi-core/commands/improve.md -- delete its exemption
EXIT=1

### after deleting the improve exemption as the message instructs
PASS: every command on the CLAUDE.md menu ships a file in a synced path
EXIT=0

extractor output (note: no `improve`):
q-debrief
wiring-check
say
q-draft
```

The red, the instructed fix, and the green all happened without the gate ever checking `improve`.

---

## What is sound

**The check is not decoration.** Mutants 1 and 2 above go red. Mutant 2 is the exact 2026-03-23 regression shape (file present, wrong directory) and it fails.

**CI actually runs it, even on a narrow diff.** `_SCAN_RE` in `change-size.py` matches the script's `for f in "$ROOT"/plugins/*/commands/...` line, so the test lands on the always-run scanner floor. Confirmed by calling the real selector:

```
scans_the_tree: True
selected_tests 85 ['q-system/.q-system/scripts/test/test_q_debrief_command_wired.sh']
```

That is for a diff touching only `plugins/kipi-core/commands/q-debrief.md`, which the script never names on an executing line. It is selected anyway. My "the gate never runs on the diff that breaks it" hypothesis is wrong.

**The command file's bash is real, not aspirational.** `loop-tracker.py open` takes argv 2 through 4 as `loop_type, target, context`, matching the three arguments the command passes. `ripple-verify.py` takes changelog plus date, matching. Both ran:

```
$ python3 q-system/.q-system/scripts/ripple-verify.py q-system/canonical/changelog.md 2026-09-20
No changelog entries for 2026-09-20. Nothing to verify.
EXIT=0
```

**The "step 6 writes graph.jsonl" claim checks out.** `debrief-template.md:266` is `### 6. Graph knowledge base -> memory/graph.jsonl`. The command file does not restate the template, so drift has no surface.

**The QROOT heuristic covers both instance shapes.** I expected `QROOT="q-system"; [ -d ] || QROOT="."` to break on instances rooted elsewhere. `instance-registry.json` says 24 of 25 instances use `subtree_prefix: q-system` and one is null, where kipi-update.sh writes to the instance root. The fallback to `.` is correct for exactly that case. `instance_q_dir` (q-consult, q-ktlyst) is a different axis and does not affect this.

**The depth assertion holds where a find-anywhere check would not.** `plugins/kipi-core/skills/research-mode/commands/q-research.md` exists on disk and `command_file_for` does not match it, so the dead nested file does not grade as wired.

**Plugin-dir scope is consistent.** I tried to find a plugins/ subdir that `command_file_for` accepts but the updater skips. All six dirs under `plugins/` are named in `.claude-plugin/marketplace.json`, and `managed_plugin_names`' `*/` glob and the check's `plugins/*/` glob agree on dot-dirs and dangling symlinks. No divergence today.

**Two I could not reproduce and am dropping.** First, whether the instance-local `plugins/` copy actually loads versus the marketplace clone (the 2026-06-20 scar in wiring-check.md). The sandbox blocks reading `~/.claude/plugins`, so I have no executed evidence either way. Second, whether the menu's bare `/q-debrief` resolves against a namespaced `kipi-core:q-debrief`. No way to query the runtime command registry from here. Both stay unstated rather than asserted.

---

## VERDICT: APPROVE WITH NITS

Only minor findings survived. The gate is real, it runs in CI on the diff that would break it, and the mutants prove it.

**Fix first:** the parser at line 73. One extra pattern for a slash-prefixed name anywhere on a `- ` menu line closes both findings at once, since `improve` becomes reachable and its exemption stops being dead. That also makes the line-30 count true.

FINDINGS:
minor|extract_commands anchors on ^- `/name` so a command advertised inline on the CLAUDE.md menu (the shape /improve already uses) escapes the gate entirely|q-system/.q-system/scripts/test/test_q_debrief_command_wired.sh:73
minor|the `improve` entry in GRANDFATHERED is never extracted, so its exemption suppresses nothing and its rot-check fires a message whose instructed fix changes no behavior|q-system/.q-system/scripts/test/test_q_debrief_command_wired.sh:66
END FINDINGS

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-387-20260923-053237.md` (7569 bytes).

```
FINDINGS:
minor|README.md claims every /q-* name now ships a command file; about eighteen names in commands.md still have none|README.md:345
minor|The menu extractor skips namespaced and glob rows, so a README row naming a nonexistent plugin command grades green|q-system/.q-system/scripts/test/test_q_debrief_command_wired.sh:88
minor|/q-debrief's allowed-tools omits every Notion tool while its own Done criteria require two Notion Actions DB checkboxes|plugins/kipi-core/commands/q-debrief.md:5
nit|"that same file" at q-calibrate.md line 15 has no antecedent; no file is named above it|plugins/kipi-core/commands/q-calibrate.md:15
nit|Emdashes in the new CLAUDE.md and q-debrief.md text, against the standing output rule|CLAUDE.md:20
END FINDINGS
```

--- reviewer output ---

Read-only review, round 2. I formed my own read from the code first. I could not read the earlier round's comments: the session's token-guard capped tool calls at 50 before I got there. Every finding below is anchored in the round-2 additions (the ten new command files, the README table shape in the extractor, the rewritten README paragraph), so re-litigation risk is low but I cannot claim zero.

## Findings

**1. minor. README.md:345 claims every `/q-*` name now ships a command file. False for about eighteen of them, in a public repo.**

The new sentence: "The `/q-*` names used to be conventions... As of ASK-1927 each one ships a file at `plugins/kipi-core/commands/<name>.md`, so they are registered commands and appear in the slash menu like any other."

`commands.md` presents 28 `/q-*` names as commands. Ten got files.

Reproducer and real output:

```
$ python3 -c "import pathlib,re; \
print('q-market-*.md on disk:', [str(x) for x in pathlib.Path('.').glob('plugins/*/commands/q-market-*.md')]); \
print('commands.md advertises:', sorted(set(re.findall(r'/q-market-[a-z]+', pathlib.Path('q-system/.q-system/commands.md').read_text()))))"

q-market-*.md on disk: []
commands.md advertises: ['/q-market-assets', '/q-market-create', '/q-market-plan',
                         '/q-market-publish', '/q-market-review', '/q-market-status']
```

Same for `/q-begin`, `/q-status`, `/q-checkpoint`, `/q-end`, `/q-sync-notion`, `/q-content-intel`, `/q-investor-update`, `/q-reality-check`, `/q-ingest-feedback`. The old sentence was true. The new one is the thing this PR exists to stop.

**2. minor. The new check passes over two row shapes, so a menu row naming a command that does not exist grades green.**

`test_q_debrief_command_wired.sh:88` character class excludes the colon and the asterisk. The header calls both skips deliberate. The consequence is that README's six `/kipi-core:*` rows are never checked, and the `/q-market-*` family is never checked.

Reproducer, running the script's own two greps against a mutated copy:

```
extracted: ['q-debrief', 'q-draft', 'q-engage', 'q-handoff', 'q-morning', 'q-research', 'q-wrap']
ghost extracted? False
q-market extracted? False
```

`| `/kipi-core:q-ghost` | a plugin command that does not exist |` is invisible. Today all six real namespaced rows resolve, so this is latent, not live.

**3. minor. `/q-debrief` cannot satisfy its own Done criteria: `allowed-tools` has no Notion tool.**

`plugins/kipi-core/commands/q-debrief.md:5` allows `Bash, Read, Write, Edit, Glob, Grep`. The file's closing line: "Every checkbox in the template's Quality Checklist is ticked or explicitly marked N/A with a reason." Two of those checkboxes need Notion.

```
$ grep -niE "notion" q-system/methodology/debrief-template.md
219:- Add to Notion Actions DB with Energy: People, Time: 15 min
319:- [ ] Any "this week" implications are added to Notion Actions DB
325:  - [ ] Follow-up Action created in Notion Actions DB (...)

$ grep -H "allowed-tools" plugins/kipi-core/commands/linear-drain.md plugins/kipi-core/commands/q-debrief.md
linear-drain.md:allowed-tools: Bash, Read, mcp__linear__list_projects, mcp__linear__list_issues, mcp__linear__save_project, mcp__linear__save_issue
q-debrief.md:allowed-tools: Bash, Read, Write, Edit, Glob, Grep
```

`linear-drain.md` in the same directory enumerates its MCP tools, so this repo treats `allowed-tools` as the operative allowlist. I could not execute the runtime denial, so I am not rating this above minor. The auto-detection path, a pasted transcript, carries no such list and is strictly stronger than the manual command it is the manual trigger for.

**4. nit. `q-calibrate.md:15` has a dangling reference.** "Read the **Post-Edit Ripple Check** section of that same file too." Lines 1 through 14 name no file. The antecedent arrives 22 lines later.

**5. nit. Emdashes in `CLAUDE.md:20` and two lines of `q-debrief.md`,** against the standing output rule. `git diff c9dfd652 HEAD -- CLAUDE.md README.md plugins/kipi-core/commands/ | grep "^+" | grep "—"` returns three hits.

## What is sound

- **The gate runs on the input it guards.** I executed `change-size.py`'s `plan()` on three synthetic diffs. `scans_the_tree` is True and there is no `covers` declaration, so the test sits on the always-run floor: `[('CLAUDE.md', 3)] target selected: True`, same for README.md and for an unrelated file. A future PR that adds a bare name to CLAUDE.md does run this check.
- **Duplicate Linear filing from an on-demand `/q-morning`.** I expected the new `python3 morning-brief.py` line, no `--dry-run`, to file a second permanent ticket for Sana on top of the 07:40 job. It does not: `alert-to-linear.py:891` fingerprints on the normalized message shape and `_file_alert_serialized` gates creation on actually holding the fingerprint lock. Dropped.
- **The 09:00 deadman being masked** by a manual run writing today's receipt. Mechanism is real, but `morning-brief-deadman.py:54` already documents receipt-versus-output as a known gap, and CLAUDE.md's "the command just runs it early" line is unchanged by this diff. Pre-existing, dropped.
- **Dangling executables.** Every script and section the ten command files point at exists: `morning-brief.py`, `ripple-verify.py`, `loop-tracker.py`, `commands.md`, `debrief-template.md`, plus `Post-Edit Ripple Check` at :44, `Evening wrap (/q-wrap)` at :766, `Session handoff (/q-handoff)` at :809.
- **The three negative self-tests all fire,** and the script is green: `PASS: every command on the CLAUDE.md and README.md menus ships a file in a synced path`, exit 0. The empty grandfather list makes the rot loop a no-op, and the header says so rather than hiding it. `ROOT` resolves four levels, correctly.

One thing I wanted to test and could not: whether the founder typing `/q-debrief` resolves, or whether the loaded name is `/kipi-core:q-debrief`. This session's own command list shows kipi-core commands namespaced, and README itself writes `/kipi-core:say` for the existing ones while CLAUDE.md writes `/say` bare. The sandbox blocks reading outside the worktree, so I could not check the loaded copy. If bare resolution does not happen, the menu still promises a name nobody can type, which is this PR's own bug class. Worth one measured session before merge.

Separate from the code: the PR is `CONFLICTING` with main, 28 commits behind, with auto-merge enabled. The rebase touches CLAUDE.md and README.md, the two files the gate parses.

## VERDICT: APPROVE WITH NITS

Fix first: the README sentence at line 345. Scope it to the ten names that shipped, or say plainly which `/q-*` names are still conventions. It is a public page, and an overclaim there is the same failure mode as the six months of stale `/q-debrief`.

FINDINGS:
minor|README.md claims every /q-* name now ships a command file; about eighteen names in commands.md still have none|README.md:345
minor|The menu extractor skips namespaced and glob rows, so a README row naming a nonexistent plugin command grades green|q-system/.q-system/scripts/test/test_q_debrief_command_wired.sh:88
minor|/q-debrief's allowed-tools omits every Notion tool while its own Done criteria require two Notion Actions DB checkboxes|plugins/kipi-core/commands/q-debrief.md:5
nit|"that same file" at q-calibrate.md line 15 has no antecedent; no file is named above it|plugins/kipi-core/commands/q-calibrate.md:15
nit|Emdashes in the new CLAUDE.md and q-debrief.md text, against the standing output rule|CLAUDE.md:20
END FINDINGS

PR #388, opened 2026-09-20
fix(commands): every name on a command menu ships a file behind it (ASK-1927)

assafkip:
Closes the grandfather list ASK-1925 opened. **Stacked on #387** (base is `sana/ask-1925`, not `main`), because the test this edits is created there.

## Per-name decision

**RESTORED (9)** — one file each at `plugins/kipi-core/commands/<name>.md`:
`q-morning`, `q-calibrate`, `q-create`, `q-plan`, `q-engage`, `q-draft`, `q-wrap`, `q-handoff`, `q-research`.

**RETIRED (1)** — `improve`. Its exemption is deleted, not honoured. It was never load-bearing: `extract_commands` matches a name at the START of a list item and `/improve` sits mid-line inside the `/q-draft` row, so it was never extracted. It is the kipi-core `improve` SKILL; `CLAUDE.md` stops writing it with a leading slash.

## Why RESTORE and not RETIRE

Measured, not assumed. `kipi-update.sh:2468,2592` copies `git archive HEAD -- q-system/` into every instance with `rsync -a --delete`, and `.q-system` is not in `INSTANCE_OWNED_SUBTREES` (`kipi-update.sh:71-79`). So `q-system/.q-system/commands.md`, which carries the full spec for all nine, already reaches the fleet. Only the load path for the NAME was missing. Retiring would have deleted nine specified, synced workflows from the menu a founder reads.

Each command file is a LOADER pointing at its `commands.md` entry, never a restatement, so it cannot drift from the spec it names — same shape as `q-debrief.md` pointing at `debrief-template.md`.

## The check now reads both menus

`README.md` writes its menu as table rows (`| `/name` |`) where `CLAUDE.md` writes list items (`- `/name``). A parser that knows one shape grades the other as empty, which reads as green. The README half gets its OWN injected synthetic row in the negative self-test: sharing one fixture would have proven only that the list shape can go red, while the table shape — the half this PR adds — stayed untested and looked covered.

`GRANDFATHERED` is now empty. The array stays rather than being deleted, because the rot check reads it; an empty list makes that loop a no-op, which is the right shape for "no exemptions", and re-adding one re-arms the check.

## Also

`plugins/kipi-core/skills/research-mode/commands/q-research.md` is deleted. A plugin's `commands/` directory is discovered at the plugin root only, so that file was inert — `/q-research` was absent from the session command list while `kipi-core:say` and `kipi-core:wiring-check` were present.

`README.md`'s "No file under a `commands/` directory backs them" is now false and is rewritten.

## Evidence

RED, before the files shipped, against nine names:

```
$ bash q-system/.q-system/scripts/test/test_q_debrief_command_wired.sh
  /q-calibrate is on the CLAUDE.md menu with no file at plugins/*/commands/q-calibrate.md
  /q-create ... /q-draft ... /q-engage ... /q-handoff ... /q-morning ...
  /q-plan ... /q-research ... /q-wrap ...
FAIL: CLAUDE.md names commands that ship no file in a path kipi-update.sh syncs
exit=1
```

GREEN, after:

```
PASS: every command on the CLAUDE.md and README.md menus ships a file in a synced path
exit=0
```

Mutation, 3/3 killed (remove one command file, re-run, restore):

```
mutant q-calibrate  exit=1
mutant q-wrap       exit=1
mutant q-research   exit=1
restored            exit=0
```

README extraction proven non-empty, so the parse floor is not silently zero:

```
q-debrief q-draft q-engage q-handoff q-morning q-research q-wrap
```

All lefthook pre-commit gates green (instruction-budget ratchet, plugin-version-bump, enforced-claim, covers-declaration, verify.sh).

## Not doing

No change to what any restored command DOES. No new sync path: `.claude/commands` stays unsynced.


assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-388-20260920-072802.md` (6355 bytes).

```
FINDINGS:
major|/q-morning writes the day's freshness receipt, so the 09:00 deadman stays silent if the 07:40 job then fails, while the same file claims the deadman still covers it|plugins/kipi-core/commands/q-morning.md:27
major|/q-calibrate runs only step 4 of the 4-step ENFORCED ripple check, and step 4 alone exits 0 "Nothing to verify"|plugins/kipi-core/commands/q-calibrate.md:28
minor|the new README table parser matches only "| " plus backtick, so ordinary table rows without that exact spacing are skipped silently|q-system/.q-system/scripts/test/test_q_debrief_command_wired.sh:92
minor|README documents three commands from the same plugins/kipi-core/commands/ directory in two different invocation forms|README.md:363
minor|check_menu reports every missing name as "on the CLAUDE.md menu" even when the failing file is README.md|q-system/.q-system/scripts/test/test_q_debrief_command_wired.sh:118
END FINDINGS
```

--- reviewer output ---

Review of PR #388 done. Five findings survived reproduction, two of them real behavior defects.

## Findings

**1. major — `/q-morning` silences the 09:00 deadman for the rest of the day, and the same file says it does not.**
`plugins/kipi-core/commands/q-morning.md:16` runs `morning-brief.py` with no flag. `main()` writes the freshness receipt on every non-dry-run invocation (`morning-brief.py:983`). The deadman's `check()` is date-only: it compares `data["date"]` to today and ignores the `at` timestamp (`morning-brief-deadman.py:110-118`). So an early manual run satisfies the watcher for the whole day. Line 27 of the same new file asserts the opposite.

Reproducer and real output:

```
$ python3 -c '... dm.check(now_at_1030, receipt stamped today at 06:00) ...'
scenario: /q-morning run manually at 06:00, then the 07:40 launchd job never fires
deadman verdict at 10:30 -> SILENT, ok=True
receipt timestamp the deadman ignored: 2026-09-20T06:00:00
```

The receipt-writing mechanism predates this PR. What this PR adds is a one-keystroke registered command that reaches it, plus a line claiming coverage that is gone the moment you follow the command above it. The cheapest fix is `--dry-run` in the command file, or `at`-aware `check()`.

**2. major — `/q-calibrate` runs step 4 of a 4-step ENFORCED check and prints a green that means nothing.**
`plugins/kipi-core/commands/q-calibrate.md:28` runs only `ripple-verify.py`. `commands.md:46-51` spells the check as four steps, and step 1 is `changelog-write.py`, which is what produces the entries `ripple-verify.py` reads. The command file never names step 1. Line 31 calls exit 1 a soft gate; exit 1 is unreachable when step 1 was skipped.

```
$ python3 q-system/.q-system/scripts/ripple-verify.py q-system/canonical/changelog.md 2026-09-20
No changelog entries for 2026-09-20. Nothing to verify.
exit=0
```

That is success reported for work that did not happen, inside a file whose closing paragraph says it does not restate the spec so it cannot drift from it. It restates one quarter of it.

**3. minor — the new README table parser matches exactly one cell spacing.**
`test_q_debrief_command_wired.sh:92` requires `| ` then a backtick. Ordinary markdown without the space is skipped silently, which is the same class of bug the PR was written to fix.

```
$ printf '| `/q-nofile-a` | ... |\n|`/q-nofile-b`| ... |\n|  `/q-nofile-c` | ... |\n' | grep -oE '^\| `/[a-z0-9-]+`'
| `/q-nofile-a`
```

Two of three rows vanish. `^\|[[:space:]]*` closes it.

**4. minor — README now documents same-directory commands two different ways.**
`README.md:349,353` write `/q-debrief` and `/q-morning` bare; `README.md:363` writes `/kipi-core:say`. All three files sit in `plugins/kipi-core/commands/`. At most one invocation form is correct, and the README asserts both. I could not settle which empirically: the sandbox blocked reading `~/.claude/plugins/`, so I am reporting the self-contradiction, not a claim about the loader.

**5. minor — the check misdirects the operator when README is the failing file.**
`test_q_debrief_command_wired.sh:118` hardcodes "is on the CLAUDE.md menu". Running `check_menu` against a README copy with an injected row:

```
  /q-ghost is on the CLAUDE.md menu with no file at plugins/*/commands/q-ghost.md
```

The name is in README. The outer `fail` names the right file, so this is bounded, but the per-name line sends you grepping the wrong one.

## What is sound

- **Every path the nine new files reference exists.** `morning-brief.py`, `loop-tracker.py`, `ripple-verify.py`, `handoff-provenance-lint.py`, `anti-hallucination.md`, `relationships.md`, `objections.md`, `current-state.md`, `.claude/rules/evidence-ledger.md`. `python3 q-system/.q-system/loop-tracker.py list` exits 0.
- **The unpinned-model scar does not apply.** `morning-brief.py:113` sets `env["ANTHROPIC_MODEL"] = BRIEF_MODEL` itself, so a manual run does not ride the default model.
- **`q-morning.md`'s "not a second implementation" is true.** `com.kipi.morning-brief.plist` `ProgramArguments` is `/usr/bin/python3 __KIPI_REPO__/q-system/.q-system/scripts/morning-brief.py`. Same script.
- **The README negative self-test is not decoration.** I re-ran the extracted functions against a README copy with a synthetic table row and it returned 1. Deleting the table grep would be caught.
- **The extractor's documented skips behave as documented.** `/q-market-*`, `/kipi-core:rca-start`, and `| nine /prd-os:*` are all correctly not extracted, for the reasons the comment gives.
- **`GRANDFATHERED=""` is safe.** The rot loop is a no-op and the `case " " in *" $name "*` guard never matches, so no name is accidentally exempted.
- **All nine restored names have a real spec row in `commands.md`.** The loader-not-restater shape holds for eight of nine; `q-calibrate` is finding 2.

## VERDICT: REQUEST CHANGES

Not BLOCK. Nothing here is permanent or unrecoverable, and the PR's core claim (nine menu names now ship files, the check reads both menus, the negative self-tests are live) holds up under mutation.

**Fix first: finding 1.** A command that quietly disables the only alarm covering a founder-facing daily job is worse than the missing command file this PR set out to fix, and the file asserting the alarm still works is what makes it invisible. `--dry-run` in `q-morning.md:16`, or make `check()` read `at`.

FINDINGS:
major|/q-morning writes the day's freshness receipt, so the 09:00 deadman stays silent if the 07:40 job then fails, while the same file claims the deadman still covers it|plugins/kipi-core/commands/q-morning.md:27
major|/q-calibrate runs only step 4 of the 4-step ENFORCED ripple check, and step 4 alone exits 0 "Nothing to verify"|plugins/kipi-core/commands/q-calibrate.md:28
minor|the new README table parser matches only "| " plus backtick, so ordinary table rows without that exact spacing are skipped silently|q-system/.q-system/scripts/test/test_q_debrief_command_wired.sh:92
minor|README documents three commands from the same plugins/kipi-core/commands/ directory in two different invocation forms|README.md:363
minor|check_menu reports every missing name as "on the CLAUDE.md menu" even when the failing file is README.md|q-system/.q-system/scripts/test/test_q_debrief_command_wired.sh:118
END FINDINGS

PR #390, opened 2026-09-20
fix(guard): the MCP denylist is keyed on the operation, not the server name (ASK-1923)

assafkip:
Closes ASK-1923 (do not merge yet: closeout runs through /issue-verify and /issue-closeout).

## What was measured

Against the **live** hook at `~/.claude/hooks/destructive-op-deny.sh`, by running the matcher on 29 real MCP tool names. Not read off it.

```
$ python3 q-system/.q-system/scripts/mcp-denylist-namespace-check.py --hook <guard>
DIRECTION 2 -- wildcards on namespaces nothing registers:
  DEAD   mcp__plugin_Notion_notion__*
  DEAD   mcp__plugin_linear_linear__*
  DEAD   mcp__plugin_vercel_vercel__*
...
FAIL: 14 case(s) wrong.
EXIT=1
```

**Over-broad.** `mcp__plugin_linear_linear__list_issues`, `__get_issue`, `mcp__plugin_vercel_vercel__list_deployments` and `mcp__plugin_Notion_notion__notion-search` all measure DENY. The only carve-out is `*authenticate*`, so read-only queries are refused.

**Under-broad, the serious half.** All three wildcarded namespaces are DEAD. The Linear server that actually loads is `mcp__linear__`, so `mcp__linear__delete_issue` measures ALLOW, against the founder's global CLAUDE.md which names Linear `*delete*` as hook-blocked regardless of mode. `mcp__supabase__delete_branch` is on no list at all. Ten destructive operations measure ALLOW in total.

## What ships

Three files. The hook itself is **not** edited: it is founder-owned and founder-applied.

- `q-system/.q-system/scripts/mcp-denylist-namespace-check.py` — the checker. Exits 1 when a known-destructive tool is allowed or a known-safe one denied; `--report` prints the enumeration and always exits 0. The denylist's namespaces are PARSED out of the hook. The expectations are not derived from it, on purpose: an oracle that read its expectations off the matcher would agree with a broken matcher. Every case row carries `source`, so no tool name here was invented to make a case.
- `q-system/.q-system/proposals/mcp-denylist-operation-split.json` — the diff. `insert_before` the existing block: deny on the destructive OPERATION for every server, and let read verbs through before the wildcards can refuse them. Apply with `bash q-system/.q-system/scripts/apply-claude-changes.sh q-system/.q-system/proposals/mcp-denylist-operation-split.json --root $HOME`.
- `q-system/.q-system/tests/test_destructive_op_mcp_namespace.py` — drives the live guard and a patched copy. READ-ONLY on `~/.claude`.

## Read this before approving the diff

**It loosens as well as tightens.** `apply-claude-changes` is additive-only precisely so an agent cannot widen a gate; an early `exit 0` is the one shape that gets around that. It is named in the inserted comment rather than smuggled. The read-verb list is the whole of what it opens, and nothing on it writes. Your call, not the tool's.

## Evidence

```
checker on the unpatched guard  -> FAIL, 14 cases wrong, 3 DEAD wildcards, exit 1
pytest test_destructive_op_mcp_namespace.py -q -> 43 passed in 3.31s
lefthook pre-commit (verify.sh --staged) -> ok, 4 checks
```

Each RED case asserts the defect is still present in its direction, and `test_the_checker_fails_on_the_unpatched_guard` is the negative control: no case here is one that has never been seen failing.

## Adjacent, not fixed here

`sp-059c80d9` describes two untracked WIP files in the `kipi-system` root that are earlier, unfinished attempts at these same two filenames. This branch is their address; the tracked versions here are green. Resolve that row at closeout.


assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-390-20260920-084532.md` (6606 bytes).

```
FINDINGS:
major|One MCP approval token unlocks any other MCP destructive op: COMMAND is empty for MCP calls so every deny hashes identically, and the shipped suite cannot see it because decide() uses a bare HOME with no capability-token.sh|q-system/.q-system/proposals/mcp-denylist-operation-split.json:10
minor|The `drop` verb denies mcp__playwright__browser_drop, a drag-and-drop gesture, with a message claiming vendor-side deletion|q-system/.q-system/proposals/mcp-denylist-operation-split.json:10
minor|MCP_MUTATION_SCOPED_SERVER is case-sensitive, so mcp__claude_ai_Vercel__* mutating ops stay ALLOW, the same server-name miss this PR exists to fix|q-system/.q-system/proposals/mcp-denylist-operation-split.json:10
minor|The inserted comment claims the read-verb list opens nothing that writes; `resolve` is on it and mcp__linear__resolve_diff_thread writes|q-system/.q-system/proposals/mcp-denylist-operation-split.json:10
END FINDINGS
```

--- reviewer output ---

Review complete. Evidence below, all reproducers run.

## MAJOR 1 — one MCP approval unlocks a different MCP delete

`q-system/.q-system/proposals/mcp-denylist-operation-split.json:10`

For an MCP tool call `COMMAND` is empty (`tool_input.command` does not exist), and `emit_deny` hashes `command + cwd`. So every MCP denial in a cwd produces the **same** hash. The deny text says "Approve THIS command out-of-band". It does not approve THIS command.

Before the patch this barely mattered: only 7 literal MCP names reached `emit_deny`. The patch routes every destructive operation on every server through it, so the approval path is now the normal path.

Reproducer, run against a throwaway HOME with the real `capability-token.sh` copied in:

```
1) founder is shown this deny for delete_comment:
    destructive-op-deny: MCP operation 'delete_comment' on server 'linear' deletes or
    unlinks state at the vendor side. ... Approve THIS command out-of-band: kipi-approve 20fca18c23f

2) hash shown for mcp__linear__delete_comment  : 20fca18c23fb19da...096aefc9
   hash shown for mcp__supabase__delete_branch : 20fca18c23fb19da...096aefc9
   SAME HASH: True

3) founder runs kipi-approve for the delete_comment hash -> approved one command, expires in 300s

4) agent then calls mcp__supabase__delete_branch -> ALLOW
```

A Supabase branch delete is not recoverable. The founder approved a Linear comment delete.

**Why no test caught it.** `decide()` at `mcp-denylist-namespace-check.py:214` runs every case in a bare `tempfile.TemporaryDirectory()` HOME. `~/.claude/bin/capability-token.sh` is absent there, so `[ -x "$_ct" ]` is false and the token branch is dead in all 43 tests. Same tool, the shipped `checker.decide`, two HOMEs:

```
shipped checker.decide (bare HOME):                                      deny
shipped checker.decide (prod HOME, token minted for the LINEAR deny):   allow
```

Fix: hash `TOOL_NAME` into the capability token for MCP calls, or refuse the token path when `COMMAND` is empty.

## MINOR 2 — a drag-and-drop gesture is denied as a vendor-side delete

`mcp-denylist-operation-split.json:10`, the `drop` verb in `MCP_DESTRUCTIVE_OP`

```
mcp__playwright__browser_drop      CURRENT allow   CANDIDATE deny   <-- CHANGED
mcp__playwright__browser_drag      CURRENT allow   CANDIDATE allow
reason: MCP operation 'browser_drop' on server 'playwright' deletes or unlinks state at the vendor side.
```

Playwright is loaded in this session. `browser_drop` is the second half of a drag gesture and touches no vendor state.

## MINOR 3 — the Vercel rule repeats the exact bug this PR fixes

`mcp-denylist-operation-split.json:10`, `MCP_MUTATION_SCOPED_SERVER='(^|_)vercel(_|$)'`

Case-sensitive, matched against the server segment. The connector this session actually lists is `claude.ai Vercel`, whose namespace follows `mcp__claude_ai_Gmail__` / `mcp__claude_ai_Supabase__`:

```
mcp__vercel__deploy                            deny
mcp__plugin_vercel_vercel__deploy              deny
mcp__claude_ai_Vercel__create_deployment       allow
mcp__claude_ai_Vercel__update_project          allow
mcp__claude_ai_Vercel__deploy_to_production    allow
```

Founder global CLAUDE.md names "Vercel mutating ops" as hook-blocked. Not a regression (it was allow before too), and nothing from that connector is authorized yet, which is why this is minor and not higher.

## MINOR 4 — "nothing on it writes" is false for the read list

`mcp-denylist-operation-split.json:10`, and the same sentence in the PR body

`resolve` is in `MCP_READ_OP`. `mcp__linear__resolve_diff_thread` matches it and exits 0 on the read branch. Resolving a review thread is a write. Measured `allow` on both current and candidate, so no behavior regression; the comment is what is wrong.

## What is sound

Attacks that did not land:

- **Fixture fidelity.** The test builds the candidate by string replace. I ran the real engine instead: `apply-claude-changes.sh ... --root <tmp>` returned rc 0, and `engine-applied == test-built candidate: True`. Byte identical. This is the class that usually breaks here, and it holds.
- **The exec-bit scar (ASK-1118).** `mode after apply: 0o755`. The apply path did not disarm the guard.
- **Anchor uniqueness.** `anchor count: 1` against the live hook.
- **Ordering.** Destructive check runs before the read exit, so a name carrying both verbs denies.
- **Boundary anchoring.** `mcp__claude_ai_Gmail__untrash_message` measured `allow`. The `trash` verb does not swallow its inverse.
- **Blast containment.** `Write` does not route into the MCP branch; the Bash denials (`rm -v -rf`, `git reset --hard`, `git push --force`, `git clean -fd`) still deny under the patched copy.
- **Exit codes as claimed.** Checker exits 1 on the live guard, 0 on the candidate, 0 on `--report`.
- **Suite runs.** `43 passed in 3.54s`, and `q-system/.q-system/tests` is a directory entry in `.verify-suites`, so CI does run it. The capability gate is red on this tree for 12 pre-existing `declared_inert` rows, none of them from this PR.
- **The negative control is real.** `test_the_checker_fails_on_the_unpatched_guard` is the right shape and does go red on the unpatched guard.

## VERDICT: REQUEST CHANGES

Not BLOCK. The patched guard is strictly better than what is live today, the hook is founder-applied, and reverting is a file edit.

**Fix first:** the capability token. Bind it to `TOOL_NAME`, or make `emit_deny` skip the token path when `COMMAND` is empty. Right now the sentence "Approve THIS command" is false on every MCP denial the patch creates, and the suite structurally cannot see it because its HOME has no token script.

FINDINGS:
major|One MCP approval token unlocks any other MCP destructive op: COMMAND is empty for MCP calls so every deny hashes identically, and the shipped suite cannot see it because decide() uses a bare HOME with no capability-token.sh|q-system/.q-system/proposals/mcp-denylist-operation-split.json:10
minor|The `drop` verb denies mcp__playwright__browser_drop, a drag-and-drop gesture, with a message claiming vendor-side deletion|q-system/.q-system/proposals/mcp-denylist-operation-split.json:10
minor|MCP_MUTATION_SCOPED_SERVER is case-sensitive, so mcp__claude_ai_Vercel__* mutating ops stay ALLOW, the same server-name miss this PR exists to fix|q-system/.q-system/proposals/mcp-denylist-operation-split.json:10
minor|The inserted comment claims the read-verb list opens nothing that writes; `resolve` is on it and mcp__linear__resolve_diff_thread writes|q-system/.q-system/proposals/mcp-denylist-operation-split.json:10
END FINDINGS

assafkip:
## Round 2 — all four findings fixed (54d8ff1e)

**major — one MCP approval token unlocked any other.** Confirmed. `emit_deny` hashes `$COMMAND` + `$CWD`, an MCP payload has no `.tool_input.command`, so every MCP denial in one cwd hashed the empty string. Reproduced on the **live** guard with the **real** `capability-token.sh` copied into a throwaway HOME: deny Gmail `delete_label`, mint its hash, and Calendar `delete_event` walks through. Fix: the insert sets `COMMAND="$TOOL_NAME ${TOOL_INPUT:-}"` at the top of the `mcp__*` branch, before either block can deny, so the server-name block below inherits the scoping too. `checker.grant_leak()` measures it and the checker exits 1 on `LEAK`. A stub token script is deliberately not used: how a grant is scoped is the thing under measurement, and a stub would encode the test's guess at it.

**minor — `drop`.** Off the destructive verb list. `mcp__playwright__browser_drop` and `browser_drag` are oracle cases now. A gate that blocks a mouse gesture gets switched off.

**minor — case-sensitive server match.** All three greps are `-Eqi`. `mcp__claude_ai_Vercel__update_project` is a deny case now.

**minor — `resolve` writes.** `resolve` is off the read list; the inserted comment now makes the narrower claim that is actually true.

Also changed: the proposal JSON is **generated** by `build-mcp-proposal.py` (`--check` is a test, and the single-writer chokepoint). Hand-editing escaped shell inside a JSON string is how a quoting defect reaches a security gate.

### Evidence

```
checker on the LIVE guard          -> 15 cases wrong, 3 DEAD wildcards, DIRECTION 3: LEAK
mcp-candidate-measure.py (patched) -> DIRECTION 3: scoped, OK every case agrees, EXIT=0
pytest                             -> 55 passed in 6.77s  (was 43)
lefthook pre-commit                -> verify.sh ok (4 checks)
```

Four mutants, one per new decision point, each flipping its check back to the finding it fixes: scoping removed gives LEAK; `resolve` re-added lets a write take the read exit; `drop` re-added denies `browser_drop`; `-Eqi` back to `-Eq` allows the capital-V connector.

The hook itself is still untouched and read-only. Founder-owned, founder-applied.

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-390-20260920-090132.md` (7421 bytes).

```
FINDINGS:
major|Applying the proposal turns 6 of the PR's own 55 tests red, including its negative control and 2 of 4 mutants|q-system/.q-system/tests/test_destructive_op_mcp_namespace.py:203
major|The checker exits 0 and prints OK on a guard that allows mcp__supabase__execute_sql with DROP TABLE and apply_migration|q-system/.q-system/scripts/mcp-denylist-namespace-check.py:331
minor|force_close on the destructive list denies the local loop_force_close tool with a false message about vendor-side deletion|q-system/.q-system/proposals/mcp-denylist-operation-split.json:16
minor|Reassigning COMMAND for the token hash also writes every MCP tool_input payload into the plaintext audit log|q-system/.q-system/proposals/mcp-denylist-operation-split.json:16
END FINDINGS
```

--- reviewer output ---

Review done. Read-only, nothing in the repo touched.

## Findings

### 1. major — applying the fix turns 6 of this PR's own 55 tests red, including its negative control

`q-system/.q-system/tests/test_destructive_op_mcp_namespace.py:203`

The suite resolves `HOOK` to the live guard when one exists. The founder applying the proposal is the entire point of this PR. The day they do, `current` is no longer unpatched and the suite breaks.

Reproducer: build an already-applied guard from the repo fixture, point the suite at it.

```
python3 -m pytest q-system/.q-system/tests/test_destructive_op_mcp_namespace.py -q
  (KIPI_DESTRUCTIVE_HOOK = a guard with the proposal already applied)
```

Real output:

```
FAILED ...TestTheGuardIsWrongInBothDirectionsToday::test_a_destructive_op_on_a_registered_namespace_is_allowed
FAILED ...TestTheGuardIsWrongInBothDirectionsToday::test_a_read_only_query_is_denied
FAILED ...TestTheGuardIsWrongInBothDirectionsToday::test_one_mcp_approval_token_unlocks_a_different_destructive_op
FAILED ...TestTheOperationKeyedSplitFixesBothDirections::test_the_checker_fails_on_the_unpatched_guard
FAILED ...TestEachFixCanGoRed::test_without_the_command_scoping_the_token_leaks_again
FAILED ...TestEachFixCanGoRed::test_a_case_sensitive_server_match_misses_the_capital_v_connector
6 failed, 49 passed in 6.05s
```

Three of those six carry the "this case has outlived it, delete it" docstring, so they are intentional. Three do not: `test_the_checker_fails_on_the_unpatched_guard` is the negative control, and two of the four mutants in `TestEachFixCanGoRed` fail because the already-applied copy decides before the mutated copy can. Those are the checks that make every other assertion in the file believable, and they are silently coupled to the guard being unpatched.

Worse at 3am: CI has no live hook, so it falls back to `tests/fixtures/destructive-op-deny.reference.sh`, which this PR does not update and which nothing pins to the live file. Grepping for consumers of that fixture returns only `probe_hook.py` and the new checker. So CI stays green forever while the founder's machine is red, and the two are measuring different guards.

### 2. major — the checker prints OK while `execute_sql "DROP TABLE users"` walks through

`q-system/.q-system/scripts/mcp-denylist-namespace-check.py:331`

The insert's comment says the rule is "keyed on the OPERATION, which is stable across servers." `mcp__supabase__` is registered and authenticated in this session, and its highest-blast-radius operations carry no destructive verb in the name.

Reproducer: drive the candidate guard on real supabase tool names, then run the checker on the same guard.

Real output:

```
tool                                           current   candidate
mcp__supabase__execute_sql                     allow     allow
mcp__supabase__apply_migration                 allow     allow
mcp__supabase__delete_branch                   allow     deny
mcp__claude_ai_Supabase__execute_sql           allow     allow

OK: every case agrees with the oracle.
checker RC 0
```

`delete_branch` is denied and `execute_sql "DROP TABLE users;"` is allowed, on the same server, and the checker calls that a pass. This is not a regression, both were allowed before. What ships here is the green measurement that reads as coverage, on a 29-name oracle the author chose. Permanent and unrecoverable at the vendor if it fires.

### 3. minor — a local bookkeeping tool is denied with a message about the vendor side

`q-system/.q-system/proposals/mcp-denylist-operation-split.json:16` (`force[_-]close` in `MCP_DESTRUCTIVE_OP`)

`mcp__plugin_kipi-core_kipi__loop_force_close` is a registered tool that closes a follow-up loop in a local JSONL. Real output from the candidate guard:

```
deny mcp__plugin_kipi-core_kipi__loop_force_close | destructive-op-deny: MCP operation
'loop_force_close' on server 'plugin_kipi-core_kipi' deletes or unlinks state at the
vendor side. Keyed on the oper...
```

No vendor, no deletion. This is the same shape the PR correctly argues against for `drop` and the mouse gesture: a gate whose message is false on a real tool is a gate someone switches off.

### 4. minor — the COMMAND reassignment starts persisting every MCP payload to the audit log

`q-system/.q-system/proposals/mcp-denylist-operation-split.json:16` (`COMMAND="$TOOL_NAME ${TOOL_INPUT:-}"`)

`log_decision` writes `"cmd": <COMMAND>` on every path, allows included. Scoping the token hash by reassigning the shared `COMMAND` also rewrites what gets logged. Real output, same call against both guards:

```
CURRENT :  ..."decision":"allow","reason":"no destructive pattern matched","cmd":"\n"
CANDIDATE: ..."decision":"allow","reason":"no destructive pattern matched",
           "cmd":"mcp__supabase__execute_sql {\"query\":\"select * from users where
           api_key = 'sk-live-SECRET'\"}\n"
```

Full tool_input now lands in plaintext in `~/.claude/audit/destructive-op-deny.log` for every MCP call. A separate variable for the hash scope keeps the token fix without the logging side effect.

## What is sound

- `bash -n` clean on the applied guard, and clean again when applied twice.
- I hunted for a write that the new read-verb exit lets past the server-name wildcards. `notion-create-pages` and `notion-update-page` on the wildcarded namespace still deny. `resolve` was already taken off the read list. Dropped the finding.
- `untrash_thread` and `unmark_thread_spam` are not caught by `trash`. The `(^|[_-])...([_-]|$)` anchors hold. Dropped.
- The token scoping genuinely closes the grant leak. The checker's DIRECTION 3 reports `scoped` on the candidate and `LEAK` on the current guard, measured with the real `capability-token.sh` copied in rather than a stub. That is the strongest part of this PR.
- The `ALLOW_DESTRUCTIVE` bypass and the whole Bash denial block sit ahead of the anchor and are untouched.
- Generating the JSON from `build-mcp-proposal.py` instead of hand-escaping shell inside a JSON string, with a `--check` test, is the right call.
- Read-only on `~/.claude` is real. Nothing in these files writes the live hook.

## VERDICT: REQUEST CHANGES

Not a BLOCK. Merging changes nothing on disk, the founder applies the proposal separately, so nothing here is unrecoverable on merge.

Fix first: finding 1. Decouple the negative control and the mutants from "the live guard is unpatched" by building `current` from a pinned unpatched copy rather than from `HOOK`, and decide what the three self-retiring cases do the day the fix lands. Right now the success of this PR is what breaks it.

FINDINGS:
major|Applying the proposal turns 6 of the PR's own 55 tests red, including its negative control and 2 of 4 mutants|q-system/.q-system/tests/test_destructive_op_mcp_namespace.py:203
major|The checker exits 0 and prints OK on a guard that allows mcp__supabase__execute_sql with DROP TABLE and apply_migration|q-system/.q-system/scripts/mcp-denylist-namespace-check.py:331
minor|force_close on the destructive list denies the local loop_force_close tool with a false message about vendor-side deletion|q-system/.q-system/proposals/mcp-denylist-operation-split.json:16
minor|Reassigning COMMAND for the token hash also writes every MCP tool_input payload into the plaintext audit log|q-system/.q-system/proposals/mcp-denylist-operation-split.json:16
END FINDINGS

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-390-20260920-091855.md` (6628 bytes).

```
FINDINGS:
major|MCP_DESTRUCTIVE_SQL requires an optional PostgreSQL keyword, so `TRUNCATE users;` and `ALTER TABLE users DROP email;` measure ALLOW while the deny message claims TRUNCATE and DROP coverage|q-system/.q-system/scripts/build-mcp-proposal.py:81
minor|the *authenticate* carve-out moved from 7 named namespaces to every MCP server and now runs before the destructive-operation check, so any op name containing `authenticate` skips the gate|q-system/.q-system/scripts/build-mcp-proposal.py:119
minor|the destructive-SQL rule matches anywhere in the payload, so an additive migration defining a function whose body contains DELETE FROM, and a SELECT with that phrase in a string literal, are refused|q-system/.q-system/scripts/build-mcp-proposal.py:81
nit|PR body evidence says 43 passed in 3.31s; the shipped suite is 66 passed in 6.65s|q-system/.q-system/tests/test_destructive_op_mcp_namespace.py:1
END FINDINGS
```

--- reviewer output ---

## PR #390 review — ASK-1923 MCP denylist

Read-only. Built the candidate by applying the proposal to the repo's reference fixture, drove it with the PR's own checker module, and ran the shipped suite. Nothing in `~/.claude` was written.

---

### major — `TRUNCATE users;` runs at ALLOW while the gate says it is covered

`q-system/.q-system/scripts/build-mcp-proposal.py:81`

`MCP_DESTRUCTIVE_SQL` requires an object keyword after the verb. In PostgreSQL, which is what Supabase runs, that keyword is optional in exactly the two forms that matter.

Reproducer and real output:

```
=== G. optional-keyword SQL forms (valid PostgreSQL) ===
  ALLOW  execute_sql       'TRUNCATE users;'                      <<< MISS, should be deny
  DENY   execute_sql       'TRUNCATE TABLE users;'                ok
  DENY   apply_migration   'ALTER TABLE users DROP COLUMN email;' ok
  ALLOW  apply_migration   'ALTER TABLE users DROP email;'        <<< MISS, should be deny
```

Same regex, separator side, constructed inputs so read this half lower:

```
  DENY   'DROP TABLE users;'      DENY   'DROP\nTABLE users;'
  ALLOW  'DROP\tTABLE users;'     ALLOW  'DROP\r\nTABLE users;'
```

Why it matters at 3am. The deny message enumerates "DROP / TRUNCATE / DELETE FROM" and the proposal's `reason` field says the same. `TRUNCATE users;` is the shortest correct spelling, not an evasion. An unattended agent that types it empties a production table and the new gate logs allow.

Why the suite did not see it. The oracle holds `DROP TABLE users;` and `ALTER TABLE users DROP COLUMN email;` only, so `checker.failures()` returns `[]` and the run prints OK. The mutant `test_without_the_payload_check_a_drop_table_walks_through` also uses `DROP TABLE`, so it proves the branch is live, never that the branch covers what the message claims. That is the oracle-picked-by-the-author shape the repo's own lessons corpus names.

Fix is one regex plus two oracle rows: make the object keyword optional after `truncate`, and match `drop` followed by an identifier inside an `ALTER TABLE`.

---

### minor — the `*authenticate*` carve-out went from 7 namespaces to every server, ahead of the destructive check

`q-system/.q-system/scripts/build-mcp-proposal.py:119`

In the old hook that carve-out lived inside the seven-pattern wildcard branch. The insert lifts it to the top of an all-servers `mcp__*` branch, before `MCP_DESTRUCTIVE_OP` runs.

```
=== N. the *authenticate* carve-out now precedes the destructive check ===
  ALLOW  mcp__plugin_vercel_vercel__authenticate
  DENY   mcp__linear__delete_issue
  ALLOW  mcp__linear__delete_authenticated_issue
  ALLOW  mcp__supabase__delete_branch_authenticate
```

Constructed input, dropped a level for it: no server in this session's list emits a name like that. The comment says the carve-out was "kept from the block below so a fix cannot quietly drop it" and does not say its scope changed.

---

### minor — the statement rule matches anywhere in the payload, so additive migrations get refused

`q-system/.q-system/scripts/build-mcp-proposal.py:81`

```
  DENY   "SELECT * FROM audit WHERE action = 'delete from users'"
  DENY   'CREATE FUNCTION purge() RETURNS void AS $$ BEGIN DELETE FROM stale; END $$;'
```

A migration that defines a cleanup function is a normal Supabase move and now costs one approval. Fail-closed, bounded, consistent with the trade this hook already makes elsewhere. Worth a line in the comment so it is a stated cost rather than a surprise.

---

### nit — PR body evidence is stale

Body says `pytest ... 43 passed in 3.31s`. Measured: `66 passed in 6.65s`.

---

## What is sound

- **Capability-token scoping.** Minted a grant for `Gmail delete_label` and offered it to `Calendar delete_event`: `grant_leak: scoped`. Same tool with a different payload earns a different hash (`263152fa…` vs `24c31ab5…`) and the AAA grant does not open BBB.
- **Audit-log hygiene.** `payload_in_audit_log: digest`. The probe values `SECRET-LABEL-42` and `hunter2` do not reach the log.
- **The delivery path actually works.** Ran `apply-claude-changes.sh` against a throwaway `--root` holding a copy of the reference guard: `exit 0`, `hooks 0->0, gates held`, insert landed, mode still `0755`. The additive-only ratchet does not refuse the early `exit 0`.
- **The read-verb exit does not open a write on a wildcarded namespace.** `mcp__plugin_linear_linear__resolve_diff_thread` measures DENY; `resolve` really is off the read list.
- **Vercel case-insensitivity.** `mcp__claude_ai_Vercel__update_project` denies; the mutant that makes the match case-sensitive flips it to allow.
- **Undo verbs are not caught.** `untrash_message` allows, `restore_issue_label` allows. The `(^|[_-])` anchoring works.
- **Blast radius outside MCP is zero.** The Bash denials still fire, `rm /tmp/one-file.txt` still runs, `Write` does not route through the MCP branch.
- **Test wiring.** `.verify-suites` names `q-system/.q-system/tests` as a directory, so the new file is collected. `build-mcp-proposal.py --check` is pinned by a test. Suite: 66 passed.
- **The negative controls are real.** `test_the_checker_fails_on_the_unpatched_guard` and the six mutants in `TestEachFixCanGoRed` each break one clause and watch the decision flip back.

This is unusually well-instrumented work. The one hole is on the axis the instrumentation could not see, which is the axis it always is.

---

**VERDICT: REQUEST CHANGES**

Fix first: make the object keyword optional in `MCP_DESTRUCTIVE_SQL` and add `TRUNCATE users;` plus `ALTER TABLE users DROP email;` as oracle rows, so the checker's OK means what the deny message says.

FINDINGS:
major|MCP_DESTRUCTIVE_SQL requires an optional PostgreSQL keyword, so `TRUNCATE users;` and `ALTER TABLE users DROP email;` measure ALLOW while the deny message claims TRUNCATE and DROP coverage|q-system/.q-system/scripts/build-mcp-proposal.py:81
minor|the *authenticate* carve-out moved from 7 named namespaces to every MCP server and now runs before the destructive-operation check, so any op name containing `authenticate` skips the gate|q-system/.q-system/scripts/build-mcp-proposal.py:119
minor|the destructive-SQL rule matches anywhere in the payload, so an additive migration defining a function whose body contains DELETE FROM, and a SELECT with that phrase in a string literal, are refused|q-system/.q-system/scripts/build-mcp-proposal.py:81
nit|PR body evidence says 43 passed in 3.31s; the shipped suite is 66 passed in 6.65s|q-system/.q-system/tests/test_destructive_op_mcp_namespace.py:1
END FINDINGS

PR #391, opened 2026-09-20
fix(covers): one glob matcher instead of a branch per `**` position (ASK-1922)

assafkip:
Closes ASK-1922. Three minors from the claude reviewer on PR #385 round 2, which the reviewer again reported as `minors captured as spillover: 0 of 3`.

## 1. The structural one

Round 1 of #385 found a MAJOR: a leading `**/` never matched a root-level file, missing all 39 root-level scripts. Round 2 found the **same defect one position over**: an interior `**/` (`q-system/**/*.sh` vs `q-system/x.sh`) misses depth zero the same way, and covers-lint printed PASS on it.

Same finding CLASS two rounds running means the fix SHAPE is wrong, not that a third case needs patching. So this is not a fourth branch.

`covers_matches` was hand-rolled `fnmatch` with special cases bolted on, and `fnmatch` has no concept of `**` at all. It only appeared to work because its `*` crosses `/`, which also made `*` and `**` synonyms and left `**` as decoration. It is now ONE translate to an anchored regex:

- `**` as a whole segment is zero or more segments, at ANY position
- `*` and `?` stop at `/`, which is what leaves `**` something to mean
- a trailing `**` is everything below that directory
- a slash-less pattern still matches the basename, the way `find -name` does
- the match is anchored at the repo root. The old third branch also tried `*/<pattern>`, so `scripts/*.py` silently claimed any `scripts/` in the tree

**Seen RED first**, against the shipped matcher:

```
AssertionError: ('q-system/**/*.sh', 'q-system/x.sh', True)
  where False = covers_matches('q-system/**/*.sh', 'q-system/x.sh')
```

`DOUBLE_STAR_CASES` is the table the DoR asks for: 24 rows over every `**` position (leading, interior, trailing, doubled), with a False row beside every True one, because a matcher that says yes to everything passes a table made only of matches.

**The one live narrowing, measured rather than assumed.** Strict `*` moved exactly one real declaration: `plugins/prd-os/scripts/*.py` went 14 tracked files to 13, dropping the `test/` subdir. That fragment now declares `**/*.py` and is back to 14. All 8 fragments that declare `covers` were re-checked, not just the one that moved.

## 2. `covers` is now read for every changed path

It used to sit inside the per-path loop AFTER that loop's `continue` for a path that is itself a declared test, so a glob naming another declared test's path was silently not honoured. Silent, which is the bad kind: covers-lint could not see it either, because such a glob does match a tracked file. It is its own pass now.

## 3. `read_declared` voids a malformed `covers` WHOLE

It kept the good entries and dropped only the bad ones, so `["**/*.plist", null]` still took the test OFF the always-run floor on a declaration nobody could read. The docstring claimed the safe-by-default posture and the code did the unsafe one. The code moved to meet the docstring, since the docstring is the one arguing for the floor.

## covers-lint

Gains the decidable half of "this glob cannot match what its author meant": a `**` INSIDE a segment (`**.py`, `a**b`) is refused, because there it degrades to a plain `*` and stops at a `/`. That is the same silent-narrowing class as rounds 1 and 2, at the one position the matcher still cannot honour. Paired with a negative self-test that `**/*.plist`, `launchd/**` and `**/launchd/**/*.plist` still pass, so the refusal is not an outage wearing a gate's name.

## Mutants

Four new decision points, four mutants, each killing a named case:

- `**` spanning zero segments (`(?:[^/]+/)*` to `[^/]+/`, the reintroduced round-1/2 bug)
- `*` stopping at a separator (`[^/]*` to `.*`)
- `covers` read for every changed path
- a malformed `covers` voiding the whole list

## Verification

```
$ python3 -m pytest q-system/.q-system/scripts/test_change_size.py \
                    q-system/.q-system/scripts/test_covers_lint.py -q
58 passed in 1.87s

$ python3 q-system/.q-system/scripts/covers-lint.py --repo-root .
covers-lint: PASS -- 8 of 278 checked fragment(s) declare coverage

$ python3 q-system/.q-system/scripts/change-size.py --base origin/main
change-size: tier L (5 files, 164 app-code lines)
  because: q-system/.q-system/scripts/change-size.py decides what runs
  tests: the FULL suite (278 declared)
```

The full suite is correct here: this diff is the machinery that decides what runs.

**Not doing** (DoR): widening what `covers` can express.


assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-391-20260920-094506.md` (6371 bytes).

```
FINDINGS:
minor|covers-lint's refusal message and the covers_matches docstring say a `**` inside a segment "stops at a `/`", which is false for a slash-less glob like `**.plist` because the basename rule matches it at any depth|q-system/.q-system/scripts/covers-lint.py:123
minor|Two new decision points in `_segment_re` (the `?` branch and the bracket-class branch) have no test case and no mutant; mutating either leaves all 58 tests green|q-system/.q-system/scripts/change-size.py:182
END FINDINGS
```

--- reviewer output ---

## What I did

Read the diff, then attacked it: measured old-versus-new matching across every real `covers` declaration, tried to crash the new regex translator, checked the new tests can go red, and ran two hand-built mutants against the suite.

---

## Finding 1 (minor) — the lint's explanation is false for the exact glob its own new test uses

`q-system/.q-system/scripts/covers-lint.py:123` tells the operator a `**` inside a segment "is just `*`, which stops at a `/`". The same claim is in the matcher docstring at `q-system/.q-system/scripts/change-size.py:254`. For a glob with no `/` in it, the basename rule at `covers_matches` fires first, so it does not stop at a `/` at all. `**.plist` is the first parametrized case in the new `test_a_double_star_inside_a_segment_is_refused`.

Reproducer:

```
python3 - <<'PY'
import importlib.util
spec = importlib.util.spec_from_file_location("cs", "q-system/.q-system/scripts/change-size.py")
cs = importlib.util.module_from_spec(spec); spec.loader.exec_module(cs)
for pat, path in [("**.py","a/b/c.py"), ("*.py","a/b/c.py"), ("launchd/**.plist","launchd/deep/a.plist")]:
    print("%-20s %-24s -> %s" % (pat, path, cs.covers_matches(pat, path)))
PY
```

Real output:

```
**.py                a/b/c.py                 -> True
*.py                 a/b/c.py                 -> True
launchd/**.plist     launchd/deep/a.plist     -> False
```

The slashed form does stop at a `/`. The slash-less form does not, and `**.py` and `**/*.py` select the identical set. The refusal itself is still right; the sentence under it is not.

---

## Finding 2 (minor) — two new branches in `_segment_re` have no case and no mutant, and survive mutation

`q-system/.q-system/scripts/change-size.py:182` (`?` becomes `[^/]`) and `:195` (the bracket class) are new decision points. `DOUBLE_STAR_CASES` has no row using `?` or `[`. The test file's own docstring says every decision point carries a mutant and a guard whose removal nothing notices is decoration.

Reproducer:

```
python3 - <<'PY'
import shutil, subprocess, tempfile
from pathlib import Path
src = Path("q-system/.q-system/scripts")
MUTANTS = [("? stops at a separator", 'out.append("[^/]")', 'out.append(".")'),
           ("a bracket class is a class",
            'out.append("[" + ("^" + body[1:] if body[:1] in "!^" else body) + "]")',
            'out.append("[^/]")')]
for label, old, new in MUTANTS:
    d = Path(tempfile.mkdtemp()); shutil.copytree(src, d / "scripts")
    f = d / "scripts" / "change-size.py"; t = f.read_text(); assert old in t
    f.write_text(t.replace(old, new, 1))
    r = subprocess.run(["python3","-m","pytest", str(d/"scripts"/"test_change_size.py"),
                        str(d/"scripts"/"test_covers_lint.py"),"-q","-p","no:cacheprovider"],
                       capture_output=True, text=True, cwd=str(d))
    print("MUTANT:", label, "->", (r.stdout or r.stderr).strip().splitlines()[-1])
PY
```

Real output:

```
MUTANT: ? stops at a separator -> 58 passed in 2.90s
MUTANT: a bracket class is a class -> 58 passed in 3.13s
```

No fragment in the repo writes a `?` or a `[` in a `covers` glob today, so nothing is wrong at runtime. Dropping a level for that: the branches are correct, just unheld.

---

## What is sound

These are attacks I ran that the code survived.

**The live-narrowing claim holds, and I checked set equality, not counts.** I loaded `covers_matches` from `origin/main` and from HEAD, ran every declared glob from both branches against `git ls-files`, and diffed the matched sets. Only `plugins/prd-os/scripts/*.py` moved. `**/*.py` restores the identical 14 files: `old-only: []`, `new-only: []`. The strict `*` alone would have dropped exactly `plugins/prd-os/scripts/test/test_spillover_ledger_root.py`, which is the one file the PR body names.

**The anchoring change loses nothing real.** The only other losses across all 8 fragments were `.review-scratch/repo/q-system/...` copies, which are committed scratch. That is the `*/<pattern>` branch being removed on purpose, which is the fix, not a regression.

**I could not make the translator crash.** Tried `[]]`, `a/[]]/b`, `a[`, `**`, `a/**/`, `**/`, `/a.py`, `./a.py`, `a**/b`. None raised; Python's `re` accepts `[]]`. And even if one did, `capability-gate.py:715` catches bare `Exception` and returns `None`, which the docstring and the gate mutant both pin as "run the full suite". A bad glob fails loud and safe, never silently to zero tests.

**The new covers-lint tests can go red.** Neutering `_bad_double_star` to return `None` makes all three refusal params exit 0, so that test fails; over-firing it makes the negative self-test fail on `**/*.plist`. Both directions are held. The repo fixture builds its paths from a real `git ls-files`, the same producer production uses, so the fixture shape is not invented.

**The mutant table is not stale.** Every row asserts its `old` string is still present in the shipped source, and the two rewritten rows (the 21-space indent change on "covers never selects a non-scanner", and the covers-own-pass row) match. Suite: `58 passed in 1.87s`.

**The width cap is not at risk.** The new covers pass can add at most the 8 declared scanners to `pulled`, against `MAX_SELECTED` of 60.

**Both PR verification claims reproduce verbatim:** `covers-lint: PASS -- 8 of 278` and `change-size: tier L (5 files, 164 app-code lines)`, full suite because the diff is the classifier itself.

**Nothing here is permanent or unrecoverable.** Pure classifier, no writes, no Linear objects, no concurrency. The full suite still runs on every push to main, so a selection that is too narrow is caught at merge. Rollback is a revert.

---

## VERDICT: APPROVE WITH NITS

Fix first: the `covers-lint.py:123` message and the `change-size.py:254` docstring, so they describe what the slash-less case actually does.

FINDINGS:
minor|covers-lint's refusal message and the covers_matches docstring say a `**` inside a segment "stops at a `/`", which is false for a slash-less glob like `**.plist` because the basename rule matches it at any depth|q-system/.q-system/scripts/covers-lint.py:123
minor|Two new decision points in `_segment_re` (the `?` branch and the bracket-class branch) have no test case and no mutant; mutating either leaves all 58 tests green|q-system/.q-system/scripts/change-size.py:182
END FINDINGS

PR #392, opened 2026-09-20
fix: the scanner floor sees recursive grep, and a lost finding says so (ASK-1921)

assafkip:
Closes the four findings the claude reviewer left on PR #377 (APPROVE WITH NITS, 2026-09-20), the last of which is that the other three reached nobody.

## 1. minor: recursive grep is a tree walk the detector could not see

`_SCAN_RE` in `change-size.py` knew glob / iterdir / listdir / scandir / walk / fnmatch / find / ls / `git ls-files` and a shell for-glob. It did not know `grep -r`. Two declared tests walk the tree that way and were therefore OFF the always-run floor, which is the safety argument for the whole selector: a test that reads files nobody named cannot be safely skipped. ASK-1918's `covers` mechanism is built on this function, so the gap propagates.

Widened against the corpus, not speculatively (the issue's Not-doing line):

```
declared tests: 278
test/test-claude-write-path.sh: grep -r  scans_the_tree=True
test/test-token-guard-hook-behavior.sh: grep -r  scans_the_tree=False
test/test-zero-safe-count-idiom.sh: grep -r  scans_the_tree=False
test_token_guard_commit_forms.py: grep -r  scans_the_tree=True
tests/test_promotion_receipt.py: grep -r  scans_the_tree=True
totals: declared tests with grep -r = 5, with an rg invocation = 0
```

`rg` is NOT added: zero declared tests invoke it (the only `rg` token in the corpus is a local variable in `test_dc_reader_verdicts.py`), and the new test asserts that boundary, so flipping it later is a visible decision rather than a drift.

The comment claiming the detector "does not need the spelling" is what made this invisible, and it now says what is actually true: fewer spellings than a pattern parser, one per WALKING TOOL rather than one per pattern.

**RED first, against the pre-fix source:**

```
HEAD~1: scans_the_tree(test/test-zero-safe-count-idiom.sh) = False
HEAD~1: scans_the_tree(test/test-token-guard-hook-behavior.sh) = False
HEAD~1: 2 of 2 declared grep -r tree-walkers are OFF the floor   (rc=1)
HEAD:   ... = True / True
HEAD:   0 of 2 declared grep -r tree-walkers are OFF the floor   (rc=0)
```

The test reads those two REAL declared tests off disk rather than a fixture, and a mutant deletes the added alternative and must make the case fail.

## 2. minor: the CAPABILITY_GATE_SKIP comment misstated the code

`.github/workflows/validate.yml` said the gate step above "just ran the full suite". On a `pull_request` it runs the SUBSET that diff earns (ASK-1749), so on a PR the full suite runs nowhere inside that job; it runs on the push to main after merge. The skip is still correct. Comment only.

## 3. nit: an empty selection reported green with zero tests run

`select_for_diff` accepted `set()` and `run_tests` then skipped every artifact and exited 0: green over zero evidence. No producer emits an empty list today because the scanner floor guarantees at least 72, which is exactly why it would be invisible on the day one does, and ASK-1918 shrinks that floor on purpose.

Two guards, one layer apart:

- empty selection returns `None`, the same full-suite answer every other unvouchable case here gets.
- `ran == 0` with nothing quarantined and nothing skipped is an error. A run that skipped everything on purpose ACCOUNTS for its silence; a run that executed nothing and cannot say why does not.

Each has a mutant that deletes it and must make its case fail.

## 4. THE MINORS PATH: `0 of 2` was never an outage

`minors captured as spillover: 0 of 2` printed on PR #377 and nothing noticed the zero. APPROVE WITH NITS is terminal, so both findings lived only in a PR comment; downstream that run was byte-identical to one that found no minors.

**This section was rewritten after round 1 of review. The first version added a LOST alert on top of the capture. The reviewer showed the capture can never succeed, so the alert was permanent.** Reproduced against the real producer:

```
refused: a minor is fixed in this change or rejected with a reason; it is never queued (founder 2026-09-12)
rc=2
```

`spillover add` was called with no `--severity`. That defaults to `minor`, which sits in `SPILLOVER_REFUSED_SEVERITIES`, so the call returned 2 on every run: `CAPTURED` was 0 **by construction**, not by outage, and an alarm built on that zero fires on every approved PR carrying a nit.

So both halves come out, and neither is re-severitied. Filing a review minor at `medium` to clear the door launders the severity the reviewer chose, and `no-orphan-findings.md` already names the only two legal ends for a NEW minor: fixed in this change, or rejected with a reason. The ledger is neither. The page goes with it: a 100-percent-rate alert on a policy refusal is the cry-wolf shape, and the genuine loss on this path keeps its own page further down (a review that never REACHED the issue).

What ships is the true statement: each minor **named** with claim and location, plus one `UNROUTED` line saying the verdict is terminal, that the ledger refuses a minor by policy, and where the findings already live (the PR comment and the FINDINGS block on the issue).

```
  minor 1: the help text omits --repo-root (a.py:10)
  minor 2: the comment misstates the code (b.yml:72)
  review minors on a terminal verdict: 2
  UNROUTED: ... nothing files them. They are on PR #392 and in the FINDINGS block on ASK-9999.
```

The test drives the SHIPPED block by awk range rather than a copy. Its exit-code knob went out with the call it drove: the stub `prd_runner.py` now ABORTS if anything calls it, the "no capture call" assertion is derived from the shipped source, and a mutant re-adds the call and must be caught. **Seen RED first** against the pre-fix source, 5 of 9 assertions failing:

```
FAIL: the block calls 'spillover add' again...
FAIL: the output tallies the minors without naming them; absent: [the help text omits --repo-root] ...
FAIL: the block called prd_runner: spillover add --source ASK-9999 --desc PR #392 review minor: ...
FAIL: an ordinary APPROVE WITH NITS with minors paged the alert path: ...
```

Green after: `ALL PASS`, rc=0.

### Still not fixed here, and captured

Nobody is ASSIGNED the fix-or-reject decision `no-orphan-findings.md` requires. Routing that is a change with real Linear inflow and its own blast radius, so it stays captured rather than bundled: sp-74e671a4, filed as **ASK-1940**.

## Checks run

- `python3 -m pytest q-system/.q-system/scripts/test_change_size.py -q` -> **48 passed** (includes the two new cases and three new mutants)
- `bash q-system/.q-system/scripts/test/test-minor-capture-signal.sh` -> **ALL PASS**, rc=0 (9 assertions, 1 mutant; registered in the capability manifest)
- `bash q-system/.q-system/scripts/test/test-severity-floor.sh` -> **PASS 201/201**, rc=0 (the neighbour that owns `pr-review-agent.sh`)
- `python3 q-system/.q-system/scripts/test_capability_gate.py` -> **ALL PASS**
- `python3 q-system/.q-system/scripts/covers-lint.py --repo-root .` -> **PASS**, 8 of 279 fragments declare coverage


assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-392-20260920-102256.md` (6393 bytes).

```
FINDINGS:
major|spillover add passes no --severity so prd_runner refuses every minor with exit 2, making CAPTURED always 0 and the new LOST alert fire on every APPROVE WITH NITS review with a minor|q-system/.q-system/scripts/pr-review-agent.sh:1218
minor|the new test's healthy-capture fixture stubs prd_runner at exit 0, a code the real producer never returns for this call, so both negative assertions cover an unreachable state|q-system/.q-system/scripts/test/test-minor-capture-signal.sh:95
minor|the alert message fingerprints identically across PRs because alert-to-linear strips all digits, so losses after the first become a counter bump on a ticket titled with a different PR|q-system/.q-system/scripts/pr-review-agent.sh:1220
END FINDINGS
```

--- reviewer output ---

Read the PR, ran the three changed paths, wrote reproducers against the real producers.

## Finding 1

**major** — the spillover capture the new alert watches can never succeed, so the alert fires on every APPROVE WITH NITS review that has a minor.

`pr-review-agent.sh:1187` calls `spillover add` with no `--severity`. `prd_runner.py:2448` refuses the default `minor` with exit 2 (`SPILLOVER_REFUSED_SEVERITIES = ("minor","low","nit")`, founder rule 2026-09-12). So `CAPTURED` is 0 by construction, not by outage. The new block at `pr-review-agent.sh:1218` reads that 0 as an exception and pages Sana.

Reproducer, the real `prd_runner.py` with the reviewer's exact argument list:

```
rc= 2
stderr: refused: a minor is fixed in this change or rejected with a reason; it is never queued (founder 2026-09-12)
A real finding at medium or above: pass --severity medium|high|major|blocker.
```

Then the shipped block itself, extracted from `pr-review-agent.sh` by the same awk range the new test uses, driven with `SKEL` pointed at this repo so the capture hits the real runner:

```
--- stdout ---
  minors captured as spillover: 0 of 2
--- stderr ---
  LOST: 2 minor finding(s) were extracted from this review and NONE reached the spillover ledger...
--- notified (Sana's queue) ---
reviewer: PR #377 (ASK-9999) verdict APPROVE WITH NITS with 2 minor finding(s) extracted and 0 captured as spillover...
```

The comment at `pr-review-agent.sh:1213` says "a capture that cannot file is not evidence the code is bad", which reads the refusal as availability. It is policy, and it is 100 percent. What a human does at 3am: reads the ticket, finds nothing actionable, closes it, gets it again on the next approved PR. That is the cry-wolf shape.

`q-system/.q-system/scripts/pr-review-agent.sh:1218`

## Finding 2

**minor** — the new test's healthy-capture fixture uses an exit code the real producer cannot return.

`test-minor-capture-signal.sh:95` runs `run_block 0`, stubbing `prd_runner.py` to exit 0. Per finding 1 the real call returns 2 for every minor, so both negative cases ("a run that captured everything raises nothing", "a healthy capture pages nobody") assert over a state production never reaches. The test is green and the defect walks past it:

```
PASS: a healthy run still reports its count (2 of 2)
PASS: a run that captured everything raises nothing
PASS: a healthy capture pages nobody
ALL PASS
```

Same run, real producer, rc=2 always (pasted above).

`q-system/.q-system/scripts/test/test-minor-capture-signal.sh:95`

## Finding 3

**minor** — the page dedupes across PRs, so after the first firing the loss is a counter bump on a ticket naming the wrong PR.

`alert-to-linear.py:133` strips every digit before hashing. Two different PRs produce one fingerprint:

```
fingerprint A: 839d23c6c20f90e4
fingerprint B: 839d23c6c20f90e4
same shape: True
title A: reviewer: PR #377 (ASK-1921) verdict APPROVE WITH NITS with 2 minor finding(s) extracted and 0 captured as spi...
```

A repeat only posts the most-recent message after `REPEAT_COMMENT_AFTER_HOURS = 12` (`alert-to-linear.py:99`). Two approved PRs in one day means PR two's loss reaches Linear as `count+1` on a ticket pointing at PR one. The stderr line is still in the run log, so it is bounded, not silent.

`q-system/.q-system/scripts/pr-review-agent.sh:1220`

## What is sound

- **The regex widening is measured, not speculative.** Loaded `HEAD~3` and `HEAD` side by side over all 279 declared tests: exactly 2 flip, `test-token-guard-hook-behavior.sh` and `test-zero-safe-count-idiom.sh`, matching the PR body. Floor after: 94 scanners.
- **I tried to find a fifth spelling in the real corpus and failed.** `grep -R` capital: 0. `egrep -r` / `fgrep -r`: 0. `grep --recursive`: 0. `rg -` invocation: 0. All four are genuine blind spots of `\bgrep\s+(?:-\w+\s+)*-[A-Za-z]*r`, and none is a shape any declared test emits, so I dropped the finding rather than construct an input nobody produces.
- **Mutants kill.** `test_change_size.py` 48 passed, including the new `recursive grep is a tree walk` mutant and the standalone empty-selection mutant that needs monkeypatch.
- **`NOTIFY` cannot abort the script.** It is bound at `pr-review-agent.sh:135` before `set -uo pipefail` matters, so the new `bash "$NOTIFY"` has no unbound-variable path.
- **The awk extraction anchors correctly.** The new inner `if` closes with an indented `fi`, so `/^fi$/` takes the outer block. The test also asserts `spillover add` is inside the extracted range.
- **`select_for_diff`'s empty guard fails in the safe direction**: returns `None`, which is full suite, same as every other unvouchable case, and the broad `except` still wraps it.
- **The manifest fragment name is right.** `capability_manifest.fragment_name` returns `q-system__.q-system__scripts__test__test-minor-capture-signal.sh.json`, byte-identical to the file added.
- **The `ran == 0` backstop has a real false branch.** Quarantined and skeleton-only runs account for their silence and stay green; I could not build a legitimate zero-execution run that this wrongly reddens.

## VERDICT

**REQUEST CHANGES.** Not a block: nothing here is permanent or unrecoverable, and the Linear side dedupes to one ticket rather than one per PR.

Fix first: `pr-review-agent.sh:1187` either passes a severity the ledger accepts, or the capture and its new alert come out together because `no-orphan-findings.md` already says a new minor is never queued. Shipping an alarm on a path that refuses by policy makes the alarm permanent and the finding still lost.

FINDINGS:
major|spillover add passes no --severity so prd_runner refuses every minor with exit 2, making CAPTURED always 0 and the new LOST alert fire on every APPROVE WITH NITS review with a minor|q-system/.q-system/scripts/pr-review-agent.sh:1218
minor|the new test's healthy-capture fixture stubs prd_runner at exit 0, a code the real producer never returns for this call, so both negative assertions cover an unreachable state|q-system/.q-system/scripts/test/test-minor-capture-signal.sh:95
minor|the alert message fingerprints identically across PRs because alert-to-linear strips all digits, so losses after the first become a counter bump on a ticket titled with a different PR|q-system/.q-system/scripts/pr-review-agent.sh:1220
END FINDINGS

assafkip:
## Rework round 1: all three findings addressed in `c374f1f3`

I reproduced finding 1 before touching anything. The reviewer's exact argument list against the real `prd_runner.py`:

```
refused: a minor is fixed in this change or rejected with a reason; it is never queued (founder 2026-09-12)
A real finding at medium or above: pass --severity medium|high|major|blocker.
rc=2
```

and the shipped block at `HEAD`, driven with 2 minors and `SKEL` pointed at this repo:

```
  minors captured as spillover: 0 of 2
  LOST: 2 minor finding(s) were extracted from this review and NONE reached the spillover ledger...
--- pages sent to Sana's queue ---
reviewer: PR #392 (ASK-9999) verdict APPROVE WITH NITS with 2 minor finding(s) extracted and 0 captured...
```

Confirmed. Nothing was down; the door refuses by policy, 100 percent of the time.

---

### Finding 1 (major) — fixed, and the fix is a removal

Taken as written, including the reasoning: an alarm on a path that refuses by policy is permanent, and the finding is still lost. I took the second of the two paths offered, because the first one is worse than it looks.

**Not `--severity medium`.** Filing a review minor at medium to clear the door launders the severity the reviewer chose. `no-orphan-findings.md` names exactly two legal ends for a NEW minor: fixed in this change, or rejected with a reason. The ledger is neither of them, which is precisely why `prd_runner.py:2448` refuses it. The capture call was never a capture; it was a no-op the code then counted.

So the call is deleted, and the page with it. A 100-percent-rate alert on a policy refusal is the cry-wolf shape you named, and `founder-notifications.md` asks for a state change once, not per-event noise. The genuine loss on this path already has its own page further down (`pr-review-agent.sh:1390`): a review that never REACHED the issue, where the findings really are unreadable by anyone.

What replaces both is the true statement:

```
  minor 1: the help text omits --repo-root (a.py:10)
  minor 2: the comment misstates the code (b.yml:72)
  review minors on a terminal verdict: 2
  UNROUTED: the 2 minor(s) above got a terminal APPROVE WITH NITS, so the rework loop
  stops here. The spillover ledger refuses a minor by policy -- it is fixed in this
  change or rejected with a reason (no-orphan-findings.md) -- so nothing files them.
  They are on PR #392 and in the FINDINGS block on ASK-9999. Owner for the
  fix-or-reject routing: ASK-1940.
```

The minors are now NAMED with claim and location rather than tallied, so the run log carries the findings themselves. They also reach two durable places without this block's help: the review comment on the PR, and the FINDINGS block the reviewer posts onto the Linear issue.

**What this still does not fix, said plainly:** nobody is ASSIGNED the fix-or-reject decision. That is a routing change with real Linear inflow and its own blast radius, so it stays captured rather than bundled: sp-74e671a4 / **ASK-1940**.

`q-system/.q-system/scripts/pr-review-agent.sh:1179-1225`

### Finding 2 (minor) — fixed; the unreachable fixture went out with the call it drove

Right, and it was the same defect wearing a test's clothes. With no external producer left in the block there is no exit-code axis to stub, so `run_block` no longer takes one — a knob for it would reintroduce exactly this.

The stub `prd_runner.py` now **aborts and records** if anything calls it, so a re-added capture fails loudly instead of passing through a friendly stub. The "no capture call" assertion is DERIVED from the shipped source (`grep` over the awk-extracted range) rather than restated, and a mutant re-adds the call to a copy and must be caught:

```
PASS: the shipped block does not call 'spillover add' (the door that refuses a minor by policy)
PASS: each minor is NAMED with its location, not only tallied
PASS: a terminal verdict with minors says so out loud (UNROUTED)
PASS: no external producer was invoked at all
PASS: a policy refusal pages nobody (no 100%-rate alert)
PASS: zero minors raises nothing
PASS: mutant killed: the runner-call check sees the re-added capture
ALL PASS
```

**Seen RED first** against the pre-fix source, 5 of 9 assertions failing, including the two that matter:

```
FAIL: the block calls 'spillover add' again...
FAIL: the output tallies the minors without naming them; absent: [the help text omits --repo-root] ...
FAIL: the block called prd_runner: spillover add --source ASK-9999 --desc PR #392 review minor: ...
FAIL: an ordinary APPROVE WITH NITS with minors paged the alert path: reviewer: PR #392 ...
```

`q-system/.q-system/scripts/test/test-minor-capture-signal.sh`

### Finding 3 (minor) — rejected with a reason, not queued

The alert it describes no longer exists, so the instance is gone with finding 1.

I checked whether the general shape survives at the other call site (`pr-review-agent.sh:1390`, the review-did-not-reach-the-issue page) and it does: `alert-to-linear.py:132` strips `\d+`, so two PRs hash the same. That is the file's stated design, not a defect in it — its own comment says "word content decides the fingerprint; nothing else does", and the collapse is the flood control it exists to provide. Two firings of ONE detector are meant to be one ticket with a counter, and the message is a repeat comment carrying the most recent text after 12 hours. Per `no-orphan-findings.md` a new minor is never queued, so this is rejected here with that reason rather than filed.

---

### Checks run

- `bash q-system/.q-system/scripts/test/test-minor-capture-signal.sh` → **ALL PASS**, rc=0 (9 assertions, 1 mutant)
- `bash q-system/.q-system/scripts/test/test-severity-floor.sh` → **PASS: 201/201**, rc=0 (the neighbour that owns `pr-review-agent.sh`)
- pre-commit: `verify.sh ok`, `covers-lint PASS`, `enforced-claim PASS`, `instruction-budget RATCHET PASS`

No manifest change: the test's path is unchanged, so `q-system__.q-system__scripts__test__test-minor-capture-signal.sh.json` still matches.

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-392-20260920-103814.md` (7034 bytes).

```
FINDINGS:
minor|The UNROUTED message claims the minors are on the PR and on the Linear issue, but it prints outside and before the `if [ "$POST" = "1" ]` block, so a dry run states it after posting nothing|q-system/.q-system/scripts/pr-review-agent.sh:1225
minor|`grep -R` and `grep --recursive` are the same walking tool as `grep -r` but the regex requires a lowercase r, and only the `rg` boundary is pinned by a test|q-system/.q-system/scripts/change-size.py:183
minor|test-findings-block-reader.sh still says extract_minor_findings feeds `spillover add` and writes permanent ledger items, a path this PR deleted|q-system/.q-system/scripts/test/test-findings-block-reader.sh:103
END FINDINGS
```

--- reviewer output ---

## PR #392 review, senior-staff adversarial pass

Three minors survived reproduction. No blocker, no major.

---

### minor 1: `grep -R` and `grep --recursive` are still invisible to `scans_the_tree`

**file:** `q-system/.q-system/scripts/change-size.py:183`

The stated design rule in the comment above the regex is "one spelling per WALKING TOOL, not one per pattern". `-R` and `--recursive` are the same tool. The regex requires a lowercase `r` inside a short-option cluster, so both spellings fall through. The test at `test_change_size.py:215` pins the `rg` boundary explicitly and says nothing about these two, so the next declared test that writes `grep -R` drops off the always-run floor exactly the way the two tests this PR rescues did.

**Reproducer (loads both module versions, run against the real repo):**

```
python3 -c '<load HEAD and ed14f6e5~1 copies of change-size.py, call scans_the_tree>'
```

**Real output:**

```
q-system/.q-system/scripts/test/test-zero-safe-count-idiom.sh  BEFORE=False  AFTER=True
q-system/.q-system/scripts/test/test-token-guard-hook-behavior.sh  BEFORE=False  AFTER=True
q-system/.q-system/scripts/test/test-claude-write-path.sh  BEFORE=True  AFTER=True
grep -R (capital) seen AFTER: False
grep --recursive seen AFTER: False
```

**Severity dropped one level on purpose.** I measured the declared corpus myself before filing this, and no producer emits either shape today:

```
declared fragments: 279
grep -r shipped regex: 6
grep -R capital: 0
grep --recursive: 0
rg invocation: 0
```

So this is a narrow false negative on a shape nobody hits yet, not a live gap. `-[A-Za-z]*[rR]` closes it in one character.

---

### minor 2: the UNROUTED line tells the operator the minors reached two places this run may never have written

**file:** `q-system/.q-system/scripts/pr-review-agent.sh:1225`

The message ends with "They are on PR #$PR and in the FINDINGS block on $ISSUE." Everything that writes those two places lives inside `if [ "$POST" = "1" ]`, which opens 58 lines later, and the minors block references `POST` nowhere.

```
$ grep -n 'UNROUTED|^if \[ "\$POST" = "1" \]; then$|DRY RUN -- nothing posted' pr-review-agent.sh
205:[ "$POST" = "1" ] || DRY_NOTE=" (DRY RUN -- nothing posted to GitHub or Linear; ...)"
1225:    echo "  UNROUTED: the $MINOR_COUNT minor(s) above ... They are on PR #$PR and in the FINDINGS block on $ISSUE. ..."
1283:if [ "$POST" = "1" ]; then
```

And the block really does emit that line, executed through the PR's own harness:

```
$ bash q-system/.q-system/scripts/test/test-minor-capture-signal.sh
PASS: a terminal verdict with minors says so out loud (UNROUTED)
...
ALL PASS
```

Two reachable shapes. A hand run without `--post` prints the claim and then prints `DRY RUN -- nothing posted to GitHub or Linear`. A `--post` run whose Linear post fails prints the claim at 1225 and then `WARN: could not post the review to $ISSUE` at 1394. The 3am action exists and is bounded, because the WARN and the page still fire, so this is a log that contradicts itself rather than a lost finding. It is the same class as this PR's own finding 2: a wrong why-line is worse than none.

---

### minor 3: `test-findings-block-reader.sh` still documents the ledger path this PR deleted

**file:** `q-system/.q-system/scripts/test/test-findings-block-reader.sh:103`

```
$ grep -n "spillover add|PERMANENT ledger|reach the ledger|permanent spillover item" test-findings-block-reader.sh
103:# extract_minor_findings feeds `spillover add`, which writes PERMANENT ledger
126:      permanent spillover item holding the standing gate red. Captured:
130:ok "only the live block's minors are captured (withdrawn findings do not reach the ledger)"
```

```
$ grep -rn "spillover add" q-system/.q-system/scripts/pr-review-agent.sh
1184:# claude review of PR #392, finding 1). It called `prd_runner.py spillover add`
```

The only remaining mention in the reviewer is the past-tense why-comment. The test's assertions are still correct and still worth keeping, but its stated blast radius (withdrawn findings becoming permanent ledger rows that hold the standing gate red) is no longer reachable from this producer, and a green `ok` line prints a claim about a ledger nothing writes.

---

## What is sound

- **The justification for deleting the capture.** `prd_runner.py:2448` reads `(args.severity or "minor")` and `SPILLOVER_REFUSED_SEVERITIES` at `:2780` is `("minor", "low", "nit")`. The refusal is real, the `0 of 2` really was by construction, and this PR is not deleting a working path.
- **Cry-wolf risk of the new `zero tests executed` error.** I traced every caller. `kipi` and `kipi-update.sh` pass no `--diff-base`, so `only` is None; instance mode carries 21 skeleton-only entries so `skipped` is never 0; and `change-size` only ever returns paths already in the manifest, so a non-empty selection cannot filter to nothing. I could not construct a false alarm.
- **The stderr move.** `kipi-dispatch.sh:1443` launches the reviewer with `stdout=log, stderr=log`, so putting the named minors on stderr does not drop them from the converge log.
- **The `covers` interaction with the widened regex.** Of the 6 declared tests carrying `grep -r`, only `test_promotion_receipt.py` also declares `covers`, and it was already a scanner before this change. Nothing moved OFF the floor.
- **The awk block extraction in the new test.** The inner `fi` is indented and the anchor is `^fi$`, so the range lands on the right block. Its mutant builds and kills, which I saw execute.
- **Regression risk in the neighbour that owns this file.** `test-severity-floor.sh` declares `covers` on `pr-review-agent.sh`: 201/201 pass.
- **ReDoS on the new alternation.** `-\w+\s+` repeated is unambiguous because `\w` and `\s` do not overlap, so there is no nested-quantifier blowup.
- **Mutation coverage of all three fixes.** 48 pass in `test_change_size.py`, including the new `recursive grep is a tree walk`, `zero executed is never green`, and the standalone empty-selection mutant.

---

## VERDICT: APPROVE WITH NITS

Fix minor 2 first. It is the only one an operator reads at 3am, and it reintroduces, in the replacement text, the same shape the PR removed: a message asserting a route the run did not take.

FINDINGS:
minor|The UNROUTED message claims the minors are on the PR and on the Linear issue, but it prints outside and before the `if [ "$POST" = "1" ]` block, so a dry run states it after posting nothing|q-system/.q-system/scripts/pr-review-agent.sh:1225
minor|`grep -R` and `grep --recursive` are the same walking tool as `grep -r` but the regex requires a lowercase r, and only the `rg` boundary is pinned by a test|q-system/.q-system/scripts/change-size.py:183
minor|test-findings-block-reader.sh still says extract_minor_findings feeds `spillover add` and writes permanent ledger items, a path this PR deleted|q-system/.q-system/scripts/test/test-findings-block-reader.sh:103
END FINDINGS

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-392-20260923-051704.md` (6958 bytes).

```
FINDINGS:
minor|scans_the_tree misses the subprocess-list spelling of recursive grep, so declared test test_lessons_recall_corpus.py is off the always-run floor, and the new comment claims that form occurs nowhere in the corpus|q-system/.q-system/scripts/change-size.py:184
minor|the UNROUTED message states the minors are on the PR and the Linear issue, but it prints outside and before the POST=1 block, so a dry run asserts a posting that never happened|q-system/.q-system/scripts/pr-review-agent.sh:1237
minor|test-findings-block-reader.sh still documents and labels extract_minor_findings as feeding spillover add and writing permanent ledger items, a path this PR deleted|q-system/.q-system/scripts/test/test-findings-block-reader.sh:103
END FINDINGS
```

--- reviewer output ---

Review complete. Here is what I found, with reproducers.

## Finding 1 — minor: the detector is still blind to the subprocess-list spelling of `grep -r`, and the new comment says that form does not exist

`q-system/.q-system/scripts/change-size.py:184` matches the shell spelling only. `q-system/.q-system/scripts/test_change_size.py:198` states, as measured fact: *"a subprocess LIST form (`["grep", "-rIl", ...]`) occurs nowhere, so it is not a form this pattern claims to see."* That is false against the declared corpus.

Reproducer:

```
python3 -c 'import importlib.util,pathlib
root=pathlib.Path(".").resolve()
spec=importlib.util.spec_from_file_location("cs",root/"q-system/.q-system/scripts/change-size.py")
cs=importlib.util.module_from_spec(spec); spec.loader.exec_module(cs)
rel="q-system/.q-system/tests/test_lessons_recall_corpus.py"
print("scans_the_tree =",cs.scans_the_tree((root/rel).read_text()))'
```

Real output:

```
line 171: callers = subprocess.run(["grep", "-rl", "lessons_recall", str(HERE.parent.parent.parent / "plugins"), str(SCRIPTS)],
scans_the_tree(q-system/.q-system/tests/test_lessons_recall_corpus.py) = False
scans_the_tree(grep list form) = False
scans_the_tree(shell form)     = True
```

That declared test greps recursively over `plugins/` and `scripts/` to discover callers and then runs their suites. It reads files nobody named, which is the definition the floor exists for. It is not on the floor:

```
declared: 280 scanners (always-run floor): 94
target on the floor: False
```

And a diff it scans skips it:

```
changed = [("plugins/kipi-core/skills/improve/scripts/recall_bridge.py", 12)]
tier: S full_suite: False selected: 94
blind scanner selected? False
```

Not a repeat of round 2's `grep -R` / `--recursive` note. I measured those against all 280 declared tests and got zero hits, so those two stay theoretical. This one the corpus actually emits.

## Finding 2 — minor: the UNROUTED message claims a posting that did not happen (round 2, still live)

Round 2 raised it, the head has not moved since, and it reproduces. `pr-review-agent.sh:1237` prints unconditionally; every poster is inside `if [ "$POST" = "1" ]` at `:1295`, and the minors block references `POST` zero times.

Reproducer: extract the shipped block and run it with `POST=0`.

```
rc = 0
--- stdout ---
  review minors on a terminal verdict: 1

--- stderr ---
  minor 1: the help text omits --repo-root (a.py:10)
  UNROUTED: ... They are on PR #392 and in the FINDINGS block on ASK-9999. Owner for the fix-or-reject routing: ASK-1940.
```

Nothing was posted. The message says it was. Severity unchanged from round 2.

## Finding 3 — minor: a neighbour test still documents the deleted path (round 2, still live)

```
$ grep -n 'spillover\|permanent' q-system/.q-system/scripts/test/test-findings-block-reader.sh
103:# extract_minor_findings feeds `spillover add`, which writes PERMANENT ledger
126:      permanent spillover item holding the standing gate red. Captured:
130:ok "only the live block's minors are captured (withdrawn findings do not reach the ledger)"
```

`spillover add` is gone from that path in this PR. The comment and the assertion label now describe code that no longer exists.

## What is sound

- **False positives on the widened regex.** Ran `scans_the_tree` at `ed14f6e5^` and at HEAD across all 280 declared tests. Floor 92 → 94, exactly the two tests named in the PR body, **zero lost**. The widening is as narrow as claimed.
- **The `rg` boundary claim.** Only hit in the corpus is `test_change_size.py` itself, which is the boundary assertion, not an invocation. Claim holds.
- **`grep -R`, `grep --recursive`, `grep --include`.** Zero hits across 280 declared tests. Round 2's minor 2 is real as regex theory and empty as corpus fact.
- **The `_EmptySelector` fixture's shape against the real producer.** `plan_for_repo` emits `app_lines, changed_files, declared_tests, full_suite, reasons, selected_tests, tier`. The fixture omits only `changed_files`, which `select_for_diff` never reads. Faithful.
- **False-alarm hunt on the new `ran == 0` error.** I tried to make it cry wolf at 3am. In instance mode `skipped` is 21 (skeleton-only) on every run, so the branch is unreachable there. `kipi-update.sh:3017` and `fleet-capability-verify.py:50` both pass `--check-only`, which never calls `run_tests`. Every road I found to `ran == 0` already carries a real error (missing files, timeouts, zero-execution). No false page.
- **Round 1's major.** `prd_runner.py:2448` reads `(args.severity or "minor")`, `SPILLOVER_REFUSED_SEVERITIES` is `("minor","low","nit")`. The deletion is right and the "0 by construction" diagnosis is right.
- **`test-severity-floor.sh`**, the neighbour declaring `covers` on `pr-review-agent.sh`: **PASS: 205/205**, rc=0. `test-minor-capture-signal.sh`: ALL PASS, mutant killed. `test_change_size.py`: 52 passed.

**One finding I dropped.** The test's `"a policy refusal pages nobody"` assertion has no mutant behind it, unlike its sibling. I tried three times to build a paging mutant and run it; this sandbox's command validator blocked each construction. Without an executed repro I am not raising it, and reading the source says it *is* reachable (the block would call `bash "$NOTIFY"`, and the stub records). Dropped.

The head has not moved since round 2. Commits stop at `c374f1f3`; `7297cc44` is a merge of main. Round 2 returned APPROVE WITH NITS, which is terminal, so its three minors got no rework pass and two of them are findings 2 and 3 above. That is this PR's own UNROUTED message describing this PR. The loop is working as documented, and the documentation is the problem it names.

No escalation happened across rounds: major (r1) → 3 minors (r2) → 3 minors (r3), one of them new. Calibration looks fine.

## VERDICT: APPROVE WITH NITS

Fix first: `change-size.py:184` either learns the subprocess-list spelling, or `test_change_size.py:198` drops the "occurs nowhere" claim and names `test_lessons_recall_corpus.py:171` as a known-blind scanner. A wrong why-comment is the exact failure this PR was opened to fix.

FINDINGS:
minor|scans_the_tree misses the subprocess-list spelling of recursive grep, so declared test test_lessons_recall_corpus.py is off the always-run floor, and the new comment claims that form occurs nowhere in the corpus|q-system/.q-system/scripts/change-size.py:184
minor|the UNROUTED message states the minors are on the PR and the Linear issue, but it prints outside and before the POST=1 block, so a dry run asserts a posting that never happened|q-system/.q-system/scripts/pr-review-agent.sh:1237
minor|test-findings-block-reader.sh still documents and labels extract_minor_findings as feeding spillover add and writing permanent ledger items, a path this PR deleted|q-system/.q-system/scripts/test/test-findings-block-reader.sh:103
END FINDINGS

PR #376, opened 2026-09-20
fix(converge): an approval survives converge's own receipt commit (ASK-1888)

assafkip:
## What was broken

converge writes the prd-os receipt as a commit on the approved branch, 4 to 11 seconds after the reviewer posts `kipi/reviewer-approved=success`. A commit status is per-sha. The approval stayed on the reviewed sha and the head GitHub gates had none, so auto-merge waited forever.

Measured 2026-09-19 (gh + the local verdict records):

- 31 approved `sana/` PRs red on their current head. 25 behind exactly this receipt commit. 30 with auto-merge armed.
- 9 budget days at 10 dispatches a day: 32 PRs opened, 1 merged.

## The fix

`receipt-carry-approval.sh` copies the reviewer's success onto origin's head when, and only when, all three hold:

1. GitHub still shows a live `success` at the reviewed sha. The floor's red on top of it is not a verdict. The local verdict record is never the source.
2. The head descends from the reviewed sha and the only path that differs is `.prd-os/receipts.jsonl`. Same rule as `receipt_only_ahead` and the receipt gate's own ledger allowance.
3. The head carries no real reviewer verdict yet. A head the reviewer already judged is never overwritten.

It has no verdict of its own to give. `reviewer-floor.sh` is untouched and still cannot post success. After this PR exactly two things can write `success` into that context: `post_reviewer_status` and `carry_post`.

converge calls it from `receipt_confirm_origin`, the one place origin is known to carry the receipt, so a fresh push and a re-run are both covered. With no owner/repo slug it posts nothing (the ASK-738 scar: `gh api` resolves placeholders from cwd).

## Proof

`bash q-system/.q-system/scripts/test/test-receipt-carry-approval.sh`

- Red before the script existed: 14 failing.
- Now 27 of 27, run 3 times.
- Fixtures are four real status payloads captured from PRs #370 and #372.
- 4 mutants killed: each of the three guards removed, and converge with the call deleted.
- Neighbour `test-converge.sh`: 21 of 21.
- `capability-gate.py` reads RED (12) on this branch and RED (12) on a clean `origin/main` checkout. None of the 12 name a file in this PR.

## Not in this PR

- The 25 PRs already stuck. They need a one-time back-fill run of the new script, after this lands.
- The 6 approved PRs whose head moved by a merge of main. Those need a real re-review; that is the floor working.
[...]

PR #393, opened 2026-09-20
fix(carry): the page is pasteable, names the right remedy, and claims only what the order earns (ASK-1905)

assafkip:
Three review nits from PR #376 (ASK-1888), held back at the time because pushing them would have moved the head and cancelled the approval, which is the exact defect that PR fixes. They ship here as the follow-up. Closes ASK-1905.

## What changed

**Nit 1, the remedy was not pasteable.** `approval_confirm`'s `CARRY_FIX` carried a literal `<pr>` while the PR number was in scope two frames up, so the operator had to go find it before running the command the page handed them. The number is now threaded `receipt_ensure` to `receipt_confirm_origin` to `approval_confirm` as an argument, rather than read off the `$PR` global: a page that is right only because a caller happened to leave a global set is right by accident.

**Nit 2, "could not ask" is not "the answer was no".** An unreadable GitHub response was paged as "the head carries no live reviewer approval" and prescribed a full re-review. `head_state` exits 1 rather than printing `none` precisely so those two stay distinguishable, and `approval_confirm` collapsed them back into one sentence. `state: unreadable` now gets its own claim and its own remedy: check `gh auth status`, check the network, re-run converge. Re-reviewing a PR fixes nothing about an expired token.

**Nit 3, the header over-claimed.** `receipt-carry-approval.sh`'s header said the staging-ref order removes the race entirely. It removes it on the NEW sha only: no verdict there can predate the copy. A REQUEST CHANGES landing on the REVIEWED sha after guard 1 read it is still buried, by converge's branch move rather than by this script. The header now states both halves, per the DoR's first option. The second option (re-read the reviewed sha before the branch move) was not taken: it changes the carry's order, which the issue's **Not doing** line forbids.

## Evidence

Reproducer first. Eight new cases against the shipped code, observed RED:

```
FAIL the header does not claim the copy can never bury a verdict (want '0', got '1')
FAIL it names what the order guarantees: no verdict on the NEW sha predates the copy
FAIL and it names the window that stays open: a refusal on the reviewed sha
FAIL the red-head remedy names the PR number, so it can be pasted
FAIL and carries no literal <pr> placeholder
FAIL an unreadable GitHub is NOT paged as a head carrying no approval
FAIL it says converge could not read the verdict, which is a different claim
FAIL and its remedy is auth plus network, never a re-review
passed 47, failed 8
```

After the fixes:

```
$ bash q-system/.q-system/scripts/test/test-receipt-carry-approval.sh
passed 58, failed 0
```

Baseline was 46 of 46, matching the DoR. Two new mutants, both killed: dropping the PR number on the way down, and collapsing `unreadable` back into the red-head sentence.

A regression guard rides along: "a genuinely red head still gets the re-review remedy" was green before and after, so nit 2's split did not take the re-review remedy away from the case that earns it.

## Also in here

`export KIPI_NOTIFY=/usr/bin/true` at the top of the test file. The mutation harness re-invokes the file against a MUTATED `converge.sh`, which is by construction a path nobody reasoned about, and converge resolves its pager through that variable. The `fable-discipline-lint` hook flagged the gap on the first edit. Scar it cites: 2026-08-01, a suite reporting 14/14 green paged the founder twice.

## After merge

`sp-3f68bde6` (nit 3) plus the two sibling rows under source ASK-1888 resolve against this issue.

## Not doing

The carry's guards and its order are untouched, per the DoR.


assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-393-20260920-110103.md` (5228 bytes).

```
FINDINGS:
minor|The operator-facing merge page still asserts a red head and prescribes a re-review when converge could only fail to read GitHub|q-system/.q-system/scripts/converge.sh:1073
minor|The new header names a refusal window that guard 2 already closes, and the new test pins that imprecise phrase|q-system/.q-system/scripts/receipt-carry-approval.sh:46
END FINDINGS
```

--- reviewer output ---

## What I found

Two minors survived reproduction. Both green suites, no blocker, no major.

---

### minor 1 — the page a human actually reads still collapses "could not ask" into "the head is red"

`q-system/.q-system/scripts/converge.sh:1073` (and `:1072` for the log line)

Nit 2 fixes `approval_confirm`'s own `CARRY_MISS` and `CARRY_FIX`. The string that reaches `bash "$NOTIFY"` is composed in the caller, and that caller hard-codes the red-head claim plus the re-review prescription around whatever `CARRY_MISS` holds.

Reproducer. `/usr/bin/false` stands in for `gh`, which is the expired-token case, so `head_state` exits 1 and converge sets `state=unreadable`. The two composition lines are grepped out of the shipped `converge.sh` and eval'd, so this cannot drift from production.

Real output:

```
== approval_confirm, the half this PR fixed ==
CARRY_MISS: converge could not read GitHub's verdict for origin's head 54a1cd13f7de (state: unreadable), so nothing here says whether it is approved or not
CARRY_FIX : gh auth status, then check the network, then re-run: kipi converge --issue ASK-1

== the line NOTIFY sends to a human ==
converge ASK-1: APPROVE after 1 round(s), PR #376 approved and its receipt landed, but
converge could not read GitHub's verdict for origin's head 54a1cd13f7de (state: unreadable),
so nothing here says whether it is approved or not. kipi/reviewer-approved is not green on
the head, so it will NOT merge by itself. Needs a review of that head: gh auth status, then
check the network, then re-run: kipi converge --issue ASK-1

FAIL page asserts the head is not green, where converge could not ask
FAIL page prescribes a review of that head for an auth/network failure
FAIL log asserts the head is red, where converge could not ask
exit=1
```

One sentence says "nothing here says whether it is approved or not" and the next says "is not green on the head". At 3am the operator reads "Needs a review of that head" and burns a review round on a dead token.

The four new tests read `CARRY_MISS` and `CARRY_FIX` directly, so none of them can see this. The value is decided in the function and shown in the caller, and the defect lives where it is shown.

---

### minor 2 — the new header names a window guard 2 already closes

`q-system/.q-system/scripts/receipt-carry-approval.sh:46-47`

The header says: "A refusal landing on the reviewed sha after guard 1 read it is invisible to this script."

Guard 1 is `delta_kind`. Guard 2 is the reviewed-sha status read at line 147. A refusal landing between them is caught:

```
SCENARIO the header now names: a REQUEST CHANGES lands on the REVIEWED sha
after guard 1 (delta_kind) read it. Header says it is 'still buried'.

carry: the live reviewer verdict at 8a063323fb2a... is 'failure', not an approval; posting nothing
exit=10
POSTs made: 0
```

The window that is genuinely open is after **guard 2**'s read, not guard 1's. `test-receipt-carry-approval.sh:73` greps for the phrase `refusal landing on the reviewed sha`, which pins the imprecise sentence in place.

Second half of the same line: "This script never buries that refusal, the branch move does, exactly as any push would." An ordinary push leaves the new head with no reviewer status, which blocks auto-merge. This script posts `success` onto it. Those are not the same outcome, and the header reads as if they are.

---

## What is sound

Attacks the code survived, all run:

- **Suite is real.** 58 of 58 green, 8 mutants killed including the two new ones. I ran it.
- **PR threading has one caller.** `grep` across `q-system/.q-system/scripts/` and `plugins/` finds no other `approval_confirm` or `receipt_confirm_origin` caller. `receipt_ensure "$SHA" ... "$PR"` at `converge.sh:1029` is the only source, so the `${3:-}` default is never the live path.
- **`state=unreadable` cannot be forged by GitHub.** `live_verdict` emits only GitHub's own state values, never the literal `unreadable`, so the new branch cannot swallow a real verdict.
- **I tried to make the header test vacuous.** Padded the file 25 lines so the honest sentences fall past the `sed -n '1,60p'` window. The negative assertion does go vacuously green, but both positive assertions go RED and the suite fails. Not a silent pass, so I dropped it.
- **No regressions.** `bash -n converge.sh` clean, `test-converge.sh` 21 of 21, `test-converge-crossrepo-receipt.sh` 3 of 3.
- **`export KIPI_NOTIFY=/usr/bin/true`** matches `converge.sh:67` (`NOTIFY="${KIPI_NOTIFY:-...}"`). The stub is real, not decorative.

---

## VERDICT: APPROVE WITH NITS

**Fix first:** `converge.sh:1073`. Split the page the same way `approval_confirm` was split, so the `unreadable` case does not assert a red head and does not say "Needs a review of that head". Right now the nit-2 fix stops one frame short of the only surface that pages a human.

FINDINGS:
minor|The operator-facing merge page still asserts a red head and prescribes a re-review when converge could only fail to read GitHub|q-system/.q-system/scripts/converge.sh:1073
minor|The new header names a refusal window that guard 2 already closes, and the new test pins that imprecise phrase|q-system/.q-system/scripts/receipt-carry-approval.sh:46
END FINDINGS

PR #395, opened 2026-09-20
fix(voiceloop): repair a model-introduced capital mid-sentence (ASK-1943)

assafkip:
## The scar

`check_capitalization` and `post_repair`'s step-2 loop both walk
`_sentence_start_offsets`. Both can therefore only ever see the FIRST word of a
sentence. A capital the model introduces mid-clause is structurally invisible to
every layer in the stack.

Four consecutive reddit reply runs shipped eleven of them past a clean gate:
"the part That would worry me", "treats The chart", "which is why They came back
mixed". The trail carried `repairs: null` and `revisions: 0`, so nothing in the
system observed any of it. The operator caught all four runs by eye.

The feature was built and measured in a long-running instance checkout and was
never committed. `plugins/kipi-core/voiceloop/` is fleet-synced FROM this repo,
so the next sync would have reverted it. That is why it lands here.

## What it adds

`MID_SENTENCE_LOWERCASE_WORDS`, `mid_sentence_cap_hits`,
`repair_mid_sentence_caps`, wired LAST in `repair()`'s pure-transform group so it
sees the emdash and contraction passes' output and still precedes the proper-noun
safety net.

It REPAIRS and never blocks. Nothing here reaches `violations()`: a mid-sentence
capital costs a lowercase letter, never a draft.

The rule is a CLOSED WORD SET, not a grammar, with three narrowing constraints:

1. the word is in the set (determiners, pronouns, relatives),
2. the preceding character is a lowercase letter, a digit or a comma on the SAME
   line,
3. a following capitalized word means a title-case run, and the run is left alone.

## Re-measured, not quoted

The prior comment's numbers were re-run against THIS tree's engine on 2026-09-20
rather than carried across. Corpus read read-only from a snapshot; the engine and
the linter are this branch's.

| Claim in the original comment | Re-measured here |
|---|---|
| zero false positives across the corpus | **reproduces.** 0 detector hits, 0 repair edits |
| 10 of 12 live defects caught | **reproduces.** 10 caught, same two misses (`run3-d`, `run4-a`), every per-fixture expectation exact |
| four words removed because they hit his writing | **reproduces.** adding back `as`, `have`, `and` or `it` each puts exactly 1 corpus hit on the board. Two more (`your`, `its`) came out in round 2 for the same class |
| "171 active corpus items" | **does NOT reproduce. 162 today**, 149,558 chars |

On the one number that moved: the corpus is live and has changed since
2026-09-10. Today 161 of 161 rows in the exemplar file are `active`, all 8 rows in
the second named file are `staged` and so contribute nothing under the active
filter, plus 1 flat file, giving 162. The claim "across all 171 active items in
these three files" overstated one of the three. The comment now carries my number
and my date. No decision in the PR rests on the difference; the zero is what the
word-list scar asks for and the zero holds.

## Round 2: the review found a major and it was real

The adversarial reviewer returned REQUEST CHANGES on the first head with
executed reproducers. I reproduced both majors myself before acting. Both hold.

**Finding 1 (major, fixed).** The pass was wired into `repair()`'s group 0,
before the sentence-start pass. Step 2 rewrites with `subn(..., count=1)`
against the raw text, so it hits the FIRST standalone lowercase occurrence of
its target word, not necessarily the sentence-start one. A draft opening a later
sentence with that word got the capital put back:

```
in   it treats The chart as disposable. the fix is simple.
out  It treats The chart as disposable. The fix is simple.
chg  ['lowercased mid-sentence capital (The) x1', ...]
```

The defect shipped while the trail claimed it was repaired. A false receipt is
worse than the `repairs: null` silence this layer exists to end. Every
`repair()`-level fixture in the suite opened with a capital, so step 2 found no
lowercase sentence start and the shape never arose. Fixed by calling the pass
after step 2 and before step 3, which keeps both properties the group-0 position
was chosen for. Held by `test_the_pass_runs_after_the_sentence_start_pass`.

**Finding 2 (major, NOT fixed here, captured).** Step 2 is itself a producer of
this defect class on ordinary lowercase-start prose, with no model involved:

```
in   the agent keeps the chart open all day. the fix is a plugin.
out  The agent keeps The chart open all day. The fix is a plugin.
```

That is pre-existing, outside this diff, and it changes the attribution of the
scar: the eleven observed capitals were blamed on the model and at least some of
that shape is manufactured by the repairer. Captured as `sp-dc022444` / ASK-1945.
Moving the pass after step 2 does mitigate it for words in the closed set, which
is why the finding-2 shapes now come back clean, but step 2 can still plant a
capital the set does not cover. The real fix is for step 2 to splice by located
offset instead of first-match `count=1`.

**Finding 4 (minor, fixed).** `your` and `its` left the word set. "Hey man, Your
invoice is attached" is a salutation with no terminator, so the word after the
comma is a sentence start the linter's offsets cannot see. `dm` and `email` are
default channel scopes. The corpus scan could not refute these two because the
corpus holds posts, not one-line DMs, which is the honest limit of a corpus gate:
it can only refute a word with text somebody already wrote. Set is now 21 words,
re-measured clean.

**Finding 3 (minor, fixed).** `test_it_repairs_and_never_blocks` passed with the
feature stubbed to identity. Both assertions were unfalsifiable. Rewritten.

**Finding 5 (minor, fixed).** The `text or ""` guard was unreachable;
`_protected_spans` raised `TypeError` on `None` first.

**Finding 6 (nit, fixed).** A test docstring's stated rationale was false.

One correction to my own PR body: I first wrote that `voiceloop/tests/` has no
CI runner, having checked `validate.yml` and `lefthook.yml` and not `verify.yml`.
The reviewer was right. `plugins/kipi-core/voiceloop` is line 71 of
`.verify-suites`, which `verify.sh --full` runs in the `verify` workflow on every
PR. The spillover item I filed on that is voided with the reason recorded.

## Rounds 3 to 5

The reviewer ran again on each new head. Every round found something real; the two
that matter most were caught by MUTATION, not by reading.

**Round 3 (approve with nits), two honesty defects in my own test file.**
`test_the_layer_runs_after_the_passes_that_CREATE_its_shape` claimed a reorder
turns it red. Measured: moving the call ahead of `repair_contractions` turns the
two ORDERING tests red and leaves that one green. The docstring now says which
test holds what. Three more docstrings still named a digit predecessor round 2
had removed, and the wiring test still said the call lives in a tuple it had left.

**Round 4 (request changes), one major, fixed.** Constraint 3 only looked FORWARD,
so the final in-set word of a title-case run was lowercased: `## The Dedup Key The
rest`. A capital immediately before is the same title-case evidence as one after.

**The round-4 review's own claim about its fix did not hold, and checking it is why
that commit has a second half.** The review stated a preceding-capital test
"excludes none of the ten caught defects; checked". It excludes run3-b: `It's That
the agent cannot tell the two apart` has a capitalized `It's` in front, so the
naive guard took the rate 10 -> 9 and three tests went red on the recorded defect
list. `_opens_a_sentence` makes the guard skip a preceding capital that is first
on its line or follows a terminator, which keeps all the heading cases clean and
keeps run3-b caught.

**Then mutation caught that the new guard was untested.** Deleting the entire
backward half left the suite GREEN (M10 SURVIVED). All three fixtures I had
written also had a capitalized word AFTER the in-set word, so the forward half was
already holding every one of them. `## The Dedup Key The Rest` is held by `Rest`,
not by `Key`. Replaced with four shapes whose following word is lowercase or
absent, so only the backward half can hold them. Both mutants now kill.

A guard whose fixtures are all covered by a different guard reads exactly like a
guard that works. The mutation run is the only thing that said otherwise.

## Mutation results

Each of the three constraints broken in turn, plus the wiring. Every mutant was
validated as APPLIED (sha and byte count both change) before its run was read.
`PYTHONDONTWRITEBYTECODE=1` throughout. Restored from the commit after each, and
the tree ends on an empty `git status`.

| Mutant | Applied | Result |
|---|---|---|
| M1 drop the closed-word-set check | 28639 -> 28557 bytes | **RED, 14 failed** |
| M2 widen the predecessor to any same-line char | 28639 -> 28636 | **RED, 7 failed** |
| M3 disable the title-case-run guard | 28639 -> 28595 | **RED, 3 failed**, exactly the three title-case-run cases |
| M4 remove the call from `repair()` entirely | 28639 -> 28550 | **RED, 5 failed** |
| M5 move the call back into group 0 (the finding-1 regression) | 28639 -> 28576 | **RED, 2 failed** |
| M6 restore the unreachable empty-body guard (finding 5) | 28639 -> 28604 | **RED, 1 failed** |
| M7 restore the dotted-name hole (`(?!\.\w)` removed) | 29623 -> 29615 | **RED, 2 failed**, the two dotted-brand cases |
| M8 restore the digit predecessor | 29623 -> 29626 | **RED, 2 failed**, the two numbered-heading cases |
| M9 run the layer before the contraction pass | 29623 -> 29560 | **RED, 2 failed**, the two ordering tests |
| M10 remove the backward half of constraint 3 | 31682 -> 31470 | **SURVIVED first, then RED, 4 failed** after the fixtures were fixed |
| M11 make `_opens_a_sentence` always False (the naive guard) | 31682 -> 31547 | **RED, 4 failed** |

Two harness notes, both worth recording because both were failure modes the
discipline caught rather than things that went smoothly:

- A mutant with a bad anchor did NOT apply. The harness reported "MUTANT NOT
  APPLIED -- this run proves NOTHING" and refused to read the result. A killed
  count is meaningless until applied is proven.
- The first harness restored with `git checkout --`, which restores to HEAD, not
  to the working state. Running it over uncommitted fixes silently reverted them
  and the following mutants then ran against the wrong code. The round-2 harness
  commits first and restores from a `cp` backup, and the tree is proven clean at
  the end.

## The half that deliberately does not live here

The corpus scan is the load-bearing check on a word list, and it CANNOT run in
this repo: kipi-system is public and the corpus is the operator's. The engine
comment was rewritten for that reason. Its original form named the corpus
filenames and quoted four fragments of his approved writing; `test_no_founder_data.py`
in this same tree went RED on the carry, correctly, and the quotes were a leak by
the same rule even where the guard's word list did not name them.

So the split is: this tree holds the mechanism and its fixtures, the instance
holds the counterexample gate against its own corpus. The comment says so, rather
than claiming a scan that is not here.

## Tests

- `plugins/kipi-core/voiceloop/tests/test_mid_sentence_caps.py`, 61 tests, new.
  Every constraint gets a named true positive AND a negative self-test proving the
  guard is a narrowing and not the whole rule.
- `pytest plugins/kipi-core/voiceloop/ -q` -> **309 passed** (248 before this PR).
- `pytest plugins/prd-os/tests/ -q` -> **709 passed, 1 skipped**.
- `pytest q-system/.q-system/tests/test_auto_commit.py -q` -> **46 passed**.
- `validate-separation.py 1` -> **PASS 46, FAIL 0, WARN 1** (with the CI capability-gate skip).
- `covers-lint.py` -> PASS.
- The new suite also passes under `env -i` with a fresh `HOME`, so it does not
  depend on local git identity or home state.

## Known gap, not introduced here

`capability-gate.py` reports RED (12) locally on this branch. It reports the
IDENTICAL 12 on a clean `origin/main` worktree, all of them `declared_inert`
scripts citing resolved spillover ids, none touching voiceloop. Pre-existing.

## Captured, not fixed here

- `sp-dc022444` / ASK-1945 (major): step 2 is itself a producer of this defect
  class. Pre-existing, and it changes the attribution of the scar.
- `sp-5468add9` / ASK-1946: two rule-scope questions (a comma or `and` before an
  in-set word defeats the forward guard; the salutation class is still live for
  the other 21 words). Both want a scan over a DM and email corpus, which is the
  population the post corpus structurally cannot see. Not patched by ad-hoc word
  removal before that instrument exists.

Closes ASK-1943.


assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-395-20260920-112336.md` (8949 bytes).

```
FINDINGS:
major|repair_mid_sentence_caps runs before step 2, which re-capitalizes the word it just lowercased, so the defect ships while changes[] reports it repaired|plugins/kipi-core/voiceloop/post_repair.py:377
major|Step 2's first-match count=1 rewrite plants mid-sentence capitals into ordinary lowercase-start prose, making the repairer itself a producer of the defect class this PR attributes to the model|plugins/kipi-core/voiceloop/post_repair.py:411
minor|test_it_repairs_and_never_blocks passes with the whole feature stubbed out; neither assertion can fail|plugins/kipi-core/voiceloop/tests/test_mid_sentence_caps.py:270
minor|A capital after a comma in a one-line salutation is lowercased, and dm/email are default channel scopes|plugins/kipi-core/voiceloop/post_repair.py:153
minor|The `text or ""` guard is unreachable because _protected_spans raises TypeError on None first|plugins/kipi-core/voiceloop/post_repair.py:318
nit|Test docstring claims deleting the predecessor check leaves the prior test green; it turns that test red|plugins/kipi-core/voiceloop/tests/test_mid_sentence_caps.py:199
END FINDINGS
```

--- reviewer output ---

# Review: PR #395 — voiceloop mid-sentence capital repair

Read the diff, read `post_repair.repair()` whole, ran the new suite (45 passed) and the full voiceloop suite (293 passed), then attacked the ordering and the word set with executed reproducers.

---

## Finding 1 — major

**The new pass runs before step 2, and step 2 puts the capital back. The change log reports a repair that did not survive.**

`plugins/kipi-core/voiceloop/post_repair.py:377`

`repair_mid_sentence_caps` is wired last in group 0. Step 2 (the sentence-start pass) runs after it and rewrites with `pattern.subn(..., count=1)` against the raw text, which hits the FIRST standalone lowercase occurrence of the target word, not the sentence-start one. When a later sentence opens with the same lowercase word, step 2 re-capitalizes the word the new layer just fixed.

Reproducer (run6-b from the PR's own measured defect list, given an ordinary second sentence):

```
cd <worktree> && python3 -c '
import os, sys
ROOT = os.getcwd()
sys.path.insert(0, os.path.join(ROOT, "plugins", "kipi-core"))
from voiceloop import post_repair
lint = post_repair._load_linter(os.path.join(ROOT, "q-system", ".q-system", "scripts", "voice-lint.py"))
t = "it treats The chart as disposable. the fix is simple."
out, ch = post_repair.repair(t, allowlist=frozenset(), linter=lint, mapping={})
print("OUT :", out); print("CHG :", ch)
print("residual:", post_repair.mid_sentence_cap_hits(out, lint))'
```

Real output:

```
OUT : It treats The chart as disposable. The fix is simple.
CHG : ['lowercased mid-sentence capital (The) x1', "capitalized sentence start 'it'", "capitalized sentence start 'the'", "capitalized sentence start 'the'"]
residual: [(10, 'The')]
```

The defect ships. The trail says it was repaired. That is the exact failure mode the PR opens with (`repairs: null` and nothing observing it), now with a false receipt instead of silence.

Every `repair()`-level fixture in the new test file opens with a capital letter, so step 2 finds no lowercase sentence start and the shape never arises. `test_the_full_repair_pass_reaches_the_new_layer` at `tests/test_mid_sentence_caps.py:250` uses `"The reason The labels move is the rebuild."` — the one input the ordering already passes.

Fix shape: move `repair_mid_sentence_caps` after step 2, or make step 2 splice by located offset instead of first-match `count=1`.

---

## Finding 2 — major

**Step 2 is itself a producer of mid-sentence capitals, on ordinary founder prose. Pre-existing, not in the diff, but it is the dominant source of the class this PR exists to close.**

`plugins/kipi-core/voiceloop/post_repair.py:411`

Same mechanism, no model involved: a draft with lowercase sentence starts (the founder's normal register) gets a capital planted mid-clause.

```
cd <worktree> && python3 -c '
import os, sys
ROOT = os.getcwd()
sys.path.insert(0, os.path.join(ROOT, "plugins", "kipi-core"))
from voiceloop import post_repair
lint = post_repair._load_linter(os.path.join(ROOT, "q-system", ".q-system", "scripts", "voice-lint.py"))
for d in ["the agent keeps the chart open all day. the fix is a plugin.",
          "nobody reads the summaries. the leftovers are what break you. they came back mixed."]:
    out, ch = post_repair.repair(d, allowlist=frozenset(), linter=lint, mapping={})
    print("OUT:", out)
    print("INJECTED:", post_repair.mid_sentence_cap_hits(out, lint))'
```

Real output:

```
OUT: The agent keeps The chart open all day. The fix is a plugin.
INJECTED: [(16, 'The')]
OUT: Nobody reads The summaries. The leftovers are what break you. They came back mixed.
INJECTED: [(13, 'The')]
```

`Nobody reads The summaries` is the same shape as the reported live defect `treats The chart`. The PR's scar section attributes these to the model. The code says the repairer manufactures them. Worth checking the eleven reported hits against this before accepting the "model-introduced" framing.

---

## Finding 3 — minor

**`test_it_repairs_and_never_blocks` passes with the entire feature deleted.**

`plugins/kipi-core/voiceloop/tests/test_mid_sentence_caps.py:270`

Assertion 1 filters `check_capitalization` output for the substring `"mid-sentence"`, which that checker never emits. Assertion 2 asserts the repaired text is clean, and the unrepaired text is already clean by that checker.

Mutation, feature stubbed to identity:

```
cd <worktree> && python3 -c '
... post_repair.repair_mid_sentence_caps = lambda text, linter: (text, [])
text = "The reason The labels move is the rebuild." ...'
```

Real output:

```
surface: []
assert1 passes: True
repaired: The reason The labels move is the rebuild.
assert2 passes: True
```

---

## Finding 4 — minor

**A capital after a comma inside a salutation is lowercased; `dm` and `email` are default channel scopes.**

`plugins/kipi-core/voiceloop/post_repair.py:153`

```
HIT  [(9, 'Your')] | Hey man, Your invoice is attached and the deck is done
                   -> Hey man, your invoice is attached and the deck is done
```

`channel_registry.py:54` lists `"dm", "email"` in `DEFAULT_SCOPES`, so a one-line DM or email is in range. Same battery also lowercased `Every.to`, `Which magazine`, and `the band Them`, but I constructed all of those and no producer in this repo emits them, so they are noted, not filed.

---

## Finding 5 — minor

**`text or ""` at line 320 is unreachable; `_protected_spans` at line 318 crashes on `None` first.**

`plugins/kipi-core/voiceloop/post_repair.py:318`

```
CRASH: TypeError expected string or bytes-like object, got 'NoneType'
```

The guard states an intent the function does not have.

---

## Finding 6 — nit

**A test docstring's stated rationale is false.**

`plugins/kipi-core/voiceloop/tests/test_mid_sentence_caps.py:199` claims deleting the predecessor check "would leave the test above passing". The predecessor check is the `(?<=[a-z0-9,])` lookbehind; deleting it turns `test_a_sentence_start_capital_is_never_touched` red on `"It broke. The chart rebuilt itself."` The negative self-test is still worth keeping, its justification is not.

---

## What is sound

Attacks the code survived, each executed:

- **Product and proper nouns.** `Google Sheet`, `PowerPoint`, `FX`, `SKU`, `PDF`, `Etsy and Square`, `JavaScript`, `Python` — zero hits. Constraint 1 is real narrowing, and the `[A-Z][a-z']+` shape makes acronyms and bare `I` structurally unreachable rather than allowlisted out.
- **Title-case runs.** `The New York Times`, `The Garden`, `The Home Depot` untouched; the paired negative (`The paper`) is caught. The guard is load-bearing, not decoration.
- **Code protection.** Inline backticks and fenced blocks are skipped via the linter's own regexes, so gate and repairer agree on what is code.
- **Offset safety.** Reverse-order splicing verified; no corruption, and the detector is idempotent on its own output.
- **Boundaries.** Newline, colon, quote, list marker, parenthesis, period predecessors all skipped. Non-ASCII predecessors are a false negative, not a corruption.
- **Concurrency and 3am.** Pure function, no disk, no state, repair-never-block. Rollback is a revert of two files.
- **Wiring.** `plugins/kipi-core/voiceloop` is a named entry in `.verify-suites`, so the new file actually runs in CI. Full suite green: 293 passed.

---

## VERDICT: REQUEST CHANGES

Not BLOCK — nothing here is unrecoverable, and the layer is strictly additive in the shapes where it works.

**Fix first:** finding 1. Move `repair_mid_sentence_caps` after step 2 (post_repair.py:377) and add a `repair()`-level fixture whose draft has a lowercase sentence start repeating the repaired word. Right now the change log says the capital was fixed while the capital is still in the shipped text, which is worse than the silence the PR set out to end.

FINDINGS:
major|repair_mid_sentence_caps runs before step 2, which re-capitalizes the word it just lowercased, so the defect ships while changes[] reports it repaired|plugins/kipi-core/voiceloop/post_repair.py:377
major|Step 2's first-match count=1 rewrite plants mid-sentence capitals into ordinary lowercase-start prose, making the repairer itself a producer of the defect class this PR attributes to the model|plugins/kipi-core/voiceloop/post_repair.py:411
minor|test_it_repairs_and_never_blocks passes with the whole feature stubbed out; neither assertion can fail|plugins/kipi-core/voiceloop/tests/test_mid_sentence_caps.py:270
minor|A capital after a comma in a one-line salutation is lowercased, and dm/email are default channel scopes|plugins/kipi-core/voiceloop/post_repair.py:153
minor|The `text or ""` guard is unreachable because _protected_spans raises TypeError on None first|plugins/kipi-core/voiceloop/post_repair.py:318
nit|Test docstring claims deleting the predecessor check leaves the prior test green; it turns that test red|plugins/kipi-core/voiceloop/tests/test_mid_sentence_caps.py:199
END FINDINGS

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-395-20260920-113635.md` (10761 bytes).

```
FINDINGS:
minor|A dotted brand name clears all three constraints and is lowercased (`Every.to` to `every.to`), the exact failure the module header calls worse than the bug|plugins/kipi-core/voiceloop/post_repair.py:160
minor|Constraint 2's digit predecessor admits a numbered section prefix, so `## 2.1 The dedup key` loses its capital|plugins/kipi-core/voiceloop/post_repair.py:160
minor|The detector reads raw text while the gate reads the linter's prose view, so blockquoted third-party words and table headers get rewritten|plugins/kipi-core/voiceloop/post_repair.py:330
minor|The salutation class that removed `your` and `its` is still live for the other 21 words, and the test only parametrises the two removed ones|plugins/kipi-core/voiceloop/post_repair.py:148
minor|The 2b comment names step 3's proper-noun pass as its safety net, but `load_proper_nouns("")` returns an empty list in every tree|plugins/kipi-core/voiceloop/post_repair.py:441
END FINDINGS
```

--- reviewer output ---

## Findings

All five reproduced end to end through the full `repair()` pipeline on this branch. Baseline first: `pytest plugins/kipi-core/voiceloop/tests/test_mid_sentence_caps.py -q` gives `50 passed`, and the whole package gives `298 passed`. Nothing below is a broken test.

Reproducer for findings 1, 2 and part of 3, run from the worktree root:

```
python3 -c 'import os,sys; sys.path.insert(0,"plugins/kipi-core")
from voiceloop import post_repair as pr
l=pr._load_linter("q-system/.q-system/scripts/voice-lint.py")
for t in ("he writes for Every.to and it shows\n",
          "## 2.1 The dedup key is the whole fix\n\nit shipped on friday.\n",
          "| Component | Loads When |\n|---|---|\n| a | b |\n",
          "he ran step 2 The chart rebuilt itself\n"):
    out,ch = pr.repair(t, allowlist=frozenset(), linter=l, mapping={})
    print(repr(t)); print(" ->", repr(out)); print("   ", ch)'
```

Real output:

```
'he writes for Every.to and it shows\n'
 -> 'He writes for every.to and it shows\n'
    ["capitalized sentence start 'he'", 'lowercased mid-sentence capital (Every) x1']
'## 2.1 The dedup key is the whole fix\n\nit shipped on friday.\n'
 -> '## 2.1 the dedup key is the whole fix\n\nIt shipped on friday.\n'
    ["capitalized sentence start 'it'", 'lowercased mid-sentence capital (The) x1']
'| Component | Loads When |\n|---|---|\n| a | b |\n'
 -> '| Component | Loads when |\n|---|---|\n| a | b |\n'
    ['lowercased mid-sentence capital (When) x1']
'he ran step 2 The chart rebuilt itself\n'
 -> 'He ran step 2 the chart rebuilt itself\n'
    ["capitalized sentence start 'he'", 'lowercased mid-sentence capital (The) x1']
```

Blocking-violation count before and after was `1 -> 0`, `1 -> 0`, `0 -> 0`, `1 -> 0`. So every one of these ships. The repair is silent in the sense that matters: the gate says the draft is clean.

**minor. A dotted brand clears all three constraints and loses its capital.** `plugins/kipi-core/voiceloop/post_repair.py:160`. `(?![\w'])` treats a following `.` as a word boundary, so `Every.to` becomes `every.to`. The module header sets the bar itself: "his own approved copy carries 'Etsy and Square', 'Google Sheet', 'PowerPoint', 'FX', 'SKU'. A repair that lowercases one of those is worse than the bug." I did not invent this input. It came out of the repo's own prose, `q-system/output/plans/claudesidian-canonical-brief-2026-06-19.md`, where the text is "featured on Every.to".

**minor. The digit predecessor admits a numbered section prefix.** Same line, `post_repair.py:160`. Constraint 2 says a list marker is excluded, and `3. The chart` is, because `.` is not in the lookbehind. But `2.1 The dedup key` ends on a digit, which the lookbehind explicitly admits, so a numbered heading loses its capital. Also from the repo's own prose, `q-system/output/plans/linear-sdlc-standard-2026-07-26.md`, "### 2.1 The dedup key is the whole".

**minor. The detector reads raw text where every other capitalization layer reads the linter's prose view.** `plugins/kipi-core/voiceloop/post_repair.py:330`. `_protected_spans` covers frontmatter, fences and inline code. `strip_code_preserving_lines` also removes `BLOCKQUOTE_RE`, and `_sentence_start_offsets` also skips `TABLE_ROW_RE`. Neither is in `_protected_spans`, so this pass edits both. The table-header case is in the run above. The blockquote case:

```
python3 -c 'import os,sys; sys.path.insert(0,"plugins/kipi-core")
from voiceloop import post_repair as pr
l=pr._load_linter("q-system/.q-system/scripts/voice-lint.py")
t="> The chart rebuilds and They come back mixed.\n\nThat is the bug. the fix is a plugin.\n"
print(repr(pr.repair(t, allowlist=frozenset(), linter=l, mapping={})))'
```

```
('> The chart rebuilds and they come back mixed.\n\nThat is the bug. The fix is a plugin.\n', ["capitalized sentence start 'the'", "capitalized sentence start 'the'", 'lowercased mid-sentence capital (The, They) x2'])
```

A reddit reply that quotes its parent gets the quoted person's words altered. Dropping this a level and saying so: `_protected_spans` predates this PR, and nothing in `plugins/kipi-core/voiceloop/` emits a `> ` line into a draft, so the producer here is a model writing reddit syntax, not repo code. Over all 353 markdown files under `q-system/` (1,719,717 chars) the detector fires 4 times and all 4 are false positives: two table headers, one dotted brand, one numbered heading. The PR's headline number is "ZERO detector hits". Against this repo's own prose the rate is 4 hits, 4 wrong.

**minor. The salutation class was fixed for two words out of twenty-three.** `plugins/kipi-core/voiceloop/post_repair.py:148`. Round 2 removed `your` and `its` because "Hey man, Your invoice is attached" is a sentence start the linter's offsets cannot see. The same shape with `that`, `these` or `their` is untouched:

```
python3 -c 'import os,sys; sys.path.insert(0,"plugins/kipi-core")
from voiceloop import post_repair as pr
l=pr._load_linter("q-system/.q-system/scripts/voice-lint.py")
for t in ("Hey man, These are the numbers you asked for.\n","Hi Chris, Their invoice went out today.\n"):
    print(repr(pr.repair(t, allowlist=frozenset(), linter=l, mapping={})))'
```

```
('Hey man, these are the numbers you asked for.\n', ['lowercased mid-sentence capital (These) x1'])
('Hi Chris, their invoice went out today.\n', ['lowercased mid-sentence capital (Their) x1'])
```

`test_a_salutation_is_not_a_mid_clause` parametrises only over `your` and `its`, and `test_the_word_set_carries_no_product_or_proper_noun` asserts the six removed words are disjoint from the set, which passes by construction and cannot see the remaining class. The reproducer named the class and the fix removed two instances of it.

**minor. The comment names a safety net that returns an empty list.** `plugins/kipi-core/voiceloop/post_repair.py:441`. The 2b block says "Step 3's proper-noun pass still runs AFTER, so it remains the safety net if this layer ever lowercases the first word of a declared multi-word proper noun." Step 3 calls `linter.load_proper_nouns("")`, and `load_proper_nouns` resolves `Path(file_path).resolve().parent`, so with `""` it looks for `canonical/proper-nouns.txt` starting one directory ABOVE the process CWD. No such file exists anywhere in the repo:

```
python3 -c 'import sys; sys.path.insert(0,"plugins/kipi-core")
from voiceloop import post_repair as pr
print(pr._load_linter("q-system/.q-system/scripts/voice-lint.py").load_proper_nouns(""))'
```

```
[]
```

Plus `find . -name proper-nouns.txt` outside `.git` returns nothing. Step 3 is a no-op loop, so the net the comment leans on is not there. Step 3 predates this PR; the claim about it does not.

## What is sound

Attacks I ran that the code survived, named so the findings above can be weighed against them.

- **The round-2 major is genuinely fixed, and its regression test is real.** I traced step 2's `subn(..., count=1)` by hand on `it treats The chart as disposable. the fix is simple.` and watched the misplacement happen and then get cleaned up by 2b. `test_the_pass_runs_after_the_sentence_start_pass` goes red if the pass moves back to group 0. That is a check that can fail for the reason it claims.
- **"Repairs, never blocks" holds, and it holds structurally, not by luck.** I went looking for an input where this pass lowercases something `check_capitalization` then flags, which would turn the layer into the stall it exists to prevent. It cannot happen: `_sentence_start_offsets` only fires after `SENTENCE_END_RE` (`[.!?]` plus optional closers) or at a line start, and the pass requires `[a-z0-9,]` immediately before same-line whitespace. The two predecessor sets are disjoint. Every case I ran, including the four false positives, went `n -> 0` or `0 -> 0` blocking violations.
- **The title-case guard, the acronym exclusion and the skip list all do what they say.** `The New York Times` survives, `AI` / `SKU` / `FX` / bare `I` are unreachable by `[A-Z][a-z']+`, and a newline, colon, quote, paren or `- ` predecessor is skipped. I re-derived the newline case rather than trusting the test: `[ \t]+` cannot span `\n`, so the lookbehind lands on `\n` and fails.
- **I tried to prove the new suite never runs in CI and failed.** `discover_tests` in `capability-gate.py` only scans `q-system/.q-system/scripts/` and `q-system/.q-system/`, and no capability fragment names this file, so I expected 50 decorative tests. `plugins/kipi-core/voiceloop` is named in `.verify-suites`, and `verify.yml` runs `verify.sh --full` on every pull request. The suite is wired. Dropping the finding.
- **Offset handling is correct.** Rewriting from the end is right, and the comment explaining why a same-length replacement does not excuse it is the useful kind.
- **The plugin version is bumped** (1.13.9 to 1.13.11), which the PR-only version-bump guard requires.

## VERDICT

**APPROVE WITH NITS.** Five minors, no blocker, no major. Nothing here is permanent, nothing pages, and everything lands in the change log where the operator can see it.

Fix first: the dotted-brand case at `post_repair.py:160`. `Every.to` to `every.to` is the exact failure the module header calls worse than the bug, it needs one character in the regex (require a real word boundary rather than `(?![\w'])`), and it came from the repo's own prose rather than from a constructed input. The numbered-heading case is the same line and the same round.

One honest note on the PR body, since the body is part of what gets reviewed: "zero false positives across the corpus" is a true statement about one corpus on one date, and it reads in the body as a property of the rule. On 1.7M characters of this repo's own English the same detector is 4 for 4 wrong. That is the number I would put next to the zero.

FINDINGS:
minor|A dotted brand name clears all three constraints and is lowercased (`Every.to` to `every.to`), the exact failure the module header calls worse than the bug|plugins/kipi-core/voiceloop/post_repair.py:160
minor|Constraint 2's digit predecessor admits a numbered section prefix, so `## 2.1 The dedup key` loses its capital|plugins/kipi-core/voiceloop/post_repair.py:160
minor|The detector reads raw text while the gate reads the linter's prose view, so blockquoted third-party words and table headers get rewritten|plugins/kipi-core/voiceloop/post_repair.py:330
minor|The salutation class that removed `your` and `its` is still live for the other 21 words, and the test only parametrises the two removed ones|plugins/kipi-core/voiceloop/post_repair.py:148
minor|The 2b comment names step 3's proper-noun pass as its safety net, but `load_proper_nouns("")` returns an empty list in every tree|plugins/kipi-core/voiceloop/post_repair.py:441
END FINDINGS

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-395-20260920-114716.md` (6620 bytes).

```
FINDINGS:
minor|A comma or `and` before an in-set word defeats the title-case guard and lowercases real brand names like `Every` and `But`|plugins/kipi-core/voiceloop/post_repair.py:175
minor|The salutation fix removed `your` and `its` but the same reproducer still fires on `the`, `this`, `that` and `they`|plugins/kipi-core/voiceloop/post_repair.py:148
minor|`test_the_layer_runs_after_the_passes_that_CREATE_its_shape` stays green under the exact reorder its docstring says turns it red|plugins/kipi-core/voiceloop/tests/test_mid_sentence_caps.py:267
nit|Three test docstrings still name a digit predecessor the round-2 narrowing removed, and one says the call lives in a tuple it no longer lives in|plugins/kipi-core/voiceloop/tests/test_mid_sentence_caps.py:181
END FINDINGS
```

--- reviewer output ---

Review done. Four findings survived reproduction, all minor or nit. Here they are.

---

## MINOR 1: a comma or an `and` before the word defeats constraint 3, and a real brand gets lowercased

`plugins/kipi-core/voiceloop/post_repair.py:175`

The title-case-run guard only looks for `[ \t]+[A-Z]` immediately after the word. A comma, a paren, or a following lowercase connector defeats it. Combined with the comma predecessor in the lookbehind, any enumeration containing an in-set word loses its capital. `Every` is a real publication name and is in the set. The round-2 narrowing fixed `Every.to` and left bare `Every` open.

I found this by scanning this repo's own prose, not by constructing it. One hit in 1.46M chars, and it is a false positive.

Reproducer and real output:

```
scanned 382 markdown files, 1461269 chars
files with hits: 1   total hits: 1

But      plugins/kipi-core/skills/founder-voice/references/voice-dna.md
         ...esh subject. Top openers: And (198), I (101), So (97), But (69), Yeah (68), Like...
```

```
EDITED  'Top openers: And (198), I (101), So (97), but (69), Yeah (68)'
          ['lowercased mid-sentence capital (But) x1']
EDITED  'they moved off Etsy, Square and every onto one ledger'
          ['lowercased mid-sentence capital (Every) x1']
EDITED  'he reads Stratechery, every and Platformer every morning'
          ['lowercased mid-sentence capital (Every) x1']
EDITED  'the tools are Notion, Linear, every, and Slack'
          ['lowercased mid-sentence capital (Every) x1']
EDITED  'he writes for every, which pays per piece'
          ['lowercased mid-sentence capital (Every) x1']
```

The second line is the PR's own `LEGITIMATE` fixture with one more brand in it. The module header calls exactly this outcome worse than the bug. It stays minor because the change log names the word it ate, so the operator gets a receipt, and there is no caller of `repair()` in this tree.

---

## MINOR 2: the salutation fix removed two words, not the shape

`plugins/kipi-core/voiceloop/post_repair.py:148` and `tests/test_mid_sentence_caps.py:304`

Round 2 pulled `your` and `its` because "Hey man, Your invoice is attached" is a sentence start the linter's offsets cannot see. The same reproducer with `the`, `this`, `that` or `they` still fires, and `SALUTATIONS` only fixtures the two words that were removed.

Real output:

```
EDITED  'Hey man, the invoice is attached and the deck is done' ['lowercased mid-sentence capital (The) x1']
EDITED  'Hi Sarah, this is the deck you asked for'              ['lowercased mid-sentence capital (This) x1']
EDITED  'Hey dude, that fix landed last night'                  ['lowercased mid-sentence capital (That) x1']
EDITED  'Hey man, they shipped it already'                      ['lowercased mid-sentence capital (They) x1']
clean   'Hey man, Your invoice is attached and the deck is done' []
```

---

## MINOR 3: one ordering test cannot fail for the reason its docstring names

`tests/test_mid_sentence_caps.py:267`

The docstring says "a reorder that puts this layer first leaves the capital standing." I ran that exact reorder. The test stays green, because in `"is not That"` the predecessor of the gap is the lowercase `t` of `not`, so the mid-sentence pass catches it with or without the contraction pass running first.

Real output:

```
MUTANT: pass placed FIRST in group 0

REAL   out="The chart isn't that reliable in practice."
       ASSERT PASS
MUTANT out="The chart isn't that reliable in practice."
       ASSERT PASS
```

The ordering property is genuinely held by `test_the_pass_runs_after_the_sentence_start_pass`, which I verified separately (see below). This one test is decoration.

---

## NIT: three stale "a digit" references after the round-2 narrowing

`tests/test_mid_sentence_caps.py:181`

The digit predecessor was removed from the regex. The constraint-2 section header, `test_a_sentence_start_capital_is_never_touched`'s docstring, and `test_the_constraint_2_predecessors_that_DO_qualify` ("each of the three predecessors it admits" over a two-case loop) all still describe the old three. Separately, `tests/test_mid_sentence_caps.py:256` says to delete the function "from the tuple in `repair()`"; after the major fix it is a standalone call at step 2b, not a tuple member.

---

## What is sound

- **The ordering fix is really pinned.** I moved the pass back into group 0 and `test_the_pass_runs_after_the_sentence_start_pass` went red with exactly the false receipt the comment describes: out was `'It treats The chart as disposable. The fix is simple.'` while changes still claimed `'lowercased mid-sentence capital (The) x1'`. That is a test that can fail, for the reason it names.
- **Idempotency and the never-blocks rule.** 200 real prose files, two passes each: `non-idempotent=0, repair-made-gate-worse=0`. The repair never turns a clean gate dirty and never oscillates.
- **The rule really is narrow.** One false positive in 1.46M chars of this repo's own prose is a genuinely low rate for a closed-word-set rewriter.
- **Acronyms, bare `I`, sentence starts, code fences, frontmatter, markdown links, numbered headings, dotted names.** I tried all of them. `[A-Z][a-z']+` and the lookbehind hold.
- **Concurrency and 3am blast radius.** Pure function, no disk, no network, no caller in this tree. Two of these running at once cannot interfere.

---

Only minor and nit findings survived reproduction. No blocker, no major.

First thing to fix: the constraint-3 guard in `post_repair.py:175`. Widen the title-case-run skip so a following comma, paren, or `and` plus capital also counts as a run, or pull `every` and `but` out of the set. Right now the one class the module header calls worse than the bug is reachable from real text.

FINDINGS:
minor|A comma or `and` before an in-set word defeats the title-case guard and lowercases real brand names like `Every` and `But`|plugins/kipi-core/voiceloop/post_repair.py:175
minor|The salutation fix removed `your` and `its` but the same reproducer still fires on `the`, `this`, `that` and `they`|plugins/kipi-core/voiceloop/post_repair.py:148
minor|`test_the_layer_runs_after_the_passes_that_CREATE_its_shape` stays green under the exact reorder its docstring says turns it red|plugins/kipi-core/voiceloop/tests/test_mid_sentence_caps.py:267
nit|Three test docstrings still name a digit predecessor the round-2 narrowing removed, and one says the call lives in a tuple it no longer lives in|plugins/kipi-core/voiceloop/tests/test_mid_sentence_caps.py:181
END FINDINGS

assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-395-20260920-115439.md` (9116 bytes).

```
FINDINGS:
major|The new pass lowercases the final in-set word of a title-case markdown heading, a regression from base that fires on this repo's own prose|plugins/kipi-core/voiceloop/post_repair.py:177
minor|test_a_salutation_is_not_a_mid_clause asserts a property the code holds for only 2 of 18 set members; the/they/that fire on the identical salutation shape|plugins/kipi-core/voiceloop/tests/test_mid_sentence_caps.py:324
minor|The new comment cites step 3's proper-noun pass as this pass's safety net, but repair() calls load_proper_nouns("") which returns zero nouns for every caller|plugins/kipi-core/voiceloop/post_repair.py:456
END FINDINGS
```

--- reviewer output ---

## Finding 1 — major

**Claim:** the new pass lowercases the last word of a title-case markdown heading, a regression from base that fires on this repo's own prose.

**Where:** `plugins/kipi-core/voiceloop/post_repair.py:177` and `:355`. `_FOLLOWED_BY_CAPITAL_RE` only looks FORWARD, so constraint 3 protects a title-case run only when the in-set word is not the run's final word. "Recommend This" and "What Breaks Them" are unprotected.

**How I found it, not constructed:** ran the detector over all 961 tracked markdown files in the tree. Five hits, all five false positives, two of them this exact heading shape.

```
FILES SCANNED: 961
TOTAL HITS: 5
./plugins/kipi-core/skills/founder-voice/references/voice-dna.md:260  [But]  ...So (97), But (69), Yeah (68)...
./q-system/output/prd-claude-md-diet-2026-04-12.md:524  [When]  ...| Est. Tokens | Loads When |...
./q-system/output/prd-claude-md-diet-2026-04-12.md:545  [When]  ...| Est. Tokens | Loads When |...
./q-system/.q-system/onboarding/guides/connect-linkedin.md:61  [This]  ...## When to Recommend This...
./q-system/.q-system/onboarding/guides/connect-gamma.md:37  [This]  ...## When to Recommend This...
```

**Reproducer** (same input through `repair()` at HEAD and at base `057a3f0a`):

```
cd /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-395 && python3 -c '
import subprocess, types
from importlib import util as iu
T = "## When to Recommend This\n\nhe shipped it on friday and nobody noticed.\n\n## Why the Labels Move, and What Breaks Them\n\nthe chart rebuilds every cycle.\n"
def load(name, src):
    m = types.ModuleType(name); m.__file__ = "plugins/kipi-core/voiceloop/post_repair.py"
    exec(compile(src, name, "exec"), m.__dict__); return m
head = open("plugins/kipi-core/voiceloop/post_repair.py").read()
base = subprocess.run(["git","show","057a3f0a:plugins/kipi-core/voiceloop/post_repair.py"],capture_output=True,text=True).stdout
for label, src in (("HEAD", head), ("BASE", base)):
    m = load("pr_"+label, src); lint = m._load_linter("q-system/.q-system/scripts/voice-lint.py")
    out, ch = m.repair(T, allowlist=frozenset(), linter=lint, mapping={})
    print("=== %s ===" % label); print(out); print("CHG:", ch)
'
```

**Real output:**

```
=== HEAD ===
## When to Recommend this

He shipped it on friday and nobody noticed.

## Why The Labels Move, and What Breaks them

The chart rebuilds every cycle.

CHG: ["capitalized sentence start 'he'", "capitalized sentence start 'the'", "capitalized sentence start 'the'", 'lowercased mid-sentence capital (Them, This) x2']
=== BASE ===
## When to Recommend This

He shipped it on friday and nobody noticed.

## Why The Labels Move, and What Breaks Them

The chart rebuilds every cycle.

CHG: ["capitalized sentence start 'he'", "capitalized sentence start 'the'", "capitalized sentence start 'the'"]
```

**Why this rank.** The module's own header sets the bar: "A repair that lowercases one of those is worse than the bug." The PR narrowed twice already for this exact class, dotted names and `## 2.1 The dedup key`, and the ordinary title-case heading survived both narrowings. The PR's load-bearing claim is zero false positives; the first second corpus I could reach produces five, and zero of them are in the recorded twelve defects. A human has to spot and re-capitalize a heading in outbound copy.

## Finding 2 — minor

**Claim:** the salutation narrowing covers two words and the test that names it asserts a property the code does not hold.

**Where:** `plugins/kipi-core/voiceloop/tests/test_mid_sentence_caps.py:324`. `your` and `its` were removed for "Hey man, Your invoice is attached". Sixteen other set members hit the identical shape and the parameterized cases never try one.

**Reproducer output** (same command shape as above, `mid_sentence_cap_hits` plus `repair_mid_sentence_caps`):

```
IN : 'Hey man, Your invoice is attached and the deck is done'
HIT: []
IN : 'Hey man, The invoice is attached and the deck is done'
HIT: [(9, 'The')]
OUT: 'Hey man, the invoice is attached and the deck is done' ['lowercased mid-sentence capital (The) x1']
IN : 'Hey man, They asked for the numbers again'
OUT: 'Hey man, they asked for the numbers again' ['lowercased mid-sentence capital (They) x1']
IN : 'Hi Sana, That is the same bug as last week'
OUT: 'Hi Sana, that is the same bug as last week' ['lowercased mid-sentence capital (That) x1']
```

The test's name is `test_a_salutation_is_not_a_mid_clause`. The code holds "a salutation is not a mid-clause for two of eighteen words." Minor rather than major because a lowercase word after a salutation comma reads correctly in his register, so the output damage is near zero. The defect is the test's claim, not the text.

## Finding 3 — minor

**Claim:** the new comment cites step 3 as the safety net for this pass; step 3 loads zero proper nouns for every caller.

**Where:** `plugins/kipi-core/voiceloop/post_repair.py:456` says "Step 3's proper-noun pass still runs AFTER, so it remains the safety net if this layer ever lowercases the first word of a declared multi-word proper noun." Line 467 is `for noun in linter.load_proper_nouns(""):`, a hardcoded empty path.

```
cd /Users/assafkipnis/.config/kipi/review-trees/assafkip_kipi-system__pr-395 && python3 -c "
from importlib import util as iu
spec = iu.spec_from_file_location('post_repair','plugins/kipi-core/voiceloop/post_repair.py')
pr = iu.module_from_spec(spec); spec.loader.exec_module(pr)
lint = pr._load_linter('q-system/.q-system/scripts/voice-lint.py')
print('proper nouns (empty path):', len(lint.load_proper_nouns('')), lint.load_proper_nouns('')[:15])
"

proper nouns (empty path): 0 []
```

The hardcoded empty path predates this PR. What is new is a comment that leans on it to justify where the new pass sits.

## What is sound

Attacks I ran that the code survived, named so the findings above can be trusted:

- **Suite runs green and has a real caller.** 55 passed in 0.04s. `.verify-suites:71` names `plugins/kipi-core/voiceloop`, and `.github/workflows/verify.yml` runs `verify.sh --full`, so this is not an unwired suite. I expected to find one and did not.
- **Finding 1's own regression test is real.** `test_the_pass_runs_after_the_sentence_start_pass` measures correctly. I reproduced the group-0 ordering bug and confirmed the fix: `it treats The chart as disposable. the fix is simple.` gives `It treats the chart as disposable. The fix is simple.` with zero residual hits.
- **Offset safety under multiple hits.** Rewriting from the end holds. Two hits in one heading pair both landed correctly.
- **Idempotence.** `repair()` twice over its own output is a fixed point, measured `True`.
- **The narrowings bite.** `Every.to`, `Which.co.uk`, `## 2.1 The dedup key`, `see 3.4 That table`, markdown links `[The Verge](...)`, bold `**That**`, quoted speech, list markers, newline predecessors, inline code and fences all return no hit. The title-case guard holds in the forward direction: `The New York Times`, `The Home Depot`, `The Browser Company`, `Notes On Everything That Broke`.
- **Acronyms and bare I are structurally unreachable**, not allowlisted out. `[A-Z][a-z']+` does it.
- **Concurrency is a non-question.** Pure transform, no disk, no shared state.
- **Attack I could not land, and am dropping.** I tried to prove the change log still emits a false receipt after the fix. It does, on a sentence with twelve mid-clause `the`s: ten "capitalized sentence start" lines, nine reverted by the new pass, second sentence start still lowercase and correctly flagged by the gate. But base produces the same ten false lines plus nine corrupted capitals, so the new pass makes it strictly better and the input was constructed. Dropped.

## VERDICT: REQUEST CHANGES

One major, two minors. Not BLOCK: nothing here is permanent or unrecoverable, and the layer repairs rather than blocks, so no draft dies.

**Fix first:** finding 1. Constraint 3 needs the backward direction too. A capitalized word immediately BEFORE the in-set word is the same title-case evidence as one after it, and it costs one lookbehind. Adding it keeps every one of the ten caught defects (checked: all ten have a lowercase-word or comma predecessor by construction of constraint 2, so a preceding-capital test excludes none of them) and takes the repo-prose false-positive count from five to zero on the two heading hits and the table hit.

FINDINGS:
major|The new pass lowercases the final in-set word of a title-case markdown heading, a regression from base that fires on this repo's own prose|plugins/kipi-core/voiceloop/post_repair.py:177
minor|test_a_salutation_is_not_a_mid_clause asserts a property the code holds for only 2 of 18 set members; the/they/that fire on the identical salutation shape|plugins/kipi-core/voiceloop/tests/test_mid_sentence_caps.py:324
minor|The new comment cites step 3's proper-noun pass as this pass's safety net, but repair() calls load_proper_nouns("") which returns zero nouns for every caller|plugins/kipi-core/voiceloop/post_repair.py:456
END FINDINGS

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-395-20260920-124614.md` (5695 bytes).

```
FINDINGS:
minor|A comma-separated list of capitalized items loses a capital because both title-case guards only see the immediately adjacent token|plugins/kipi-core/voiceloop/post_repair.py:382
minor|test_a_salutation_is_not_a_mid_clause claims a general property that holds only for the two words removed; 12+ remaining set members fire on the identical shape|plugins/kipi-core/voiceloop/tests/test_mid_sentence_caps.py:324
END FINDINGS
```

--- reviewer output ---

## Review of PR #395

Read-only. 961 markdown files scanned, full `repair()` run over 955 of them, five guard mutations, 309-test suite run the way `verify.sh` runs it.

---

### FINDING 1 — minor: a comma-separated list of capitalized items loses a capital

**Claim:** both title-case guards only look at the immediately adjacent token, so a parenthetical between list members blinds them and a capitalized in-set word inside a list gets lowercased.

**File:** `plugins/kipi-core/voiceloop/post_repair.py:382`

**Reproducer:** found by scanning this repo's own prose, not constructed.

```
python3 -c "import sys; sys.path.insert(0,'plugins/kipi-core')
from voiceloop import post_repair as P
L=P._load_linter('q-system/.q-system/scripts/voice-lint.py')
t='Top openers: And (198), I (101), So (97), But (69), Yeah (68), Like (55).'
print(P.repair_mid_sentence_caps(t,L))"
```

**Real output:**

```
IN : Top openers: And (198), I (101), So (97), But (69), Yeah (68), Like (55).
OUT: Top openers: And (198), I (101), So (97), but (69), Yeah (68), Like (55).
RCP: ['lowercased mid-sentence capital (But) x1']
```

The producer is `plugins/kipi-core/skills/founder-voice/references/voice-dna.md`. Bounded: this was the ONLY hit in 4,717,768 characters across 961 files. It is a reference doc, not a draft, so the blast radius is one lowercase letter in prose nobody publishes.

---

### FINDING 2 — minor: the salutation test asserts a property the code does not have

**Claim:** `your` and `its` were removed from the set because a salutation comma is a sentence start the linter cannot see. 12 or more of the remaining 22 members fire in the identical shape, so the test's stated property holds only for the two words that were deleted.

**File:** `plugins/kipi-core/voiceloop/tests/test_mid_sentence_caps.py:324`

```
  That      -> LOWERCASED
  There     -> LOWERCASED
  Then      -> LOWERCASED
  These     -> LOWERCASED
  They      -> LOWERCASED
  Their     -> LOWERCASED
  This      -> LOWERCASED
  Each      -> LOWERCASED
  Another   -> LOWERCASED
  Because   -> LOWERCASED
  When      -> LOWERCASED
  Every     -> LOWERCASED
```

(input shape: `Hey man, <Word> invoice thing we talked about is done`)

Output text stays grammatical in every case, so the harm is the docstring, not the prose. The honest name is "two words that hit the salutation shape were removed", not "a salutation is not a mid-clause".

---

## What is sound

Attacks I ran that the code survived. Naming them because a review that only lists faults is not calibrated.

- **False-positive rate on real prose.** 4,717,768 chars, 961 files, ONE detector hit. That is the number the word-list scar asks for and it holds on a corpus the author did not pick.
- **Does the new pass turn the gate red?** Ran full `repair()` on 955 files twice, once with `repair_mid_sentence_caps` live and once stubbed to identity, then diffed `capitalization_violations`. Result: `2b INCREASES cap violations in: 0`.
- **False receipts.** The round-2 major was a change log claiming a repair the text did not keep. Checked every file whose receipt claimed a lowercasing, then re-ran the detector on the output. Result: `FALSE RECEIPT (claimed lowercase, still present after): 0`.
- **All five guards are load-bearing.** Killed each one in-process and watched the fixtures go red: backward guard killed, all four TITLE_CASE_TAILS fire; `_opens_a_sentence` forced False, run3-b drops from caught to missed; forward guard killed, both title-case runs fire; dotted lookahead dropped, `Every.to` fires; digit predecessor restored, `## 2.1 The dedup key` fires. No mutation survived.
- **Word-set calibration.** Re-measured against repo prose: current set 1 hit; adding `and` gives 2, adding `as` gives 3, adding `so` gives 2. The set is at a local minimum, and the removed words are cheap to keep out.
- **Non-idempotence, dropped.** `repair(repair(x)) != repair(x)` on 5 of 955 files, and the new pass looked responsible. Traced it: `mid_sentence_cap_hits(a)` is `[]` after pass one, so this layer IS a fixpoint for its own class. The second pass's edits come from step 2's `for _ in range(10)` bound, which is pre-existing and fixes at most ten sentence starts per call. Not this PR.
- **Test wiring, dropped.** Suspected the 61 new tests never run. `.verify-suites` names `plugins/kipi-core/voiceloop`. Ran it the way `verify.sh` does: `309 passed in 0.39s`.
- **Hyphen boundary, dropped.** `the The-Dude plugin` becomes `the the-Dude plugin`, same class as the dotted-name narrowing that was fixed. Zero instances of that shape in 961 files, so no producer emits it.
- **Bare publication names, dropped.** `he writes for Every these days` gets lowercased. Zero instances in the repo scan.

## VERDICT: APPROVE WITH NITS

Both findings are bounded, reproducible, and cost a lowercase letter in prose that is not a published draft. Nothing here is permanent or unrecoverable, and there is nothing a human has to do at 3am.

**Fix first:** Finding 2. It is a two-line docstring correction and it is the one that will mislead the next person, because the next person reading `test_a_salutation_is_not_a_mid_clause` will believe the shape is handled and add a word back.

FINDINGS:
minor|A comma-separated list of capitalized items loses a capital because both title-case guards only see the immediately adjacent token|plugins/kipi-core/voiceloop/post_repair.py:382
minor|test_a_salutation_is_not_a_mid_clause claims a general property that holds only for the two words removed; 12+ remaining set members fire on the identical shape|plugins/kipi-core/voiceloop/tests/test_mid_sentence_caps.py:324
END FINDINGS

PR #396, opened 2026-09-20
fix(update): the long path cannot start without the one-second answer

assafkip:
## The scar

A full `kipi-update.sh` run costs 20-25 minutes and reports only the **first**
refusal per instance. So anyone diagnosing "why will the fleet not sync" by
running it learns exactly one blocker per 25 minutes.

Measured across 2026-09-19 and 2026-09-20: **five runs**, each surfacing one new
blocker (dirty tree, then a skeleton one commit behind `origin/main`, then four
dangling symlinks). The fleet was unsynced at the end of both days.

`fleet-reach-audit.py` answers the same question for every instance at once,
read-only, in about a second. It was written **2026-08-14 (ASK-797/ASK-803)**
after this same pattern burned that rollout. Neither 2026-09 session opened it.

A memory file, a rule and a comment all existed. None of them fired. So the fix
is not more documentation, it is the invocation.

## What changed

`reach_preflight()` runs before the instance loop:

- **prints the verdict every run**, including a clean one. A preflight that
  speaks only when unhappy cannot be told from one that did not run.
- **refuses when any instance would not sync**, in both modes, naming every
  blocker and the `fleet-unblock.py` remedy. Hatch `--skip-reach-preflight` is
  named in the refusal, so one permanently-stuck instance cannot wedge the other 24.
- **aborts when the audit is missing or mute**, matching how this file already
  treats the propagation leak gate. A gate that cannot run must not pass.
- **disarms with an announcement** when there is no `instance-registry.json` —
  which is every `kipi-update` test fixture.

## Verification

`bash test-kipi-update-reach-preflight.sh` — 12 checks, all green, exit 0 read
without a pipe.

Case F mutates out the call site, asserts the mutant **applied** by byte count
(149636 -> 149620), and confirms the refusal stops firing without it.

Only `--dry-run` is invoked, against throwaway skeletons and instances.


assafkip:
## Verdict: REQUEST CHANGES

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-396-20260920-134726.md` (6086 bytes).

```
FINDINGS:
major|An unclearable BLOCKED-FLEET instance aborts every full-fleet run while the remedy the abort names refuses the path and exits 0|kipi-update.sh:1752
minor|MISSING and NOT-A-REPO verdicts count against the reach denominator but are never named in the output|kipi-update.sh:1731
END FINDINGS
```

--- reviewer output ---

# Review: PR #396 — reach preflight in `kipi-update.sh`

Two findings survived reproduction. One major, one minor.

---

## MAJOR — an unclearable `BLOCKED-FLEET` instance wedges every full-fleet run, and the remedy the ABORT prints is a no-op that exits 0

**File:** `kipi-update.sh:1752` (the remedy line), refusal decided at `kipi-update.sh:1739`

**Claim.** `classify()` in `fleet-reach-audit.py:189` returns `fleet-written` when the skeleton wrote **either** the index blob **or** the worktree blob. A staged skeleton write with a founder edit on top therefore yields kind `fleet-written` only, so the verdict is `BLOCKED-FLEET`, so the preflight aborts the whole run. But `fleet-unblock.py:183` **refuses** that exact shape by design. The remedy cannot clear it, and `fleet-unblock.py:473` treats `refused` as success and exits 0.

Round 1 of this PR fixed this shape for `BLOCKED-FOUNDER`. The same shape survives under `BLOCKED-FLEET`.

**Reproducer** (real `fleet-reach-audit.py` and real `fleet-unblock.py`, throwaway repos under `tempfile.mkdtemp()`, nothing in the repo touched):

```
python3 -c '<builds skeleton with q-system/foo at V1 then V2;
            instance HEAD=V1, index=V2 (skeleton blob), worktree="FOUNDER EDIT";
            runs the real audit, the real unblocker --apply, then re-measures>'
```

**Real output:**

```
=== 1. REAL fleet-reach-audit.py --json ===
verdict: BLOCKED-FLEET  kinds: ['fleet-written']
=== 2. the remedy the ABORT prints: fleet-unblock.py --apply ===
MODE: APPLY

inst  [BLOCKED-FLEET]
    REFUSE  q-system/foo
              index holds skeleton blob 55bee9c205a5 but the worktree holds f7e427497d93, which the skeleton never wrote; committing the index would leave that edit for the next sync to overwrite

acted on 0 path(s); refused 1 path(s)

fleet-unblock exit code: 0
=== 3. re-measure, exactly as the ABORT text instructs ===
verdict AFTER the remedy: BLOCKED-FLEET
```

Case B of the PR's own harness already proves `BLOCKED-FLEET` reaches `ABORT`, so the loop closes: abort, run the remedy, remedy reports success while changing nothing, re-measure, abort again.

**What a human does about it.** The only escape the refusal offers is `--skip-reach-preflight`, which is the gate off. A founder can also clear it by committing their edit, but the refusal text never says so: the `BLOCKED-FOUNDER` branch prints "correctly refused until committed" and lets the run proceed, while this branch hard-aborts and names a tool that will not act. On a 25-instance fleet one such path stops all 25.

Recoverable, so major and not blocker.

---

## MINOR — `MISSING` and `NOT-A-REPO` count against the denominator and are never named

**File:** `kipi-update.sh:1731`

`audit_instance()` emits five verdicts. The preflight buckets three. `MISSING` and `NOT-A-REPO` land in `len(rows)` and in no print loop, so the liveness line the whole design rests on under-reports reach with no explanation attached.

**Reproducer** (extracts the real heredoc out of `kipi-update.sh` and feeds it real-shaped rows):

```
python3 -c '<reads the PYEOF body from kipi-update.sh, runs it on
            MISSING + NOT-A-REPO + WOULD-SYNC rows>'
```

**Real output:**

```
--- REAL heredoc extracted from kipi-update.sh, 930 bytes ---
reach preflight: 1 of 3 would sync now
exit: 0 (0 = preflight lets the 25-minute run proceed)
```

Two instances will not sync, neither is named. Bounded: the live registry has zero such entries today (checked, 25 entries, 0 missing paths, 0 non-repos), and the updater handles a missing path with `SKIP:` rather than a failure. It is a wrong number, not a wrong action.

---

- **Timing claim.** PR says "about a second". `time python3 fleet-reach-audit.py --json` on the live 25-entry registry: **0.614s total**. Holds.
- **Fixture shape.** The stub at `test-kipi-update-reach-preflight.sh:100` emits a list of `{name, path, prefix, blocked_by[{status,path,staged,kind}], verdict}`. That is byte-for-byte the shape `audit_instance()` builds. Not an invented fixture.
- **The mutation case is real.** Case F asserts the mutant applied by byte count before testing it, and I watched it fire: `mutant applied (152443 -> 152427 bytes)` then `the mutant is killed`. Not decoration.
- **Ordering against the skeleton-freshness gate.** `reach_preflight` is called at line 1763, after the `origin/$SKELETON_BRANCH` equality abort at lines 739-797. The audit's `SkeletonBlobs` therefore reads a HEAD already proven current. I looked for a stale-HEAD misclassification and there is none.
- **`set -e` interaction.** The `|| rc=$?` reasoning in the comment is correct and I could not break it: a python exit 2 reaches the refusal rather than killing the script.
- **CI.** `validate` pass, `verify` pass, `reviewer-floor` pass. `test-propagation-entrypoints.py` exit 0, `test_update_propagation.py` 6 passed.
- **The new harness is not in the capability manifest** — and neither is any other root-level `test-*.sh` (0 of 278 declarations). Pre-existing convention, not introduced here. Dropped.
- **`$UPDATE_ARGV` word-splitting** on an argument containing a space: I could not produce an input any real caller emits (instance names in the registry have none). Dropped per the construct-your-own-input rule.

---

## VERDICT: REQUEST CHANGES

A major finding, recoverable, so changes requested rather than a block.

**Fix first:** the `BLOCKED-FLEET` refusal at `kipi-update.sh:1739`. Refuse only on rows `fleet-unblock.py` would actually act on, or drop the abort to a named warning for rows it refuses. Shipping as-is trains the operator to type `--skip-reach-preflight` on every run, which is the gate off, which is the outcome round 1 of this PR was written to prevent.

FINDINGS:
major|An unclearable BLOCKED-FLEET instance aborts every full-fleet run while the remedy the abort names refuses the path and exits 0|kipi-update.sh:1752
minor|MISSING and NOT-A-REPO verdicts count against the reach denominator but are never named in the output|kipi-update.sh:1731
END FINDINGS

assafkip:
## Verdict: APPROVE WITH NITS

Reviewer engine: `claude`. Full review on disk: `/Users/assafkipnis/.config/kipi/pr-reviews/assafkip_kipi-system__pr-396-20260920-140303.md` (5805 bytes).

```
FINDINGS:
minor|The every-run "N of N would sync now" line claims reach coverage the audit does not have, and the caveat prints only inside the abort branch|kipi-update.sh:1738
minor|The audit's stderr is discarded by 2>/dev/null, so a broken audit aborts the whole fleet sync with an exit code and no cause|kipi-update.sh:1694
minor|The comment claims the gate is pinned by a harness that no workflow, hook, or capability manifest entry runs|kipi-update.sh:1672
END FINDINGS
```

--- reviewer output ---

# Review: PR #396, the reach preflight

Three findings, all minor. Every one reproduced against a real fixture; the two design attacks I expected to land did not.

---

## Finding 1 (minor): "N of N would sync now" claims coverage the audit does not have

`kipi-update.sh:1738` prints the headline on every run. It is derived from one refusal class only, the dirty tree inside the synced pathspec. The honest caveat ("does NOT cover a stale skeleton or a dangling symlink") lives at `kipi-update.sh:1751`, inside the ABORT branch, so it prints only when the run is already refused. On the path the operator actually takes, the line reads as a prediction that the run will succeed.

Reproducer: fixture skeleton plus one instance carrying a dangling escaping symlink under `q-system/`, audit stub reporting WOULD-SYNC (a true answer to its own question).

Real output:

```
reach preflight: 1 of 1 would sync now
--- clean (subtree) ---
unsafe dangling symlink escapes the disposable model: q-system/dangling -> /var/.../r1/gone/x
  ERROR: unsafe symlink prevents isolated dry-run modeling
  Updated: 0
  Failed:  1
```

That is blocker class three out of the three named in this PR's own scar text. The wording fix is one line: say what was measured, for example "reach preflight: 1 of 1 clear of dirty-tree blockers (stale skeleton and symlinks are separate gates)".

## Finding 2 (minor): the audit's own diagnosis is discarded, and the fleet sync stops with a bare exit code

`kipi-update.sh:1694` runs the audit with `2>/dev/null`. `fleet-reach-audit.py` raises a written-for-this-moment RuntimeError when it cannot parse `INSTANCE_OWNED_SUBTREES` out of `kipi-update.sh`. The preflight then aborts the entire run and prints the exit number only.

Reproducer: a cosmetic edit to the fixture updater, a trailing comment on the array opener. The updater itself is unaffected; the audit's regex is not.

Real output, both halves:

```
--- what the audit says when run by hand ---
RuntimeError: INSTANCE_OWNED_SUBTREES not found in /private/var/.../skeleton/kipi-update.sh; the audit cannot build the guard's pathspec and refuses to guess it

--- what the operator sees from the updater ---
ABORT: the reach preflight produced no verdict (exit 1).
A preflight that cannot run must not pass. Fix it, or pass
--skip-reach-preflight to proceed without it.
```

"Fix it" with the cause deleted. Dropping `2>/dev/null`, or capturing stderr and echoing it under the ABORT, restores the message somebody already wrote for this case.

## Finding 3 (minor): nothing runs the only test for this gate

`kipi-update.sh:1672` says "Pinned by test-kipi-update-reach-preflight.sh". The harness passes by hand, 12 checks, exit 0, mutation case F kills its mutant. No automated caller exists:

```
grep -rn "test-kipi-update" lefthook.yml q-system/.q-system/verify.sh .github/workflows/
  -> no match for the new harness
grep -rln "test-kipi-update-reach-preflight" .
  -> kipi-update.sh (a comment), the harness itself, test-propagation-entrypoints.py (an exemption)
```

CI runs pytest directories named in `.verify-suites`. Thirteen sibling update harnesses live under `q-system/.q-system/scripts/test/` with a `capability/expected_tests/*.json` entry each; this one sits at repo root with neither. The root-level harnesses are all unregistered today, so this is the repo's existing posture rather than a regression, but a registered lane exists and the new gate is outside it.

---

## What is sound

**Fixture shape matches the real producer.** The stub emits name, path, prefix, blocked_by with status/path/staged/kind, verdict. I ran the real `fleet-reach-audit.py --json` and it emits exactly that. No invented-key mutex here.

**No false alarm on the live fleet.** Real audit against the real registry: 24 rows, 23 WOULD-SYNC, 1 BLOCKED-FOUNDER (travel-agent). Founder work prints as information and does not refuse, so today's every-invocation behavior is one extra line, not a wedge.

**The preflight does not wedge the updater's own suite.** The three existing harnesses that copy the real `*.py` into their fixture and write a registry run the armed preflight for real. `test-kipi-update-dirty-guard-scope.sh`, `test-kipi-update-bash32-empty-array.sh` and `test-kipi-update-cache-exclusion.sh` all still print PASS.

**Audit set and sync set agree.** I checked all 25 registry entries for a row the audit includes but the updater would skip (standalone, null prefix, skeleton_managed false). Zero mismatches, so no run can be refused over an instance it would never touch.

**Cost claim holds.** `python3 fleet-reach-audit.py --json` takes 0.55s wall on the real fleet, against the stated 20 to 25 minutes.

**Rollback and concurrency are clean.** The abort fires before the instance loop and before any mutation, the EXIT trap is already installed, and every audit call is a read, so two updaters racing changes nothing.

**The `|| rc=$?` reasoning in the comment is true.** Case D (audit present, exits 3, silent) aborts rather than passing, which is the branch the comment claims.

---

Fix first: finding 1. The headline line is printed on every run and is the one sentence an operator will trust at 3am; it currently promises more than the audit measured.

FINDINGS:
minor|The every-run "N of N would sync now" line claims reach coverage the audit does not have, and the caveat prints only inside the abort branch|kipi-update.sh:1738
minor|The audit's stderr is discarded by 2>/dev/null, so a broken audit aborts the whole fleet sync with an exit code and no cause|kipi-update.sh:1694
minor|The comment claims the gate is pinned by a harness that no workflow, hook, or capability manifest entry runs|kipi-update.sh:1672
END FINDINGS

